PR #15048, opened 2020-11-05
[Fileshare] Share enable protocol share squash root

tasherif-msft:


sdk/storage/azure-storage-file-share/azure/storage/fileshare/_models.py
```diff
@@ -332,6 +334,10 @@ class ShareProperties(DictMixin):
     :ivar int remaining_retention_days:
         To indicate how many remaining days the deleted share will be kept.
         This is a service returned value, and the value will be set when list shared including deleted ones.
+    :ivar ~azure.storage.fileshare.models.ShareRootSquash or str root_squash:
+        Possible values include: 'NoRootSquash', 'RootSquash', 'AllSquash'.
+    :ivar ~azure.storage.fileshare.models.ShareProtocols or str protocols:
```
xiafu-msft, line 339: type should be a list

sdk/storage/azure-storage-file-share/azure/storage/fileshare/_share_client.py
```diff
@@ -314,6 +316,13 @@ def create_share(self, **kwargs):
 
         :keyword int timeout:
             The timeout parameter is expressed in seconds.
+        :keyword protocols:
+            Protocols to enable on the share. Only one protocol can be enabled on the share.
+        :paramtype protocols: str or ~azure.storage.fileshare.ShareProtocols
```
xiafu-msft, line 321: type should be a list?

tasherif-msft: No it shouldn't, as discussed with @annatisch, we will opt for using a string since users are allowed to use only one protocol at the moment. It doesn't make sense to have them pass ['SMB'] or ['NFS']. As Anna recommended we will stick to string for now and then in the future we can simply change the type to `Union[str, [protocols]]`. However, it is important to note that we are in fact deserializing the enabled protocols on a share into a list, i.e in `ShareProperties` (as discussed in the archboard meeting)

sdk/storage/azure-storage-file-share/azure/storage/fileshare/_models.py
tasherif-msft: It shouldn't be a list since the service returns a string of enabled protocols and we will deserialize that into a list for the user.

tasherif-msft: It shouldn't since the service returns a string of protocols and deserialize it into a list in `_models.py` here https://github.com/Azure/azure-sdk-for-python/blob/4592d8537d5d07d6ea667b6530f2b2c28993d30e/sdk/storage/azure-storage-file-share/azure/storage/fileshare/_models.py#L360

sdk/storage/azure-storage-file-share/azure/storage/fileshare/_share_client.py
xiafu-msft: as long as you have a plan to make it a list in the future I'm good with it!

xiafu-msft (approved):

PR #15058, opened 2020-11-06
Nightly publish to dev feed should run regardless of test outcome

scbedd:
We should always run integration to publish from our artifacts. If the build pipeline crashed too early in the pipeline, this will fail regardless, due to inability to pull the artifacts down.

While this doesn't address the chicken/egg problem caused from checking in a version bump of azure-core + a dependency version bump in the same day, it will definitely enable the pipeline to self-mitigate that problem. It'll fail for a night, then resume normal operations the next day.


weshaggard (approved):
I'm OK with this change although it is a little odd so it is probably worth adding a comment about why this condition is this way.

Issue #14645, opened 2020-10-20
[ServiceBus] Fix sphinx warnings when generating docs


yunhaoling:
When generating the docs locally, sphinx would output multiple warnings complaining that some sample code snippet misplacement or wrong type.

we should check the sphinx output and try the best to resolve them.

PR #15059, opened 2020-11-06
[ServiceBus] Sphinx/Docstring/Code snippets Fix

yunhaoling:
addressing issue: https://github.com/azure/azure-sdk-for-python/issues/14645

### DIFF
--- a/sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_client.py
+++ b/sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_client.py
@@ -224,9 +224,9 @@ class ServiceBusClient(object):
         :paramtype receive_mode: ~azure.servicebus.ReceiveMode
         :keyword Optional[float] max_wait_time: The timeout in seconds between received messages after which the
          receiver will automatically stop receiving. The default value is None, meaning no timeout.
-        :keyword Optional[AutoLockRenewer] auto_lock_renewer: An AutoLockRenewer can be provided such that messages are
-         automatically registered on receipt.  If the receiver is a session receiver, it will apply to the session
-         instead.
+        :keyword Optional[~azure.servicebus.AutoLockRenewer] auto_lock_renewer: An ~azure.servicebus.AutoLockRenewer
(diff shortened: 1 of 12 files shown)


sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_client.py
```diff
@@ -224,9 +224,9 @@ def get_queue_receiver(self, queue_name, **kwargs):
         :paramtype receive_mode: ~azure.servicebus.ReceiveMode
         :keyword Optional[float] max_wait_time: The timeout in seconds between received messages after which the
          receiver will automatically stop receiving. The default value is None, meaning no timeout.
-        :keyword Optional[AutoLockRenewer] auto_lock_renewer: An AutoLockRenewer can be provided such that messages are
-         automatically registered on receipt.  If the receiver is a session receiver, it will apply to the session
-         instead.
+        :keyword Optional[~azure.servicebus.AutoLockRenewer] auto_lock_renewer: An ~azure.servicebus.AutoLockRenewer
```
KieranBrantnerMagee, line 227: So for my own understanding; does a local typename _never_ work?  Because I swear we used type-aliases elsewhere in these docstrings, and am not sure how we'd work around that otherwise.

KieranBrantnerMagee (approved):

sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_client.py
yunhaoling: it's only the AutoLockRenewer that sphinx is warning "duplicate reference found".
usually we use the pattern:
```
:keyword xxx:
:paramtype xxx: ~azure.servicebus.xxx
```
which contains the namespace.

so just `AutoLockRenewer` alone makes sphinx confused.

KieranBrantnerMagee: Thanks for the clarity. That's silly.  Have to assume it's something to do with aio but would expect to see it more broadly.  Oh well, at least I get the gist of what's going on now, this all seems totally reasonable.

Issue #14739, opened 2020-10-23
[ServiceBus] Session Receiver should not retry (reconnect) in cases of connection issues


yunhaoling:
When the sessionful receiver failed an operation due to connection/link issues, reconnect to that session is different from normal entity and it may trigger other session errors as Josh mentioned:

> The issue is that when you attempt to reconnect, the user has to handle more potential exceptions - instead of just SessionLockLost, they also have to handle SessionCannotBeLocked and ServiceTimeout. 

we'd rather let user handle session error by their best judgement.
and this is cross-language alignment, all languages all other language decide not to reconnect in cases of session connection issues.

PR #15060, opened 2020-11-06
[ServiceBus] Do not retry on connection issue when it's a sessionful receiver

yunhaoling:
Addressing issue: https://github.com/Azure/azure-sdk-for-python/issues/14739

### DIFF
--- a/sdk/servicebus/azure-servicebus/azure/servicebus/exceptions.py
+++ b/sdk/servicebus/azure-servicebus/azure/servicebus/exceptions.py
@@ -122,6 +122,12 @@ def _handle_amqp_connection_error(logger, exception, handler):
         error = ServiceBusConnectionError(message, error=exception)
         error_need_raise, error_need_close_handler = True, False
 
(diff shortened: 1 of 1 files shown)


KieranBrantnerMagee (approved):
LGTM; I've expressed my lukewarm-ness regarding this semantic to you OOB but putting on the record that if we get user feedback that the differing session behavior is a pitfall more than a helpful guardrail I'd likely use that as an opportunity to push back on the powers that be WRT this normalization. (That's not on you, just grumbling out loud :P)

yunhaoling:
> LGTM; I've expressed my lukewarm-ness regarding this semantic to you OOB but putting on the record that if we get user feedback that the differing session behavior is a pitfall more than a helpful guardrail I'd likely use that as an opportunity to push back on the powers that be WRT this normalization. (That's not on you, just grumbling out loud :P)

we'll gather feedback from users and can simply remove these 6 lines of code to add retry for session back :)

PR #15083, opened 2020-11-06
throws an error on invalid types when creating an entity

seankane-msft:


sdk/tables/azure-data-tables/azure/data/tables/_serialize.py
```diff
@@ -207,6 +207,9 @@ def _add_entity_properties(source):
 
     properties = {}
 
+    # if the entity is one from the service, remove _metadata and proceed
+    source.pop("_metadata", None)
```
annatisch, line 211: I'm not sure we should pop - as this will remove it from the customers copy of the entity as well.
Unless we're already copying the underlying dictionary?
Otherwise we _should_ be able to do something like:
```python
to_send = dict(source)  # this should do a shallow copy
to_send.pop('_metadata', None)  # Remove metadata from our copy
```
```

annatisch (approved):

PR #15089, opened 2020-11-06
[ServiceBus] Migration guide update

yunhaoling:


KieranBrantnerMagee (approved):
LGTM.  Echoing our OOB, but we may want to make more precise sample long-form links in the future or if prompted by user feedback, but for now since your examples are somewhat self contained this should be sufficient.

PR #15090, opened 2020-11-06
[Storage][STG75]Lease on Get/Set Tags & Find Blob by Tags

xiafu-msft:


sdk/storage/azure-storage-blob/tests/test_blob_tags.py
```diff
@@ -103,6 +103,29 @@ def test_set_blob_tags(self, resource_group, location, storage_account, storage_
...
+        blob_client, _ = self._create_block_blob()
+        lease = blob_client.acquire_lease()
+
+        # Act
+        blob_tags = {"tag1": "firsttag", "tag2": "secondtag", "tag3": "thirdtag"}
+        with self.assertRaises(HttpResponseError):
+            blob_client.set_blob_tags(blob_tags)
+        blob_client.set_blob_tags(blob_tags, lease=lease)
+
+        blob_client.get_blob_tags()
```
tasherif-msft, line 119: Is this supposed to work? the blob client is leased yet it doesn't need it? I assume read operations don't need the lease? (fails if a wrong lease is given instead)

sdk/storage/azure-storage-blob/tests/test_blob_tags.py
```diff
@@ -276,8 +299,8 @@ def test_list_blobs_returns_tags(self, resource_group, location, storage_account
             for key, value in blob.tags.items():
                 self.assertEqual(tags[key], value)
 
-    @GlobalResourceGroupPreparer()
-    @StorageAccountPreparer(random_name_enabled=True, location="canadacentral", name_prefix='pytagstorage')
```
tasherif-msft, line 280: Why are we removing live testing from `canadacentral`?

tasherif-msft:
LGTM besides the two questions I had :)

sdk/storage/azure-storage-blob/tests/test_blob_tags.py
tasherif-msft: Resolved offline.

sdk/storage/azure-storage-blob/tests/test_blob_tags.py
tasherif-msft: Resolved.

tasherif-msft (approved):

Issue #14990, opened 2020-11-03
Function kwargs not rendering on docs.ms.com


scbedd:
Boils down to during processing within the `sphinx-docfx-yaml` extension, we're seeing the following earning when we enable debug output:

```
Can't get argspec for <class 'function'>: azure.mgmt.storage.v2015_06_15.aio.operations.StorageAccountsOperations.begin_create
Can't get argspec for <class 'function'>: azure.mgmt.storage.v2015_06_15.aio.operations.StorageAccountsOperations.check_name_availability
```

I know it's firing from `extension.py`, but I don't have a root cause as to WHY this is yet.

Note that `class` kwargs aren't having this same problem.

scbedd:
After spending a bunch more time with this, I believe this issue is at least twofold. Context here is that Johan raised this as a general issue. It both IS and ISN'T a general issue. We _definitely do_ document some function keyword arguments. The problem is we don't seem to be doing it consistently. There also isn't a great way to programatically pull this detail out of the rendered docs. 

For the specific issue that Johan raised, like [here](https://review.docs.microsoft.com/en-us/python/api/azure-storage-blob/azure.storage.blob.aio.blobclient?view=azure-python&branch=smoke-test#download-blob-offset-none--length-none----kwargs-)

You'll notice that SOME keyword params are rendered. And NONE of the `param` types are. That's because this doc string is broken _in general_.

The `sphinx` itself not being able to parse the doc string for azure-storage-blob. See our own [github.io.](https://azuresdkdocs.blob.core.windows.net/$web/python/azure-storage-blob/12.5.0/azure.storage.blob.html#azure.storage.blob.ContainerClient.download_blob)

![image](https://user-images.githubusercontent.com/45376673/98615596-e1550a00-22af-11eb-8afa-c21d1c206582.png)

I am going to try a couple local sphinx runs. I'm thinking the easy answer will be to replace the `"` with `'`, but I'm digging and will submit a PR that resolves this on the packages that I can find that do this.

As for the "only some" function keyword params, I do have [another PR](https://github.com/docascode/sphinx-docfx-yaml/pull/102) that addresses other issues we've found as part of this investigation.

PR #15179, opened 2020-11-10
Resolve missing storage kwparams

scbedd:
Turns out this _isn't_ a holistic problem with function keyword parameters. See #14990 for further discussion.

Before the changes included in this changeset. This is what the _current_ sphinx docs look like for `BlobClient.download_blob`

image

Notice how the keyword params are broken up? Everything after the `if_tags_match_condition` is rendered as ANOTHER SET of params. This is because the `\"` that we were using was seriously messing with our sphinx output. 

image

@xiafu-msft I just did a find replace to easily fix the docs (and so you can see the difference in the generated `documentation` artifacts). I welcome adjustments as long as they properly render in standard sphinx output.


tasherif-msft:
This has been resolved by this https://github.com/Azure/azure-sdk-for-python/pull/15229#event-3995863851, correct?
If so, we can close this PR! :)

scbedd:
@tasherif-msft 100%. Thanks!

PR #15195, opened 2020-11-10
[T1] avs 2020 11 11

msyyc:
https://github.com/Azure/sdk-release-request/issues/956
--


sdk/compute/azure-mgmt-avs/azure/mgmt/avs/_avs_client.py
```diff
@@ -56,7 +62,7 @@ def __init__(
         super(AVSClient, self).__init__(self.config.credentials, self.config)
 
         client_models = {k: v for k, v in models.__dict__.items() if isinstance(v, type)}
-        self.api_version = '2020-03-20'
+        self.api_version = '2020-07-17-preview'
```
ctaggart, line 65: It would still be using the 2020-03-20 API.

ctaggart (changes requested):
It should be using the GA API, which is 2020-03-20.

Issue #15216, opened 2020-11-10
[Storage] process_storage_error should use raise_with_traceback


annatisch:
The `process_storage_error` (currently in the `_shared.response_handlers` module in all of the various Storage SDKs) should add raise_with_traceback to avoid losing the original error context.

https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/storage/azure-storage-blob/azure/storage/blob/_shared/response_handlers.py#L147

This line should be updated to:
```python
raise error.raise_with_traceback()
```
This method should always be available, as all of the constructed error objects are flavours of AzureError.

PR #15263, opened 2020-11-11
Added raise_with_traceback to storage errors

tasherif-msft:
This resolves https://github.com/Azure/azure-sdk-for-python/issues/15216.


sdk/storage/azure-storage-blob/azure/storage/blob/_shared/response_handlers.py
```diff
@@ -144,7 +144,7 @@ def process_storage_error(storage_error):
     error = raise_error(message=error_message, response=storage_error.response)
     error.error_code = error_code
     error.additional_info = additional_data
-    raise error
+    raise error.raise_with_traceback()
```
annatisch, line 147: I thought `raise_with_traceback` already called `raise` - does it not?

tasherif-msft: you are absolutely correct

xiafu-msft (approved):

PR #15314, opened 2020-11-13
[ServiceBus] Message body type validation

KieranBrantnerMagee:
Add validation such that Message construction fails fast if body is not a valid type.  Add changelog notes and a test for this scenario as well.


sdk/servicebus/azure-servicebus/azure/servicebus/_common/message.py
```diff
@@ -75,7 +77,7 @@ class ServiceBusMessage(object):  # pylint: disable=too-many-public-methods,too-
     """
 
     def __init__(self, body, **kwargs):
-        # type: (Union[str, bytes], Any) -> None
+        # type: (Optional[Union[str, bytes]], Any) -> None
```
rakshith91, line 80: do we want to make it optional? I know we should allow none, but thought we dont really want to document it.

rakshith91 (approved):

sdk/servicebus/azure-servicebus/azure/servicebus/_common/message.py
KieranBrantnerMagee: If there was a way to not make MyPy complain from passing None without this type, I'd agree, but I don't want to make life harder for folks who have that and want to pass None because it's not technically incorrect.  Since it's a type signature I don't believe we're tied to backcompat too, if we end up wanting to change this/it causes confusion (as the underlying behavior won't change), but I'll make sure we have the flexibility to reassess this before I merge it.

sdk/servicebus/azure-servicebus/CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 **Breaking Changes**
 
 * `ServiceBusSender` and `ServiceBusReceiver` are no more reusable and will raise `ValueError` when trying to operate on a closed handler.
+* `ServiceBusMessage` will now raise a `TypeError` when provided an invalid body type.  Valid bodies are strings, bytes, None, or a list of string or bytes items.
```
yunhaoling, line 8: I'd rather not mention we support "a list of string or bytes items", because our current uamqp implementation would concatenate them into a single string which I think is a bit unexpected behavior.

(Besides we had discussion on supporting AmqpValue before, in which case a list of string/bytes should be a list not concatenation.)

sdk/servicebus/azure-servicebus/azure/servicebus/_common/message.py
```diff
@@ -109,11 +111,20 @@ def __str__(self):
         return str(self.message)
 
     def _build_message(self, body):
+        def _validate_body_type(body, allow_none=False):
+            if not (isinstance(body, (six.string_types, six.binary_type)) or (body is None and allow_none)):
+                raise TypeError("ServiceBusMessage body must be a string, bytes, None, or a list of string or bytes "
```
yunhaoling, line 116: same comment as the changelog one..
not mentioning "or a list of string or bytes element" would be my preference.

sdk/servicebus/azure-servicebus/azure/servicebus/_common/message.py
```diff
@@ -109,11 +111,20 @@ def __str__(self):
...
+                                "elements. Got instead: {}".format(type(body) if allow_none else "[... {} ...]"
+                                                                                                 .format(type(body))))
+
         if isinstance(body, list) and body:  # TODO: This only works for a list of bytes/strings
+            _validate_body_type(body[0])
             self.message = uamqp.Message(body[0], properties=self._amqp_properties, header=self._amqp_header)
             for more in body[1:]:
+                _validate_body_type(more)
                 self.message._body.append(more)  # pylint: disable=protected-access
```
yunhaoling, line 125: I tend to just remove the "if isinstance(body, list) and body" check part based on what I mentioned above

sdk/servicebus/azure-servicebus/azure/servicebus/_common/message.py
KieranBrantnerMagee: As above, agreed, good call.

sdk/servicebus/azure-servicebus/azure/servicebus/_common/message.py
KieranBrantnerMagee: ++ I've reached out to Josh to see what we're normalizing on here.

KieranBrantnerMagee: The answer is "don't take a list"; that may be a weirdness historical to our UAMQP.  This is a nice alignment with what the user studies found, that some folks got tripped up by this when they tried to send a list of messages in this fashion and "it worked" (but not as they expected)

sdk/servicebus/azure-servicebus/azure/servicebus/_common/message.py
KieranBrantnerMagee: (confirmed that we can be flexible with this)

yunhaoling (approved):

yunhaoling (approved):

PR #15317, opened 2020-11-13
[Communication] Update version for Preview 3 release

ankitarorabit:


sdk/communication/azure-communication-sms/CHANGELOG.md
```diff
@@ -1,7 +1,7 @@
 # Release History
 
-## 1.0.0b4 (Unreleased)
-
+## 1.0.0b4 (2020-11-16)
+- Updated `azure-communication-sms` version.
```
DominikMe, line 4: Instead of the updated version comment here and in chat, we could mention that we changed or fixed the azure-core dependency to <2.0.0,>=1.6.0.

Your call which message you think makes more sense.

DominikMe (approved):
Thanks for making the updates!

One optional comment to consider if you like.

sdk/communication/azure-communication-sms/CHANGELOG.md
ankitarorabit: I thought about this and talked to @turalf but his opinion is that this is more like and internal bug fix which we don't really need to mention in the changelog

PR #15324, opened 2020-11-13
[ServiceBus]Evaluate renaming ReceiveMode to ServiceBusReceiveMode in line with dotnet

KieranBrantnerMagee:
Pending feedback from architects given potential inconsistency, CC @annatisch


sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_client.py
```diff
@@ -254,11 +255,14 @@ def get_queue_receiver(self, queue_name, **kwargs):
                 "To connect to the sub queue of a sessionful queue, "
                 "please set sub_queue only as sub_queue does not support session."
             )
-        if sub_queue and sub_queue in SubQueue:
+        try:
             queue_name = generate_dead_letter_entity_name(
                 queue_name=queue_name,
-                transfer_deadletter=(sub_queue == SubQueue.TransferDeadLetter)
+                transfer_deadletter=(ServiceBusSubQueue(sub_queue) == ServiceBusSubQueue.TransferDeadLetter)
```
annatisch, line 261: This works if `sub_queue` is both str and an enum instance? I think it does IIRC - but just double checking

sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_client.py
```diff
@@ -322,15 +326,16 @@ def get_subscription_receiver(self, topic_name, subscription_name, **kwargs):
          sessionful subscription, otherwise it must be None. In order to receive messages from the next available
          session, set this to ~azure.servicebus.NEXT_AVAILABLE_SESSION.
         :paramtype session_id: Union[str, ~azure.servicebus.NEXT_AVAILABLE_SESSION]
-        :keyword Optional[SubQueue] sub_queue: If specified, the subqueue this receiver will connect to.
+        :keyword Optional[Union[ServiceBusSubQueue, str]] sub_queue: If specified, the subqueue this receiver will
+         connect to.
```
annatisch, line 330: Do we want to add to the docstring something like: Possible values include 'deadletter', 'transferdeadletter'.

sdk/servicebus/azure-servicebus/tests/async_tests/test_queues_async.py
```diff
@@ -586,9 +586,9 @@ async def test_async_queue_by_servicebus_client_receive_batch_with_deadletter(se
 
             async with sb_client.get_queue_receiver(
                     servicebus_queue.name,
-                    sub_queue = SubQueue.DeadLetter,
+                    sub_queue = ServiceBusSubQueue.DeadLetter.value,
```
annatisch, line 589: Why the `.value` here? Is this to test it using strings?
Might be good to have some dedicate tests for not using enums. That tests things like case sensitivity etc.

sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_client.py
KieranBrantnerMagee: Correct.

sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_client.py
KieranBrantnerMagee: Mmm, do you worry that clutters it vs. someone being able to eyeball the enum?  Adds maintenance overhead too to keep those in parity.  (I hate the fact that I have to make this counterargument, if there was _any_ sort of docstring formatting I'd feel better adding more content to it.  If you feel like this'd be a big help or we have precedent let me know and I'll simply align.

sdk/servicebus/azure-servicebus/tests/async_tests/test_queues_async.py
KieranBrantnerMagee: Correct, it is to test using strings.  Truthfully I didn't go too much to edge cases because we were relying on the built-in enum string parsing functionality, but it may be more defensive.  I can slip something in there.

(Making sure we're on the same page though, due to the nature of the parse strategy, it is _not_ case insensitive.  Just making sure that was your expectation as well.)

annatisch: Yes that's my expectation. Though in the autorest generated SDKs we now have fancy case-insensitive enums - so we can adopt that as a future extension :)

sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_client.py
KieranBrantnerMagee: I used sqlalchemy as my tiebreaker.  It lists enum parameters, e.g. ISOLATION_LEVEL.  Add it is, solid call++

sdk/servicebus/azure-servicebus/CHANGELOG.md
```diff
@@ -2,8 +2,15 @@
 
 ## 7.0.0b9 (Unreleased)
 
+**New Features**
+
+* `sub_queue` and `receive_mode` may now be passed in as a valid string (as defined by their respective enum type) as well as their enum form.
```
yunhaoling, line 7:
```suggestion
* `sub_queue` and `receive_mode` may now be passed in as a valid string (as defined by their respective enum type) as well as their enum form when constructing `ServiceBusReceiver`.
```

sdk/servicebus/azure-servicebus/azure/servicebus/_common/constants.py
```diff
@@ -114,18 +114,21 @@
 TRANSFER_DEAD_LETTER_QUEUE_SUFFIX = '/$Transfer' + DEAD_LETTER_QUEUE_SUFFIX
 
 
-class ReceiveMode(Enum):
-    PeekLock = constants.ReceiverSettleMode.PeekLock
-    ReceiveAndDelete = constants.ReceiverSettleMode.ReceiveAndDelete
+class ServiceBusReceiveMode(Enum):
```
yunhaoling, line 117: I remember in other sdks like schema registry or storage blob, the class would actually inherit from both str and Enum like this:
```suggestion
class ServiceBusReceiveMode(str, Enum):
```

so that we could enable the str comparison:
like 
```python
"PeekLock" == ServiceBusReceiveMode.PeekLock # which would be True
```

sdk/servicebus/azure-servicebus/azure/servicebus/_common/constants.py
```diff
@@ -114,18 +114,21 @@
...
+    PeekLock = "peeklock"
```
yunhaoling, line 118: Does case matter here -- "PeekLock" vs "peeklock" or this is intentionally to be of lower case.

sdk/servicebus/azure-servicebus/azure/servicebus/_common/receiver_mixins.py
```diff
@@ -46,10 +46,7 @@ def _populate_attributes(self, **kwargs):
 
         self._auth_uri = "sb://{}/{}".format(self.fully_qualified_namespace, self.entity_path)
         self._entity_uri = "amqps://{}/{}".format(self.fully_qualified_namespace, self.entity_path)
-        self._receive_mode = kwargs.get("receive_mode", ReceiveMode.PeekLock)
-        # While we try to leave failures to the service, in this case the errors lower down the stack are less clear.
-        if not isinstance(self._receive_mode, ReceiveMode):
-            raise TypeError("Parameter 'receive_mode' must be of type ReceiveMode")
+        self._receive_mode = ServiceBusReceiveMode(kwargs.get("receive_mode", ServiceBusReceiveMode.PeekLock))
```
yunhaoling, line 49: case sensitivity when user passes string?
e.g. if user passes "PeekLock" instead "peeklock", I assume there would be a ValueError?
do we want to convert the string to lower case first?

sdk/servicebus/azure-servicebus/azure/servicebus/_common/constants.py
KieranBrantnerMagee: I have this fixed in a local change, just been waiting to get an OK on the enum naming inconsistencies with MGMT before I push it in, but good call :)

sdk/servicebus/azure-servicebus/azure/servicebus/_common/constants.py
KieranBrantnerMagee: Intentionally lower case.  See the discussion with the architects and the followup I pinged to the messaging archs. last night regarding -mgmt consistency.

KieranBrantnerMagee: Closing on this btw, the -mgmt enums are _magical_ and support case-insensitivity on the attribute side.  (Thanks laurent/anna/johan for the confirmation)

https://github.com/Azure/autorest.python/blob/49afdc2f15db14c48c2b7b4a9b37f5c4c7175f73/test/azure/Expected/AcceptanceTests/StorageManagementClient/storage/models/_storage_management_client_enums.py#L12-L35

sdk/servicebus/azure-servicebus/azure/servicebus/_common/constants.py
KieranBrantnerMagee: See comment below for full details on the resolution, but this is aligned.

sdk/servicebus/azure-servicebus/azure/servicebus/_common/receiver_mixins.py
KieranBrantnerMagee: It will ValueError, correct.  I'm not apt to convert because eventually if we onboard to "Fancy Future Enums" that'll come for us, and would be additive?  If you agree that it would be non-breaking I'm inclined to let that come via autorest if only because it being... basically code-complete day :P

KieranBrantnerMagee: (Edit, I literally just re-read the discussion with Anna I had above and she seems already on board with this plan, so LMK if you concur)

yunhaoling: yep, I see the conversation, sounds good to me

sdk/servicebus/azure-servicebus/tests/async_tests/test_queues_async.py
KieranBrantnerMagee: Dedicated tests added++

sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_client.py
```diff
@@ -386,7 +395,9 @@ def get_subscription_receiver(self, topic_name, subscription_name, **kwargs):
                 retry_backoff_max=self._config.retry_backoff_max,
                 **kwargs
             )
-        else:
+        except ValueError:
+            if sub_queue: # If we got here and sub_queue is defined, it's an incorrect value or something unrelated.
+                raise
```
yunhaoling, line 400: I think it should be moved ahead of "handler = ServiceBusReceiver".
otherwise if it's a normal queue if any value raised by ServiceBusReceiver constructors would be omitted.

yunhaoling: (I see get_queue_receiver() you're checking ahead)

yunhaoling: never mind, your code is good to handle the case

annatisch (approved):

PR #15326, opened 2020-11-14
[core] new paging story

iscai-msft:


sdk/core/azure-core/azure/core/async_paging.py
```diff
@@ -85,26 +88,45 @@ def __init__(
...
+        self._response = None  # type: Optional[ResponseType]
+        self._current_page = None  # type: Optional[Iterable[ReturnType]]
 
     async def __anext__(self):
-        if self.continuation_token is None and self._did_a_call_already:
+        if self._paging_method.finished(self.continuation_token):
             raise StopAsyncIteration("End of paging")
         try:
-            self._response = await self._get_next(self.continuation_token)
+            try:
```
xiangyan99, line 119: Why nested try?

iscai-msft: good point, un-nested

iscai-msft: had to re-nest @xiangyan99 because flattening the tries doesn't allow me to enter multiple except statements. In the case of a legacy call to paging with an `AzureError`, I need to enter both the `TypeError` except, and the `AzureError` except, which won't happen unless it's nested

sdk/core/azure-core/azure/core/async_paging.py
```diff
@@ -70,13 +71,15 @@ async def __anext__(self) -> ReturnType:
...
             [Optional[str]], Awaitable[ResponseType]
-        ],
-        extract_data: Callable[
+        ]] = None,
+        extract_data: Optional[Callable[
             [ResponseType], Awaitable[Tuple[str, AsyncIterator[ReturnType]]]
-        ],
+        ]] = None,
         continuation_token: Optional[str] = None,
+        paging_method=None,
```
johanste, line 81: I assume that this is an `Optional[str]`? Did we consider overloading the get_next to take a `str | Callable`?

sdk/core/azure-core/azure/core/async_paging_method.py
```diff
@@ -0,0 +1,234 @@
...
+from typing import Any, Callable, Iterable, Tuple
+from .paging import ResponseType, ReturnType
+from ._pipeline_client import PipelineClient
+from .pipeline.transport import HttpRequest, HttpResponse
+
+from .exceptions import (
+    HttpResponseError, ClientAuthenticationError, ResourceExistsError, ResourceNotFoundError, map_error
+)
+
+class AsyncPagingMethodABC:
```
johanste, line 35: Why didn't we make this into a "real" abstract class? See https://six.readthedocs.io/#six.add_metaclass and https://docs.python.org/3.2/library/abc.html#abc.ABCMeta for more info on how to make it py2 compatible....

johanste: ..which also begs the question, is this intended to be py3 only code?

sdk/core/azure-core/azure/core/paging_method.py
```diff
@@ -0,0 +1,225 @@
...
+    def __init__(self):
+        super(DifferentNextOperationPagingMethod, self).__init__()
+        self._prepare_next_request = None
+
+    def initialize(self, client, deserialize_output, next_link_name, **kwargs):
+        # type: (PipelineClient, Callable, str, Any) -> None
+        super(DifferentNextOperationPagingMethod, self).initialize(
+            client, deserialize_output, next_link_name, **kwargs
+        )
+        self._prepare_next_request = kwargs.pop("prepare_next_request", None)
```
johanste, line 202:
```python
try:
    self._prepare_next_request = kwargs.pop("prepare_next_request")
except KeyError:
    raise TypeError("<insert prefix here> is missing required keyword-only argument 'prepare_next_request'"
```

johanste: (in other words, TypeError for *missing* parameter, ValueError for present, but wrong value...)

sdk/core/azure-core/azure/core/paging_method.py
```diff
@@ -0,0 +1,225 @@
...
+        # type: (Any, HttpRequest) -> HttpResponse
+        if not self.did_a_call_already:
+            request = initial_request
+            self.did_a_call_already = True
+        else:
+            request = self.get_next_request(continuation_token, initial_request)
+        response = self._client._pipeline.run(request, stream=False)  # pylint: disable=protected-access
+
+        http_response = response.http_response
+        if not 200 <= http_response.status_code < 300:
```
johanste, line 144: Nitpick: reading doing `status_code < 200 or status_code >= 300` is marginally easier to read. At least that is what I have to mentally translate the line of code to say. But maybe that is my Swedish accent coming into play.

sdk/core/azure-core/azure/core/async_paging.py
```diff
@@ -85,26 +88,45 @@ def __init__(
...
-        if self.continuation_token is None and self._did_a_call_already:
+        if self._paging_method.finished(self.continuation_token):
             raise StopAsyncIteration("End of paging")
         try:
-            self._response = await self._get_next(self.continuation_token)
+            try:
+                self._response = await self._paging_method.get_page(self.continuation_token, self._initial_request)
+            except TypeError:
+                # legacy doesn't support passing initial request into get_page
+                self._response = await self._paging_method.get_page(self.continuation_token)
```
johanste, line 123: Is there a reason that the TypeError leaks out of the legacy methods get_page? Who else calls the method? If it is only intended to be called from here, and we always ignore the failure, then the method itself might as well ignore it...

johanste: ...which incidentally would get rid of the nested `try's :)

iscai-msft: In the new design, we define a `get_page` on the paging method, that takes in a cont token like it used to, but also takes in the initial request so it's more readily available for users to play with when they're overriding the paging method class. So, if we pass in initial request to the legacy callbacks people used to send, we get a `TypeError` if we pass in the initial request, which is why I catch it. However, after writing this whole chunk of response, I'm thinking I can deal with this in the `_LegacyPagingMethod` and get rid of the nested trys

sdk/core/azure-core/azure/core/paging_method.py
iscai-msft: sounds good, I def see where you're coming from

sdk/core/azure-core/azure/core/async_paging.py
iscai-msft: Sorry it took me so logn to address. this comment. `paging_method` is an optional `PagingMethod` class, I've added type hints for this. `get_next` is still a `callable`, we've just made it optional, since passing in `get_next` is the older way of paging

sdk/core/azure-core/azure/core/paging.py
```diff
@@ -24,65 +24,360 @@
...
+        :param optional[str] continuation_token_location: Property name on the response object that houses the
+         continuation token. Defaults to `None`, as `None` is an acceptable value for a continuation token location.
+         It means that there's no continuation token on the response object.
+        :return: The continuation token
+        :rtype: any
+        """
+
+        raise NotImplementedError("This method needs to be implemented")
+
+class NextLinkPagingMethod(PagingMethodABC):
```
iscai-msft, line 199: Sadly couldn't keep `CallbackPagingMethod` as the base, since `NextLinkPagingMethod` requires the `client` that is passed to us through the `get_next_request` method signature. Made `NextLlinkPagingMetjhod` the base bc it has no required parameters

sdk/core/azure-core/tests/async_tests/test_paging_async.py
```diff
@@ -105,39 +267,189 @@ async def extract_data(response):
...
+            return ["changedByCls"] * len(list_of_obj)
+
+        item_paged = AsyncItemPaged(
+            deserialize_output=deserialize_output,
+            client=client,
+            page_iterator_class=page_iterator,  # have to add this arg since I'm overriding PageIterator (vast majority won't use this param)
+            initial_state=http_request,
+            paging_method=ContinueWithNextLink(),
+            continuation_token_location="next_link",
+            _cls=cls,
```
iscai-msft, line 452: this is def showing the need to find a way around the `cls` issues. going to look into this more tomorrow

sdk/core/azure-core/azure/core/async_paging/_async_page_iterator.py
```diff
@@ -0,0 +1,107 @@
...
+         callables for get_next and extract_data.
+        :type paging_method: ~azure.core.paging_method.PagingMethodABC
+        """
+        if get_next or extract_data:
+            if paging_method:
+                raise ValueError(
+                    "You can't pass in both a paging method and a callback for get_next or extract_data. "
+                    "We recommend you only pass in a paging method, since passing in callbacks is legacy."
+                )
+            if not get_next and extract_data:
```
annatisch, line 76: I don't think this if statement works - wouldn't it need to be:
`if not (get_next and extract_data):`

sdk/core/azure-core/azure/core/paging/_paging_method.py
```diff
@@ -0,0 +1,230 @@
...
+        """Get the continuation token from the current page. This operation returning None signals the end of paging.
+        Continuation token can be mutated here as well.
+
+        :param pipeline_response: The immediate response returned from the pipeline
+        :type pipeline_response: ~azure.core.pipeline.transport.HttpResponse
+        :param object deserialized: The deserialized pipeline_response
+        :param optional[str] continuation_token_location: Property name on the response object that houses the
+         continuation token. Defaults to `None`, as `None` is an acceptable value for a continuation token location.
+         It means that there's no continuation token on the response object.
+        :return: The continuation token
```
annatisch, line 100: It might be good to specify here that this value can be of any type - and will simply be passed as-is to the `get_next_request` method?

sdk/core/azure-core/azure/core/paging/_paging_method.py
```diff
@@ -0,0 +1,230 @@
...
+                        continuation_token_location
+                    )
+                )
+            return getattr(deserialized, continuation_token_location)
+        return None
+
+
+class ContinueWithCallback(ContinueWithNextLink):  # pylint: disable=too-many-instance-attributes
+    def __init__(self, next_request_callback, **kwargs):
+        """Base paging method. Accepts the callback for the next request as an init arg.
```
annatisch, line 185: Seems like this isn't the `Base paging method`?

sdk/core/azure-core/azure/core/async_paging/_async_page_iterator.py
iscai-msft: yep, def right, thanks!

sdk/core/azure-core/azure/core/async_paging/_async_page_iterator.py
```diff
@@ -0,0 +1,107 @@
...
+from ._async_paging_method_handler import _AsyncPagingMethodHandler
+from ._async_list import AsyncList
+from ..paging._utils import ReturnType, ResponseType
+from ..paging import PagingMethodABC
+from ..exceptions import AzureError
+from ..pipeline._tools_async import await_result as _await_result
+
+_LOGGER = logging.getLogger(__name__)
+
+class AsyncPageIterator(AsyncIterator[AsyncIterator[ReturnType]]):
```
johanste, line 46: Shouldn't this be `AsyncPageIterator(AsyncIterator[typing.List[ReturnType]]):`? Individual pages are not asynchronously retrieved... Only each page is...

sdk/core/azure-core/azure/core/async_paging/_async_page_iterator.py
```diff
@@ -0,0 +1,107 @@
...
+    def __init__(
+        self,
+        get_next: Optional[Callable[
+            [Optional[str]], Awaitable[ResponseType]
+        ]] = None,
+        extract_data: Optional[Callable[
+            [ResponseType], Awaitable[Tuple[str, AsyncIterator[ReturnType]]]
+        ]] = None,
+        continuation_token: Optional[str] = None,
+        paging_method: Optional[PagingMethodABC] = None,
```
johanste, line 56: We should make at least `paging_method` a keyword-only argument for the AsyncPageIterator. Per our guidelines, all optional arguments should have been keyword-only, but we shouldn't make a breaking change just for that.

sdk/core/azure-core/azure/core/paging/_paging_method.py
```diff
@@ -0,0 +1,241 @@
...
+        Callable,
+        Iterable,
+        Optional,
+    )
+    from ._utils import ReturnType, ResponseType
+    from ..pipeline.transport import HttpResponse, HttpRequest
+    from .._pipeline_client import PipelineClient
+
+@add_metaclass(ABCMeta)
+class PagingMethodABC():
```
johanste, line 42: `class PagingMethodABC(object):`?

sdk/core/azure-core/azure/core/paging/_paging_method.py
```diff
@@ -0,0 +1,241 @@
...
+
+    def get_next_request(self, continuation_token, initial_request, client):
```
johanste, line 44: Why isn't this marked as an `@abstractmethod`?

sdk/core/azure-core/azure/core/paging/_paging_method.py
```diff
@@ -0,0 +1,241 @@
...
+        # type: (HttpResponse, ResponseType, str) -> Iterable[ReturnType]
+        """Extract the list elements from the current page to return to users
+
+        :param pipeline_response: The immediate response returned from the pipeline
+        :type pipeline_response: ~azure.core.pipeline.transport.HttpResponse
+        :param object deserialized: The deserialized pipeline_response
+        :return: The iterable of items extracted out of paging response
+        :rtype: iterable[object]
+        """
+        if hasattr(deserialized, item_name):
```
johanste, line 140: Nit: you can simply try to access the name and catch AttributeError. Minor performance gain (no need to look up the attribute twice.

sdk/core/azure-core/azure/core/paging/_paging_method.py
```diff
@@ -0,0 +1,241 @@
...
+
+class ContinueWithNextLink(ContinueWithCallback):  # pylint: disable=too-many-instance-attributes
+
+    def __init__(self, path_format_arguments=None, **kwargs):
+        """Most common paging method. Uses the continuation token as the URL for the next call.
+        """
+        def _next_request_callback(continuation_token, initial_request, client):
+            request = initial_request
+            next_link = continuation_token
+            if self._path_format_arguments:
```
johanste, line 189: If the default value for `path_format_arguments` was `{}`, then you wouldn't have to make this check. Is there a specific reason we want the default value to be `None`?

sdk/core/azure-core/azure/core/paging/_paging_method.py
```diff
@@ -0,0 +1,241 @@
...
+                next_link = client.format_url(next_link, **self._path_format_arguments)
+            else:
+                next_link = client.format_url(next_link)
+            request.url = next_link
+            return request
+
+        super(ContinueWithNextLink, self).__init__(
+            next_request_callback=_next_request_callback, **kwargs
+        )
+        self._path_format_arguments = path_format_arguments
```
johanste, line 199: Do we really need this instance attribute? The argument passed in is already captured in the `_next_request_callback`, so is there a specific need to access if from `self`? I don't see it being mutated anywhere along the way...

sdk/core/azure-core/azure/core/paging/_paging_method_handler.py
```diff
@@ -0,0 +1,110 @@
...
+        if isinstance(self._initial_state, PipelineResponse):
+            return self._initial_state.http_response.request
+        return self._initial_state
+
+    def _handle_response(self, continuation_token, response):
+        http_response = response.http_response
+        status_code = http_response.status_code
+        if status_code < 200 or status_code >= 300:
+            if self._error_map:
+                map_error(status_code=status_code, response=http_response, error_map=self._error_map)
```
johanste, line 72: Does `map_error` populate the continuation_token?

sdk/core/azure-core/azure/core/paging/_paging_method_handler.py
```diff
@@ -0,0 +1,110 @@
...
+        client,
+        initial_state,
+        **kwargs
+    ):
+        self._paging_method = paging_method
+        self._deserialize_output = deserialize_output
+        self._client = client
+        self._initial_state = initial_state
+        self._cls = kwargs.pop("_cls", None)
+        self._error_map = {
```
johanste, line 53: What about making self._error_map a `DefaultDict` with `HttpResponseError` as the default/missing key value? That way, `map_error` would always raise and you don't need to check if the status code is known or not...

sdk/core/azure-core/azure/core/paging/_paging_method_handler.py
```diff
@@ -0,0 +1,110 @@
...
+            401: ClientAuthenticationError, 404: ResourceNotFoundError, 409: ResourceExistsError
+        }
+        self._error_map.update(kwargs.pop('error_map', {}))
+        self._item_name = kwargs.pop("item_name", "value")
+        self._continuation_token_location = kwargs.pop("continuation_token_location", None)
+        self._operation_config = kwargs
+
+    @property
+    def _initial_request(self):
```
johanste, line 62: Are subclasses supposed to override any of the private methods or access any private attributes? If so, they should be made public.

sdk/core/azure-core/azure/core/async_paging/_async_page_iterator.py
iscai-msft: You're right, I don't htink it should be `AsyncIterator`[`AsyncIterator`], I had copied that from what we have in core right now. However, we return an `AsyncList` to users, so I think it should be `AsyncIterator`[`AsyncList`]

sdk/core/azure-core/azure/core/async_paging/_async_page_iterator.py
iscai-msft: made them both kwargs for good measure

sdk/core/azure-core/azure/core/paging/_paging_method.py
iscai-msft: no, i can make the default {}, thanks!

sdk/core/azure-core/azure/core/paging/_paging_method.py
iscai-msft: good point, won't catch it

sdk/core/azure-core/azure/core/paging/_paging_method_handler.py
iscai-msft: that's a really good point, need to make sure that happens

sdk/core/azure-core/azure/core/paging/_paging_method_handler.py
iscai-msft: great idea, thanks!

iscai-msft: side note: it doesn't work bc `map_error` calls .get() on the dictionary, which overrides the default value of the dictionary to `None`

sdk/core/azure-core/azure/core/paging/_paging_method_handler.py
iscai-msft: it doesn't, but it gets populated in the page iterator, where we catch all `AzureError`s and set the continuation token

PR #15348, opened 2020-11-16
Update arm changelog for Nov release

ankitarorabit:


turalf (commented):
We also need to change the version in the version.py file

turalf (approved):

PR #15364, opened 2020-11-17
[ServiceBus] Condition based error redesign + client validation alignment

yunhaoling:
https://gist.github.com/yunhaoling/1c13d4e50f81be25d35962edb4430106#gistcomment-3525780


sdk/servicebus/azure-servicebus/azure/servicebus/_base_handler.py
```diff
@@ -193,9 +195,11 @@ def _convert_connection_string_to_kwargs(cls, conn_str, **kwargs):
 
         entity_in_kwargs = queue_name or topic_name
         if entity_in_conn_str and entity_in_kwargs and (entity_in_conn_str != entity_in_kwargs):
-            raise ServiceBusAuthenticationError(
-                "Entity names do not match, the entity name in connection string is {};"
-                " the entity name in parameter is {}.".format(entity_in_conn_str, entity_in_kwargs)
+            raise ServiceBusAuthenticationError(  # TODO: should this be a ValueError?
```
annatisch, line 198: Yes I think this should be a value error rather than an auth error

sdk/servicebus/azure-servicebus/azure/servicebus/_common/message.py
```diff
@@ -161,6 +161,9 @@ def session_id(self):
     @session_id.setter
     def session_id(self, value):
         # type: (str) -> None
+        if value and len(value) > 128:
```
KieranBrantnerMagee, line 164: Nit: const 128? (MAX_ID_LENGTH or something so you can use one for all?)

if you think this is overkill ignore me, this is VERY bikesheddy.

sdk/servicebus/azure-servicebus/azure/servicebus/_common/message.py
```diff
@@ -202,6 +205,13 @@ def partition_key(self):
     @partition_key.setter
     def partition_key(self, value):
         # type: (str) -> None
+        if value and len(value) > 128:
+            raise ValueError("partition_key cannot be longer than 128 characters.")
+
+        if value and value != self.session_id:
```
KieranBrantnerMagee, line 211: To confirm, we don't want this logic in session_id as well for the inverse relationship?

sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_sender.py
```diff
@@ -169,7 +169,7 @@ def _from_connection_string(
         :rtype: ~azure.servicebus.ServiceBusSender
 
         :raises ~azure.servicebus.ServiceBusAuthenticationError: Indicates an issue in token/identity validity.
-        :raises ~azure.servicebus.ServiceBusAuthorizationError: Indicates an access/rights related failure.
+        :raises ~azure.servicebus.UnauthorizedAccessError: Indicates an access/rights related failure.
```
KieranBrantnerMagee, line 172: @annatisch from OOB convo with Adam, this was motivated by normalization with dotnet, but it doesn't sit well with me, especially in line with our broader leaning towards prefixing especially for ambiguous errors. (To be precise, I'd revert to ServiceBusAuthorization, or, if the alignment was desirable, meet in the middle with (Oh goodness.) ServiceBusUnauthorizedAcceessError)  How do you feel about this one?

sdk/servicebus/azure-servicebus/azure/servicebus/aio/_base_handler_async.py
```diff
@@ -114,23 +115,42 @@ async def __aenter__(self):
...
-        error, error_need_close_handler, error_need_raise = \
-            _create_servicebus_exception(_LOGGER, exception, self, **kwargs)
-        if error_need_close_handler:
+    async def _handle_exception(self, exception):
+        # pylint: disable=protected-access
+        error = _create_servicebus_exception(_LOGGER, exception)
+
+        try:
+            # If SessionLockLostError or ServiceBusConnectionError happen when a session receiver is running,
+            # the receiver can no longer be used and should create a new session receiver
```
KieranBrantnerMagee, line 124:
```suggestion
            # the receiver should no longer be used and should create a new session receiver
```
So the one thing I'd note here (which the above tweak gets at but doesn't complete because <waves fist at github> no multi-line suggestions yet; is our motivation for disallowing this: namely that there are pitfalls WRT both next session IDs, and the diversity of session failure modes, that motivates us to disallow this. (as opposed to the original verbage of "can no longer be used" which to me implied that there's a service reason for it.)

sdk/servicebus/azure-servicebus/azure/servicebus/exceptions.py
```diff
@@ -228,108 +193,245 @@ class ServiceBusError(AzureError):
...
+        self._condition = kwargs.pop("condition", None)
+        self._status_code = kwargs.pop("status_code", None)
+        try:
+            message = message.decode('UTF-8')
+        except AttributeError:
+            pass
+
+        if self._condition:
+            try:
+                condition = self._condition.decode('UTF-8')
```
KieranBrantnerMagee, line 208: If we're storing message on self decoded, why not do the same with condition?

sdk/servicebus/azure-servicebus/azure/servicebus/exceptions.py
```diff
@@ -228,108 +193,245 @@ class ServiceBusError(AzureError):
...
+            retryable=True,
+            shutdown_handler=True,
+            **kwargs
+        )
 
 
 class ServiceBusAuthenticationError(ServiceBusError):
     """An error occurred when authenticate the connection."""
+    def __init__(self, **kwargs):
+        message = kwargs.pop("message", None) or "An error occurred when authenticate the connection."
```
KieranBrantnerMagee, line 232:
```suggestion
        message = kwargs.pop("message", None) or "An error occurred when authenticating the connection."
```

sdk/servicebus/azure-servicebus/azure/servicebus/exceptions.py
```diff
@@ -228,108 +193,245 @@ class ServiceBusError(AzureError):
...
 
-class ServiceBusMessageError(ServiceBusError):
-    """An error occurred when an operation on a message failed because the message is in an incorrect state."""
-
-
-class MessageContentTooLarge(ServiceBusMessageError, ValueError):
+class MessageSizeExceededError(ServiceBusError, ValueError):
     """Message content is larger than the service bus frame size."""
+    def __init__(self, **kwargs):
+        message = kwargs.pop("message", None) or "Message content is larger than the service bus frame size."
```
KieranBrantnerMagee, line 268: Can we give the value here as part of the message?

sdk/servicebus/azure-servicebus/azure/servicebus/exceptions.py
```diff
@@ -228,108 +193,245 @@ class ServiceBusError(AzureError):
...
+        super(MessageSizeExceededError, self).__init__(
+            message,
+            retryable=False,
+            shutdown_handler=False,
+            **kwargs
+        )
 
 
-class MessageAlreadySettled(ServiceBusMessageError):
+class MessageAlreadySettled(ValueError):
```
KieranBrantnerMagee, line 277:
```suggestion
# ValueError was preferred to ServiceBusMessageError/ServiceBusError as a base
# since this arises only from client validation.
class MessageAlreadySettled(ValueError):
```

Feel free to take it or leave it, but as with the other comment, I'm just big on "writing down why things that are different are different so that those who come across this after understand"

sdk/servicebus/azure-servicebus/azure/servicebus/exceptions.py
```diff
@@ -228,108 +193,245 @@ class ServiceBusError(AzureError):
...
+        message = kwargs.pop("message", None) or "An entity with the same name exists under the same namespace."
+        super(MessagingEntityAlreadyExistsError, self).__init__(
+            message,
+            retryable=False,
+            shutdown_handler=False,
+            **kwargs
+        )
+
+
+class QuotaExceededError(ServiceBusError):
```
KieranBrantnerMagee, line 379: Is this "generic"?  I worry in an Azure situation, folks may wonder if it's an Azure quota or SB specific.  cc @annatisch

sdk/servicebus/azure-servicebus/azure/servicebus/exceptions.py
```diff
@@ -74,145 +68,116 @@ def _error_handler(error):
...
+        try:
+            # handling AMQP Errors that have the condition field
+            condition = exception.condition
+            description = exception.description
+            exception = _handle_amqp_exception_with_condition(logger, condition, description, exception=exception)
+        except AttributeError:
+            # handling AMQP Errors that don't have the condition field
+            exception = _handle_amqp_exception_without_condition(logger, exception)
+    elif not isinstance(exception, ServiceBusError):
+        logger.info("Unexpected error occurred (%r). Handler shutting down.", exception)
```
KieranBrantnerMagee, line 156:
```suggestion
        logger.exception("Unexpected error occurred (%r). Handler shutting down.", exception)
```
Assuming this is as exceptional as this seems, may be worth using the above (or error, take your pick.  "exception" is supposed to only be used in an exception handler, which this technically is, if deeply embedded, so should be fine.)

sdk/servicebus/azure-servicebus/azure/servicebus/exceptions.py
```diff
@@ -74,145 +68,116 @@ def _error_handler(error):
...
+    status_code=None
+):
+    #
+    # handling AMQP Errors that have the condition field or the mgmt handler
+    logger.info("AMQP error occurred: (%r), condition: (%r), description: (%r).", exception, condition, description)
+    if condition == AMQPErrorCodes.NotFound:
+        # handle NotFound error code
+        error_cls = ServiceBusCommunicationError \
+            if isinstance(exception, AMQPErrors.AMQPConnectionError) else MessagingEntityNotFoundError
+    elif condition == AMQPErrorCodes.ClientError and 'timed out' in str(exception):
```
KieranBrantnerMagee, line 97: Just checking there's no more reliable way to check the timeout than string comparison?  That's just... more fragile than I'd like if there's any other approach, but if we're stuck we're stuck.  (I see we had this in the past so can feel free to add this as a post-GA issue, can even be up-for-grabs to investigate.)

KieranBrantnerMagee: As an aside though, if we're OK with this, it makes me wonder if we could do the same to distinguish the different types of session timeouts that we used to have, if they gave any feedback in that exception body.

sdk/servicebus/azure-servicebus/azure/servicebus/_common/message.py
yunhaoling: ++, updated

sdk/servicebus/azure-servicebus/azure/servicebus/_common/message.py
yunhaoling: right, it's one-way

sdk/servicebus/azure-servicebus/azure/servicebus/aio/_base_handler_async.py
yunhaoling: put your words into the comments++

sdk/servicebus/azure-servicebus/azure/servicebus/exceptions.py
yunhaoling: This is just a default message, we do the "provide from higher in the callstack" thing.

sdk/servicebus/azure-servicebus/azure/servicebus/exceptions.py
yunhaoling: This is somewhat limited by our uamqp implementation as in uamqp send messages we would wrap any error into ClientError and this could be improved in uamqp.

yunhaoling: issue created: https://github.com/Azure/azure-uamqp-python/issues/184

sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_sender.py
yunhaoling: as per oob discussion:
UnauthorizedAccessError back to ServiceBusAuthorizationError

sdk/servicebus/azure-servicebus/azure/servicebus/exceptions.py
yunhaoling: as per oob discussion:

QuotaExceededError -> ServiceBusQuotaExceededError
ServiceBusyError -> ServiceBusServerBusyError

sdk/servicebus/azure-servicebus/azure/servicebus/_common/receiver_mixins.py
```diff
@@ -107,21 +94,24 @@ def _get_source(self):
...
+
+        if self._receive_mode == ReceiveMode.ReceiveAndDelete:
+            raise ValueError(
+                "The operation {} is not supported in 'ReceiveAndDelete' receive mode.".format(action)
+            )
+
         if message._settled:
-            raise MessageAlreadySettled(action)
-        try:
-            if message._lock_expired:
```
annatisch, line 116: I thought we were keeping some kind of artificial client-side check here until we could fix uamqp?

yunhaoling: yep, we only check in the message settlement now

KieranBrantnerMagee (dismissed):
LGTM.  This was not an easy set of changes, thank you @yunhaoling profusely for enduring all the flux that went into them.

(Noting for posterity:  Intent to incrementally add some sort of SettleFailed/SendFailed "grouping" errors for ease-of-use, hopefully those helpers are not going to be gone for long, only nixed for now as we got the base scaffolding set to be backcompat-safe for GA.)

KieranBrantnerMagee (approved):

PR #15420, opened 2020-11-18
[T1]Cdn 2020-11-18

msyyc:
https://github.com/Azure/sdk-release-request/issues/961
--


sdk/cdn/azure-mgmt-cdn/azure/mgmt/cdn/models/_models_py3.py
```diff
@@ -3392,7 +3379,7 @@ class RequestHeaderMatchConditionParameters(Model):
 
     odatatype = "#Microsoft.Azure.Cdn.Models.DeliveryRuleRequestHeaderConditionParameters"
 
-    def __init__(self, *, selector: str, operator, match_values, negate_condition: bool=None, transforms=None, **kwargs) -> None:
+    def __init__(self, *, operator, selector: str=None, negate_condition: bool=None, match_values=None, transforms=None, **kwargs) -> None:
```
lsmith130, line 3382: Unfortunately this looks like a breaking change. In the C# SDK, we're customizing the generated code to avoid a breaking change here, is that possible with the python SDK? If not let's make sure this is released as a new major version.

msyyc: @lsmith130 
image
picture above indicates that the change makes the parameters optional. Is there anything wrong?

lsmith130: I see, these are keyword-only arguments. In that case, please ignore my earlier comment. This looks good to me.

00Kai0 (approved):

PR #15427, opened 2020-11-18
[Communication] Remove user_credential_tests from the shared folder(s)

turalf:


ankitarorabit (commented):
Don't you need to move these tests to some different folder? Why are they removed instead?

turalf:
> Don't you need to move these tests to some different folder? Why are they removed instead?

Yes. This test file was living in a shared folder, that is why was duplicated across packages. Now in the administration package I removed it from the _shared folder and added to the list of tests. With that way we don't have to have this test file in other packages.

ankitarorabit (approved):

PR #15447, opened 2020-11-19
Pin Pip 20.2 

scbedd:
I don't expect breaks from 20.3, but just to be certain eh?


kristapratico (approved):

mitchdenny:
Log an issue to do the upgrade to ```pip==20.3``` when it releases?

PR #15455, opened 2020-11-19
ACS Chat: Changes for API version: 2020-11-01-preview3

Leoaqr:
Several changes required by `2020-11-01-preview3` of the API:

- Rename `Thread Members` to `Participants`
- Rename function `UpdateThread` to `UpdateTopic` 
- Rename `ReadReceipt` to `ChatMessageReadReceipt` in model
- Return message id (string) instead of SendMessageResult (object) when calling `send_message`
- Add a convenience method `add_participant` for adding 1 participant
- Add failed participants check in response header in `create_thread` function


sdk/communication/azure-communication-chat/azure/communication/chat/_shared/response.py
```diff
@@ -0,0 +1,8 @@
+# ------------------------------------------------------------------------
```
fangchen0601, line 1: the _shared folder is shared between all service SDKs, like SMS SDK has the same folder and same files/methods. Which means adding one method/module to this folder, other teams must do the same. 
So I suggest:
1. Talk to Tural to ask if above is the same case, if yes, ask if this will become a common method that other teams will use, if yes, then your PR is good
2. If this module and method is only for Chat team, suggest move it to azure/communication/chat/_util.py after talked to Tural.

fangchen0601 (commented):
lgtm. please make sure unit test and live test passed locally. and upload recorded yaml files if needed.

sdk/communication/azure-communication-chat/azure/communication/chat/_shared/response.py
Leoaqr: Thanks for the feedback. 
Currently, I don't commit the test files in order to reduce the unnecessary review workload. Yes, I will run and upload new test files after review.  

For the helper file in shared folder, what do you think? @turalf

turalf: I guess this file is specific to the chat module. If that is the case please keep this file outside of the `_shared`  folder. 
Also it seems that this method is used only in 1 file. If you are not planning to use it somewhere else in the future, better to keep in that file.

turalf (commented):
Please make sure to add recording files after running live tests locally, once they are available

Leoaqr:
> Please make sure to add recording files after running live tests locally, once they are available

turalf (approved):
Lgtm

sdk/communication/azure-communication-chat/azure/communication/chat/_chat_thread_client.py
```diff
@@ -246,8 +246,8 @@ def send_message(
         :keyword str sender_display_name: The display name of the message sender. This property is used to
           populate sender name for push notifications.
         :keyword callable cls: A custom type or function that will be passed the direct response
-        :return: SendChatMessageResult, or the result of cls(response)
-        :rtype: ~azure.communication.chat.SendChatMessageResult
+        :return: str, or the result of cls(response)
+        :rtype: str
```
annatisch, line 250: This change should not have been made - could we fix it before the next release goes out?
Instead we should use something like this:
`return { 'id': send_chat_message_result.id }`

sdk/communication/azure-communication-chat/azure/communication/chat/_chat_thread_client.py
```diff
@@ -113,7 +113,7 @@ def thread_id(self):
         return self._thread_id
 
     @distributed_trace
-    def update_thread(
+    def update_topic(
         self,
         topic=None,  # type: Optional[str]
```
annatisch, line 118: If we are updating the topic-  the `topic` parameter should no longer be optional.

sdk/communication/azure-communication-chat/azure/communication/chat/_chat_client.py
```diff
@@ -142,28 +144,29 @@ def create_chat_thread(
...
+            response_header = response.http_response.headers
+            if ('Azure-Acs-InvalidParticipants' in response_header.keys() and
+                response_header['Azure-Acs-InvalidParticipants'] is not None):
+                invalid_participant_and_reason_list = response_header['Azure-Acs-InvalidParticipants'].split('|')
+                errors = []
+                for invalid_participant_and_reason in invalid_participant_and_reason_list:
+                    participant, reason = invalid_participant_and_reason.split(',', 1)
+                    errors.append('participant ' + participant + ' failed to join thread '
+                    + create_chat_thread_result.id + ' return statue code ' + reason)
+                raise ValueError(errors)
```
annatisch, line 165: This should not be a ValueError

sdk/communication/azure-communication-chat/azure/communication/chat/_chat_thread_client.py
Leoaqr: Will change accordingly in the next iteration

PR #15461, opened 2020-11-20
[formrecognizer] skip aad tests

kristapratico:


sdk/formrecognizer/azure-ai-formrecognizer/tests/test_mgmt_async.py
```diff
@@ -22,6 +22,7 @@ class TestManagementAsync(AsyncFormRecognizerTest):
 
     @GlobalFormRecognizerAccountPreparer()
     @pytest.mark.live_test_only
+    @pytest.mark.skip
```
maririos, line 25: do u usually have to put a reason? or just skip is fine

kristapratico:
AAD is now supported so this is unnecessary

PR #15467, opened 2020-11-20
[SeviceBus] Entity name validation when spawning handler from client

yunhaoling:


sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_client.py
```diff
@@ -359,6 +380,13 @@ def get_subscription_receiver(self, topic_name, subscription_name, **kwargs):
 
         """
         # pylint: disable=protected-access
+
+        if self._entity_name and topic_name != self._entity_name:
```
KieranBrantnerMagee, line 384: for sub receiver is it still just topic_name?  and not like topic_name/subscription_name?

KieranBrantnerMagee: (from OOB no, topic-only connstr as far as entity-specific goes.)

yunhaoling: right, connection string only contains the topic name as the entity path

KieranBrantnerMagee (approved):

KieranBrantnerMagee (approved):
One of the old client tests had the service side exception expect for the client side check we just put in, was a simple/safe fix, reapproving/running++

PR #15484, opened 2020-11-20
[Communication] Remove test_list_read_receipts e2e test for chat 

turalf:
This tests fails from time to time due to the fact that the api is eventually consistent. 
There is already a unit test covering this case. Therefore removing this until we have a proper way of doing e2e tests for this case


ankitarorabit (approved):
Not sure why are these removed?

turalf:
> Not sure why are these removed?



> Not sure why are these removed?

This test is unreliable when we do it in live mode. I.e. When the message is sent, it is not immediately available in the backend. Therefore, we had to wait before sending read receipt. That causes unreliable test runs. Therefore, we are  doing this through mocking network call. Which is already in `test_chat_thread_client.py` file.

PR #15485, opened 2020-11-20
Move to eng/common docindex reference 

sima-zhu:


sima-zhu:
closed this one as we opened a new one

PR #15488, opened 2020-11-20
add python 3.9 to live test matrix

kristapratico:


eng/pipelines/templates/jobs/archetype-sdk-tests.yml
```diff
@@ -25,9 +25,9 @@ parameters:
     Linux_PyPy3:
       OSVmImage: 'ubuntu-18.04'
       PythonVersion: 'pypy3'
-    Linux_Python38:
+    Linux_Python39:
       OSVmImage: 'ubuntu-18.04'
-      PythonVersion: '3.8'
+      PythonVersion: '3.9.0'
```
weshaggard, line 30: Not sure it matters but is there a reason this is 3.9.0 instead of 3.9?

weshaggard (approved):
Change looks reasonable to me assuming we are good with removing the 38 testing and replace it with 3.9.

eng/pipelines/templates/jobs/archetype-sdk-tests.yml
kristapratico: Hm not sure. I copied this from the servicebus tests.yml. Will wait on final say from @scbedd

scbedd: IMO this is fine. I will check if `py39` has been added to the actual images themselves. We would use `3.9` like you're asking if it is already on the machine Wes.

scbedd: Given that this would need to be changed in 3 or so locations. This should probably just merge.

scbedd (approved):

PR #15489, opened 2020-11-20
[ServiceBus] Pre-GA Bug Bash fixes

KieranBrantnerMagee:
- Small fix found during bug bash, an already disposed handler during iter finalizer causes issues on iter termination.
- Given the risks the above exposed, made the explicit getter which necessitated the iter wrapper voodoo internal so as to give time to ensure we're not locking ourselves into a dangerous pattern.  This was simply a helper and the same functionality can still be accessed by iterating on the base receiver.
- Adjusts docs, samples, tests.
- Clarify one changelog point (Thanks for noting Yunhaoling)
- Fixes up some stress tests that had been missed in prior refactors and ensure all green.


sdk/servicebus/azure-servicebus/CHANGELOG.md
```diff
@@ -36,7 +36,7 @@
   - Introduced `SessionCannotBeLockedError` which would be raised when the requested session cannot be locked.
 * `ServiceBusClient.get_queue/topic_sender` and `ServiceBusClient.get_queue/subscription_receiver` will now
 raise `ValueError` if the `queue_name` or `topic_name` does not match the `EntityPath` in the connection string used to construct the `ServiceBusClient`.
-
+* `get_streaming_message_iter` has been made internal for the time being to assess use patterns before comitting to back-compatibility; messages may still be iterated over in equivelent fashion by iterating on the receiver itself.
```
yunhaoling, line 39: nit: sorry about being paranoid on adding the class name to make it more explicit :P
```suggestion
* `ServiceBusReceiver.get_streaming_message_iter` has been made internal for the time being to assess use patterns before comitting to back-compatibility; messages may still be iterated over in equivelent fashion by iterating on the receiver itself.
```

sdk/servicebus/azure-servicebus/tests/stress_tests/test_stress_queues.py
```diff
@@ -299,3 +301,50 @@ def test_stress_queue_close_and_reopen(self, servicebus_namespace_connection_str
         result = stress_test.run()
         assert(result.total_sent > 0)
         assert(result.total_received > 0)
+
+    class DroppedMessageCheckerStressTestRunner(StressTestRunner):
```
yunhaoling, line 305: hmmm, can you explain a bit what this stress test is doing?

yunhaoling:
LGTM

yunhaoling (dismissed):

sdk/servicebus/azure-servicebus/tests/stress_tests/test_stress_queues.py
KieranBrantnerMagee: Yup!  (And I assume you mean in comment too so I can certainly do that, but for progeny here: ) It's a check motivated by Harsha having observed that over a million messages they observed some get lost due to <problems> on their end, and it seemed like (given our having users with this concern in EH) a reasonable thing to have validation for.

sdk/servicebus/azure-servicebus/CHANGELOG.md
KieranBrantnerMagee: Nono, it's my fault for not realizing, I even made the change to adjust this on the other line and then apparently my brain shut off.  Always ++ for clarity.

sdk/servicebus/azure-servicebus/tests/stress_tests/test_stress_queues.py
KieranBrantnerMagee:
```suggestion
    # This test validates that all individual messages are received contiguously over a long time period.
    # (e.g. not dropped for whatever reason, not sent, or not received)
    class DroppedMessageCheckerStressTestRunner(StressTestRunner):
```

yunhaoling (dismissed):

yunhaoling (approved):

rakshith91 (approved):

PR #15510, opened 2020-11-23
[Perfstress] Added framework

### LINKED ISSUE #12581: Add Python PerfStress framework
Move Python PerfStress framework from https://github.com/mikeharder/azure-sdk-perf to https://github.com/Azure/azure-sdk-for-python.

annatisch:
- Fixes #12581


sdk/storage/azure-storage-file-share/tests/perfstress/_test_base.py
```diff
@@ -0,0 +1,93 @@
...
+
+    async def close(self):
+        await self.async_service_client.close()
+        await super().close()
+
+    @staticmethod
+    def add_arguments(parser):
+        super(_ServiceTest, _ServiceTest).add_arguments(parser)
+        parser.add_argument('--max-range-size', nargs='?', type=int, help='Maximum size of file uploading in single HTTP PUT. Defaults to 4*1024*1024', default=4*1024*1024)
+        parser.add_argument('--service-client-per-instance', action='store_true', help='Create one ServiceClient per test instance.  Default is to share a single ServiceClient.', default=False)
```
mikeharder, line 46: Why is there a parameter for this?  Do we want to test one shared ServiceClient vs one instance per test class?  Do we expect to see different perf?  What do we recommend customers do?

sdk/storage/azure-storage-file-share/tests/perfstress/download.py
```diff
@@ -0,0 +1,36 @@
...
+class DownloadTest(_ShareTest):
+    def __init__(self, arguments):
+        super().__init__(arguments)
+        file_name = "downloadtest"
+        self.sharefile_client = self.share_client.get_file_client(file_name)
+        self.async_sharefile_client = self.async_share_client.get_file_client(file_name)
+
+    async def global_setup(self):
+        await super().global_setup()
+        data = b'a' * self.args.size
```
mikeharder, line 18: We should probably use random bytes instead of `aaa...`, for one thing to ensure it is non-compressable.  We already have a `random_stream` helper class, could we use it here?  We may also want a `random_array` helper class/method for scenarios which do need an array.

sdk/storage/azure-storage-file-share/tests/perfstress/download.py
```diff
@@ -0,0 +1,36 @@
...
+        await self.async_sharefile_client.upload_file(data)
+
+    def run_sync(self):
+        stream = self.sharefile_client.download_file(max_concurrency=self.args.parallel)
```
mikeharder, line 22: You cannot use the `parallel` argument for this purpose, since it's already being used by the framework internally to run `parallel` instances of your test class in parallel.  If you do want an argument for `max_concurrency`, you will need to add a new argument for it, so the user can set `parallel` and `max_concurrency` independently.

sdk/storage/azure-storage-file-share/tests/perfstress/download.py
```diff
@@ -0,0 +1,36 @@
...
+        await stream.readall()
+
+    async def close(self):
+        await self.async_sharefile_client.close()
+        await super().close()
+
+    @staticmethod
+    def add_arguments(parser):
+        super(DownloadTest, DownloadTest).add_arguments(parser)
+        parser.add_argument('-s', '--size', nargs='?', type=int, help='Size of files to download.  Default is 10240.', default=10240)
```
mikeharder, line 36: The .NET framework provides a few base classes with common options like `size`:

https://github.com/Azure/azure-sdk-for-net/blob/master/common/Perf/Azure.Test.Perf/SizeOptions.cs

The python command-line arguments work differently (no base class), but do you think Python should add helper methods for common options, or just duplicate the option in every class that needs it?

sdk/storage/azure-storage-file-share/tests/perfstress/download_to_file.py
```diff
@@ -0,0 +1,40 @@
...
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for license information.
+# --------------------------------------------------------------------------------------------
+
+import tempfile
+
+from ._test_base import _ShareTest
+
+
+class DownloadToFileTest(_ShareTest):
```
mikeharder, line 11: Does this test add any value over `DownloadTest`?  It looks like the only difference is calling `stream.readall()` instead of `stream.readinto(fp)`, and we are probably not interested in measuring the perf difference between these two APIs, unless we have reason to believe the stream we return might have different perf in these scenarios.

The important thing here is we are calling the exact same SDK API, the only difference is what we do with the returned stream.  If the SDK provided an API to download directly to file, that might be more interesting to test.

mikeharder: I see that Track1 did have an API to download directly to a file, so it probably does make sense to have an equivalent in Track2.

mikeharder (changes requested):

sdk/storage/azure-storage-file-share/tests/perfstress/_test_base_legacy.py
```diff
@@ -0,0 +1,52 @@
...
+# --------------------------------------------------------------------------------------------
+
+import os
+import uuid
+
+from azure_devtools.perfstress_tests import PerfStressTest
+
+from azure.storage.file import FileService
+
+class _LegacyServiceTest(PerfStressTest):
```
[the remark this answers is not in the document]
annatisch: I've moved the T1 tests into a separate directory for clarity.

sdk/storage/azure-storage-file-share/tests/perfstress/_test_base.py
```diff
@@ -0,0 +1,93 @@
...
+
+class _ServiceTest(PerfStressTest):
+    service_client = None
+    async_service_client = None
+
+    def __init__(self, arguments):
+        super().__init__(arguments)
+
+        connection_string = os.environ.get("AZURE_STORAGE_CONNECTION_STRING")
+        if not connection_string:
```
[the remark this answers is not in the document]
annatisch: Good plan - added!

sdk/storage/azure-storage-file-share/tests/perfstress/download.py
annatisch: I can add helper methods to make adding the arguments more straightforward - though I would prefer to keep the addition of arguments on an "as needed" basis.

sdk/storage/azure-storage-file-share/tests/perfstress/download_to_file.py
annatisch: Yeah - T1 had dedicated file APIs for upload and download - which we have removed and demonstrated with samples for handling files. So would be good to ensure we haven't removed perf in the process.

sdk/storage/azure-storage-file-share/tests/perfstress/download.py
mikeharder: In .NET it was fairly natural to create base classes like `SizeOptions` that tests can use if they want a `Size` option.  If this doesn't really make sense in Python you can skip it.

sdk/storage/azure-storage-file-share/tests/perfstress/_test_base.py
annatisch: Hah - this was here because you had it in the original blobs test prototypes - so I assumed you had a particular reason and kept it around. Though I can remove it :)

sdk/storage/azure-storage-file-share/tests/perfstress/download.py
annatisch: Done!

sdk/storage/azure-storage-file-share/tests/perfstress/download.py
annatisch: On closer inspection I don't think we gain a lot by adding helpers for adding the arguments - so I think I will leave this as is

sdk/storage/azure-storage-file-share/tests/perfstress/_test_base.py
mikeharder: I recall it was something we wanted to test at one point.  Maybe related to connection sharing?  Let's keep this enabled for now and get results both ways and compare them.

annatisch: Yeah - that sounds familiar. If nothing else it will be interesting to run as a comparison.
I've added it back it, but renamed it to something a bit easier to type on the cmd line (`--no-client-share`)

tools/azure-devtools/src/azure_devtools/perfstress_tests/random_stream.py
```diff
@@ -0,0 +1,35 @@
+# --------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for license information.
+# --------------------------------------------------------------------------------------------
+
+import os
+
+def get_random_bytes(buffer_length):
```
mikeharder, line 8: Since this is a one-line function call either way, I would be fine directly using `os.urandom()` in the test code.  Or we can keep this helper function.  Whatever you prefer.

mikeharder (approved):

tools/azure-devtools/src/azure_devtools/perfstress_tests/random_stream.py
annatisch: I might keep the function for now in case I want to later use `bytearray` instead of `bytes`

scbedd (approved):
TBH I don't see anything wrong here. Honestly quite like a bunch of the abstraction :) 

I think my final question is this. What does it look like to configure an actual test in our runs?

Just create a new test file under whichever package directory (inheriting from the appropriate test case) and just writing your test up as any other `AzureMgmtTestCase`?

PR #15549, opened 2020-11-25
Adding .env file support for testing

seankane-msft:
adding python-dotenv to the test-tools, and loading in init of `AzureTestCase`

### DIFF
--- a/tools/azure-sdk-tools/devtools_testutils/azure_testcase.py
+++ b/tools/azure-sdk-tools/devtools_testutils/azure_testcase.py
@@ -15,6 +15,7 @@ except ImportError:
     from inspect import getargspec as get_arg_spec
 
 import pytest
+from dotenv import load_dotenv, find_dotenv
 
 from azure_devtools.scenario_tests import (
     ReplayableTest, AzureTestError,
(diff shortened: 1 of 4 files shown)


tools/azure-sdk-tools/devtools_testutils/azure_testcase.py
```diff
@@ -101,6 +102,7 @@ def __init__(self, method_name, config_file=None,
         config_file = config_file or os.path.join(self.working_folder, TEST_SETTING_FILENAME)
         if not os.path.exists(config_file):
             config_file = None
+        load_dotenv(find_dotenv())
```
weshaggard, line 105: What type of env variables do we expect to be loaded? We need to be very careful about having secrets in env files that might get committed to the repo.

seankane-msft: I will add the .env file to the .gitignore as well. We're expecting devs to put any variables that they use for authenticating so it will be important that the `.env` file is not committed to the repo.

weshaggard: Adding it to the .gitignore is a good helper step but we should try and encourage folks to not even put the .env file under the git repo root. They should go into the user protected settings somewhere.

seankane-msft: I agree that is a great way to do it, I ran a quick test and placing my `.env` file in one directory before `azure-sdk-for-python` the environment variables loaded perfectly.

weshaggard: Great so when we document this we should suggest people put that file outside of the repo. Preferably in a user directory if that is a probing path of the python-dotenv library but if not at least in a higher level would be more secure.

lmazuel (approved):

PR #15591, opened 2020-12-01
Disable Version Increment -- Management Packages

scbedd:
Pending agreement that this doesn't fit in their workflow.

### DIFF
--- a/eng/versioning/version_shared.py
+++ b/eng/versioning/version_shared.py
@@ -34,7 +34,7 @@ DEV_STATUS_REGEX = r'(classifiers=\[(\s)*)(["\']Development Status :: .*["\'])'
 logging.getLogger().setLevel(logging.INFO)
 
 def path_excluded(path):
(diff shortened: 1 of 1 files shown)


eng/versioning/version_shared.py
```diff
@@ -34,7 +34,7 @@
 logging.getLogger().setLevel(logging.INFO)
 
 def path_excluded(path):
```
weshaggard, line 36: What all does this exclude? I suspect it disables more then just the version increment.

weshaggard: Nevermind I guess I should look at the script name. I only looked at the method name.

weshaggard (approved):
Change looks reasonable. I must say I was expecting a change somewhere in the yml files though but I suppose this will do the job as well. 

Of course this is assuming they agree they don't want this versioning strategy.

scbedd:
They agreed. Disabling.

Issue #15555, opened 2020-11-26
[ServiceBus][7.0.0][Bug] get_queue_receiver will lock 1 message


DzcleisureCN:
- **Package Name**: 
   azure-servicebus
- **Package Version**: 
   7.0.0
- **Operating System**:
   Centos
- **Python Version**: 
   3.7
**Describe the bug**
   get_queue_receiver will lock 1 message in the queue.

**To Reproduce**
1. #There are 5 messages in the queue if use below command can get 5 message.
```
with ServiceBusClient.from_connection_string(connstr) as client:
    with client.get_queue_receiver(queue_name, prefetch=10) as receiver:
        received_message_array = receiver.receive_messages(max_message_count=5, max_wait_time=10)
```
2. But if the use below command can only get 4 messages it seems get_queue_receiver will lock 1 message.
```
#There are 5 messages in the queue if use below command can get 4 message.
with ServiceBusClient.from_connection_string(connstr) as client:
    with client.get_queue_receiver(queue_name, prefetch=10) as receiver:
        pass
    with client.get_queue_receiver(queue_name, prefetch=10) as receiver:
        received_message_array = receiver.receive_messages(max_message_count=5, max_wait_time=10)
```
3. Why I ask this because I want separate "receive the message and complete message",  but now the complete message must have a receiver, so I use a get_queue_receiver but do not use receive_messages to delete a message, but get_queue_receiver  will lock 1 message in the queue.
```
#I just want like SQS can delete the message anywhere not after receive_messages then delete.
with ServiceBusClient.from_connection_string(connstr) as client:
   #just open a receiver and(do not use receive_messages)  to complete the message but seems lock 1 message in the queue.
    with client.get_queue_receiver(queue_name, prefetch=10) as receiver:
        receiver.complete_message(message)--> this message is input.
```
**Expected behavior**
Also can get 5 messages.
```
#There are 5 messages in the queue if use below command can get 4 message.
with ServiceBusClient.from_connection_string(connstr) as client:
    with client.get_queue_receiver(queue_name, prefetch=10) as receiver:
        pass
    with client.get_queue_receiver(queue_name, prefetch=10) as receiver:
        received_message_array = receiver.receive_messages(max_message_count=5, max_wait_time=10)
```

**Screenshots**

**Additional context**

DzcleisureCN:
Add another question 
```
with ServiceBusClient.from_connection_string(connstr) as client:
[...]

PR #15622, opened 2020-12-03
[ServiceBus] Prevent receiver from receiving messages during connection establishment

yunhaoling:
Current sb receiver would receive message (the amount is decided by the prefetch count) during connection establishment (see issue https://github.com/Azure/azure-sdk-for-python/issues/15555).
This is because the c implementation would immediately send a `FLOW` out when a receiver get attached and the service would send the client messages when the flow link credit > 0, see [link.c](https://github.com/Azure/azure-uamqp-python/blob/master/src/vendor/azure-uamqp-c/src/link.c#L330-L360)

This is a workaround/trying to address the issue.

Wondering if this could be moved into uamqp to be more cohesive. But need to be very careful as it might break existing uamqp behavior.

In worst case, this could just be a fixing in service bus. other sdks like event hub won't be affected by this problem as there's no lock/settlement feature.


yunhaoling:
Closing this PR as the fix would be carried in the release of uamqp 1.2.13.

PR #15629, opened 2020-12-03
[T2] azurestack

00Kai0:


jsntcy:
@00Kai0 do we also need update readme.md and _version.py?

PR #15698, opened 2020-12-08
[T2] Automanage

00Kai0:


00Kai0:
Automanage already has a track2 sdk, so I will close this PR.

PR #15712, opened 2020-12-08
Create Initial cut at an early preview (machine generated) version of python SDK.

LarryOsterman:
Created an extremely early preview of the Microsoft Azure Attestation service's dataplane (administrative) API surface.

This is an early preview of the SDK, basically just the machine generated API surface for python. Early in 2021, this preview SDK will be updated with the functionality more appropriate for a high quality Azure dataplane SDK.


sdk/attestation/azure-security-attestation/setup.py
```diff
@@ -0,0 +1,88 @@
...
+        'Programming Language :: Python :: 3.7',
+        'Programming Language :: Python :: 3.8',
+        'License :: OSI Approved :: MIT License',
+    ],
+    zip_safe=False,
+    packages=find_packages(exclude=[
+        'tests',
+        # Exclude packages that will be covered by PEP420 or nspkg
+        'azure',
+        'azure.mgmt',
```
johanste, line 78: @lmazuel, is this coming from a template? Or is it plain ole' copy/paste errors that creep in? Either way, we need to have an `azure-security-nspkg` package in addition to changing the `azure.mgmt` to `azure.security` for the list of packages to `exclude`.

LarryOsterman: copy&paste. Fixing.

LarryOsterman: Where would an `azure-security-nspkg package` come from?

johanste: > Where would an azure-security-nspkg package come from?

From @lmazuel :). In all seriousness, the central azure-sdk team needs to publish one.

LarryOsterman: Is the absence of this going to prove to be a problem? I can rename the client azure.attestation instead of azure.security.attestation for now (since it's an early preview sdk anyway)..

lmazuel (approved):

PR #15715, opened 2020-12-08
[Cosmos] Add option for disabling auto id generation in create_item

zfoster:
Replaces hardcoded automaticIdGeneration flag with option in container `create_item`.

Does anyone familiar with this project know why this was flagged True?

cc @annatisch @Rodrigossz @southpolesteve


sdk/cosmos/azure-cosmos/azure/cosmos/container.py
```diff
@@ -462,6 +462,7 @@ def create_item(
         pre_trigger_include=None,  # type: Optional[str]
         post_trigger_include=None,  # type: Optional[str]
         indexing_directive=None,  # type: Optional[Any]
+        disable_automatic_id_generation=None, #type: Optional[bool]
```
annatisch, line 465: Could this be moved into `kwargs`?
(The `:param` label in the docstring would move to `:keyword` to match.

Rodrigossz: Maybe it's true to make it easier for those that don't have an id to use.

zfoster: @annatisch I see no reason why not, however all of the other similar args follow this pattern

annatisch: Thanks @zfoster - we already have more positional optional parameters in the method signatures than we would generally advise, as making these parameters keyword only gives us more options for non-breaking changes in the future.
So going forward we should generally add new parameters to `kwargs` unless there's a very compelling reason not to.

zfoster: @annatisch do you happen to know why this is the case in python?

https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/cosmos/azure-cosmos/test/test_crud.py#L698

zfoster: or by chance know who _would_ know? :)

Rodrigossz: Spark Connector creates an id by default. Maybe it's for consistency.

zfoster: It's doing the opposite here I think? We have hardcoded to DISABLE ID generation

southpolesteve: @zfoster I think that comment is just wrong. We should match whatever the JS SDK does here.

sdk/cosmos/azure-cosmos/azure/cosmos/container.py
```diff
@@ -475,6 +475,7 @@ def create_item(
         :param pre_trigger_include: trigger id to be used as pre operation trigger.
         :param post_trigger_include: trigger id to be used as post operation trigger.
         :param indexing_directive: Indicate whether the document should be omitted from indexing.
+        :keyword bool disable_automatic_id_generation: Enable automatic id generation if no id present.
```
annatisch, line 478: The default value is True right? If we want non-breaking behaviour, we will probably need to keep automatic generation as disabled by default. So I think we could do that one of two ways:
- Keep the parameter called `disable_auto_id_gen`, and document the default as being `True` (and set it as such if not present).
- Rename the parameter to `enable_auto_id_gen`, document the default as being `False`, and set the request option to the opposite.

sdk/cosmos/azure-cosmos/azure/cosmos/container.py
```diff
@@ -488,7 +489,6 @@ def create_item(
         request_options = build_options(kwargs)
         response_hook = kwargs.pop('response_hook', None)
 
-        request_options["disableAutomaticIdGeneration"] = True
```
annatisch, line 491: As I mentioned above - we probably need to keep this here to ensure consistent behaviour for existing customers, allowing the parameter overwrite:
```python
request_options["disableAutomaticIdGeneration"] = kwargs.pop('disable_auto_id_gen', True)
```

annatisch (changes requested):

sdk/cosmos/azure-cosmos/azure/cosmos/container.py
zfoster: that makes sense - I was thinking this wouldn't affect users since it requires them to use an ID already, but I can see people relying on getting an error when they do not pass an ID

zfoster:
@annatisch how does this look now?

zfoster:
@annatisch can you take another look at this?

sdk/cosmos/azure-cosmos/azure/cosmos/container.py
annatisch: Thanks @zfoster - just to clarify, when you say "requires them to use an ID already", whereabouts is that?

zfoster:
``` 
request_options["disableAutomaticIdGeneration"] = True
````

having this hardcoded means you have to supply an ID in the args when creating items

Comment (zfoster) [reply sdk/cosmos/azure-cosmos/azure/cosmos/container.py:491]: if i'm reading it right

Comment (annatisch) [reply sdk/cosmos/azure-cosmos/azure/cosmos/container.py:491]: Interesting - in that case what is the keyword argument people would need to have been supplying? And is that what our tests are doing? (sorry I should have dug deeper into this earlier)

Comment (zfoster) [reply sdk/cosmos/azure-cosmos/azure/cosmos/container.py:491]: we didn't have a kwarg for it afaict. honestly, I'm not sure why this is the way it is, it's unique to this sdk. Our tests do supply IDs and from this we can assume our users do too, however I can flesh out why I think your comment about ensuring consistent behavior is still important:

We may have users who generate IDs or create them programmatically, and rely on the fact that we don't ever automatically generate them. The way this is set up now, we keep this consistent and never generate IDs unless they explicitly pass the kwarg in (same behavior as before)

**Review by southpolesteve — APPROVED** [review]
Approved. As I understand it, this is keeping the existing behavior and adding an option to disable it. 

I get why this if confusing. I think this was just done backwards in the original SDK. If we ever do a 5.0 we should look at flipping the behavior to be in sync with the other Cosmos SDKs.

File: sdk/cosmos/azure-cosmos/azure/cosmos/container.py (hunk 2 above)
Comment (annatisch) [reply sdk/cosmos/azure-cosmos/azure/cosmos/container.py:478]: @zfoster - on thinking about this a bit more - I think we should use the latter option - as I believe it conveys the behaviour more intuitively. i.e. update the kwarg name to `enable_auto_id_generation`, and update the docstring to state explicitly that this is `False` by default.
It would mean also updating the implementation to the opposite:
```
request_options["disableAutomaticIdGeneration"] = not kwargs.pop('disable_auto_id_gen', True)
```
Could we please also add a test (or update an existing test) to make sure that this new flag is tested?

File: sdk/cosmos/azure-cosmos/azure/cosmos/container.py [hunk 4]
```diff
@@ -488,7 +489,7 @@ def create_item(
...
+        request_options["enableAutomaticIdGeneration"] = kwargs.pop('enable_automatic_id_generation', False)
```
annatisch, line 492: The request_options key here is a service value - not sure it can be changed like this?
I think you'll need something like:
```python
        request_options["disableAutomaticIdGeneration"] = not kwargs.pop('enable_automatic_id_generation', False)
```

zfoster: yeah definitely -  I can't push at the moment, having issues authenticating to github. I slapped something together and then was testing pushing. I'll reping this when ready for review

annatisch: Awesome, thanks! :) 
Good luck with the github trouble - let me know if you need me to make any updates from my end.

zfoster:
@annatisch I think this is all set and good for another review

sdk/cosmos/azure-cosmos/azure/cosmos/_cosmos_client_connection.py
```diff
@@ -1072,20 +1072,22 @@ def CreateItem(self, database_or_container_link, document, options=None, **kwarg
...
         # link in case of client side partitioning
-        if base.IsItemContainerLink(database_or_container_link):
-            options = self._AddPartitionKey(database_or_container_link, document, options)
-
         collection_id, document, path = self._GetContainerIdWithPathForItem(
             database_or_container_link, document, options
         )
+
+        if base.IsItemContainerLink(database_or_container_link):
+            print('checking link', document)
```
annatisch, line 1088: A print statement ;)
Additionally - why was this block moved down? Does the ordering change here affect the `options`?

sdk/cosmos/azure-cosmos/azure/cosmos/container.py
```diff
@@ -508,8 +509,10 @@ def create_item(
         if indexing_directive is not None:
             request_options["indexingDirective"] = indexing_directive
 
+        passthrough_args = {k: v for k, v in kwargs.items() if k != "enable_automatic_id_generation"}
```
annatisch, line 512: This like shouldn't be necessary as `enable_automatic_id_generation` has already been `popped` from the `kwargs` above.

zfoster: ah I think I saw this error before I understood what the difference between the `whl` and `sdist` tests were. I never rebuilt so I kept getting the error :) thanks!

sdk/cosmos/azure-cosmos/azure/cosmos/_cosmos_client_connection.py
zfoster: Oops, I thought that print statement was already there when I reordered them. Let me remove! The order changed because the ID is generated and added to the body in the statement above this one. This statement then adds the `partitionKey` to the options. In the case that `partitionKey` is `/id`, we won't have the `partitionKey` in the options unless the statements are in this order

zfoster: Fixed

sdk/cosmos/azure-cosmos/azure/cosmos/container.py
zfoster: fixed

annatisch (approved):

PR #15733, opened 2020-12-09
[Communication] Remove live test_send_typing_notification tests for chat SDK

turalf:


ankitarorabit (approved):
I don't know why these are being removed but assuming there is a good reason, the change is fine.

PR #15750, opened 2020-12-10
[Communication] - Identity- Added AAD support

jbeauregardb:


sdk/communication/azure-communication-administration/azure/communication/administration/_communication_identity_client.py
```diff
@@ -152,4 +153,16 @@ def revoke_tokens(
             user.identifier if user else None,
             tokens_valid_from=issued_before,
             **kwargs)
- 
\ No newline at end of file
+    
+    def get_authentication_policy(self, endpoint, credential):
```
turalf, line 157: it seems that this method is identical with the async version. Instead of copying I would recommend, creating `_utils.py` file and add that method there

sdk/communication/azure-communication-administration/tests/test_communication_identity_client.py
```diff
@@ -65,3 +75,6 @@ def test_delete_user(self, connection_string):
         identity_client.delete_user(user)
 
         assert user.identifier is not None
+    
+    def get_endpoint_from_connection_string(self, connection_string):
```
turalf, line 79: inside `_shared/utils` there is a method for parsing connection string. Let's use that.

sdk/communication/azure-communication-administration/azure/communication/administration/_communication_identity_client.py
turalf: Actually, it is a better idea to add this to the `_shared/utils` as it seems other packages might use that as well?

jbeauregardb: Solved it!

sdk/communication/azure-communication-administration/azure/communication/administration/_shared/utils.py
```diff
@@ -68,3 +68,21 @@ def create_access_token(token):
         return AccessToken(token, datetime.fromtimestamp(payload['exp']).replace(tzinfo=TZ_UTC))
     except ValueError:
         raise ValueError(token_parse_err_msg)
+
+def get_authentication_policy(
+        endpoint, # type: str
+        credential # type: TokenCredential or str
+    ):
```
turalf, line 75: it would also be useful to add a documentation for this method, and let the caller decide without checking the content of the method. Something like [this](https://github.com/Azure/azure-sdk-for-python/blob/0c5b1408d876464e268258d13cab43b965a58bc9/sdk/communication/azure-communication-administration/azure/communication/administration/_phone_number_administration_client.py#L115)

sdk/communication/azure-communication-administration/azure/communication/administration/_shared/utils.py
```diff
@@ -68,3 +68,21 @@ def create_access_token(token):
...
+    # type: (...) -> SansIOHTTPPolicy
```
turalf, line 76: It seems it does not return SansIOHTTPPolicy, right?

turalf (approved):
lgtm

PR #15758, opened 2020-12-11
Enforce LF-only line endings

bluca:
Yet more files slipped in with CRLF line endings, breaking everything downstream.
Force everything to be committed only with LF to solve the problem once and
for all.


bluca:
ping

PR #15764, opened 2020-12-11
[Key Vault] Add migration guide for azure-keyvault-certificates

mccoyp:
Part of #15118.

### DIFF
--- a/sdk/keyvault/azure-keyvault-certificates/migration_guide.md
+++ b/sdk/keyvault/azure-keyvault-certificates/migration_guide.md
@@ -59,7 +59,7 @@ credentials = ServicePrincipalCredentials(
     tenant="tenant id"
 )
 
-certificate_client = KeyVaultClient(credentials)
+client = KeyVaultClient(credentials)
 ```
 
 Now in `azure-keyvault-certificates` you can create a `CertificateClient` using any credential from [`azure-identity`](https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/identity/azure-identity/README.md). Below is an example using [`DefaultAzureCredential`](https://docs.microsoft.com/python/api/azure-identity/azure.identity.defaultazurecredential?view=azure-python):
@@ -101,7 +101,7 @@ async with certificate_client:
 In `azure-keyvault` you could create a certificate by using `KeyVaultClient`'s `create_certificate` method, which required a vault endpoint and certificate name. This method returned a `CertificateOperation`.
 
 ```python
-operation = certificate_client.create_certificate(
+operation = client.create_certificate(
     vault_base_url="https://my-key-vault.vault.azure.net/",
     certificate_name="cert-name"
 )
@@ -126,13 +126,13 @@ If you would like to check the status of your certificate creation, you can call
 In `azure-keyvault` you could retrieve a certificate (as a `CertificateBundle`) by using `get_certificate` and specifying the desired vault endpoint, certificate name, and certificate version. You could retrieve the versions of a certificate with the `get_certificate_versions` method, which returned an iterator-like object.
 
 ```python
-certificate_versions = certificate_client.get_certificate_versions(
+certificate_versions = client.get_certificate_versions(
     vault_base_url="https://my-key-vault.vault.azure.net/",
     certificate_name="cert-name"
 )
 
 for certificate_version in certificate_versions:
-    certificate = certificate_client.get_certificate(
+    certificate = client.get_certificate(
         vault_base_url="https://my-key-vault.vault.azure.net/",
         certificate_name="cert-name",
         certificate_version=certificate_version
@@ -160,7 +160,7 @@ certificate = certificate_client.get_certificate_version(certificate_name="cert-
 In `azure-keyvault` you could list the properties of certificates in a specified vault with the `get_certificates` method. This returned an iterator-like object containing `CertificateItem` instances.
 
 ```python
-certificates = certificate_client.get_certificates(vault_base_url="https://my-key-vault.vault.azure.net/")
+certificates = client.get_certificates(vault_base_url="https://my-key-vault.vault.azure.net/")
 
 for certificate in certificates:
     print(certificate.attributes.created)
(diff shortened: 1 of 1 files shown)


sdk/keyvault/azure-keyvault-certificates/migration_guide.md
```diff
@@ -0,0 +1,288 @@
...
+- [`azure-keyvault-keys`](https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/keyvault/azure-keyvault-keys/README.md) contains `KeyClient` for working with keys and `CryptographyClient` for performing cryptographic operations.
+- [`azure-keyvault-secrets`](https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/keyvault/azure-keyvault-secrets/README.md) contains `SecretClient` for working with secrets.
+
+### Client constructors
+
+Across all modern Azure client libraries, clients consistently take an endpoint or connection string along with token credentials. This differs from `KeyVaultClient`, which took an authentication delegate and could be used for multiple Key Vault endpoints.
+
+#### Authenticating
+
+Previously in `azure-keyvault` you could create a `KeyVaultClient` by using `ServicePrincipalCredentials` from `azure-common-credentials`:
```
chlowell, line 50: This is kind of tricky. Typically, you'd import the class from `azure.common` but actually it's [defined in msrestazure](https://github.com/Azure/azure-sdk-for-python/blob/3808ad57daced04b57ff43018045b977105573ab/sdk/core/azure-common/azure/common/credentials.py#L116) and [the documentation](https://docs.microsoft.com/python/api/msrestazure/msrestazure.azure_active_directory?view=azure-python) is associated with that package. And azure-common-credentials doesn't exist. So maybe it's clearest to use the namespace here:
```suggestion
Previously in `azure-keyvault` you could create a `KeyVaultClient` by using `ServicePrincipalCredentials` from `azure.common`:
```

sdk/keyvault/azure-keyvault-certificates/migration_guide.md
```diff
@@ -0,0 +1,288 @@
...
+
+### Create a certificate
+
+In `azure-keyvault` you could create a certificate by using `KeyVaultClient`'s `create_certificate` method, which required a vault endpoint and certificate name. This method returned a `CertificateOperation`.
+
+```python
+from azure.common.credentials import ServicePrincipalCredentials
+from azure.keyvault import KeyVaultClient
+
+credentials = ServicePrincipalCredentials(
```
chlowell, line 107: Do we need to include authentication in every example?

sdk/keyvault/azure-keyvault-certificates/migration_guide.md
```diff
@@ -0,0 +1,288 @@
...
+
+credential = DefaultAzureCredential()
+
+certificate_client = CertificateClient(vault_url="https://my-key-vault.vault.azure.net/", credential=credential)
+
+create_certificate_poller = certificate_client.begin_create_certificate(
+    certificate_name="cert-name",
+    policy=CertificatePolicy.get_default()
+)
+certificate = create_certificate_poller.result()
```
chlowell, line 135: Might warrant a comment that this blocks until the operation completes (because that could take days).

sdk/keyvault/azure-keyvault-certificates/migration_guide.md
```diff
@@ -0,0 +1,288 @@
...
+certificate_client = KeyVaultClient(credentials)
+
+deleted_certificate = certificate_client.delete_certificate(
+    vault_base_url="https://my-key-vault.vault.azure.net/",
+    certificate_name="cert-name"
+)
+
+# wait for the deletion to complete
+
+certificate_client.purge_deleted_certificate(
```
chlowell, line 262:
```suggestion
# this would fail when deletion hadn't finished
certificate_client.purge_deleted_certificate(
```
i.e. continuing the idea of "the old client had this pitfall we've bridged in the new client"

sdk/keyvault/azure-keyvault-certificates/migration_guide.md
mccoyp: This makes sense, thanks for the context! I wasn't sure about the authentication story since I never used the track 1 library, so this is really helpful

sdk/keyvault/azure-keyvault-certificates/migration_guide.md
mccoyp: Probably not, since there's a section about authenticating the client. I'll go ahead and shorten the examples

sdk/keyvault/azure-keyvault-certificates/migration_guide.md
```diff
@@ -0,0 +1,210 @@
...
+)
+
+# this purge would fail if deletion hadn't finished
+client.purge_deleted_certificate(
+    vault_url="https://my-key-vault.vault.azure.net/",
+    certificate_name="cert-name"
+)
+```
+
+Now in `azure-keyvault-certificates` you can delete a certificate with `begin_delete_certificate`, which returns a long operation poller object that can be used to wait/check on the operation much like you would with the poller received from `begin_create_certificate`. Calling `result()` on the poller will return information about the deleted certificate (as a `DeletedCertificate`) without waiting for the operation to complete, but calling `wait()` will wait for the deletion to complete. Again, `purge_deleted_certificate` will permanently delete your deleted certificate and make it unrecoverable.
```
chlowell, line 197: I might not mention `begin_create_certificate` here, to avoid muddying the waters. `result()` is blocking on that poller but not on the one returned by `begin_delete_certificate`.

sdk/keyvault/azure-keyvault-certificates/migration_guide.md
```diff
@@ -0,0 +1,210 @@
...
+certificates = certificate_client.list_properties_of_certificates()
+
+for certificate in certificates:
+    print(certificate.name)
+    print(certificate.x509_thumbprint)
+```
+
+### Delete a certificate
+
+In `azure-keyvault` you could delete all versions of a certificate with the `delete_certificate` method. This returned information about the deleted certificate (as a `DeletedCertificateBundle`), but you you could not poll the deletion operation to know when it completed. This would be valuable information if you intended to permanently delete the deleted certificate with `purge_deleted_certificate`.
```
chlowell, line 182:
```suggestion
In `azure-keyvault` you could delete all versions of a certificate with the `delete_certificate` method. This returned information about the deleted certificate (as a `DeletedCertificateBundle`), but you could not poll the deletion operation to know when it completed. This would be valuable information if you intended to permanently delete the deleted certificate with `purge_deleted_certificate`.
```

chlowell (approved):

PR #15782, opened 2020-12-14
[Storage]Avoid incorrect padding Error when base64 decode account key

xiafu-msft:


xf58:
@tasherif-msft it looks like CLI got a workaround so this pr is not needed.

PR #15786, opened 2020-12-14
[EventHubs] Add unit test for 32 partition greedy ownership to attempt and validate issue #13546 having been fixed

### LINKED ISSUE #13546: Not all partitions are assigned to consumer group if partitions % instances != 0
- **Package Name**: azure-eventhub
- **Package Version**:
azure-eventhub==5.1.0
azure-eventhub-checkpointstoreblob==1.1.0
- **Operating System**: Linux
- **Python Version**: 3.6

**Describe the bug**
Having 6 instances of a process consuming records from an EventHub topic with 32 partitions (using the same consumer group) results in every instance being assigned to 5 partitions, with 32 - 6 * 5 = 2 partitions staying unassigned to any consumer, even hours after the processes have been launched. No partitions remain unassigned if either 4 or 8 instances are used instead (we hypothesize the reason is that 32 % 8 == 0, 32 % 4 == 0, but 32 % 6 != 0).

**To Reproduce**
1. Fill in the EVENT_HUB_MONITOR_* constants within the given Python code with proper values.
2. Execute 6 instances of the given Python code in parallel.
3. Put the output of every instance together to figure out how many partitions remain unassigned.

The source code of the program follows:

[EventHubMonitor.txt](https://github.com/Azure/azure-sdk-for-python/files/5169353/EventHubMonitor.txt)

**Expected behavior**
The expected behavior would be that 4 instances are assigned to 5 partitions, and 2 instances are assigned to 6 partitions, for a total of 4 * 5 + 2 * 6 = 32 assigned partitions.

**Screenshots**
I include screenshots that show the output of each one of the 6 instances after 5 minutes have passed since they were launched.

Screenshot_20200903_164612
Screenshot_20200903_164635
Screenshot_20200903_164647
Screenshot_20200903_164655
Screenshot_20200903_164704
Screenshot_20200903_164711

**Additional context**
While we are using the Azure Blob check-pointing mechanism to balance the consumers, we do not use it to commit reading offsets. In practice, to run several instances of this process in parallel we use a ReplicationController within a Kubernetes cluster.

KieranBrantnerMagee:
Tests seem to indicate it's been dealt with via. latest load balancing changes.

closes #13546


sdk/eventhub/azure-eventhub/tests/livetest/synctests/test_eventprocessor.py
```diff
@@ -522,6 +522,8 @@ def claim_ownership(self, ownsership):
         (['ownership_expired', 'ownership_released'], ["0", "1", "2", "3"], 4),
         (['ownership_active0', 'ownership_active1'], ["0", "1"], 0),
         (['ownership_active0', 'ownership_self_owned'], ["0", "1"], 1),
+        (['ownership_active0', 'ownership_active1'], [str(i) for i in range(32)], 11),
```
yunhaoling, line 525: dumb question regarding to our greedy load balance algorithm:
why is this 11 (I guess it‘s 32/3 round up 11 ) not 16 as there are only two active owners?

yunhaoling: answered by yijun, it's two active ownership, so when running the code, there're actually three ownerships, the max expected per owner is 11

yunhaoling (approved):

Issue #15653, opened 2020-12-04
[Tables] Delete Entity Returning Unauthorized, Query Entities Not


sejohnson-at-griffis:
- **Package Name**:  azure-data-tables (reproducable in azure-cosmosdb-tables)
- **Package Version**: 12.0.0b3
- **Operating System**: Windows 10
- **Python Version**: 3.8.6

**Describe the bug**
After creating a TableClient from azure.data.tables.aio, a call to table_client.query_entities is successful. Just a few lines later, a call to table_client.delete_entity is not, and returns a ClientAutheniticationError: Server failed to authenticate the request. Make sure the value of Authorization header is formed correctly including the signature.

I've tried this with azure.cosmosdb.table as well: Using similar calls, the query call is successful, but the delete call returns an AzureSigningError: Incorrect padding.

I have verified that the key is 88 characters long, ends with ==, and is DIRECTLY copied from the storage account in question. It obviously works for the query_entities call.

**To Reproduce**
1. Create an Azure Storage Account, add a table to it, and name the table "Locator". Add a few entities with a column called LocatorHash, with any string value you like.
2. Run the following code:

```
    from azure.data.tables.aio import TableClient
    from azure.core import MatchingConditions

    table_name = 'Locator'
    email_hash = '...' #Can be any string value contained in the LocatorHash column
    table_client = TableClient(account_url='https://<YOURACCOUNTNAME>.table.core.windows.net', table_name=table_name, credential=os.environ['STORAGE_ACCOUNT_ACCESS_KEY'])
    async with table_client:
        locator_entities = table_client.query_entities(filter=f"LocatorHash eq '{email_hash}'")
        coroutines = []
        async for locator_entity in locator_entities:
            logging.info(f"Deleting record: {locator_entity}")
            # error occurs on this line
            await table_client.delete_entity(row_key=locator_entity['RowKey'], partition_key=locator_entity['PartitionKey'], etag=locator_entity['_metadata']['etag'], match_condition=MatchConditions.Unconditionally)
```

**Expected behavior**
Entities are deleted rather than receiving an AuthError.

**Screenshots**

*When using azure.data.tables.aio*
![image](https://user-images.githubusercontent.com/73712865/101196573-27d32580-361e-11eb-85e7-64d1bd034982.png)

*When using azure.cosmosdb.table -- note, had to edit code to be synchronous*
[...]

PR #15813, opened 2020-12-15
Customer at symbol issue

seankane-msft:
Addressed customer issue #15653


annatisch (approved):
Looks great! Could you add an update to the changelog documenting the bug fix with reference to the issue number?

PR #15814, opened 2020-12-15
Better error from InteractiveBrowserCredential for invalid redirect_uri

chlowell:
InteractiveBrowserCredential's optional `redirect_uri` argument must be a URL with a port. That's okay but the credential doesn't fail until it first tries to use the URL, so when you construct a credential with an invalid URL, you won't know it's invalid until you try to get a token and the error message is unhelpful:
```
>>> credential = InteractiveBrowserCredential(redirect_uri="http://localhost")
>>> credential.get_token("...")
azure.core.exceptions.ClientAuthenticationError: Authentication failed: an integer is required (got type NoneType)
```

This PR refactors the credential's usage of this argument such that the credential can validate the URL at initialization and raise a better error:
```
>>> credential = InteractiveBrowserCredential(redirect_uri="http://localhost")
ValueError: "redirect_uri" must be a URL with port number, for example "http://localhost:8400"
```

### DIFF
--- a/sdk/identity/azure-identity/azure/identity/_credentials/browser.py
+++ b/sdk/identity/azure-identity/azure/identity/_credentials/browser.py
@@ -50,6 +50,7 @@ class InteractiveBrowserCredential(InteractiveCredential):
     :keyword bool allow_unencrypted_cache: if True, the credential will fall back to a plaintext cache on platforms
           where encryption is unavailable. Default to False. Has no effect when `enable_persistent_cache` is False.
     :keyword int timeout: seconds to wait for the user to complete authentication. Defaults to 300 (5 minutes).
+    :raises ValueError: invalid `redirect_uri`
     """
 
     def __init__(self, **kwargs):
--- a/sdk/identity/azure-identity/azure/identity/_credentials/browser.py
+++ b/sdk/identity/azure-identity/azure/identity/_credentials/browser.py
@@ -6,6 +6,8 @@ import socket
 import uuid
 import webbrowser
 
+from six.moves.urllib_parse import urlparse
+
 from azure.core.exceptions import ClientAuthenticationError
 
 from .. import CredentialUnavailableError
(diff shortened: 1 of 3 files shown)


sdk/identity/azure-identity/azure/identity/_credentials/browser.py
```diff
@@ -52,7 +54,14 @@ class InteractiveBrowserCredential(InteractiveCredential):
 
     def __init__(self, **kwargs):
         # type: (**Any) -> None
-        self._redirect_uri = kwargs.pop("redirect_uri", None)
+        redirect_uri = kwargs.pop("redirect_uri", None)
+        if redirect_uri:
+            self._parsed_url = urlparse(redirect_uri)
+            if not (self._parsed_url.hostname and self._parsed_url.port):
+                raise ValueError('"redirect_uri" must be a URL with port number, for example "http://localhost:8400"')
```
mccoyp, line 61: Should there be a field added to the docstring documenting this error? Something like:
```
:raises ValueError: if `redirect_uri` is not a URL with a port number
```

sdk/identity/azure-identity/azure/identity/_credentials/browser.py
```diff
@@ -64,17 +73,17 @@ def _request_token(self, *scopes, **kwargs):
 
         # start an HTTP server to receive the redirect
         server = None
-        redirect_uri = self._redirect_uri
-        if redirect_uri:
+        if self._parsed_url:
             try:
-                server = self._server_class(redirect_uri, timeout=self._timeout)
+                redirect_uri = "http://{}:{}".format(self._parsed_url.hostname, self._parsed_url.port)
```
mccoyp, line 78: Really nitpicky, but should this line be moved into the `except` block since `redirect_uri` isn't used outside the error message?

mccoyp (approved):

sdk/identity/azure-identity/azure/identity/_credentials/browser.py
chlowell: It's not in the diff, but this is used on line 99 as an argument to msal's `get_authorization_request_url()`.

PR #15819, opened 2020-12-16
[Key Vault] Add migration guide for azure-keyvault-keys

mccoyp:
Part of #15118.

This also includes some (very) minor fixes to the azure-keyvault-keys README.

### DIFF
--- a/sdk/keyvault/azure-keyvault-keys/migration_guide.md
+++ b/sdk/keyvault/azure-keyvault-keys/migration_guide.md
@@ -252,7 +252,7 @@ operation_result = client.encrypt(
 ciphertext = operation_result.result
 ```
 
-Now in `azure-keyvault-keys` you can perform these cryptographic operations by using a `CryptographyClient`. The key used to create the client will be used for these operations. Cryptographic operations are now performed locally by the client, rather than remotely by Key Vault.
+Now in `azure-keyvault-keys` you can perform these cryptographic operations by using a `CryptographyClient`. The key used to create the client will be used for these operations. Cryptographic operations are now performed locally by the client when it's intialized with the necessary key material or is able to get that material from Key Vault, and are only performed by the Key Vault service when required key material is unavailable.
 
 ```python
 from azure.keyvault.keys.crypto import CryptographyClient, EncryptionAlgorithm
--- a/sdk/keyvault/azure-keyvault-keys/migration_guide.md
+++ b/sdk/keyvault/azure-keyvault-keys/migration_guide.md
@@ -1,0 +1,271 @@
+# Guide for migrating to azure-keyvault-keys from azure-keyvault
+
+This guide is intended to assist in the migration to `azure-keyvault-keys` from `azure-keyvault`. It will focus on side-by-side comparisons for similar operations between the two packages.
+
+Familiarity with the `azure-keyvault` package is assumed. For those new to the Key Vault client library for Python, please refer to the [README for `azure-keyvault-keys`](https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/keyvault/azure-keyvault-keys/README.md) rather than this guide.
+
+## Table of contents
+
+* [Migration benefits](#migration-benefits)
+* [Important changes](#important-changes)
+    - [Separate packages and clients](#separate-packages-and-clients)
+    - [Client constructors](#client-constructors)
+    - [Async operations](#async-operations)
+    - [Create a key](#create-a-key)
+    - [Retrieve a key](#retrieve-a-key)
+    - [List properties of keys](#list-properties-of-keys)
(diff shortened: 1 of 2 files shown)


sdk/keyvault/azure-keyvault-keys/migration_guide.md
```diff
@@ -0,0 +1,271 @@
...
+    kty="EC"
+)
+key = key_bundle.key
+```
+
+Now in `azure-keyvault-keys` there are multiple ways to create keys. You can provide a key name and type to the general `create_key` method, or provide just a name to `create_rsa_key` or `create_ec_key`. These methods all return the created key as a `KeyVaultKey`.
+
+```python
+# create a key with specified type
+key = key_client.create_key(name="key-name", key_type="oct")
```
chlowell, line 135: "oct" will get an error from a standard Key Vault. Also may be useful to show the `KeyType` enum here:
```suggestion
from azure.keyvault.keys import KeyType

# create a key with specified type
key = key_client.create_key(name="key-name", key_type=KeyType.ec)
```

sdk/keyvault/azure-keyvault-keys/migration_guide.md
```diff
@@ -0,0 +1,271 @@
...
+print(key.name)
+print(key.key_type)
+
+# create an RSA key
+rsa_key = key_client.create_rsa_key(name="rsa-key-name", size=2048)
+
+# create an elliptic curve key
+ec_key = key_client.create_ec_key(name="ec-key-name", curve="P-256")
```
chlowell, line 143: Suggest showing KeyCurveName here

sdk/keyvault/azure-keyvault-keys/migration_guide.md
```diff
@@ -0,0 +1,271 @@
...
+```python
+from azure.keyvault import KeyId
+
+key_items = client.get_key_versions(
+    vault_base_url="https://my-key-vault.vault.azure.net/",
+    key_name="key-name"
+)
+
+for key_item in key_items:
+    key_id = KeyId(key_item.kid)
```
chlowell, line 159: This isn't mentioned in the text or used in the example?

sdk/keyvault/azure-keyvault-keys/migration_guide.md
```diff
@@ -0,0 +1,271 @@
...
+    vault_base_url="https://my-key-vault.vault.azure.net/",
+    key_name="key-name",
+    key_version=key_version,
+    algorithm="RSA-OAEP-256",
+    value=plaintext
+)
+ciphertext = operation_result.result
+```
+
+Now in `azure-keyvault-keys` you can perform these cryptographic operations by using a `CryptographyClient`. The key used to create the client will be used for these operations.
```
chlowell, line 253: Another difference is that `CryptographyClient` performs operations locally when it has or can get the key material whereas every `KeyVaultClient` crypto operation is performed by Key Vault.

sdk/keyvault/azure-keyvault-keys/migration_guide.md
mccoyp: This variable isn't used directly, but I split up the process of fetching the key version because I wanted to clearly show how the `KeyId` was being used. I don't think condensing lines 159/160 into `key_version = KeyId(key_item.kid).version` would be too complicated though, if you think that's neater

chlowell: Ah, my bad, it's used in the sense I meant on the next line, it's fine like this 🤓

sdk/keyvault/azure-keyvault-keys/migration_guide.md
```diff
@@ -0,0 +1,273 @@
...
+    vault_base_url="https://my-key-vault.vault.azure.net/",
+    key_name="key-name",
+    key_version=key_version,
+    algorithm="RSA-OAEP-256",
+    value=plaintext
+)
+ciphertext = operation_result.result
+```
+
+Now in `azure-keyvault-keys` you can perform these cryptographic operations by using a `CryptographyClient`. The key used to create the client will be used for these operations. Cryptographic operations are now performed locally by the client, rather than remotely by Key Vault.
```
chlowell, line 255: This implies CryptographyClient does everything locally, but actually it falls back to Key Vault when it can't get the key material.

mccoyp: Hm, that makes sense. Is there documentation of the CryptographyClient's behavior that I could refer/link to, or would this have to be gleaned from the source code?

chlowell: Good question. It should be documented but is not: #15859

chlowell (approved):

PR #15833, opened 2020-12-16
[Perfstress][Storage] Added Blobs perf tests

annatisch:


tools/azure-devtools/src/azure_devtools/perfstress_tests/random_stream.py
```diff
@@ -29,7 +29,34 @@ def read(self, size=None):
...
+    def seek(self, index):
+        self._position = index
 
     def remaining(self):
-        return self._remaining
\ No newline at end of file
+        return self._remaining
+
+
+class WriteStream:
```
mikeharder, line 45: This is probably fast enough, but I wonder if it would still work and be even faster if it "cheated" and didn't bother to track length and position?  Where `write()` is just a no-op, `seekable` is false, `tell` always returns 0, etc.

mikeharder: With this implementation, I also think all tests could re-use a single static WriteStream.

mikeharder: But if the `WriteStream` needs to support length and seeking for correctness, then each test instance will need its own, but maybe it could just "reset" the stream between operations instead of creating a new one.

sdk/storage/azure-storage-blob/tests/perfstress_tests/T1_legacy_tests/list_blobs.py
```diff
@@ -0,0 +1,31 @@
...
+from azure_devtools.perfstress_tests import get_random_bytes
+
+from ._test_base_legacy import _LegacyContainerTest
+
+
+class LegacyListBlobsTest(_LegacyContainerTest):
+
+    async def global_setup(self):
+        await super().global_setup()
+        for i in range(self.args.num_blobs):
```
mikeharder, line 15: It should be faster to create many blobs in parallel using a reasonable number of threads (like 32).  Here's a helper function we wrote in JS for this purpose, Python should be able to use similar (with threads instead of async tasks):

https://github.com/mikeharder/blob-upload-parallel/pull/2/files#diff-0634662c6970258c08f765262734d321f014ad9a55ce51c0b9519b34fcbd6d7aR26-R49

sdk/storage/azure-storage-blob/tests/perfstress_tests/T1_legacy_tests/upload.py
```diff
@@ -0,0 +1,27 @@
...
+from ._test_base_legacy import _LegacyContainerTest
+
+
+class LegacyUploadTest(_LegacyContainerTest):
+    def __init__(self, arguments):
+        super().__init__(arguments)
+        self.blob_name = "blobtest-" + str(uuid.uuid4())
+
+    def run_sync(self):
+        data = RandomStream(self.args.size)
```
mikeharder, line 19: Can `RandomStream` be created once per test instance and reused?

sdk/storage/azure-storage-blob/tests/perfstress_tests/T1_legacy_tests/upload_block.py
```diff
@@ -0,0 +1,26 @@
...
+class LegacyUploadBlockTest(_LegacyContainerTest):
+    def __init__(self, arguments):
+        super().__init__(arguments)
+        self.blob_name = "blobblocktest-" + str(uuid.uuid4())
+
+    def run_sync(self):
+        self.service_client.put_block(
+            container_name=self.container_name,
+            blob_name=self.blob_name,
+            block=get_random_bytes(self.args.size),
```
mikeharder, line 22: Can random bytes be created once per instance and reused?

sdk/storage/azure-storage-blob/tests/perfstress_tests/T1_legacy_tests/upload_block.py
```diff
@@ -0,0 +1,26 @@
...
+            block_id=str(uuid.uuid4()))
```
mikeharder, line 23: Can we put to the same block id each time, or do they need to be unique each time?

sdk/storage/azure-storage-blob/tests/perfstress_tests/_test_base.py
```diff
@@ -0,0 +1,92 @@
...
+
+    async def global_setup(self):
+        await super().global_setup()
+        try:
+            await self.async_blob_client.delete_blob()
+        except ResourceNotFoundError:
+            pass
+
+    async def global_cleanup(self):
+        try:
```
mikeharder, line 84: Unnecessary since super will delete whole container

sdk/storage/azure-storage-blob/tests/perfstress_tests/download.py
```diff
@@ -0,0 +1,35 @@
...
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for license information.
+# --------------------------------------------------------------------------------------------
+
+from azure_devtools.perfstress_tests import get_random_bytes, WriteStream
+
+from ._test_base import _ContainerTest
+
+
+class DownloadTest(_ContainerTest):
```
mikeharder, line 11: Should this extend `_BlobTest` instead of `_ContainerTest`?

sdk/storage/azure-storage-blob/tests/perfstress_tests/download.py
```diff
@@ -0,0 +1,35 @@
...
+        self.blob_client = self.container_client.get_blob_client(blob_name)
+        self.async_blob_client = self.async_container_client.get_blob_client(blob_name)
+
+    async def global_setup(self):
+        await super().global_setup()
+        data = get_random_bytes(self.args.size)
+        await self.async_blob_client.upload_blob(data)
+
+    def run_sync(self):
+        download = WriteStream()
```
mikeharder, line 24: Can `WriteStream` be reused?

sdk/storage/azure-storage-blob/tests/perfstress_tests/list_blobs.py
```diff
@@ -0,0 +1,41 @@
...
+from azure_devtools.perfstress_tests import get_random_bytes
+
+from ._test_base import _ContainerTest
+
+
+class ListBlobsTest(_ContainerTest):
+
+    async def global_setup(self):
+        await super().global_setup()
+        pending = (asyncio.ensure_future(self.async_container_client.upload_blob("listtest" + str(i), data=b"")) for i in range(self.args.num_blobs))
```
mikeharder, line 17: I'm not sure how this works, but if it is creating the blobs in parallel and it works fine, it seems good to me :).

sdk/storage/azure-storage-blob/tests/perfstress_tests/upload.py
```diff
@@ -0,0 +1,28 @@
...
+# Licensed under the MIT License. See License.txt in the project root for license information.
+# --------------------------------------------------------------------------------------------
+
+from ._test_base import _BlobTest
+
+from azure_devtools.perfstress_tests import RandomStream, get_random_bytes
+from azure_devtools.perfstress_tests import AsyncRandomStream
+
+
+class UploadTest(_BlobTest):
```
mikeharder, line 12: I think your Upload and Download tests have the pattern reversed.  For Dowload, all instances can use the same blob in the service.  For upload, each instance should upload to its own blob.  In general, for read-only it's OK to share, but we should avoid concurrent writes to the same resource.

sdk/storage/azure-storage-blob/tests/perfstress_tests/upload.py
```diff
@@ -0,0 +1,28 @@
...
+
+    def run_sync(self):
+        data = RandomStream(self.args.size)
```
mikeharder, line 15: Can RandomStream be reused?

sdk/storage/azure-storage-blob/tests/perfstress_tests/upload_block.py
```diff
@@ -0,0 +1,27 @@
...
+# Licensed under the MIT License. See License.txt in the project root for license information.
+# --------------------------------------------------------------------------------------------
+import uuid
+
+from ._test_base import _BlobTest
+
+from azure_devtools.perfstress_tests import get_random_bytes
+
+
+class UploadBlockTest(_BlobTest):
```
mikeharder, line 12: Same feedback as upload

sdk/storage/azure-storage-blob/tests/perfstress_tests/upload_from_file.py
```diff
@@ -0,0 +1,37 @@
...
+
+import os
+import tempfile
+
+from azure_devtools.perfstress_tests import get_random_bytes
+
+from ._test_base import _BlobTest
+
+
+class UploadFromFileTest(_BlobTest):
```
mikeharder, line 14: Same feedback as Upload.

sdk/storage/azure-storage-blob/tests/perfstress_tests/upload_from_file.py
```diff
@@ -0,0 +1,37 @@
...
+        with tempfile.NamedTemporaryFile(delete=False) as temp_file:
+            UploadFromFileTest.temp_file = temp_file.name
+            temp_file.write(data)
+
+    async def global_cleanup(self):
+        os.remove(UploadFromFileTest.temp_file)
+        await super().global_cleanup()
+
+    def run_sync(self):
+        with open(UploadFromFileTest.temp_file, 'rb') as fp:
```
mikeharder, line 29: Is it interesting to test uploading from a file pointer if we are alreading testing upload from RandomStream?

mikeharder (changes requested):

tools/azure-devtools/src/azure_devtools/perfstress_tests/random_stream.py
annatisch: Yeah unfortunately the SDK makes use of seekable and tell for multi-threaded downloads - so I think to get the best indication of download perf we should include testing of those mechanisms.
Additionally I could add a "reset" feature to the stream to both this and the RandomStream - but I think I will use a dedicated `reset` function that will be run outside of the clock for each test iteration.

sdk/storage/azure-storage-blob/tests/perfstress_tests/T1_legacy_tests/upload_block.py
annatisch: That's how it originally was - however I then hit memory limits with 32 concurrent instances all holding a GB of data....

sdk/storage/azure-storage-blob/tests/perfstress_tests/T1_legacy_tests/upload_block.py
annatisch: Turns out it can be the same. Updating :)

sdk/storage/azure-storage-blob/tests/perfstress_tests/T1_legacy_tests/upload_block.py
annatisch: Actually I've made some improvements to hopefully resolve the memory issues

sdk/storage/azure-storage-blob/tests/perfstress_tests/upload.py
annatisch: I think I have the pattern correct (what you describe is what it should be doing) - as use of the base class `_BlobTest` indicates a unique blob to that instance.

sdk/storage/azure-storage-blob/tests/perfstress_tests/download.py
annatisch: As mentioned above when discussing the upload/download resource patter - the `_BlobTest` base represents a unique endpoint to that instance - whereas for download we don't need a unique resource.

sdk/storage/azure-storage-blob/tests/perfstress_tests/download.py
annatisch: I've added a reset to do this, so we can exclude this from the timed function

sdk/storage/azure-storage-blob/tests/perfstress_tests/list_blobs.py
annatisch: Yup - this just keeps the maximum concurrent operations at 16, and keeps adding new uploads to the pool when one finishes.

sdk/storage/azure-storage-blob/tests/perfstress_tests/upload.py
mikeharder: You are correct, somehow _I_ got this reversed when reading the code.

sdk/storage/azure-storage-blob/tests/perfstress_tests/T1_legacy_tests/download.py
```diff
@@ -0,0 +1,34 @@
...
+            blob_name=self.blob_name,
+            blob=data)
+
+    def run_sync(self):
+        self.service_client.get_blob_to_stream(
+            container_name=self.container_name,
+            blob_name=self.blob_name,
+            stream=self.download_stream,
+            max_connections=self.args.max_concurrency)
+        self.download_stream.reset()
```
mikeharder, line 31: NIT: I prefer to reset before use, since it ensures you are working with a reset stream.  But before or after are logically the same so whatever you prefer is also fine.

sdk/storage/azure-storage-blob/tests/perfstress_tests/T1_legacy_tests/list_blobs.py
mikeharder: Especially since this is Track 1, and we use async for the Track 2 version of the test, I'm fine not changing this.

mikeharder (approved):

sdk/storage/azure-storage-blob/tests/perfstress_tests/README.md
```diff
@@ -0,0 +1,72 @@
...
+```
+Using the `perfstress` command alone will list the available perf tests found. Note that the available tests discovered will vary depending on whether your environment is configured for the T1 or T2 SDK.
+
+### Common perf command line options
+These options are available for all perf tests:
+- `--duration=10` Number of seconds to run as many operations (the "run" function) as possible. Default is 10.
+- `--iterations=1` Number of test iterations to run. Default is 1.
+- `--parallel=1` Number of tests to run in parallel. Default is 1.
+- `--no-client-share` Whether each parallel test instance should share a single client, or use their own. Default is False (sharing).
+- `--warm-up=5` Number of seconds to spend warming up the connection before measuing begins. Default is 5.
```
tasherif-msft, line 37: minor nit :)
```suggestion
- `--warm-up=5` Number of seconds to spend warming up the connection before measuring begins. Default is 5.
```

sdk/storage/azure-storage-blob/tests/perfstress_tests/T1_legacy_tests/_test_base_legacy.py
```diff
@@ -0,0 +1,51 @@
...
+
+from azure.storage.blob import BlockBlobService
+
+class _LegacyServiceTest(PerfStressTest):
+    service_client = None
+    async_service_client = None
+
+    def __init__(self, arguments):
+        super().__init__(arguments)
+        connection_string = self.get_from_env("AZURE_STORAGE_CONNECTION_STRING")
```
tasherif-msft, line 19: another minor nit: could we mention that its fetching the connection string from an env variable in the README?

tasherif-msft (approved):
Besides the two small nits this looks great :)

sdk/storage/azure-storage-blob/tests/perfstress_tests/T1_legacy_tests/_test_base_legacy.py
annatisch: Great catch - forgot all about that!

PR #15834, opened 2020-12-16
[Perfstress][Storage] Added FileShare perf tests

annatisch:


sdk/storage/azure-storage-file-share/tests/perfstress_tests/T1_legacy_tests/upload.py
```diff
@@ -0,0 +1,29 @@
...
+from ._test_base import _LegacyShareTest
+
+
+class LegacyUploadTest(_LegacyShareTest):
+    def __init__(self, arguments):
+        super().__init__(arguments)
+        self.file_name = "sharefiletest-" + str(uuid.uuid4())
+
+    def run_sync(self):
+        data = RandomStream(self.args.size)
```
mikeharder, line 19: Can RandomStream be reused?

sdk/storage/azure-storage-file-share/tests/perfstress_tests/_test_base.py
```diff
@@ -0,0 +1,89 @@
...
+        self.async_sharefile_client = self.async_share_client.get_file_client(file_name)
+
+    async def global_setup(self):
+        await super().global_setup()
+        try:
+            self.sharefile_client.delete_file()
+        except ResourceNotFoundError:
+            pass
+
+    async def global_cleanup(self):
```
mikeharder, line 80: Should be unnecessary since super cleanup will delete whole share.

sdk/storage/azure-storage-file-share/tests/perfstress_tests/download.py
```diff
@@ -0,0 +1,35 @@
...
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for license information.
+# --------------------------------------------------------------------------------------------
+
+from azure_devtools.perfstress_tests import get_random_bytes
+
+from ._test_base import _ShareTest
+
+
+class DownloadTest(_ShareTest):
```
mikeharder, line 11: Same feedback as blob tests, download should re-use server file across all instances, while upload each instance should use its own server file.

sdk/storage/azure-storage-file-share/tests/perfstress_tests/download_to_file.py
```diff
@@ -0,0 +1,37 @@
...
+        self.sharefile_client = self.share_client.get_file_client(file_name)
+        self.async_sharefile_client = self.async_share_client.get_file_client(file_name)
+
+    async def global_setup(self):
+        await super().global_setup()
+        data = get_random_bytes(self.args.size)
+        await self.async_sharefile_client.upload_file(data)
+
+    def run_sync(self):
+        with tempfile.TemporaryFile() as fp:
```
mikeharder, line 26: Is this creating a unique temp file every call?  If so, can one temp file per instance be re-used?

mikeharder: Same question as blob, is this test useful if we are just consuming the data differently (write to file vs write to null)?

mikeharder (changes requested):

sdk/storage/azure-storage-file-share/tests/perfstress_tests/download.py
annatisch: They are all downloading the same file? Only one upload is done in `global_setup`? Or am I misunderstanding?

sdk/storage/azure-storage-file-share/tests/perfstress_tests/download_to_file.py
annatisch: Only useful in the sense that in T1 we had a dedicated API for downloading to a file, that did the file creation.
In that sense, creating a unique temp file per download call is probably a close approximation.
However if we feel there's no real need to test this comparative scenario I'm fine with removing it altogether.

sdk/storage/azure-storage-file-share/tests/perfstress_tests/T1_legacy_tests/upload.py
annatisch: Not currently - I would need to add some kind of "reset" feature to it - which would end up being as about as expensive as simply re-instantiating I think...

sdk/storage/azure-storage-file-share/tests/perfstress_tests/_test_base.py
```diff
@@ -0,0 +1,82 @@
...
+    def __init__(self, arguments):
+        super().__init__(arguments)
+        file_name = "sharefiletest-" + str(uuid.uuid4())
+        self.sharefile_client = self.share_client.get_file_client(file_name)
+        self.async_sharefile_client = self.async_share_client.get_file_client(file_name)
+
+    async def global_setup(self):
+        await super().global_setup()
+        try:
+            self.sharefile_client.delete_file()
```
mikeharder, line 76: Should this be `create_file()`?

sdk/storage/azure-storage-file-share/tests/perfstress_tests/download.py
mikeharder: You are correct, I had it backwards.

sdk/storage/azure-storage-file-share/tests/perfstress_tests/upload.py
```diff
@@ -0,0 +1,30 @@
...
+            self.upload_stream,
+            length=self.args.size,
+            max_concurrency=self.args.max_concurrency)
+
+    async def run_async(self):
+        self.upload_stream_async.reset()
+        await self.async_sharefile_client.upload_file(
+            self.upload_stream_async,
+            length=self.args.size,
+            max_concurrency=self.args.max_concurrency)
```
mikeharder, line 30: NIT: Missing newline

mikeharder (changes requested):

sdk/storage/azure-storage-file-share/tests/perfstress_tests/download_to_file.py
mikeharder: I'm fine keeping the test, we can always decide to not run it in automation if we think the scenario is no longer interesting.

sdk/storage/azure-storage-file-share/tests/perfstress_tests/_test_base.py
mikeharder: Or maybe `global_setup` should just be removed completely?  I believe the pattern we are using is `FileTest` creates a unique filename per instance, and clients mapped to that filename, but it doesn't actually create the file on the server.  This would be up to the actual test subclassing `FileTest`.

mikeharder (changes requested):

sdk/storage/azure-storage-file-share/tests/perfstress_tests/_test_base.py
annatisch: You're right - I'm not sure why this is here... as it's a new share, so there shouldn't be any cleanup necessary....

mikeharder (approved):

tasherif-msft:
Besides similar comments to Blobs, this looks great!

tasherif-msft (approved):

Issue #14217, opened 2020-10-02
[EngSys][ServiceBus/Tables] Remove workaround in preparers.py after fixing cosmos non-recording tests.


KieranBrantnerMagee:
For the hotfix [here](https://github.com/Azure/azure-sdk-for-python/pull/14195)

This fix should be deprecated and the underlying problem addressed.  It was initially constructed to handle tests that were marked live but did not have a recording/did not generate HTTP traffic, but if used in cached mode, this would cause problems with how the in_recording variable is set. (it sets a random name which is then preserved in the cache)

In conversation with @lmazuel there is a more appropriate angle to fix this:  Tests that do not generate recordings _simply should not be AzureTestCases._  Either the work to set up the resource should be in the test, not a preparer, (if you actually wanted a live resource) or it should be mocked and unit-testable.

One of these fixes should be applied to the cosmos tests in question, and the workaround clause should be yanked.

PR #15858, opened 2020-12-17
Investigate workaround

seankane-msft:
Verifying a hotfix documented in #14217 . Raising an error if the conditional is true to see how often tests might produce this error


seankane-msft:
I changed from raising an exception and instead am logging a warning.

iscai-msft (approved):

PR #15861, opened 2020-12-18
[Perfstress][Storage] Added Datalake perf tests

annatisch:


kasobol-msft:
@jaschrep-msft

jaschrep-msft:
I see no concerns.

sdk/storage/azure-storage-file-datalake/tests/perfstress_tests/append.py
```diff
@@ -0,0 +1,35 @@
...
+
+from azure_devtools.perfstress_tests import RandomStream, AsyncRandomStream
+
+from ._test_base import _FileSystemTest
+
+
+class AppendTest(_FileSystemTest):
+    def __init__(self, arguments):
+        super().__init__(arguments)
+        file_name = "filetest-" + str(uuid.uuid4())
```
mikeharder, line 16: Each instance should append to its own file

sdk/storage/azure-storage-file-datalake/tests/perfstress_tests/append.py
```diff
@@ -0,0 +1,35 @@
...
+        self.file_client = self.fs_client.get_file_client(file_name)
+        self.async_file_client = self.async_fs_client.get_file_client(file_name)
+
+    async def setup(self):
+        await self.async_file_client.create_file()
+
+    def run_sync(self):
+        data = RandomStream(self.args.size)
```
mikeharder, line 24: Can RandomStream be reused?

sdk/storage/azure-storage-file-datalake/tests/perfstress_tests/append.py
```diff
@@ -0,0 +1,35 @@
...
+        self.file_client.append_data(
```
mikeharder, line 25: Can appending to the same file over and over cause issues with service perf or correctness (e.g. is there a max file size)?

sdk/storage/azure-storage-file-datalake/tests/perfstress_tests/read.py
```diff
@@ -0,0 +1,36 @@
...
+        self.async_file_client = self.async_fs_client.get_file_client(file_name)
+
+    async def global_setup(self):
+        await super().global_setup()
+        data = get_random_bytes(self.args.size)
+        await self.async_file_client.create_file()
+        await self.async_file_client.upload_data(data, overwrite=True)
+
+    def run_sync(self):
+        download = WriteStream()
```
mikeharder, line 25: Can WriteStream be reused?

sdk/storage/azure-storage-file-datalake/tests/perfstress_tests/read.py
```diff
@@ -0,0 +1,36 @@
...
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for license information.
+# --------------------------------------------------------------------------------------------
+
+from azure_devtools.perfstress_tests import get_random_bytes, WriteStream
+
+from ._test_base import _FileSystemTest
+
+
+class DownloadTest(_FileSystemTest):
```
mikeharder, line 11: Extend `_FileTest`?

sdk/storage/azure-storage-file-datalake/tests/perfstress_tests/upload.py
```diff
@@ -0,0 +1,28 @@
...
+# Licensed under the MIT License. See License.txt in the project root for license information.
+# --------------------------------------------------------------------------------------------
+
+from ._test_base import _FileTest
+
+from azure_devtools.perfstress_tests import RandomStream
+from azure_devtools.perfstress_tests import AsyncRandomStream
+
+
+class UploadTest(_FileTest):
```
mikeharder, line 12: Same feedback as blob, download can use single service file, upload each instance must use own file.

sdk/storage/azure-storage-file-datalake/tests/perfstress_tests/upload_from_file.py
```diff
@@ -0,0 +1,40 @@
...
+        with tempfile.NamedTemporaryFile(delete=False) as temp_file:
+            UploadFromFileTest.temp_file = temp_file.name
+            temp_file.write(data)
+
+    async def global_cleanup(self):
+        os.remove(UploadFromFileTest.temp_file)
+        await super().global_cleanup()
+
+    def run_sync(self):
+        with open(UploadFromFileTest.temp_file, 'rb') as fp:
```
mikeharder, line 29: Same feedback as blob, is uploading from file worth testing?

mikeharder (changes requested):

sdk/storage/azure-storage-file-datalake/tests/perfstress_tests/append.py
annatisch: It is - this constructor is scoped to the instance, not global. And the `setup` method constructs the file specific to this instance.

sdk/storage/azure-storage-file-datalake/tests/perfstress_tests/append.py
annatisch: I will check in with the Storage team to see if they can shed light on that

sdk/storage/azure-storage-file-datalake/tests/perfstress_tests/read.py
annatisch: No really, as it maintains positional state - which in turn is used by the SDK for multi-threaded downloads

sdk/storage/azure-storage-file-datalake/tests/perfstress_tests/upload.py
annatisch: I think that's what it's already doing?

sdk/storage/azure-storage-file-datalake/tests/perfstress_tests/upload_from_file.py
annatisch: I'm happy to remove it if we don't feel need to test parity against the T1 API for this.
We can also remove if from the table of results in onenote.

sdk/storage/azure-storage-file-datalake/tests/perfstress_tests/read.py
annatisch: `_FileTest` base currently creates a unique destination filename (and so is used for the upload tests).
So for the downloads tests I have only used the `_FileSystemTest` base and have each instance of the `DownloadTest` simply hit the same filename endpoint.

annatisch: (But I did have the `global_setup/cleanup` for `_FileTest` messed up - so have fixed that)

annatisch: There's really not much benefit to the `_FileTest` base - I could refactor it out

mikeharder: Keep or remove `_FileTest` is up to you, it might be used more in the future, and .NET has a similar class, so my vote is to keep it.

sdk/storage/azure-storage-file-datalake/tests/perfstress_tests/upload_from_file.py
mikeharder: Let's keep it for now

mikeharder (approved):

sdk/storage/azure-storage-file-datalake/tests/perfstress_tests/append.py
tasherif-msft: There should be no problems. When you keep re-appending, you would be rewriting the appended data (i.e the last append wins - assuming same length and offset). As for the maximum size, its 4.75 TiB (100 MiB X 50,000 blocks) for version 2016-05-31 and later, and 195 GiB (4 MiB X 50,000 blocks) for all older versions (since a datalake file is a block blob).

tasherif-msft (approved):

PR #15862, opened 2020-12-18
[Storage][LiveTest]replace the fixed blob names

xiafu-msft:


kasobol-msft (approved):

tasherif-msft:
LGTM I think some more tests need to be rerecorded for CI to be green.

tasherif-msft (approved):

sdk/storage/azure-storage-blob/tests/test_block_blob.py
```diff
@@ -48,7 +48,10 @@ def _setup(self, storage_account, key, container_name='utcontainer'):
         self.container_name = self.get_resource_name(container_name)
 
         if self.is_live:
-            self.bsc.create_container(self.container_name)
+            try:
+                self.bsc.create_container(self.container_name)
+            except:
```
tasherif-msft, line 53: Good point. I think we should catch a more specific exception just to be safe, so something like `except ResourceExistsError:`

PR #15865, opened 2020-12-18
Run smoke tests against latest published versions in release pipeline

benbp:
This PR updates the package release pipelines to run smoke tests against the latest published versions (GA and/or beta) after the release stage in the release pipeline. This only applies for packages that are included as dependencies in the smoke tests (keyvault, storage, eventhub).

Azure/azure-sdk#1999


scbedd (approved):
I don't see any issues with this. The template looks mostly the same except for the two conditional injections.

I'd like to see a successful run of the template w/ these new changes, just to confirm we're not missing any requirements gotchas in the release stages.

benbp:
@scbedd here's a run of the normal smoke tests pipeline: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=660094&view=results

Here's a run of just the smoke test portion of the release pipeline (with actual build/release steps commented out): https://dev.azure.com/azure-sdk/internal/_build/results?buildId=660110&view=results

If you can think of a way to exercise the whole release pipeline in a non-consequential way, I'd love to do that too.

scbedd:
@benbp Dozie has some _excellent_ automation in place to execute a release of the template package (including all stages) with a version tagged by buildNumber. That's one way we can actually exercise it.

scbedd:
You're good to just merge this IMO though. That test you did with the actual stuff commented out works for me as well.

common/smoketest/requirements-release.txt
```diff
@@ -0,0 +1,8 @@
+azure-core>=0.0.0b1
```
weshaggard, line 1: Are we certain this will automatically pick the latest pre-release version? I'm not an expert but if I'm reading https://www.python.org/dev/peps/pep-0440/#handling-of-pre-releases correctly it will exclude pre-releases for these version identifiers unless they are already installed.

benbp: It does, as I think appending `b1` causes this to happen. For example, as of this writing `azure-storage-blob` has versions `12.6.0` and `12.7.0b1` published upstream. By setting the version to `>= 0.0.0b1`, pip installs `12.7.0b1`, with no cache:

```
⇉ ⇉ ⇉ python -m venv ./prereleasetest
⇉ ⇉ ⇉ cd prereleasetest && . ./bin/activate
⇉ ⇉ ⇉ cd ~/sdk/azure-sdk-for-python/common/smoketest
⇉ ⇉ ⇉ pip3 install -r requirements-release.txt --no-deps --upgrade  --no-cache-dir | grep storage
Collecting azure-storage-blob>=0.0.0b1
  Downloading azure_storage_blob-12.7.0b1-py2.py3-none-any.whl (332 kB)
...
⇉ ⇉ ⇉ pip3 freeze | grep storage
azure-storage-blob==12.7.0b1
```

benbp: @scbedd might have more context on the underlying behavior, as he showed me this tip originally.

weshaggard: If it works then great.

rakshith91 (approved):

benbp:
@weshaggard I updated this PR to pull in versions from the build artifacts, similar to the net and js PRs.

weshaggard (approved):
Looks good.

PR #15872, opened 2020-12-18
Use powershell preparer

seankane-msft:
Switches over to the arm template. next step is to add the samples as part of the pipeline


seankane-msft:
Also fixes #12614

lmazuel (approved):

Issue #15796, opened 2020-12-15
Broken link in azure-keyvault v1 readme


ramya-rao-a:
The [readme for azure-keyvault v1](https://github.com/Azure/azure-sdk-for-python/tree/azure-keyvault_1.1.0/azure-keyvault) is linked from the [Key Vault Overview page for Python packages](https://docs.microsoft.com/en-us/python/api/overview/azure/key-vault-index?view=azure-python).

This readme has a broken link https://docs.microsoft.com/python/api/overview/azure/key-vault
The problem is in the hypen between the words "key" and "vault", the right url is https://docs.microsoft.com/python/api/overview/azure/keyvault

Our options are
- Create a redirect link such that the wrong one leads to the right one
- Or, create a branch from the tag `azure-keyvault_1.1.0`, fix the link, delete the tag, add the tag to this branch.

Found in #15110
Nice catch @mccoyp!

mccoyp:
As mentioned in the pull request meant to resolve this, it turns out that the `azure-keyvault_1.1.0` tag containing the README is immutable. Unless we do a release of a version 1.1.1, our remaining option for fixing the link is to create a redirect link.

mccoyp:
Resolved via MicrosoftDocs/azure-docs-sdk-python#1124.

PR #15873, opened 2020-12-18
[Key Vault] Link fix for azure-keyvault 1.1.0 README

mccoyp:
Part of fix for #15796.

After getting the fixed link checked in, we'll need to replace the current `azure-keyvault_1.1.0` tag with one from this commit.

### DIFF
--- a/azure-keyvault/README.rst
+++ b/azure-keyvault/README.rst
@@ -31,7 +31,7 @@ Usage
 =====
 
 For code examples, see `Key Vault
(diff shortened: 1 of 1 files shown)


mccoyp:
The `azure-keyvault_1.1.0` tag turns out to be immutable, so fixing the link with a code change would require a version 1.1.1. If there's additional fixing up or deprecation note adding that needs to be done, we could do a release -- otherwise, it's probably easier to add a redirect from the broken link to the desired endpoint.

PR #15878, opened 2020-12-18
use getattr to lazy load aiohttp and trio

xiangyan99:


sdk/core/azure-core/azure/core/pipeline/transport/__init__.py
```diff
@@ -47,22 +47,67 @@
...
-        ])
-    except ImportError:
-        pass  # Trio not installed
+    if sys.version_info >= (3, 7):
+        def __getattr__(name):
+            if name == 'AioHttpTransport':
+                try:
+                    from ._aiohttp import AioHttpTransport
+                    if 'AioHttpTransport' not in __all__:
+                        __all__.extend([
```
lmazuel, line 56: `__all__` is a pre-declaration of what exists, mostly to help IDE and doc to know what they would show. This means that I think this should be available even if the actualy import is not done. But putting this inside the `__getattr__`, it means `dir` is now not accurate:
image

I _believe_ it should be possible to have `__all__` correct, with a static list of strings outside of the `__getattr__` but inside your `if sys.version_info >= (3, 7)`, in order to get the regular tooling able to list the possible importable types.

lmazuel: @xiangyan99 I see your latest commit, doesn't fix `dir`, so read the spec this time and not guessing and there is a `__dir__`:
https://www.python.org/dev/peps/pep-0562/#specification

We're close :)

lmazuel: Something like:
```python
if python > 3.7:
    ...
    def __dir__():
        return __all__
```

xiangyan99: image


Now it works. :)

lmazuel: You still get the perf improvement right? None of the previous commits breaks anything?

lmazuel: @brettcannon If you're curious, we're playing with PEP562 here ;).

xiangyan99: Yes. We still get the perf improvement with the changes.

brettcannon: @lmazuel I figured since I wrote a proof-of-concept of this 2.5 years ago with you in mind 😄  https://pypi.org/project/modutil/ (which also touches on `__all__` and `dir()` as you are discovering 😁 ).

And to be clear about `__all__`, it exists to control `from ... import *` specifically, but has come to mean for some what the public API is.

xiangyan99:
#9399

sdk/core/azure-core/azure/core/pipeline/transport/__init__.py
johanste: Yes, we (the azure client libraries) are part of the "assume `__all__` means what the public API surface area is crowd". Primarily this is because that is what sphinx/autodoc, by default, uses it as well. So it's convenient.

@brettcannon, I'm not aware of any downsides of making this assumption - are there gotchas that we should look out for? I personally don't find the `import  *` as a mechanism to import the "recommended"/most commonly used subset of the full public API surface area for a package/module to be a common (or good) practice.

brettcannon: `import *` is most definitely not meant to convey "recommended/commonly used"; it is entirely to make it easier to play with things in the REPL and you can't make that judgement call for people as to what is common (and `import *` one of the things I would remove from the language if I could).

johanste: Yeah, the "most commonly used" argument is probably the only argument that I had heard from people recommending the use of `import *`. But I don't think it is a common argument. The use in REPL is a much better argument, but I would agree that, given a time machine, I would still personally support a quick trip back in time and remove the support for it altogether.

lmazuel (approved):

Issue #12743, opened 2020-07-27
[Tables] Make full use of generated code


seankane-msft:
The table client and table service client have some parts (creating pipeline and configuration) that are done explicitly when it's possible this was taken care of by the generated code.

seankane-msft:
Remove `create_configuration` in `_base_client.py` and have this taken care of by the generated code.

PR #15908, opened 2020-12-22
removing create_configuration and using generated code instead

seankane-msft:
Addresses #12743

### DIFF
--- a/sdk/tables/azure-data-tables/azure/data/tables/_base_client.py
+++ b/sdk/tables/azure-data-tables/azure/data/tables/_base_client.py
@@ -17,53 +17,27 @@ from typing import (  # pylint: disable=unused-import
 )
 import logging
 from uuid import uuid4
-
-
-
+import six
 try:
     from urllib.parse import parse_qs, quote
 except ImportError:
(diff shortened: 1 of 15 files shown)


sdk/tables/azure-data-tables/azure/data/tables/_base_client.py
```diff
@@ -231,7 +232,15 @@ def _create_pipeline(self, credential, **kwargs):
         elif credential is not None:
             raise TypeError("Unsupported credential: {}".format(credential))
 
-        config = kwargs.get("_configuration") or create_configuration(**kwargs)
+        config = kwargs.get("_configuration") or AzureTableConfiguration(
```
iscai-msft, line 235: I didn't see this whole code to create a pipeline, you don't really need to do that anymore

I would actually suggest taking a look at kv: https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_shared/client_base.py#L46. We used to have this same code, where we would list all the policies, etc, but we changed it to this. Should get rid of. all of this creating a pipeline / transport / config before calling into thegeneratedclient

PR #15918, opened 2020-12-24
Fix put_encrypted_blob function not encrypting blob

griseau:
The function put_encrypted_blob was setting key_resolver_function instead of key_encryption_key. Due to this, the blob uploaded to the container was not encrypted.


xiafu-msft (approved):

xiafu-msft:
Thanks so much for the pr! It looks great

PR #15946, opened 2020-12-29
Add AzureSasCredential

kasobol-msft:
**In this PR:**
- Add `AzureSasCredential` per https://github.com/Azure/azure-sdk/issues/1954
- `AzureSasCredential` is the name that has been settled on the end of discussion.
- Add `AzureSasCredentialPolicy` that appends SAS to query

**Remarks:**
- Some service (like storage in the Portal) present SAS with leading "?". This has to be stripped before appending
- The validation if serviceUri already contain sas (mentioned [here](https://github.com/Azure/azure-sdk/issues/1954#issuecomment-718190611)) will be responsibility of service clients:
    - the format varies between services (i.e. Event Grid SAS and Storage SAS are vastly different)
    - it would be good to fail fast (at client creation) rather than late (at request send).

**References**
- [.NET PR](https://github.com/Azure/azure-sdk-for-net/pull/17636)


annatisch:
Thanks @kasobol-msft - looks good.
Could you please update release notes? I assume this will initially be going out as a preview version?

kasobol-msft:
> Thanks @kasobol-msft - looks good.
> Could you please update release notes? I assume this will initially be going out as a preview version?

As discussed offline. We intend to ship this in GA. I've added release notes entry.

xiangyan99:
What's the difference betweeen AzureSasCredential and AzureKeyCredential? Can we just create an alias of AzureKeyCredential?

sdk/core/azure-core/CHANGELOG.md
```diff
@@ -3,6 +3,10 @@
 
 ## 1.9.1 (Unreleased)
```
xiangyan99, line 4: Given this is a new feature, it should be 1.10.0

kasobol-msft: fixed.

kasobol-msft:
> What's the difference betweeen AzureSasCredential and AzureKeyCredential? Can we just create an alias of AzureKeyCredential?

There isn't much today. However per https://github.com/Azure/azure-sdk/issues/1954 the AzureKeyCredential will get extra property (name). The outcome from board discussion was to keep these types separate.

annatisch (approved):

PR #15987, opened 2021-01-05
Skip automatic API review for management packages

praveenkuttappan:
Skip automatic review creation for management packages since these are not currently reviewed by architects

### DIFF
--- a/eng/scripts/Language-Settings.ps1
+++ b/eng/scripts/Language-Settings.ps1
@@ -185,6 +185,13 @@ function Find-python-Artifacts-For-Apireview($artifactDir, $artifactName)
 {
   # Find wheel file in given artifact directory
   # Make sure to pick only package with given artifact name
+  # Skip auto API review creation for management packages
+  if ($artifactName -match "mgmt")
+  {
+    Write-Host "Skipping automatic API review for management artifact $($artifactName)"
+    return $null
+  }
+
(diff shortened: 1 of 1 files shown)


scbedd (approved):

eng/scripts/Language-Settings.ps1
```diff
@@ -185,6 +185,13 @@ function Find-python-Artifacts-For-Apireview($artifactDir, $artifactName)
 {
   # Find wheel file in given artifact directory
   # Make sure to pick only package with given artifact name
+  # Skip auto API review creation for management packages
+  if ($artifactName -match "mgmt")
```
weshaggard, line 189: Can we please get confirmation from some one on the mgmt team so they are aware this is happening and can confirm this is what they want?

praveenkuttappan: I have discussed this with Anna and Johan about this and making this change as per their request. Will definitely send a note to mgmt team also. I have a call with management team tomorrow. But I would like to disable auto reviews for the time being for management plane until they agree to approve reviews.

weshaggard: I just want to be sure we are including them in the conversation and not just making decisions on their behalf. If it is easier to disable them for now that is fine just be sure to start the conversation with them.

PR #16010, opened 2021-01-07
Setting up template to use 1ES hosted pool.

mitchdenny:
This PR modifies the Python pipelines to use the 1ES hosted pools for WIndows/Linux build/test legs. It has a few aspects:

1. Change to the stage template to define a WindowsPool and LinuxPool.
2. Change to the job template to define a WindowsPool and LinuxPool
3. Maps WindowsLinux and LinuxPool to variables in the test job so that they can be used in the matrix.

### DIFF
--- a/eng/pipelines/templates/steps/build-test.yml
+++ b/eng/pipelines/templates/steps/build-test.yml
@@ -24,8 +24,6 @@ steps:
     parameters:
       versionSpec: '${{ parameters.PythonVersion }}'
 
-  - template: /eng/common/pipelines/templates/steps/verify-agent-os.yml
-
   - script: |
       python -m pip install pip==20.1
       pip install -r eng/ci_tools.txt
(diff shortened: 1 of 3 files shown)


eng/pipelines/templates/steps/build-test.yml
```diff
@@ -24,8 +24,6 @@ steps:
     parameters:
       versionSpec: '${{ parameters.PythonVersion }}'
 
-  - template: /eng/common/pipelines/templates/steps/verify-agent-os.yml
```
weshaggard, line 27: I'm OK with us disabling this temporarily to onboard the pool but please file an issue to track enabling some way to validate the machines are as we expect. We should either have something that runs in the pipeline or have some external pipeline that can run on the different pools to assert they are giving use the correct OS images.

weshaggard (approved):

eng/pipelines/templates/steps/build-test.yml
mitchdenny: https://github.com/Azure/azure-sdk-tools/issues/1313

PR #16015, opened 2021-01-07
Additional Blob/Datalake regression storage pipeline fix

tasherif-msft:
There was another regression failure related to the migration to the new generator. This PR should resolve it.

### DIFF
--- a/eng/pipelines/templates/jobs/archetype-sdk-client.yml
+++ b/eng/pipelines/templates/jobs/archetype-sdk-client.yml
@@ -178,7 +178,6 @@ jobs:
               BuildTargetingString: ${{ parameters.BuildTargetingString }}
 
   - job: 'RunRegression'
-    condition: and(succeededOrFailed(), or(eq(variables['Run.Regression'], 'true'), and(eq(variables['Build.Reason'], 'Schedule'), eq(variables['System.TeamProject'],'internal'))))
     displayName: 'Run Regression'
     timeoutInMinutes: 120
     variables:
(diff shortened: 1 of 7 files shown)


eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -178,7 +178,6 @@ jobs:
               BuildTargetingString: ${{ parameters.BuildTargetingString }}
 
   - job: 'RunRegression'
-    condition: and(succeededOrFailed(), or(eq(variables['Run.Regression'], 'true'), and(eq(variables['Build.Reason'], 'Schedule'), eq(variables['System.TeamProject'],'internal'))))
```
weshaggard, line 181: @scbedd @praveenkuttappan I assume we don't want to remove this condition.

tasherif-msft: I'm removing it temporarily - for testing purposes (forcing regression to run). I will undo this before merging.

eng/tox/install_depend_packages.py
```diff
@@ -28,7 +28,8 @@
 logging.getLogger().setLevel(logging.INFO)
 
 MINIMUM_VERSION_SUPPORTED_OVERRIDE = {
-    'azure-common': '1.1.10'
+    'azure-common': '1.1.10',
+    'azure-storage-file-datalake': ' 12.2.0'
```
scbedd, line 32:
```suggestion
    'azure-storage-file-datalake': '12.2.0'
```

scbedd (commented):
I'll approve after condition gets add back!

eng/tox/install_depend_packages.py
tasherif-msft: woops

scbedd (approved):

PR #16026, opened 2021-01-07
Added test pass info for python

LarryOsterman:
Added Test Pass Information for MAA SDK.


LarryOsterman:
Ping. It would be nice to get this committed.

LarryOsterman:
Ping. we're up to 3 months that this has been open :(.

weshaggard (approved):

sdk/attestation/tests.yml
```diff
@@ -0,0 +1,36 @@
...
+  none
+
+jobs:
+ - template: ../../eng/pipelines/templates/jobs/archetype-sdk-tests.yml
+   parameters:
+     ServiceDirectory: attestation
+     Location: westus
+     DeployArmTemplate: true
+     EnvVars:
+       ATTESTATION_CLIENT_ID: "$(ATTESTATION_CLIENT_ID)"
```
weshaggard, line 11: Where are these variables coming from? I don't see them in the arm template nor do I see them in the variable group hooked up to the pipeline. I suspect these are dead code and if so it would be good to remove them.

weshaggard:
Sorry @LarryOsterman this one was lost in the noise. I'm going to go ahead and merge this so you can get the test pipeline running but it would be good if you can follow-up on whether or not you need those env variables passed.

sdk/attestation/tests.yml
LarryOsterman: The variables come from the test-resources.json file and the test-resources-pre.ps1 script. MAA requires some fairly specific configuration before it can work :(.

PR #16032, opened 2021-01-08
Fix test

xiafu-msft:


sdk/storage/azure-storage-blob/tests/test_block_blob.py
```diff
@@ -67,7 +67,10 @@ def _get_blob_reference(self):
     def _create_blob(self, tags=None, data=b'', **kwargs):
         blob_name = self._get_blob_reference()
         blob = self.bsc.get_blob_client(self.container_name, blob_name)
-        blob.upload_blob(data, tags=tags, **kwargs)
+        try:
+            blob.upload_blob(data, tags=tags, **kwargs)
+        except:
+            blob.upload_blob(data, tags=tags, overwrite=True, **kwargs)
```
kasobol-msft, line 73: should we just always overwrite? I'm assuming that upload_blob is covered by some dedicated tests and this function is just a helper for other scenarios.

xiafu-msft: yeah we can do that, this is just a quick fix to avoid re-recording every test in the file.

xiafu-msft: it seems we don't need to re-record!

tasherif-msft (approved):

kasobol-msft (approved):

PR #16033, opened 2021-01-08
[eventgrid] add tests for samples

rakshith91:
Samples are triggered in the live run
Example: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=679428&view=results

Fixes https://github.com/Azure/azure-sdk-for-python/issues/14322

### DIFF
--- a/scripts/devops_tasks/test_run_samples.py
+++ b/scripts/devops_tasks/test_run_samples.py
@@ -21,7 +21,11 @@ logging.getLogger().setLevel(logging.INFO)
 root_dir = os.path.abspath(os.path.join(os.path.abspath(__file__), "..", "..", ".."))
 
 IGNORED_SAMPLES = {
-    "azure-eventgrid": ["__init__.py"]
+    "azure-eventgrid": [
+        "__init__.py",
+        "consume_cloud_events_from_eventhub.py",
+        "consume_cloud_events_from_service_bus_queue.py",
+        "consume_cloud_events_from_storage_queue.py"]
 }
 
 
--- a/scripts/devops_tasks/test_run_samples.py
+++ b/scripts/devops_tasks/test_run_samples.py
@@ -25,7 +25,8 @@ IGNORED_SAMPLES = {
         "__init__.py",
         "consume_cloud_events_from_eventhub.py",
         "consume_cloud_events_from_service_bus_queue.py",
-        "consume_cloud_events_from_storage_queue.py"]
+        "consume_cloud_events_from_storage_queue.py",
+        "cs2_publish_custom_events_to_a_domain_topic"]
 }
 
 
--- a/scripts/devops_tasks/test_run_samples.py
+++ b/scripts/devops_tasks/test_run_samples.py
@@ -26,7 +26,7 @@ IGNORED_SAMPLES = {
         "consume_cloud_events_from_eventhub.py",
         "consume_cloud_events_from_service_bus_queue.py",
         "consume_cloud_events_from_storage_queue.py",
-        "cs2_publish_custom_events_to_a_domain_topic"]
+        "cs2_publish_custom_events_to_a_domain_topic.py"]
 }
 
 
--- a/sdk/eventgrid/azure-eventgrid/samples/champion_scenarios/cs3_consume_system_events.py
+++ b/sdk/eventgrid/azure-eventgrid/samples/champion_scenarios/cs3_consume_system_events.py
@@ -15,7 +15,7 @@ import json
 from azure.eventgrid import EventGridConsumer
 
 consumer = EventGridConsumer()
-path = os.path.abspath(os.path.join(os.path.abspath(__file__), "..", "./cs3_system_events.json"))
+path = os.path.abspath(os.path.join(os.path.abspath(__file__), "..", "./cs3_system_event.json"))
 
 with open(path, 'r') as f:
     eg_event_received_message = json.loads(f.read())
--- a/sdk/eventgrid/azure-eventgrid/samples/champion_scenarios/cs3_consume_system_events.py
+++ b/sdk/eventgrid/azure-eventgrid/samples/champion_scenarios/cs3_consume_system_events.py
@@ -15,8 +15,9 @@ import json
 from azure.eventgrid import EventGridConsumer
 
 consumer = EventGridConsumer()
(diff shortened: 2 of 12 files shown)


sdk/eventgrid/azure-eventgrid/samples/champion_scenarios/cs3_consume_system_events.py
```diff
@@ -15,8 +15,9 @@
 from azure.eventgrid import EventGridConsumer
 
 consumer = EventGridConsumer()
+path = os.path.abspath(os.path.join(os.path.abspath(__file__), "..", "./cs3_consume_system_events.json"))
```
rakshith91, line 18:
```suggestion
path = os.path.abspath(os.path.join(os.path.abspath(__file__), "..", "./cs3_system_events.json"))
```

rakshith91:
```suggestion
path = os.path.abspath(os.path.join(os.path.abspath(__file__), "..", "./cs3_consume_system_event.json"))
```

sdk/eventgrid/azure-eventgrid/samples/champion_scenarios/cs3_consume_system_events.py
```diff
@@ -15,8 +15,9 @@
 from azure.eventgrid import EventGridConsumer
 
 consumer = EventGridConsumer()
+path = os.path.abspath(os.path.join(os.path.abspath(__file__), "..", "./cs3_system_events.json"))
```
rakshith91, line 18:
```suggestion
path = os.path.abspath(os.path.join(os.path.abspath(__file__), "..", "./cs3_system_event.json"))
```

scripts/devops_tasks/test_run_samples.py
```diff
@@ -21,7 +21,12 @@
 root_dir = os.path.abspath(os.path.join(os.path.abspath(__file__), "..", "..", ".."))
 
 IGNORED_SAMPLES = {
-    "azure-eventgrid": ["__init__.py"]
+    "azure-eventgrid": [
+        "__init__.py",
+        "consume_cloud_events_from_eventhub.py",
+        "consume_cloud_events_from_service_bus_queue.py",
+        "consume_cloud_events_from_storage_queue.py",
+        "cs2_publish_custom_events_to_a_domain_topic"]
```
rakshith91, line 29: Will be fixed in this https://github.com/Azure/azure-sdk-for-python/issues/16042

scripts/devops_tasks/test_run_samples.py
```diff
@@ -21,7 +21,12 @@
 root_dir = os.path.abspath(os.path.join(os.path.abspath(__file__), "..", "..", ".."))
 
 IGNORED_SAMPLES = {
-    "azure-eventgrid": ["__init__.py"]
+    "azure-eventgrid": [
+        "__init__.py",
+        "consume_cloud_events_from_eventhub.py",
+        "consume_cloud_events_from_service_bus_queue.py",
+        "consume_cloud_events_from_storage_queue.py",
```
rakshith91, line 28: it's not practical to test these these 4 samples - they have a dependency on storage, servicebus, EH, azure functions.
it's creating a lot of maintainance in tests.yml and in adding more dependcies to dev_requirements. Also, it's not simple to deploy azure functions on CI sample run and moreover, we do have consume samples which mimic the data we receive.

scripts/devops_tasks/test_run_samples.py
rakshith91, line 29:
```suggestion
        "cs2_publish_custom_events_to_a_domain_topic.py"]
```

scripts/devops_tasks/test_run_samples.py
yunhaoling: you're farsighted, thanks for calling it out!

but I'm afraid that in the long term we want to have tests against those integration samples. Because it helps us verify if there're any changes in EH/SB/Storage SDK and we'll need to upgrade these EG samples accordingly.

I'm not sure how other SDKs is doing, but I think it would be nice to have tests for integration samples.
I didn't intent to block the PR, we could open separate issue and PRs when we have time.

thoughts? @rakshith91 @lmazuel

scripts/devops_tasks/test_run_samples.py
```diff
@@ -21,7 +21,12 @@
 root_dir = os.path.abspath(os.path.join(os.path.abspath(__file__), "..", "..", ".."))
 
 IGNORED_SAMPLES = {
-    "azure-eventgrid": ["__init__.py"]
+    "azure-eventgrid": [
+        "__init__.py",
```
yunhaoling, line 25: dumb question for my own learning: why here is a `__init__.py`, I don't see there is a `__init__.py` inside the sample folder?

rakshith91: the __init__.py is from the azurefunctions appi was talking about in my previous comment
https://github.com/Azure/azure-sdk-for-python/tree/master/sdk/eventgrid/azure-eventgrid/samples/consume_samples/e2e_samples/functionsapp

scripts/devops_tasks/test_run_samples.py
rakshith91: that's a good point - i agree that integration tests can be useful in the long run when we eventually have some breaking changes - what i can do is add these tests once we move the test framework to use the arm template in the march milestone - that way, we have less maintenance on tests.yml

yunhaoling: sounds good to me!

scripts/devops_tasks/test_run_samples.py
yunhaoling: oh I see, then this is for EG only, make sense to me

scripts/devops_tasks/test_run_samples.py
rakshith91: https://github.com/Azure/azure-sdk-for-python/issues/16057

kristapratico (approved):
LGTM in terms of running samples

yunhaoling (approved):
LGTM, do we need to kick off a livetest run against this PR before merging?

rakshith91:
> do we need to kick off a livetest run against this PR before merging?

already did and shared the link on the pr description

PR #16049, opened 2021-01-08
Update generated

seankane-msft:
Use more generated code for config and pipeline creation

### DIFF
--- a/sdk/tables/azure-data-tables/azure/data/tables/_base_client.py
+++ b/sdk/tables/azure-data-tables/azure/data/tables/_base_client.py
@@ -125,9 +125,7 @@ class StorageAccountHostsMixin(object):  # pylint: disable=too-many-instance-att
         self.key_encryption_key = kwargs.get("key_encryption_key")
         self.key_resolver_function = kwargs.get("key_resolver_function")
 
-        self._configure_credential(
-            self.credential, storage_sdk=service, **kwargs
-        )
+        self._configure_credential(self.credential)
 
         self._config = kwargs.get("_configuration") or AzureTableConfiguration(
             url=self.url,
@@ -264,7 +262,7 @@ class StorageAccountHostsMixin(object):  # pylint: disable=too-many-instance-att
             credential = None
         return query_str.rstrip("?&"), credential
 
-    def _configure_credential(self, credential, **kwargs):
+    def _configure_credential(self, credential):
         # type: (Any, **Any) -> Tuple[Configuration, Pipeline]
         self._credential_policy = None
         if hasattr(credential, "get_token"):
@@ -307,7 +305,7 @@ class StorageAccountHostsMixin(object):  # pylint: disable=too-many-instance-att
             boundary="batch_{}".format(uuid4()),
         )
 
-        pipeline_response = self._client._client._pipeline.run(request, **kwargs)
+        pipeline_response = self._client._client._pipeline.run(request, **kwargs)  # pylint:disable=protected-access
         response = pipeline_response.http_response
 
         if response.status_code == 403:
--- a/sdk/tables/azure-data-tables/azure/data/tables/_base_client.py
+++ b/sdk/tables/azure-data-tables/azure/data/tables/_base_client.py
@@ -125,33 +125,21 @@ class StorageAccountHostsMixin(object):  # pylint: disable=too-many-instance-att
         self.key_resolver_function = kwargs.get("key_resolver_function")
 
         self._configure_credential(self.credential)
-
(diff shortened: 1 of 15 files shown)


seankane-msft:
Addresses #12743

sdk/tables/azure-data-tables/azure/data/tables/_base_client.py
```diff
@@ -124,10 +124,39 @@ def __init__(
...
         self.key_encryption_key = kwargs.get("key_encryption_key")
         self.key_resolver_function = kwargs.get("key_resolver_function")
-        self._config, self._pipeline = self._create_pipeline(
-            self.credential, storage_sdk=service, **kwargs
+
+        self._configure_credential(self.credential)
+
+        self._config = kwargs.get("_configuration") or AzureTableConfiguration(
+            url=self.url,
+            headers_policy=StorageHeadersPolicy(**kwargs),
```
iscai-msft, line 132: i see you're passing in your own list. of policies because you're adding policies that configuration doesn't have properties for. In that case, I wouldn't bother wit all of this `Configurationi` creation, since you're just going to override all of. this configuration policy information with the policies you have defined under `self._policies`. I would instead just use. the generate code's configuration

sdk/tables/azure-data-tables/azure/data/tables/_base_client.py
```diff
@@ -124,10 +124,39 @@ def __init__(
...
+            user_agent_policy=UserAgentPolicy(sdk_moniker=SDK_MONIKER, **kwargs),
+            retry_policy=kwargs.get("retry_policy") or TablesRetryPolicy(**kwargs),
+            logging_policy=StorageLoggingPolicy(**kwargs),
+            proxy_policy=ProxyPolicy(**kwargs)
         )
 
+        kwargs.setdefault("connection_timeout", CONNECTION_TIMEOUT)
+        kwargs.setdefault("read_timeout", READ_TIMEOUT)
+        self._config.transport = kwargs.get("transport") or RequestsTransport(**kwargs)
```
iscai-msft, line 141: transport is a property on the pipeline, not on the config. [Here's][https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/core/azure-core/azure/core/_pipeline_client.py#L101] the code for the pipeline client, and it appears to do this code for you, so you shouldn't need this line, just make sure you pass the kwargs to the generated client and this shoudl be handled for you

sdk/tables/azure-data-tables/azure/data/tables/_base_client.py
```diff
@@ -461,37 +464,37 @@ def parse_connection_str(conn_str, credential, service, keyword_args):
...
-    config.max_page_size = kwargs.get("max_page_size", 4 * 1024 * 1024)
-
-    # Blob downloads
-    config.max_single_get_size = kwargs.get("max_single_get_size", 32 * 1024 * 1024)
-    config.max_chunk_get_size = kwargs.get("max_chunk_get_size", 4 * 1024 * 1024)
-
-    # File uploads
-    config.max_range_size = kwargs.get("max_range_size", 4 * 1024 * 1024)
-    return config
+# def create_configuration(**kwargs):
```
iscai-msft, line 467: should this code be deleted?

sdk/tables/azure-data-tables/azure/data/tables/_table_service_client.py
```diff
@@ -69,7 +69,11 @@ def __init__(
         super(TableServiceClient, self).__init__(
             account_url, service="table", credential=credential, **kwargs
         )
-        self._client = AzureTable(self.url, pipeline=self._pipeline)
+        self._client = AzureTable(
+            self.url,
+            transport=self._config.transport,
```
iscai-msft, line 74: i don't thinkt his line is right, just pass in **kwargs and that should be enouh

sdk/tables/azure-data-tables/azure/data/tables/aio/_table_client_async.py
```diff
@@ -77,7 +78,14 @@ def __init__(
             loop=loop,
             **kwargs
         )
-        self._client = AzureTable(self.url, pipeline=self._pipeline, loop=loop)
+        kwargs['connection_timeout'] = kwargs.get('connection_timeout') or CONNECTION_TIMEOUT
+        self._configure_policies(**kwargs)
+        self._client = AzureTable(
+            self.url,
+            policies=kwargs.pop('policies', None) or self._policies,
```
iscai-msft, line 85: tiny nit: you can just do `kwargs.pop('policies, self._policies)`

iscai-msft (approved):

PR #16066, opened 2021-01-10
[Perfstress][ServiceBus] Added ServiceBus perf tests

annatisch:


sdk/servicebus/azure-servicebus/tests/perf_tests/T1_legacy_tests/_test_base.py
```diff
@@ -0,0 +1,139 @@
...
+    def __init__(self, arguments):
+        super().__init__(arguments)
+        connection_string = self.get_from_env("AZURE_SERVICEBUS_CONNECTION_STRING")
+        connection_props = parse_connection_string(connection_string)
+        self.mgmt_client = ServiceBusService(
+            service_namespace=connection_props['namespace'],
+            shared_access_key_name=connection_props['shared_access_key_name'],
+            shared_access_key_value=connection_props['shared_access_key']
+        )
+        queue = Queue(max_size_in_megabytes=40960)
```
mikeharder, line 69: I think this code should move to `global_setup()`, in part because ctors should not contain code which makes network requests, and also for symmetry with `global_cleanup()`.

sdk/servicebus/azure-servicebus/tests/perf_tests/T1_legacy_tests/receive_message_stream.py
```diff
@@ -0,0 +1,38 @@
...
+            async for msg in self.async_receiver:
+                if count >= self.args.num_messages:
+                    break
+                count += 1
+                await msg.complete()
+        else:
+            async for msg in self.async_receiver:
+                if count >= self.args.num_messages:
+                    break
+                count += 1
```
mikeharder, line 38: NIT: Missing newline

sdk/servicebus/azure-servicebus/tests/perf_tests/T1_legacy_tests/send_message.py
```diff
@@ -0,0 +1,25 @@
...
+from azure.servicebus.aio import Message as AsyncMessage
+
+
+class LegacySendMessageTest(_SendTest):
+    def __init__(self, arguments):
+        super().__init__(arguments)
+        self.data = get_random_bytes(self.args.message_size)
+
+    def run_sync(self):
+        message = Message(self.data)
```
mikeharder, line 20: Can `Message` be reused or is a new instance required for each call to `send()`?

sdk/servicebus/azure-servicebus/tests/perf_tests/T1_legacy_tests/send_message_batch.py
```diff
@@ -0,0 +1,25 @@
...
+
+from azure_devtools.perfstress_tests import get_random_bytes
+
+from azure.servicebus import BatchMessage
+
+
+class LegacySendMessageBatchTest(_SendTest):
+
+    def run_sync(self):
+        data = get_random_bytes(self.args.message_size)
```
mikeharder, line 16: Can `data`, `messages`, or `batch` be reused?

sdk/servicebus/azure-servicebus/tests/perf_tests/send_message.py
```diff
@@ -0,0 +1,23 @@
...
+
+from azure.servicebus import ServiceBusMessage
+
+class SendMessageTest(_SendTest):
+    def __init__(self, arguments):
+        super().__init__(arguments)
+        self.data = get_random_bytes(self.args.message_size)
+
+    def run_sync(self):
+        message = ServiceBusMessage(self.data)
```
mikeharder, line 18: Can `message` be reused?

sdk/servicebus/azure-servicebus/tests/perf_tests/send_message_batch.py
```diff
@@ -0,0 +1,39 @@
...
+
+from azure_devtools.perfstress_tests import get_random_bytes
+
+from azure.servicebus import ServiceBusMessage
+
+
+class SendMessageBatchTest(_SendTest):
+
+    def run_sync(self):
+        data = get_random_bytes(self.args.message_size)
```
mikeharder, line 16: Can `data` or `batch` be reused?

mikeharder (changes requested):

sdk/servicebus/azure-servicebus/tests/perf_tests/send_message_batch.py
annatisch: Data yes, but batch no

sdk/servicebus/azure-servicebus/tests/perf_tests/T1_legacy_tests/_test_base.py
```diff
@@ -66,8 +68,14 @@ def __init__(self, arguments):
...
         )
+
+    async def global_setup(self):
+        await super().global_setup()
         queue = Queue(max_size_in_megabytes=40960)
         self.mgmt_client.create_queue(self.queue_name, queue=queue)
+
+    async def setup(self):
+        await super().setup()
         self.queue_client = self.service_client.get_queue(self.queue_name)
```
mikeharder, line 79: NIT: Does getting a queue_client make a network call, or is it a client-only operation?  If it's client-only, the pattern we've typically been using is to put it into the ctor.  In short, ctor is for any code which doesn't make a network call, then async setup/global_setup are for code which does make network calls or is otherwise slow.

sdk/servicebus/azure-servicebus/tests/perf_tests/T1_legacy_tests/_test_base.py
```diff
@@ -107,24 +116,19 @@ def __init__(self, arguments):
             mode=mode,
             prefetch=self.args.num_messages,
             idle_timeout=self.args.max_wait_time)
-    
+        self.receiver.open()
+        await self.async_receiver.open()
+        await self._preload_queue()
```
mikeharder, line 121: The call to `_preload_queue` moved from `global_setup` to instance `setup`.  Intentional or mistake?

mikeharder (changes requested):

sdk/servicebus/azure-servicebus/tests/perf_tests/T1_legacy_tests/send_message.py
mikeharder: Customers are not likely to reuse `Message`, so our perf test should not either.

sdk/servicebus/azure-servicebus/tests/perf_tests/T1_legacy_tests/_test_base.py
annatisch: Yes - for T1 this does make a network call - and needs to happen after the queue is created in `global_setup`

sdk/servicebus/azure-servicebus/tests/perf_tests/T1_legacy_tests/_test_base.py
annatisch: You're right - this should be global

mikeharder (approved):

sdk/servicebus/azure-servicebus/tests/perf_tests/T1_legacy_tests/_test_base.py
```diff
@@ -0,0 +1,148 @@
...
+    def __init__(self, arguments):
+        super().__init__(arguments)
+
+        connection_string = self.get_from_env("AZURE_SERVICEBUS_CONNECTION_STRING")
+        if not _ServiceTest.service_client or self.args.service_client_per_instance:
+            _ServiceTest.service_client = ServiceBusClient.from_connection_string(connection_string)
+            _ServiceTest.async_service_client = AsyncServiceBusClient.from_connection_string(connection_string)
+
+        self.service_client = _ServiceTest.service_client
+        self.async_service_client =_ServiceTest.async_service_client
```
yunhaoling, line 47: I should admit that it took me a while to understand how it works, and I found it's compact but perfect.

My only concern here is that in the case of `self.args.service_client_per_instance`, we keep updating the class variable `_ServiceTest.service_client`, would that affect the same class `_ServiceTest` used for `not self.args.service_client_per_instance` case -- or test running are isolated, updating class variable is okay

sdk/servicebus/azure-servicebus/tests/perf_tests/T1_legacy_tests/_test_base.py
```diff
@@ -0,0 +1,148 @@
...
+
+    @staticmethod
+    def add_arguments(parser):
+        super(_ServiceTest, _ServiceTest).add_arguments(parser)
+        parser.add_argument('--message-size', nargs='?', type=int, help='Size of a single message. Defaults to 100 bytes', default=100)
+        parser.add_argument('--service-client-per-instance', action='store_true', help='Create one ServiceClient per test instance.  Default is to share a single ServiceClient.', default=False)
+        parser.add_argument('--num-messages', nargs='?', type=int, help='Number of messages to send or receive. Defaults to 100', default=100)
```
yunhaoling, line 54: dumb question: `message-size` or `message_size` or `-` and `_` don't make a different when parsing arguments

sdk/servicebus/azure-servicebus/tests/perf_tests/T1_legacy_tests/_test_base.py
```diff
@@ -0,0 +1,148 @@
...
+
+    async def _preload_queue(self):
+        data = get_random_bytes(self.args.message_size)
+        async_queue_client = self.async_service_client.get_queue(self.queue_name)
+        async with async_queue_client.get_sender() as sender:
+            for i in range(self.args.preload):
+                sender.queue_message(Message(data))
+                if i % 1000 == 0:
+                    print("Loaded {} messages".format(i))
+                    await sender.send_pending_messages()
```
yunhaoling, line 135: if message_size * 1000 exceeds max allowed size on the link, then we'll get an error.

sdk/servicebus/azure-servicebus/tests/perf_tests/_test_base.py
```diff
@@ -0,0 +1,121 @@
...
+        connection_string = self.get_from_env("AZURE_SERVICEBUS_CONNECTION_STRING")
+        if not _ServiceTest.service_client or self.args.service_client_per_instance:
+            _ServiceTest.service_client = ServiceBusClient.from_connection_string(conn_str=connection_string)
+            _ServiceTest.async_service_client = AsyncServiceBusClient.from_connection_string(conn_str=connection_string)
+
+        self.service_client = _ServiceTest.service_client
+        self.async_service_client =_ServiceTest.async_service_client
+
+    async def close(self):
+        await self.async_service_client.close()
```
yunhaoling, line 31:
```suggestion
        self.service_client.close()
        await self.async_service_client.close()
```

sdk/servicebus/azure-servicebus/tests/perf_tests/T1_legacy_tests/receive_message_batch.py
```diff
@@ -0,0 +1,21 @@
...
+
+from ._test_base import _ReceiveTest
+
+
+class LegacyReceiveMessageBatchTest(_ReceiveTest):
+    def run_sync(self):
+        batch = self.receiver.fetch_next(max_batch_size=self.args.num_messages)
+        if self.args.peeklock:
+            for msg in batch:
+                msg.complete()
```
yunhaoling, line 16: if `num_messages` is very large, then we are not able to receive all the messages in a single call.

should it be like the push mode style?
```python
count = 0
while count < self.args.num_messages:
     batch = self.receiver.fetch_next(max_batch_size=self.args.num_messages)
     if self.args.peeklock:
            for msg in batch:
                msg.complete()
     count += len(batch)
```

sdk/servicebus/azure-servicebus/tests/perf_tests/T1_legacy_tests/_test_base.py
annatisch: For the command line argument name - it should use "-", however when this value is accessed programmatically, the "-" is automagically converted to "_"  :)

sdk/servicebus/azure-servicebus/tests/perf_tests/T1_legacy_tests/_test_base.py
annatisch: I can't remember whether the T1 SDK handled this already - but given that this is just for adding messages to the queue before running the test, I'm not too concerned if an error is raised here. I've tried running this test with large messages and I don't think It's caused issues yet.

sdk/servicebus/azure-servicebus/tests/perf_tests/T1_legacy_tests/_test_base.py
yunhaoling: I didn't know that before, thanks!

sdk/servicebus/azure-servicebus/tests/perf_tests/_test_base.py
annatisch: Good catch!

sdk/servicebus/azure-servicebus/tests/perf_tests/T1_legacy_tests/receive_message_batch.py
annatisch: Yeah good plan. Updated!

sdk/servicebus/azure-servicebus/tests/perf_tests/T1_legacy_tests/_test_base.py
annatisch: You're right in that it's not overly clean - but I don't think it's going to cause problems. As soon as we assign to `self` on lines 46/47 then we're using local instance attribute.
However you're right that this is probably not best practice - I will clean it up a bit...

sdk/servicebus/azure-servicebus/tests/perf_tests/T1_legacy_tests/_test_base.py
yunhaoling: aha yes, I was thinking wrong, I thought they would go in a single batch message but they are separate messages. so it should be good.
sorry about making the noise :P

yunhaoling (approved):
LGTM

PR #16075, opened 2021-01-11
[Communication] - Phone Number Management - Added support for AAD auth

jbeauregardb:


sdk/communication/azure-communication-administration/README.md
```diff
@@ -24,9 +24,18 @@ pip install azure-communication-administration
...
 
 connection_str = os.getenv('AZURE_COMMUNICATION_SERVICE_CONNECTION_STRING')
+endpoint = os.getenv('AZURE_COMMUNICATION_SERVICE_ENDPOINT')
+
+# To use Azure Active Directory Authentication (DefaultAzureCredential) make sure to have
+# AZURE_TENANT_ID, AZURE_CLIENT_ID and AZURE_CLIENT_SECRET as env variables.
+phone_number_administration_client = PhoneNumberAdministrationClient(endpoint, DefaultAzureCredential())
+
+# You can also authenticate using your connection string
 phone_number_administration_client = PhoneNumberAdministrationClient.from_connection_string(connection_str)
```
sacheu, line 37: Do we need to remove this line (line 37), when there is an addition of line 34?

sdk/communication/azure-communication-administration/tests/test_phone_number_administration_client.py
```diff
@@ -12,15 +12,23 @@
...
 from phone_number_testcase import PhoneNumberCommunicationTestCase
 from _shared.testcase import BodyReplacerProcessor
+from azure.identity import DefaultAzureCredential
+from azure.communication.administration._shared.utils import parse_connection_str
 
 SKIP_PHONE_NUMBER_TESTS = True
 PHONE_NUMBER_TEST_SKIP_REASON= "Phone Number Administration live tests infra not ready yet"
 
-class PhoneNumberAdministrationClientTest(PhoneNumberCommunicationTestCase):
+class FakeTokenCredential(object):
```
sacheu, line 25: I see this test and the async version of the test both use this class `FakeTokenCredential`,  can we move this class into common code

sdk/communication/azure-communication-administration/README.md
jbeauregardb: We're keeping both forms of authentication, so line 37 is still relevant

sdk/communication/azure-communication-administration/tests/test_phone_number_administration_client.py
```diff
@@ -125,6 +128,19 @@ def setUp(self):
...
+    @pytest.mark.live_test_only
+    @pytest.mark.skipif(SKIP_PHONE_NUMBER_TESTS, reason=PHONE_NUMBER_TEST_SKIP_REASON)
+    def test_list_all_phone_numbers_from_managed_identity(self):
+        endpoint, access_key = parse_connection_str(self.connection_str)
+        from devtools_testutils import is_live
+        if not is_live():
+            credential = FakeTokenCredential()
+        else:
+            credential = DefaultAzureCredential()
+        phone_number_client = PhoneNumberAdministrationClient(endpoint, credential)
```
sacheu, line 140: Can you move this logic of creating a FakeToken Credential VS DefaultAzureCredential into the test utils too?
I see we are doing the same in this test and the async version of the test

sacheu:
PR looks good.

sdk/communication/azure-communication-administration/tests/test_phone_number_administration_client_async.py
```diff
@@ -126,6 +128,20 @@ def setUp(self):
...
+    async def test_list_all_phone_numbers_from_managed_identity(self):
+        endpoint, access_key = parse_connection_str(self.connection_str)
+        credential = create_token_credential()
+        phone_number_client = PhoneNumberAdministrationClient(endpoint, credential)
+        async with phone_number_client:
+            pages = phone_number_client.list_all_phone_numbers()
+            items = []
+            async for item in pages:
+                items.append(item)
+        assert len(items) > 0
```
sacheu, line 143: should we add assert to check for what we get in item? instead of just making sure the len(items) is not empty?

sdk/communication/azure-communication-administration/tests/test_phone_number_administration_client.py
```diff
@@ -125,6 +127,15 @@ def setUp(self):
...
             self.release_id = "release_id"
 
+    @pytest.mark.live_test_only
+    @pytest.mark.skipif(SKIP_PHONE_NUMBER_TESTS, reason=PHONE_NUMBER_TEST_SKIP_REASON)
+    def test_list_all_phone_numbers_from_managed_identity(self):
+        endpoint, access_key = parse_connection_str(self.connection_str)
+        credential = create_token_credential()
+        phone_number_client = PhoneNumberAdministrationClient(endpoint, credential)
+        pages = phone_number_client.list_all_phone_numbers()
+        assert pages.next()
```
sacheu, line 137: should we check if the # of phone numbers we get match with what we expected?

sdk/communication/azure-communication-administration/README.md
ankitarorabit: True, but it is quite weird to look at it in the code snippet so maybe keep only one in the first snippet.. then write a regular text saying that you could also authenticate using the connection string and then have a separate snippet containing just the separate constructor.. that way it would easier to look at and understand better.

sacheu (commented):
Looks good. Left a few suggestions.

sdk/communication/azure-communication-administration/azure/communication/administration/aio/_phone_number_administration_client_async.py
```diff
@@ -70,7 +69,7 @@ def __init__(
         self._endpoint = endpoint
         self._phone_number_administration_client = PhoneNumberAdministrationClientGen(
             self._endpoint,
-            authentication_policy=HMACCredentialsPolicy(endpoint, credential),
```
ankitarorabit, line 73: wondering if we really need the isAsync flag here?

ankitarorabit (approved):

sdk/communication/azure-communication-administration/azure/communication/administration/aio/_phone_number_administration_client_async.py
jbeauregardb: yes, because the get_authentication_policy function receives that optional parameter and passes it to the HMACCredentialPolicy object. Like this:

`return HMACCredentialsPolicy(endpoint, credential, decode_url=is_async)`

sdk/communication/azure-communication-administration/tests/test_phone_number_administration_client.py
jbeauregardb: We don't know which (or how many) phone numbers we're going to receive exactly, that's why this test only checks that we indeed returned something. This test is also copied from the current list_all_phone_numbers test from this same file, just with the added managed identity authentication.

sdk/communication/azure-communication-administration/tests/test_phone_number_administration_client_async.py
jbeauregardb: Same as other comment.

PR #16077, opened 2021-01-11
[Communication] Move core dependency to setup.py from dev_requirements in communication management package

turalf:
We had this [PR ](https://github.com/Azure/azure-sdk-for-python/pull/15969) to solve the dependency issue. 
However the tests still failing for the scheduled runs of the pipeline. Moving that into setup.py

### DIFF
--- a/sdk/communication/azure-mgmt-communication/setup.py
+++ b/sdk/communication/azure-mgmt-communication/setup.py
@@ -84,6 +84,7 @@ setup(
         'msrest>=0.5.0',
         'msrestazure>=0.4.32,<2.0.0',
         'azure-common~=1.1',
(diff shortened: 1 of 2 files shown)


sdk/communication/azure-mgmt-communication/setup.py
```diff
@@ -84,6 +84,7 @@
         'msrest>=0.5.0',
         'msrestazure>=0.4.32,<2.0.0',
         'azure-common~=1.1',
+        "azure-core<2.0.0,>=1.2.2"
```
sacheu, line 87: nit, change from " to ' in order to be consistent with line 84-86

memontic-ms (approved):

PR #16170, opened 2021-01-14
breaking changes run in ci

kristapratico:
Python is the only language that doesn't have a check for API breaks in CI: https://dev.azure.com/azure-sdk/internal/_wiki/wikis/internal.wiki/367/Capabilities-and-State. This is an effort to fix that.

Example run in Form Recognizer here: https://dev.azure.com/azure-sdk/public/_build/results?buildId=686253&view=logs&j=b70e5e73-bbb6-5567-0939-8415943fadb9&t=efe9ac6d-3fdb-50af-24e1-4319f36eecb4

This is an opt-in check for now which uses an allowlist. Long term plans will be to integrate this into the APIview tool itself.


iscai-msft (approved):
ba

eng/tox/tox.ini
```diff
@@ -266,3 +266,16 @@ deps =
 commands =
     {envbindir}/python -m pip freeze
     {envbindir}/python {toxinidir}/../../../scripts/devops_tasks/test_run_samples.py -t {toxinidir}
+
+
+[testenv:breaking]
+skipsdist = false
```
scbedd, line 272:
```suggestion
skipsdist = true
```

eng/tox/tox.ini
```diff
@@ -266,3 +266,16 @@ deps =
...
+skip_install = false
```
scbedd, line 273:
```suggestion
skip_install = true
```

eng/tox/tox.ini
```diff
@@ -266,3 +266,16 @@ deps =
...
+usedevelop = true
```
scbedd, line 274: `usedevelop` in toxland tells it to install with `-e`. We don't want to install anything for this tox env, so probably set to false? Honestly it doesn't _really_ matter given that the previous two comments shut off this being used anyway.

scbedd (approved):
Small nitpicks, on the `tox.ini`, but the CI pieces look solid!

eng/tox/tox.ini
kristapratico: Thanks Scott! I'm laughing at how I was 0/3 on those config options 😆

kristapratico: My bad for not understanding these options, we definitely need to install the current package for this check to work.

https://dev.azure.com/azure-sdk/public/_build/results?buildId=688098&view=logs&j=b70e5e73-bbb6-5567-0939-8415943fadb9&t=efe9ac6d-3fdb-50af-24e1-4319f36eecb4

PR #16181, opened 2021-01-15
[Blob]Add batch request body serialization

xiafu-msft:


sdk/storage/azure-storage-blob/azure/storage/blob/_shared/request_handlers.py
```diff
@@ -145,3 +149,121 @@ def add_metadata_headers(metadata=None):
         for key, value in metadata.items():
             headers['x-ms-meta-{}'.format(key.strip())] = value.strip() if value else value
     return headers
+
+
+def serialize_batch_body(requests, batch_id):
```
tasherif-msft, line 154: This is great, this would actually enable us to use the generated `batch_submit()` method.

sdk/storage/azure-storage-blob/azure/storage/blob/_shared/base_client.py
```diff
@@ -284,10 +282,17 @@ def _batch_send(
             enforce_https=False
         )
 
+        Pipeline._prepare_multipart_mixed_request(request) # pylint: disable=protected-access
```
tasherif-msft, line 285: What is the point of this line? isn't `multipart_mixed_info` already set [here](https://github.com/Azure/azure-sdk-for-python/pull/16181/files#diff-e6635abc4e1ab372c04ecded61ffecea0315144b8a7e3b97ffcce2b9e8b8e977L277)?

xiafu-msft, line 285: I can create a` _prepare_multipart_mixed_request` in azure-storage-blob package later, while we still cannot avoid this hacky way to deserialize the response.
```
request.multipart_mixed_info = temp
```

tasherif-msft (approved):

PR #16187, opened 2021-01-15
Refactor Artifacts at rest for safe/easy deployment

scbedd:
No more of this `stage-filtered-artifacts` stuff. We'll generate the whl/sdist/documents into the appropriate folder within the artifact, categorized by package name like we see over in .NET.


eng/pipelines/templates/stages/archetype-python-release.yml
```diff
@@ -40,24 +41,26 @@ stages:
...
                           ForRelease: true
-                    - template: /eng/pipelines/templates/steps/stage-filtered-artifacts.yml
-                      parameters:
-                        SourceFolder: ${{parameters.ArtifactName}}
-                        TargetFolder: ${{artifact.safeName}}
-                        PackageName: ${{artifact.name}}
+
                     - pwsh: |
-                        Get-ChildItem -Recurse $(Pipeline.Workspace)/${{artifact.safeName}}
+                        $packageDirectory = "${{artifact.name}}".Replace("_", "-")
```
weshaggard, line 53: Do we need to do this replacement? Can we just use the artifact name?

scbedd: Discussed offline. This is placed in lieu of touching every `ci.yml` in this PR. Will work with an admin to push the updating commit directly to master after this PR goes in.

weshaggard (approved):
Changes look reasonable based on our offline discussion about the artifact names being switched to match the package name in a follow-up.

eng/pipelines/templates/stages/archetype-python-release.yml
```diff
@@ -280,22 +283,17 @@ stages:
...
+              echo "##vso[task.setvariable variable=Package.Name]$packageDirectory"
+          - pwsh: |
+              $fileCount = (Get-ChildItem $(Pipeline.Workspace)/${{parameters.ArtifactName}}/$packageDirectory  | Measure-Object).Count
               if ($fileCount -eq 0) {
                 Write-Host "No alpha packages for ${{artifact.safeName}} to publish."
                 exit 0
               }
 
-              twine upload --repository $(DevFeedName) --config-file $(PYPIRC_PATH) $(Pipeline.Workspace)/${{artifact.safeName}}/*a*.whl
+              twine upload --repository $(DevFeedName) --config-file $(PYPIRC_PATH) $(Pipeline.Workspace)/${{parameters.ArtifactName}}/$(Package.Name)/*a*.whl
```
mitchdenny, line 295: Not really what this change is about, but I wonder if this wildcard is too permissive.

mitchdenny (approved):
Looks reasonable. I think it might be worth merging this onto a feature branch in upstream and testing the release process end to end on template.

eng/pipelines/templates/stages/archetype-python-release.yml
scbedd: What makes you say that?

scbedd:
@mitchdenny already done here: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=747165&view=logs&j=3c922a7d-1305-5297-ba3f-884e0ad84366

PR #16202, opened 2021-01-15
append code coverage over each other

seankane-msft:


eng/pipelines/templates/stages/archetype-sdk-tests.yml
```diff
@@ -0,0 +1,117 @@
...
+        AllocateResourceGroup: ${{ parameters.AllocateResourceGroup }}
+        DeployArmTemplate: ${{ parameters.DeployArmTemplate }}
+        TestTimeoutInMinutes: ${{ parameters.TestTimeoutInMinutes }}
+        TestSamples: ${{ parameters.TestSamples }}
+        Location: ${{ parameters.Location }}
+        Matrix: ${{ parameters.Matrix }}
+        CloudConfigurations: ${{ parameters.CloudConfigurations }}
+
+
+  - stage: Publish_Coverage
```
weshaggard, line 110: Why make this a separate stage/job?

eng/pipelines/templates/jobs/archetype-sdk-tests.yml
```diff
@@ -1,49 +1,85 @@
 parameters:
```
weshaggard, line 1: FYI @benbp I know you are working on refactoring this file and so you will likely conflict with this PR.

benbp: @weshaggard I chatted with Sean and was waiting for these changes to land first, hence why I did Java instead.

eng/pipelines/templates/stages/archetype-sdk-tests.yml
```diff
@@ -0,0 +1,117 @@
...
+  - name: TestSamples
+    type: boolean
+    default: false
+  - name: Location
+    type: string
+    default: ''
+  - name: Matrix
+    type: object
+    default:
+      Linux_Python35:
```
benbp, line 53: This is duplicated with the job parameters. I think you can keep it here and remove it there (though I'm going to delete both subsequently, so its optional)

eng/pipelines/templates/stages/archetype-sdk-tests.yml
benbp: I would call the stage something other than `Build`. Maybe `LiveTest` or just leave it blank

eng/pipelines/templates/stages/archetype-sdk-tests.yml
```diff
@@ -0,0 +1,117 @@
...
+        DeployArmTemplate: ${{ parameters.DeployArmTemplate }}
+        TestTimeoutInMinutes: ${{ parameters.TestTimeoutInMinutes }}
+        TestSamples: ${{ parameters.TestSamples }}
+        Location: ${{ parameters.Location }}
+        Matrix: ${{ parameters.Matrix }}
+        CloudConfigurations: ${{ parameters.CloudConfigurations }}
+
+
+  - stage: Publish_Coverage
+    dependsOn: Build
```
benbp, line 111: The dependsOn is unnecessary, since it will implicitly do the stages in list order. But as Wes asked, you can probably just have this in the job list instead.

eng/pipelines/templates/stages/archetype-sdk-tests.yml
seankane-msft: @weshaggard the code coverage is dependent on all the Python 3.9 jobs.

eng/pipelines/templates/jobs/publish-coverage.yml
```diff
@@ -0,0 +1,69 @@
...
+
+      - task: PublishCodeCoverageResults@1
+        displayName: 'Publish Code Coverage to DevOps'
+        continueOnError: true
+        condition: and(succeededOrFailed(), ${{ parameters.RunCoverage }})
+        inputs:
+          codeCoverageTool: Cobertura
+          summaryFileLocation: '$(Build.SourcesDirectory)/coverage.xml'
+
+  - job: Show_all_coverage_files
```
weshaggard, line 65: Why not do this as another step in the previous job? We don't want to use another agent machine to just do that.

eng/pipelines/templates/stages/archetype-sdk-tests.yml
weshaggard: So this this is an aggregate coverage report publishing? Is there advantage to do it this way as opposed to letting each job publish the results individually? If there is an advantage to doing an aggregate then this approach should work but I'd suggest updating the name to indicate that is what is doing.

seankane-msft: If each job publishes individually they will overwrite previous reports. This method guarantees that all jobs have an accurate code coverage report.

weshaggard: I'm not opposed to doing an aggregate report and publishing but just curious why we couldn't solve this problem by giving each job a unique name?

seankane-msft: @weshaggard yes, the code coverage reports are currently over-writing each other each time they are published. This ensures that one report is published with all the code coverage for the package in one place.

eng/pipelines/templates/steps/build-test.yml
```diff
@@ -55,17 +55,29 @@ steps:
...
+      }
+    displayName: 'Show .coverage files'
+    condition: and(succeededOrFailed(), ${{ parameters.RunCoverage }})
+
+  - task: PublishPipelineArtifact@1
+    displayName: 'Publish .coverage files'
+    continueOnError: true
+    condition: and(succeededOrFailed(), ${{ parameters.RunCoverage }})
+    inputs:
+      targetPath: '$(Build.SourcesDirectory)/_coverage'
```
weshaggard, line 71: Does this work? DevOps generally doesn't like you uploading an artifact with the same name from multiple jobs/stages. Usually you have to have a unique name.

seankane-msft: Yes it uploads the artifacts in a folder named after the job so each job is unique.

eng/pipelines/templates/jobs/publish-coverage.yml
```diff
@@ -0,0 +1,74 @@
...
+        continueOnError: false
+        condition: and(succeededOrFailed(), ${{ parameters.RunCoverage }})
+
+      - task: DownloadPipelineArtifact@2
+        inputs:
+          source: current
+          path: '$(Build.SourcesDirectory)/_all_coverage_files'
+
+      - script: |
+          python -m pip install pip==20.1
```
scbedd, line 40: Can we instead place `pip==20.2` in the `eng/ci_tools.txt`? That way it's pinned in an expected location.

sdk/storage/tests_invoke.yml
```diff
@@ -3,7 +3,7 @@ parameters:
   InjectedPackages: ''
 
 jobs:
-  - template: ../../eng/pipelines/templates/jobs/archetype-sdk-tests.yml
```
scbedd, line 6: Is this file necessary anymore? We removed the reference to tests_invoke.yml in the storage tests.yml IIRC.

scbedd: Aha. I'd prefer that `tests.yml` call into `tests_invoke.yml`. 

The reason that it was set up a bit non-standard was to minimize the duplication of the environment variables being passed to `archetype-sdk-tests`.

scripts/devops_tasks/tox_harness.py
```diff
@@ -112,7 +112,7 @@ def collect_tox_coverage_files(targeted_packages):
 
     # coverage report has paths starting .tox and azure
     # coverage combine fixes this with the help of tox.ini[coverage:paths]
-    combine_coverage_files(targeted_packages)
+    # combine_coverage_files(targeted_packages) # Skipping this, upload all .coverage-{package-name} files
```
scbedd, line 115: I think we need agree on how `coverage` artifacts will be handled. Specifically, in some places, we comment out `combine` of coverage. However, we still talk about as if we were combining down below

```
logging.info("Visible uncombined .coverage files: {}".format(coverage_files))
```

scbedd (changes requested):
Couple structural questions to do with storage live tests. Need to clean up the bounds around `coverage` in setup_execute and tox_harness. 

The vast majority is great! Couple questions and let's get this merged.

scripts/devops_tasks/create_coverage.py
```diff
@@ -0,0 +1,74 @@
...
+        run_check_call(commands, root_dir, always_exit = False)
+    else:
+        logging.error("Coverage file is not available in {} to generate coverage XML".format(coverage_dir))
+
+
+def fix_dot_coverage_file(coverage_file):
+    print("running 'fix_dot_coverage_file' on {}".format(coverage_file))
+
+    out = None
+    with open(coverage_file) as cov_file:
```
scbedd, line 63: This specific function makes me side-eye it.

I know you've been dumping content in the publish-coverage job. You know what's up, but this specific line really scares me.

Can you talk me through it in a quick call?

scripts/devops_tasks/tox_harness.py
seankane-msft: Changed this to logging information about what files are found.

sdk/storage/tests_invoke.yml
seankane-msft: fixed!

scripts/devops_tasks/create_coverage.py
seankane-msft: Result: change the replace to replace anything with the regex `/.tox/**/site-packages` to a blank string

scbedd (approved):
Pending our discussed items. G2g.

eng/pipelines/templates/jobs/publish-coverage.yml
```diff
@@ -0,0 +1,72 @@
...
+  - name: ServiceDirectory
+    type: string
+    default: ''
+  - name: Artifacts
+    type: object
+    default: []
+  - name: DependsOn
+    type: string
+    default: ''
+  - name: RunCoverage
```
weshaggard, line 11: Is this interesting any more? This step isn't doing any running but only publishing.

seankane-msft: Nope, removing

eng/pipelines/templates/jobs/publish-coverage.yml
```diff
@@ -0,0 +1,69 @@
...
+        displayName: Show all coverage files
+
+      - task: PythonScript@0
+        displayName: 'Create Coverage Report'
+        inputs:
+          scriptPath: 'scripts/devops_tasks/create_coverage.py'
+
+      - pwsh: |
+          Get-Content $(Build.SourcesDirectory)\.coverage
+        continueOnError: true
```
weshaggard, line 57: If you If you plan to keep these steps you should given them a displayname so it is clear what they are doing in the logs and it might be worth combining these 2.

sdk/cosmos/tests.yml
```diff
@@ -1,7 +1,7 @@
 trigger: none
 
 jobs:
```
benbp, line 3: You need this top level `jobs` field to be renamed to `stages` too for all these `tests.yml` files. Looks like you did it for a few services but not for the majority.

benbp (dismissed):

PR #16211, opened 2021-01-16
[Blob] Added support for container SAS on batch operations

tasherif-msft:


sdk/storage/azure-storage-blob/azure/storage/blob/_shared/base_client.py
```diff
@@ -262,17 +262,20 @@ def _create_pipeline(self, credential, **kwargs):
         return config, Pipeline(config.transport, policies=policies)
 
     def _batch_send(
-        self, *reqs,  # type: HttpRequest
+        self,
+        *reqs,  # type: HttpRequest
+        container_name,  # type: str
```
xiafu-msft, line 267: _batch_send will not work if we add delete_blobs on blob_service_client because there won't be a container name there

sdk/storage/azure-storage-blob/tests/recordings/test_container.test_batch_blobs_with_container_sas.yaml
```diff
@@ -0,0 +1,311 @@
+interactions:
```
xiafu-msft, line 1: this file can be removed

sdk/storage/azure-storage-blob/azure/storage/blob/_shared/base_client.py
tasherif-msft: This is only intended for container level batch calls. Once we start having discussions about service level I feel we will probably change the current flow anyway (and maybe even remove or change `batch_send` greatly) and we might utilize generated code. However if you want to avoid passing container name here I can put another workaround. Let me know

xiafu-msft: it's in base_client, that means the change should be in every package, we should make it more generic.
Also we should serialize the query parameter instead of hardcode it. Then less work will be there if we want to add batch on blob_service_client.
If we make it specific for container, higher risk to break container.delete_blobs(), since you need to touch two apis to make them both work.
We can avoid using batch_send if the generated batch_submit() works. It's your call

tasherif-msft: offline: will proceed to make it more generic and later on we will utilize the `batch_submit` operation

sdk/storage/azure-storage-blob/tests/test_container.py
```diff
@@ -1166,9 +1166,47 @@ def test_delete_blobs_simple(self, resource_group, location, storage_account, st
         assert response[1].status_code == 202
         assert response[2].status_code == 202
 
+    @pytest.mark.live_test_only
     @pytest.mark.skipif(sys.version_info < (3, 0), reason="Batch not supported on Python 2.7")
-    @GlobalResourceGroupPreparer()
-    @StorageAccountPreparer(random_name_enabled=True, location="canadacentral", name_prefix='storagename')
```
xiafu-msft, line 1171: it seems tags only works in canadacentral (at least previously), are we sure removing this will not break live test?

xiafu-msft (approved):

Issue #16094, opened 2021-01-12
[Event Hubs] Update docs/README


swathipil:
* In the [Event Hubs README](https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/eventhub/azure-eventhub/README.md), under section `Create client using the azure-identity library`, add the azure-identity installation step: `pip install azure-identity`
> similar to the [form recognizer README AAD section](https://github.com/Azure/azure-sdk-for-python/tree/master/sdk/formrecognizer/azure-ai-formrecognizer#create-the-client-with-an-azure-active-directory-credential)

yunhaoling:
we also need to mention async transport support (aiohttp) if people want to use azure-identity in async world, could just be a short sentence.

PR #16215, opened 2021-01-16
[Event Hubs] updated README with install steps for azure-identity and aiohttp

swathipil:
addressing: #16094

### DIFF
--- a/sdk/eventhub/azure-eventhub/README.md
+++ b/sdk/eventhub/azure-eventhub/README.md
@@ -56,6 +56,14 @@ Alternately, one can use a Credential object to authenticate via AAD with the az
 [TokenCredential](https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/core/azure-core/azure/core/credentials.py)
 protocol. There are implementations of the `TokenCredential` protocol available in the
 [azure-identity package](https://pypi.org/project/azure-identity/). The host name is of the format `<yournamespace.servicebus.windows.net>`.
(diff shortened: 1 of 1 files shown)


sdk/eventhub/azure-eventhub/README.md
```diff
@@ -56,6 +56,14 @@ Alternately, one can use a Credential object to authenticate via AAD with the az
...
 protocol. There are implementations of the `TokenCredential` protocol available in the
 [azure-identity package](https://pypi.org/project/azure-identity/). The host name is of the format `<yournamespace.servicebus.windows.net>`.
+- To use the `TokenCredential` type, or other credential types provided with the Azure SDK, please install the `azure-identity` package:
+
+```pip install azure-identity```
+
+- Additionally, to use the async API supported on Python 3.5+, please install the async transport [`aiohttp`](https://pypi.org/project/aiohttp/):
+
+```pip install aiohttp```
+
```
yunhaoling, line 66: nitpicky: I don't think the blank lines are needed?
also tweak the words a bit, just want to be clear on that `TokenCredential` is a protocol which is not a concrete credential class, and aiohttp is one option for async transport support, users could have install other async transport lib.

```suggestion
- To use the credential types provided by `azure-identity`, please install the package:
```pip install azure-identity```
- Additionally, to use the async API supported on Python 3.5+, you must first install an async transport, such as [`aiohttp`](https://pypi.org/project/aiohttp/):
```pip install aiohttp```
```

yunhaoling (approved):

yunhaoling:
I think it would be nice if we could check servicebus readme and update servicebus readme accordingly

PR #16253, opened 2021-01-20
Download code cov

seankane-msft:
Steps to do this:

- [ ]  Create coverage.xml report
- [ ]  Step for renaming coverage.xml -> coverage-{package-name}.xml
- [ ]  Publish the xml file to pipeline artifacts
- [ ]  After completion of all runs, download all pipeline xml artifacts
- [ ]  Run python script to combine all xml files into one single coverage.xml file
- [ ]  Run the PublishCodeCoverageResults@1 step to publish final results to Code Coverage tab

### DIFF
--- a/eng/pipelines/templates/steps/build-test.yml
+++ b/eng/pipelines/templates/steps/build-test.yml
@@ -53,14 +53,6 @@ steps:
         ${{ parameters.ToxEnvParallel }}
     env: ${{ parameters.EnvVars }}
 
-  - pwsh: |
-      mv coverage.xml coverage-$(BuildTargetingString).xml
-    displayName: 'Rename coverage.xml file'
-
-  - task: PublishPipelineArtifact@1
-    inputs:
-      targetPath: $(System.DefaultWorkingDirectory)/coverage-*.xml
-
   - ${{ parameters.AfterTestSteps }}
 
   - task: PublishTestResults@2
--- a/eng/pipelines/templates/steps/build-test.yml
+++ b/eng/pipelines/templates/steps/build-test.yml
@@ -33,7 +33,7 @@ steps:
 
   - ${{if eq(variables['System.TeamProject'], 'internal') }}:
     - template: ../steps/auth-dev-feed.yml
-      parameters: 
+      parameters:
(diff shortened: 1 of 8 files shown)


eng/pipelines/templates/steps/publish-coverage.yml
```diff
@@ -9,10 +9,22 @@ steps:
     continueOnError: true
     condition: and(ne(variables['codecov-python-repository-token'], ''), succeededOrFailed(), ${{ parameters.RunCoverage }})
 
-  - task: PublishCodeCoverageResults@1
-    displayName: 'Publish Code Coverage to DevOps'
-    continueOnError: true
+  - pwsh: |
+      mv coverage.xml coverage-$(BuildTargetingString).xml; ls
```
weshaggard, line 13: Why do you need to rename the file? 

If you do need to rename and you continue to use pwsh to do it please not use alias's like mv and ls as those might do things differently on different OS's. It is generally advised to use the ps cmdlets instead (see http://powershell-guru.com/powershell-best-practice-1-use-full-cmdlet-name-not-alias/).

seankane-msft: Some pipelines test multiple packages at once and these overwrite the coverage.xml file. I will change to the ps cmdlets, thanks for the suggestion.

weshaggard: Is it possible to have the tool that produces the coverage report to output the file directly with the new name instead of needing to do a rename? If not then this should be fine.

eng/pipelines/templates/steps/build-test.yml
```diff
@@ -50,6 +50,7 @@ steps:
         --service="${{ parameters.ServiceDirectory }}"
         --toxenv="${{ parameters.ToxTestEnv }}"
         --injected-packages="${{ parameters.InjectedPackages }}"
+        --package-name="{{ parameters.BuildTargetingString }}"
```
weshaggard, line 53: I suspect it might be more flexible to just pass the name of the coverage file down instead of passing the package name. Especially given you use that full file name in other places it would make it easier to keep things in sync, and there will be less coupling in your script.

Issue #16047, opened 2021-01-08
[EventHubs] Custom End Point Support


yunhaoling:
azure-sdk board issue: https://github.com/Azure/azure-sdk/issues/1367

design:
https://apiview.dev/Assemblies/Review/fa294a29022e42c9b9b1b4d998e67821?diffRevisionId=3ed3b2c427c94b89bd7b00bcd6aafdea

sample:
https://gist.github.com/jsquire/f0948c9a5ffe2741b03d312dcf8d899e#gistcomment-3353486

js pr:
https://github.com/Azure/azure-sdk-for-js/pull/12909

test setting up:
https://gist.github.com/chradek/d4d6ce32ef4c5b46bf55d6e1b979b991

PR #16295, opened 2021-01-22
[EventHubs] Support for Custom endpoint adddress and custom certificate 

yunhaoling:
addressing issues: https://github.com/Azure/azure-sdk-for-python/issues/16047
also see gist: https://gist.github.com/jsquire/f0948c9a5ffe2741b03d312dcf8d899e

# Champion Scenarios
## Connecting on a restricted network
A local hospital employs a strict set of network policies to ensure the safety of their network and the security of their patient information. As part of this strategy, all traffic is routed through a security proxy which performs a threat analysis and logging of connections.

In order to allow for trusted internal applications to embrace cloud services, a special bank of IP addresses has been reserved which passes through a secure DMZ path rather than the security proxy. In order for trusted connections to work appropriately, they must resolve to one of the reserved IP addresses using a CNAME configured in the local DNS service.

Because of these restrictions, applications are unable to perform direct connections to the Azure Event Hubs service using the standard endpoint address and require the ability to specify a custom host name to ensure they route through the proper intermediary for the connection to be made.

## sample code:
```python
# use constructor
consumer_client = EventHubConsumerClient(
        "<FULLY QUALIFIED NAMESPACE>",
        "<EVENT HUB NAME>",
        "<CONSUMER GROUP>",
        credential,
        custom_endpoint_address="sb://eventhubs.mycompany.local",
        connection_verify="path_to_cert_file"
)

# use connection string
consumer_client = EventHubConsumerClient.from_connection_string (
        "<CONNECTION STRING>",
        "<CONSUMER GROUP>",
        custom_endpoint_address="sb://eventhubs.mycompany.local",
        connection_verify="path_to_cert_file"
)
```

## Test:
by following the steps in
- https://github.com/Azure/azure-sdk-for-js/pull/12909#issue-540828683
- https://gist.github.com/chradek/d4d6ce32ef4c5b46bf55d6e1b979b991
I'm able to set an Azure Applicate Gateway and verify this feature

## Pending:
pending uamqp v1.2.14 release: https://github.com/Azure/azure-uamqp-python/pull/202


sdk/eventhub/azure-eventhub/azure/eventhub/_consumer_client.py
```diff
@@ -102,7 +102,9 @@ class EventHubConsumerClient(ClientBase):
      evaluation regardless of the load balancing strategy.
      Greedy strategy is used by default.
     :paramtype load_balancing_strategy: str or ~azure.eventhub.LoadBalancingStrategy
-
+    :keyword str custom_endpoint_address: The custom endpoint address to connect to. Default is None.
+    :keyword str connection_verify: Custom SSL certificate verification. It could be set to the path to a CA_BUNDLE
```
yunhaoling, line 106: add a note that it only works for custom endpoint

sdk/eventhub/tests.yml
```diff
@@ -5,6 +5,7 @@ jobs:
     parameters:
       AllocateResourceGroup: 'false'
       ServiceDirectory: eventhub
+      InjectedPackages: $(WheelLocationsList)
```
yunhaoling, line 8: MARK: this is for testing unreleased uamqp v1.2.14 wheels only, will be removed after uamqp 1.2.14 is released

sdk/eventhub/azure-eventhub/azure/eventhub/_consumer_client.py
```diff
@@ -102,7 +102,10 @@ class EventHubConsumerClient(ClientBase):
      evaluation regardless of the load balancing strategy.
      Greedy strategy is used by default.
     :paramtype load_balancing_strategy: str or ~azure.eventhub.LoadBalancingStrategy
-
+    :keyword str custom_endpoint_address: The custom endpoint address to connect to. Default is None.
```
annatisch, line 105: I think this description could probably be more descriptive about the specific scenario this supports. I don't want people to fall into it by accident :)

annatisch (approved):

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_client_base_async.py
```diff
@@ -149,6 +152,9 @@ async def _create_auth_async(self) -> authentication.JWTTokenAsync:
             timeout=self._config.auth_timeout,
             http_proxy=self._config.http_proxy,
             transport_type=self._config.transport_type,
+            hostname=self._config.custom_endpoint_hostname,
+            port=self._config.connection_port,
+            verify=self._config.connection_verify
```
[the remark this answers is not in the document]
yunhaoling: nah.. actually `verify` in uamqp only supports str (path to the cert file) and we read data from the file.

https://github.com/Azure/azure-uamqp-python/blob/master/uamqp/authentication/cbs_auth.py#L203-L204
```python
    :param verify: The path to a user-defined certificate.
    :type verify: str
```

```python
self.cert_file = verify
...
cert = self.cert_file or certifi.where()
        with open(cert, 'rb') as cert_handle:
            cert_data = cert_handle.read()
```

sdk/eventhub/azure-eventhub/azure/eventhub/_consumer_client.py
yunhaoling: description updated

swathipil (approved):
lgtm

PR #16292, opened 2021-01-22
rename list_dimension_values

xiangyan99:

PR #16298, opened 2021-01-22
[metricsadvisor] fix for live tests

kristapratico:
https://dev.azure.com/azure-sdk/internal/_build/results?buildId=697835&view=ms.vss-test-web.build-test-results-tab&runId=16138210&resultId=100151&paneView=debug


xiangyan99:
It is covered in #16292

PR #16302, opened 2021-01-22
[Perf] Include running average in per-second output

mikeharder:
Same change in .NET: Azure/azure-sdk-for-net#18071


annatisch (approved):
LGTM :)

PR #16309, opened 2021-01-22
[Datalake] Removed list_paths manual paging and deserialization

tasherif-msft:
Using the new generator the list paths generated operation is able to handle paging and property deserialization which allows us to remove a lot of manually added code. This resolves https://github.com/Azure/azure-sdk-for-python/issues/16230


sdk/storage/azure-storage-blob/tests/test_largest_block_blob.py
```diff
@@ -259,7 +259,7 @@ def test_create_largest_blob_from_stream_single_upload_without_network(self, res
         payload_dropping_policy = PayloadDroppingPolicy()
         credential_policy = _format_shared_key_credential(storage_account.name, storage_account_key)
         self._setup(storage_account, storage_account_key, [payload_dropping_policy, credential_policy],
-                    max_single_put_size=LARGEST_SINGLE_UPLOAD_SIZE)
+                    max_single_put_size=LARGEST_SINGLE_UPLOAD_SIZE+1)
```
kasobol-msft, line 262: why?

sdk/storage/azure-storage-file-datalake/azure/storage/filedatalake/aio/_file_system_client_async.py
```diff
@@ -385,7 +384,7 @@ def get_paths(self, path=None,  # type: Optional[str]
                         recursive=True,  # type: Optional[bool]
                         max_results=None,  # type: Optional[int]
                         **kwargs):
-        # type: (...) -> ItemPaged[PathProperties]
+        # type: (...) -> AsyncItemPaged[PathProperties]
```
kasobol-msft, line 387: isn't this a break? what if somebody checks type? should we alias?

tasherif-msft: We have previously fixed incorrect type annotations. The return object in this method is `AsyncItemPaged` so this was a mistake that needed to be fixed.

sdk/storage/azure-storage-blob/tests/test_largest_block_blob.py
tasherif-msft: I added this in an attempt to resolve the flaky test `test_create_largest_blob_from_stream_single_upload_without_network` which seems to sometimes do a `PUT` twice rather than once causing one of the assertions to fail.

kasobol-msft (approved):

PR #16316, opened 2021-01-25
[Core] Added enum meta class

annatisch:


lmazuel (approved):

xiangyan99:
Because of https://github.com/PyCQA/astroid/issues/713

pylint does not like class MyCustomEnum(with_metaclass(CaseInsensitiveEnumMeta, str, Enum)):

Do you think we want to add this information into dev_doc?

PR #16318, opened 2021-01-25
Send method changes 

rakshith91:
Fixes https://github.com/Azure/azure-sdk-for-python/issues/16350

- Rename send_events to send
- Add custom event dictionary support 
-  Check only first instance of list

### DIFF
--- a/sdk/eventgrid/azure-eventgrid/azure/eventgrid/_helpers.py
+++ b/sdk/eventgrid/azure-eventgrid/azure/eventgrid/_helpers.py
@@ -95,7 +95,7 @@ def _is_cloud_event(event):
 
 def _is_eventgrid_event(event):
     # type: (Any) -> bool
-    required = ('subject', 'event_type', 'data', 'data_version')
+    required = ('subject', 'event_type', 'data', 'data_version', 'id', 'event_time')
     try:
         return all([prop in event for prop in required])
     except TypeError:
--- a/sdk/eventgrid/azure-eventgrid/azure/eventgrid/_helpers.py
+++ b/sdk/eventgrid/azure-eventgrid/azure/eventgrid/_helpers.py
@@ -97,7 +97,7 @@ def _is_eventgrid_event(event):
     # type: (Any) -> bool
     required = ('subject', 'event_type', 'data', 'data_version')
     try:
-        return all([_ in event for _ in required])
+        return all([prop in event for prop in required])
     except TypeError:
         return False
 
--- a/sdk/eventgrid/azure-eventgrid/azure/eventgrid/_helpers.py
+++ b/sdk/eventgrid/azure-eventgrid/azure/eventgrid/_helpers.py
@@ -93,6 +93,14 @@ def _is_cloud_event(event):
     except TypeError:
         return False
 
+def _is_eventgrid_event(event):
+    # type: (Any) -> bool
+    required = ('subject', 'event_type', 'data', 'data_version')
+    try:
+        return all([_ in event for _ in required])
+    except TypeError:
+        return False
+
 def _eventgrid_data_typecheck(event):
     try:
         data = event.get('data')
--- a/sdk/eventgrid/azure-eventgrid/azure/eventgrid/_publisher_client.py
+++ b/sdk/eventgrid/azure-eventgrid/azure/eventgrid/_publisher_client.py
@@ -27,6 +27,7 @@ from ._helpers import (
     _get_endpoint_only_fqdn,
     _get_authentication_policy,
     _is_cloud_event,
+    _is_eventgrid_event,
     _eventgrid_data_typecheck
 )
 from ._generated._event_grid_publisher_client import EventGridPublisherClient as EventGridPublisherClientImpl
@@ -98,7 +99,7 @@ class EventGridPublisherClient(object):
         return policies
 
     @distributed_trace
-    def send_events(self, events, **kwargs):
+    def send(self, events, **kwargs):
         # type: (SendType, Any) -> None
         """Sends event data to topic hostname specified during client initialization.
 
@@ -113,24 +114,22 @@ class EventGridPublisherClient(object):
         if not isinstance(events, list):
             events = cast(ListEventType, [events])
 
-        if all(isinstance(e, CloudEvent) for e in events) or all(_is_cloud_event(e) for e in events):
+        if isinstance(events[0], CloudEvent) or _is_cloud_event(events[0]):
             try:
                 events = [cast(CloudEvent, e)._to_generated(**kwargs) for e in events] # pylint: disable=protected-access
             except AttributeError:
                 pass # means it's a dictionary
             kwargs.setdefault("content_type", "application/cloudevents-batch+json; charset=utf-8")
-            self._client.publish_cloud_event_events(
+            return self._client.publish_cloud_event_events(
                 self._endpoint,
                 cast(List[InternalCloudEvent], events),
                 **kwargs
                 )
-        elif all(isinstance(e, EventGridEvent) for e in events) or all(isinstance(e, dict) for e in events):
-            kwargs.setdefault("content_type", "application/json; charset=utf-8")
-            for event in events:
-                _eventgrid_data_typecheck(event)
-            self._client.publish_events(self._endpoint, cast(List[InternalEventGridEvent], events), **kwargs)
-        elif all(isinstance(e, CustomEvent) for e in events):
-            serialized_events = [dict(e) for e in events] # type: ignore
-            self._client.publish_custom_event_events(self._endpoint, cast(List, serialized_events), **kwargs)
         else:
-            raise ValueError("Event schema is not correct.")
+            kwargs.setdefault("content_type", "application/json; charset=utf-8")
+            if isinstance(events[0], EventGridEvent) or _is_eventgrid_event(events[0]):
+                for event in events:
+                    _eventgrid_data_typecheck(event)
+            elif isinstance(events[0], CustomEvent):
+                events = [dict(e) for e in events] # type: ignore
+            return self._client.publish_custom_event_events(self._endpoint, cast(List, events), **kwargs)
--- a/sdk/eventgrid/azure-eventgrid/azure/eventgrid/_publisher_client.py
+++ b/sdk/eventgrid/azure-eventgrid/azure/eventgrid/_publisher_client.py
@@ -33,7 +33,7 @@ from ._helpers import (
 from ._generated._event_grid_publisher_client import EventGridPublisherClient as EventGridPublisherClientImpl
 from ._policies import CloudEventDistributedTracingPolicy
 from ._version import VERSION
-from ._generated.models import CloudEvent as InternalCloudEvent, EventGridEvent as InternalEventGridEvent
+from ._generated.models import CloudEvent as InternalCloudEvent
 
 if TYPE_CHECKING:
     # pylint: disable=unused-import,ungrouped-imports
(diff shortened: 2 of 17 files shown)


sdk/eventgrid/azure-eventgrid/azure/eventgrid/_helpers.py
```diff
@@ -93,6 +93,14 @@ def _is_cloud_event(event):
     except TypeError:
         return False
 
+def _is_eventgrid_event(event):
+    # type: (Any) -> bool
+    required = ('subject', 'event_type', 'data', 'data_version')
```
yunhaoling, line 98: dumb question: I see that in the `EventGridEvent` class, we have other required props:
```python
    _validation = {
        'id': {'required': True},
        'subject': {'required': True},
        'data': {'required': True},
        'event_type': {'required': True},
        'event_time': {'required': True},
        'metadata_version': {'readonly': True},
        'data_version': {'required': True},
    }
```

so we don't need id, event_time here?

sdk/eventgrid/azure-eventgrid/azure/eventgrid/_helpers.py
```diff
@@ -93,6 +93,14 @@ def _is_cloud_event(event):
...
+    try:
+        return all([_ in event for _ in required])
```
yunhaoling, line 100: nitpicky:

we could improve the readability by naming:
```python
return all([prop in event for prop in required_prop])
```

sdk/eventgrid/azure-eventgrid/azure/eventgrid/_helpers.py
rakshith91: right - `id` is required, but when not provided, it's a random uuid4 ; similarly `event_time` is the current datetime.

rakshith91: We document id and event_time as optional and have them as kwargs

sdk/eventgrid/azure-eventgrid/azure/eventgrid/_helpers.py
rakshith91, line 100:
```suggestion
        return all([prop in event for prop in required])
```

sdk/eventgrid/azure-eventgrid/azure/eventgrid/_helpers.py
rakshith91, line 98:
```suggestion
    required = ('subject', 'event_type', 'data', 'data_version', 'id', 'event_time')
```

sdk/eventgrid/azure-eventgrid/azure/eventgrid/_publisher_client.py
```diff
@@ -113,27 +117,24 @@ def send_events(self, events, **kwargs):
         if not isinstance(events, list):
             events = cast(ListEventType, [events])
 
-        if all(isinstance(e, CloudEvent) for e in events) or all(_is_cloud_event(e) for e in events):
+        if isinstance(events[0], CloudEvent) or _is_cloud_event(events[0]):
```
swathipil, line 120: are we only checking that the first event is a cloud event, rather than all the events, because the service will catch it if they are not all cloud events being sent?

sdk/eventgrid/azure-eventgrid/azure/eventgrid/_publisher_client.py
```diff
@@ -113,27 +117,24 @@ def send_events(self, events, **kwargs):
...
-            self._client.publish_cloud_event_events(
+            return self._client.publish_cloud_event_events(
                 self._endpoint,
                 cast(List[InternalCloudEvent], events),
                 **kwargs
                 )
-        elif all(isinstance(e, EventGridEvent) for e in events) or all(isinstance(e, dict) for e in events):
-            kwargs.setdefault("content_type", "application/json; charset=utf-8")
+        kwargs.setdefault("content_type", "application/json; charset=utf-8")
+        if isinstance(events[0], EventGridEvent) or _is_eventgrid_event(events[0]):
```
swathipil, line 132: same as above: are we only checking that the first event, rather than all the events, is an eg event because the service will catch it if they are not all eg events being sent?

sdk/eventgrid/azure-eventgrid/azure/eventgrid/_publisher_client.py
```diff
@@ -113,27 +117,24 @@ def send_events(self, events, **kwargs):
...
             for event in events:
                 _eventgrid_data_typecheck(event)
-            self._client.publish_events(self._endpoint, cast(List[InternalEventGridEvent], events), **kwargs)
-        elif all(isinstance(e, CustomEvent) for e in events):
-            serialized_events = [dict(e) for e in events] # type: ignore
-            self._client.publish_custom_event_events(self._endpoint, cast(List, serialized_events), **kwargs)
-        else:
-            raise ValueError("Event schema is not correct.")
+        elif isinstance(events[0], CustomEvent):
```
swathipil, line 135: same question as above for checking a single custom event

sdk/eventgrid/azure-eventgrid/azure/eventgrid/_publisher_client.py
```diff
@@ -113,27 +117,24 @@ def send_events(self, events, **kwargs):
...
                 cast(List[InternalCloudEvent], events),
                 **kwargs
                 )
-        elif all(isinstance(e, EventGridEvent) for e in events) or all(isinstance(e, dict) for e in events):
-            kwargs.setdefault("content_type", "application/json; charset=utf-8")
+        kwargs.setdefault("content_type", "application/json; charset=utf-8")
+        if isinstance(events[0], EventGridEvent) or _is_eventgrid_event(events[0]):
             for event in events:
                 _eventgrid_data_typecheck(event)
-            self._client.publish_events(self._endpoint, cast(List[InternalEventGridEvent], events), **kwargs)
```
swathipil, line 131: did you remove this line with `publish_events` for EventGridEvents since `publish_custom_event_events` accomplishes the same thing?

sdk/eventgrid/azure-eventgrid/azure/eventgrid/aio/_publisher_client_async.py
```diff
@@ -28,10 +28,11 @@
     _get_endpoint_only_fqdn,
     _get_authentication_policy,
     _is_cloud_event,
+    _is_eventgrid_event,
```
swathipil, line 31: same questions as the sync publisher client

sdk/eventgrid/azure-eventgrid/tests/test_eg_publisher_client_async.py
```diff
@@ -287,4 +287,4 @@ async def test_send_and_close_async_session(self, resource_group, eventgrid_topi
                     data = "cloudevent",
                     type="Sample.Cloud.Event"
                     )
-            await client.send_events(cloud_event)
\ No newline at end of file
+            await client.send_events(cloud_event)
```
swathipil, line 290: I think you meant this?
```suggestion
            await client.send(cloud_event)
```

sdk/eventgrid/azure-eventgrid/azure/eventgrid/_publisher_client.py
rakshith91: this is intentional - service doesn't have partial success - so either all are cloud events are not - and we avoid more client side validation.

sdk/eventgrid/azure-eventgrid/azure/eventgrid/_publisher_client.py
rakshith91: same as above :)

sdk/eventgrid/azure-eventgrid/azure/eventgrid/_publisher_client.py
rakshith91: same as above :)

sdk/eventgrid/azure-eventgrid/azure/eventgrid/_publisher_client.py
rakshith91: that's right - this was we are able to allow a dict representation for a customevent  as well

sdk/eventgrid/azure-eventgrid/azure/eventgrid/aio/_publisher_client_async.py
rakshith91: similar design as sync

sdk/eventgrid/azure-eventgrid/azure/eventgrid/_publisher_client.py
swathipil: I'm assuming that this means `publish_custom_event_events` and `publish_events` are interchangeable. And, in that case, is there a need for separate `publish_events` and `publish_custom_event_events` methods? 

would it be better to only have the `publish_events` method generated by the swagger since it can be used for sending both eg events and custom events? Or is that not something we want to do because we shouldn't change the swagger since it is used by all languages?

swathipil: ^^ to address above and summarize discussion with @rakshith91 : these are not the same methods, since they accept different types of objects. We can use `publish_custom_event_events` for sending both since it accepts `List[object]`. Can't use `publish_events` for both since it only accepts `List[EventGridEvent]`. The separate methods are generated by the swagger and needed for other languages.

swathipil (approved):
lgtm

sdk/eventgrid/azure-eventgrid/azure/eventgrid/_publisher_client.py
```diff
@@ -113,27 +117,24 @@ def send_events(self, events, **kwargs):
         if not isinstance(events, list):
             events = cast(ListEventType, [events])
```
yunhaoling, line 118: do we want to replace this cast/wrapper of non-list events with raising `ValueError` as in the description we say that we only accept list of events?

rakshith91: we still want to accept a single event - the purpose of removing from documentation is that we want to encourage users to use a list instead - there's no reason not to accept a single event otherwise

sdk/eventgrid/azure-eventgrid/azure/eventgrid/_publisher_client.py
```diff
@@ -113,27 +117,24 @@ def send_events(self, events, **kwargs):
...
 
-        if all(isinstance(e, CloudEvent) for e in events) or all(_is_cloud_event(e) for e in events):
+        if isinstance(events[0], CloudEvent) or _is_cloud_event(events[0]):
             try:
                 events = [cast(CloudEvent, e)._to_generated(**kwargs) for e in events] # pylint: disable=protected-access
             except AttributeError:
                 pass # means it's a dictionary
             kwargs.setdefault("content_type", "application/cloudevents-batch+json; charset=utf-8")
```
yunhaoling, line 125: non-blocking: we could consider constifying those strings if we use them across the sdk

rakshith91: hmm, we can, but this is the only method where we use it

yunhaoling: if there're only one or two places, I'm good with not constifying the string :)

yunhaoling (approved):

PR #16323, opened 2021-01-25
Support CAE in azure-identity

chlowell:
With this, user credentials configure MSAL with client capability "CP1", and accept an optional "claims" argument to get_token.

### DIFF
--- a/sdk/identity/azure-identity/azure/identity/_internal/interactive.py
+++ b/sdk/identity/azure-identity/azure/identity/_internal/interactive.py
@@ -103,8 +103,8 @@ class InteractiveCredential(MsalCredential):
         This method is called automatically by Azure SDK clients.
 
         :param str scopes: desired scopes for the access token. This method requires at least one scope.
-        :keyword str claims_challenge: a claims challenge returned by a resource provider following an authorization
-          failure
+        :keyword str claims: additional claims required in the token, such as those returned in a resource provider's
+          claims challenge following an authorization failure
         :rtype: :class:`azure.core.credentials.AccessToken`
         :raises CredentialUnavailableError: the credential is unable to attempt authentication because it lacks
           required data, state, or platform support
@@ -190,7 +190,7 @@ class InteractiveCredential(MsalCredential):
 
                 now = int(time.time())
                 result = app.acquire_token_silent_with_error(
-                    list(scopes), account=account, claims_challenge=kwargs.get("claims_challenge")
+                    list(scopes), account=account, claims_challenge=kwargs.get("claims")
                 )
                 if result and "access_token" in result and "expires_in" in result:
                     return AccessToken(result["access_token"], now + int(result["expires_in"]))
--- a/sdk/identity/azure-identity/azure/identity/_internal/interactive.py
+++ b/sdk/identity/azure-identity/azure/identity/_internal/interactive.py
@@ -103,6 +103,8 @@ class InteractiveCredential(MsalCredential):
         This method is called automatically by Azure SDK clients.
 
         :param str scopes: desired scopes for the access token. This method requires at least one scope.
+        :keyword str claims_challenge: a claims challenge returned by a resource provider following an authorization
+          failure
         :rtype: :class:`azure.core.credentials.AccessToken`
         :raises CredentialUnavailableError: the credential is unable to attempt authentication because it lacks
           required data, state, or platform support
@@ -187,7 +189,9 @@ class InteractiveCredential(MsalCredential):
                     continue
 
                 now = int(time.time())
-                result = app.acquire_token_silent_with_error(list(scopes), account=account, **kwargs)
+                result = app.acquire_token_silent_with_error(
+                    list(scopes), account=account, claims_challenge=kwargs.get("claims_challenge")
+                )
                 if result and "access_token" in result and "expires_in" in result:
                     return AccessToken(result["access_token"], now + int(result["expires_in"]))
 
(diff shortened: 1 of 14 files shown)


sdk/identity/azure-identity/tests/test_shared_cache_credential.py
```diff
@@ -800,6 +800,44 @@ def mock_send(request, **_):
...
+    """get_token should pass any claims challenge to MSAL token acquisition APIs"""
+
+    expected_claims = '{"access_token": {"essential": "true"}'
+
+    record = AuthenticationRecord("tenant-id", "client_id", "authority", "home_account_id", "username")
+
+    msal_app = Mock()
+    msal_app.get_accounts.return_value = [{"home_account_id": record.home_account_id}]
+    msal_app.acquire_token_silent_with_error.return_value = dict(
+        build_aad_response(access_token="**", id_token=build_id_token())
```
mccoyp, line 828: Does the `SharedTokenCacheCredential` not have an `id_token_claims` portion of the return value (like [this one](https://github.com/Azure/azure-sdk-for-python/pull/16323/files?file-filters%5B%5D=.py#diff-310b4541094eb834960847f01d92d60f5f9123fb62d181388e74d54ae6493d46R170)) because it's the only credential of these that doesn't subclass `InteractiveCredential`?

sdk/identity/azure-identity/tests/test_cae.py
```diff
@@ -0,0 +1,139 @@
...
+                "authority": os.environ.get("CAE_AUTHORITY", AzureAuthorityHosts.AZURE_PUBLIC_CLOUD),
+                "graph_url": os.environ.get("CAE_GRAPH_URL", "https://graph.microsoft.com"),
+                "password": os.environ.get("CAE_PASSWORD"),
+                "tenant_id": os.environ["CAE_TENANT_ID"],
+                "username": os.environ.get("CAE_USERNAME"),
+            }
+            real = urlparse(self.cae_settings["arm_url"])
+            self.scrubber.register_name_pair(real.netloc, "management.azure.com")
+            self.scrubber.register_name_pair(self.cae_settings["tenant_id"], "tenant")
+            self.scrubber.register_name_pair(self.cae_settings["username"], "username")
```
mccoyp, line 58:
```suggestion
            self.scrubber.register_name_pair(self.cae_settings["username"], "username")
            self.scrubber.register_name_pair(self.cae_settings["password"], "password")
```
Would this be necessary as well, or are passwords already hidden in recordings?

mccoyp (commented):
I see recordings for `test_device_code` and `test_username_password` from `test_cae.py`, but are there any other recordings that can be/need to be added?

sdk/identity/azure-identity/tests/test_shared_cache_credential.py
chlowell: Yes. That's relevant because MSAL token acquisition methods return a dict containing both the raw ID token and its claims. `InteractiveCredential` [uses these claims to build an `AuthenticationRecord`](https://github.com/Azure/azure-sdk-for-python/blob/79a64bb9d40610d1696819a2595a49d18d7a7ad6/sdk/identity/azure-identity/azure/identity/_internal/interactive.py#L51). `SharedTokenCacheCredential` does not, because it already has that information.

sdk/identity/azure-identity/tests/recording_processors.py
```diff
@@ -0,0 +1,77 @@
...
+    def process_response(self, response):
+        """Changes the "exp" claim of recorded id tokens to be in the future during playback
+
+        This is necessary because msal always validates id tokens, raising an exception when they've expired.
+        """
+        try:
+            # decode the recorded token
+            body = json.loads(six.ensure_str(response["body"]["string"]))
+            header, encoded_payload, signed = body["id_token"].split(".")
+            decoded_payload = base64.b64decode(encoded_payload + "=" * (4 - len(encoded_payload) % 4))
```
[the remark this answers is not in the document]
chlowell: A base64 encoded string should be padded with "=" to make its length divisible by 4. CPython's `base64` strictly requires padding. However, because a decoder can infer the padding, encoders commonly omit it.

sdk/identity/azure-identity/tests/test_cae.py
chlowell: Passwords are in request bodies, which aren't recorded.

chlowell:
I didn't record `InteractiveBrowserCredential` or `SharedTokenCacheCredential` because playback is complex for them (note the elaborate mocking in their tests). I think manual testing is acceptable for now, for this feature. I expect to revisit recorded tests in the future, to investigate covering more scenarios with them.

mccoyp (approved):

sdk/identity/azure-identity/azure/identity/_internal/interactive.py
```diff
@@ -187,7 +189,9 @@ def _acquire_token_silent(self, *scopes, **kwargs):
                     continue
 
                 now = int(time.time())
-                result = app.acquire_token_silent_with_error(list(scopes), account=account, **kwargs)
+                result = app.acquire_token_silent_with_error(
+                    list(scopes), account=account, claims_challenge=kwargs.get("claims")
+                )
```
jiasli, line 194: The removal of `**kwargs` will make acquiring SSH certificate stop working as `data` is passed in `kwargs`. 

Azure CLI is aiming to support acquiring SSH certificate for Service Principal as well (https://github.com/Azure/azure-sdk-for-python/pull/16397).

chlowell: Summarizing offline discussion here, there are two orthogonal concerns: acquiring SSH certificates, and allowing applications to pass keyword arguments through get_token to MSAL. Neither is supported today. Acquiring SSH certificates may be supported in a future version. Passing MSAL arguments through get_token is possible in some cases but really isn't supportable because routine maintenance requires internal changes (e.g. #16449) that can break applications relying on this behavior. My goal with this change is to prevent applications from taking a dependency on such implementation details.

PR #16326, opened 2021-01-25
Revise token cache configuration API

chlowell:
This PR replaces the current `enable_persistent_cache` and `allow_unencrypted_cache` keyword arguments to credentials with a new class, `TokenCachePersistenceOptions`, that allows the same configuration and enables an application to isolate its persistent cache from other applications.

```py
# configuring a credential to use the default persistent cache
cache_options = TokenCachePersistenceOptions()
credential = InteractiveBrowserCredential(cache_persistence_options=cache_options)

# specify a cache name to isolate the cache from other applications
TokenCachePersistenceOptions(name="my_application")

# configure the cache to fall back to unencrypted storage when encryption isn't available
TokenCachePersistenceOptions(allow_unencrypted_storage=True)
```


sdk/identity/azure-identity/samples/user_authentication.py
```diff
@@ -32,11 +34,13 @@
 client = SecretClient(VAULT_URL, credential)
 secret_names = [s.name for s in client.list_properties_of_secrets()]
 
-# With persistent caching enabled, an authentication record stored by your application enables
-# credentials to access data from past authentications. If the cache contains sufficient data,
-# this eliminates the need for your application to prompt for authentication every time it runs.
+# An authentication record stored by your application enables other credentials to access data from
+# past authentications. If the cache contains sufficient data, this eliminates the need for your
+# application to prompt for authentication every time it runs.
```
mccoyp, line 39: I would say that maybe we should mention that allowing other credentials to access this data is optional, except that's probably outside the scope of a short sample. Also, that information is covered in the code snippet for `TokenCachePersistenceOptions`, so I think that should be sufficient

mccoyp (approved):

schaabs (approved):

PR #16331, opened 2021-01-25
Mindep test all dependencies except nspkg

scbedd:
Instead of just azure packages.

Right now, the only one I know of that the previous excluded was `msrest`, but we'll see after a run of `python - all`!


scbedd:
Kicked a [full build](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=704312&view=results) based on this update. 

Just need to walk through the errors, file issues for minimum msrest against the couple management packages that have issues, and merge this.

praveenkuttappan (approved):
Nice. So this tests against min and max of external packages also.

chlowell (approved):

kristapratico (approved):

PR #16332, opened 2021-01-25
Resolve Eventhub Import Failure

scbedd:
azure-eventhub-checkpointstore is failing `depends` check because the following file can't be loaded by Python:

`D:\\vss-agent-2.181.0\\_work\\1\\s\\sdk\\eventhub\\azure-eventhub-checkpointstoreblob\\.tox\\depends\\Lib\\site-packages\\azure\\eventhub\\extensions\\checkpointstoreblob\\_vendor\\storage\\blob\\_generated\\aio\\operations_async\\__pycache__\\_append_blob_operations_async.cpython-35.pyc'`

This is obviously a longpath issue on windows, given that linux and mac both check out.


weshaggard:
@mitchdenny this looks like it might be caused by an increase in the root path with the switch to the 1ES agent pools do we know if there is a way to specify that in the pool configurations?

@scbedd I'm not a fan of enabling the longpath support in CI as it just sets up trouble for contributors doing work in the repo on windows. Is there any way we can figure out how to shorten the path?

scbedd:
@weshaggard I'm with ya. It's not really an option anyway, given that setting of the environment variable doesn't actually resolve the error that the tox environment is hitting.

rakshith91 (approved):

mitchdenny:
Incoming fix from the 1ES team may help to resolve this.

scbedd:
Not the solution. Trying something else. Will re-open if if I go this route again.

PR #16365, opened 2021-01-26
Move mindependency and latestdependency common dependencies to a file

scbedd:
- Process that file as part of CI to ensure we re-use the same wheel (instead of hitting issues when multiple tox environments try to build it at once)
- In `mindependency` or `latestdependency` we only use use default tool install + your package reqs. NO dev_requirements are used.

This is the piece that I forgot when merging the [the other PR](https://github.com/Azure/azure-sdk-for-python/pull/16313). We need it to work both with a default `tox` install, but also in CI where parallelism comes into play.

This PR should meet the needs of both.

### DIFF
--- a/eng/tox/tox.ini
+++ b/eng/tox/tox.ini
@@ -214,6 +214,9 @@ commands =
 
 [testenv:mindependency]
 deps = 
+  azure-mgmt-keyvault<7.0.0
+  azure-mgmt-resource<15.0.0
+  azure-mgmt-storage<15.0.0
   {[dependencytools]deps}
   {[tools]deps}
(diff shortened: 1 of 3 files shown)


scbedd:
Confirmed on internal run.

image

praveenkuttappan (approved):

eng/tox/tox.ini
```diff
@@ -179,8 +178,6 @@ commands =
 
 
 [testenv:devtest]
-pre-deps =
```
scbedd, line 182: pre-deps literally isn't a thing. Just cleaning it up.

PR #16384, opened 2021-01-27
CertificateCredential supports PKCS12 certs

### LINKED ISSUE #13540: Identity - support for client certificates in formats other than PEM
When users wish to use the Identity SDK for Python in conjunction with Azure Batch, they need to implement extra logic to convert their certificate from .pfx/.cer to .pem. Can we add logic so that the Identity SDK supports other certificate formats innately? 

Azure Batch documentation on certificates: https://docs.microsoft.com/en-us/azure/batch/batch-certificates
Suspected code where the current restriction for PEM-only certificates is implemented: https://github.com/Azure/azure-sdk-for-python/blob/eea845a56aa792f5b03b9188cd0180d42a3de4bb/sdk/identity/azure-identity/azure/identity/_internal/aadclient_certificate.py#L19
Python SDK reference describing the PEM-only restriction: https://azuresdkdocs.blob.core.windows.net/$web/python/azure-identity/1.4.0/index.html#environment-variables

### LINKED ISSUE #14055: SDK should support reading certificate from buffers instead of file paths.
Using Azure SDK to authenticate via certificateCredential requires parameter **filePath to pem cert including private key**. This is not desirable in an environment where it is not safe to have pem certs that include private key.

If the user has pem bytes in the variable, he should be able to pass it and authenticate. 

One use case is Azure Databricks Environment. User does not want to store the certificate inside a file on a cluster. Instead he can get the certificate from mounted keyvault and use it for authentication. 

Because the code is just reading pem bytes from file path, let it also consume pem bytes from user if the user has it.

https://github.com/Azure/azure-sdk-for-python/blob/e2cac0341ebb1d1d56c9e4535cae3f9c107154c4/sdk/identity/azure-identity/azure/identity/_credentials/certificate.py#L52

### LINKED ISSUE #14854: Address Client Certificate Credential Consistency Public API Review feedback
Following changes need to be done, based on the review discussion:

- [ ] Add support for PFX certs (#13540)

chlowell:
Closes #13540


sdk/identity/azure-identity/tests/test_live.py
```diff
@@ -24,23 +24,29 @@ def get_token(credential):
...
-        live_certificate_with_password["client_id"],
-        live_certificate_with_password["cert_path"],
-        password=live_certificate_with_password["password"],
+        tenant_id, client_id, certificate_bytes=cert["cert_with_password_bytes"], password=cert["password"],
     )
     get_token(credential)
 
 
+@pytest.mark.skip("...")
 def test_client_secret_credential(live_service_principal):
```
mccoyp, line 50: Why are this and `test_default_credential` being skipped?

chlowell:
I split the cert-as-bytes part of this into #16410, will refocus this PR on PKCS12 support.

sdk/identity/azure-identity/tests/test_live.py
chlowell: To simplify my debugging. Definitely not something I want to merge, thanks for pointing it out!

sdk/identity/azure-identity/azure/identity/_credentials/certificate.py
```diff
@@ -84,6 +62,67 @@ def extract_cert_chain(pem_bytes):
...
+    private_key, cert, additional_certs = pkcs12.load_key_and_certificates(
+        certificate_bytes, password=password, backend=default_backend()
+    )
+
+    if password:
+        encryption = BestAvailableEncryption(password)  # type: KeySerializationEncryption
+    else:
+        encryption = NoEncryption()
+
+    key_bytes = private_key.private_bytes(Encoding.PEM, PrivateFormat.TraditionalOpenSSL, encryption)
```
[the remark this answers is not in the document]
chlowell: No. So far as I can tell we have no reason to prefer either, so I'll follow cryptography's preference for PKCS8.

sdk/identity/azure-identity/azure/identity/_credentials/certificate.py
```diff
@@ -76,6 +77,42 @@ def extract_cert_chain(pem_bytes):
...
+    key_bytes = private_key.private_bytes(Encoding.PEM, PrivateFormat.PKCS8, NoEncryption())
+    pem_sections = [key_bytes] + [c.public_bytes(Encoding.PEM) for c in [cert] + additional_certs]
+    pem_bytes = b"".join(pem_sections)
+
+    fingerprint = cert.fingerprint(hashes.SHA1())  # nosec
+
+    return _Cert(pem_bytes, private_key, fingerprint)
+
+
 def get_client_credential(certificate_path, password=None, certificate_data=None, send_certificate_chain=False, **_):
```
xiangyan99, line 116: Seems like certificate_path is not required. Why not give it a default value?

chlowell: Either certificate_path or certificate_data is required. `CertificateCredential.__init__` will always have a value for certificate_path because it was originally a required parameter.

xiangyan99: Then if user wants to use certificate_data, they have to explicitly set certificate_path to None?

My concern (though not very strong) is if we imply user must set the value of certificate_path and if certificate_path is not None, setting certificate_data raises ValueError. It will cause confusion if user wants to use certificate_data to get_client_credential?

chlowell: This (get_client_credential) is an internal helper method called only by `CertificateCredential.__init__`, so I doubt it implies anything to users. This is the public API:
```py
# load a cert from disk
CertificateCredential(tenant_id, client_id, "/cert.pem")

# load a cert from bytes
CertificateCredential(tenant_id, client_id, certificate_data=some_bytes)
```

xiangyan99: Get it. Then never mind.

Thanks for the clarification.

sdk/identity/azure-identity/tests/test_live.py
```diff
@@ -24,26 +23,24 @@ def get_token(credential):
     assert token.expires_on
 
 
-def test_certificate_credential(live_certificate):
-    tenant_id = live_certificate["tenant_id"]
-    client_id = live_certificate["client_id"]
+@pytest.mark.parametrize("certificate_fixture", ("live_pem_certificate", "live_pfx_certificate"))
+def test_certificate_credential(certificate_fixture, request):
```
mccoyp, line 27: What's `request` here?

chlowell: One of pytest's built-in fixtures: https://docs.pytest.org/en/6.2.x/reference.html?highlight=fixture#request

sdk/identity/azure-identity/azure/identity/_credentials/certificate.py
```diff
@@ -76,6 +77,42 @@ def extract_cert_chain(pem_bytes):
...
+        raise ValueError("Failed to deserialize certificate in PEM or PKCS12 format")
+
+    # This serializes the private key without any encryption it may have had. Doing so doesn't violate security
+    # boundaries because this representation of the key is kept in memory. We already have the key and its
+    # password, if any, in memory.
+    key_bytes = private_key.private_bytes(Encoding.PEM, PrivateFormat.PKCS8, NoEncryption())
+    pem_sections = [key_bytes] + [c.public_bytes(Encoding.PEM) for c in [cert] + additional_certs]
+    pem_bytes = b"".join(pem_sections)
+
+    fingerprint = cert.fingerprint(hashes.SHA1())  # nosec
```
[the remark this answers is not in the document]
chlowell: The tool is bandit. Its warnings are rare enough that I've never run it locally. It runs as a stage of the Analyze job, so I expect a bandit warning to fail a CI run. You suppress the warning when you really need to do something bandit considers insecure and have confirmed there is no danger. For example, the issue here is that a collision attack against SHA1 is feasible. In this context that means an attacker could create a forged cert having the same fingerprint as a legitimate one. That's okay because AAD has a copy of the legitimate cert and uses the fingerprint only to identify that copy. An attacker can't authenticate with such a forged cert because its signatures won't verify against the legitimate cert.

mccoyp (approved):

xiangyan99 (approved):

DhavalShewale:
@chlowell By when can we expect support for PKCS12 certs to be available in the azure-identity library

chlowell:
This change will be in 1.7.0b4, which we plan to release September 8

DhavalShewale:
@chlowell Tks for the quick revert. I presume 1.7.0b4 will be a beta release.

Do we have clarity on when a stable release for version 1.7.0 will be available?

chlowell:
October at the earliest. If you need something in the meantime, it's not too difficult to convert PKCS12 to PEM with cryptography ([for example](https://github.com/Azure/azure-sdk-for-python/blob/1f0fdea5cb745ca916a8a60b817d4fe6dfdb6fb3/sdk/identity/azure-identity/samples/key_vault_cert.py#L60)).

PR #16397, opened 2021-01-28
[Identity] Let GetTokenMixin.get_token pass kwargs to MSAL

jiasli:
`GetTokenMixin` is inherited by 

- `ClientSecretCredential`
- `CertificateCredential`
- `ManagedIdentityCredential` 

However, `GetTokenMixin.get_token` discards `**kwargs` when calling MSAL, making it impossible to get an SSH certificate with a Service Principal credential.

After applying this PR, the network trace shows `req_cnf` is sent correctly:

```
azure.core.pipeline.policies._universal: Request URL: 'https://login.microsoftonline.com/54826b22-38d6-4fb2-bad9-b7b93a3e9c5a/oauth2/v2.0/token'
azure.core.pipeline.policies._universal: Request method: 'POST'
...
azure.core.pipeline.policies._universal: Request body:
azure.core.pipeline.policies._universal: 
{
    'client_id': 'ee01ddbe-2865-4d12-95ef-cd75ac9285a8', 
    'grant_type': 'client_credentials', 
    'client_info': 1, 
    'client_secret': 'UKx11...', 
    'token_type': 'ssh-cert', 
    'req_cnf': '{
        "kty": "RSA", 
        "n": "...", 
        "e": "AQAB", 
        "kid": "a83576e..."
    }', 
    'key_id': 'a83576e2adc1a2654003f3744486fa1f2595851e5539c6d213011854be1377fb', 
    'scope': 'https://pas.windows.net/CheckMyAccess/Linux/.default'
}
```

### DIFF
--- a/sdk/identity/azure-identity/azure/identity/_internal/managed_identity_client.py
+++ b/sdk/identity/azure-identity/azure/identity/_internal/managed_identity_client.py
@@ -84,8 +84,8 @@ class ManagedIdentityClientBase(ABC):
 
         return token
 
-    def get_cached_token(self, *scopes):
-        # type: (*str) -> Optional[AccessToken]
+    def get_cached_token(self, *scopes, **kwargs):   # pylint:disable=unused-argument
+        # type: (*str, **Any) -> Optional[AccessToken]
         resource = _scopes_to_resource(*scopes)
         tokens = self._cache.find(TokenCache.CredentialType.ACCESS_TOKEN, target=[resource])
         for token in tokens:
--- a/sdk/identity/azure-identity/azure/identity/_credentials/app_service.py
+++ b/sdk/identity/azure-identity/azure/identity/_credentials/app_service.py
@@ -38,9 +38,9 @@ class AppServiceCredential(GetTokenMixin):
             )
         return super(AppServiceCredential, self).get_token(*scopes, **kwargs)
 
-    def _acquire_token_silently(self, *scopes):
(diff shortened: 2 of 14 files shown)


sdk/identity/azure-identity/azure/identity/_internal/managed_identity_client.py
```diff
@@ -84,8 +84,8 @@ def _process_response(self, response, request_time):
 
         return token
 
-    def get_cached_token(self, *scopes):
-        # type: (*str) -> Optional[AccessToken]
+    def get_cached_token(self, *scopes, **kwargs):   # pylint:disable=unused-argument
```
jiasli, line 87: I think this is more complicated than the change itself:

- Service Principal credentials support `data`, but 
- Managed Identity credentials don't. (They may support `data` in the future.)

Making them share the same `GetTokenMixin` adds unused `kwargs` to `ManagedIdentityClientBase.get_cached_token`.

chlowell:
Whether a credential uses MSAL, and how, are implementation details. Routine maintenance makes it impossible to guarantee two versions of azure-identity will call the same MSAL methods, or require the same version of MSAL (see e.g. #16449). An application trying to pass arguments to an MSAL method through `get_token` therefore risks breaking every time it upgrades azure-identity. I don't think we should help applications take that risk.

Whether we support acquiring SSH certificates is a separate question. We'll consider doing that in a future version.

Issue #14394, opened 2020-10-09
New Analyze Check


scbedd:
Per #13903 

> Having some files using CRLF makes things extremely messy when
using tarballs on Linux, as the git repository gets automatically
converted, but the tarball stores the original line ending,
so the two do not match anymore.

Add a step in the `Analyze` Step that runs `dos2unix` after all the others. This should ensure that no more CRLF readmes make it into our repo.

PR #16398, opened 2021-01-28
adding step to test for crlf line endings

seankane-msft:
#14394

### DIFF
--- a/eng/pipelines/templates/steps/analyze.yml
+++ b/eng/pipelines/templates/steps/analyze.yml
@@ -30,8 +30,7 @@ steps:
       Get-ChildItem $(Build.SourcesDirectory) -Filter "*.py" |
       Foreach-Object {
         If ((Get-Content $_ -Raw) -match "\r\n") {
-          Write-Ouput $_
-          Throw "Found a CRLF line ending"
+          Write-Error "Found CRLF line ending in `$_`"
         }
       }
 
--- a/eng/pipelines/templates/steps/analyze.yml
+++ b/eng/pipelines/templates/steps/analyze.yml
@@ -28,7 +28,7 @@ steps:
 
   - script: |
       find ./ -type f -name "*.py" -o -name "*.md" -print0 | while IFS= read -r -d '' file; do
-          [[ $(file $file) =~ "CRLF" ]] && printf 'CRLF line ending in %s\n' "$file"
+          [[ $(file $file) =~ "CRLF" ]] && echo -n "CRLF line ending in: " && echo "$file"
       done
     displayName: Check for CRLF Line endings
 
--- a/eng/pipelines/templates/steps/analyze.yml
+++ b/eng/pipelines/templates/steps/analyze.yml
@@ -29,11 +29,11 @@ steps:
   - pwsh: |
       Get-ChildItem .\ -Filter "*.py" |
       Foreach-Object {
-        (Get-Content $_ -Raw) -match "\r\n$" && Write-Output $_
+        (Get-Content $_ -Raw) -match "\r\n" && Write-Output $_
       }
       Get-ChildItem .\ -Filter "*.md" |
       Foreach-Object {
-        (Get-Content $_ -Raw) -match "\r\n$" && Write-Output $_
+        (Get-Content $_ -Raw) -match "\r\n" && Write-Output $_
       }
     displayName: Check for CRLF Line endings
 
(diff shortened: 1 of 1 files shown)


seankane-msft:
\azp run python - appconfiguration

eng/pipelines/templates/steps/analyze.yml
```diff
@@ -26,6 +26,17 @@ steps:
       python find_invalid_versions.py --always-succeed --service=${{parameters.ServiceDirectory}}
     displayName: Find Invalid Versions
 
+  - pwsh: |
+      Get-ChildItem .\ -Filter "*.py" |
+      Foreach-Object {
+        (Get-Content $_ -Raw) -match "\r\n$" && Write-Output $_
```
weshaggard, line 32: Does this work? It looks like it might only check the end of the file for crlf, as I believe Raw will cause you to return the full file contents. If you want to check for it everywhere I think you want to remove the `$` from the regex.

eng/pipelines/templates/steps/analyze.yml
```diff
@@ -26,6 +26,17 @@ steps:
...
+      }
+      Get-ChildItem .\ -Filter "*.md" |
+      Foreach-Object {
+        (Get-Content $_ -Raw) -match "\r\n$" && Write-Output $_
+      }
+    displayName: Check for CRLF Line endings
```
weshaggard, line 38: Did you want this step to fail if you find any files with CRLF?

seankane-msft: Yes @weshaggard

seankane-msft: I added a `Throw` chaining to these two statements

weshaggard: You can just call exit 1 as well and that will cause the step to fail.

eng/pipelines/templates/steps/analyze.yml
```diff
@@ -26,6 +26,24 @@ steps:
       python find_invalid_versions.py --always-succeed --service=${{parameters.ServiceDirectory}}
     displayName: Find Invalid Versions
 
+  - pwsh: |
+      Get-ChildItem .\ -Filter "*.py" |
```
weshaggard, line 30: You should either pass in the base directly instead of relying on the current directory which could be different depending on other steps. Or you can explicitly pass the workingdirectory as an option to this step.

eng/pipelines/templates/steps/analyze.yml
```diff
@@ -26,6 +26,24 @@ steps:
       python find_invalid_versions.py --always-succeed --service=${{parameters.ServiceDirectory}}
     displayName: Find Invalid Versions
 
+  - pwsh: |
+      Get-ChildItem $(Build.SourcesDirectory) -Filter "*.py" |
+      Foreach-Object {
+        If ((Get-Content $_ -Raw) -match "\r\n") {
+          Write-Ouput $_
+          Throw "Found a CRLF line ending"
```
weshaggard, line 34:
```suggestion
          Write-Error "Found CRLF line ending in `$_`"
```

eng/pipelines/templates/steps/analyze.yml
```diff
@@ -26,6 +26,24 @@ steps:
...
+        }
+      }
+
+      Get-ChildItem $(Build.SourcesDirectory) -Filter "*.md" |
+      Foreach-Object {
+        If ((Get-Content $_ -Raw) -match "\r\n") {
+          Write-Ouput $_
+          Throw "Found a CRLF line ending"
```
weshaggard, line 42:
```suggestion
        If ((Get-Content $_ -Raw) -match "\r\n") {
          Write-Error "Found CRLF line ending in `$_`"
```

eng/pipelines/templates/steps/analyze.yml
```diff
@@ -26,6 +26,24 @@ steps:
       python find_invalid_versions.py --always-succeed --service=${{parameters.ServiceDirectory}}
     displayName: Find Invalid Versions
 
+  - pwsh: |
+      Get-ChildItem $(Build.SourcesDirectory) -Filter "*.py" |
+      Foreach-Object {
+        If ((Get-Content $_ -Raw) -match "\r\n") {
```
weshaggard, line 32:
```suggestion
        if ((Get-Content $_ -Raw) -match "\r\n") {
```

seankane-msft:
\azp run python - appconfiguration

weshaggard (approved):

PR #16415, opened 2021-01-28
Regenerate baseline because last one break.

sima-zhu:
Regenerate baseline: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=708546&view=results


eng/python.gdnbaselines
```diff
@@ -1,5613 +1,5730 @@
...
-      "name": "baseline",
-      "createdDate": "2021-01-15 22:07:40Z",
-      "lastUpdatedDate": "2021-01-15 22:07:40Z"
+    "demobaseline": {
+      "name": "demobaseline",
+      "createdDate": "2021-01-29 00:29:11Z",
+      "lastUpdatedDate": "2021-01-29 00:29:11Z"
     }
   },
   "results": {
```
weshaggard, line 10: Can we wait and update these baseline files once we get better suppression rules? I don't want these baseline files to constantly need updated based on any change and I'd like them to be much smaller if we get good placeholder values.

sima-zhu: We changed this because guardian team introduced some breaking changes which they do not recognize the errors in baseline file. 

The baseline file does not affect how we do the suppression. It is better to merge the baseline file first, and  come up with the suppression strategy in the meanwhile

eng/pipelines/aggregate-reports.yml
```diff
@@ -53,8 +34,32 @@ jobs:
       pwsh: true
       workingDirectory: $(Build.SourcesDirectory)
       filePath: eng/common/scripts/Verify-Resource-Ref.ps1
+      
+  - template: /eng/pipelines/templates/steps/analyze_dependency.yml
```
weshaggard, line 38: What was this order changed? I think this needs to be moved back.

eng/python.gdnbaselines
weshaggard: My expectation is that this baseline file will be much smaller once we do the suppressions. If we plan to update it then that is fine but I want to be careful that we don't have to have lots of folks updating this baseline file all the time.

sima-zhu: Currently, we have two purposes for putting baseline file here:
1. Leave time for owner to address the existing errors.
2. Catch new issue introduced to packages.

I think #2 is more crucial for our cred scan steps right now. The baseline file help us achieve the purpose, so it is necessary to check in now. 
As for #1, I plan to do some suppression on some obvious fake ones. Then we indeed need help from language owner.

eng/pipelines/aggregate-reports.yml
sima-zhu: Do you mean analyze_dependency should go after run cred scan step?

weshaggard (approved):

eng/pipelines/aggregate-reports.yml
```diff
@@ -54,7 +37,29 @@ jobs:
...
+      GdnBreakGdnToolCredScan: true
+      GdnBreakGdnToolCredScanSeverity: Error
+      GdnBreakBaselineFiles: $(Build.SourcesDirectory)\eng\python.gdnbaselines
+      GdnBreakBaselines: baseline   
+      # Used for generating baseline file.   
+      # GdnBreakOutputBaselineFile: python
+      # GdnBreakOutputBaseline: baseline
+  - task: securedevelopmentteam.vss-secure-development-tools.build-task-publishsecurityanalysislogs.PublishSecurityAnalysisLogs@3
+    displayName: 'Publish Security Analysis Logs'
+    continueOnError: true
```
weshaggard, line 59: We should remove this now that the baseline is updated.

Issue #14565, opened 2020-10-16
ResourceWarning: unclosed <ssl.SSLSocket fd=20, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=6, laddr=('10.1.0.62', 52098), raddr=('52.239.140.10', 443)>


lukin0110:
- **azure-storage-blob**: 
- **12.5.0**: 
- **Linux 39a68fd14e65 4.19.76-linuxkit #1 SMP Tue May 26 11:42:35 UTC 2020 x86_64 GNU/Linux**:
- **Python 3.8.3**: 

**Describe the bug**
We've enabled a few warning filters in our application with `warning.simplefilter` and we're seeing a python warning when we download a file from the blob storage. `ResourceWarning: unclosed <ssl.SSLSocket fd=20, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=6, laddr=('10.1.0.62', 52098), raddr=('52.239.140.10', 443)>`

**To Reproduce**
```
client = BlobServiceClient.from_connection_string("<connection string here>").get_blob_client(container="<container name>", blob="<blob name>")
raw_data = client.download_blob().readall()
with gzip.open(BytesIO(raw_data), "rb") as fh:
    data = pickle.load(fh)  # noqa: S301
```

**Expected behavior**
No warning or a proper way to close the stream.

**Screenshots**
NA

**Additional context**
NA

xiafu-msft:
Hi @lukin0110 

So you also saw this warning when only executing this two lines?
```
client = BlobServiceClient.from_connection_string("<connection string here>").get_blob_client(container="<container name>", blob="<blob name>")
raw_data = client.download_blob().readall()
```

lukin0110:
Hi @xiafu-msft 

The correct way to reproduce is:
```
import warnings
from azure.storage.blob import BlobServiceClient

warnings.simplefilter("default", ResourceWarning)

CONNECTION = "<connection string here>"
CONTAINER = "<container name here>"
BLOB_NAME = "<blob name here>"

client = BlobServiceClient.from_connection_string(CONNECTION).get_blob_client(container=CONTAINER, blob=BLOB_NAME)
raw_data = client.download_blob().readall()
```

xiafu-msft:
Hi @lukin0110 

Thanks for your code!
can you try this and see if you still get this warning:
```
[...]

PR #16428, opened 2021-01-29
[Bug][Blob][Datalake][Queue][Fileshare]Fixed socket not closing after session closure across parent clients

tasherif-msft:
This resolves: https://github.com/Azure/azure-sdk-for-python/issues/14565
Currently sockets are not being closed after context manager ends session or `close()` is called. This is happening because `TransportLayer` is used which does not have a proper `close()`  (currently just passes) and even if we added `self._transport.close()` it would beat the purpose of the class since this would close parent sessions as well causing them to break (sharing the same transport object/session). 

For example:
```
        with BlobServiceClient(self.account_url(storage_account, "blob"), credential=storage_account_key) as bsc:
            with bsc.get_blob_client("test", "blob.txt") as bc:
                bc.download_blob().readall()
            with bsc.get_blob_client("test", "blob.txt") as bc1:  #--------> fails here
                bc1.download_blob().readall()
```
Would fail since bsc session would also be closed/set to None.
Instead of using TransportLayer, do a `deepcopy`

### DIFF
--- a/sdk/storage/azure-storage-blob/azure/storage/blob/_blob_service_client.py
+++ b/sdk/storage/azure-storage-blob/azure/storage/blob/_blob_service_client.py
@@ -3,8 +3,8 @@
 # Licensed under the MIT License. See License.txt in the project root for
 # license information.
 # --------------------------------------------------------------------------
-
 import functools
+from copy import deepcopy
 from typing import (  # pylint: disable=unused-import
     Union, Optional, Any, Iterable, Dict, List,
     TYPE_CHECKING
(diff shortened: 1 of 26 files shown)


abhishek12tri:
Please Try:
`blobServClient = BlobServiceClient.from_connection_string({{CONNECTION}})`
`blobClient = blobServClient.get_blob_client(container="test", blob="{{BLOB_FILE}}")`
`raw_files = blobClient.download_blob().readall()`
`blobServClient.close()`

Need to close the socket after use

Issue #16433, opened 2021-01-30
[EG arch feedback] Task list


rakshith91:
- [ ] Kill Custom Events
- [ ] Kill Deserializer
- [ ] Cloud Event make data optional
- [ ] Support Data / data_base64 (not both)data must be optional ; data_base64 is optional too data can be bytes(convert to data_base64) data and data_base64 at the same time raises.
- [ ] SystemEventMappings to SystemEvents
- [ ] change order of EG event

PR #16441, opened 2021-02-01
Arch preview feedback

rakshith91:
Fixes https://github.com/Azure/azure-sdk-for-python/issues/16433

-  Kill Custom Events
-  Kill Deserializer
-  Cloud Event make data optional
-  Support Data / data_base64 (not both)data must be optional ; data_base64 is optional too data can be bytes(convert to data_base64) data and data_base64 at the same time raises.
-  SystemEventMappings to SystemEvents
-  change order of EG event


sdk/eventgrid/azure-eventgrid/README.md
```diff
@@ -60,10 +60,7 @@ Information about the key concepts on Event Grid, see [Concepts in Azure Event G
 
 ### EventGridPublisherClient
 `EventGridPublisherClient` provides operations to send event data to topic hostname specified during client initialization.
-Either a list or a single instance of CloudEvent/EventGridEvent/CustomEvent can be sent.
-
-### EventGridDeserializer
-`EventGridDeserializer` is used to desrialize an event received.
+Either a list or a single instance of CloudEvent/EventGridEvent can be sent.liAlternatively, a lis or a single instance of the dict representation of CloudEvent/EventGridEvent or a Custom Schema can also be published.
```
yunhaoling, line 63:
```suggestion
Either a list or a single instance of CloudEvent/EventGridEvent can be sent. Alternatively, a list or a single instance of the dict representation of CloudEvent/EventGridEvent or a Custom Schema can also be published.
```

rakshith91: but how? :O

sdk/eventgrid/azure-eventgrid/azure/eventgrid/_models.py
```diff
@@ -36,13 +35,16 @@ class CloudEvent(EventMixin):   #pylint:disable=too-many-instance-attributes
     """Properties of an event published to an Event Grid topic using the CloudEvent 1.0 Schema.
 
     All required parameters must be populated in order to send to Azure.
+    If data is of binary type, data_base64 can be used alternatively. Note that data and data_base64
+    cannot be present at the same time.
```
yunhaoling, line 39: may I ask when the user should use `data_base64` and when should the user just use `data`?
it looks like `data` is enough to handle bytes.

rakshith91: yes - data is enough to handle bytes - when user passes bytes using data, we internally send data_base64 instead - something like this
 ```
if isinstance(self.data, six.binary_type):
            data_base64 = self.data
            data = None
```

but, we should not stop users from sending data_base64 directly as a keyword instead of data

File: sdk/eventgrid/azure-eventgrid/azure/eventgrid/_models.py [hunk 3]
```diff
@@ -80,19 +88,23 @@ class CloudEvent(EventMixin):   #pylint:disable=too-many-instance-attributes
...
+        self.data = kwargs.pop("data", None)
         self.datacontenttype = kwargs.pop("datacontenttype", None)
         self.dataschema = kwargs.pop("dataschema", None)
         self.subject = kwargs.pop("subject", None)
+        self.data_base64 = kwargs.pop("data_base64", None)
         self.extensions = {}
         self.extensions.update(dict(kwargs.pop('extensions', {})))
+        if self.data is not None and self.data_base64 is not None:
+            raise ValueError("data and data_base64 cannot be provided at the same time.\
+                Use data_base64 only if you are sending bytes, and use data otherwise.")
```
yunhaoling, line 107: do we need to check if both of them are None?
```python
if not self.data and not self.data_base64:
    raiseValueError("...")
```

yunhaoling: any do we need to check if `data_base64` is of bytes type?

sdk/eventgrid/azure-eventgrid/azure/eventgrid/_models.py
```diff
@@ -211,8 +218,8 @@ class EventGridEvent(InternalEventGridEvent, EventMixin):
         'data_version': {'key': 'dataVersion', 'type': 'str'},
     }
 
-    def __init__(self, data, subject, event_type, data_version, **kwargs):
-        # type: (object, str, str, str, Any) -> None
+    def __init__(self, subject, event_type, data, data_version, **kwargs):
```
yunhaoling, line 221: I guess this is also a breaking change worth to note in the changelog as the position of positional argument changes.

sdk/eventgrid/azure-eventgrid/tests/test_eg_publisher_client.py
```diff
@@ -113,6 +114,51 @@ def test_send_cloud_event_data_dict(self, resource_group, eventgrid_topic, event
...
+        cloud_event = CloudEvent(
+                source = "http://samplesource.dev",
+                data = b'cloudevent',
+                type="Sample.Cloud.Event"
+                )
+
+        def callback(request):
+            req = json.loads(request.http_request.body)
+            assert req[0].get("data_base64") is not None
+            assert req[0].get("data") is None
```
yunhaoling, line 131: now I'm getting a bit confused by this request body.. because you set the `data` in the `CloudEvent` while the http request body `get("data")` is None,

so if I pass bytes to the `data` argument, which will internally set the `data_base64` variable?
```
cloud_event = CloudEvent(
    data = b'cloudevent',
)
# which of the following will have the value set?
cloud_event.data 
cloud_event.data_base64
```

or it's just implementation detail that users don't need to care about.

sdk/eventgrid/azure-eventgrid/azure/eventgrid/aio/_publisher_client_async.py
```diff
@@ -60,6 +57,8 @@ class EventGridPublisherClient():
     :param credential: The credential object used for authentication which implements
      SAS key authentication or SAS token authentication.
     :type credential: ~azure.core.credentials.AzureKeyCredential or ~azure.core.credentials.AzureSasCredential
+    :rtype: None
+    :raises: :class:`ValueError`, when events do not follow specified SendType.
```
yunhaoling, line 61: is this part miss placed? looks like it's related to the `send` methods

sdk/eventgrid/azure-eventgrid/samples/champion_scenarios.md
```diff
@@ -0,0 +1,333 @@
...
+topic_key = os.environ["EG_ACCESS_KEY"]
+endpoint = os.environ["EG_TOPIC_HOSTNAME"]
+
+credential = AzureKeyCredential(topic_key)
+client = EventGridPublisherClient(endpoint, credential)
+```
+
+## Send Scenarios
+
+### Send a Single EventGridEvent as a strongly typed object
```
yunhaoling, line 40: nit:
```suggestion
### Send a single EventGridEvent as a strongly typed object
```

sdk/eventgrid/azure-eventgrid/samples/champion_scenarios.md
```diff
@@ -0,0 +1,333 @@
...
+            "itemSku": "Contoso Item SKU #1"
+        },
+        subject="Door1",
+        data_version="2.0"
+    )
+
+client.send(event)
+```
+
+### Send a Single CloudEvent as a strongly typed object
```
yunhaoling, line 65: nit:
```suggestion
### Send a single CloudEvent as a strongly typed object
```

sdk/eventgrid/azure-eventgrid/samples/champion_scenarios.md
```diff
@@ -0,0 +1,333 @@
...
+## Receive Scenarios
+
+### Deserialize EventGridEvents from storage queue
+
+```Python
+from azure.eventgrid import CloudEvent
+from azure.storage.queue import QueueServiceClient, BinaryBase64DecodePolicy
+import os
+import json
+from base64 import b64decode
```
yunhaoling, line 267: emmm, is `b64decode` used?

sdk/eventgrid/azure-eventgrid/azure/eventgrid/aio/_publisher_client_async.py
rakshith91, line 61:
```suggestion
```

sdk/eventgrid/azure-eventgrid/azure/eventgrid/_models.py
yunhaoling: is there any reason why we want to add `data_base64` argument to emphasize the `base64` encoded bytes?
I still feel that if we're being smart to handle all the input data type, then this `data_base64` seems redundant.

sdk/eventgrid/azure-eventgrid/azure/eventgrid/_models.py
rakshith91: yes - data can totally be an empty dict for instance - so we need to check if both are not none explicitly.

for second part, not necessarily, we will eventually get a good 
`TypeError: a bytes-like object is required, not 'str'`

sdk/eventgrid/azure-eventgrid/azure/eventgrid/_models.py
rakshith91: well, this is just reverted back to the original form :( 
i had a pr before that changed this, but we decided not to anyway

sdk/eventgrid/azure-eventgrid/azure/eventgrid/aio/_publisher_client_async.py
rakshith91: yes - fixing it

sdk/eventgrid/azure-eventgrid/samples/champion_scenarios.md
rakshith91, line 267:
```suggestion
```

rakshith91: removing this

sdk/eventgrid/azure-eventgrid/tests/test_eg_publisher_client.py
rakshith91: internally, `cloud_event.data_base64` is set when we we pass bytes regardless of what we use in `data`, `data_base64` to send bytes.

sdk/eventgrid/azure-eventgrid/azure/eventgrid/_models.py
rakshith91: we want a data_base64 since https://github.com/cloudevents/spec says so :)

sdk/eventgrid/azure-eventgrid/README.md
```diff
@@ -60,10 +60,7 @@ Information about the key concepts on Event Grid, see [Concepts in Azure Event G
 
 ### EventGridPublisherClient
 `EventGridPublisherClient` provides operations to send event data to topic hostname specified during client initialization.
-Either a list or a single instance of CloudEvent/EventGridEvent/CustomEvent can be sent.
-
-### EventGridDeserializer
-`EventGridDeserializer` is used to desrialize an event received.
+Either a list or a single instance of CloudEvent/EventGridEvent can be sent. Alternatively, a list or a single instance of the dict representation of CloudEvent/EventGridEvent or a Custom Schema can also be published.
```
swathipil, line 63: The wording is a little confusing. Might be more clear to separate the idea that CloudEvent/EGEvents can be sent as dicts and that events in a custom schema can be sent as a dict or list of dicts. Would be useful to add links for either the samples of sending all these different formats or from the `champion_scenarios.md` . Something like?

```suggestion
The following formats of events are allowed to be sent:
- a [list of strongly-typed EventGridEvents](add link to list of strongly typed EGE from champion_scenarios.md) 
- a [list of strongly-typed CloudEvents](add link to scenario from champion_scenarios.md) 
- a [single strongly-typed EventGridEvent instance](add link to single strongly typed EGE from champion_scenarios.md) 
- a [single strongly-typed CloudEvent instance](add link to scenario from champion_scenarios.md)
- a list of the [dict representation of EventGridEvents](add link) 
- a list of the [dict representation of CloudEvents](add link)
- a single instance of the [dict representation of EventGridEvent](url)
- a single instance of the [dict representation of CloudEvent](url)
- a list of [dicts in a custom schema](link)
- a single [dict in a custom schema](url)
```

sdk/eventgrid/azure-eventgrid/samples/champion_scenarios.md
```diff
@@ -0,0 +1,331 @@
...
+    "type":"Azure.Sdk.Sample",
+    "time":"2020-08-07T02:06:08.11969Z",
+    "specversion":"1.0"
+}]"""
+
+deserialized_dict_events = [CloudEvent(**msg) for msg in json.loads(cloud_custom_dict)]
+
+for event in deserialized_dict_events:
+    print(event.data)
+    print(type(event))
```
swathipil, line 330: out of curiosity, did you run all these samples outlined here? Or did you copy them from the actual samples and change things to fit each of the scenarios?

sdk/eventgrid/azure-eventgrid/azure/eventgrid/_models.py
```diff
@@ -36,13 +35,15 @@ class CloudEvent(EventMixin):   #pylint:disable=too-many-instance-attributes
...
+    If data is of binary type, data_base64 can be used alternatively. Note that data and data_base64
+    cannot be present at the same time.
 
     :param source: Required. Identifies the context in which an event happened. The combination of id and source must
         be unique for each distinct event. If publishing to a domain topic, source must be the domain name.
     :type source: str
     :param type: Required. Type of event related to the originating occurrence.
     :type type: str
-    :param data: Required. Event data specific to the event type.
+    :keyword data: Optional. Event data specific to the event type. Only one of the `data` or `data_base64` argument must be present.
```
yunhaoling, line 46: as OOB discussion:
we could add a note saying that `if data is of bytes type, we'll send it as `data_base64` in the outgoing request`

By this we add clarity on what users should expect on the data of outgoing request e.g. request.get("data") is None, request.get("data_base64") has value.

sdk/eventgrid/azure-eventgrid/tests/test_eg_publisher_client.py
swathipil: Based on the above, it might be useful to add a sample where you deserialize the CloudEvent that has the `data` field set to a binary value. Then, after deserializing, call `cloud_event.data_base64` and comment that it returns a not None value. Like [this sample](https://github.com/rakshith91/azure-sdk-for-python/blob/arch_preview_feedback/sdk/eventgrid/azure-eventgrid/samples/champion_scenarios/sample_consume_custom_payload.py), except with the data field set to `b'cloudevent'` like the test above.

swathipil (commented):
One more thing: you probably will need to update the `samples/README.md` file to reflect the changes here (which you may already be planning on doing).

sdk/eventgrid/azure-eventgrid/README.md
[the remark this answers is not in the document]
swathipil: RE-ADDED COMMENT BELOW AFTER NEW CHANGES MADE

sdk/eventgrid/azure-eventgrid/samples/champion_scenarios.md
rakshith91: good questions - if you noticed, the from_dict isn't implemented yet :P - other than the two of those, everything else is running and works fine. this file is like a quick reference till i add more relevant samples

sdk/eventgrid/azure-eventgrid/README.md
rakshith91: it would be worth it - i'll tackle it in the samples issue though since i fell this is a non blocker. changing the wording a bit for now

rakshith91, line 63:
```suggestion
CloudEvents and EventGridEvents can be sent either as a single event or a list of respective typed objects or their equivalent dict representations. To send a custom schema, a dict representation can be used. Please have a look at the [samples](https://github.com/Azure/azure-sdk-for-python/tree/master/sdk/eventgrid/azure-eventgrid/samples) for detailed examples.
```

sdk/eventgrid/azure-eventgrid/tests/test_eg_publisher_client.py
rakshith91: we do not have a `from_dict` implementation yet - it'd make more sense to add the samples after we have a symmetric workflow (plus i would like to unblock the breaking changes first)

rakshith91:
Here's the queued live build https://dev.azure.com/azure-sdk/internal/_build/results?buildId=716122&view=results
the CI link: https://dev.azure.com/azure-sdk/public/_build/results?buildId=716137&view=results

sdk/eventgrid/azure-eventgrid/azure/eventgrid/_models.py
```diff
@@ -36,13 +35,16 @@ class CloudEvent(EventMixin):   #pylint:disable=too-many-instance-attributes
...
 
     :param source: Required. Identifies the context in which an event happened. The combination of id and source must
-        be unique for each distinct event. If publishing to a domain topic, source must be the domain name.
+     be unique for each distinct event. If publishing to a domain topic, source must be the domain name.
     :type source: str
     :param type: Required. Type of event related to the originating occurrence.
     :type type: str
-    :param data: Required. Event data specific to the event type.
+    :keyword data: Optional. Event data specific to the event type. Only one of the `data` or `data_base64`
+     argument must be present. If data is of bytes type, we'll send it as data_base64 in the outgoing request.
```
yunhaoling, line 47: tiny tweak :P
```suggestion
     argument must be present. If data is of bytes type, it will be sent as data_base64 in the outgoing request.
```

sdk/eventgrid/azure-eventgrid/samples/publish_samples/publish_custom_schema_events_to_topic_sample.py
```diff
@@ -33,8 +33,8 @@ def publish_event():
     client = EventGridPublisherClient(endpoint, credential)
 
     custom_schema_event = {
-        "customSubject": "sample",
-        "customEventType": "sample.event",
+        "custom_subject": "sample",
+        "custom_EventType": "sample.event",
```
rakshith91, line 37:
```suggestion
        "customSubject": "sample",
        "customEventType": "sample.event",
```

yunhaoling (approved):

PR #16455, opened 2021-02-01
Resolve Regression Failures

scbedd:
Azure-sdk-tools is no longer installed as part of test_tools.txt.

Given that, a standard regression test will fail unless it's been included in the dev_requirements.txt. Some packages _did_ have that, but the regression uses the code as it was when the package released. There a couple packages that relied on `azure-sdk-tools` being present in the test_tools.txt to install it at test run-time. Since we use the newest copy of the regression env, test_tools doesn't include it, and previous tags fail.

Just filter it, then include it ourselves.


scripts/devops_tasks/build_packages.py
```diff
@@ -31,7 +31,7 @@ def build_packages(targeted_packages, distribution_directory, is_dev_build=False
         print("Generating Package Using Python {}".format(sys.version))
         run_check_call(
             [
-                "python",
+                sys.executable,
```
scbedd, line 34: Discovered when running locally that python calling python doesn't nest properly unless you're using system PATH updates like how the VSAgents do them.

Using sys.executable is the proper way to ensure that you're always in the same environment.

praveenkuttappan (approved):
Looks good.

scbedd:
[Manually triggered a quick verification build here.](https://dev.azure.com/azure-sdk/public/_build/results?buildId=713188&view=results)

PR #16460, opened 2021-02-01
[Azure Identity] Add support and documentation for SSH key injection

mccoyp:
Resolves an issue that's come up in [Azure/azure-sdk #901](https://github.com/Azure/azure-sdk/issues/901).

SSH keys are now required for Linux VM deployment -- but only for Microsoft-internal subscriptions. Extra security requirements are never a bad thing though, and in all honesty this will be only be used internally, so the sample has been changed to require SSH key injection by default.


sdk/identity/azure-identity/tests/managed-identity-live/service-fabric/arm-templates/cluster.parameters.json
```diff
@@ -17,6 +17,9 @@
     "certificateThumbprint": {
         "value": "<REPLACE THIS: Hexadecimal string that can be found as Thumbprint or X.509 SHA-1 Thumbprint (in hex) in the certificate's properties>"
     },
+    "sshKeyData": {
+        "value": "<REPLACE THIS: If providing an SSH key, this should be the base64-encoded SSH public key (the content of the key's .pub file, with no trailing whitespace). Looks like ssh-rsa {base64 content} {user}@{machine}>"
```
chlowell, line 21: It must be provided, and is it only the base64 encoded content or an SSH public key one-liner?
```suggestion
        "value": "<REPLACE THIS: an SSH public key (the content of the key's .pub file, with no trailing whitespace). Looks like ssh-rsa {base64 content} {user}@{machine}>"
```

chlowell (approved):
Looks good to me. There's no rush to merge, if you want another manual test pass first.

sdk/identity/azure-identity/tests/managed-identity-live/service-fabric/arm-templates/cluster.parameters.json
mccoyp: Good point -- "base64-encoded SSH public key" came from ARM's documentation, but it should but the full public key. Thanks!

PR #16464, opened 2021-02-02
[SB] combine conn str parser logic in base handler and _common

### LINKED ISSUE #13683: [ServiceBus] Combine various connection string parsing logic
Currently, parse_conn_str is implemented in _VERY_ similar manner in utils.py and in base_handler.py

One of these is for the data-plane ServiceBusAdministration code, the other is for the data-plane itself (servicebusclient and family.)

They can likely be combined significantly to reduce code duplication and improve maintainability.

In addition to this, _create_credential_from_connection_string_parameters is a VERY strange flow right now in order to provide sync vs async dispatch of the credential objects, coupling their respective implementation to the corrosponding sync or async handler.  If there's a clever way to move this all into utils we're game for improvements in this space.  (Considered using a functional callback to pass in the respective sync or async credential builder as a param to the create_credential function, but that seemed like a lateral move at best.)

swathipil:
Fixes #13683

### DIFF
--- a/sdk/servicebus/azure-servicebus/azure/servicebus/_base_handler.py
+++ b/sdk/servicebus/azure-servicebus/azure/servicebus/_base_handler.py
@@ -59,7 +59,7 @@ def _parse_conn_str(conn_str, check_case=False):
     shared_access_signature_expiry = None  # type: Optional[int]
 
     # split connection string into properties
-    conn_properties = [s.split("=", 1) for s in conn_str.rstrip(";").split(";")]
+    conn_properties = [s.split("=", 1) for s in conn_str.strip().rstrip(";").split(";")]
     if any(len(tup) != 2 for tup in conn_properties):
         raise ValueError("Connection string is either blank or malformed.")
     conn_settings = dict(conn_properties)   # type: ignore
--- a/sdk/servicebus/azure-servicebus/azure/servicebus/_base_handler.py
+++ b/sdk/servicebus/azure-servicebus/azure/servicebus/_base_handler.py
@@ -14,6 +14,11 @@ try:
 except ImportError:
     from urllib.parse import quote_plus
 
+try:
+    from urllib.parse import urlparse
+except ImportError:
+    from urlparse import urlparse  # type: ignore
+
 import uamqp
 from uamqp import utils, compat
 from uamqp.message import MessageProperties
@@ -83,17 +88,27 @@ def _parse_conn_str(conn_str):
             ):  # Fallback since technically expiry is optional.
                 # An arbitrary, absurdly large number, since you can't renew.
                 shared_access_signature_expiry = int(time.time() * 2)
-    if not (
-        all((endpoint, shared_access_key_name, shared_access_key))
-        or all((endpoint, shared_access_signature))
-    ) or all(
-        (shared_access_key_name, shared_access_signature)
-    ):  # this latter clause since we don't accept both
+
+    if not endpoint:
+        raise ValueError("Connection string is either blank or malformed.")
+    parsed = urlparse(endpoint.rstrip("/"))
+    if not parsed.netloc:
+        raise ValueError("Invalid Endpoint on the Connection String.")
+    if any([shared_access_key, shared_access_key_name]) and not all(
+        [shared_access_key, shared_access_key_name]
+    ):
         raise ValueError(
-            "Invalid connection string. Should be in the format: "
-            "Endpoint=sb://<FQDN>/;SharedAccessKeyName=<KeyName>;SharedAccessKey=<KeyValue>"
-            "\nWith alternate option of providing SharedAccessSignature instead of SharedAccessKeyName and Key"
+            "Connection string must have both SharedAccessKeyName and SharedAccessKey."
         )
+    if shared_access_signature is not None and shared_access_key is not None:
+        raise ValueError(
+            "Only one of the SharedAccessKey or SharedAccessSignature must be present."
+        )
+    if shared_access_signature is None and shared_access_key is None:
+        raise ValueError(
+            "At least one of the SharedAccessKey or SharedAccessSignature must be present."
+        )
+
     entity = cast(str, entity_path)
     host = cast(str, strip_protocol_from_uri(cast(str, endpoint)))
 
--- a/sdk/servicebus/azure-servicebus/azure/servicebus/_base_handler.py
+++ b/sdk/servicebus/azure-servicebus/azure/servicebus/_base_handler.py
@@ -10,13 +10,9 @@ from datetime import timedelta
 from typing import cast, Optional, Tuple, TYPE_CHECKING, Dict, Any, Callable
 
 try:
+    from urllib.parse import quote_plus, urlparse
+except ImportError:
     from urllib import quote_plus  # type: ignore
-except ImportError:
-    from urllib.parse import quote_plus
-
-try:
-    from urllib.parse import urlparse
-except ImportError:
     from urlparse import urlparse  # type: ignore
 
 import uamqp
@@ -53,7 +49,7 @@ if TYPE_CHECKING:
 _LOGGER = logging.getLogger(__name__)
 
 
-def _parse_conn_str(conn_str):
+def _parse_conn_str(conn_str, check_case=False):
     # type: (str) -> Tuple[str, Optional[str], Optional[str], str, Optional[str], Optional[int]]
     endpoint = None
     shared_access_key_name = None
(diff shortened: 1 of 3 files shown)


sdk/servicebus/azure-servicebus/azure/servicebus/_common/_connection_string_parser.py
```diff
@@ -71,39 +67,14 @@ def parse_connection_string(conn_str):
...
-    conn_settings = [s.split("=", 1) for s in conn_str.split(";")]
-    if any(len(tup) != 2 for tup in conn_settings):
-        raise ValueError("Connection string is either blank or malformed.")
-    conn_settings = dict(conn_settings)
-    shared_access_signature = None
-    for key, value in conn_settings.items():
-        if key.lower() == "sharedaccesssignature":
-            shared_access_signature = value
-    shared_access_key = conn_settings.get("SharedAccessKey")
-    shared_access_key_name = conn_settings.get("SharedAccessKeyName")
```
yunhaoling, line 83: the connection string parser actually cares about the case of the string while the `_parse_conn_str`does not: e.g. `sharedaccesskey` vs `SharedAccessKey`.

so we need to take this into account when combining the parse logic.

sdk/servicebus/azure-servicebus/azure/servicebus/_common/_connection_string_parser.py
```diff
@@ -71,39 +67,14 @@ def parse_connection_string(conn_str):
...
-        raise ValueError("Connection string is either blank or malformed.")
-    parsed = urlparse(endpoint.rstrip("/"))
-    if not parsed.netloc:
-        raise ValueError("Invalid Endpoint on the Connection String.")
-    namespace = parsed.netloc.strip()
+    namespace, policy, key, entity, signature = _parse_conn_str(conn_str)[:-1]
+    endpoint = "sb://" + namespace + "/"
     props = {
         "fully_qualified_namespace": namespace,
         "endpoint": endpoint,
```
yunhaoling, line 74:
```suggestion
    fully_qualified_namespace, policy, key, entity, signature = _parse_conn_str(conn_str)[:-1]
    endpoint = "sb://" + namespace + "/"
    props = {
        "fully_qualified_namespace": fully_qualified_namespace,
        "endpoint": endpoint,
```

sdk/servicebus/azure-servicebus/azure/servicebus/_base_handler.py
```diff
@@ -83,17 +88,27 @@ def _parse_conn_str(conn_str):
...
+            "Connection string must have both SharedAccessKeyName and SharedAccessKey."
         )
+    if shared_access_signature is not None and shared_access_key is not None:
+        raise ValueError(
+            "Only one of the SharedAccessKey or SharedAccessSignature must be present."
+        )
+    if shared_access_signature is None and shared_access_key is None:
+        raise ValueError(
+            "At least one of the SharedAccessKey or SharedAccessSignature must be present."
+        )
```
yunhaoling, line 110:
```suggestion
    if shared_access_signature and shared_access_key:
        raise ValueError(
            "Only one of the SharedAccessKey or SharedAccessSignature must be present."
        )
    if not shared_access_signature and not shared_access_key:
        raise ValueError(
            "At least one of the SharedAccessKey or SharedAccessSignature must be present."
        )
```

sdk/servicebus/azure-servicebus/azure/servicebus/_common/_connection_string_parser.py
yunhaoling: also I think it's worth to add tests against the case of the string in the two scenarios

sdk/servicebus/azure-servicebus/azure/servicebus/_common/_connection_string_parser.py
```diff
@@ -71,39 +67,14 @@ def parse_connection_string(conn_str):
...
-        "entity_path": conn_settings.get("EntityPath"),
```
yunhaoling, line 104: same case problem here: `EntityPath` vs `entitypath`

sdk/servicebus/azure-servicebus/azure/servicebus/_base_handler.py
```diff
@@ -14,6 +14,11 @@
 except ImportError:
     from urllib.parse import quote_plus
 
+try:
+    from urllib.parse import urlparse
+except ImportError:
+    from urlparse import urlparse  # type: ignore
```
yunhaoling, line 20: we should combine the import logic into line 12-15

sdk/servicebus/azure-servicebus/azure/servicebus/_common/_connection_string_parser.py
```diff
@@ -71,39 +67,14 @@ def parse_connection_string(conn_str):
...
-        )
-    endpoint = conn_settings.get("Endpoint")
-    if not endpoint:
-        raise ValueError("Connection string is either blank or malformed.")
-    parsed = urlparse(endpoint.rstrip("/"))
-    if not parsed.netloc:
-        raise ValueError("Invalid Endpoint on the Connection String.")
-    namespace = parsed.netloc.strip()
+    namespace, policy, key, entity, signature = _parse_conn_str(conn_str)[:-1]
+    endpoint = "sb://" + namespace + "/"
```
yunhaoling, line 71: could this be returned from `_parse_conn_str` instead of we manually rebuild it?

sdk/servicebus/azure-servicebus/azure/servicebus/_base_handler.py
```diff
@@ -83,17 +88,27 @@ def _parse_conn_str(conn_str):
...
-        or all((endpoint, shared_access_signature))
-    ) or all(
-        (shared_access_key_name, shared_access_signature)
-    ):  # this latter clause since we don't accept both
+
+    if not endpoint:
+        raise ValueError("Connection string is either blank or malformed.")
+    parsed = urlparse(endpoint.rstrip("/"))
+    if not parsed.netloc:
+        raise ValueError("Invalid Endpoint on the Connection String.")
```
yunhaoling, line 96: I think we're checking the fully qualified namespace here, wondering if there's any change we could combine it with `host = cast(str, strip_protocol_from_uri(cast(str, endpoint)))`

sdk/servicebus/azure-servicebus/azure/servicebus/_common/_connection_string_parser.py
swathipil: Yes it can be returned. However, it would also require every other call to _parse_conn_str, including the all of the clients, to be changed as well. I'm not sure whether this would be worth the trade off, especially since this is the only method that would use the returned endpoint value and all the other ones would need to receive it and not store it.

sdk/servicebus/azure-servicebus/azure/servicebus/_base_handler.py
swathipil: since `parsed.netloc` already strips the protocol, I think that can be substituted for `strip_protocol_from_uri(cast(str, endpoint))`

sdk/servicebus/azure-servicebus/azure/servicebus/_base_handler.py
```diff
@@ -48,27 +49,32 @@
...
-        elif key.lower() == "sharedaccesskeyname":
-            shared_access_key_name = value
-        elif key.lower() == "sharedaccesskey":
-            shared_access_key = value
-        elif key.lower() == "entitypath":
-            entity_path = value
-        elif key.lower() == "sharedaccesssignature":
+
+    # split connection string into properties
+    conn_settings = [s.split("=", 1) for s in conn_str.rstrip(";").split(";")]
```
yunhaoling, line 62: This looks neat! :P
may I ask why we want to rstrip trailing ";" here, as we previously just have `conn_str.strip()`.

yunhaoling (commented):
just left one comment, looks good!
also there're some mypy errors we need to fix :P

sdk/servicebus/azure-servicebus/azure/servicebus/_base_handler.py
swathipil: I saw that in [this test case](https://github.com/Azure/azure-sdk-for-python/blob/db4892a29c45fd9d65da1319d65ffd3269f13898/sdk/servicebus/azure-servicebus/tests/test_sb_client.py#L259), the manually built conn str ends with `;`. I would have changed the test case if we had the assumption that customers can only get the conn str from the portal (where conn strings don't end in `;`). However, since users can manually create the conn strings, they may end the connection string with `;`. This would add an empty string to `conn_settings` when splitting on `;` and raise the `ValueError`. So I kept the trailing `;` in the test case, and accounted for it here. Would it have been better to change the test case instead, and say that a connection string cannot end with `;`?

yunhaoling: I think we should change the test case instead, keeping `conn_str.strip()` first.
we didn't know whether a conn_str with trailing ';' is valid or not, it's better the keep the status quo until we get more input :P

swathipil: It looks like .NET does strip trailing `;` : https://github.com/Azure/azure-sdk-for-net/blob/5fedc94bca9fb9d1a941eda7a8af1eceebb5809a/sdk/servicebus/Microsoft.Azure.ServiceBus/tests/ServiceBusConnectionStringBuilderTests.cs#L75. For consistency, should we also `rstrip(';')`. So something like `conn_str.strip().rstrip(';').split(';')`.

yunhaoling: I think we're good now, thanks for addressing my bikeshedding!

sdk/servicebus/azure-servicebus/azure/servicebus/_common/_connection_string_parser.py
yunhaoling: you're more considerate than me :P
I think we're good for now

yunhaoling (approved):
LGTM

PR #16474, opened 2021-02-02
[Key Vault] Test keys library against a shared vault

### LINKED ISSUE #15435: [Key Vault] Update tests to work on a shared vault (keys)
Part of solution to #15371.

Divided into library-specific issues for manageability.

mccoyp:
Resolves #15435.

This drops KV's existing preparers, and instead uses a `PowerShellPreparer` to fetch an existing vault URL from an environment variable. 
`test-resources.json` and `test-resources-post.ps1` are added to make resource group, vault, and managed HSM creation as easy as [running a script](https://github.com/Azure/azure-sdk-tools/tree/master/eng/common/TestResources). 
`test-resources-cleanup` cleans up test resources from a vault, which are identified by a "livekvtest" prefix -- this is automatically added by `get_resource_name()`.


sdk/keyvault/azure-keyvault-keys/tests/_shared/test_case.py
```diff
@@ -20,6 +22,16 @@ def setUp(self):
         self.list_test_size = 7
         super(KeyVaultTestCase, self).setUp()
 
+    def create_kv_client(self, vault_uri, **kwargs):
+        credential = self.get_credential(KeyClient)
+        return self.create_client_from_credential(
+            KeyClient, credential=credential, vault_url=vault_uri, **kwargs
```
chlowell, line 28: _shared is identical across libraries, so this kind of thing belongs somewhere else. Here's a thought:

```py
class KeyVaultTestCase(AzureTestCase):
    def __init__(self, *args, **kwargs):
        self.client = self.create_client(vault_uri)

class KeyClientTests(KeyVaultTestCase):
    def create_client(self, vault_uri):
        return self.create_client_from_credential(KeyClient, ...)

    # looks like this pattern enables removing the preparer
    def test_something(self):
        self.client.do_something()

        # or, if you don't like touching every client call
        client = self.client
```

sdk/keyvault/test-resources-cleanup.py
```diff
@@ -0,0 +1,47 @@
+import os
```
chlowell, line 1: Needs a license header

sdk/keyvault/test-resources-cleanup.py
```diff
@@ -0,0 +1,47 @@
...
+if "KEYVAULT_CLIENT_SECRET" not in os.environ:
+    raise EnvironmentError("Missing a client secret for Key Vault")
+
+credential = ClientSecretCredential(
+    tenant_id=os.environ.get("KEYVAULT_TENANT_ID"),
+    client_id=os.environ.get("KEYVAULT_CLIENT_ID"),
+    client_secret=os.environ.get("KEYVAULT_CLIENT_SECRET")
+)
+
+if "AZURE_KEYVAULT_URL" in os.environ:
```
chlowell, line 23: This will always be True (if it were False, we would have raised on line 9)

sdk/keyvault/test-resources-cleanup.py
```diff
@@ -0,0 +1,47 @@
...
+    raise EnvironmentError("Missing a tenant ID for Key Vault")
+if "KEYVAULT_CLIENT_ID" not in os.environ:
+    raise EnvironmentError("Missing a client ID for Key Vault")
+if "KEYVAULT_CLIENT_SECRET" not in os.environ:
+    raise EnvironmentError("Missing a client secret for Key Vault")
+
+credential = ClientSecretCredential(
+    tenant_id=os.environ.get("KEYVAULT_TENANT_ID"),
+    client_id=os.environ.get("KEYVAULT_CLIENT_ID"),
+    client_secret=os.environ.get("KEYVAULT_CLIENT_SECRET")
```
chlowell, line 20: Nitpicking: no need to call `get`, we know these keys have values.

sdk/keyvault/test-resources-cleanup.py
```diff
@@ -0,0 +1,47 @@
+import os
+
+from azure.identity import ClientSecretCredential
+from azure.keyvault.certificates import CertificateClient
+from azure.keyvault.keys import KeyClient
+from azure.keyvault.secrets import SecretClient
+
```
chlowell, line 7: Should this call load_dotenv?

sdk/keyvault/test-resources-cleanup.py
mccoyp: Thanks for the catch -- that was a remnant of when I had prototyped this to handle MHSM cleanup as well, but I decided that was best left for another time

sdk/keyvault/azure-keyvault-keys/tests/_shared/test_case.py
mccoyp: True, something this would obviate the preparer if I fetch the vault URL before running tests. I actually used a pattern like this for that reason in a prototype, but Laurent would like the SDK to adopt the PowerShellPreparer for common infrastructure and to move away from a dependency on test classes. Moving the client creation out of _shared since the client differs makes sense though

mccoyp: What makes the most sense to me is to keep the preparer behavior, but to just move the client creation method into KeyClientTests, etc. What do you think?

sdk/keyvault/azure-keyvault-keys/tests/test_crypto_client.py
```diff
@@ -35,6 +41,14 @@ def __init__(self, *args, **kwargs):
 
     plaintext = b"5063e6aaa845f150200547944fd199679c98ed6f99da0a0b2dafeaf1f4684496fd532c1c229968cb9dee44957fcef7ccef59ceda0b362e56bcd78fd3faee5781c623c0bb22b35beabde0664fd30e0e824aba3dd1b0afffc4a3d955ede20cf6a854d52cfd"
 
+    def create_client(self, vault_uri, **kwargs):
```
chlowell, line 44:
```suggestion
    def create_key_client(self, vault_uri, **kwargs):
```

sdk/keyvault/azure-keyvault-keys/tests/test_keys_async.py
```diff
@@ -215,18 +222,45 @@ async def test_key_crud_operations(self, client, **kwargs):
...
+        created_bundle = await client.create_key(key_name, key_type)
+        self.assertEqual(key_type, created_bundle.key_type)
+
+        # backup key
+        key_backup = await client.backup_key(created_bundle.name)
+        self.assertIsNotNone(key_backup, "key_backup")
+
+        # delete key
+        await client.delete_key(created_bundle.name)
+        # can add test case to see if we do get_deleted should return error
```
chlowell, line 243: 🤔

sdk/keyvault/azure-keyvault-keys/tests/test_samples_keys.py
```diff
@@ -35,21 +37,22 @@ def test_create_key_client():
...
+        key_name = self.get_resource_name("key-name")
 
         # [START create_key]
         from dateutil import parser as date_parse
 
         expires_on = date_parse.parse("2050-02-02T08:00:00.000Z")
 
         # create a key with optional arguments
-        key = key_client.create_key("key-name", "RSA-HSM", expires_on=expires_on)
+        key = key_client.create_key(key_name, "RSA-HSM", expires_on=expires_on)
```
chlowell, line 55:
```suggestion
        key = key_client.create_key(key_name, KeyType.rsa_hsm, expires_on=expires_on)
```

sdk/keyvault/azure-keyvault-keys/tests/test_samples_keys_async.py
```diff
@@ -47,7 +54,7 @@ async def test_example_key_crud_operations(self, client, **kwargs):
 
         # create a key with optional arguments
         key = await key_client.create_key(
-            "key-name", "RSA", size=key_size, key_operations=key_ops, expires_on=expires_on
+            key_name, "RSA", size=key_size, key_operations=key_ops, expires_on=expires_on
```
chlowell, line 57: Don't recall why we ended up with a string here. It seems better to demonstrate the enum in snippets:
```suggestion
            key_name, KeyType.rsa, size=key_size, key_operations=key_ops, expires_on=expires_on
```

sdk/keyvault/azure-keyvault-keys/tests/test_examples_crypto_async.py
```diff
@@ -2,23 +2,38 @@
...
+    azure_keyvault_url="https://vaultname.vault.azure.net"
+)
 
 
 class TestCryptoExamples(KeyVaultTestCase):
     def __init__(self, *args, **kwargs):
         kwargs["match_body"] = False
         super(TestCryptoExamples, self).__init__(*args, **kwargs)
 
+    def create_client(self, vault_uri, **kwargs):
```
chlowell, line 24:
```suggestion
    def create_key_client(self, vault_uri, **kwargs):
```

chlowell (commented):
Were the recordings made with the code as it is in the latest revision of this PR (it looks like they were recorded by test cases sharing a single client instance)?

mccoyp:
These recordings were all made after other changes were made, so I'm not sure why that would be the case

sdk/keyvault/azure-keyvault-keys/tests/test_keys_async.py
mccoyp: 🤔, indeed

I only moved this test case up to match the ordering of the synchronous tests, so I don't think I wrote this line... in any case, `test_purge` covers what I think is being described here

chlowell:
It's because the ChallengeAuthenticationPolicy uses a process-wide cache of challenge information. In a live run, the first test case caches challenge information and subsequent test cases--all using the same vault, whose URL is the cache key--use that cached information rather than send a request to get it from Key Vault. Therefore only one recording contains the initial empty request, which recording being determined by order of execution. This implies that when you play the tests back in a different order than they were recorded, you can expect failure because the challenge cache will be empty until the test case with the recorded challenge runs. Each test that runs before the challenge is played back will send a request with no recorded response: vcr should raise.

But in fact this doesn't happen; the tests pass because:
1. they don't configure vcr to match request bodies (because the bodies contain random data)
2. the initial, empty request the client sends to elicit a challenge from Key Vault therefore matches the subsequent, authorized request so far as vcr is concerned
3. when it gets the empty request, vcr replays the response to the authorized request
4. that response doesn't have a WWW-Authenticate header, so the auth policy fails to parse a challenge from it
5. the policy's failure mode in this case is to return the response (because it doesn't know what else to do with it)
6. the returned response isn't an error, so the pipeline doesn't complain
7. actually the response is exactly what the test asked for when it called `client.do_whatever()`
8. success!

So these recordings create an error condition that probably should, but coincidentally does not, cause test failures 😆 

The easiest solution is to clear the cache between tests:
https://github.com/Azure/azure-sdk-for-python/blob/0763539490824f7ffcf32a899d27c3e90aed8306/sdk/keyvault/azure-keyvault-keys/tests/test_challenge_auth.py#L33
...something the preparer could handle?

mccoyp:
I can ask Sean about adding it to the PowerShellPreparer (or we could subclass the preparer and do this from there), but I've verified locally that this could also be done in a tearDown method of each test class. The former would be cleaner though 🤔

chlowell:
It's a Key Vault only thing. I thought tearDown was called after all the test cases have run?

mccoyp:
One would think that would be the case, but it's called after each test method finishes up -- kind of convenient, kind of confusing

chlowell (approved):

PR #16476, opened 2021-02-02
Added package properties SDKType and NewSDK

praveenkuttappan:
Added a wrapper method in common_tasks to parse package name, version and whether it is track 2 or not.

### DIFF
--- a/scripts/devops_tasks/common_tasks.py
+++ b/scripts/devops_tasks/common_tasks.py
@@ -466,5 +466,5 @@ def get_package_properties(setup_py_path):
     """Parse setup.py and return package details like package name, version, whether it's new SDK
     """
     pkgName, version, _, requires = parse_setup(setup_py_path)
-    is_new_pkg = pkgName in NEW_REQ_PACKAGES or any(map(lambda x: (parse_require(x)[0] in NEW_REQ_PACKAGES), requires))
-    return pkgName, version, is_new_pkg
+    is_new_sdk = pkgName in NEW_REQ_PACKAGES or any(map(lambda x: (parse_require(x)[0] in NEW_REQ_PACKAGES), requires))
+    return pkgName, version, is_new_sdk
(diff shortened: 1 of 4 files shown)


scripts/devops_tasks/common_tasks.py
```diff
@@ -461,3 +462,9 @@ def get_installed_packages(paths = None):
     ws = WorkingSet(paths) if paths else working_set  
     return ["{0}=={1}".format(p.project_name, p.version) for p in ws]
 
+def get_package_properties(setup_py_path):
+    """Parse setup.py and return package details like package name, version, whether it's track1 or track2
+    """
+    pkgName, version, _, requires = parse_setup(setup_py_path)
+    is_track2_pkg = pkgName in TRACK2_IDENTIFIER_REQ_PACKAGES or any(map(lambda x: (parse_require(x)[0] in TRACK2_IDENTIFIER_REQ_PACKAGES), requires))
```
weshaggard, line 469: I suggest calling this is_new_sdk.

scripts/devops_tasks/common_tasks.py
```diff
@@ -32,6 +32,7 @@
 
 DEV_REQ_FILE = "dev_requirements.txt"
 NEW_DEV_REQ_FILE = "new_dev_requirements.txt"
+TRACK2_IDENTIFIER_REQ_PACKAGES = ["azure-core", "azure-mgmt-core"]
```
weshaggard, line 35: Lets try avoid using the name "track2". Lets use something like NEW_REQ_PACKAGES

weshaggard (approved):

scripts/devops_tasks/common_tasks.py
```diff
@@ -461,3 +462,9 @@ def get_installed_packages(paths = None):
     ws = WorkingSet(paths) if paths else working_set  
     return ["{0}=={1}".format(p.project_name, p.version) for p in ws]
 
+def get_package_properties(setup_py_path):
```
scbedd, line 465: So it's a new SDK if the package is azure-core, azure-mgmt-core, or if any of the requirements are one of those? 

What?

scbedd (approved):

scripts/devops_tasks/common_tasks.py
weshaggard: For reference we are following https://github.com/Azure/azure-sdk/issues/1261#issuecomment-691327942. The idea is we needed a way to detect if a package is track 2 (aka new).

PR #16479, opened 2021-02-02
Create/Update methods stop calling get

xiangyan99:


kristapratico (approved):

kristapratico:
I think we need to update the README samples as well :)

PR #16496, opened 2021-02-03
bump six dependencies in some libraries

iscai-msft:


kristapratico (approved):
update changelogs?

iscai-msft:
@kristapratico will do in another PR, this ci has been so painful to run

PR #16504, opened 2021-02-03
Fix min dependency tests - update azure core

rakshith91:
trigger - https://dev.azure.com/azure-sdk/internal/_build/results?buildId=718018&view=results


scbedd (approved):
Remember to mention in changelog! @rakshith91

PR #16509, opened 2021-02-03
Implement from_dict

rakshith91:


sdk/eventgrid/azure-eventgrid/azure/eventgrid/_helpers.py
```diff
@@ -110,3 +110,12 @@ def _eventgrid_data_typecheck(event):
     if isinstance(data, six.binary_type):
         raise TypeError("Data in EventGridEvent cannot be bytes. Please refer to"\
             "https://docs.microsoft.com/en-us/azure/event-grid/event-schema")
+
+def _decode(content):
+    try:
+        return base64.b64decode(content.encode('utf-8'))
```
yunhaoling, line 116: dumb question: what does the `content` look like, utf-8 base64 string?

rakshith91: "data_base64":'Y2xvdWRldmVudA==',

sdk/eventgrid/azure-eventgrid/azure/eventgrid/_models.py
```diff
@@ -106,6 +107,31 @@ def __init__(self, source, type, **kwargs): # pylint: disable=redefined-builtin
...
                 Use data_base64 only if you are sending bytes, and use data otherwise.")
 
+    @classmethod
+    def from_dict(cls, event, **kwargs):
+        # type: (Dict, Any) -> CloudEvent
+        """
+        Returns the deserialized CloudEvent object when a dict is provided.
+
+        :param event: The dict representation of the event which needs to be deserialized.
+        :type event: dict
```
yunhaoling, line 117: I think it would be helpful if we could add description on the key-value items we accepts -- but it's a doc stuff

sdk/eventgrid/azure-eventgrid/azure/eventgrid/_models.py
```diff
@@ -228,3 +254,26 @@ def __init__(self, subject, event_type, data, data_version, **kwargs):
...
+    @classmethod
+    def from_dict(cls, event, **kwargs):
+        # type: (Dict, Any) -> EventGridEvent
+        """
+        Returns the deserialized EventGridEvent object when a dict is provided.
+
+        :param event: The dict representation of the event which needs to be deserialized.
+        :type event: dict
+
+        :rtype: EventGridEvent
```
yunhaoling, line 267: same comment as above regarding to the description of the dict accepted key-value items.

sdk/eventgrid/azure-eventgrid/tests/test_from_dict.py
```diff
@@ -0,0 +1,56 @@
...
+    def test_cloud_custom_dict_with_extensions(self, **kwargs):
+        event = CloudEvent.from_dict(cloud_custom_dict_with_extensions)
+        assert event.data == {"team": "event grid squad"}
+        assert event.__class__ == CloudEvent
+        assert event.extensions == {"ext1": "example", "ext2": "example2"}
+
+    def test_cloud_custom_dict_base64(self, **kwargs):
+        event = CloudEvent.from_dict(cloud_custom_dict_base64)
+        assert event.data == b'cloudevent'
+        assert event.data_base64 == None
```
yunhaoling, line 46: as I mentioned above in the `from_dict`, the dictionary item is `data_base64` while we actually set it to the `data` which I think kinda breaks consistency from the experience of using the client constructor.

sdk/eventgrid/azure-eventgrid/azure/eventgrid/_models.py
```diff
@@ -106,6 +107,31 @@ def __init__(self, source, type, **kwargs): # pylint: disable=redefined-builtin
...
+        :type event: dict
+
+        :rtype: CloudEvent
+        """
+        return cls(
+        id=event.pop("id", None),
+        source=event.pop("source", None),
+        type=event.pop("type", None),
+        specversion=event.pop("specversion", None),
+        data=event.pop("data", None) or _decode(event.pop("data_base64", None)),
```
yunhaoling, line 126: In my understanding, to use the `from_dict` for `CloudEvent`, users should be doing this:

```python
dict = {
    data_base64 = b'<base64 bytes>'
}

# or

dict = {
    data = '<data>'
}
```

which means internally we do:
```
data=event.pop("data", None)
data_base64=event.pop("data_base64", None),
```

is this by design?

sdk/eventgrid/azure-eventgrid/azure/eventgrid/_models.py
rakshith91: hmm, the payload normally comes directly from a storage queue - i can put a link to the relevant spec

sdk/eventgrid/azure-eventgrid/azure/eventgrid/_models.py
swathipil: I think it is by design, and mirrors [this](https://github.com/Azure/azure-sdk-for-python/pull/16441#discussion_r569031063). Similar to sending `data` that is of bytes type (sent as `data_base64` in the outgoing request), when `from_dict` is used to deserialize, it sets `data_base64` back to `data`.

yunhaoling: I'm thinking of the following two creation case which makes me feel the inconsistency, from intuition I would expect two cases to return the "same" `CloudEvent` :
```python
event_created_from_class = CloudEvent(data_base64=b'bytes_data')
event_created_from_class.data is None
event_created_from_class.data_base64 is not None

event_created_from_dict = CloudEvent.from_dict({'data_base64':b'bytes_data'})
event_created_from_class.data is not None
event_created_from_dict.data_base64 is None
```

apart from that, I notice that `from_dict` could take both `data` and `data_base64`, and if both are set `data_base64` gets ignored:
```python
# I think this should be an invalid input?
event_created_from_dict = CloudEvent.from_dict({'data': 'somedata', 'data_base64':b'bytes_data'})
```

yunhaoling: (I could also be wrong, might be missing some context here, just thinking out loud)

sdk/eventgrid/azure-eventgrid/azure/eventgrid/_models.py
swathipil: If not all the key-value pairs, at least mentioning that `data_base64` will always be decoded and set as the value for the `data` field would be helpful. In general, maybe mentioning that `from_dict` will never result in the `data_base64` field being set in the resulting `CloudEvent` object (if I'm getting that right).

sdk/eventgrid/azure-eventgrid/azure/eventgrid/_models.py
swathipil: That two ways that `data_base64` can be deserialized is a little confusing since, no matter what, only `data` will end up being set. Maybe clarifying this in the docstring of `from_dict` would be useful?

yunhaoling: we could also do a cross-language check

sdk/eventgrid/azure-eventgrid/azure/eventgrid/_helpers.py
rakshith91: so we decided not to expose data_base64 to users at all 
see https://github.com/Azure/azure-sdk-for-python/pull/16661

sdk/eventgrid/azure-eventgrid/azure/eventgrid/_models.py
rakshith91: update: we decided not to expose data_base64
https://github.com/Azure/azure-sdk-for-python/pull/16661

rakshith91:
Closing in favor of https://github.com/Azure/azure-sdk-for-python/pull/16661

PR #16547, opened 2021-02-04
Adding missing supported fields to invoice sample

### LINKED ISSUE #15792: [formrecognizer] show all available fields in invoice sample
This sample form should return all 26 fields: https://docs.microsoft.com/en-us/azure/cognitive-services/form-recognizer/media/sample-invoice.jpg

We should update our [sync](https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/formrecognizer/azure-ai-formrecognizer/samples/sample_recognize_invoices.py) / [async](https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/formrecognizer/azure-ai-formrecognizer/samples/async_samples/sample_recognize_invoices_async.py) invoice samples to use this sample form and print all available fields.

Supported fields by name can be found on this page: https://westcentralus.dev.cognitive.microsoft.com/docs/services/form-recognizer-api-v2-1-preview-2/operations/5ed8c9acb78c40a2533aee83

catalinaperalta:
Fixes #15792.


catalinaperalta:
I've updated the sample to use the invoice linked in the issue, but I could create another sample that uses the other invoice as well. Not sure how much value it would add since that one only shows a subset of the fields that the new invoice adds.

kristapratico (approved):
Looks good!

kristapratico:
> I've updated the sample to use the invoice linked in the issue, but I could create another sample that uses the other invoice as well. Not sure how much value it would add since that one only shows a subset of the fields that the new invoice adds.

I think one sample per sync/async for invoices is enough.

iscai-msft (approved):
oh my gosh @catalinaperalta so much work

PR #16549, opened 2021-02-04
adding step to upload tox logs on a failure

### LINKED ISSUE #9168: Create Additional "Debugging" Artifact on Builds
Right now, the debugging experience is limited to what you can see in stdout/stderr on the build agent window. 

We should additional attach:

- `tox` installation logs
- any additional caught exception information (if available)

Upload in an attached artifact of loose file artifacts.

seankane-msft:
Fixes #9168

### DIFF
--- a/eng/pipelines/templates/steps/build-test.yml
+++ b/eng/pipelines/templates/steps/build-test.yml
@@ -53,9 +53,14 @@ steps:
         ${{ parameters.ToxEnvParallel }}
     env: ${{ parameters.EnvVars }}
 
+  - pwsh: |
+      Get-ChildItem .\_tox_logs\ -Recurse
+    displayName: 'Show tox logs files'
+
   - task: PublishPipelineArtifact@1
-    condition: failed()
     displayName: Publish Tox Logs
+    condition: failed()
+    continueOnError: true
     inputs:
       targetPath: '$(Build.SourcesDirectory)/_tox_logs'
       artifactType: 'pipeline'
--- a/eng/pipelines/templates/steps/build-test.yml
+++ b/eng/pipelines/templates/steps/build-test.yml
@@ -53,10 +53,6 @@ steps:
         ${{ parameters.ToxEnvParallel }}
     env: ${{ parameters.EnvVars }}
 
-  - pwsh: |
-      Get-ChildItem .\_tox_logs\ -Recurse
-    displayName: 'Show tox logs files'
-
   - task: PublishPipelineArtifact@1
     displayName: Publish Tox Logs
     condition: failed()
--- a/eng/pipelines/templates/steps/build-test.yml
+++ b/eng/pipelines/templates/steps/build-test.yml
@@ -53,6 +53,13 @@ steps:
         ${{ parameters.ToxEnvParallel }}
     env: ${{ parameters.EnvVars }}
 
+  - task: PublishPipelineArtifact@1
+    condition: failed()
+    displayName: Publish Tox Logs
+    inputs:
+      targetPath: '$(Build.SourcesDirectory)/_tox_logs'
+      artifactType: 'pipeline'
+
   - ${{ parameters.AfterTestSteps }}
 
   - pwsh: |
--- a/scripts/devops_tasks/setup_execute_tests.py
+++ b/scripts/devops_tasks/setup_execute_tests.py
@@ -222,7 +222,7 @@ if __name__ == "__main__":
     parser.add_argument(
         "--tparallel",
         default=False,
-        help=("Flag  that enables parallel tox invocation."),
+        help=("Flag that enables parallel tox invocation."),
         action="store_true",
     )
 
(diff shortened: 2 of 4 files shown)


eng/pipelines/templates/steps/build-test.yml
```diff
@@ -53,6 +53,13 @@ steps:
         ${{ parameters.ToxEnvParallel }}
     env: ${{ parameters.EnvVars }}
 
+  - task: PublishPipelineArtifact@1
+    condition: failed()
```
weshaggard, line 57: Don't these need an artifactName as well?

scripts/devops_tasks/setup_execute_tests.py
```diff
@@ -307,3 +335,5 @@ def execute_global_install_and_test(
         execute_global_install_and_test(args, targeted_packages, extended_pytest_args)
     else:
         prep_and_run_tox(targeted_packages, args, extended_pytest_args)
+
+    collect_log_files()
```
scbedd, line 339: Am concerned. 

`execute_tox_serial` in `tox_harness.py:277` will blow away the tox directory after it finishes up with each run. This is because the CI machines _do not_ have very much space and a full venv is pretty big size wise.

Gonna need to exclude it from the cleanup over there.

scbedd: Locally you probably haven't seen this because you're running in a way that doesn't trigger the `in_ci` bool.

To try in pshell, call `$Env:TF_BUILD=whatever_text_you_want_needs_to_be_non_empty`

eng/pipelines/templates/steps/build-test.yml
seankane-msft: No, it's uploading a folder and Azure Pipelines will create the new folder and upload all the contents from the folder given.

scripts/devops_tasks/setup_execute_tests.py
seankane-msft: I believe I have worked around this now

eng/pipelines/templates/steps/build-test.yml
weshaggard: What name does it end up giving the artifact?

seankane-msft: It will be named after the job name, Python39 for example.

scripts/devops_tasks/tox_harness.py
```diff
@@ -274,11 +274,66 @@ def replace_dev_reqs(file, pkg_root):
         f.write("\n".join(adjusted_req_lines))
 
 
+def collect_log_files(working_dir):
+    logging.info("Collecting log files from {}".format(working_dir))
+    package = working_dir.split('/')[-1]
+    # collect all the log files into one place for publishing in case of tox failure
+    test_envs = ["whl", "sdist", "depends", "latestdependency", "mindependency"]
```
scbedd, line 281: I kinda feel like we should do a surface level glob of the folders under the `.tox` folder, excluding `.tmp`. `glob.glob` totally works on py27 as long as we don't need recursion. In this case, we do not.

That way, if we ever add another environment, we won't have to remember to update this list.

scripts/devops_tasks/tox_harness.py
```diff
@@ -274,11 +274,66 @@ def replace_dev_reqs(file, pkg_root):
...
+        log_directory, package
+    )
+
+    try:
+        os.mkdir(log_directory)
+    except OSError:
+        logging.info("'{}' directory already exists".format(log_directory))
+
+    log_directory = os.path.join(
+        log_directory, sys.version.split()[0]
```
scbedd, line 302: So the final result log directory looks like:

~~`$(Build.SourcesDirectory)/_tox_logs/<packagename>/3/*.txt` If I understand this correctly.~~

This is totally fine. `sys.version.split()` returned something I did not expect.

scripts/devops_tasks/tox_harness.py
```diff
@@ -291,6 +346,8 @@ def execute_tox_serial(tox_command_tuples):
         if result is not None and result != 0:
             return_code = result
 
+        collect_log_files(cmd_tuple[1])
```
scbedd, line 349: This should be within `in_ci` I think.

scbedd (changes requested):
Couple functionality changes.

scripts/devops_tasks/tox_harness.py
scbedd: We don't currently move coverage files around etc unless that's the case at least. I could be talked down from that :)

scripts/devops_tasks/tox_harness.py
scbedd: Just as an aside. I would also deliberately include a failing assertion in a random test just to double check the artifact publishing. If you have already done that 👍 good by me.

seankane-msft:
I added a failing test to confirm that the [logs are uploaded properly](https://dev.azure.com/azure-sdk/public/_build/results?buildId=733617&view=results)

scbedd (approved):
LGTM. This will be really nice to use on a failure.

PR #16557, opened 2021-02-05
Fix replace keys for nested json

lsundaralingam:


sdk/communication/azure-communication-identity/tests/_shared/testcase.py
```diff
@@ -51,13 +51,12 @@ def process_response(self, response):
     
     def _replace_keys(self, body):
         def _replace_recursively(dictionary):
-            if type(dictionary) != dict:
```
beltr0n, line 54: i wonder if somehow using type instead of isinstance did something

sacheu (approved):
Thanks for the fix.

PR #16563, opened 2021-02-05
Coalesce all packages around tools/azure-devtools instead of pypi azure-devtools

seankane-msft:


eng/pipelines/templates/jobs/update_pr.yml
```diff
@@ -19,11 +19,13 @@ steps:
     versionSpec: 3.7
 
 - script: |
-    python3 -m pip install azure-devtools[ci_tools]>=1.1.0
+    python3 -m pip install -e "git+https://github.com/Azure/azure-sdk-for-python#subdirectory=tools/azure-devtools&egg=azure-devtools"
```
weshaggard, line 22: What branch does it end up pulling from in a case like this? We should try and ensure we pull from a consistent branch instead of always pulling from tip as that might cause issues with older builds in some cases.

seankane-msft: This will pull from master. I will talk with Scott B on this.

PR #16565, opened 2021-02-05
[Key Vault] Add local-only mode to CryptographyClient

### LINKED ISSUE #16619: Remove LocalCryptographyClient in favor of CryptographyClient local mode
Per [archboard feedback](https://github.com/Azure/azure-sdk/issues/2221) and subsequent discussions, remove (or internalize) the `LocalCryptographyClient` class in favor of making a local-only "mode" of the existing `CryptographyClient`. This should be done in a constructor or language equivalent, and take only a `JsonWebKey` and, optionally (only if need to), a `CryptographyClientOptions`. For example, .NET uses the Diagnostic options for tracing.

mccoyp:
Resolves #16619.

Aligns Python's CryptographyClient with other languages, in providing a local-only mode by means of a factory method.

### DIFF
--- a/sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_models.py
+++ b/sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_models.py
@@ -245,8 +245,7 @@ class KeyVaultKey(object):
 
     def __init__(self, key_id, jwk=None, **kwargs):
         # type: (str, Optional[dict], **Any) -> None
-        if not kwargs.pop("_local_only", False):
-            self._properties = kwargs.pop("properties", None) or KeyProperties(key_id, **kwargs)
+        self._properties = kwargs.pop("properties", None) or KeyProperties(key_id, **kwargs)
         if isinstance(jwk, dict):
             if any(field in kwargs for field in JsonWebKey._FIELDS):  # pylint:disable=protected-access
                 raise ValueError(
--- a/sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_models.py
+++ b/sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_models.py
@@ -104,23 +104,25 @@ class KeyProperties(object):
 
     @property
     def name(self):
-        # type: () -> str
+        # type: () -> Optional[str]
         """The key's name
 
         :rtype: str
         """
         if self._vault_id:
             return self._vault_id.name
+        return None
 
     @property
     def version(self):
-        # type: () -> str
+        # type: () -> Optional[str]
         """The key's version
 
         :rtype: str
         """
         if self._vault_id:
             return self._vault_id.version
+        return None
 
     @property
     def enabled(self):
@@ -169,13 +171,14 @@ class KeyProperties(object):
 
     @property
     def vault_url(self):
-        # type: () -> str
+        # type: () -> Optional[str]
         """URL of the vault containing the key
 
         :rtype: str
         """
         if self._vault_id:
             return self._vault_id.vault_url
+        return None
 
     @property
     def recoverable_days(self):
--- a/sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_models.py
+++ b/sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_models.py
@@ -68,12 +68,9 @@ class KeyProperties(object):
         # type: (str, Optional[_models.KeyAttributes], **Any) -> None
         self._attributes = attributes
         self._id = key_id
+        self._vault_id = parse_key_vault_id(key_id)
         self._managed = kwargs.get("managed", None)
         self._tags = kwargs.get("tags", None)
-        try:
-            self._vault_id = parse_key_vault_id(key_id)
-        except:
-            self._vault_id = None
 
     def __repr__(self):
         # type () -> str
@@ -104,25 +101,21 @@ class KeyProperties(object):
 
     @property
     def name(self):
-        # type: () -> Optional[str]
+        # type: () -> str
         """The key's name
 
         :rtype: str
         """
-        if self._vault_id:
-            return self._vault_id.name
-        return None
+        return self._vault_id.name
 
     @property
     def version(self):
-        # type: () -> Optional[str]
+        # type: () -> str
         """The key's version
 
         :rtype: str
         """
-        if self._vault_id:
-            return self._vault_id.version
-        return None
+        return self._vault_id.version
 
     @property
     def enabled(self):
@@ -171,14 +164,12 @@ class KeyProperties(object):
 
     @property
     def vault_url(self):
-        # type: () -> Optional[str]
+        # type: () -> str
         """URL of the vault containing the key
 
         :rtype: str
         """
-        if self._vault_id:
-            return self._vault_id.vault_url
-        return None
+        return self._vault_id.vault_url
 
     @property
     def recoverable_days(self):
@@ -254,7 +245,8 @@ class KeyVaultKey(object):
 
     def __init__(self, key_id, jwk=None, **kwargs):
         # type: (str, Optional[dict], **Any) -> None
-        self._properties = kwargs.pop("properties", None) or KeyProperties(key_id, **kwargs)
+        if not kwargs.pop("_local_only", False):
+            self._properties = kwargs.pop("properties", None) or KeyProperties(key_id, **kwargs)
         if isinstance(jwk, dict):
             if any(field in kwargs for field in JsonWebKey._FIELDS):  # pylint:disable=protected-access
                 raise ValueError(
--- a/sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_models.py
+++ b/sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_models.py
@@ -68,9 +68,12 @@ class KeyProperties(object):
         # type: (str, Optional[_models.KeyAttributes], **Any) -> None
         self._attributes = attributes
         self._id = key_id
-        self._vault_id = parse_key_vault_id(key_id)
         self._managed = kwargs.get("managed", None)
         self._tags = kwargs.get("tags", None)
+        try:
+            self._vault_id = parse_key_vault_id(key_id)
+        except:
+            self._vault_id = None
 
     def __repr__(self):
         # type () -> str
@@ -106,7 +109,8 @@ class KeyProperties(object):
 
         :rtype: str
         """
-        return self._vault_id.name
+        if self._vault_id:
+            return self._vault_id.name
 
     @property
     def version(self):
@@ -115,7 +119,8 @@ class KeyProperties(object):
 
         :rtype: str
         """
-        return self._vault_id.version
+        if self._vault_id:
+            return self._vault_id.version
 
     @property
     def enabled(self):
@@ -169,7 +174,8 @@ class KeyProperties(object):
 
         :rtype: str
         """
-        return self._vault_id.vault_url
+        if self._vault_id:
+            return self._vault_id.vault_url
 
     @property
     def recoverable_days(self):
--- a/sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/crypto/_client.py
+++ b/sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/crypto/_client.py
@@ -142,6 +142,13 @@ class CryptographyClient(KeyVaultClientBase):
         :param jwk: the key's cryptographic material, as a JsonWebKey or dictionary.
         :type jwk: JsonWebKey or dict
         :rtype: CryptographyClient
+
+        .. literalinclude:: ../tests/test_examples_crypto.py
+            :start-after: [START from_jwk]
+            :end-before: [END from_jwk]
+            :caption: Create a CryptographyClient from a JsonWebKey
+            :language: python
+            :dedent: 8
         """
         if isinstance(jwk, JsonWebKey):
             key = vars(jwk)
--- a/sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/crypto/_client.py
+++ b/sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/crypto/_client.py
@@ -108,8 +108,8 @@ class CryptographyClient(KeyVaultClientBase):
             self._key_id = parse_key_vault_id(key)
             self._keys_get_forbidden = None  # type: Optional[bool]
         elif self._jwk:
-            self._key_id = key.get("kid", "")
-            self._key = KeyVaultKey(None, jwk=key, _local_only=True)
+            self._key = key
+            self._key_id = key.kid
         else:
             raise ValueError("'key' must be a KeyVaultKey instance or a key ID string including a version")
 
@@ -126,6 +126,8 @@ class CryptographyClient(KeyVaultClientBase):
     def key_id(self):
         # type: () -> str
         """The full identifier of the client's key.
+
+        This property may be None when a client is constructed with `CryptographyClient.from_jwk`.
 
         :rtype: str
         """
@@ -150,9 +152,9 @@ class CryptographyClient(KeyVaultClientBase):
             :dedent: 8
         """
         if isinstance(jwk, JsonWebKey):
-            key = vars(jwk)
-        else:
             key = jwk
+        else:
+            key = JsonWebKey(**jwk)
         return cls(key, object(), _jwk=True)
 
     @distributed_trace
--- a/sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/crypto/_client.py
+++ b/sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/crypto/_client.py
@@ -128,7 +128,7 @@ class CryptographyClient(KeyVaultClientBase):
                 self._local_provider = get_local_cryptography_provider(self._key)
                 self._initialized = True
             except Exception as ex:  # pylint:disable=broad-except
-                raise ValueError("The provided jwk is not valid for local cryptography: {}".format(ex))
+                six.raise_from(ValueError("The provided jwk is not valid for local cryptography"), ex)
         else:
             self._local_provider = NoLocalCryptography()
             self._initialized = False
@@ -145,7 +145,7 @@ class CryptographyClient(KeyVaultClientBase):
 
         This property may be None when a client is constructed with :func:`from_jwk`.
 
-        :rtype: str
+        :rtype: str or None
         """
         if not self._jwk:
             return self._key_id.source_id
@@ -158,7 +158,7 @@ class CryptographyClient(KeyVaultClientBase):
 
         This property may be None when a client is constructed with :func:`from_jwk`.
 
-        :rtype: str
+        :rtype: str or None
         """
         return self._vault_url
 
@@ -170,13 +170,6 @@ class CryptographyClient(KeyVaultClientBase):
         :param jwk: the key's cryptographic material, as a JsonWebKey or dictionary.
         :type jwk: JsonWebKey or dict
         :rtype: CryptographyClient
-
-        .. literalinclude:: ../tests/test_examples_crypto.py
-            :start-after: [START from_jwk]
-            :end-before: [END from_jwk]
-            :caption: Create a CryptographyClient from a JsonWebKey
-            :language: python
-            :dedent: 8
         """
         if not isinstance(jwk, JsonWebKey):
             jwk = JsonWebKey(**jwk)
@@ -202,7 +195,7 @@ class CryptographyClient(KeyVaultClientBase):
 
         # if we have the key material, create a local crypto provider with it
         if self._key:
-            self._local_provider = get_local_cryptography_provider(self._key, _key_id=self.key_id)
+            self._local_provider = get_local_cryptography_provider(self._key)
             self._initialized = True
         else:
             # try to get the key again next time unless we know we're forbidden to do so
--- a/sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/crypto/_client.py
+++ b/sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/crypto/_client.py
@@ -102,7 +102,12 @@ class CryptographyClient(KeyVaultClientBase):
 
         if isinstance(key, KeyVaultKey):
             self._key = key
-            self._key_id = parse_key_vault_id(key.id)
+            try:
+                self._key_id = parse_key_vault_id(key.id)
+            except ValueError:
+                if not self._local_only:
+                    raise
+                self._key_id = None
         elif isinstance(key, six.string_types):
             self._key = None
             self._key_id = parse_key_vault_id(key)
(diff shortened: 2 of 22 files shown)


sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/crypto/_client.py
```diff
@@ -126,14 +127,24 @@ def key_id(self):
...
+
     @distributed_trace
     def _initialize(self, **kwargs):
         # type: (**Any) -> None
         if self._initialized:
             return
 
         # try to get the key material, if we don't have it and aren't forbidden to do so
-        if not (self._key or self._keys_get_forbidden):
+        if not (self._key or self._keys_get_forbidden or self._local_only):
```
mccoyp, line 147: Since key material is required to create a local-only client, this is theoretically redundant since self._key shouldn't be None. A user _could_ set self._key = None though, so I added this check just in case. It doesn't always make sense to account for the various ways users could interfere with the class though, so are "just in case" checks like this unnecessary entirely?

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/crypto/_client.py
```diff
@@ -114,6 +114,7 @@ def __init__(self, key, credential, **kwargs):
 
         self._local_provider = NoLocalCryptography()
         self._initialized = False
+        self._local_only = kwargs.pop("local_only", False)
```
chlowell, line 117: If you don't want this argument to be public, its name should make that clear:
```suggestion
        self._local_only = kwargs.pop("_local_only", False)
```

sdk/keyvault/azure-keyvault-keys/tests/test_crypto_client.py
```diff
@@ -491,6 +491,67 @@ def test_calls_service_for_operations_unsupported_locally():
...
+
+
+def test_local_only_mode_raise():
+    """A local-only CryptographyClient should raise an exception if an operation can't be performed locally"""
+
+    jwk = {"kty":"RSA", "key_ops":["decrypt", "verify", "unwrapKey"], "n":b"10011", "e":b"10001"}
+    client = CryptographyClient.from_jkw(key_id="https://localhost/fake/key/version", jwk=jwk)
+
+    # Algorithm not supported locally
+    with pytest.raises(NotImplementedError):
```
chlowell, line 536: Do you want to test some aspect of the error message, like that it includes the algorithm?

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/crypto/_client.py
chlowell: Yes. An underscored attribute is private, makes no compatibility promises.

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/crypto/_client.py
```diff
@@ -180,11 +191,13 @@ def encrypt(self, algorithm, plaintext, **kwargs):
         _validate_arguments(operation=KeyOperation.encrypt, algorithm=algorithm, iv=iv, aad=aad)
         self._initialize(**kwargs)
 
-        if self._local_provider.supports(KeyOperation.encrypt, algorithm):
+        if self._local_provider.supports(KeyOperation.encrypt, algorithm) or self._local_only:
```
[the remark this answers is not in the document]
mccoyp: That's a good point -- I was on the fence about this, since it is counterintuitive (and potentially bad, as you point out) to proceed with an operation that we know isn't supported. The intention was to raise the resulting error message -- and the minimal refactoring was appealing -- but I agree that it's probably better to raise the NotImplementedError here

sdk/keyvault/azure-keyvault-keys/tests/test_crypto_client.py
mccoyp: I had considered it and decided not to, but it doesn't seem too implementation-dependent to expect this kind of error to return the algorithm name. I'll add that check, thanks!

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/crypto/_client.py
```diff
@@ -185,7 +196,13 @@ def encrypt(self, algorithm, plaintext, **kwargs):
             try:
                 return self._local_provider.encrypt(algorithm, plaintext)
             except Exception as ex:  # pylint:disable=broad-except
+                if self._local_only:
+                    raise
```
chlowell, line 200: Should we log the warning first?

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/crypto/_client.py
```diff
@@ -126,6 +127,16 @@ def key_id(self):
...
         return self._key_id.source_id
 
+    @classmethod
+    def from_jwk(cls, jwk):
+        # type: (dict) -> CryptographyClient
+        """Creates a client that can only perform cryptographic operations locally.
+
+        :param dict jwk: the key's cryptographic material, as a dictionary.
+        """
+        key_id = "https://key-vault.vault.azure.net/keys/local-key"
```
chlowell, line 137: Using a real URL as a signal value makes me nervous. Also, I think we should try to use "kid" from the JWK, if it has one. That entails relaxing KeyProperties to return None for the properties it would normally parse from a Key Vault identifier, which seems okay to me, and doing so would also allow defaulting to None here 🤔

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/crypto/_client.py
```diff
@@ -126,6 +127,16 @@ def key_id(self):
...
+        return cls(KeyVaultKey(key_id, jwk), object, _local_only=True)
```
chlowell, line 138: Is there a reason to pass the type `object` rather than an instance of it?

sdk/keyvault/azure-keyvault-keys/tests/test_crypto_client.py
```diff
@@ -491,6 +491,82 @@ def test_calls_service_for_operations_unsupported_locally():
     assert supports_nothing.wrap_key.call_count == 0
 
 
+def test_local_only_mode_no_service_calls():
+    """A local-only CryptographyClient shouldn't call the service if an operation can't be performed locally"""
+
+    mock_client = mock.Mock()
+    jwk = mock.Mock(spec=JsonWebKey)
+    client = CryptographyClient.from_jwk(jwk=jwk)
```
chlowell, line 499: We expect no request, so I would use a mock transport that raises when the client tries to send e.g.
```py
mock.Mock(send=mock.Mock(side_effect=Exception("client shouldn't send a request"))
```

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/crypto/_client.py
```diff
@@ -126,6 +127,16 @@ def key_id(self):
         """
         return self._key_id.source_id
 
+    @classmethod
+    def from_jwk(cls, jwk):
+        # type: (dict) -> CryptographyClient
```
chlowell, line 132: `Union[azure.keyvault.keys.JsonWebKey, dict]`?

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/crypto/_client.py
mccoyp: I wasn't sure if logging before throwing would be redundant, but it doesn't hurt to do both

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/crypto/_client.py
mccoyp: None in particular -- probably makes more sense to pass an instance

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/crypto/_client.py
mccoyp: That's fair, I'd prefer not to use a valid URL either. But there's then also the question of how we initialize the `KeyVaultClientBase` without a vault URL 🤔

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/crypto/_client.py
```diff
@@ -98,33 +98,65 @@ class CryptographyClient(KeyVaultClientBase):
...
+            :caption: Create a CryptographyClient from a JsonWebKey
+            :language: python
+            :dedent: 8
+        """
+        if isinstance(jwk, JsonWebKey):
+            key = vars(jwk)
+        else:
+            key = jwk
+        key_id = key.get("kid")
+        return cls(KeyVaultKey(key_id, key), object(), _local_only=True)
```
chlowell, line 159: Please forgive me for launching another expedition into design space, but seeing the KeyVaultKey changes necessary for this to work makes me curious about alternatives. For example, what if `_local_only` were `_jwk`, and the client is responsible for initializing itself with the JWK?

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_models.py
```diff
@@ -101,21 +104,25 @@ def id(self):
 
     @property
     def name(self):
-        # type: () -> str
+        # type: () -> Optional[str]
```
chlowell, line 107: I worry this encourages users to write defensive code they will never need. We tell users this property could be `None` but they don't need to consider that, because it only occurs in our internal usage of this class, never through the public API (modulo the influence of bugs we would want to know about).

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/crypto/_client.py
mccoyp: That would make sense, and it would be great to avoid making changes to existing models. I think there's a question about how clean the abstraction is if we were to switch from `_local_only` to `_jwk` in the cases where we avoid making remote calls, considering it might be reasonable to want a crypto client from a JWK that still uses the service. If you wanted that then you could just create a client with a KeyVaultKey though, so I think it's reasonable to say that a client created with a JWK is inherently local-only

mccoyp: After tinkering with this, it might be too much work to have a CryptographyClient's key material as a JWK. The underlying providers would need to account for the key material coming from a JsonWebKey instead of a KeyVaultKey, and while it might be feasible, it's ultimately more work and more error-prone than creating a KeyVaultKey without an `id`

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_models.py
```diff
@@ -245,7 +245,8 @@ class KeyVaultKey(object):
 
     def __init__(self, key_id, jwk=None, **kwargs):
         # type: (str, Optional[dict], **Any) -> None
-        self._properties = kwargs.pop("properties", None) or KeyProperties(key_id, **kwargs)
+        if not kwargs.pop("_local_only", False):
```
mccoyp, line 248: I feel like `_local_only` suits the abstraction better than `_jwk` for this class since it already has a `jwk` parameter

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/crypto/_client.py
```diff
@@ -106,16 +107,20 @@ def __init__(self, key, credential, **kwargs):
             self._key = None
             self._key_id = parse_key_vault_id(key)
             self._keys_get_forbidden = None  # type: Optional[bool]
+        elif self._jwk:
+            self._key_id = key.get("kid", "")
```
mccoyp, line 111: By defaulting to an empty string, we can leave the `key_id` property return type as a string. This feels a little questionable though -- what do you think?

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/crypto/_client.py
```diff
@@ -124,7 +129,31 @@ def key_id(self):
...
+            :end-before: [END from_jwk]
+            :caption: Create a CryptographyClient from a JsonWebKey
+            :language: python
+            :dedent: 8
+        """
+        if isinstance(jwk, JsonWebKey):
+            key = vars(jwk)
+        else:
+            key = jwk
+        return cls(key, object(), _jwk=True)
```
mccoyp, line 156: This goes against the prescribed type for the key parameter, but I feel like that's reasonable for the implementation

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/crypto/_client.py
chlowell: My gut says let it be None, and document that it may be None when the client was constructed by `from_jwk`.

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/crypto/_client.py
```diff
@@ -106,16 +107,20 @@ def __init__(self, key, credential, **kwargs):
             self._key = None
             self._key_id = parse_key_vault_id(key)
             self._keys_get_forbidden = None  # type: Optional[bool]
+        elif self._jwk:
+            self._key_id = key.get("kid", "")
+            self._key = KeyVaultKey(None, jwk=key, _local_only=True)
```
chlowell, line 112: Everything we need for local crypto is in the JWK. Do we really need to jam that into KeyVaultKey?

mccoyp: All the local crypto providers and operations assume that they're working with a KeyVaultKey, so we'd have to do so unless we refactor them to accept a JsonWebKey unless I'm mistaken

mccoyp: I don't think it's unreasonable by any means to make local providers accept JWKs, but it seems like a design question of whether we do that (and preserve the current definition of a KeyVaultKey) or make a smaller tweak to accept KeyVaultKeys intended for local-only use

chlowell: Right, that's what I'm getting at. Their expectation of KeyVaultKey appears to be driving additional complexity here but that expectation isn't a requirement. Actually they only need the JWK. If they took that up front, maybe it wouldn't be necessary to complicate KeyVaultKey?

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/crypto/_providers/__init__.py
```diff
@@ -8,22 +8,28 @@
...
 
 if TYPE_CHECKING:
+    from typing import Union
     from ... import KeyVaultKey
 
 
 def get_local_cryptography_provider(key):
-    # type: (KeyVaultKey) -> LocalCryptographyProvider
-    if key.key_type in (KeyType.ec, KeyType.ec_hsm):
+    # type: (Union[JsonWebKey, KeyVaultKey]) -> LocalCryptographyProvider
```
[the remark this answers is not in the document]
mccoyp: I also thought about whether we should just use the JsonWebKey for local crypto, and the only snag I can think of is that a JsonWebKey's `kid` could be different from a KeyVaultKey's `id`. It seems unlikely that someone would intentionally assign those properties different values to some end, but it's not impossible

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/crypto/_key_validity.py
```diff
@@ -29,7 +30,7 @@ def dst(self, dt):
 
 
 def raise_if_time_invalid(key):
-    # type: (KeyVaultKey) -> None
+    # type: (Union[JsonWebKey, KeyVaultKey]) -> None
```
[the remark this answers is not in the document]
mccoyp: Should I throw a `# type: ignore` on calls that would upset mypy? I agree that it is silly to include JsonWebKey in there

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/crypto/_providers/__init__.py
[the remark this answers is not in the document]
mccoyp: This could probably be resolved by passing a keyword argument corresponding to KeyVaultKey.id like `_key_id` to get_local_cryptography_provider, and preferring that value over the JsonWebKey's `kid` in LocalCryptographyProvider.__init__

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/crypto/_client.py
```diff
@@ -98,33 +100,69 @@ class CryptographyClient(KeyVaultClientBase):
...
 
-        if not self._key_id.version:
+        if not (self._jwk or self._key_id.version):
             raise ValueError("'key' must include a version")
 
         self._local_provider = NoLocalCryptography()
         self._initialized = False
 
-        super(CryptographyClient, self).__init__(vault_url=self._key_id.vault_url, credential=credential, **kwargs)
+        vault_url = "vault_url" if self._jwk else self._key_id.vault_url
```
chlowell, line 129: I struggle to think up a great value for this. None seems to make the most sense but won't do for the base class. What do you think of overriding the `vault_url` property in this class to make it optional?

sdk/keyvault/azure-keyvault-keys/tests/test_examples_crypto.py
```diff
@@ -18,6 +18,18 @@
 )
 
 
+def test_create_client_from_jwk():
+    # [START from_jwk]
+    # create a CryptographyClient using a JsonWebKey instance
+    key = JsonWebKey(kty="RSA")
```
chlowell, line 24: If someone copies this, it will definitely not work 😜 

Which points to an interesting user experience problem. If I give CryptographyClient a garbage JWK, it won't complain until I ask it to perform an operation (and then it just says the key doesn't support that operation). This delayed error was the best we could do before, because it wasn't feasible to determine at construction what would be possible with a given key or key ID. That's no longer true in local-only mode though. The client could raise ValueError when given a key that it can't get a local provider for, for example.

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/crypto/_providers/local_provider.py
```diff
@@ -44,12 +45,12 @@ def supports(self, operation, algorithm):
...
     @property
     def key_id(self):
-        # type: () -> str
+        # type: () -> Optional[str]
         """The full identifier of the provider's key.
 
         :rtype: str
         """
-        return self._key.id
+        return self._key_id
```
chlowell, line 53: This will be either a valid Key Vault identifier or whatever the user gave us:
```suggestion
        return self._key.get("kid")
```
Does that remove the need to ask a caller to give us `_key_id` separately?

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/crypto/_client.py
```diff
@@ -98,33 +100,69 @@ class CryptographyClient(KeyVaultClientBase):
...
+        self._jwk = kwargs.pop("_jwk", False)
+        self._not_before = None  # type: Optional[datetime]
+        self._expires_on = None  # type: Optional[datetime]
+        self._key_id = None  # type: Optional[KeyVaultResourceId]
 
         if isinstance(key, KeyVaultKey):
-            self._key = key
+            self._key = key.key
             self._key_id = parse_key_vault_id(key.id)
+            if key.properties._attributes:  # pylint:disable=protected-access
```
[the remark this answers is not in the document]
mccoyp: Unless attributes are provided to KeyProperties these properties raise an AttributeError, since `KeyProperties._attributes` is None by default

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/crypto/_providers/local_provider.py
mccoyp: My thinking here is that a KeyVaultKey provided to CryptographyClient could have a valid Key Vault identifier for its `id`, but that the JsonWebKey could have a different `kid`. I assumed it made more sense to return the Key Vault identifier in the operation result if possible, which is why I added the `_key_id` keyword argument

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/crypto/_client.py
mccoyp: That would be nice, since None does make more sense. Would that entail providing a non-None value to the base class, but overriding the property with an optional value in CryptographyClient?

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/crypto/aio/_client.py
```diff
@@ -5,6 +5,7 @@
 import logging
 from typing import TYPE_CHECKING
 
+import six
```
chlowell, line 8: Unused

sdk/keyvault/azure-keyvault-keys/tests/test_crypto_client.py
```diff
@@ -276,6 +276,23 @@ def test_encrypt_local(self, azure_keyvault_url, **kwargs):
             result = crypto_client.decrypt(result.algorithm, result.ciphertext)
             self.assertEqual(result.plaintext, self.plaintext)
 
+    @KeyVaultPreparer()
+    def test_encrypt_local_from_jwk(self, azure_keyvault_url, **kwargs):
```
chlowell, line 280: It's also possible to test private operations locally, using an imported key for public operations (the neighboring `keys.py` has suitable keys). Doesn't need to be in this PR but it's a test gap to fill before the next stable release. Also, need to cover EC keys as well.

sdk/keyvault/azure-keyvault-keys/tests/recordings/test_crypto_client_async.test_wrap_local_from_jwk.yaml
```diff
@@ -0,0 +1,170 @@
...
+        resource="https://vault.azure.net"
+      x-content-type-options: nosniff
+      x-ms-keyvault-network-info: conn_type=Ipv4;addr=172.92.159.124;act_addr_fam=InterNetwork;
+      x-ms-keyvault-region: westus2
+      x-ms-keyvault-service-version: 1.2.191.0
+      x-powered-by: ASP.NET
+    status:
+      code: 401
+      message: Unauthorized
+    url: https://mcpatinokv.vault.azure.net/keys/livekvtestwrap-localfe02144b/create?api-version=7.2-preview
```
chlowell, line 37: Do you know why we've stopped replacing the vault name?

chlowell (approved):

sdk/keyvault/azure-keyvault-keys/tests/test_crypto_client.py
chlowell: (That is, if it would be a bug for the client to send a request in local-only mode.)

mccoyp: This is one that I've made a note of to revisit -- I'm not too familiar with Mock and wasn't able to randomly turn enough dials to get a better understanding. I figure that asserting that the underlying client doesn't call get_key is sufficient for the time being

Issue #16183, opened 2021-01-15
Resolve architect board feedback for Key Vault


heaths:
Here are a few things that came up during the [archboard review](https://github.com/Azure/azure-sdk/issues/2221). Note that there is a discussion about whether to expose a `LocalCryptographyClient` or make it a "mode" of `CryptographyClient`, but that should only apply to .NET and Java. That is not listed below as prototyping is necessary. I will open a new bug shortly.

- [x] ~Add service version for "7.2". Don't bother with "7.2-preview". It won't be supported anyway, most likely, and "7.2" already is supported by both Key Vault and Managed HSM (rather, Managed HSM ignores it).~ Changed to keeping "7.2-preview"
- [x] ~Rename `EncryptOptions` to `EncryptParameters`.~ ~Create Python equivalent: e.g. EncryptionAlgorithmConfiguration with factory methods (.rsa1_5_encrypt, etc.) or similar~ Not implementing this in Python for now
- [x] ~Rename `DecryptOptions` to `DecryptParameters`.~ ~Using above idea, add factories for decrypt: e.g. EncryptionAlgorithmConfiguration.rsa1_5_decrypt~ Not implementing this in Python for now
- [x] `AdditionalAuthenticatedData should be a read-only property (no setters).
- [x] `KeyVaultKeyIdentifier` (as well as for secrets and certificates) should not have "parse" methods, but instead take the URI into the constructor/creator. This pattern will, of course, vary from language to language.
- [x] Make sure all properties on `JsonWebKey` are settable (came up indirectly, but this should be the case already).

PR #16567, opened 2021-02-05
[Key Vault] Add EncryptParameters analog in Python

mccoyp:
Resolves part of #16183.

This adds an `EncryptionAlgorithmConfiguration` class that contains factory methods intended to lead users to discover which parameters are applicable for a given algorithm. Encrypt and decrypt methods accept this configuration via the `algorithm` argument.


sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/crypto/_models.py
```diff
@@ -8,6 +8,131 @@
...
+    @classmethod
+    def rsa_oaep_256_decrypt(cls):
+        return cls(EncryptionAlgorithm.rsa_oaep_256)
+
+    @classmethod
+    def rsa1_5_decrypt(cls):
+        return cls(EncryptionAlgorithm.rsa1_5)
+
+    @classmethod
+    def a128_gcm_decrypt(cls, authentication_tag=None, additional_authenticated_data=None):
```
chlowell, line 88: This gives me pause. The algorithm is the same regardless of whether you're encrypting or decrypting but the tag is an encryption product only used in decryption. Having two methods for the same algorithm is weird (doubly so when their signatures are identical) but combining them entails asking someone about to encrypt for an optional useless argument, which is contrary to our purpose. Having `_encrypt` and `_decrypt` methods for only some algorithms is also weird, as is adding a `DecryptionAlgorithmConfiguration` class. Well, that last thought isn't totally out there. But all this has me thinking we need some more ideas here 🤔

mccoyp: The last option would be most similar to what other languages are doing with `EncryptParameters` and `DecryptParameters`. I can't recall why I was opposed to having `EncryptionAlgorithmConfiguration` and `DecryptionAlgorithmConfiguration` -- maybe because they both take `EncryptionAlgorithm`s, and the latter obfuscates that -- but it's definitely one of the better options

PR #16569, opened 2021-02-05
replace recursion in replacer

beltr0n:
this pattern is more in-line with what i've seen in other azure sdks (e.g. form recognizer AI)
the recursion one is more generic/extensible so if that works I'd prefer we keep that instead


beltr0n:
Closing for now, we are using a different approach

PR #16577, opened 2021-02-05
Testing samples

swathipil:


scripts/devops_tasks/test_run_samples.py
```diff
@@ -25,7 +25,30 @@
...
+        "iot_hub_connection_string_receive_async.py",
+        "proxy_async.py",
+        "receive_batch_with_checkpoint_async.py",
+        "recv_async.py",
+        "recv_track_last_enqueued_event_prop_async.py",
+        "recv_with_checkpoint_by_event_count_async.py",
+        "recv_with_checkpoint_by_time_interval_async.py",
+        "recv_with_checkpoint_store_async.py",
+        "recv_with_custom_starting_position_async.py",
+        "sample_code_eventhub_async.py"
```
yunhaoling, line 50: I've added two extra sample for connection to custom endpoint address, we should ignore those two as well.

sdk/eventhub/azure-eventhub/samples/async_samples/client_identity_authentication_async.py
```diff
@@ -60,6 +60,7 @@ async def run():
                     # New EventDataBatch object can be created here to send more data.
                     break
             await producer.send_batch(event_data_batch)
+    await credential.close()
```
yunhaoling, line 63: we actually have `async with credential:` on line 48, I think this `close` is not needed?

swathipil: you're right, I thought I was getting an error but I guess I wasn't 🤦

yunhaoling (approved):
LGTM

sdk/eventhub/test-resources.json
```diff
@@ -0,0 +1,202 @@
...
+      "defaultValue": "10",
+      "metadata": {
+        "description": "The maximum duration, in minutes, that a single test is permitted to run before it is considered at-risk for being hung."
+      }
+    }
+  },
+  "variables": {
+    "ehVersion": "2017-04-01",
+    "contributorRoleId": "b24988ac-6180-42a0-ab88-20f7382dd24c",
+    "eventHubsDataOwnerRoleId": "f526a384-b230-433a-b45c-95f59c4a2dec",
```
yunhaoling, line 69: where do these two ids come from -- are they public constants?
(my concern is whether we're leaking kinda credential info here, tho they should just be randomly generated uuid?)

swathipil: yes here: https://docs.microsoft.com/en-us/azure/role-based-access-control/built-in-roles

yunhaoling: thanks!

yunhaoling (approved):
LGTM

PR #16579, opened 2021-02-05
Support matrix config import, and non sparse parameter arguments

benbp:
This PR adds support for config import and non-sparse parameters to job matrix generation. These features will enable a couple scenarios I have found while trying to convert Python live test configurations to use the job matrix generation scripts.

The import feature allows a matrix configuration to depend on another matrix configuration, for example if a live test wanted to run our base platform matrix, but add a set of additional test arguments to it. There are multiple places in the python sdk tests.yml files where we duplicate the platform matrix in order to make some minor modifications to it.

The non-sparse parameter feature allows us to generate sparse matrices while making sure that we still include all combinations of one or more parameters. This is also useful in conjunction with the import feature in order to enable some platform configurations that need to add multiple parameters to the base, but only need the full set of one parameter, otherwise the test matrix would be very large.

See the README changes for more in-depth details.

Example that will import a common sparse matrix and combine it with another sparse matrix, and then combine that with all values of `testFieldC`:
```
{
    "matrix": {
        "$IMPORT": "./test-import-matrix.json",
        "testFieldA": [ "test1A", "test2A" ],
        "testFieldB": [ "test1B", "test2B" ],
        "testFieldC": [ "test1C", "test2C", "test3C" ]
    }
}

Create-JobMatrix ./matrix.json -selection sparse -nonSparseParameters testFieldC
```

### DIFF
--- a/eng/scripts/job-matrix/Create-JobMatrix.ps1
+++ b/eng/scripts/job-matrix/Create-JobMatrix.ps1
@@ -20,7 +20,7 @@ param (
 
 $config = GetMatrixConfigFromJson (Get-Content $ConfigPath)
 # Strip empty string filters in order to be able to use azure pipelines yaml join()
-$Filters = $Filters | Where-Object { $_ -ne "" }
+$Filters = $Filters | Where-Object { $_ }
 
 [array]$matrix = GenerateMatrix `
     -config $config `
--- a/eng/scripts/job-matrix/Create-JobMatrix.ps1
+++ b/eng/scripts/job-matrix/Create-JobMatrix.ps1
@@ -12,14 +12,23 @@ param (
     [Parameter(Mandatory=$True)][string] $ConfigPath,
     [Parameter(Mandatory=$True)][string] $Selection,
     [Parameter(Mandatory=$False)][string] $DisplayNameFilter,
-    [Parameter(Mandatory=$False)][array] $Filters
+    [Parameter(Mandatory=$False)][array] $Filters,
+    [Parameter(Mandatory=$False)][array] $NonSparseParameters
 )
 
 . $PSScriptRoot/job-matrix-functions.ps1
 
 $config = GetMatrixConfigFromJson (Get-Content $ConfigPath)
+# Strip empty string filters in order to be able to use azure pipelines yaml join()
+$Filters = $Filters | Where-Object { $_ -ne "" }
 
-[array]$matrix = GenerateMatrix $config $Selection $DisplayNameFilter $Filters
+[array]$matrix = GenerateMatrix `
+    -config $config `
+    -selectFromMatrixType $Selection `
+    -displayNameFilter $DisplayNameFilter `
+    -filters $Filters `
+    -nonSparseParameters $NonSparseParameters
+
 $serialized = SerializePipelineMatrix $matrix
 
 Write-Output $serialized.pretty
--- a/eng/scripts/job-matrix/README.md
+++ b/eng/scripts/job-matrix/README.md
@@ -2,18 +2,20 @@
 
 * [Usage in a pipeline](#usage-in-a-pipeline)
 * [Matrix config file syntax](#matrix-config-file-syntax)
-   * [Fields](#fields)
-	  * [matrix](#matrix)
-	  * [include](#include)
-	  * [exclude](#exclude)
-	  * [displayNames](#displaynames)
+ * [Fields](#fields)
+	* [matrix](#matrix)
+	* [include](#include)
+	* [exclude](#exclude)
+	* [displayNames](#displaynames)
+	* [$IMPORT](#import)
 * [Matrix Generation behavior](#matrix-generation-behavior)
-  * [all](#all)
-  * [sparse](#sparse)
-  * [include/exclude](#includeexclude)
-  * [displayNames](#displaynames-1)
-  * [Filters](#filters)
-  * [Under the hood](#under-the-hood)
+	* [all](#all)
+	* [sparse](#sparse)
+	* [include/exclude](#includeexclude)
+	* [displayNames](#displaynames-1)
+	* [Filters](#filters)
+	* [NonSparseParameters](#nonsparseparameters)
+	* [Under the hood](#under-the-hood)
 * [Testing](#testing)
 
 
@@ -38,6 +40,7 @@ jobs:
       - Name: base_product_matrix
         Path: /eng/pipelines/matrix.json
         Selection: sparse
+		NonSparseParameters: <csv of parameter names for which all combinations should be included>
 		GenerateVMJobs: true
       - Name: sdk_specific_matrix
         Path: /sdk/foobar/matrix.json
@@ -144,6 +147,24 @@ readable value here. For example:
 }
 ```
 
+#### $IMPORT
+
+Matrix configs can also import another matrix config. The effect of this is the imported matrix will be generated,
+and then the importing config will be combined with that matrix (as if each entry of the imported matrix was a parameter).
+To import a matrix, add a parameter with the key `$IMPORT`:
+
+```
+"matrix": {
+  "$IMPORT": "path/to/matrix.json",
+  "JavaVersion": [ "1.8", "1.11" ]
+}
+```
+
+Importing can be useful, for example, in cases where there is a shared base matrix, but there is a need to run it
+once for each instance of a language version.
+
+Includes and excludes for the importing matrix will still get processed after the imported matrix is combined with the base matrix.
+
 ## Matrix Generation behavior
 
 #### all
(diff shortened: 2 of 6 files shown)


eng/scripts/job-matrix/Create-JobMatrix.ps1
```diff
@@ -12,14 +12,23 @@ param (
...
-    [Parameter(Mandatory=$False)][array] $Filters
+    [Parameter(Mandatory=$False)][array] $Filters,
+    [Parameter(Mandatory=$False)][array] $NonSparseParameters
 )
 
 . $PSScriptRoot/job-matrix-functions.ps1
 
 $config = GetMatrixConfigFromJson (Get-Content $ConfigPath)
+# Strip empty string filters in order to be able to use azure pipelines yaml join()
+$Filters = $Filters | Where-Object { $_ -ne "" }
```
weshaggard, line 23:
```suggestion
$Filters = $Filters | Where-Object { !$_ }
```

That should handle both empty and null cases.

eng/scripts/job-matrix/README.md
```diff
@@ -144,6 +147,24 @@ readable value here. For example:
...
+"matrix": {
+  "$IMPORT": "path/to/matrix.json",
+  "JavaVersion": [ "1.8", "1.11" ]
+}
+```
+
+Importing can be useful, for example, in cases where there is a shared base matrix, but there is a need to run it
+once for each instance of a language version.
+
+Includes and excludes for the importing matrix will still get processed after the imported matrix is combined with the base matrix.
```
weshaggard, line 166: I think I'm fine with this behavior but I thought we were going to apply exclude/includes on the imported matrix before we cross it with the matrix that is importing it. Or is this yet another case where I'm confused by the terminology :) Let me go see the code...

eng/scripts/job-matrix/job-matrix-functions.modification.tests.ps1
```diff
@@ -0,0 +1,151 @@
...
+        $matrix[4].name | Should -Be test12_test22_foo2_bar2
+        $matrix[4].parameters.testField1 | Should -Be "test12"
+        $matrix[4].parameters.testField2 | Should -Be "test22"
+        $matrix[4].parameters.Foo | Should -Be "foo2"
+    }
+
+    It "Should import a sparse matrix with import, include, and exclude" {
+        $matrixJson = @'
+{
+    "matrix": {
```
weshaggard, line 125: From a testing prospective and an understanding prospective it would be good to add a test where you you are using the import and compare it to one where you didn't do the import so we can understand how things would look if we didn't import it.

eng/scripts/job-matrix/job-matrix-functions.ps1
```diff
@@ -154,45 +185,83 @@ function ProcessExcludes([Array]$matrix, [Array]$excludes)
...
+    return $matrix, $importedMatrix
+}
+
+function CombineMatrices([Array]$matrix1, [Array]$matrix2)
+{
+    $combined = @()
+    if (-not $matrix1) {
+        return $matrix2
+    }
+    if (-not $matrix2) {
```
weshaggard, line 220: nit: I generally prefer using the less verbose `!` operator over `-not`, they are the same but `!` feels more natural to me.

eng/scripts/job-matrix/job-matrix-functions.ps1
```diff
@@ -154,45 +185,83 @@ function ProcessExcludes([Array]$matrix, [Array]$excludes)
...
+        foreach ($entry2 in $matrix2) {
+            $newEntry = @{
+                name = $entry1.name
+                parameters = CloneOrderedDictionary $entry1.parameters
+            }
+            foreach($param in $entry2.parameters.GetEnumerator()) {
+                if (-not $newEntry.Contains($param.Name)) {
+                    $newEntry.parameters[$param.Name] = $param.Value
+                } else {
+                    Write-Warning "Skipping duplicate parameter `"$($param.Name)`" when combining matrix."
```
weshaggard, line 234: I wonder if we should instead of skipping add the new values if any?

eng/scripts/job-matrix/job-matrix-functions.modification.tests.ps1
benbp: Great idea

eng/scripts/job-matrix/job-matrix-functions.ps1
benbp: This has the effect of favoring top level keys over imported keys, which I think is the right behavior. We can't have both.

benbp: Or are you saying include all combinations with both?

eng/scripts/job-matrix/README.md
benbp: Yeah I need to work on wording. Here the "importing" matrix is different from the "imported" matrix 😁

eng/scripts/job-matrix/job-matrix-functions.ps1
weshaggard: I was suggesting including all values from both sets.

eng/scripts/job-matrix/job-matrix-functions.modification.tests.ps1
```diff
@@ -119,6 +119,46 @@ Describe "Platform Matrix Import" -Tag "import" {
         $matrix[4].parameters.Foo | Should -Be "foo2"
     }
 
+    It "Should generate different matrices when importing and not importing" {
```
weshaggard, line 122: I was hoping to see another matrix entry here that is equal to the one doing the import so that we I could easily see what the full matrix would look like.

weshaggard:
```
{
  "matrix": {
    "Foo_Bar" [ ... ]
    "Foo_Bar_Baz" [ ...]
    "testField1": [ "test11", "test12" ]
  },
}
```

Something like that, or what it actually represents.

eng/scripts/job-matrix/job-matrix-functions.ps1
benbp: I think it makes more sense to have them combine as you suggest, though I'm inclined to do it as a follow-on change since the code might get complicated.

benbp:
@weshaggard I added import data structure examples to the README

eng/scripts/job-matrix/README.md
```diff
@@ -165,11 +165,124 @@ once for each instance of a language version.
...
+          "client": "netty"
+        }
+    ]
+}
+```
+
+1. The base matrix is generated (sparse in this example):
+    ```
+    {
+      "storage_18": {
```
weshaggard, line 196: Where does storage come from in your example?

eng/scripts/job-matrix/README.md
```diff
@@ -165,11 +165,124 @@ once for each instance of a language version.
...
 
-1. The base matrix is generated
-1. The imported base matrix is generated
+Given a matrix and import matrix like below:
+```
+{
+    "matrix": {
+        "$IMPORT": "example-matrix.json",
+        "JavaVersion": [ "1.8", "1.11" ],
+        "TestFromSource": [ false, true ]
```
weshaggard, line 174: Based on your walk through I assume this should be service as "storage"d and "cosmos"

eng/scripts/job-matrix/README.md
```diff
@@ -165,11 +165,124 @@ once for each instance of a language version.
...
+        "JavaVersion": "1.11",
+        "operatingSystem": "mac",
+        "client": "netty"
+      }
+    }
+    ```
+1. Includes/excludes from the top-level matrix get applied to the multiplied matrix, so the below element will be added
+   to the above matrix, for an output matrix with 7 elements:
+    ```
+    "windows_18_TestFromSource": {
```
weshaggard, line 280: I also don't see this include in your sample matrix.

eng/scripts/job-matrix/README.md
benbp: Yeah these values are just pseudo-real

eng/scripts/job-matrix/README.md
benbp: Ah whoops, forgot to fully copy the json I was testing against back into the markdown...

benbp: fixed both readme issues

eng/scripts/job-matrix/README.md
benbp: fixed

weshaggard (approved):

PR #16583, opened 2021-02-05
[Communication] - Simplify identifier json models

sacheu:


jbeauregardb (approved):

beltr0n (approved):

sdk/communication/azure-communication-chat/tests/_shared/test_communication_identifier_serializer.py
```diff
@@ -6,92 +6,86 @@
 import unittest
```
beltr0n, line 6: (not related to your fixes)
is there a reason this uses unittest instead of pytest?

sacheu: I don't know why we use unittest instead of pytest. I need to look into this.

PR #16559, opened 2021-02-05
[EventHubs] update vendor storage blob code

yunhaoling:
trying to resolve issues: https://github.com/Azure/azure-sdk-for-python/issues/16192
use the storage-blob version v12.7.1: https://github.com/Azure/azure-sdk-for-python/tree/azure-storage-blob_12.7.1, commit id: e8edeb85bcb9f92cb3c6299b893eb99be55bdd51

PR #16596, opened 2021-02-08
[Storage Blob] Use relative import in storage blob shared access signature module

yunhaoling:
This is an improvement I found that could be made when I was trying to update the azure-storage-blob vendor code in eventhub-checkpointstore(-aio): https://github.com/Azure/azure-sdk-for-python/pull/16559

To vendor the storage blob code, I have to use the relative important instead of `azure.storage.blob.xxx` otherwise the module could not be found.

I found most of the storage blob codes are using relative import, I think this part we could also relative important or is there a reason why we don't use relative import here?


annatisch:
Looks okay to me - will defer to @tasherif-msft and @xiafu-msft

xiafu-msft (approved):

PR #16604, opened 2021-02-08
added contains to DictMixin

mohamedshabanofficial:
where to add unit tests in the project?
i mean like almost all tests are integration tests


iscai-msft:
@mohamedshabanofficial we don't have a separate folder for unittests, we just add them in the same folder. This is def something we can look into changing, but for now I would jsut add them in the same folder. Example unit tests [here](https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/textanalytics/azure-ai-textanalytics/tests/test_repr.py)

iscai-msft:
and @mohamshabanofficial this looks good! I'm going to hold off on merging any of these code-changing TA PRs until after our release on Tuesday. Thanks!

iscai-msft (approved):

Issue #14841, opened 2020-10-28
Meaningful error message when sending empty batch to eventhub


arache:
When sending empty batch to eventhub, the SDK just raised the error like below:
```
azure.eventhub.exceptions.EventDataSendError: Message sender failed to add message data to outgoing queue.
ErrorCodes.ClientError: Message sender failed to add message data to outgoing queue.

```
I believe a more clear error message would be helpful here.

KieranBrantnerMagee:
Hey @arache, thanks for reaching out.

Totally reasonable ask, and I find myself wanting the clarity as much as you. (I believe we even fixed this comparable specific issue in ServiceBus just recently, so more than glad to pull this in.)

Expect to see a fix in during our next EH release, this issue will track the PR as it is produced and checked in.

yunhaoling:
hey @arache , sorry for keeping you waiting.

I did a cross-language check and think this actually should be a no-op instead of raising error when sending empty event data batch.

I had a PR out for this: https://github.com/Azure/azure-sdk-for-python/pull/16638, let me know if you have concern on the change.

yunhaoling:
hey @arache, we decided to make sending empty batch being a no-op instead of raising error, we will be having a preview release next week which you could try out.

I'm closing this issue now, don't hesitate to reach out to us or re-open the issue if you have any other questions!

arache:
Thanks @yunhaoling , nice fix.

PR #16638, opened 2021-02-09
[EventHubs] no-op instead of raising error when sending out empty event data batch

yunhaoling:
addressing: https://github.com/Azure/azure-sdk-for-python/issues/14841

- doing no-op instead of raising error when sending empty event batch
  - this aligns with event hubs in other languages and service bus in python
- This should be a non-breaking change.

references:

JS EH:
https://github.com/Azure/azure-sdk-for-js/blob/fc443f9b35c1cce112f9862685c07bed2bbae854/sdk/eventhub/event-hubs/src/eventHubSender.ts#L304-L316

.NET EH:
https://github.com/Azure/azure-sdk-for-net/blob/b493f7ad5487fd0d01f004e9f7ebdf50bfec0b96/sdk/eventhub/Azure.Messaging.EventHubs/src/Producer/EventHubProducerClient.cs#L573-L576

JAVA EH: can not find the code :(

### DIFF
--- a/sdk/eventhub/azure-eventhub/CHANGELOG.md
+++ b/sdk/eventhub/azure-eventhub/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## 5.3.1 (Unreleased)
 
+**Bug fixes**
+
+- Sending emtpy `event_data_batch` will be a no-op now instead of raising error.
 
 ## 5.3.0 (2021-02-08)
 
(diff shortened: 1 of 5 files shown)


sdk/eventhub/azure-eventhub/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## 5.3.1 (Unreleased)
 
+**Bug fixes**
+
+- Sending emtpy `event_data_batch` will be a no-op now instead of raising error.
```
swathipil, line 7: small typo:
```suggestion
- Sending empty `event_data_batch` will be a no-op now instead of raising error.
```

swathipil (approved):
lgtm

PR #16648, opened 2021-02-09
Bump msrest

seankane-msft:
previous bump didn't work for python 2.7, this takes care of that.


iscai-msft (approved):
looks good. once you regenerate with autorest htough, you'll need to bump the version to `21`

PR #16657, opened 2021-02-10
Skip dumping package proprties to file in test phase

praveenkuttappan:
Common script  to extract package properties and dump it into file is failing on pypy3 for azure communication pipeline. This PR is to unblock release by skipping this step from test phase. Common change will be submitted to azure-sdk-tools repo in another PR after unblocking python release.

### DIFF
--- a/eng/common/pipelines/templates/steps/daily-dev-build-variable.yml
+++ b/eng/common/pipelines/templates/steps/daily-dev-build-variable.yml
@@ -2,8 +2,10 @@
 # is used when this pipeline is going to be generating and publishing daily dev builds.
 parameters:
   ServiceDirectory: ''
+  DumpPackageProperties: true
+
 steps:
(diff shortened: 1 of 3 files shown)


eng/common/pipelines/templates/steps/daily-dev-build-variable.yml
```diff
@@ -2,8 +2,10 @@
 # is used when this pipeline is going to be generating and publishing daily dev builds.
 parameters:
   ServiceDirectory: ''
+  DumpPackageProperties: true
```
weshaggard, line 5: This feels like the wrong switch. Why does this fail in test legs?

praveenkuttappan:
Closing this PR. Issue was transient and reattempt has succeeded. I will check and add a flag in common script to ensure this is triggered only during build step.

PR #16667, opened 2021-02-10
Verify autorest in pipeline

seankane-msft:
#13724


eng/pipelines/templates/steps/verify-autorest.yml
```diff
@@ -0,0 +1,40 @@
...
+
+steps:
+  - task: UsePythonVersion@0
+    inputs:
+      versionSpec: '$(PythonVersion)'
+
+  - task: NodeTool@0
+    displayName: 'Install Node.js'
+    inputs:
+      versionSpec: '$(NodeVersion_'
```
scbedd, line 24: not accessing variable properly 

(sorry I know it's draft but it jumped out at me lol)

eng/pipelines/templates/steps/verify-autorest.yml
```diff
@@ -0,0 +1,40 @@
...
+  - script: |
+      pip install -r eng/autorest_req.txt
+      git clone $(auto_rest_clone_url)
+      cd autorest.python
+      git checkout $(repo_branch)
+      npm install
+    displayName: 'Prepare Environment'
+
+  - script: |
+      python scripts/devops_tasks/verify_autorest.py service_directory ${{ parameters.ServiceDirectory }}
```
scbedd, line 40: Invoke this with a python script task

eng/pipelines/templates/steps/verify-autorest.yml
```diff
@@ -0,0 +1,66 @@
...
+    displayName: "Install autorest"
+
+  - script: |
+      python --version
+      python -m pip install pip==20.1
+      pip install -r eng/autorest_req.txt
+      git clone ${{ parameters.auto_rest_clone_url }}
+      cd autorest.python
+      npm install
+      cd ..
```
scbedd, line 42: nit: You can drop the `cd ..`. context is not saved when you transition to a new step. It will always default to `$(System.DefaultWorkingDirectory)`.

eng/pipelines/templates/steps/verify-autorest.yml
```diff
@@ -0,0 +1,66 @@
...
+    displayName: 'Prepare Environment'
+
+  - pwsh: |
+      Get-ChildItem ./
+    displayName: 'List contents'
+    condition: always()
```
scbedd, line 48:
```suggestion
    condition: succeededOrFailed()
```

eng/pipelines/templates/steps/verify-autorest.yml
```diff
@@ -0,0 +1,66 @@
...
+
+  - task: PythonScript@0
+    displayName: 'Run autorest Python'
+    inputs:
+      scriptPath: 'scripts/devops_tasks/verify_autorest.py'
+      arguments: >-
+        --service_directory="${{ parameters.ServiceDirectory }}"
+
+  - pwsh: |
```
scbedd, line 57: Feel free to rip this entire step out. We get a flattened agent with every new run, so you don't need to spend time cleaning up.

sdk/containerregistry/ci.yml
```diff
@@ -40,4 +40,5 @@ extends:
     - name: azure-mgmt-containerregistry
       safeName: azuremgmtcontainerregistry
     - name: azure-containerregistry
-      safeName: azurecontainerregistry
\ No newline at end of file
+      safeName: azurecontainerregistry
+    VerifyAutorest: false
```
scbedd, line 44: In these the udpated ci.yml, `VerifyAutorest` is set to `false`. Are these ones that you KNOW will have issues, so you're kinda pushing it off? That's totally fine, but what about the rest of the repo? Have you run `VerifyAutorest` against all the service directories?

Trying to get a feel for the impact if we check this in. 

Also, have you let the mgmt folks know about this new check?

scripts/devops_tasks/verify_autorest.py
```diff
@@ -0,0 +1,82 @@
...
+    swagger_folders = find_swagger_folders(service_dir)
+
+    for working_dir in swagger_folders:
+        os.chdir(working_dir)
+        f = os.path.join(working_dir, "README.md")
+        if os.path.exists(f):
+            reset_command = ["autorest", "--reset"]
+            run_check_call(reset_command, root_dir)
+
+            command = ["autorest", "--python", os.path.join(working_dir, "README.md"), "--verbose"]
```
scbedd, line 38:
```suggestion
            command = ["autorest", "--python", f, "--verbose"]
```

eng/pipelines/templates/steps/verify-autorest.yml
```diff
@@ -0,0 +1,66 @@
...
+    type: string
+    default: 'client'
+  - name: ServiceDirectory
+    type: string
+  - name: NodeVersion
+    type: string
+    default: '12.x'
+  - name: PythonVersion
+    type: string
+    default: '3.9'
```
scbedd, line 15: We have a python version default available in globals.yml. Right now the default we're packages n stuff with is `3.6`. Feel free to default to `'$(PythonVersion)'` here. Just so it's cohesive. If it makes a huge difference for you, that may also mean bumping the default that we use to 3.9.

eng/pipelines/templates/steps/verify-autorest.yml
```diff
@@ -0,0 +1,66 @@
...
+      versionSpec: ${{ parameters.NodeVersion }}
+
+  - script: |
+      npm install -g autorest
+      autorest --help
+    displayName: "Install autorest"
+
+  - script: |
+      python --version
+      python -m pip install pip==20.1
```
scbedd, line 37: We have pip pinned in `ci_tools.txt`. It's 20.2 there. I understand why you have it separate here, as you don't need every single package that is in `ci_tools.txt`.

eng/pipelines/templates/steps/verify-autorest.yml
```diff
@@ -0,0 +1,66 @@
...
+      pip install -r eng/autorest_req.txt
+      git clone ${{ parameters.auto_rest_clone_url }}
```
scbedd, line 39: nit, as I don't believe autorest repo is that big. but can we use a sparse checkout?

```
  - template: /eng/common/pipelines/templates/steps/sparse-checkout.yml
    parameters:
      Paths:
        - "/"
      Repositories:
        - Name: "Azure/autorest.python"
          Commitish: "master"
          WorkingDirectory: "$(Build.SourcesDirectory)/autorest.python"
      SkipDefaultCheckout: true
```

I'm like 99% certain this will work. This will basically grab zero history for the commitsh you ask for. I'm adding path "/" to tell it to grab every file in the repo.

scbedd (changes requested):
Looks super solid Sean. Some pointers and nitpicks, but overall 👍

eng/pipelines/templates/steps/verify-autorest.yml
scbedd: I don't believe that will negatively affect `build-artifacts.yml` which is really the only piece I'm super concerned with.

eng/pipelines/templates/steps/verify-autorest.yml
scbedd: I'm testing this locally to confirm that it's good advice. Feel free to ignore this comment for now.

scbedd: Ok I tested this locally.

This will not work for what I'm asking you to do until I update the `sparse-checkout.yml` to NOT grab the folder `eng` by default. It would work if the pattern was simply:

```
        git clone --no-checkout --filter=tree:0 git://github.com/Azure/autorest.python .
        git sparse-checkout init
        git sparse-checkout add /
        git checkout <targetcommit> (master in your case)
```

But because by default it has a 

```
       git sparse-checkout set eng
```

Run before it adds the path, it actually won't work for git repos without an `eng` folder. It just quietly doesn't checkout properly.

sdk/containerregistry/ci.yml
seankane-msft: I know this will cause issues with containerregistry and storage from testing it on both. I will reach out to the storage team to see if they want me to default it on. I have not let mgmt team know about it, but will reach out.

eng/pipelines/templates/steps/verify-autorest.yml
scbedd: [PR that will enable this](https://github.com/Azure/azure-sdk-for-python/pull/16667)

benbp: @scbedd I would set the path to `/*` instead of `/` which will work with `eng`. This is what git sparse-checkout defaults to using anyway before we run `git sparse-checkout set`

scbedd: Ah ok. I'll definitely give it a shot locally.

scbedd: @seankane-msft this works. I updated the comment with the example. I believe it should be a drop in replacement.

scbedd (approved):
I have a feeling this will fail more than a few. If you're ok with dealing with the fallout, I won't block on this. all the requested code changes are done :)

seankane-msft:
> I have a feeling this will fail more than a few. If you're ok with dealing with the fallout, I won't block on this. all the requested code changes are done :)

I need to revisit this @scbedd and make sure it's only target appconfig for now. Then I will merge. I don't want to disrupt anyone other myself 😃

PR #16670, opened 2021-02-10
Refactor live test configs to use matrix generation, standardized yaml

benbp:
This PR refactors the live test configurations to be consistent with the config model we are rolling out across the repositories (better multi-cloud handling, container handling, less duplication, etc.). It also introduces the usage of dynamically generated platform matrices, instead of the duplicated/embedded matrix definitions that currently exist across the tests.yml files. See the [matrix README](https://github.com/Azure/azure-sdk-for-python/blob/master/eng/scripts/job-matrix/README.md) for more details.

There are a few special cases in python where the service live test configs do their own stage definitions. I've simplified these but the experience should be mostly the same. For formrecognizer, the new way to select canary and/or prod tests in the UI, is to click "stages to run" and make your selection, e.g.

image


sdk/formrecognizer/tests.yml
```diff
@@ -1,104 +1,46 @@
...
+      TestTimeoutInMinutes: 200
+      TestSamples: true
+      Clouds: 'Prod,Canary'
+      # This is a specific request from the formrecognizer service team
+      # their claim is that the full matrix ends up stress-testing their service.
+      # As such, the canary test runs should run on a reduced matrix.
+      CloudConfig:
+        Prod:
+          SubscriptionConfiguration: $(sub-config-azure-cloud-test-resources)
+          Location: 'westus2'
```
kristapratico, line 18: Can we use the pipeline variable here? For form recognizer, we switch the prod region quite frequently based on service team recommendation and it's nice to have this configurable when triggering a pipeline.

```suggestion
      Location: $(LOCATION)
```

benbp: Ah looks like I missed this due to rebase timing. Will update.

sdk/monitor/tests.yml
```diff
@@ -12,19 +12,8 @@ stages:
...
-          PythonVersion: '3.7'
-          CoverageArg: '--disablecov'
-        Linux_Python39:
-          Pool: azsdk-pool-mms-ubuntu-1804-general
-          OSVmImage: MMSUbuntu18.04
-          PythonVersion: '3.9'
-          CoverageArg: ''
+      MatrixConfigs:
+        - Name: Monitor_live_test
+          Path: sdk/monitor/platform-matrix.json
```
weshaggard, line 17: Do we know if there is a good reason the monitor has a unique matrix?

sdk/eventhub/tests.yml
```diff
@@ -5,27 +5,8 @@ stages:
...
-          OSVmImage: MMS2019
-          PythonVersion: '2.7'
-          CoverageArg: '--disablecov'
-        MacOs_Python37:
-          Pool: Azure Pipelines
-          OSVmImage: 'macOS-10.15'
-          PythonVersion: '3.7'
-          CoverageArg: '--disablecov'
+      MatrixFilters:
+        - PythonVersion=^(?!pypy3).*
```
weshaggard, line 9: This seems like a pretty common filter. That matrix entry seems like we might want to flip it to be opt-in instead of opt-out.

benbp: There are three services excluding it, and eighteen using the default that includes it.

sdk/monitor/tests.yml
benbp: The matrix they set up is slightly different in how the python versions match up against the operating systems. I was aiming for full parity on a first pass. We could import from the base matrix with filters/excludes/overrides, but it wouldn't be exactly the same. Might be worth a conversation about what their matrix needs are and whether we can get them more in line with the base.

benbp: @rakshith91 gave some input on this, so I'll update the matrix to only exclude 2.7 versions. I'm still thinking about how to get this from the base matrix without excluding macos. This might a reason in favor of adding a mode for "import and combine sparse" in addition to the existing "import and combine full."

sdk/formrecognizer/tests.yml
kristapratico: Thank you!

kristapratico (approved):
form recognizer changes LGTM

sdk/monitor/tests.yml
rakshith91: @benbp On a second thought, do add python 2.7 as well to the test matrix - i'll create a separate PR that excludes only monitor exporter and not other monitor related packages from running 2.7

sdk/keyvault/tests.yml
```diff
@@ -1,7 +1,7 @@
 trigger: none
 
 stages:
-  - template: ../../eng/pipelines/templates/stages/archetype-sdk-tests.yml
+  - template: /eng/pipelines/templates/stages/archetype-sdk-tests.yml
```
mccoyp, line 4: Key Vault uses templates for each of its packages (keys, secrets, and certificates) -- would this change, and the removal of `EnvVars`, need to be made for each of them?

benbp: Ack, looks like this file is incomplete, must have missed a couple changes there. Good catch. I'll normalize all the paths and add the env vars back to the certificates tests.

mccoyp (approved):
Changes look good for Key Vault; thanks so much!

sdk/monitor/tests.yml
benbp: @rakshith91 ok great. In that case I'll just have it run the base matrix then I think? That would be:

image

benbp:
Sorry for the rebase/force-push spam.

weshaggard (approved):

PR #16684, opened 2021-02-10
Ensure errors during test cases cause failure

chlowell:
If a test on 3.x raises ImportError or SyntaxError due to a bug, AbstractPreparer swallows the exception, causing the test to succeed, perhaps with a warning about an unscheduled coroutine. That is to say, it makes failure look like success 😬

### DIFF
--- a/tools/azure-devtools/src/azure_devtools/scenario_tests/preparers.py
+++ b/tools/azure-devtools/src/azure_devtools/scenario_tests/preparers.py
@@ -145,13 +145,14 @@ You must specify use_cache=True in the preparer decorator""".format(test_class_i
             try:
                 try:
                     import asyncio
+                except ImportError:
+                    fn(test_class_instance, **trimmed_kwargs)
+                else:
                     if asyncio.iscoroutinefunction(fn):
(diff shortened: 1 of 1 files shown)


tools/azure-devtools/src/azure_devtools/scenario_tests/preparers.py
```diff
@@ -145,13 +145,14 @@ def _preparer_wrapper(test_class_instance, **kwargs):
             try:
                 try:
                     import asyncio
+                except ImportError:
```
chlowell, line 148: I don't understand the prior comment--I'm thinking we don't need to handle `SyntaxError` because shouldn't this always be `ImportError` on 2.7 (modulo someone's custom `asyncio` module)?

seankane-msft: I concur

seankane-msft (approved):
:shipit:

mccoyp (approved):
I've verified with live local tests that this un-hides errors appropriately 👍

PR #16694, opened 2021-02-10
Document CertificateCredential requires an RSA private key

chlowell:
Mypy is warning about `AadClientCertificate.sign()` because it uses a private key returned by cryptography's `load_pem_private_key()` for signing, and that method returns one of several possible types all of whose `sign` methods take different arguments. So those are legitimate warnings, thank you mypy, that call does indeed raise e.g. `TypeError: sign() takes 3 positional arguments but 4 were given` when a user passes in a non-RSA private key.

However, no user should do that because Azure AD expects clients to sign JWT assertions with RS256, which is to say the client's certificate must have an RSA private key and `load_pem_private_key()` must have returned `RSAPrivateKey`, or the client can't authenticate anyway, even if it could sign an assertion. The cryptic `TypeError` won't lead a user with the wrong cert back onto the right path, so this PR has CertificateCredential instead raise `ValueError` with a clear message, documents its requirement for an RSA private key, and makes the situation clear to mypy with a cast.


sdk/identity/azure-identity/azure/identity/_internal/aadclient_certificate.py
```diff
@@ -19,14 +20,18 @@
...
-    :param bytes pem_bytes: bytes of a a PEM-encoded certificate including the private key
+    :param bytes pem_bytes: bytes of a a PEM-encoded certificate including the (RSA) private key
     :param bytes password: (optional) the certificate's password
     """
     def __init__(self, pem_bytes, password=None):
         # type: (bytes, Optional[bytes]) -> None
+        private_key = serialization.load_pem_private_key(pem_bytes, password=password, backend=default_backend())
+        if not isinstance(private_key, RSAPrivateKey):
+            raise ValueError("CertificateCredential requires an RSA private key because it uses RS256 for signing")
+        self._private_key = cast(RSAPrivateKey, private_key)
```
mccoyp, line 31: I'm assuming this cast is to appease mypy?

mccoyp (approved):

sdk/identity/azure-identity/azure/identity/_internal/aadclient_certificate.py
chlowell: I don't actually need the cast, it's a vestige of my first attempt, which mypy couldn't figure out:
```py
self._private_key = serialization.load_pem_private_key(pem_bytes, password=password, backend=default_backend())
if not isinstance(self._private_key, RSAPrivateKey):
    raise ValueError("CertificateCredential requires an RSA private key because it uses RS256 for signing")
```
I suppose because the assignment came before the `isinstance` check?

PR #16701, opened 2021-02-11
Adjust Injected Requirement Generation

scbedd:
We produce builds that are equivalent to <base_version>aYYYYMMDDN, actually require that when we take a dep on an unpublished package

Example. azure-storage-file-datalake (in a PR) is taking a dep on a new version of azure-storage-blob. `azure-storage-blob>=12.8.0b1`

The issue is that when we produce daily packages, we publish artifacts with versioning that looks like `12.8.0a20210210003` and the like.

Our dynamic `req` replacer that resolves this for the test environments that _don't_ install dev_requirements had a minor bug! Before this change, we would be replacing the requirement for the `alpha` package with something like the following `azure-storage-blob<13.0.0,>=12.8.0b1a`

Notice the fact that `a` is just added on? Yep. Bad. Needs to _replace_ the `pre` titling before that.

### DIFF
--- a/eng/tox/sanitize_setup.py
+++ b/eng/tox/sanitize_setup.py
@@ -92,7 +92,7 @@ def process_requires(setup_py_path):
 
         if not is_required_version_on_pypi(pkg_name, spec):
             old_req = str(req)
-            version = get_version(pkg_name)
+            version = get_version(pkg_name).base_version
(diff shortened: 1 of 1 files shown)


tasherif-msft (approved):

praveenkuttappan (approved):
Looks good

PR #16715, opened 2021-02-11
[formrecognizer] Clean up tests

catalinaperalta:
Cleaning up redundant authentication key successful tests.


catalinaperalta:
Leaving this as draft for now meanwhile I keep looking through tests.

sdk/formrecognizer/azure-ai-formrecognizer/tests/test_business_card.py
```diff
@@ -418,6 +364,40 @@ def test_business_card_jpg_include_field_elements(self, client):
         for name, field in business_card.fields.items():
             for f in field.value:
                 self.assertFieldElementsHasValues(f.value_data.field_elements, business_card.page_range.first_page_number)
+
```
catalinaperalta, line 367: @kristapratico let me know what you're opinion on this is, I see pros and cons to combining these jpg tests into one. 
I will say that at the end of the day the test_business_card_jpg_include_field_elements already tests that a jpg can be passed to the endpoint successfully. However, I also think that the checks on the returned model are important which is why I've combined those here.

iscai-msft (approved):

sdk/formrecognizer/azure-ai-formrecognizer/tests/test_receipt_from_url.py
```diff
@@ -195,7 +180,6 @@ def test_receipt_url_jpg(self, client):
         receipt_type = receipt.fields.get("ReceiptType")
         self.assertIsNotNone(receipt_type.confidence)
         self.assertEqual(receipt_type.value, 'Itemized')
-        self.assertReceiptItemsHasValues(receipt.fields["Items"].value, receipt.page_range.first_page_number, False)
```
kristapratico, line 198: I would keep this line, but delete assertFormPagesHasValues 4 lines above since we already call that directly after the poller result (same in async file)

kristapratico (commented):
Looks good, thank you! I would just delete all the recordings for the tests that we're consolidating since we don't need them anymore :)

sdk/formrecognizer/azure-ai-formrecognizer/tests/test_receipt_from_url.py
catalinaperalta: Done!

PR #16724, opened 2021-02-12
Trigger scripts

seankane-msft:
Per @mikeharder 's recommendation, we should trigger some pipelines on changes to the `scripts` directory to validate changes, `appconfiguration` is our quickest _real_ library and I included `template` at mike's suggestion.


mikeharder (changes requested):
What is the value in running `template` in addition to `appconfiguration`?

scbedd (approved):
I won't block, but would prefer just template.

PR #16735, opened 2021-02-13
[Communication] Generate identifier Models from new swagger

sacheu:


RezaJooyandeh (approved):

sdk/communication/azure-communication-chat/azure/communication/chat/_shared/communication_identifier_serializer.py
```diff
@@ -4,15 +4,17 @@
 # license information.
 # --------------------------------------------------------------------------
 
-from .models import (
+from .._generated.models import (
```
RezaJooyandeh, line 7: This file should be moved out of the shared folder. The shared folder across all packages should be identical. Considering that we are removing the models from the shared folder, we should move this class to the non-shared part of the chat code as well

RezaJooyandeh (changes requested):
Let's remove the models from the shared folder of all pacakges

sdk/communication/azure-communication-chat/azure/communication/chat/_shared/communication_identifier_serializer.py
sacheu: Thanks. Updated in new iteration.

RezaJooyandeh (approved):

Issue #12427, opened 2020-07-07
[Event Hubs] Idempotent producer


YijunXieMS:
Feature spec (.NET Design Proposal): https://gist.github.com/jsquire/1cc4db6b3ca4ef13d26dc5315483555b
Python gist: https://gist.github.com/yunhaoling/c17ddd3c1bfbe5c4ba156a8eb0e54e7d
.NET Sample gist: https://gist.github.com/jsquire/0f3bc5701388fe03ccd027e86f1994f1

.NET Implementation PR:

https://github.com/Azure/azure-sdk-for-net/pull/15034/files
https://github.com/Azure/azure-sdk-for-net/pull/15125/files
https://github.com/Azure/azure-sdk-for-net/pull/15332/files

tasks:
- [x] API review with arch
  - [x] gist (sample code)
  - [x] apiview
- [x] implementation
- [x] test
- [x] sample
- [x] doc

KieranBrantnerMagee:
Update: OstensiblynNov timespan for python, .net will be going live with reference implementation with the other languages following.

PR #16756, opened 2021-02-16
[EventHubs] Idempotent producer

yunhaoling:
gist: https://gist.github.com/yunhaoling/c17ddd3c1bfbe5c4ba156a8eb0e54e7d
issue: https://github.com/Azure/azure-sdk-for-python/issues/12427

pending uamqp v1.2.15 release: https://github.com/Azure/azure-uamqp-python/pull/210


sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
```diff
@@ -392,6 +417,20 @@ def size_in_bytes(self):
         """
         return self._size
 
+    @property
+    def starting_published_sequence_number(self):
```
annatisch, line 421: I still feel this needs a rename.... 
And I think my preference is to go with `published_sequence_number` have the documentation indicate that this refers to the first in the batch.

annatisch: IIRC batches are transactional right? There's no partial failure here?

sdk/eventhub/azure-eventhub/azure/eventhub/_configuration.py
```diff
@@ -51,3 +56,41 @@ def __init__(self, **kwargs):
...
+     publishing.
+     The starting sequence number should be in the range from 0 to max signed integer (2^31 - 1) as
+     required by the service.
+
+    """
+    def __init__(self, **kwargs):
+        validate_producer_client_partition_config(kwargs)
+        self.owner_level = kwargs.get("owner_level")  # type: Optional[int]
+        self.producer_group_id = kwargs.get("producer_group_id")  # type: Optional[int]
+        self.starting_sequence_number = kwargs.get("starting_sequence_number")  # type: Optional[int]
```
annatisch, line 96: I thought all these were required?

sdk/eventhub/azure-eventhub/azure/eventhub/_producer.py
```diff
@@ -185,41 +277,29 @@ def _on_outcome(self, outcome, condition):
 
     def _wrap_eventdata(
         self,
-        event_data,  # type: Union[EventData, EventDataBatch, Iterable[EventData]]
+        event_data_batch,  # type: EventDataBatch
         span,  # type: Optional[AbstractSpan]
         partition_key,  # type: Optional[AnyStr]
     ):
-        # type: (...) -> Union[EventData, EventDataBatch]
-        if isinstance(event_data, EventData):
```
annatisch, line 193: Couldn't see where the processing of a single EventData got moved to?

annatisch: Actually I guess you can only send by batch at the moment?

sdk/eventhub/azure-eventhub/azure/eventhub/_producer.py
```diff
@@ -57,8 +65,72 @@ def _set_trace_message(event_datas, parent_span=None):
...
+            return
+        self._owner_level = properties.get(PRODUCER_EPOCH)
+        self._producer_group_id = properties.get(PRODUCER_ID)
+        self._starting_sequence_number = properties.get(PRODUCER_SEQUENCE_NUMBER)
+
+    def _set_idempotent_producer_link_configs(self):
+        self._desired_capabilities = utils.data_factory(types.AMQPArray([IDEMPOTENT_PRODUCER_SYMBOL]))
+        if self._producer_group_id is not None or\
+                self._owner_level is not None or\
+                self._starting_sequence_number is not None:
```
annatisch, line 86: If one of these is None isn't that an error?

sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py
```diff
@@ -55,6 +87,19 @@ class EventHubProducerClient(ClientBase):
     :keyword str connection_verify: Path to the custom CA_BUNDLE file of the SSL certificate which is used to
      authenticate the identity of the connection endpoint.
      Default is None in which case `certifi.where()` will be used.
+    :keyword bool enable_idempotent_partitions: Indicates whether or not the producer should enable idempotent
+     publishing to the Event Hub partitions. If enabled, the producer will only be able to publish directly
+     to partitions; it will not be able to publish to the Event Hubs gateway for automatic partition routing
+     nor using a partition key. Default is False.
+    :keyword partition_configs: The set of configurations that can be specified to influence publishing behavior
```
annatisch, line 94: If we use "config" - I would make it singular.

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_producer_client_async.py
```diff
@@ -391,6 +441,37 @@ async def get_partition_properties(self, partition_id: str) -> Dict[str, Any]:
...
 
+    async def get_partition_publishing_properties(self, partition_id: str) -> Dict[str, Any]:
+        """Get the information about the state of publishing for a partition as observed by
+        the `EventHubProducerClient`. This data can always be read, but will only be populated with
+        information relevant to the active features for the producer client.
+            - `enable_idempotent_publishing` (bool)
+            - `partition_id` (str)
+            - `last_published_sequence_number` (Optional[int])
+            - `producer_group_id` (Optional[int])
+            - `owner_level` (Optional[int]
```
annatisch, line 452: I think the list might need spaces around it to render correctly

sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
yunhaoling: yes, batches are transactional, either all events in the batch get sent or failed.

yunhaoling: I don't have a strong preference on the name -- personally I think the "starting_" prefix adds some clarity for a batch object while with just `published_sequence_number` would give us the benefit of consistency with `EventData`.

@jsquire , would you be able to provide some context on the "starting_" prefix being chosen instead of just `starting_published_sequence_number`?

cc: @chradek @conniey

sdk/eventhub/azure-eventhub/azure/eventhub/_configuration.py
yunhaoling: AKAIK those are optional for the advanced creation scenario, have played with the different configs.
If those settings are mis-configured, the service would return kinda "invalid state" error.

sdk/eventhub/azure-eventhub/azure/eventhub/_producer.py
yunhaoling: correct, currently we only accept the batch/list object for the public send API.
so this is a simplification for the internal logic.

sdk/eventhub/azure-eventhub/azure/eventhub/_producer.py
yunhaoling: nope, I have tested the scenario, those three don't have to be all set at the same time.

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_producer_client_async.py
yunhaoling: I'll run a local sphinx to and do the necessary changes to make it being rendered properly.

sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
jsquire: "Starting" was chosen to denote that it was the sequence number applied to the first event in the batch.   The batch itself is not assigned a sequence number, nor is it considered a single item by the service.   

To determine the next sequence number that the service expects,  one has to take the batch's starting sequence number and increment it by the number of events that were in the batch.   This caused confusion for our beta users that expected the "LastPublishedSequenceNumber" returned as part of the partition publishing properties did not align with the "PublishedSequenceNumber" that we assigned to the batch.   It was agreed that the prefix "Starting" helped as a mnemonic.

sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py
```diff
@@ -360,6 +444,40 @@ def get_partition_properties(self, partition_id):
...
+
+        :param partition_id: The target partition ID.
+        :type partition_id: str
+        :rtype: dict
+        """
+        # pylint:disable=protected-access
+        output = {
+            'enable_idempotent_publishing': self._config.enable_idempotent_partitions,
+            'partition_id': partition_id,
+            'last_published_sequence_number': None,
```
swathipil, line 467: if the counterpart is called `starting_published_sequence_number` then should this be called `ending_published_sequence_number` for consistency? or `first_published_sequence_number` if you decide to keep `last_published_sequence_number`.
(Might not be the most important/relevant of details.)

sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py
```diff
@@ -87,7 +132,10 @@ def __init__(
         }  # type: Dict[str, Optional[EventHubProducer]]
         self._max_message_size_on_link = 0
         self._partition_ids = None  # Optional[List[str]]
+        self._partition_configs = kwargs.get("partition_configs") or {}
```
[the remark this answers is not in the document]
swathipil: also the name for the method in `_utils` is singular - `validate_producer_client_partition_config` --> so more reason for singular

sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py
```diff
@@ -55,6 +87,19 @@ class EventHubProducerClient(ClientBase):
...
+     to partitions; it will not be able to publish to the Event Hubs gateway for automatic partition routing
+     nor using a partition key. Default is False.
+    :keyword partition_configs: The set of configurations that can be specified to influence publishing behavior
+     specific to the configured Event Hub partition. These configurations are not necessary in the majority of
+     scenarios and are intended for use with specialized scenarios, such as when recovering the state used for
+     idempotent publishing.
+     It is highly recommended that these configurations only be specified if there is a proven need to do so;
+     Incorrectly configuring these values may result in an `EventHubProducerClient` instance that is unable to
+     publish to the Event Hubs. These configurations are ignored when publishing to the Event Hubs gateway for
+     automatic routing or when using a partition key.
```
swathipil, line 101: This might be self-explanatory (and not needed b/c, as you mentioned, our users don't need to be spoon-fed), but adding something at the end like the following could provide clarity:

If a dict is provided, instead of a PartitionPublishingConfiguration object, it may contain the optional parameters `producer_group_id`, `owner_level`, and `starting_sequence_number`.

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_producer_client_async.py
```diff
@@ -11,8 +11,11 @@
 from ..exceptions import ConnectError, EventHubError
 from ._client_base_async import ClientBaseAsync
 from ._producer_async import EventHubProducer
+from .._producer_client import validate_outgoing_event_data
 from .._constants import ALL_PARTITIONS
 from .._common import EventDataBatch, EventData
+from .._configuration import PartitionPublishingConfiguration
+from .._utils import validate_producer_client_partition_config
```
swathipil, line 19: same comments from sync apply to async

sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py
swathipil: looks like in the sample you say `ending`

sdk/eventhub/azure-eventhub/tests/livetest/asynctests/test_send_async.py
```diff
@@ -223,3 +225,191 @@ async def test_send_batch_pid_pk_async(invalid_hostname, partition_id, partition
...
+
+    partition_publishing_properties = await client.get_partition_publishing_properties("0")
+
+    # test out-of-sequence
+    partition_config_for_producer_2 = {
+        "0": {
+            "producer_group_id": partition_publishing_properties["producer_group_id"],
+            "owner_level": partition_publishing_properties["owner_level"],
+            "starting_sequence_number": partition_publishing_properties["last_published_sequence_number"] - 5
+        }
```
swathipil, line 399: what happens if you send in an empty dict --> even though you set `enable_idempotent_partitions=True`?

sdk/eventhub/azure-eventhub/tests/livetest/synctests/test_send.py
```diff
@@ -229,3 +231,190 @@ def test_send_batch_pid_pk(invalid_hostname, partition_id, partition_key):
...
+    client_2.send_batch(event_data_list, partition_id="0")
+    partition_publishing_properties_2 = client_2.get_partition_publishing_properties("0")
+    # with the new owner_level, the sequence number starts from 0 again
+    assert partition_publishing_properties_2["last_published_sequence_number"] == 1
+    assert partition_publishing_properties_2["producer_group_id"] == 0
+    assert partition_publishing_properties_2["owner_level"] == partition_publishing_properties["owner_level"] + 1
+
+    # test producer-epoch-stolen
+    with pytest.raises(EventDataSendError):
+        # the first producer will get exploded as the link is stolen by the second producer with higher owner level
```
swathipil, line 337: probably dumb question, but is it possible for owner_levels to be the same ever? for example, if the user passes in the same level for both producers.

swathipil (commented):
left a bunch of comments, but mostly about doc cleanup - lmk if comments look wrong or don't make sense :)

sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py
yunhaoling: it's actually composed of multiple partition configs -- a dictionary, the concern I have is that singular is not precise.

maybe it should be called `partitions_config`?

sdk/eventhub/azure-eventhub/tests/livetest/synctests/test_send.py
yunhaoling: yes, according to the service spec and my test, onwer_level could be set to the same value, it's the <producer-id, owner-level> pair identifies a producer.

sdk/eventhub/azure-eventhub/tests/livetest/asynctests/test_send_async.py
yunhaoling: the empty dict will simply be ignored, no settings on the producer link

sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py
```diff
@@ -135,12 +183,22 @@ def _start_producer(self, partition_id, send_timeout):
...
-    def _create_producer(self, partition_id=None, send_timeout=None):
-        # type: (Optional[str], Optional[Union[int, float]]) -> EventHubProducer
+    def _create_producer(
+        self,
+        partition_id=None,
+        send_timeout=None,
+        enable_idempotent_partitions=False,
+        partition_config=None
+    ):
+        # type: (Optional[str], Optional[Union[int, float]], bool, PartitionPublishingConfigType) -> EventHubProducer
```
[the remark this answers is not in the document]
yunhaoling: I already have the type definition:
```
PartitionPublishingConfigType = Optional[Union[PartitionPublishingConfiguration, dict]]
```

and for the bool, according to the [typing docs](https://docs.python.org/3/library/typing.html#typing.Optional), I don't think it has to be optional?
```
typing.Optional
Optional type.

Optional[X] is equivalent to Union[X, None].

Note that this is not the same concept as an optional argument, which is one that has a default. An optional argument with a default does not require the Optional qualifier on its type annotation just because it is optional. For example:

def foo(arg: int = 0) -> None:
    ...
On the other hand, if an explicit value of None is allowed, the use of Optional is appropriate, whether the argument is optional or not. For example:

def foo(arg: Optional[int] = None) -> None:
    ...
```

sdk/eventhub/azure-eventhub/azure/eventhub/_configuration.py
```diff
@@ -51,3 +56,41 @@ def __init__(self, **kwargs):
...
+
+class PartitionPublishingConfiguration(DictMixin):
+    """
+    The set of configurations that can be specified for an `EventHubProducerClient`
+    to influence its behavior when publishing directly to an Event Hub partition.
+
+    :ivar int producer_group_id: The identifier of the producer group that this producer is associated with when
+     publishing to the associated partition.
+    :ivar int owner_level: The owner level indicates that a publishing is intended to be performed exclusively for
+     events in the requested partition in the context of the associated producer group.
```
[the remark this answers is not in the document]
yunhaoling: each idempotent producer has a producer-group-id, that's the "associated producer group"

(to be honest I copy those docs from .NET, I am not a good writer :P)

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_producer_client_async.py
yunhaoling: sorry.. I'm a bit losing the context (too many comments) -- which comments are you referring to?

sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py
yunhaoling: So personally I think `last_published_sequence_number` is more suitable for a `EventHubProducerClient` and in contrast `ending_published_sequence_number` sounds reasonable to a `EventDataBatch` which has the `starting_published_sequence_number`.

But I think this a good starter for discussion on the prefix of those two object, I have created a separate issue for the naming discussion: https://github.com/Azure/azure-sdk-for-python/issues/16994, feel free to leave thoughts

sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py
yunhaoling: updated, thanks!

sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py
yunhaoling: name updated

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_producer_client_async.py
yunhaoling: you're right! spaces added, thanks,

sdk/eventhub/azure-eventhub/azure/eventhub/_configuration.py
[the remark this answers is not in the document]
swathipil: ah yes that makes sense (I'm also a big fan of copy/paste magic XD )

sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py
swathipil: both of those sound reasonable to me

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_producer_client_async.py
swathipil: sorry, should have stated explicitly. It was just the same naming comments, like `partition_config` instead of `partition_configs` etc. Looks like those have all been resolved so you're good!

swathipil (approved):
looks good to me (and my amqp-novice eyes)

sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
yunhaoling: I have created a separate issue for further discussion on the namings: https://github.com/Azure/azure-sdk-for-python/issues/16994

annatisch: Yeah we can keep as-is for preview and settle before GA

sdk/eventhub/azure-eventhub/azure/eventhub/_producer.py
```diff
@@ -80,6 +153,11 @@ class EventHubProducer(
      periods of inactivity. The default value is `None`, i.e. no keep alive pings.
     :keyword bool auto_reconnect: Whether to automatically reconnect the producer if a retryable error occurs.
      Default value is `True`.
+    :keyword bool enable_idempotent_partitions: Indicates whether or not the producer should enable idempotent
+     publishing to the Event Hub partitions.
+    :keyword dict partition_config: The optional producer configuration. This must be a dictionary that
+     contains the following optional configurations for the partition: `'owner_level'` (int value),
+     `'producer_group_id '` (int value) and `'starting_sequence_number'` (int value).
```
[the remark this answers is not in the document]
annatisch: For the record, I'm still not a fan of these being two separate parameters. We can go to preview with it, but I'd like to review before GA.

sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py
```diff
@@ -14,13 +14,45 @@
...
+        if is_idempotent_publishing:
+            if event_data_batch._partition_id is None:
+                raise ValueError("The EventDataBatch object must have the partition_id set when performing "
+                                 "idempotent publishing. Please create an EventDataBatch object with partition_id.")
+            if event_data_batch.starting_published_sequence_number is not None:
+                raise ValueError("EventDataBatch object that has already been published by idempotent producer"
+                                 "cannot be published again. Please create a new object.")
+        if partition_id or partition_key:
+            raise TypeError("partition_id and partition_key should be None when sending an EventDataBatch "
+                            "because type EventDataBatch itself may have partition_id or partition_key")
```
annatisch, line 46: Is raising this error a change in behaviour for anyone? Or were we previously raising it somewhere else?

sdk/eventhub/azure-eventhub/tests/livetest/asynctests/test_receive_async.py
```diff
@@ -135,7 +135,8 @@ async def on_event(partition_context, event):
         ed = EventData("Event Number {}".format(i))
         ed.properties = app_prop
         event_list.append(ed)
-    senders[0].send(event_list)
+
+    senders[0].send(EventDataBatch._from_batch(event_list))
```
annatisch, line 139: I find it funny that in our tests (both EH and SB), we use these private shortcuts to create a batch...
I wonder how many customers do the same :P

sdk/eventhub/azure-eventhub/azure/eventhub/_producer.py
```diff
@@ -185,41 +278,29 @@ def _on_outcome(self, outcome, condition):
...
+            add_link_to_send(event, span)
+            if self._enable_idempotent_partitions:
+                self._populate_idempotent_event_annotations(event, i)
+
+        event_data_batch.message.on_send_complete = self._on_outcome
+        return event_data_batch
 
     def send(
         self,
-        event_data,  # type: Union[EventData, EventDataBatch, Iterable[EventData]]
```
annatisch, line 222: This wasn't a public typehint right?

sdk/eventhub/azure-eventhub/tests/livetest/synctests/test_consumer_client.py
```diff
@@ -11,8 +11,8 @@
 @pytest.mark.liveTest
 def test_receive_no_partition(connstr_senders):
     connection_str, senders = connstr_senders
-    senders[0].send(EventData("Test EventData"))
-    senders[1].send(EventData("Test EventData"))
```
annatisch, line 15: So previously it looks like we did support sending a single batch, even through it wasn't part of the public API?
I wonder if it might be dangerous to remove the logic that supported this...

sdk/eventhub/azure-eventhub/azure/eventhub/_producer.py
yunhaoling: nope, the `_producer` module is internal

sdk/eventhub/azure-eventhub/tests/livetest/asynctests/test_receive_async.py
yunhaoling: yea.. I remember there was once a discussion on enabling the Batch object to accept a list, however the pitfall here is when trying to adding a list into batch which exceeds the batch size, then users still have to loop over the list and add event one by one to avoid exceeding the size limit.

sdk/eventhub/azure-eventhub/azure/eventhub/_producer.py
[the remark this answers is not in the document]
yunhaoling: this is the internal producer for a single partition, so it accepts a dict which contains the key-value pair configs for that partition.

sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py
yunhaoling: the latter one, it is still raised in the `create_batch` method -- I did some refactor to reduce code duplication, extracting the same logic from the `create_batch` from both of sync and async producer client and put them in a helper method `validate_outgoing_event_data` which is called by `create_batch` and could be used by both of the clients.

sdk/eventhub/azure-eventhub/tests/livetest/synctests/test_consumer_client.py
yunhaoling: in our current implement from the public api, a single event is not accepted, an error will be raised:
```
    to_send_batch._load_events(event_data_batch)  # pylint:disable=protected-access
  File "d:\projects\azure-sdk-for-python\sdk\eventhub\azure-eventhub\azure\eventhub\_common.py", line 382, in _load_events
    for event_data in events:
TypeError: 'EventData' object is not iterable
```
so I think it's safe remove the logic.
apart from that, I'd like to keep the implementation compact and simple as I find they're not used. We could always bring the code back when needed.

may I ask what's the concern you have -- do you think that it will break users' usage?

sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
yunhaoling: the preview may stay for a while (or even be pulled out) so we should have time to discuss among languages :P

annatisch (approved):

yunhaoling:
livetest passed which could be found in the `python - eventhub - tests 20210304.3`

PR #16771, opened 2021-02-16
Add FeatureFlagConfigurationSetting, SecretReferenceConfigurationSetting

seankane-msft:


sdk/appconfiguration/azure-appconfiguration/azure/appconfiguration/_models.py
```diff
@@ -26,40 +29,428 @@ class ConfigurationSetting(Model):
...
+            last_modified=key_value.last_modified,
+            tags=key_value.tags,
+            read_only=key_value.locked,
+            etag=key_value.etag,
+            feature_filters=filters,
+        )
+
+    def _to_generated(self):
+        # type: (...) -> KeyValue
+        value = {
```
pakrym, line 197: Won't this overwrite the customer's value if it's not well-formed JSON?

seankane-msft: No, this does not overwrite the customer's value. `key` and `description` should be strings, and `enabled` should be a boolean so there is no fear of invalid json

pakrym: The service doesn't validate the value of the `value` property to be JSON at all even if the content type and the prefix are both set to represent a feature flag. So imagine you get `{ "value":"3" }` from the service. When re-serializing this code would replace it with the full feature-flag definitions `{ "value": "{ \"isEnabled\":false..... }" }` losing the original.

sdk/appconfiguration/azure-appconfiguration/azure/appconfiguration/_models.py
```diff
@@ -26,40 +29,428 @@ class ConfigurationSetting(Model):
...
+            users=dict_repr["Audience"]["Users"],
+            groups=dict_repr["Audience"]["Groups"],
+        )
+
+    def _to_generated(self):
+        # type: (...) -> dict
+        return {
+            "name": self._name,
+            "parameters": {
+                "Audience": {
```
pakrym, line 381: The schema seems to be different from https://apiview.dev/Assemblies/Review/d89b0cb38b474b07ad545fd89df0a4a2?diffRevisionId=ead773d194c94a3f8e72a482135355e6&doc=False&diffOnly=False#@azure/app-configuration!FeatureFlagTargetingClientFilter:interface


Sample JSON
```
{
   "id":"MyFeature",
   "description":"",
   "enabled":true,
   "conditions":{
      "client_filters":[
         {
            "name":"Microsoft.Targeting",
            "parameters":{
               "Audience":{
                  "Users":[
                     "sadf",
                     "aaa"
                  ],
                  "Groups":[
                     {
                        "Name":"asdf",
                        "RolloutPercentage":100
                     },
                     {
                        "Name":"fdas",
                        "RolloutPercentage":29
                     }
                  ],
                  "DefaultRolloutPercentage":72
               }
            }
         }
      ]
   }
}
```

sdk/appconfiguration/azure-appconfiguration/azure/appconfiguration/_models.py
```diff
@@ -26,40 +29,428 @@ class ConfigurationSetting(Model):
...
+
+    @classmethod
+    def _from_generated(cls, feature_filter):
+        if feature_filter["name"] == "Microsoft.Targeting":
+            return TargetingFeatureFilter.from_service(feature_filter["parameters"])
+        if feature_filter["name"] == "Microsoft.TimeWindow":
+            return TimeWindowFeatureFilter.from_service(feature_filter["parameters"])
+        if feature_filter["name"] == "Microsoft.Percentage":
+            return CustomFeatureFilter.from_service(feature_filter["parameters"])
+        raise ValueError(
```
pakrym, line 333: We can't throw here, customers are free to add custom filters of their own.

sdk/appconfiguration/azure-appconfiguration/azure/appconfiguration/_models.py
```diff
@@ -26,40 +28,313 @@ class ConfigurationSetting(Model):
...
-    def _from_key_value(cls, key_value):
+    def _from_generated(cls, key_value):
         # type: (KeyValue) -> ConfigurationSetting
         if key_value is None:
             return None
+        if key_value.content_type is not None:
+            if key_value.content_type.startswith(
+                FeatureFlagConfigurationSetting.feature_flag_content_type
+            ):
+                return FeatureFlagConfigurationSetting._from_generated(  # pylint: disable=protected-access
```
xiangyan99, line 63: What if we fail to interpret the value into a FeatureFlagConfigurationSetting? Should we just return an instance of ConfigurationSetting?

seankane-msft: I will add this in a try/except wrapper

sdk/appconfiguration/azure-appconfiguration/azure/appconfiguration/_models.py
```diff
@@ -26,40 +28,313 @@ class ConfigurationSetting(Model):
...
+        self.tags = kwargs.get("tags", {})
 
     @classmethod
-    def _from_key_value(cls, key_value):
+    def _from_generated(cls, key_value):
         # type: (KeyValue) -> ConfigurationSetting
         if key_value is None:
             return None
+        if key_value.content_type is not None:
+            if key_value.content_type.startswith(
```
pakrym, line 60: feature flag requires both content type and prefix.

pakrym (approved):
LGTM

sdk/appconfiguration/azure-appconfiguration/azure/appconfiguration/__init__.py
```diff
@@ -19,4 +24,9 @@
     "AzureAppConfigurationClient",
     "ConfigurationSetting",
     "ResourceReadOnlyError",
+    "FeatureFlagConfigurationSetting",
+    "SecretReferenceConfigurationSetting",
+    "PERCENTAGE",
```
xiangyan99, line 29: Should we have a enum for these constants?

sdk/appconfiguration/azure-appconfiguration/azure/appconfiguration/_azure_appconfiguration_client.py
```diff
@@ -277,16 +280,20 @@ def get_configuration_setting(
             raise binascii.Error("Connection string secret has incorrect padding")
 
     @distributed_trace
-    def add_configuration_setting(self, configuration_setting, **kwargs):
-        # type: (ConfigurationSetting, dict) -> ConfigurationSetting
+    def add_configuration_setting(
+        self,
+        configuration_setting,  # type: Union[ConfigurationSetting,SecretReferenceConfigurationSetting,FeatureFlagConfigurationSetting] # pylint: disable=line-too-long
```
xiangyan99, line 285: Do we want to list all of them?

My concern is if in the future we want to add more types, this will increase our maintenance cost.

sdk/appconfiguration/azure-appconfiguration/azure/appconfiguration/_azure_appconfiguration_client.py
```diff
@@ -277,16 +280,20 @@ def get_configuration_setting(
...
+        **kwargs  # type: dict
+    ):
+        # type: (...) -> ConfigurationSetting
 
         """Add a ConfigurationSetting into the Azure App Configuration service.
 
         :param configuration_setting: the ConfigurationSetting object to be added
         :type configuration_setting: :class:`ConfigurationSetting<azure.appconfiguration.ConfigurationSetting>`
```
xiangyan99, line 293: We may need to update here.

sdk/appconfiguration/azure-appconfiguration/azure/appconfiguration/_models.py
```diff
@@ -39,29 +42,259 @@ class ConfigurationSetting(Model):
...
+    def _from_generated(cls, key_value):
+        # type: (KeyValue) -> FeatureFlagConfigurationSetting
+        try:
+            if key_value is None:
+                return None
+            if key_value.value:
+                try:
+                    key_value.value = json.loads(key_value.value)
+                except json.decoder.JSONDecodeError:
+                    pass
```
xiangyan99, line 172: If we fail to decode the payload, should we just return ConfigSetting?

sdk/appconfiguration/azure-appconfiguration/azure/appconfiguration/_models.py
```diff
@@ -39,29 +42,259 @@ class ConfigurationSetting(Model):
...
+        self.enabled = enabled
+        self.label = kwargs.get("label", None)
+        self.content_type = kwargs.get("content_type", self._feature_flag_content_type)
+        self.last_modified = kwargs.get("last_modified", None)
+        self.tags = kwargs.get("tags", {})
+        self.read_only = kwargs.get("read_only", None)
+        self.etag = kwargs.get("etag", None)
+        self.description = kwargs.get("description", None)
+        self.display_name = kwargs.get("display_name", None)
+        self.filters = filters or []
```
xiangyan99, line 160: How do you handle use modifies filters property and value property needs to get updated as well?

sdk/appconfiguration/azure-appconfiguration/azure/appconfiguration/__init__.py
seankane-msft: Anna/Johan recommended a constant string rather than an Enum.

sdk/appconfiguration/azure-appconfiguration/azure/appconfiguration/_azure_appconfiguration_client.py
seankane-msft: I can change to `ConfigurationSetting`, the other two inherit from `ConfigurationSetting`.

sdk/appconfiguration/azure-appconfiguration/azure/appconfiguration/_models.py
seankane-msft: The `value` parameter is recreated by the `_to_generated` method.

sdk/appconfiguration/azure-appconfiguration/azure/appconfiguration/_models.py
seankane-msft: Not always. If the prefix and content_type correspond to FeatureFlagConfigSetting we try that. If the json deserial fails then we store the json string in the value variable. We can change this on both FeatureFlag and SecretReference easily if we want to go that route.

sdk/appconfiguration/azure-appconfiguration/azure/appconfiguration/_models.py
xiangyan99: _to_generated() is called when you talk to server.

What if:

```py
config_setting = FeatureFlagConfigurationSetting()     # with some value
config_setting.filters = <new value>
print(config_setting.value)       <-- expect updated value
```

seankane-msft: This will print `None`, a FeatureFlagConfigurationSetting value is stored in the `enabled` variable.

sdk/appconfiguration/azure-appconfiguration/azure/appconfiguration/_models.py
```diff
@@ -193,13 +193,47 @@ def value(self, new_value):
...
+    def filters(self, new_filters):
+        if self._value is None:
+            self._value = {
+                "conditions": {
+                    "client_filters": [
+                        self._filters
+                    ]
+                }
+            }
+        elif isinstance(self._value, dict):
```
xiangyan99, line 224: If not isinstance(self._value, dict), your self._value will conflict with self._filters. Then what is the expected behavior for _to_generated()?

xiangyan99: My understanding is 
if user sets self._value to an non-dict, the object is not a FeatureFlagConfigurationSetting any longer. Hence if they want to get/set FeatureFlagConfigurationSetting properties, we raise AttributeError.

sdk/appconfiguration/azure-appconfiguration/azure/appconfiguration/_models.py
```diff
@@ -174,23 +176,24 @@ def enabled(self, value):
...
         return self._value
 
     @value.setter
     def value(self, new_value):
         # type: (Dict[str, Any]) -> None
-        try:
-            self._enabled = new_value.get("enabled", self._enabled)
-        except AttributeError:
-            pass
+        if not new_value:
```
xiangyan99, line 194: Are None & empty string valid for new_value?

sdk/appconfiguration/azure-appconfiguration/CHANGELOG.md
```diff
@@ -5,6 +5,8 @@
 
 ## 1.2.0b1 (Unreleased)
 
+Azure Appconfiguratin requires Python 2.7 or Python 3.6+ since this release.
```
xiangyan99, line 8: We need a GA release with the warning before we can remove the 3.5 support. :)

sdk/appconfiguration/azure-appconfiguration/azure/appconfiguration/_models.py
seankane-msft: No they are not.

xiangyan99: Why? I think we can set value to None or empty string for basic config settings?

seankane-msft: For `ConfigurationSetting` it is valid, but for `SecretReference` and `FeatureFlag` the `._value` has to be a dictionary.

xiangyan99: So it is a valid scenario that users treat it as basic config setting and set the value to empty string, right? We cannot silently change the value to {}.

seankane-msft: Yes, I am adding a test for it now.

sdk/appconfiguration/azure-appconfiguration/azure/appconfiguration/_models.py
```diff
@@ -10,6 +11,8 @@ class ConfigurationSetting(Model):
     """A configuration value.
     Variables are only populated by the server, and will be ignored when
     sending a request.
+    :ivar value: The value of the configuration setting
+    :vartype value: str
```
xiangyan99, line 15: if you add value here, we may want to remove 

:param value:
:type value: str

below.

sdk/appconfiguration/azure-appconfiguration/azure/appconfiguration/_models.py
```diff
@@ -39,29 +42,362 @@ class ConfigurationSetting(Model):
...
         self.key = kwargs.get("key", None)
         self.label = kwargs.get("label", None)
-        self.content_type = kwargs.get("content_type", None)
         self.value = kwargs.get("value", None)
+        self.etag = kwargs.get("etag", None)
+        self.content_type = kwargs.get("content_type", self.content_type)
         self.last_modified = kwargs.get("last_modified", None)
         self.read_only = kwargs.get("read_only", None)
-        self.tags = kwargs.get("tags", None)
+        self.tags = kwargs.get("tags", {})
```
xiangyan99, line 57: Why we give default value {}?

Given the case user creates a ConfigSetting w/o tags, read it and send it back, will this change the value of tag?

sdk/appconfiguration/azure-appconfiguration/azure/appconfiguration/_models.py
```diff
@@ -39,29 +42,362 @@ class ConfigurationSetting(Model):
...
+    )
+    kind = "FeatureFlag"
+
+    def __init__(self, key, enabled, filters=None, **kwargs):
+        # type: (str, bool, Optional[List[Dict[str, Any]]]) -> None
+        self._enabled = enabled
+        super(FeatureFlagConfigurationSetting, self).__init__(**kwargs)
+        self.key = key
+        if not key.startswith(self.key_prefix):
+            self.key = self.key_prefix + key
```
xiangyan99, line 151: This is interesting. What's the behavior of other languages?

sdk/appconfiguration/azure-appconfiguration/azure/appconfiguration/_models.py
```diff
@@ -39,29 +42,362 @@ class ConfigurationSetting(Model):
...
+        self._enabled = self._value.get("enabled") or self._enabled
+        return self._value
+
+    @value.setter
+    def value(self, new_value):
+        # type: (Dict[str, Any]) -> None
+        if not new_value:
+            new_value = {}
+        self._enabled = new_value.get("enabled") or self._enabled
+        self._value = new_value
```
xiangyan99, line 197: Looks like if we set value=None then get enabled, the value will be changed to {"enabled": self._enabled}?

sdk/appconfiguration/azure-appconfiguration/azure/appconfiguration/_models.py
seankane-msft: No this will not.

sdk/appconfiguration/azure-appconfiguration/azure/appconfiguration/_models.py
seankane-msft: This only happens when a user creates one in a Python program. On deserialization we only deserialize into `FeatureFlagConfigurationSetting` if the prefix and content_type are the correct value.

sdk/appconfiguration/azure-appconfiguration/azure/appconfiguration/_models.py
seankane-msft: if `new_value = None` this will raise an `AttributeError` on line 196 (`new_value.get("enabled")`).

xiangyan99: {}.get('enabled')

returns None. :)

sdk/appconfiguration/azure-appconfiguration/azure/appconfiguration/_models.py
xiangyan99: Taken offline.

sdk/appconfiguration/azure-appconfiguration/azure/appconfiguration/_models.py
xiangyan99: I mean if user creates a FeatureFlagConfigurationSetting with key ".appconfig.featureflag/abc"

How can we know he/she wants ".appconfig.featureflag/abc" or ".appconfig.featureflag/.appconfig.featureflag/abc"?

seankane-msft: On the other hand do we want a user to have to create a feature flag by doing:
```python
my_flag = FeatureFlagConfigurationSetting(".appconfig.featureflag/my_flag", enabled=True)
```
or
```python
my_flag = FeatureFlagConfigurationSetting("my_flag", enabled=True)
```
@xiangyan99

xiangyan99: I don't have a strong opinion on this. Let's align the behavior with other languages. :)

sdk/appconfiguration/azure-appconfiguration/azure/appconfiguration/_models.py
seankane-msft: Ahh, good catch. I will remove the `if not new_value` lines, then it will throw an `AttributeError`.

sdk/appconfiguration/azure-appconfiguration/azure/appconfiguration/_models.py
seankane-msft: Talked with other languages. I will make the key prefix a constant and export it.

sdk/appconfiguration/azure-appconfiguration/azure/appconfiguration/_models.py
```diff
@@ -147,8 +147,6 @@ def __init__(self, key, enabled, filters=None, **kwargs):
         self._enabled = enabled
         super(FeatureFlagConfigurationSetting, self).__init__(**kwargs)
         self.key = key
-        if not key.startswith(self.key_prefix):
```
xiangyan99: Do we want to check if key.startswith(self.key_prefix)?

seankane-msft: Do we still want to do this? I removed it to align with java and javascript/typescript.

sdk/appconfiguration/azure-appconfiguration/azure/appconfiguration/_models.py
```diff
@@ -39,29 +40,362 @@ class ConfigurationSetting(Model):
...
+        self.etag = kwargs.get("etag", None)
+        self.description = kwargs.get("description", None)
+        self.display_name = kwargs.get("display_name", None)
+        self._filters = filters or []
+        self._value = kwargs.get("value", {})
+
+    @property
+    def enabled(self):
+        # type: () -> bool
+        if self._value is None:
```
xiangyan99, line 164: Not  raise AttributeError("value parameter is expected to be a dictionary") if value is None?

sdk/appconfiguration/azure-appconfiguration/azure/appconfiguration/_models.py
```diff
@@ -39,29 +40,362 @@ class ConfigurationSetting(Model):
...
+            raise AttributeError("value parameter is expected to be a dictionary")
+        else:
+            self._value["enabled"] = self._enabled
+        return self._enabled
+
+    @enabled.setter
+    def enabled(self, value):
+        # type: (bool) -> bool
+        if self._value is None:
+            self._value = {"enabled": value}
```
xiangyan99, line 176: If self._value is None, it is not a FeatureFlagConfigurationSetting, right? :)

sdk/appconfiguration/azure-appconfiguration/azure/appconfiguration/_models.py
seankane-msft: Are you suggesting that we raise that error if the user tries to set the value to None?

sdk/appconfiguration/azure-appconfiguration/azure/appconfiguration/_models.py
seankane-msft: FeatureFlagConfigurationSetting is determined by the key_prefix and content_type, The value can be anything, there is no service side validation.

sdk/appconfiguration/azure-appconfiguration/azure/appconfiguration/_models.py
xiangyan99: if not key.startswith(self.key_prefix):
    raise ValueError. :)

sdk/appconfiguration/azure-appconfiguration/azure/appconfiguration/_models.py
xiangyan99: If _value is not in right schema or key not startswith(self.key_prefix), it is not FeatureFlagConfigurationSetting any longer. Hence enabled & filters properties are invalid.

sdk/appconfiguration/azure-appconfiguration/azure/appconfiguration/_models.py
seankane-msft: Added!

sdk/appconfiguration/azure-appconfiguration/azure/appconfiguration/_models.py
xiangyan99: Do you mean enabled property is Nullable?

sdk/appconfiguration/azure-appconfiguration/azure/appconfiguration/_models.py
seankane-msft: There was some confusion in our discussion, the standard is to prefix no matter what. That is what Java and C# do, I will add the same to Python

sdk/appconfiguration/azure-appconfiguration/azure/appconfiguration/_models.py
seankane-msft: Yes it is nullable

sdk/appconfiguration/azure-appconfiguration/azure/appconfiguration/_models.py
xiangyan99: self._value is the real value we store. 

In a GETTER, we cannot change the value.

```py
if self._value is None:
    return None
```

same in 
```py
else:
            self._value["enabled"] = self._enabled
```

the value can only be changed in a setter method which means user explicitly modify the value.

xiangyan99: And I would suggest we have a private method to validate if the value and key are valid. It will be called for multiple times. (any time user reads FeatureFlagConfigurationSetting specific properties, we need to validate it first.)

seankane-msft: To be clear, we will
1. validate anytime `my_key.value = <new_value>` and `my_key.enabled = <bool>` occur, (validating on a set)?
2. Validate anytime a user tries to read the `value` or `enabled` properties as well (validate on a get)?
3. If they are valid, we update the internal dictionary or enabled value 
4. if they are not valid (ie not a dictionary, not a boolean) we raise an error?

sdk/appconfiguration/azure-appconfiguration/azure/appconfiguration/_models.py
```diff
@@ -39,29 +40,352 @@ class ConfigurationSetting(Model):
...
+            self._value["conditions"]["client_filters"] = new_filters
+        except KeyError:
+            self._value["conditions"] = {
+                "client_filters": new_filters
+            }
+
+    @property
+    def value(self):
+        # type: () -> Dict[str, Any]
+        self._validate()
```
xiangyan99, line 208: We don't need validate() here. value is a property of basic config setting. It is always valid.

sdk/appconfiguration/azure-appconfiguration/azure/appconfiguration/_models.py
```diff
@@ -39,29 +40,352 @@ class ConfigurationSetting(Model):
...
+        return self._value
+
+    @value.setter
+    def value(self, new_value):
+        # type: (Dict[str, Any]) -> None
+        if not isinstance(new_value, dict) and new_value is not None:
```
xiangyan99, line 214: Same a basic property is always set-able.

sdk/appconfiguration/azure-appconfiguration/azure/appconfiguration/_models.py
```diff
@@ -39,29 +40,352 @@ class ConfigurationSetting(Model):
...
+        "read_only": {"key": "read_only", "type": "bool"},
+        "tags": {"key": "tags", "type": "{str}"},
+    }
+    key_prefix = ".appconfig.featureflag/"
+    _feature_flag_content_type = (
+        "application/vnd.microsoft.appconfig.ff+json;charset=utf-8"
+    )
+    kind = "FeatureFlag"
+
+    def __init__(self, key, enabled, filters=[], **kwargs):  # pylint: disable=dangerous-default-value
```
xiangyan99, line 145: If we add prefix for user, I would suggest not using the name "key" here.

xiangyan99: Seems like .net uses "featureId"

seankane-msft: @xiangyan99 what do you suggest?

sdk/appconfiguration/azure-appconfiguration/azure/appconfiguration/_models.py
```diff
@@ -39,29 +40,346 @@ class ConfigurationSetting(Model):
...
+            self._value = {}
+        self._value["enabled"] = new_value
+
+    @property
+    def filters(self):
+        # type: () -> Union[None, List[Any]]
+        self._validate()
+        if self._value is None:
+            return None
+        return self._value["conditions"]["client_filters"]
```
xiangyan99, line 190: You may want to handle KeyError

sdk/appconfiguration/azure-appconfiguration/azure/appconfiguration/_models.py
```diff
@@ -39,29 +40,346 @@ class ConfigurationSetting(Model):
...
+            self._value["secret_uri"] = value
+        else:
+            raise ValueError("Expect 'value' to be a dictionary.")
+
+    def _validate(self):
+        # type: () -> None
+        if not (self._value is None or isinstance(self._value, dict)):
+            raise ValueError("Expect 'value' to be a dictionary or None.")
+
+    @property
```
xiangyan99, line 340: Given

```py
@property
    def value(self):
        # type: () -> Dict[str, Any]
        return self._value

    @value.setter
    def value(self, value):
        # type: (Dict[str, Any]) -> None
        self._value = value
```
Maybe we don't need _value. :)

xiangyan99: In fact, after I re-think about it. ConfigSetting.value is a string. why here it is a dict?

Should we return a string in getter?

seankane-msft: It is stored as json in the service I think it makes the most sense to treat it as a dictionary to make it the most valuable to the user.

xiangyan99:
Btw, don't forget to update changelog. :)

xiangyan99 (approved):

Issue #16723, opened 2021-02-12
[Storage] Blob download: short reads


jochen-ott-by:
- **Package Name**: azure-storage-blob 
- **Package Version**: 12.5.0
- **Operating System**: Debian 9
- **Python Version**: python3.6

Other relevant packages: azure-core==1.10.0

**Describe the bug**

Reading the blob content can return too few bytes, leading to data corruption.

**To Reproduce**
1. create some blobs with fake data (not too small: best beyond the maximum get blob size, which has a default of 32MB)
2. download these blobs completely via `BlobClient.download_blob().readall()`. If the download returned without error, compare the content with the expected one from the upload in step 1.

Note that you have to repeat step 2. about 1 million times (depending on network, may be more or less; I estimate this is the order of probability we observe); alternatively, take other measures to increase the likelihood of a connection error while downloading the blob content.

**Expected behavior**

When the blob download completed without exception, the download content must match exactly with what was uploaded before, including the length.

**Additional context**

We observe this mainly when downloading parts of blobs, i.e. when passing non-zero offset and length parameters to `BlobClient.download_blob()`

jochen-ott-by:
I wrote a test for the bug I believe is responsible for it, see
https://github.com/Azure/azure-sdk-for-python/compare/master...jochen-ott-by:reproduce-gh-16723?expand=1

The believe the bug is in the retry behavior of `azure.core.pipeline.transport._request_basic.StreamDownloadGenerator.__next__`:

https://github.com/Azure/azure-sdk-for-python/blob/56193e95eac71933cbaebea56d63c48c4a01d333/sdk/core/azure-core/azure/core/pipeline/transport/_requests_basic.py#L122-L148

(`self.iter_content_func` is `response.iter_content(self.block_size)`, where `response` is the `requests.Response` object)

So what can happen is that L124 fetches the next chunk from the network. This usually is fine. However, assume there is a ConnectionError. This invokes the retry behavior starting at L134. If nothing else goes wrong, the next interesting thing that happens is L143: the code actually calls `next` on a generator that already raised an exception! This will always raise a `StopIteration`. This exception propagates and the caller of `StreamDownloadGenerator.__next__` assumes that the download is complete (although it's not).

Some other observations:
[...]

PR #16783, opened 2021-02-17
Raise exception rather than swallowing it if there is something wrong…

xiangyan99:
… in retry stream downloading

### DIFF
--- a/sdk/core/azure-core/azure/core/pipeline/transport/_aiohttp.py
+++ b/sdk/core/azure-core/azure/core/pipeline/transport/_aiohttp.py
@@ -228,17 +228,24 @@ class AioHttpStreamDownloadGenerator(AsyncIterator):
             except _ResponseStopIteration:
                 self.response.internal_response.close()
                 raise StopAsyncIteration()
-            except (ChunkedEncodingError, ConnectionError):
+            except (ChunkedEncodingError, ConnectionError) as ex:
                 retry_total -= 1
                 if retry_total <= 0:
                     retry_active = False
                 else:
                     await asyncio.sleep(retry_interval)
                     headers = {'range': 'bytes=' + str(self.downloaded) + '-'}
-                    resp = await self.pipeline.run(self.request, stream=True, headers=headers)
-                    if resp.http_response.status_code == 416:
-                        raise
-                    chunk = await self.response.internal_response.content.read(self.block_size)
+                    try:
+                        resp = await self.pipeline.run(self.request, stream=True, headers=headers)
+                        if not resp.http_response:
+                            raise
+                        if resp.http_response.status_code == 416:
+                            raise
+                        chunk = await self.response.internal_response.content.read(self.block_size)
+                    except Exception as err:  # pylint: disable=broad-except
+                        _LOGGER.warning("Unable to stream download: %s", err)
+                        self.response.internal_response.close()
+                        raise ex
                     if not chunk:
                         raise StopAsyncIteration()
                     self.downloaded += len(chunk)
--- a/sdk/core/azure-core/azure/core/pipeline/transport/_requests_basic.py
+++ b/sdk/core/azure-core/azure/core/pipeline/transport/_requests_basic.py
@@ -130,17 +130,24 @@ class StreamDownloadGenerator(object):
                 self.response.internal_response.close()
                 raise StopIteration()
             except (requests.exceptions.ChunkedEncodingError,
-                    requests.exceptions.ConnectionError):
+                    requests.exceptions.ConnectionError) as ex:
                 retry_total -= 1
                 if retry_total <= 0:
                     retry_active = False
                 else:
                     time.sleep(retry_interval)
                     headers = {'range': 'bytes=' + str(self.downloaded) + '-'}
-                    resp = self.pipeline.run(self.request, stream=True, headers=headers)
-                    if resp.http_response.status_code == 416:
-                        raise
-                    chunk = next(self.iter_content_func)
+                    try:
+                        resp = self.pipeline.run(self.request, stream=True, headers=headers)
+                        if not resp.http_response:
+                            raise
+                        if resp.http_response.status_code == 416:
+                            raise
+                        chunk = next(self.iter_content_func)
+                    except Exception as err:   # pylint: disable=broad-except
+                        _LOGGER.warning("Unable to stream download: %s", err)
+                        self.response.internal_response.close()
+                        raise ex
                     if not chunk:
                         raise StopIteration()
                     self.downloaded += len(chunk)
--- a/sdk/core/azure-core/CHANGELOG.md
+++ b/sdk/core/azure-core/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## 1.11.1 (Unreleased)
 
+### Bug Fixes
+
+- Raise exception rather than swallowing it if there is something wrong in retry stream downloading  #16723
 
 ## 1.11.0 (2021-02-08)
 
--- a/sdk/core/azure-core/tests/async_tests/test_stream_generator_async.py
+++ b/sdk/core/azure-core/tests/async_tests/test_stream_generator_async.py
@@ -2,10 +2,13 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
 # ------------------------------------
+import requests
 from azure.core.pipeline.transport import (
     HttpRequest,
     AsyncHttpResponse,
     AsyncHttpTransport,
+    AsyncioRequestsTransportResponse,
+    AioHttpTransport,
 )
 from azure.core.pipeline import AsyncPipeline
 from azure.core.pipeline.transport._aiohttp import AioHttpStreamDownloadGenerator
@@ -105,3 +108,50 @@ async def test_connection_error_416():
     with mock.patch('asyncio.sleep', new_callable=AsyncMock):
         with pytest.raises(ConnectionError):
             await stream.__anext__()
+
+@pytest.mark.asyncio
+async def test_response_streaming_error_behavior():
+    # Test to reproduce https://github.com/Azure/azure-sdk-for-python/issues/16723
+    block_size = 103
+    total_response_size = 500
+    req_response = requests.Response()
+    req_request = requests.Request()
+
+    class FakeStreamWithConnectionError:
+        # fake object for urllib3.response.HTTPResponse
+
+        def stream(self, chunk_size, decode_content=False):
+            assert chunk_size == block_size
+            left = total_response_size
+            while left > 0:
+                if left <= block_size:
+                    raise ConnectionError()
+                data = b"X" * min(chunk_size, left)
+                left -= len(data)
+                yield data
+
+        def close(self):
+            pass
+
+    req_response.raw = FakeStreamWithConnectionError()
+
+    response = AsyncioRequestsTransportResponse(
+        req_request,
+        req_response,
+        block_size,
+    )
+
+    async def mock_run(self, *args, **kwargs):
+        return PipelineResponse(
+            None,
+            requests.Response(),
+            None,
+        )
+
+    transport = AioHttpTransport()
+    pipeline = AsyncPipeline(transport)
+    pipeline.run = mock_run
+    downloader = response.stream_download(pipeline)
+    with pytest.raises(ConnectionError):
+        while True:
+            await downloader.__anext__()
--- a/sdk/core/azure-core/tests/test_stream_generator.py
+++ b/sdk/core/azure-core/tests/test_stream_generator.py
@@ -7,6 +7,8 @@ from azure.core.pipeline.transport import (
     HttpRequest,
     HttpResponse,
     HttpTransport,
+    RequestsTransport,
+    RequestsTransportResponse,
 )
 from azure.core.pipeline import Pipeline, PipelineResponse
 from azure.core.pipeline.transport._requests_basic import StreamDownloadGenerator
@@ -99,3 +101,48 @@ def test_connection_error_416():
     with mock.patch('time.sleep', return_value=None):
         with pytest.raises(requests.exceptions.ConnectionError):
             stream.__next__()
+
+def test_response_streaming_error_behavior():
+    # Test to reproduce https://github.com/Azure/azure-sdk-for-python/issues/16723
+    block_size = 103
+    total_response_size = 500
+    req_response = requests.Response()
+    req_request = requests.Request()
+
+    class FakeStreamWithConnectionError:
+        # fake object for urllib3.response.HTTPResponse
+
+        def stream(self, chunk_size, decode_content=False):
+            assert chunk_size == block_size
+            left = total_response_size
+            while left > 0:
+                if left <= block_size:
+                    raise requests.exceptions.ConnectionError()
+                data = b"X" * min(chunk_size, left)
+                left -= len(data)
+                yield data
+
+        def close(self):
+            pass
+
+    req_response.raw = FakeStreamWithConnectionError()
+
+    response = RequestsTransportResponse(
+        req_request,
+        req_response,
+        block_size,
+    )
+
+    def mock_run(self, *args, **kwargs):
+        return PipelineResponse(
+            None,
+            requests.Response(),
+            None,
+        )
+
+    transport = RequestsTransport()
+    pipeline = Pipeline(transport)
+    pipeline.run = mock_run
+    downloader = response.stream_download(pipeline)
+    with pytest.raises(requests.exceptions.ConnectionError):
+        full_response = b"".join(downloader)


xiangyan99:
#16723

sdk/core/azure-core/azure/core/pipeline/transport/_requests_basic.py
```diff
@@ -130,17 +130,24 @@ def __next__(self):
...
-                    if resp.http_response.status_code == 416:
-                        raise
-                    chunk = next(self.iter_content_func)
+                    try:
+                        resp = self.pipeline.run(self.request, stream=True, headers=headers)
+                        if not resp.http_response:
+                            raise
+                        if resp.http_response.status_code == 416:
+                            raise
+                        chunk = next(self.iter_content_func)
```
jochen-ott-by, line 146: There is no point calling `next(self.iter_content_func)` at this point, as this line is only reached if the previous call to `next(self.iter_content_func)` in L124 raised an exception.

Note that this follows from the specificaiton of python generators. `self.iter_content_func` is a python generator and according to the python generator specification [1]:

```
If an unhandled exception [...] is raised by, or passes through, a generator function, then the exception is passed on to the caller in the usual way, and subsequent attempts to resume the generator function raise StopIteration. In other words, an unhandled exception terminates a generator's useful life.
```

Therefore: please remove this call. It has no purpose.


[1] https://www.python.org/dev/peps/pep-0255/#specification-generators-and-exception-propagation

xiangyan99: @jochen-ott-by Thanks for the comments.

The logic behind it is:

If there is a ConnectionError raises during the downloading, we will wait for a while,
```py
time.sleep(retry_interval)
```

Then try to re-connect and continue the downloading by setting the header
```py
headers = {'range': 'bytes=' + str(self.downloaded) + '-'}
```

and yes, if we hit failures again, we will pop it up to the caller (this is what we fix in this PR).

sdk/core/azure-core/azure/core/pipeline/transport/_requests_basic.py
```diff
@@ -130,17 +130,24 @@ def __next__(self):
...
+                    except Exception as err:   # pylint: disable=broad-except
+                        _LOGGER.warning("Unable to stream download: %s", err)
```
annatisch, line 148: Optional nit:
From a customers perspective - will this code always be in the context of a "download"? Do we stream in other customer usecases?
Maybe it could be more generic with something like "Unable to stream response content"..... but maybe that sounds too HTTP-ish.

sdk/core/azure-core/azure/core/pipeline/transport/_aiohttp.py
```diff
@@ -228,17 +228,24 @@ async def __anext__(self):
...
-                        raise
-                    chunk = await self.response.internal_response.content.read(self.block_size)
+                    try:
+                        resp = await self.pipeline.run(self.request, stream=True, headers=headers)
+                        if not resp.http_response:
+                            raise
+                        if resp.http_response.status_code == 416:
+                            raise
+                        chunk = await self.response.internal_response.content.read(self.block_size)
+                    except Exception as err:  # pylint: disable=broad-except
```
annatisch, line 245: I'm curious - prior to your change - which `except` clause was catching and swallowing these errors?

sdk/core/azure-core/azure/core/pipeline/transport/_requests_basic.py
```diff
@@ -130,17 +130,24 @@ def __next__(self):
...
                 else:
                     time.sleep(retry_interval)
                     headers = {'range': 'bytes=' + str(self.downloaded) + '-'}
-                    resp = self.pipeline.run(self.request, stream=True, headers=headers)
-                    if resp.http_response.status_code == 416:
-                        raise
-                    chunk = next(self.iter_content_func)
+                    try:
+                        resp = self.pipeline.run(self.request, stream=True, headers=headers)
+                        if not resp.http_response:
```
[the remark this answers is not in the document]
xiangyan99: It has __bool__. :)

https://github.com/psf/requests/blob/master/requests/models.py#L671

sdk/core/azure-core/azure/core/pipeline/transport/_requests_basic.py
xiangyan99: I tried to be consistent to https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/core/azure-core/azure/core/pipeline/transport/_requests_basic.py#L152. But I am open to change it.

sdk/core/azure-core/azure/core/pipeline/transport/_aiohttp.py
xiangyan99: > I'm curious - prior to your change - which `except` clause was catching and swallowing these errors?

It fails in https://github.com/psf/requests/blob/master/requests/models.py#L920,

self.status_code is None.

sdk/core/azure-core/azure/core/pipeline/transport/_requests_basic.py
jochen-ott-by: Thanks for the answer. However, your description does not match with what the code does.

"Then try to re-connect and continue the downloading" this is misleading, as it's *not* a new connection that is used for the attempt to continue the downloading. Rather, the original value of `self.iter_content_func` is used, the one that just raised an exception and can no longer be used for anything that makes sense.
The only failure reported here is a 416 "range not satisfiable".

And as already commented, the code makes assumptions it cannot safely make: StreamDownloader is also used for requests that already have a (non-trivial) range header set. So what happens if the original range header is, say, "bytes=1000-1999", chunking is 200 and there is a connection error after the first chunk?
This code will then go on a make a request for "range: bytes=200-", which does not overlap with the original request.

This makes no sense either.

sdk/core/azure-core/azure/core/pipeline/transport/_aiohttp.py
```diff
@@ -228,17 +228,24 @@ async def __anext__(self):
...
-                    if resp.http_response.status_code == 416:
-                        raise
-                    chunk = await self.response.internal_response.content.read(self.block_size)
+                    try:
+                        resp = await self.pipeline.run(self.request, stream=True, headers=headers)
+                        if not resp.http_response:
+                            raise
+                        if resp.http_response.status_code == 416:
+                            raise
+                        chunk = await self.response.internal_response.content.read(self.block_size)
```
jochen-ott-by, line 244: This line re-uses the original response that just raised an exception in L223 (it does *not* use the new response created on l. 239!). But: what is the point to re-try a read after a `ConnectionError`? How can that ever succeed? Even if for `ConnectionError`, there might be cases it makes sense, it's clearly wrong for a `ChunkedEncodingError`: this just means there is a serious error in the data stream parsing. Re-using the data stream that just observed such a parsing error is wrong.

sdk/core/azure-core/azure/core/pipeline/transport/_requests_basic.py
```diff
@@ -130,17 +130,25 @@ def __next__(self):
...
+                            raise
+                        if resp.http_response.status_code == 416:
+                            raise
+                        self.iter_content_func = resp.http_response.iter_content(self.block_size)
```
jochen-ott-by, line 146: This last change of re-setting `self.iter_content_func` seems to make sense superficially, because now `next(self.iter_content_func)` but it really does not help for the reasons I explained already in some depth, so I'm just citing myself here:
```
And as already commented, the code makes assumptions it cannot safely make: StreamDownloader is also used for requests that already have a (non-trivial) range header set. So what happens if the original range header is, say, "bytes=1000-1999", chunking is 200 and there is a connection error after the first chunk?
This code will then go on a make a request for "range: bytes=200-", which does not overlap with the original request.
```

The only way out I can see here is to remove the retry entirely. This function just does not have enough context information to re-try safely. So the best it can do is to raise and let the upper layers handle it.

sdk/core/azure-core/azure/core/pipeline/transport/_aiohttp.py
```diff
@@ -228,17 +228,25 @@ async def __anext__(self):
...
                     headers = {'range': 'bytes=' + str(self.downloaded) + '-'}
-                    resp = await self.pipeline.run(self.request, stream=True, headers=headers)
-                    if resp.http_response.status_code == 416:
-                        raise
-                    chunk = await self.response.internal_response.content.read(self.block_size)
+                    try:
+                        resp = await self.pipeline.run(self.request, stream=True, headers=headers)
+                        if not resp.http_response:
+                            raise
+                        if resp.http_response.status_code == 416:
```
jochen-ott-by, line 242: This will never be true, as `not resp.http_response` will always be fulfilled if status_code is 416

sdk/core/azure-core/azure/core/pipeline/transport/_requests_basic.py
```diff
@@ -130,17 +130,25 @@ def __next__(self):
...
                 raise StopIteration()
             except (requests.exceptions.ChunkedEncodingError,
-                    requests.exceptions.ConnectionError):
+                    requests.exceptions.ConnectionError) as ex:
                 retry_total -= 1
                 if retry_total <= 0:
                     retry_active = False
                 else:
                     time.sleep(retry_interval)
                     headers = {'range': 'bytes=' + str(self.downloaded) + '-'}
```
jochen-ott-by, line 139: I'm starting to repeat myself at least 3 times over, but again: the original request saved in `self.request` can already have a range set, e.g. `range: bytes=1000-`. So what does this new range request do, exactly?

Also note that there are REST APIs that have special range requests, e.g. azure blob store GET used `x-ms-range, for more information please refer to the documentation here: https://docs.microsoft.com/en-us/rest/api/storageservices/specifying-the-range-header-for-blob-service-operations

So it would not even help to try parse this header of the original request here, because one can never be sure to catch all the special cases of APIs here.

Again, please do not retry, this cannot be done safely.

sdk/core/azure-core/azure/core/pipeline/transport/_requests_basic.py
xiangyan99: Thank you.

Yes. I was about to answer this. :)

As you know, the purpose for azure-core is to support other SDK libraries. We have not heard any real scenarios from libraries that need to read bytes like "bytes=1000-1999". In storage, we always download entire file. This made us leaning towards having retry rather than failing directly. 

But yes, you are right. It is not correct in case "bytes=1000-1999". I will open an issue to track it.

In the meantime, do you have real scenarios that need this?

That can help us prioritize the issue.

Thanks.

sdk/core/azure-core/azure/core/pipeline/transport/_requests_basic.py
jochen-ott-by: I should also add that, in general `self.downloaded` is unreliable, because it is computed inconsistently, sometimes
as
```
self.downloaded += len(chunk)
```
and sometimes
```
self.downloaded += self._block_size
```

Both of them can be wrong, in general: requests configures urllib3 to do decoding based on the content-encoding header of the response (e.g. gunzip the data stream). It tries to read block_size bytes, then decodes them. However, for chunked http responses, the block_size is not actually used; instead, the chunks are iterated over.

So in essence, there is no way to reliably know how many of the original bytes were downloaded unless it's clear there is the special case of no http response chunking and no content decoding. This cannot be assumed at this point, as this is a general-purpose library.

sdk/core/azure-core/azure/core/pipeline/transport/_requests_basic.py
jochen-ott-by: Surely you are aware this is used in the azure-storage-blob library, which *does* set range headers by default, as it will use multi-threading to download large blobs in chunks, in parallel?

Yes, we have real scenarios we use this. As you might be aware, we use azure storage blob for big data analysis, processing dozens of TB daily. We had several customer-facing incidents that are very likely due to this bug. We spent person-weeks tracking this down and working around this.

sdk/core/azure-core/azure/core/pipeline/transport/_requests_basic.py
hoffmann: Hey @xiangyan99,

regarding the real word scenarios. We are using the azure blob storage in our machine learning applications that pull parquet data from the azure blob storage. Parquet files implement row group statistics, so a common optimisation is to only read the parts of the parquet file that have the row group statistics and then do pruning based on predicates. 

See https://github.com/mbr/simplekv/blob/master/simplekv/net/_azurestore_new.py#L194 for a file like interface to access the azure blob storage and http://peter-hoffmann.com/2020/understand-predicate-pushdown-on-rowgroup-level-in-parquet-with-pyarrow-and-python.html or an explanation about row group pruning.

xiangyan99: Thank you @jochen-ott-by @hoffmann , your information is really helpful.

Let me re-think about it. :)

sdk/core/azure-core/azure/core/pipeline/transport/_requests_basic.py
```diff
@@ -119,39 +154,58 @@ def __next__(self):
...
-                    if resp.http_response.status_code == 416:
-                        raise
-                    chunk = next(self.iter_content_func)
+                    # todo handle pre-set range & x-ms-range
+                    headers = self.request.headers
+                    if not self.range_header:
+                        range_header = {'range': 'bytes=' + str(self.downloaded) + '-'}
+                    else:
+                        range_header = {self.range_header: make_range_header(self.range, self.downloaded)}
+                    if self.etag:
```
jochen-ott-by, line 181: Combining the responses from multiple requests with different ranges must be done only if there is a strong validator, such as a non-weak etag (cf. section 3.3 of RFC 7234 ("Combining Partial Content") and section 4.3. of RFC 7233 ("Combining Ranges")).
So to be on the safe side, if there is no (non-weak) etag, there should be no retry.

(To see why, imagine a server response that dynamically generates some json data that does not guarantee the same order of keys in the json objects on the next request: this is a valid server behavior, but the client here would produce corrupt data).

sdk/core/azure-core/azure/core/pipeline/transport/_requests_basic.py
```diff
@@ -119,39 +154,58 @@ def __next__(self):
...
+                        range_header.update({'If-Match': self.etag})
+                    headers.update(range_header)
```
jochen-ott-by, line 183: Better make a copy here, as this might loose the original request information, which can be pretty confusing, e.g. in case some error is reported later that includes the "original request".

sdk/core/azure-core/azure/core/pipeline/transport/_requests_basic.py
```diff
@@ -108,6 +130,19 @@ def __init__(self, pipeline, response):
         self.iter_content_func = self.response.internal_response.iter_content(self.block_size)
         self.content_length = int(response.headers.get('Content-Length', 0))
         self.downloaded = 0
+        headers = response.request.headers
+        if "x-ms-range" in headers:
+            self.range_header = "x-ms-range"
+            self.range = headers["x-ms-range"]
```
jochen-ott-by, line 136: A more robust to set `self.range` is the *response* headers, as this tells you which range the server actually responses with (keep in mind that the server can legitimately ignore the range header and return the full content!). See https://tools.ietf.org/html/rfc7233#section-4.1

jochen-ott-by:
This looks like good progress, @xiangyan99 ! However, I'm still concerned about the complexity of this solution, which does not deal with the fact that (in general) `self.downloaded` cannot be computed reliably (at least for the `requests` backend). Getting this right will be next to impossible, so I think it's worth to take a step back and ask again what the current state is, how it can be improved short term, and what longer term possibilities are.

Here is my view:

 * The latest released version has a serious bug that produces data corruption.
 * While the currently released version does have code that was written with the *intention* to re-try an interrupted streamed download, this code path never lived up to the intention. So effectively, the latest released version has no re-tries in real life
 * Data corruption bugs are important to fix ASAP.

So again, please consider removing the retry code altogether to get a fix for the data corruption bug ASAP. Note that removing the code that was intended to re-try does not really make anything worse for anyone, as the code never actually helped retrying anything (!). So removing it is a clear net win for everyone.

After this immediate fix, one could evaluate again weather the complexity of a re-try at this point of the code is worth the effort. (In my personal opinion, it's not worth it; instead, better re-try some levels up in the stack, where it's much easier, as much more context information is available how to re-try best).

xiangyan99:
> This looks like good progress, @xiangyan99 ! However, I'm still concerned about the complexity of this solution, which does not deal with the fact that (in general) `self.downloaded` cannot be computed reliably (at least for the `requests` backend). Getting this right will be next to impossible, so I think it's worth to take a step back and ask again what the current state is, how it can be improved short term, and what longer term possibilities are.
> 
> Here is my view:
> 
> * The latest released version has a serious bug that produces data corruption.
> * While the currently released version does have code that was written with the _intention_ to re-try an interrupted streamed download, this code path never lived up to the intention. So effectively, the latest released version has no re-tries in real life
> * Data corruption bugs are important to fix ASAP.
> 
> So again, please consider removing the retry code altogether to get a fix for the data corruption bug ASAP. Note that removing the code that was intended to re-try does not really make anything worse for anyone, as the code never actually helped retrying anything (!). So removing it is a clear net win for everyone.
> 
> After this immediate fix, one could evaluate again weather the complexity of a re-try at this point of the code is worth the effort. (In my personal opinion, it's not worth it; instead, better re-try some levels up in the stack, where it's much easier, as much more context information is available how to re-try best).

Thanks for your inputs. I am not quite clear why self.downloaded cannot be computed reliably.

Let's say the file has 1mb and each time we read 1k. First read we will try get 0~1k. If succeeds, we record 1k is downloaded (the return may be less than 1k because of decoding). If we need to re-connect, we want to tell the service to continue from 1k because from service's perspective, it always deals with pre-decoded bits. right?

We release our SDKs on monthly cadence and next release will be in the week of 3/8. (I thought you already talked to your customer support team and confirmed 3/8 worked for you?)

jochen-ott-by:
> Thanks for your inputs. I am not quite clear why self.downloaded cannot be computed reliably.

I already explained this in some depth here: https://github.com/Azure/azure-sdk-for-python/pull/16783#discussion_r579411043


I'm citing the relevant part here:

> requests configures urllib3 to do decoding based on the content-encoding header of the response (e.g. gunzip the data stream):
> https://github.com/psf/requests/blob/bdc00eb0978dd3cdd43f7cd1f95ced7aa75fab39/requests/models.py#L753-L754
> So this will (usually) read block_size bytes from the underlying raq data stream, then decode these bytes to return chunk. > So self.downloaded += len(chunk) will be wrong if such decoding is happening.

> However, note that urllib3's stream function that is documented here:
> https://urllib3.readthedocs.io/en/1.26.3/reference/urllib3.response.html#urllib3.response.HTTPResponse.stream
> actually does not promise to always read exactly block_size bytes from the underlying stream. And in general, it won't: in > the current version, for chunked http responses, the block_size is not actually used; instead, the chunks are iterated over.

> So in essence, there is no way to reliably know how many of the original bytes were downloaded unless it's clear there is > the special case of no http response chunking and no content decoding, and relying on implementation details of urllib3.


You wrote:

> Let's say the file has 1mb and each time we read 1k.

The point is that the `requests`/`urllib3` library does not promise anything about the chunk size it downloads. In other words: even if you say `response.iter_content(1000)`, iterating over the stream might actually only read chunks of, say, 786 bytes. There is no guarantee whatsoever from requests that the chunk size passed to `iter_content` is respected exactly.

> First read we will try get 0~1k. If succeeds, we record 1k is downloaded (the return may be less than 1k because of decoding). If we need to re-connect, we want to tell the service to continue from 1k because from service's perspective, it always deals with pre-decoded bits. right?

No. What can happen is that the first chunk from `response.iter_content(1000)` actually read only 786 bytes. Then, this code goes on in the re-try and starts from bytes position 1000. This corrupts data.

> We release our SDKs on monthly cadence and next release will be in the week of 3/8. (I thought you already talked to your customer support team and confirmed 3/8 worked for you?)

Release-wise, this would work. I'm just not confident that the current approach of implementing re-tries is something that can be done properly by then.

xiangyan99:
> > Thanks for your inputs. I am not quite clear why self.downloaded cannot be computed reliably.
> 
> I already explained this in some depth here: [#16783 (comment)](https://github.com/Azure/azure-sdk-for-python/pull/16783#discussion_r579411043)
> 
> I'm citing the relevant part here:
> 
> > requests configures urllib3 to do decoding based on the content-encoding header of the response (e.g. gunzip the data stream):
> > https://github.com/psf/requests/blob/bdc00eb0978dd3cdd43f7cd1f95ced7aa75fab39/requests/models.py#L753-L754
> > So this will (usually) read block_size bytes from the underlying raq data stream, then decode these bytes to return chunk.  So self.downloaded += len(chunk) will be wrong if such decoding is happening.

Yes. len(chunk) is not reliable and here we modified the code to only use block_size.

> 
> > However, note that urllib3's stream function that is documented here:
> > https://urllib3.readthedocs.io/en/1.26.3/reference/urllib3.response.html#urllib3.response.HTTPResponse.stream
> > actually does not promise to always read exactly block_size bytes from the underlying stream. And in general, it won't: in  the current version, for chunked http responses, the block_size is not actually used; instead, the chunks are iterated over.

in https://urllib3.readthedocs.io/en/1.26.3/reference/urllib3.response.html#urllib3.response.HTTPResponse.stream

it says:

stream(amt=65536, decode_content=None)
A generator wrapper for the read() method. __A call will block until amt bytes have been read from the connection or until the connection is closed.__

So it reads block_size or the connection is closed (connection error or done).

If there is a connection error, we don call
self.downloaded += self.block_size

if it is done, everything is fine.

> 
> > So in essence, there is no way to reliably know how many of the original bytes were downloaded unless it's clear there is  the special case of no http response chunking and no content decoding, and relying on implementation details of urllib3.
> 
> You wrote:
> 
> > Let's say the file has 1mb and each time we read 1k.
> 
> The point is that the `requests`/`urllib3` library does not promise anything about the chunk size it downloads. In other words: even if you say `response.iter_content(1000)`, iterating over the stream might actually only read chunks of, say, 786 bytes. There is no guarantee whatsoever from requests that the chunk size passed to `iter_content` is respected exactly.
> 
> > First read we will try get 0~1k. If succeeds, we record 1k is downloaded (the return may be less than 1k because of decoding). If we need to re-connect, we want to tell the service to continue from 1k because from service's perspective, it always deals with pre-decoded bits. right?
> 
> No. What can happen is that the first chunk from `response.iter_content(1000)` actually read only 786 bytes. Then, this code goes on in the re-try and starts from bytes position 1000. This corrupts data.

Yes. To make sure we are one same page, you are saying the return (after decoding) is 786, but the raw bytes (before decoding) it reads is 1000, right? So next time we should tell it to read start from 1000.

> 
> > We release our SDKs on monthly cadence and next release will be in the week of 3/8. (I thought you already talked to your customer support team and confirmed 3/8 worked for you?)
> 
> Release-wise, this would work. I'm just not confident that the current approach of implementing re-tries is something that can be done properly by then.

jochen-ott-by:
> > > Thanks for your inputs. I am not quite clear why self.downloaded cannot be computed reliably.
> > 
> > 
> > I already explained this in some depth here: [#16783 (comment)](https://github.com/Azure/azure-sdk-for-python/pull/16783#discussion_r579411043)
> > I'm citing the relevant part here:
> > > requests configures urllib3 to do decoding based on the content-encoding header of the response (e.g. gunzip the data stream):
> > > https://github.com/psf/requests/blob/bdc00eb0978dd3cdd43f7cd1f95ced7aa75fab39/requests/models.py#L753-L754
> > > So this will (usually) read block_size bytes from the underlying raq data stream, then decode these bytes to return chunk.  So self.downloaded += len(chunk) will be wrong if such decoding is happening.
> 
> Yes. len(chunk) is not reliable and here we modified the code to only use block_size.
> 
> > > However, note that urllib3's stream function that is documented here:
> > > https://urllib3.readthedocs.io/en/1.26.3/reference/urllib3.response.html#urllib3.response.HTTPResponse.stream
> > > actually does not promise to always read exactly block_size bytes from the underlying stream. And in general, it won't: in  the current version, for chunked http responses, the block_size is not actually used; instead, the chunks are iterated over.
> 
> in https://urllib3.readthedocs.io/en/1.26.3/reference/urllib3.response.html#urllib3.response.HTTPResponse.stream
> 
> it says:
> 
> stream(amt=65536, decode_content=None)
> A generator wrapper for the read() method. **A call will block until amt bytes have been read from the connection or until the connection is closed.**

It also says: **amt: How much of the content to read. The generator will return up to much data per iteration, but may return less.**

So the documentation is incomplete here.

Also, if you look how it's *implemented*, you'll see that for chunked encoding, this will always iterate over the chunks returned by the server. This is what I specifically wrote above. I actually tested it. It really does not respect this parameter.

> So it reads block_size or the connection is closed (connection error or done).

No, in general, it does not.


> > > So in essence, there is no way to reliably know how many of the original bytes were downloaded unless it's clear there is  the special case of no http response chunking and no content decoding, and relying on implementation details of urllib3.
> > 
> > 
> > You wrote:
> > > Let's say the file has 1mb and each time we read 1k.
> > 
> > 
> > The point is that the `requests`/`urllib3` library does not promise anything about the chunk size it downloads. In other words: even if you say `response.iter_content(1000)`, iterating over the stream might actually only read chunks of, say, 786 bytes. There is no guarantee whatsoever from requests that the chunk size passed to `iter_content` is respected exactly.
> > > First read we will try get 0~1k. If succeeds, we record 1k is downloaded (the return may be less than 1k because of decoding). If we need to re-connect, we want to tell the service to continue from 1k because from service's perspective, it always deals with pre-decoded bits. right?
> > 
> > 
> > No. What can happen is that the first chunk from `response.iter_content(1000)` actually read only 786 bytes. Then, this code goes on in the re-try and starts from bytes position 1000. This corrupts data.
> 
> To make sure we are one same page, you are saying the return (after decoding) is 786, but the raw bytes (before decoding) it reads is 1000, right? So next time we should tell it to read start from 1000.

No. Assume there is no compression decoding (like gzip)  at all. You pass amt=1000 to `urllib3.response.Response.stream`. The server, however, may respond with chunked encoding, yielding e.g. chunks of the size 768. So the first chunk of data yielded by `stream` will read 768 raw bytes. Yet, the request that re-tries starts reading at 1000, so bytes 768-999 never reach the client.

BTW, there is another potential issue here I have not yet tested: what happens if the content-encoding is gzip and you try to re-download a partial content after a failure? The first request will (successfully) decompress the response. But for the partial data, the client cannot really do that ...

kekatahi:
@xiangyan99  
 Do you have the ETA about this PR?
We are waiting for this fix. #16980

xiangyan99:
> @xiangyan99
> Do you have the ETA about this PR?
> We are waiting for this fix. #16980

We have merged one to disable the retry. 

Could you help to describe your scenario? 

That can help us prioritize the issue.

Thank you.

kekatahi:
@xiangyan99 We download the blob from databricks and sometime there are network error with no exception and the downloaded file seems incomplete compare to source one. Do you mean that you already fix those behavior(making incomplete file with no exception ) ? So if there are no exception, those gap never occur?

xiangyan99:
> @xiangyan99 We download the blob from databricks and sometime there are network error with no exception and the downloaded file seems incomplete compare to source one. Do you mean that you already fix those behavior(making incomplete file with no exception ) ? So if there are no exception, those gap never occur?

We used to have the issue that we return incomplete downloads with a "succeed" tag. 

With the fix, current behavior is:
either the download succeeds and we get everything or we raise error.

Please let me know if this does not meet your requirements.

PR #16792, opened 2021-02-18
[Key Vault] Test certificates library against a shared vault

### LINKED ISSUE #15434: [Key Vault] Update tests to work on a shared vault (certificates)
Part of solution to #15371.

Divided into library-specific issues for manageability.

mccoyp:
Resolves #15434.

This implements the same preparer changes from #16474 in the certificates library. This also updates `tests.yml` to deploy ARM templates for certificates tests, and tweaks the test resource cleanup script to purge pending certificates (so that deleted certificates that were cancelled during creation can actually get purged).


sdk/keyvault/azure-keyvault-certificates/tests/test_examples_certificates_async.py
```diff
@@ -4,17 +4,20 @@
 # ------------------------------------
 import asyncio
 import functools
+import pytest
```
chlowell, line 7: Grouping pytest with standard library imports contravenes the [PEP 8](https://www.python.org/dev/peps/pep-0008/#imports) style we usually enforce with pylint (but not on test files so I'm nitpicking here)

chlowell (approved):

PR #16795, opened 2021-02-18
[text analytics] Updated tests failing in live mode

abhahn:
I used [this](https://dev.azure.com/azure-sdk/590cfd2a-581c-4dcb-a12e-6568ce786175/_apis/build/builds/736680/logs/126) log as a guide for which fixes to make.  

I was able to reproduce all of the failures except for `test_analyze.test_all_successful_passing_string_pii_entities_task` which is still passing for me when running in live mode locally, so more investigation may be needed to determine why it's failing.


sdk/textanalytics/azure-ai-textanalytics/tests/test_analyze_healthcare_async.py
```diff
@@ -294,7 +303,7 @@ async def test_out_of_order_ids(self, client):
             self.assertEqual(resp.id, expected_order[idx])
 
     @GlobalTextAnalyticsAccountPreparer()
-    @TextAnalyticsClientPreparer()
+    @TextAnalyticsClientPreparer(client_kwargs={"text_analytics_account_key": os.environ.get("AZURE_TEXT_ANALYTICS_KEY"), "text_analytics_account": os.environ.get("AZURE_TEXT_ANALYTICS_ENDPOINT")})
```
iscai-msft: hey @abhahn can you get rid of these environ variables in the future? i'm going to just push to this branch, but I think you adding this means your tests aren't fully set up correctly, can hop on a call today to figure this out more.

abhahn: Hey, looks like you did this already.  I have had some issues getting the tests to run correctly in WSL, so it might help for us to review what I'm doing to see if there's something wrong.

iscai-msft (approved):

PR #16809, opened 2021-02-18
Appconfig perf test

xiangyan99:


sdk/appconfiguration/azure-appconfiguration/tests/perfstress_tests/README.md
```diff
@@ -0,0 +1,43 @@
...
+(env) ~/azure-appconfiguration/tests> perfstress
+```
+Using the `perfstress` command alone will list the available perf tests found. 
+
+### Common perf command line options
+These options are available for all perf tests:
+- `--duration=10` Number of seconds to run as many operations (the "run" function) as possible. Default is 10.
+- `--iterations=1` Number of test iterations to run. Default is 1.
+- `--parallel=1` Number of tests to run in parallel. Default is 1.
+- `--no-client-share` Whether each parallel test instance should share a single client, or use their own. Default is False (sharing).
```
annatisch, line 35: The `no-client-share` option isn't actually a default option. If you want to support this you would need to add it to the perf test. You can see the implementation used for ServiceBus here:
https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/servicebus/azure-servicebus/tests/perf_tests/_test_base.py#L17-L46

It's up to you if you wish to support client sharing initially, though it might give a better indication of the perf of the client-wrapper/pipeline code without the overhead of the new connections with each test.
If you don't want to add it - you can simply undocument this cmd option :)

sdk/appconfiguration/azure-appconfiguration/tests/perfstress_tests/get_set.py
```diff
@@ -0,0 +1,42 @@
...
+
+class GetSetTest(PerfStressTest):
+    service_client = None
+    async_service_client = None
+
+    def __init__(self, arguments):
+        super().__init__(arguments)
+        connection_string = self.get_from_env("AZURE_APP_CONFIG_CONNECTION_STRING")
+        self.service_client = SyncAppConfigClient.from_connection_string(connection_string=connection_string)
+        self.async_service_client = AsyncAppConfigClient.from_connection_string(connection_string=connection_string)
```
annatisch, line 22: If you're planning to add a suite of tests for AppConfig, you can move this into a test base (like I linked to for ServiceBus above). However if you're only planning 1-2 tests, then it's probably simpler to keep as-is.

sdk/appconfiguration/azure-appconfiguration/tests/perfstress_tests/get_set.py
```diff
@@ -0,0 +1,42 @@
...
+
+    async def close(self):
+        # await self.async_service_client.close()
```
annatisch, line 25: I think this should be uncommented?

sdk/appconfiguration/azure-appconfiguration/tests/perfstress_tests/get_set.py
```diff
@@ -0,0 +1,42 @@
...
+        await super().close()
+
+    def run_sync(self):
+        kv = ConfigurationSetting(
+            key="KEY",
+            value="VALUE",
+        )
+        self.service_client.set_configuration_setting(kv)
+        self.service_client.get_configuration_setting(key=kv.key)
```
annatisch, line 34: Was it decided cross-language to use both these operations in a single perf test?
Typically for the other SDKs we have tried to make the perf test as singular as possible.
For example - if we determine that Python is running this scenario slower than JS, is it because of the Get operation or the Set operation?
I would break this into two separate tests, each one may need some setup/cleanup:
- Get config setting, where each class has setup function that does the Set operation with a UUID for key, then a cleanup operation to remove that setting
- Set config setting, where each class has a unique UUID key value to set, which is then cleaned up after completion.

sdk/appconfiguration/azure-appconfiguration/tests/perfstress_tests/get_set.py
```diff
@@ -0,0 +1,42 @@
...
+        self.async_service_client = AsyncAppConfigClient.from_connection_string(connection_string=connection_string)
+
+    async def close(self):
+        # await self.async_service_client.close()
+        await super().close()
+
+    def run_sync(self):
+        kv = ConfigurationSetting(
+            key="KEY",
+            value="VALUE",
```
annatisch, line 31: It looks like we're setting the same key/value for every run - do we know if there's a perf difference from the service if we are calling Set on an existing setting?

sdk/appconfiguration/azure-appconfiguration/tests/perfstress_tests/get_set.py
xiangyan99: Unfortunately, we don't implement close() for the async client. It is a bug and we will fix it soon. But for now, we cannot uncomment it. :(

xiangyan99: Btw, that's the reason I comment it rather than deleting it...

sdk/appconfiguration/azure-appconfiguration/tests/perfstress_tests/get_set.py
xiangyan99: No. This is my proposed scenario.

In app config, set and get almost do nothing other than call the API can translate the object.

The most (only) tricky part is to parse the sync-token.

So I want to try set (the service will generate a new sync-token) and then get (client will use the sync-token to get the new value).

This is open and your comments are more than welcome.

seankane-msft (commented):
LGTM :shipit:
Leaving up to Anna for final approval

sdk/appconfiguration/azure-appconfiguration/tests/perfstress_tests/set.py
```diff
@@ -0,0 +1,42 @@
...
+        self.async_service_client = AsyncAppConfigClient.from_connection_string(connection_string=connection_string)
+
+    async def close(self):
+        # await self.async_service_client.close()
+        await super().close()
+
+    def run_sync(self):
+        kv = ConfigurationSetting(
+            key=self.key,
+            value="VALUE" + str(uuid.uuid4()),
```
annatisch, line 33: Nit: We could move the `str(uuid.uuid4())` into the constructor to avoid generating it in the perf test

annatisch (approved):

PR #16840, opened 2021-02-20
[Key Vault] Add EncryptParameters analog in Python

mccoyp:
Adds `EncryptionAlgorithmConfiguration` and `DecryptionAlgorithmConfiguration` classes, which are accepted as an `algorithm` argument in `encrypt` and `decrypt`, respectively, to lead users to provide appropriate arguments to encryption algorithms.

### DIFF
--- a/sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/crypto/_models.py
+++ b/sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/crypto/_models.py
@@ -4,8 +4,10 @@
 # ------------------------------------
 from typing import TYPE_CHECKING
 
+from ._enums import EncryptionAlgorithm
+
 if TYPE_CHECKING:
-    from . import EncryptionAlgorithm, KeyWrapAlgorithm, SignatureAlgorithm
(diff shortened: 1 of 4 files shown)


sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/crypto/_models.py
```diff
@@ -6,7 +6,141 @@
...
+
+
+class EncryptionAlgorithmConfiguration():
+    """An encryption algorithm, including parameters applicable to using it for encryption.
+    """
+
+    def __init__(self, algorithm, **kwargs):
+        self.algorithm = algorithm
+        self.iv = kwargs.pop("iv", None)
+        self.additional_authenticated_data = kwargs.pop("additional_authenticated_data", None)
```
mccoyp, line 18: Question here: it's easy to imagine a user using these constructors directly. In that case, should we use the argument validation helper from `_client.py` in `__init__`? We could also validate arguments during calls to `encrypt`/`decrypt` like we do when an `EncryptionAlgorithm` is provided, but that wouldn't give us the chance to raise an error when the configuration is defined.

PR #16847, opened 2021-02-22
adds support for milliseconds on datetime objects

seankane-msft:


sdk/tables/azure-data-tables/azure/data/tables/_deserialize.py
```diff
@@ -96,11 +96,26 @@ def tzname(self, dt):
...
+    value = value.split(".")
+    ms = ""
+    if len(value) == 2:
+        ms = value[-1].replace("Z", "")
+        if len(ms) > 6:
+            ms = ms[:6]
+        ms = ms + "Z"
+        value = ".".join([value[0], ms])
+    else:
+        value = value[0]
```
annatisch, line 108: This logic is for filtering out the .Net extra decimal place right?
If that's the case - I would add comments for it... maybe even throw it into a separate function called `clean_up_dotnet_timestamps` just for fun ;)

sdk/tables/azure-data-tables/azure/data/tables/_deserialize.py
```diff
@@ -96,11 +96,26 @@ def tzname(self, dt):
...
+
+    try:
+        return datetime.datetime.strptime(value, "%Y-%m-%dT%H:%M:%S.%fZ").replace(
+            tzinfo=Timezone()
+        )
+    except ValueError:
+        # if value[-9:] == ".0000000Z":x
```
annatisch, line 115: Left over comment?

annatisch (approved):

PR #16859, opened 2021-02-22
Make default retryable methods configurable

### LINKED ISSUE #16284: Make default retryable methods configurable
RetryPolicy allows configuring which HTTP methods are retryable only for individual requests. If a client wants always to retry requests whose methods aren't in the policy's [hardcoded list of retryable methods](https://github.com/Azure/azure-sdk-for-python/blob/1ac1524f40e0095a1d5bc6294fa4619a32ce7ecc/sdk/core/azure-core/azure/core/pipeline/policies/_retry.py#L76), it must pass `retry_on_methods` with every request:
```py
response = self._pipeline.run(request, retry_on_methods=["POST"])
```
It would be more ergonomic and performant to allow configuring that list of retryable methods, for example:
```py
RetryPolicy(retry_on_methods=["GET", "POST"])
```

chlowell:
Closes #16284 by adding a `retry_on_methods` kwarg to Async/RetryPolicy, allowing a caller to set the default retryable verbs. This is useful when a client wants always to retry verbs that are not in the policy's hardcoded default list, such as POST (Azure AD token requests use POST):
```py
# before: pass the argument in every pipeline.run()
response = pipeline.run(request, retry_on_methods=["POST"])

# after: optionally set the argument at construction
policy = RetryPolicy(retry_on_methods=["POST"])
```


xiangyan99:
You may also need to update RetryPolicy.configure_retries().

chlowell:
I renamed "self._method_whitelist" to "self._retry_methods" in that method, did I miss something else that needs an update?

xiangyan99:
I was thinking should we have same code like

```py
methods = kwargs.pop('retry_on_methods', None) or ['HEAD', 'GET', 'PUT', 'DELETE', 'OPTIONS', 'TRACE']
self._retry_on_methods = frozenset(methods)
```

in config_retries()

Btw, to make sure I understand the scope of this change,

w/o the change, user needs:
```py
policy = RetryPolicy()
policy.configure_retries(retry_on_methods=["POST"])
```

and w/ the change, they only need
```py
policy = RetryPolicy(retry_on_methods=["POST"])
```

Is this correct?

chlowell:
`configure_retries` actually doesn't affect the policy's behavior. It just merges kwargs with the policy's default values into a new dictionary, which it returns ([see here](https://github.com/Azure/azure-sdk-for-python/blob/9f1a27bcfa7f10af53719b8699700b7254c75d66/sdk/core/azure-core/azure/core/pipeline/policies/_retry.py#L86)). Applying those kwargs to the policy's defaults could break code, and would contravene our convention of `pipeline.run()` kwargs affecting only a single request, so I think we should leave it alone.

Today, you can only configure retryable methods for a single request, with `pipeline.run(req, retry_on_methods=...)`. This PR makes it possible to configure them when creating the policy, without changing the behavior of `pipeline.run(req, retry_on_methods=...)`:
```py
# new configuration from this PR: policy will retry only POST by default
policy = RetryPolicy(retry_on_methods=["POST"])

# this request is retried iff it's a POST
pipeline.run(request)

# this also still works: policy will retry this request iff it's a GET
pipeline.run(request, retry_on_methods=["GET"])

# back to default: request is retried iff it's a POST
pipeline.run(request)
```

sdk/core/azure-core/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## 1.11.1 (Unreleased)
```
xiangyan99, line 3: 1.12.0 please. :)

xiangyan99:
Could you also update CLIENT_LIBRARY_DEVELOPER.md?

sdk/core/azure-core/CHANGELOG.md
chlowell: Should I update `__version__` also?

xiangyan99: Yes. They need to match.

PR #16886, opened 2021-02-23
[text analytics] default to API v4

iscai-msft:
Some test changes arose bc of differences in v4. Also deleted quite a few healthcare tests, so we can run all of our tests within the 2 hour nightly live test run


sdk/textanalytics/azure-ai-textanalytics/tests/test_analyze_healthcare.py
```diff
@@ -267,7 +133,7 @@ def test_show_stats_and_model_version(self, client):
 
         self.assertIsNotNone(response)
         self.assertIsNotNone(response.model_version)
-        self.assertEqual("2021-01-11", response.model_version)
+        self.assertIsNotNone(response.model_version)
```
abhahn, line 136: Duplicate line

sdk/textanalytics/azure-ai-textanalytics/tests/test_analyze_healthcare.py
```diff
@@ -189,68 +114,9 @@ def test_output_same_order_as_input(self, client):
...
-            {"id": "1", "text": "Microsoft was founded by Bill Gates and Paul Allen."},
-            TextDocumentInput(id="2", text="I did not like the hotel we stayed at. It was too expensive."),
-            u"You cannot mix string input with the above inputs"
-        ]
-        with self.assertRaises(TypeError):
-            response = client.begin_analyze_healthcare_entities(docs, polling_interval=self._interval())
-
-    @GlobalTextAnalyticsAccountPreparer()
-    @TextAnalyticsClientPreparer()
-    def test_out_of_order_ids(self, client):
```
abhahn, line 238: We could consider keeping only this test (instead of the other one above testing the ordering) in order to also ensure that we are always sorting based on the input order rather than naively (i.e. alphanumerically).

iscai-msft: good point, i'll just get rid of the alphanumeric one then

sdk/textanalytics/azure-ai-textanalytics/tests/test_analyze_healthcare.py
iscai-msft: oh wait, this actually isn't a duplicated line, I just replaced `self.assertEqual("2021-01-11", response.model_version)` with `self.assertIsNotNone(response.model_version)`

iscai-msft: nvm i'm actually going to go with that specific model version, since we actually pass the model version in during our call

abhahn (approved):
Looks good, thanks!

PR #16925, opened 2021-02-25
adjust the filecount in integration stage so we don't attempt to push non-alpha packages 

scbedd:
I have verified with a local copy of an artifact set. Properly filters out non-alpha packages.

@seankane-msft  when you check it an approve, can you check-enforcer override? No checks will trigger for this boyo.

### DIFF
--- a/eng/pipelines/templates/stages/archetype-python-release.yml
+++ b/eng/pipelines/templates/stages/archetype-python-release.yml
@@ -286,7 +286,8 @@ stages:
               $packageDirectory = "${{artifact.name}}".Replace("_", "-")
               echo "##vso[task.setvariable variable=Package.Name]$packageDirectory"
           - pwsh: |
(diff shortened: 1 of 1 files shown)


seankane-msft (approved):

eng/pipelines/templates/stages/archetype-python-release.yml
```diff
@@ -286,7 +286,8 @@ stages:
               $packageDirectory = "${{artifact.name}}".Replace("_", "-")
               echo "##vso[task.setvariable variable=Package.Name]$packageDirectory"
           - pwsh: |
-              $fileCount = (Get-ChildItem $(Pipeline.Workspace)/${{parameters.ArtifactName}}/$packageDirectory  | Measure-Object).Count
+              $fileCount = (Get-ChildItem $(Pipeline.Workspace)/${{parameters.ArtifactName}}/$packageDirectory -Filter $packageDirectory-[0-9]*.[0-9]*.[0-9]*a[0-9]*  | Measure-Object).Count
```
weshaggard, line 289: @scbedd why was this an issue with the artifact refactoring? I would have expected it to be an issue in either case.

Issue #16910, opened 2021-02-24
Rethink the credential we used in Python test source code and recording files


sima-zhu:
**Current Status**
We already setup the cred scan steps in aggregate-report, and we can check the error report here: [link](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=750499&view=sariftools.scans.build-tab)
However, team keeps checked in new files containing new keys which failed the pipeline.

**Open Issues to address**
We have the [suppression file](https://github.com/Azure/azure-sdk-for-python/blob/master/eng/CredScanSuppression.json) to suppress some of the existing issues. 
It is a huge effort if we suppress all newly detected keys in suppression file, since most reported cases are false positive. However, blindly suppress the key probably hide the real issues. Therefore, we have to rethink how we generate those keys and reuse them as many as possible.

**Facts**
Currently, we basically have 3 types of errors:
* Placeholder key/secret in test src code. E.g password = "123".
We'd better to reuse the fake key inside of the suppression file instead of generating new ones for new tests.
* Files only which contain the key, mostly appears in keyvault and identity. E.g client.jks, pemCert.pem
If any new service needs to reuse the same key, it is better to move the key files into some common space, so we do not need to suppress them one by one for each service.
* (Important) SAS, share key, web hook token, hashkey etc
For this category, we saw a lot of random generalized key/secret/token. Team is better to review how each service generates those keys, and whether we can share the same type of key for all services, so that we can reuse and suppress certain amount of keys only. Can we reuse redact string instead of real ones?

seankane-msft:
Update 2/26: Storage is in the process of adding `self.scrubber.register(key, <scrubbed_value>)` which will remove ~40% of these issues.

seankane-msft:
Update 3/2: Added several `fake_mgmt_settings.py` files to be ignored.

mccoyp:
We now have a strategy for handling these scenarios that should satisfy this issue. There's a central, shared [fake_credentials.py](https://github.com/Azure/azure-sdk-for-python/blob/main/tools/azure-sdk-tools/devtools_testutils/fake_credentials.py) file where multiple packages can share sanitization strings, and this is being publicized by a [CredScan process guide](https://github.com/Azure/azure-sdk-for-python/blob/main/doc/dev/credscan_process.md) that describes best practices for resolving warnings. There's an ongoing effort to make `fake_credentials.py` use more widespread, but the process is in place.

PR #16944, opened 2021-02-25
Addressing issues with CredScan

seankane-msft:
#16910

### DIFF
--- a/eng/CredScanSuppression.json
+++ b/eng/CredScanSuppression.json
@@ -35,9 +35,11 @@
         {
             "file":[
                 "sdk/keyvault/azure-keyvault-certificates/tests/ca.key",
-                "sdk/identity/azure-identity/tests/ec-certificate.pem"
+                "sdk/identity/azure-identity/tests/ec-certificate.pem",
+                "sdk/core/azure-servicemanagement-legacy/tests/legacy_mgmt_settings_fake.py",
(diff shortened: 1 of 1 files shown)


eng/CredScanSuppression.json
```diff
@@ -35,9 +35,13 @@
         {
             "file":[
                 "sdk/keyvault/azure-keyvault-certificates/tests/ca.key",
-                "sdk/identity/azure-identity/tests/ec-certificate.pem"
+                "sdk/identity/azure-identity/tests/ec-certificate.pem",
+                "sdk/core/azure-servicemanagement-legacy/tests/legacy_mgmt_settings_fake.py",
```
weshaggard, line 39: We want to be careful about excluding entire files because that means if there ever becomes an individual secret in one of them that we will not detect it. We are trying to only exclude files if the entire file is a fake key/certificate that is used for testing.

sima-zhu: I have checked this file. The purpose of this file is to add fake key in common place and use variables in tests. If people do not abuse the file with real key, then it is supposed to suppress in this way.

weshaggard: OK yeah I think those cases are also reasonable. The approach also might be interesting for other languages if they need something similar.

sima-zhu (approved):
LGTM

PR #16949, opened 2021-02-25
Add perf tests for eventgrid

rakshith91:
Fixes https://github.com/Azure/azure-sdk-for-python/issues/16541

### DIFF
--- a/sdk/eventgrid/azure-eventgrid/tests/perfstress_tests/README.md
+++ b/sdk/eventgrid/azure-eventgrid/tests/perfstress_tests/README.md
@@ -30,7 +30,6 @@ These options are available for all perf tests:
 - `--duration=10` Number of seconds to run as many operations (the "run" function) as possible. Default is 10.
 - `--iterations=1` Number of test iterations to run. Default is 1.
 - `--parallel=1` Number of tests to run in parallel. Default is 1.
-- `--no-client-share` Whether each parallel test instance should share a single client, or use their own. Default is False (sharing).
 - `--warm-up=5` Number of seconds to spend warming up the connection before measuring begins. Default is 5.
 - `--sync` Whether to run the tests in sync or async. Default is False (async).
 - `--no-cleanup` Whether to keep newly created resources after test run. Default is False (resources will be deleted).
--- a/sdk/eventgrid/azure-eventgrid/tests/perfstress_tests/README.md
+++ b/sdk/eventgrid/azure-eventgrid/tests/perfstress_tests/README.md
@@ -1,0 +1,49 @@
+# EventGrid Performance Tests
+
+In order to run the performance tests, the `azure-devtools` package must be installed. This is done as part of the `dev_requirements`.
+Start by creating a new virtual environment for your perf tests. This will need to be a Python 3 environment, preferably >=3.7.
+
+### Setup for test resources
+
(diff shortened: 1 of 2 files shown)


sdk/eventgrid/azure-eventgrid/tests/perfstress_tests/README.md
```diff
@@ -0,0 +1,49 @@
...
+```
+
+### Common perf command line options
+These options are available for all perf tests:
+- `--duration=10` Number of seconds to run as many operations (the "run" function) as possible. Default is 10.
+- `--iterations=1` Number of test iterations to run. Default is 1.
+- `--parallel=1` Number of tests to run in parallel. Default is 1.
+- `--no-client-share` Whether each parallel test instance should share a single client, or use their own. Default is False (sharing).
+- `--warm-up=5` Number of seconds to spend warming up the connection before measuring begins. Default is 5.
+- `--sync` Whether to run the tests in sync or async. Default is False (async). This flag must be used for Storage legacy tests, which do not support async.
```
rakshith91, line 35:
```suggestion
- `--sync` Whether to run the tests in sync or async. Default is False (async).
```

sdk/eventgrid/azure-eventgrid/tests/perfstress_tests/README.md
```diff
@@ -0,0 +1,49 @@
...
+(env) ~/azure-eventgrid> cd tests
+(env) ~/azure-eventgrid/tests> perfstress
+```
+
+### Common perf command line options
+These options are available for all perf tests:
+- `--duration=10` Number of seconds to run as many operations (the "run" function) as possible. Default is 10.
+- `--iterations=1` Number of test iterations to run. Default is 1.
+- `--parallel=1` Number of tests to run in parallel. Default is 1.
+- `--no-client-share` Whether each parallel test instance should share a single client, or use their own. Default is False (sharing).
```
annatisch, line 33: You can remove this option unless you want to implement sharing a single client between test instances.
Probably not needed for now.

sdk/eventgrid/azure-eventgrid/tests/perfstress_tests/send.py
```diff
@@ -0,0 +1,69 @@
...
+            credential=AzureKeyCredential(topic_key)
+        )
+
+        services = ["EventGrid", "ServiceBus", "EventHubs", "Storage"]
+        self.event_list = []
+        for _ in range(self.args.num_events):
+            self.event_list.append(EventGridEvent(
+                event_type="Contoso.Items.ItemReceived",
+                data={
+                    "services": random.sample(services, k=random.randint(1, 4))
```
annatisch, line 34: This will test a different data set every time - we should probably keep each test consistent with throughput.
If you expect there to be variation in consuming each of these values that you want to compare - you could make it a cmd flag:
```python
parser.add_argument('-e', '--event-service', nargs='?', type=str, help='The event service type. Default is "EventGrib"', default='EventGrid')
```
But otherwise I think we can just stick to a single value.

annatisch (approved):

sdk/eventgrid/azure-eventgrid/tests/perfstress_tests/send.py
rakshith91: sticking to single value should be good

PR #16968, opened 2021-02-26
Update azure-core minimum version

craigktreasure:
- This change updates the minimum version of Azure Core to 1.4.0 to make sure we get a package version that includes AzureKeyCredential to fix the nightly release builds.

### DIFF
--- a/sdk/mixedreality/azure-mixedreality-authentication/setup.py
+++ b/sdk/mixedreality/azure-mixedreality-authentication/setup.py
@@ -66,7 +66,7 @@ setup(
         'azure.mixedreality'
     ]),
     install_requires=[
-        'azure-core<2.0.0,>=1.2.2',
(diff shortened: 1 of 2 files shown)


seankane-msft (approved):
LGTM :shipit:

craigktreasure:
@seankane-msft Build failed with:
```
The following libraries declare requirement 'azure-core<2.0.0,>=1.4.0' which does not match the frozen requirement 'azure-core<2.0.0,>=1.2.2':
  * azure-mixedreality-authentication
```

Any suggestions?

scbedd:
@craigktreasure I'll come up with something for you. This error doesn't make sense.

craigktreasure:
Thanks @scbedd. Looks like that did the trick. Hopefully that version has everything I need.

PR #16972, opened 2021-02-26
[core] add HttpRequest and HttpResponse reprs

### LINKED ISSUE #16952: [core] add good reprs for our HttpRequest and HttpResponses

iscai-msft:
fixes #16952


sdk/core/azure-core/azure/core/pipeline/transport/_aiohttp.py
```diff
@@ -267,7 +267,7 @@ def __init__(self, request: HttpRequest, aiohttp_response: aiohttp.ClientRespons
         self.status_code = aiohttp_response.status
         self.headers = CIMultiDict(aiohttp_response.headers)
         self.reason = aiohttp_response.reason
-        self.content_type = aiohttp_response.headers.get('content-type')
+        self.content_type = self.headers.get('content-type')
```
iscai-msft, line 270: callout: I think this should be `self.headers` instead of `aiohttp_response.headers`, bc we made `self.headers` case insensitive

johanste: Technically, from a functional perspective, this shouldn't matter since `aiohttp_response.headers` is also case-insensitive.

iscai-msft: yeah, @lmazuel brought that up too. tbh, i'll just go ahead and delete it, it doesn't need to be there

sdk/core/azure-core/azure/core/pipeline/transport/_base.py
```diff
@@ -592,6 +594,15 @@ def raise_for_status(self):
         if self.status_code >= 400:
             raise HttpResponseError(response=self)
 
+    def __repr__(self):
+        # there doesn't have to be a content type
+        content_type_str = (
```
johanste, line 599: Nitpick: I would exclude the content type key/value altogether in the repr if we don't know what it is (i.e. not add `, Content-Type: ` to the output...

kristapratico (approved):
Verbal confirmation of approval for this PR was given, I'm just approving so we can merge.

iscai-msft:
thanks @kristapratico , @johanste and @lmazuel have already given their verbal approval

xiangyan99 (approved):

PR #16978, opened 2021-02-27
Update folder name to correct variable format, update filtering logic in integration stage

scbedd:


praveenkuttappan (approved):
Looks good

Issue #14994, opened 2020-11-04
Ability to traverse blob hierarchy upwards. BlobClient->BlobContainer


kasobol-msft:
Propagating https://github.com/Azure/azure-sdk-for-net/issues/16359 from .NET SDK for parity.

PR #16981, opened 2021-02-28
Added support for BlobClient -> ContainerClient -> BlobServiceClient Traversal 

tasherif-msft:
This resolves https://github.com/Azure/azure-sdk-for-python/issues/14994

### DIFF
--- a/sdk/storage/azure-storage-blob/azure/storage/blob/_blob_client.py
+++ b/sdk/storage/azure-storage-blob/azure/storage/blob/_blob_client.py
@@ -18,11 +18,12 @@ except ImportError:
     from urllib2 import quote, unquote  # type: ignore
 
 import six
+from azure.core.pipeline import Pipeline
 from azure.core.tracing.decorator import distributed_trace
 from azure.core.exceptions import ResourceNotFoundError, HttpResponseError
 
 from ._shared import encode_base64
-from ._shared.base_client import StorageAccountHostsMixin, parse_connection_str, parse_query
+from ._shared.base_client import StorageAccountHostsMixin, parse_connection_str, parse_query, TransportWrapper
 from ._shared.encryption import generate_blob_encryption_data
 from ._shared.uploads import IterStreamer
 from ._shared.request_handlers import (
@@ -3765,8 +3766,16 @@ class BlobClient(StorageAccountHostsMixin):  # pylint: disable=too-many-public-m
                 :caption: Get container client from blob object.
         """
         from ._container_client import ContainerClient
+        if not isinstance(self._pipeline._transport, TransportWrapper):
+            _pipeline = Pipeline(
+                transport=TransportWrapper(self._pipeline._transport), # pylint: disable = protected-access
+                policies=self._pipeline._impl_policies # pylint: disable = protected-access
+            )
+        else:
+            _pipeline = self._pipeline   # pylint: disable = protected-access
         return ContainerClient(
             "{}://{}".format(self.scheme, self.primary_hostname), container_name=self.container_name,
             credential=self._raw_credential, api_version=self.api_version, _configuration=self._config,
-            _location_mode=self._location_mode, _hosts=self._hosts, require_encryption=self.require_encryption,
-            key_encryption_key=self.key_encryption_key, key_resolver_function=self.key_resolver_function)
+            _pipeline=_pipeline, _location_mode=self._location_mode, _hosts=self._hosts,
+            require_encryption=self.require_encryption, key_encryption_key=self.key_encryption_key,
+            key_resolver_function=self.key_resolver_function)
--- a/sdk/storage/azure-storage-blob/azure/storage/blob/_blob_client.py
+++ b/sdk/storage/azure-storage-blob/azure/storage/blob/_blob_client.py
@@ -3766,7 +3766,7 @@ class BlobClient(StorageAccountHostsMixin):  # pylint: disable=too-many-public-m
                 :caption: Get container client from blob object.
         """
         from ._container_client import ContainerClient
-        if not isinstance(self._pipeline._transport, TransportWrapper):
+        if not isinstance(self._pipeline._transport, TransportWrapper): # pylint: disable = protected-access
             _pipeline = Pipeline(
                 transport=TransportWrapper(self._pipeline._transport), # pylint: disable = protected-access
                 policies=self._pipeline._impl_policies # pylint: disable = protected-access
--- a/sdk/storage/azure-storage-blob/azure/storage/blob/_blob_client.py
+++ b/sdk/storage/azure-storage-blob/azure/storage/blob/_blob_client.py
@@ -167,7 +167,7 @@ class BlobClient(StorageAccountHostsMixin):  # pylint: disable=too-many-public-m
                 self.snapshot = snapshot or path_snapshot
 
         # This parameter is used for the hierarchy traversal. Give precedence to credential.
-        self.client_credential = credential if credential else sas_token
+        self._raw_credential = credential if credential else sas_token
         self._query_str, credential = self._format_query_string(sas_token, credential, snapshot=self.snapshot)
         super(BlobClient, self).__init__(parsed_url, service='blob', credential=credential, **kwargs)
         self._client = AzureBlobStorage(self.url, pipeline=self._pipeline)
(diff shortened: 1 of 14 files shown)


sdk/storage/azure-storage-blob/azure/storage/blob/_blob_client.py
```diff
@@ -164,6 +166,8 @@ def __init__(
             except TypeError:
                 self.snapshot = snapshot or path_snapshot
 
+        # This parameter is used for the hierarchy traversal. Give precedence to credential.
+        self.client_credential = credential if credential else sas_token
```
xiafu-msft, line 170: why don't we use self.credential on base_client?

tasherif-msft: Because it fails in the following case:
1. User creates a blob client with a SAS token (`BlobClient(.., credential=sas,..)`)
2. Eventually the method `self._format_query_string(sas_token, credential, snapshot=self.snapshot)` is called
3. This method would set `credential=None` and add the sas token to the URI/query parameters
4. So when we call `self.credential` we would get None which would create a ContainerClient with no credential

xiafu-msft: My concern is user will see self.credential and self.client_credential on the same object

sdk/storage/azure-storage-blob/azure/storage/blob/_blob_client.py
```diff
@@ -3741,3 +3745,39 @@ def seal_append_blob(self, **kwargs):
...
+            .. literalinclude:: ../samples/blob_samples_containers.py
+                :start-after: [START get_container_client_from_blob_client]
+                :end-before: [END get_container_client_from_blob_client]
+                :language: python
+                :dedent: 8
+                :caption: Get container client from blob object.
+        """
+        from ._container_client import ContainerClient
+        _pipeline = Pipeline(
+            transport=TransportWrapper(self._pipeline._transport), # pylint: disable = protected-access
```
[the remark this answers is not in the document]
tasherif-msft: I'm not sure I follow, if a user gets a container client from a bsc we wrap and then if we get a blob client from that container client then we wrap again - isn't the same as this case except we're just going upwards rather than downwards?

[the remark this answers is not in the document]
xiafu-msft: oh I didn't paste the link successfully https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/storage/azure-storage-blob/azure/storage/blob/_shared/authentication.py#L81-L83
if you run test_create_blob_with_equal_sign and put debug point on line 81 probably you would know what I mean

sdk/storage/azure-storage-blob/azure/storage/blob/_blob_client.py
xiafu-msft: probably make it self._raw_credential?

sdk/storage/azure-storage-blob/azure/storage/blob/_blob_client.py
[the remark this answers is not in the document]
tasherif-msft: OH haha I see why it works in the case of bsc -> cc -> bc 
Hmmm the problem is if we don't wrap there will be issues with context managers - I think. I will investigate this more

sdk/storage/azure-storage-blob/azure/storage/blob/_blob_client.py
xiafu-msft: probably we can keep raw url, which probably has the sas, while later we can just pass the raw url back to ContainerClient

sdk/storage/azure-storage-blob/azure/storage/blob/_blob_client.py
```diff
@@ -3741,3 +3745,39 @@ def seal_append_blob(self, **kwargs):
...
+            policies=self._pipeline._impl_policies # pylint: disable = protected-access
+        )
+
+        return ContainerClient(
+            "{}://{}".format(self.scheme, self.primary_hostname), container_name=self.container_name,
```
xiafu-msft, line 3779: you can pass the raw url here and don't need to worry if credential is None

tasherif-msft: raw_url would fail in the case:
If the user wants to change the credential from SAS to AzureSasCredential when getting their container client. It would throw:
```
        if sas_token and isinstance(credential, AzureSasCredential):
            raise ValueError(
                "You cannot use AzureSasCredential when the resource URI also contains a Shared Access Signature.")
```
Therefore we needed this new parameter I added and passing the URI with only scheme and primary hostname

sdk/storage/azure-storage-blob/azure/storage/blob/_blob_client.py
tasherif-msft: Comment on this below

sdk/storage/azure-storage-blob/azure/storage/blob/_blob_client.py
xiafu-msft: why does the user want to pass another credential when get the parent client?

tasherif-msft: I feel this extra flexibility would be nice and doesn't cost us much. This would give the user to change the credential on the fly if they choose to rather than instantiating a whole new container client just because they want to change credentials.

sdk/storage/azure-storage-blob/azure/storage/blob/_blob_client.py
```diff
@@ -3741,3 +3745,39 @@ def seal_append_blob(self, **kwargs):
...
             process_storage_error(error)
+
+    @distributed_trace
+    def get_container_client(self, **kwargs):
+        # type: (...) -> ContainerClient
+        """Get a client to interact with the blob's parent container.
+
+        The container need not already exist. Defaults to current blob's credentials.
+
+        :keyword credential:
```
xiafu-msft, line 3756: @kasobol-msft @annatisch how do you feel about having credential option when get_container_client from blob client?

sdk/storage/azure-storage-blob/azure/storage/blob/_blob_client.py
xiafu-msft: while if you put credential option there, that would cause confusion for users who are using shared key or AAD credential, they would think why they can pass another credential here

sdk/storage/azure-storage-blob/azure/storage/blob/_blob_client.py
kasobol-msft: I think we should keep it simple. It's a reverse operation to the `container.get_blob_client`, so it seems that this should be parameter-less/option-less ?

https://github.com/Azure/azure-sdk-for-python/blob/6286f95b28942b25da258ea45eba8c00c386f176/sdk/storage/azure-storage-blob/azure/storage/blob/_container_client.py#L1474-L1500

kasobol-msft: btw. it's not [yagni](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it)

sdk/storage/azure-storage-blob/azure/storage/blob/_blob_client.py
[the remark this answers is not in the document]
tasherif-msft: I have updated the PR, the methods now create a newly constructed pipeline (and transport) - this means that their context is independent from the child class that created them. 
The other approach was to pass the child class transport object (without wrapping)  - this means that in context managers would get closed if the child got closed (if the child was not wrapped) and would always remain open if the child class was wrapped (since child classes only close when their parent closes)

I felt the best approach would just be to create a new pipeline when these constructors are used. Let me know what you think

xiafu-msft (approved):

kasobol-msft (approved):

sdk/storage/azure-storage-blob/azure/storage/blob/_blob_client.py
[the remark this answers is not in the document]
xiafu-msft: By creating new pipeline and transport object, we will not be able to reuse the existing connection pool.
Can we check if it's a transportwrapper and if it's already a transport wrapper we can directly reuse the pipeline? by specifying self._pipeline.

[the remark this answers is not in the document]
tasherif-msft: I thought of this before, my concern with this is that if we spin up BSC from CC for example, then if CC is closed the BSC will get closed as well. Is this ok? - As I'm writing this I realized that its ok to do that actually

PR #16982, opened 2021-02-28
Update OT to 1.0.0

rakshith91:
Fixes https://github.com/Azure/azure-sdk-for-python/issues/17539

usage would be like this:
```
from azure.core.tracing.ext.opentelemetry_span import OpenTelemetrySpan
from azure.core.settings import settings
from azure.core.tracing import Link
settings.tracing_implementation = OpenTelemetrySpan
span_impl = settings.tracing_implementation()

links = Link({'traceparent': '2346782uehdcfhvj'})

span = span_impl(name="example", kind=SpanKind.CLIENT,  links=[links])
```

### DIFF
--- a/sdk/core/azure-core-tracing-opencensus/azure/core/tracing/ext/opencensus_span/__init__.py
+++ b/sdk/core/azure-core-tracing-opencensus/azure/core/tracing/ext/opencensus_span/__init__.py
@@ -46,8 +46,8 @@ _config_integration.trace_integrations(['threading'])
 class OpenCensusSpan(HttpSpanMixin, object):
     """Wraps a given OpenCensus Span so that it implements azure.core.tracing.AbstractSpan"""
 
-    def __init__(self, span=None, name="span"):
-        # type: (Optional[Span], Optional[str]) -> None
+    def __init__(self, span=None, name="span", **kwargs):
+        # type: (Optional[Span], Optional[str], Any) -> None
         """
         If a span is not passed in, creates a new tracer. If the instrumentation key for Azure Exporter is given, will
         configure the azure exporter else will just create a new tracer.
@@ -58,7 +58,7 @@ class OpenCensusSpan(HttpSpanMixin, object):
         :type name: str
         """
         tracer = self.get_current_tracer()
-        self._span_instance = span or tracer.start_span(name=name)
+        self._span_instance = span or tracer.start_span(name=name, **kwargs)
 
     @property
     def span_instance(self):
--- a/sdk/core/azure-core-tracing-opencensus/azure/core/tracing/ext/opencensus_span/__init__.py
+++ b/sdk/core/azure-core-tracing-opencensus/azure/core/tracing/ext/opencensus_span/__init__.py
@@ -56,6 +56,9 @@ class OpenCensusSpan(HttpSpanMixin, object):
         :type span: :class: opencensus.trace.Span
         :param name: The name of the OpenCensus span to create if a new span is needed
         :type name: str
+        :keyword SpanKind kind: The span kind of this span.
+        :keyword links: The list of links to be added to the span.
+        :paramtype links: list[~azure.core.tracing.Link]
         """
         tracer = self.get_current_tracer()
         self._span_instance = span or tracer.start_span(name=name, **kwargs)
@@ -74,6 +77,9 @@ class OpenCensusSpan(HttpSpanMixin, object):
         Create a child span for the current span and append it to the child spans list in the span instance.
         :param name: Name of the child span
         :type name: str
+        :keyword SpanKind kind: The span kind of this span.
+        :keyword links: The list of links to be added to the span.
+        :paramtype links: list[~azure.core.tracing.Link]
         :return: The OpenCensusSpan that is wrapping the child span instance
         """
         return self.__class__(name=name, **kwargs)
--- a/sdk/core/azure-core-tracing-opencensus/azure/core/tracing/ext/opencensus_span/__init__.py
+++ b/sdk/core/azure-core-tracing-opencensus/azure/core/tracing/ext/opencensus_span/__init__.py
@@ -68,15 +68,15 @@ class OpenCensusSpan(HttpSpanMixin, object):
         """
         return self._span_instance
 
-    def span(self, name="span"):
-        # type: (Optional[str]) -> OpenCensusSpan
+    def span(self, name="span", **kwargs):
+        # type: (Optional[str], Any) -> OpenCensusSpan
         """
         Create a child span for the current span and append it to the child spans list in the span instance.
         :param name: Name of the child span
         :type name: str
         :return: The OpenCensusSpan that is wrapping the child span instance
         """
-        return self.__class__(name=name)
+        return self.__class__(name=name, **kwargs)
 
     @property
     def kind(self):
--- a/sdk/core/azure-core-tracing-opentelemetry/azure/core/tracing/ext/opentelemetry_span/__init__.py
+++ b/sdk/core/azure-core-tracing-opentelemetry/azure/core/tracing/ext/opentelemetry_span/__init__.py
@@ -120,7 +120,7 @@ class OpenTelemetrySpan(HttpSpanMixin, object):
         )
         if kind is None:
             raise ValueError("Kind {} is not supported in OpenTelemetry".format(value))
-        self.span_instance.kind = kind
+        self.span_instance.set_attribute("kind", kind)
 
     def __enter__(self):
         """Start a span."""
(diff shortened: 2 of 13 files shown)


sdk/core/azure-core-tracing-opencensus/azure/core/tracing/ext/opencensus_span/__init__.py
```diff
@@ -68,15 +68,15 @@ def span_instance(self):
         """
         return self._span_instance
 
-    def span(self, name="span"):
-        # type: (Optional[str]) -> OpenCensusSpan
+    def span(self, name="span", **kwargs):
```
johanste, line 71: Not a new issue, but I'm a bit surprised that this isn't a `@classmethod`...

johanste, line 71: You are missing doccomment on what `kwargs` are used for...

rakshith91: i think the reason is this is supposed to create a child  span  - so `parent_span.span` is more intuitive than `SpanImplementation.span`? not sure what's the thought process behind it when abstract span was created

sdk/core/azure-core-tracing-opentelemetry/azure/core/tracing/ext/opentelemetry_span/__init__.py
```diff
@@ -150,7 +153,7 @@ def to_header(self): # pylint: disable=no-self-use
         :return: A key value pair dictionary
         """
         temp_headers = {} # type: Dict[str, str]
-        inject(_set_headers_from_http_request_headers, temp_headers)
+        inject(temp_headers)
```
johanste, line 156: I don't quite understand what the purpose of this is?

sdk/core/azure-core/azure/core/tracing/_abstract_span.py
```diff
@@ -45,15 +45,15 @@ class SpanKind(Enum):
 class AbstractSpan(Protocol):
     """Wraps a span from a distributed tracing implementation."""
 
-    def __init__(self, span=None, name=None):  # pylint: disable=super-init-not-called
-        # type: (Optional[Any], Optional[str]) -> None
+    def __init__(self, span=None, name=None, **kwargs):  # pylint: disable=super-init-not-called
```
johanste, line 48: This is a breaking change - although it is unlikely that anyone has actually created custom span types...

sdk/core/azure-core-tracing-opentelemetry/azure/core/tracing/ext/opentelemetry_span/__init__.py
rakshith91: it's to propagate values to a http client

sdk/core/azure-core/azure/core/tracing/_abstract_span.py
rakshith91: how is adding kwargs breaking?

sdk/core/azure-core/azure/core/tracing/_abstract_span.py
```diff
@@ -129,6 +129,18 @@ def span_instance(self):
         Returns the span the class is wrapping.
         """
 
+    @staticmethod
```
johanste, line 132: Why does a protocol type has a static method? We generally try to avoid `@staticmethod`s, and having it on a protocol seem doubly odd? Especially since it doesn't have any documentation on what it is supposed to return other than a "link"?

rakshith91: This method is supposed to create a link given headers so that it can be passed to the child span while creating the span instance. 

the other idea i had is have a custom `Link` class in azure-core exposed along with abstract span.

```
class Link:
  def __init__(self, headers, attributes):
    self.headers = headers
    self.attributes = attributes
```

and ask the users to use this instead of OT's Link
```
from azure.core.tracing import Link
link = Link(headers, attributes)
```
internally, in the OT span __init__, we change this into Opentelemetry's span.k

This felt like an overkill personally
- and right now, this method only exists to create a lin

sdk/core/azure-core-tracing-opencensus/azure/core/tracing/ext/opencensus_span/__init__.py
johanste: You can call `@classmethod`s through an instance, though. But that is not a change we can make now...

shared_requirements.txt
```diff
@@ -158,8 +158,8 @@ avro<2.0.0,>=1.10.0
 opencensus>=0.6.0
 opencensus-ext-threading
 opencensus-ext-azure>=0.3.1
-opentelemetry-api==0.17b0
-opentelemetry-sdk==0.17b0
+opentelemetry-api==1.0.0
```
johanste, line 161: Why are we pinning an exact version?

sdk/core/azure-core-tracing-opentelemetry/setup.py
```diff
@@ -57,9 +56,9 @@
     packages=[
         'azure.core.tracing.ext.opentelemetry_span',
     ],
-    python_requires=">=3.5.0",
+    python_requires=">=3.6.0",
     install_requires=[
-        'opentelemetry-api==0.17b0',
+        'opentelemetry-api==1.0.0',
```
johanste, line 61: Why are we pinning an exact version?

sdk/core/azure-core-tracing-opentelemetry/setup.py
```diff
@@ -57,9 +56,9 @@
     packages=[
         'azure.core.tracing.ext.opentelemetry_span',
     ],
-    python_requires=">=3.5.0",
+    python_requires=">=3.6.0",
```
johanste, line 59: We are indicating that we support Python 2/2.7 in our classifiers, yet we require python>=3.6. This seem odd...

rakshith91: opentelemetry requires python 3.6 or above - there is no python 2 support for the OT plugin

sdk/core/azure-core-tracing-opentelemetry/setup.py
rakshith91: so far we have pinned since it was in beta and things were breaking randomly - let's keep it this way till monitor exporter is released and we come up with a plan on how to keep both versions identical

shared_requirements.txt
rakshith91: so far we have pinned since it was in beta and things were breaking randomly - let's keep it this way till monitor exporter is released and we come up with a plan on how to keep both versions identical

sdk/core/azure-core-tracing-opentelemetry/azure/core/tracing/ext/opentelemetry_span/__init__.py
```diff
@@ -39,37 +39,50 @@
...
+
+        links = kwargs.pop('links', None)
+        if links:
+            try:
+                ot_links = []
+                for link in links:
+                    ctx = extract(link.headers)
+                    span_ctx = get_span_from_context(ctx).get_span_context()
+                    ot_links.append(OpenTelemetryLink(span_ctx, link.attributes))
+                    kwargs.setdefault('links', ot_links)
```
johanste, line 80: Why is this done inside the loop? Looks like it should be outside, no?

rakshith91, line 80:
```suggestion
                kwargs.setdefault('links', ot_links)
```

rakshith91: this should be outside - moved it (y)

sdk/core/azure-core-tracing-opentelemetry/setup.py
johanste: We pin preview versions since they don't follow semver. We don't pin GA versions for libraries that do follow semver since that locks the application to a very specific version. Do we have specific reasons to believe that OpenTelemetry will not follow semver?

rakshith91: no - OT will follow the semver - un-pinned it

sdk/core/azure-core-tracing-opencensus/azure/core/tracing/ext/opencensus_span/__init__.py
rakshith91: added

sdk/core/azure-core-tracing-opentelemetry/setup.py
xiangyan99: Do you mean if customers are using 2.7, they are completely broken?

rakshith91: customers cannot use 2.7 with opentelemtry - they should use opencencus for 3.5 and less

johanste (approved):

Issue #16992, opened 2021-03-01
Method for removing SAS from recordings


seankane-msft:
SAS are stored in plain text in the recordings currently. Adding the SAS token with a `self.scrubber.register_name_pair(token, "fake_token")` removes the token from the recording but the new URL will fail when running in playback mode.

PR #17024, opened 2021-03-02
Scrub sas

seankane-msft:
#16992


tools/azure-sdk-tools/devtools_testutils/azure_testcase.py
```diff
@@ -347,4 +352,45 @@ def run(test_class_instance, *args, **kwargs):
...
+            elif part.startswith("st"):
+                key = part.split("=")
+                key[1] = "start"
+                parts[idx] = "=".join(key)
+            elif part.startswith("se"):
+                key = part.split("=")
+                key[1] = "end"
+                parts[idx] = "=".join(key)
+
+        return "&".join(parts)
```
seankane-msft, line 396: @ShivangiReja @maririos @YijunXieMS @jeremymeng @sima-zhu 

Here is an example of an addition I made that scrubs out SAS tokens in recordings.

PR #17030, opened 2021-03-02
Scrub batch shared keys

tasherif-msft:
Resolves https://github.com/Azure/azure-sdk-for-python/issues/17054. Currently shared keys are leaking in batch operations since we have no way of replacing these values. This PR adds a new method to detect a batch operation and filter out the storage account name and shared keys from its payload using a passed regex expression.

### DIFF
--- a/tools/azure-devtools/src/azure_devtools/scenario_tests/recording_processors.py
+++ b/tools/azure-devtools/src/azure_devtools/scenario_tests/recording_processors.py
@@ -4,7 +4,7 @@
 # --------------------------------------------------------------------------------------------
 import six
 
-from .utilities import is_text_payload, is_json_payload
+from .utilities import is_text_payload, is_json_payload, is_batch_payload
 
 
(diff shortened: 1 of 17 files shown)


tools/azure-devtools/src/azure_devtools/scenario_tests/recording_processors.py
```diff
@@ -195,6 +195,13 @@ def process_request(self, request):
                 if old in body:
                     request.body = body.replace(old, new)
 
+            if request.body and request.uri and is_batch_payload(request):
+                import re
+                body = six.ensure_str(request.body)
+                matched_objects = set(re.findall(old, body))
```
tasherif-msft, line 201: Use `set` to ensure that we don't iterate over the already replaced substring (no duplicates)

kasobol-msft (approved):

seankane-msft (approved):
LGTM :shipit:

PR #17032, opened 2021-03-02
[Key Vault] Make test resource cleanup script asynchronous

### LINKED ISSUE #16473: [Key Vault] Efficient test resource cleanup script
Key Vault tests will create resources in an existing vault/MHSM moving forward, and we need a way to clean up those resources between test runs. Synchronously deleting and purging resources of each type takes a long time to complete, so this should be done asynchronously when possible.

mccoyp:
Resolves #16473.

This makes test resource cleanup significantly faster. Certificates are deleted before keys and secrets so that certificate-related key and secret deletion requests don't raise service errors.

### DIFF
--- a/sdk/keyvault/test-resources-cleanup.py
+++ b/sdk/keyvault/test-resources-cleanup.py
@@ -2,13 +2,14 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
 # ------------------------------------
-import os
+import asyncio
 from dotenv import load_dotenv
+import os
 
-from azure.identity import ClientSecretCredential
-from azure.keyvault.certificates import CertificateClient
-from azure.keyvault.keys import KeyClient
-from azure.keyvault.secrets import SecretClient
+from azure.identity.aio import ClientSecretCredential
+from azure.keyvault.certificates.aio import CertificateClient
+from azure.keyvault.keys.aio import KeyClient
+from azure.keyvault.secrets.aio import SecretClient
 
 load_dotenv()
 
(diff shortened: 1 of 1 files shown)


sdk/keyvault/test-resources-cleanup.py
```diff
@@ -31,25 +32,60 @@
...
+async def close_sessions():
+    await credential.close()
+    await cert_client.close()
+    await key_client.close()
+    await secret_client.close()
+
+loop = asyncio.get_event_loop()
+loop.run_until_complete(delete_certificates())
+loop.run_until_complete(delete_keys_and_secrets())
+loop.run_until_complete(purge_resources())
```
chlowell, line 89: The script might be faster if it scheduled all these coroutines at once.

chlowell (approved):
I imagine if we use this in CI, we can choose which version of Python runs it?

sdk/keyvault/test-resources-cleanup.py
mccoyp: It might be, but Key Vault raises an error when a coroutine is scheduled to delete a key or secret that's created for a certificate. Do you know of any way around that?

chlowell: Keys and secrets have a "managed" property to indicate whether they were so created (which we don't expose for secrets! #17129 😨). You could skip deleting and purging the managed ones, assuming Key Vault deletes/purges them with their associated certificate (naively seems like it must do this but I haven't checked). I expect the decrease in total requests would have a much greater impact than scheduling all the coroutines at once. Anyway, it's not something you need to investigate, unless you're too curious 😸

Issue #16535, opened 2021-02-04
[formrecognizer] unskip prebuilt url tests


kristapratico:
We are currently skipping all prebuilt model tests that use a URL as input. We have seen an increased incidence of "InvalidimageUrl" errors returned from the service despite sending valid URLs. We see this behavior across languages / regions both with image URLs hosted from github and with blob URLs.

We also need to unskip the changes made in this PR to disable the receipt URL samples: https://github.com/Azure/azure-sdk-for-python/pull/17036

Service team is notified and looking into it.

PR #17036, opened 2021-03-02
[formrecognizer] temp disable sample tests until service bug fixed

kristapratico:
Disabling URL samples so that we may get some green live tests. This change is being tracked in the overall issue for this service bug here: https://github.com/Azure/azure-sdk-for-python/issues/16535

### DIFF
--- a/sdk/formrecognizer/azure-ai-formrecognizer/samples/sample_authentication.py
+++ b/sdk/formrecognizer/azure-ai-formrecognizer/samples/sample_authentication.py
@@ -36,7 +36,7 @@ import os
 
 class AuthenticationSample(object):
 
-    url = "https://raw.githubusercontent.com/Azure/azure-sdk-for-python/master/sdk/formrecognizer/azure-ai-formrecognizer/tests/sample_forms/receipt/contoso-receipt.png"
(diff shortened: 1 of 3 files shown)


sdk/formrecognizer/azure-ai-formrecognizer/samples/sample_authentication.py
```diff
@@ -47,7 +47,7 @@ def authentication_with_api_key_credential_form_recognizer_client(self):
 
         form_recognizer_client = FormRecognizerClient(endpoint, AzureKeyCredential(key))
         # [END create_fr_client_with_key]
-        poller = form_recognizer_client.begin_recognize_receipts_from_url(self.url)
+        poller = form_recognizer_client.begin_recognize_content_from_url(self.url)
         receipt = poller.result()
```
catalinaperalta, line 51: Should `receipt` be updated to `form`? (just for clarity) Same question for the change below.

catalinaperalta (approved):
LGTM! Just left one suggested update to a variable name for clarity.

sdk/formrecognizer/azure-ai-formrecognizer/samples/sample_authentication.py
kristapratico: good point, going to change to `result` to match async sample

Issue #16782, opened 2021-02-17
Add ConnectionString parsing


nfx:
In plenty of cases we receive connection strings for azure resources. We should port `ConnectionString` parser from iot sdk, so that it's easier and more consistent to work with APIs https://github.com/Azure/azure-iot-sdk-python/blob/master/azure-iot-device/azure/iot/device/common/auth/connection_string.py

xiangyan99:
Thanks for the feedback. Are you suggesting moving the connection_string parsing into azure.core?

One concern here is different services may have different schemas for the connection string.

e.g. app config service connection string is different from the one for iot. hence _validate_keys may not work correctly for app config.

Could you help give more data points which services can share the parser? That can help us prioritize the work.

Thank you.

nfx:
@xiangyan99 i'd mostly need connection string parser for storage and event hubs. that's it :) you can add more services later. typical use case would be getting azure event hub connection string and configuring pyspark consumer with it. or accessing adls gen1/gen2/blob through python sdk directly. 

the point is that connection strings are usually the outputs from Terraform & they land in secret stores and stored over there as secret string as a whole. it would be nicer not to write same (even small) amount of code again and again. 

you can introduce a `ConnectionString` class that has generic methods to access keys in a free form and then have subclasses with specific service schema validation. Those subclasses should also extend `Credentials` class (that usually has SPN client/secret).

xiangyan99:
@nfx Thank you for the information. It is very helpful.

nfx:
Looking forward to see it and use it :) make sure it propagates to azureeventhubs and storage docs

swathipil:
Hi @nfx,

The feature has been added and will be available in the next release. We did consider your proposed option of some sort of `ConnectionString` class, but ultimately decided that a parsing function is the best solution, as it provides validation that is generic enough for parsing connection strings across services and is simplest to use.

If you are looking for a `ConnectionString` class, specifically, Event Hubs already has the [`EventHubConnectionStringProperties`](https://github.com/Azure/azure-sdk-for-python/blob/a70c40cbce1bc091c6cf961c69f82e78db8511ce/sdk/eventhub/azure-eventhub/azure/eventhub/_connection_string_parser.py#L13) class, which I believe is similar to what you are looking for. Storage does not seem to have a similar class, but we can create an issue for this if you still find this necessary.

Thank you for your feedback!

nfx:
[...]

PR #17049, opened 2021-03-03
[DO NOT MERGE] core connection string parser

swathipil:
This is an attempt to create a generic connection string parsing function that will be used across SDKs to minimize redundant logic and validation, addressing #16782. 

This parsed connection string will be returned as a dict of connection string keys (ie. `Endpoint`, `SharedAccessKey`, etc.) and values, with an option for allowing case insensitive keys. This option is useful when checking a connection string with keys that don't match expected case, without looping through all the keys. For example:
```
conn_str = "Endpoint=XXXXENDPOINT;ACcEsSKEY=XXXXKEY"
parsed = parse_connection_string(conn_str, case_sensitive_keys=False)
endpoint = parsed.get("endpoint")   # value is XXXXENDPOINT
accesskey = parsed.get("accesskey")  # value is XXXXKEY
```

THIS PR WILL NOT BE MERGED. Once approved, the PR will be cherry-picked for the core changes which includes all connection string parsing related changes in `azure-core`. 

The rest of service the changes ARE ONLY FOR REFERENCE, for use for each service developer if they choose to use this new feature.

gist: https://gist.github.com/swathipil/0640998161b2b9fb41dee053e16a770c


sdk/core/azure-core/azure/core/_connection_string_parser.py
```diff
@@ -0,0 +1,31 @@
...
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+# --------------------------------------------------------------------------
+from typing import Dict
+
+CS_DELIMITER = ";"
+CS_VAL_SEPARATOR = "="
+
+def parse_connection_string_to_dict(conn_str):
```
rakshith91, line 11: is this method exposed to users? I'm not quite sure having a generic parser in azure core provides much value -it'd prolly be easier for users to write a `.split(;)` for themselves than find this API.

swathipil: So I exposed it to users since it needs to be available across all the SDKs that use parsing. I wasn't sure if there was another way to make it available to all those SDKs without making it public.

So the original request for this parser came from #16782. In order to actually make it generic enough so that it can be used by all of the other parsers, we couldn't do specific key validation here. There are only a couple validations that are generic enough that they can be added here without really breaking any of the other parsers. For some SDK parsers, this does provide extra checks that weren't in place before, so I think there is some value there.

xiangyan99:
You need to pump the core version (and also in those packages' setup.py) to enable the new feature for those packages. :)

xiangyan99:
LGTM for the azure-core changes.

sdk/core/azure-core/tests/test_connection_string_parsing.py
```diff
@@ -0,0 +1,182 @@
...
+        assert parse_result.get("accesskey") is None
+        assert parse_result.get("accesskey", "XXothertestkeyXX=") == "XXothertestkeyXX="
+
+    def test_case_insensitive_keys_method(self):
+        conn_str = 'enDpoiNT=XXXXENDPOINTXXXX;sharedaccesskeyname=XXXXPOLICYXXXX;SHAREDACCESSKEY=THISISATESTKEYXXXXXXXXXXXXXXXXXXXXXXXXXXXX='
+        parse_result = parse_connection_string(conn_str, False)
+        keys = parse_result.keys()
+        assert len(keys) == 3
+        assert "enDpoiNT" in keys
+        if sys.version_info >= (3, 5):
```
swathipil, line 109: The behavior of the `.keys()` method differs in Python 2.7 vs. 3.5+ for CaseInsensitiveDict. Is this concerning behavior?

PR #17053, opened 2021-03-03
[EG] Regenerate Code

rakshith91:


sdk/eventgrid/azure-eventgrid/azure/eventgrid/_generated/aio/operations/_event_grid_publisher_client_operations.py
```diff
@@ -22,7 +22,7 @@ class EventGridPublisherClientOperationsMixin:
     async def publish_events(
         self,
         topic_hostname: str,
-        events: List["models.EventGridEvent"],
+        events: List["_models.EventGridEvent"],
```
yunhaoling, line 25: random question: how would this be rendered on doc/type hint?

wondering if users would get confused that how they create `EventGridEvent` from `_models` module.
(ignore me if this could be rendered corrected by python/sphinx)

sdk/eventgrid/azure-eventgrid/azure/eventgrid/_event_mappings.py
```diff
@@ -12,28 +12,28 @@ class SystemEventNames(str, Enum):
     visit https://docs.microsoft.com/azure/event-grid/system-topics.
     """
 
-    ACSChatMemberAddedToThreadWithUserEventName = (
+    AcsChatMemberAddedToThreadWithUserEventName = (
```
yunhaoling, line 15: Is is something public in the previous release?
naming change seems to be a breaking change to me, probably something worth to mention in the changelog.

If those are just added in this release, then I have no concern.

sdk/eventgrid/azure-eventgrid/azure/eventgrid/_generated/aio/operations/_event_grid_publisher_client_operations.py
rakshith91: I'm not much concerned about this - this is an autogenerated method from swagger which is not used anywhere in our code. (we just need this in the swagger for the sake of other languages)

sdk/eventgrid/azure-eventgrid/azure/eventgrid/_event_mappings.py
rakshith91: i was on fence to add this in the changelog, but i agree -updated

yunhaoling (approved):

PR #17055, opened 2021-03-03
[Perf] Small fixes to storage-blob

mikeharder:
- Removed unnecessary code in UploadTest (copied from DownloadTest)
- Renamed option in ListBlobsTest to match other languages


annatisch (approved):
LGTM - thanks! :)

tasherif-msft (approved):

Issue #12353, opened 2020-07-02
[Azure Event Hub]- Sending Event Message with Partition Key of type int is DANGEROUS


aliyevmirzakhan:
Hi, a few days ago I wrote script that sends the real time events to Event Hub. I used Azure Durable Functions to implement parallelism, and in doing so I used partition_key to ensure events with same id is directed to same partition. For some reason I thought the key can be single number and thus I used int types as partition keys, and while sending I never had any error about it. Later I implemented another function App that listens to Event Hub, and triggered when new message is received in Even Hub, and while doing so I was getting the following error at each firing:
`Unable to cast object of type 'System.Int32' to type 'System.String`

At first I thought may be this is due to encoding of str files that send which took me quite a long time to figure out this was because my usage of int value as partition key due to which Event Hub backend could not generate partition key hashes. 

I am wondering why no Exception is raised if int type is passed as partition_key argument at the time of sending message string, and may be this is bug to be considered.

kaerm:
@aliyevmirzakhan thanks for reporting this, @yunhaoling, @YijunXieMS can you take a look at this?

yunhaoling:
Hello @aliyevmirzakhan , 

May I ask first confirm which language and what version of the EventHub sdk are you using?

Because from the error information `Unable to cast object of type 'System.Int32' to type 'System.String`, it looks like you're using  EventHub sdk of .NET, while this repo mainly focuses on Python SDKs.

johanste:
I believe the issue is more likely somewhere in the azure event hubs binding for functions. I don't know the protocol well enough to definitively state if an integer is a valid partition key or not - but if my understanding is correct, the value gets transmitted all the way to the receiver, which would imply that it is.

jsquire:
Hi @aliyevmirzakhan.   Would you be so kind as to help clarify a bit more about your scenario so that we can make sure to get the right folks involved to assist?  It would be helpful to understand:

- How is the event being published to Event Hubs?  The .NET SDKs only allow a `string` to be used as a partition key when publishing events, so it would seem that the event is being published using some other means.  Is it the Python SDK being used to publish the event?

- Are the Azure Functions that you're using both Python, both .NET or a mix?   Can you clarify which stack is generating the exception?

YijunXieMS:
@johanste Yes the protocol supports string, int (and maybe other types) from our test.
We decided to make it a string for all languages back then. Python has type annotation and documentation to indicate it's a string like below.
```python
[...]

PR #17057, opened 2021-03-03
[EventHubs] add logging.info to warn the usage of partition key of non-string type

yunhaoling:
The PR is to further address the issue with stronger warning in logging: https://github.com/Azure/azure-sdk-for-python/issues/12353

### DIFF
--- a/sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py
+++ b/sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py
@@ -246,6 +246,12 @@ class EventHubProducerClient(ClientBase):
         """
         partition_id = kwargs.get("partition_id")
         partition_key = kwargs.get("partition_key")
+
+        if partition_key and not isinstance(partition_key, str):
+            _LOGGER.info("WARNING: Please DO NOT pass a partition_key of non-string type to the send_batch method."
+                         "The Event Hub service ignores partition_key of non-string type, "
+                         "in which case events will be assigned to all partitions using round-robin.")
+
         if isinstance(event_data_batch, EventDataBatch):
             if partition_id or partition_key:
                 raise TypeError("partition_id and partition_key should be None when sending an EventDataBatch "
@@ -305,6 +311,11 @@ class EventHubProducerClient(ClientBase):
         max_size_in_bytes = kwargs.get("max_size_in_bytes", None)
         partition_id = kwargs.get("partition_id", None)
         partition_key = kwargs.get("partition_key", None)
+
+        if partition_key and not isinstance(partition_key, str):
+            _LOGGER.info("WARNING: Please DO NOT pass a partition_key of non-string type to the create_batch method."
+                         "The Event Hub service ignores partition_key of non-string type, "
+                         "in which case events will be assigned to all partitions using round-robin.")
 
         if max_size_in_bytes and max_size_in_bytes > self._max_message_size_on_link:
             raise ValueError(
--- a/sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py
+++ b/sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py
@@ -4,6 +4,7 @@
 # --------------------------------------------------------------------------------------------
 import logging
 import threading
+import six
 
 from typing import Any, Union, TYPE_CHECKING, Dict, List, Optional, cast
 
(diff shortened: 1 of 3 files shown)


swathipil (approved):
:shipit:

sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py
```diff
@@ -246,6 +246,12 @@ def send_batch(self, event_data_batch, **kwargs):
         """
         partition_id = kwargs.get("partition_id")
         partition_key = kwargs.get("partition_key")
+
+        if partition_key and not isinstance(partition_key, str):
+            _LOGGER.info("WARNING: Please DO NOT pass a partition_key of non-string type to the send_batch method."
```
rakshith91, line 251: why not `_looger.warning`?

sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py
```diff
@@ -306,6 +312,11 @@ def create_batch(self, **kwargs):
         partition_id = kwargs.get("partition_id", None)
         partition_key = kwargs.get("partition_key", None)
 
+        if partition_key and not isinstance(partition_key, str):
+            _LOGGER.info("WARNING: Please DO NOT pass a partition_key of non-string type to the create_batch method."
```
rakshith91, line 316: same here

sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py
rakshith91: this applies everywhere - is there a specific reason why you restricted it to info than a warning

yunhaoling: according to our guideline warning is used to indicate error and error should be raised:
> DO use the WARNING logging level when a function fails to perform its intended task. The function should also raise an exception.

while in special case passing partition_key of non-string type doesn't cause any error, the service simply ignore, so making it a logging info.

annatisch: Yeah I think `info` is probably appropriate here - we can bump it up a level later if we get feedback, or messaging from the service team that this should be treated more like an error. Though currently the service is not treating this like an error.
Do we want to tell people _not_ to do this? Or be informative, e.g. "By settings a non-string value, this partition key will be ignored by the service"
(Just asking, not suggesting :) )

sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py
```diff
@@ -306,6 +312,11 @@ def create_batch(self, **kwargs):
         partition_id = kwargs.get("partition_id", None)
         partition_key = kwargs.get("partition_key", None)
 
+        if partition_key and not isinstance(partition_key, str):
```
annatisch, line 315: Does this need to be `six.string_types`? Or would 2.7 not support a unicode value here?

sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py
yunhaoling: based on:
- what Serkant mentioned in the issue #12353 :
> Service doesn't do any validation on the delivered type and assumes sender library already conducted sanitary checks. This applies to every entry in the annotations map as far as I know.
- the customer described having trouble with Azure Function receiving messages which uses SDKs from a different language trying to parse the value:
```Unable to cast object of type 'System.Int32' to type 'System.String```

I think users had better not to set the partition_key of non-string type.

However, the wording could be improved in a "softer" way, I'm thinking of this way to be less imperative and more descriptive.

`Setting a partition_key of non-string type is discouraged in which case the partition_key will be ignored by the Event Hub service ignores and events will be assigned to all partitions using round-robin. Furthermore, SDKs from other languages might fail to parse the non-string value if they only expect partition_key to be string type`

How does it look to you?

File: sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py (hunk 3 above)
Comment (yunhaoling) [reply sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py:315]: just tested unicode partition_key in python2.7, they are working well and can be distributed among different partitions (absolutely not in a round-robin way)

Comment (yunhaoling) [reply sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py:315]: I think this should actually be `isinstance(partition_key, (six.text_type, six.binary_type))` depending on our uamqp util.data_factory implementation which is used to encode the `partiton_key` inside the message annotation, six.text_type and six.binary_type will be interpreted into AMQPString value.

I have also tested partition_key as bytes, they're working as well.

https://github.com/Azure/azure-uamqp-python/blob/e1051ae1c8f93ed2792a852857d261a4f34b76f5/uamqp/utils.py#L117-L120

```python
def data_factory(value, encoding='UTF-8'):
    ...
    elif isinstance(value, six.text_type):
        result = c_uamqp.string_value(value.encode(encoding))
    elif isinstance(value, six.binary_type):
        result = c_uamqp.string_value(value)
    ...
```

sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py
annatisch: Looks good! Thanks!

PR #17063, opened 2021-03-03
Adapt to azure core's cloud event

rakshith91:
Fixes https://github.com/Azure/azure-sdk-for-python/issues/16585

### DIFF
--- a/sdk/eventgrid/azure-eventgrid/azure/eventgrid/_helpers.py
+++ b/sdk/eventgrid/azure-eventgrid/azure/eventgrid/_helpers.py
@@ -18,6 +18,10 @@ from azure.core.credentials import AzureKeyCredential, AzureSasCredential
 from ._signature_credential_policy import EventGridSasCredentialPolicy
 from . import _constants as constants
 
+from ._generated.models import (
+    CloudEvent as InternalCloudEvent,
+)
+
 if TYPE_CHECKING:
     from datetime import datetime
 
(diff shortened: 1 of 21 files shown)


sdk/eventgrid/azure-eventgrid/azure/eventgrid/_helpers.py
```diff
@@ -134,3 +138,25 @@ def _eventgrid_data_typecheck(event):
             "Data in EventGridEvent cannot be bytes. Please refer to"
             "https://docs.microsoft.com/en-us/azure/event-grid/event-schema"
         )
+
+def _cloud_event_to_generated(cloud_event, **kwargs):
```
rakshith91, line 142: I'm on the fence to put it in the _helpers file  vs leaving this as a standalone method in _models.py - any strong opinions here? 

Nevertheless, it's an implementation detail and not blocking

sdk/eventgrid/azure-eventgrid/tests/test_eg_publisher_client.py
```diff
@@ -132,32 +135,27 @@ def test_send_cloud_event_data_dict(self, resource_group, eventgrid_topic, event
                 )
         client.send(cloud_event)
 
+    @pytest.mark.skip("https://github.com/Azure/azure-sdk-for-python/issues/16993")
     @CachedResourceGroupPreparer(name_prefix='eventgridtest')
     @CachedEventGridTopicPreparer(name_prefix='cloudeventgridtest')
-    def test_send_cloud_event_data_base64_using_data(self, resource_group, eventgrid_topic, eventgrid_topic_primary_key, eventgrid_topic_endpoint):
+    def test_send_cloud_event_data_NULL(self, resource_group, eventgrid_topic, eventgrid_topic_primary_key, eventgrid_topic_endpoint):
```
rakshith91, line 141: this test technically "passes", but with a NULL object sent in the request - skipping it for now

sdk/eventgrid/azure-eventgrid/azure/eventgrid/_helpers.py
lmazuel: don't care  :p

sdk/eventgrid/azure-eventgrid/tests/test_eg_publisher_client.py
lmazuel: You should be able to use "raw_request_hook" to inspect the body and confirm null was sent

rakshith91: `AssertionError: assert '<azure.core.serialization._Null object at 0x048B9670>' is None`

lmazuel (approved):

yunhaoling:
I think we also need to update samples/docs referring to `azure.eventgrid.CloudEvent`, e.g. sample code in readme: 
image

(also it's worthwhile to mention in readme something like `please install azure-core to use CloudEvent` and ref links etc

but this could be put into a separate PR.

yunhaoling:
minor comment:

we could add some references/links in the readme sending cloud event sample saying something like `to send a CloudEvent, you need to install azure-core first`

yunhaoling (approved):

rakshith91:
>>> to send a CloudEvent, you need to install azure-core first

azure-core is a dependency - so no need to mention it i guess :)

PR #17076, opened 2021-03-04
Add NULL to readme

rakshith91:


sdk/core/azure-core/README.md
```diff
@@ -179,6 +179,17 @@ class MyCustomEnum(with_metaclass(CaseInsensitiveEnumMeta, str, Enum)):
...
 
+#### Null Sentinel Value
+
+A falsy sentinel object which is supposed to be used to specify attributes
+with no data. This gets serialized to `null` on the wire.
+
+```python
+from azure.core.serialization import NULL
+
+assert bool(NULL) is False
```
rakshith91, line 190:
```suggestion
from azure.core.serialization import NULL

assert bool(NULL) is False

model = Model(
    attr=NULL
)
```

sdk/core/azure-core/README.md
```diff
@@ -179,6 +179,21 @@ class MyCustomEnum(with_metaclass(CaseInsensitiveEnumMeta, str, Enum)):
     BAR = 'bar'
 ```
 
+#### Null Sentinel Value
```
xiangyan99, line 182: Just want to double check we expect this to be used by end user or developer or both?

rakshith91: we expect this to be used by end user - if they want to pass an explicit null on the wire

rakshith91: it's a rare edge case (which cloud events happen to allow), so decided to add a sentinel

xiangyan99 (approved):

PR #17080, opened 2021-03-04
[ServiceBus] Object mapping support

annatisch:


sdk/servicebus/azure-servicebus/azure/servicebus/_common/utils.py
```diff
@@ -222,20 +214,37 @@ def transform_messages_to_sendable_if_needed(messages):
...
+            )
+        )
+
+
 def create_messages_from_dicts_if_needed(messages, message_type):
-    # type: (DictMessageType, type) -> DictMessageReturnType
+    # type: (MessagesType, Type[ServiceBusMessage]) -> Union[ServiceBusMessage, List[ServiceBusMessage]]
     """
     This method is used to convert dict representations
     of messages to a list of ServiceBusMessage objects or ServiceBusBatchMessage.
```
swathipil, line 237: since `ServiceBusMessageBatch` is not of `MessagesType`, should it be removed from this description?
```suggestion
    of messages to a list of ServiceBusMessage objects.
```

sdk/servicebus/azure-servicebus/azure/servicebus/_common/utils.py
```diff
@@ -222,20 +214,37 @@ def transform_messages_to_sendable_if_needed(messages):
...
-    :param DictMessageType messages: A list or single instance of messages of type ServiceBusMessages or
+
+    :param Messages messages: A list or single instance of messages of type ServiceBusMessages or
         dict representations of type ServiceBusMessage. Also accepts ServiceBusBatchMessage.
```
swathipil, line 240: Same here, I think?
```suggestion
        dict representations of type ServiceBusMessage.
```

swathipil (commented):
left a couple questions!

sdk/servicebus/azure-servicebus/azure/servicebus/_common/utils.py
annatisch: Thanks - fixed!

sdk/servicebus/azure-servicebus/azure/servicebus/_common/utils.py
annatisch: Fixed :)

swathipil (approved):
LGTM 👍

sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_sender.py
```diff
@@ -368,48 +378,44 @@ def send_messages(self, message, **kwargs):
...
             if (
-                isinstance(message, ServiceBusMessageBatch) and len(message) == 0
+                isinstance(obj_message, ServiceBusMessageBatch) and len(obj_message) == 0
             ):  # pylint: disable=len-as-condition
                 return  # Short circuit noop if an empty list or batch is provided.
-            if not isinstance(message, (ServiceBusMessageBatch, ServiceBusMessage)):
-                raise TypeError(
-                    "Can only send azure.servicebus.<ServiceBusMessageBatch, ServiceBusMessage> "
-                    "or lists of ServiceBusMessage."
-                )
```
yunhaoling, line 405: would this removed part of instance type checking be covered by the `_single_message_from_dict` method?

sdk/servicebus/azure-servicebus/tests/async_tests/mgmt_tests/test_mgmt_queues_async.py
```diff
@@ -329,11 +329,11 @@ async def test_async_mgmt_queue_update_invalid(self, servicebus_namespace_connec
         queue_description = await mgmt_service.create_queue(queue_name)
         try:
             # handle a null update properly.
-            with pytest.raises(AttributeError):
+            with pytest.raises(TypeError):
                 await mgmt_service.update_queue(None)
 
             # handle an invalid type update properly.
-            with pytest.raises(AttributeError):
+            with pytest.raises(TypeError):
```
yunhaoling, line 336: is it a breaking change -- raising TypeError instead of AttributeError?

sdk/servicebus/azure-servicebus/azure/servicebus/management/_utils.py
```diff
@@ -326,14 +321,18 @@ def _validate_topic_subscription_and_rule_types(
...
-    :rtype: DictPropertiesReturnType
+    :rtype: PropertiesType
     """
-    return_properties = sb_resource_type(**properties) if isinstance(properties, dict) else properties
-    return return_properties
+    if isinstance(properties, sb_resource_type):
+        return properties
+    try:
+        return sb_resource_type(**cast(Mapping[str, Any], properties))
+    except TypeError:
```
yunhaoling, line 337: dumb question: where is this TypeError getting raised from, the `cast`?

sdk/servicebus/azure-servicebus/azure/servicebus/management/_utils.py
```diff
@@ -326,14 +321,18 @@ def _validate_topic_subscription_and_rule_types(
...
+        raise TypeError("Update input must be an instance of {}, or a mapping".format(sb_resource_type.__name__))
```
yunhaoling, line 338: in the constructor of `*Properties`, a method called `validate_extraction_missing_args` will be called to check if there's any required keyword argument missing, if there is, then it will also raise TypeError, wouldn't that message be swallowed and replaced by this `TypeError("Update input must be an instance of {}, or a mapping".format(sb_resource_type.__name__))`?

the code 
```
def validate_extraction_missing_args(extraction_missing_args):
    if extraction_missing_args:
        raise TypeError(
            "__init__() missing {} required keyword arguments: {}".format(
                len(extraction_missing_args),
                " and ".join(["'" + e + "'" for e in extraction_missing_args]),
            )
        )
```

sdk/servicebus/azure-servicebus/azure/servicebus/management/_utils.py
annatisch: No, attempting to splat an object (`properties`) that is not a Mapping results in a TypeError :)
Do you think it would aid readability to move the `cast` to a separate line?

sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_sender.py
annatisch: Yes that was my intention - it seems we perform the same type checks a number of times, so I tried to slim it down a bit.

sdk/servicebus/azure-servicebus/tests/async_tests/mgmt_tests/test_mgmt_queues_async.py
annatisch: As discussed in the chat - we can probably call this a bugfix

sdk/servicebus/azure-servicebus/azure/servicebus/management/_utils.py
annatisch: Yeah this is a really good point - I will sleep on it...

sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_sender.py
yunhaoling: yea, I re-read the code, should be covered :)

sdk/servicebus/azure-servicebus/azure/servicebus/management/_utils.py
yunhaoling: I'm good with it :) after trying the code I understand how it works, thanks for the explanation!

sdk/servicebus/azure-servicebus/tests/async_tests/mgmt_tests/test_mgmt_queues_async.py
yunhaoling: so I guess the change is worth a line in the changelog?
something like:
```
Fixed a bug that `update_*` methods on `ServiceBusManagementClient` should raise `TypeError` when given value of invalid types instead of `AttributeError`.
```

sdk/servicebus/azure-servicebus/azure/servicebus/management/_utils.py
yunhaoling: I think what we could do is:

1. have a separate instance check at the front:
```python
if not isinstance(properties, Mapping):
    raise TypeError("Update input must be an instance of {}, or a mapping".format(sb_resource_type.__name__))
return sb_resource_type(**cast(Mapping[str, Any], properties))
```
2. in the except block check the message of the exception:
```python
except TypeError as e:
    if "missing" in str(e):
        raise e
    raise TypeError("Update input must be an instance of {}, or a mapping".format(sb_resource_type.__name__))
```

annatisch: Unfortunately there's no such thing as a type check for a protocol.... as this kind of negates the reason to accept a protocol :)
I suspect option 2 would be the only way to surface the original error message.

sdk/servicebus/azure-servicebus/tests/async_tests/mgmt_tests/test_mgmt_queues_async.py
annatisch: Added! :)

sdk/servicebus/azure-servicebus/azure/servicebus/_common/utils.py
```diff
@@ -222,20 +214,39 @@ def transform_messages_to_sendable_if_needed(messages):
...
+
+def _single_message_from_dict(message, message_type):
+    # type: (Union[ServiceBusMessage, Mapping[str, Any]], Type[ServiceBusMessage]) -> ServiceBusMessage
+    if isinstance(message, message_type):
+        return message
+    try:
+        return message_type(**cast(Mapping[str, Any], message))
+    except TypeError as e:
+        if "required keyword arguments" in str(e):
+            raise e
```
yunhaoling, line 226: oops, I think this is added in the wrong place :p

should be in the `sdk/servicebus/azure-servicebus/azure/servicebus/management/_utils.py`

annatisch: Hahahahaha thank you thank you :D

yunhaoling (approved):
LGTM

sdk/servicebus/azure-servicebus/tests/async_tests/test_queues_async.py
```diff
@@ -1785,6 +1786,53 @@ async def test_queue_async_send_dict_messages(self, servicebus_namespace_connect
                         received_messages.append(message)
                 assert len(received_messages) == 6
 
+    @pytest.mark.liveTest
+    @pytest.mark.live_test_only
+    @CachedResourceGroupPreparer(name_prefix='servicebustest')
+    @CachedServiceBusNamespacePreparer(name_prefix='servicebustest')
+    @ServiceBusQueuePreparer(name_prefix='servicebustest')
+    async def test_queue_async_send_mapping_messages(self, servicebus_namespace_connection_string, servicebus_queue, **kwargs):
```
swathipil, line 1794: out of curiosity, is there only a test here because the changes made for Mapping were the same everywhere? so if this tests passes/fails, it applies to the mgmt `update_*` as well?

annatisch: Yeah - my thinking was that the approach I took was the same for both messages as properties:
```
try:
    obj_type(**random_value)
except TypeError:
    ...
```
We already had coverage for objects and dictionaries across all scenarios. I mostly wanted to affirm that this pattern could handle a non-dict Mapping.

swathipil: ah, that makes sense, thank you!

PR #17101, opened 2021-03-05
search perf tests

xiangyan99:


sdk/search/azure-search-documents/tests/perfstress_tests/search_documents.py
```diff
@@ -0,0 +1,36 @@
...
+        super().__init__(arguments)
+        api_key = self.get_from_env("AZURE_SEARCH_API_KEY")
+        service_endpoint = os.getenv("AZURE_SEARCH_SERVICE_ENDPOINT")
+        index_name = os.getenv("AZURE_SEARCH_INDEX_NAME")
+        key = os.getenv("AZURE_SEARCH_API_KEY")
+        self.service_client = SyncAppConfigClient(service_endpoint, index_name, AzureKeyCredential(api_key))
+        self.async_service_client = AsyncAppConfigClient(service_endpoint, index_name, AzureKeyCredential(api_key))
+
+    async def global_setup(self):
+        await super().global_setup()
```
annatisch, line 26: Do we need to pre-load any documents into the index in order for the search to actually return results?
Otherwise the perf test might be skipping some results deserialization logic?
And if the returned results are pagable, then iterating through them would also be an important part. Maybe add a configuable `--num-documents` flag to set up the number of results we would expect to be returned by the search?

sdk/search/azure-search-documents/tests/perfstress_tests/suggest.py
```diff
@@ -0,0 +1,32 @@
...
+        key = os.getenv("AZURE_SEARCH_API_KEY")
+        self.service_client = SyncAppConfigClient(service_endpoint, index_name, AzureKeyCredential(api_key))
+        self.async_service_client = AsyncAppConfigClient(service_endpoint, index_name, AzureKeyCredential(api_key))
+
+    async def close(self):
+        await self.async_service_client.close()
+        await super().close()
+
+    def run_sync(self):
+        self.service_client.suggest(search_text="mot", suggester_name="sg")
```
annatisch, line 29: Same comment as for search - do we need to set anything up in order for this to actually return results?
Otherwise we might be not be testing deserialization logic.
(Same question for the auto-complete test)

sdk/search/azure-search-documents/tests/perfstress_tests/search_documents.py
xiangyan99: Yes. This is covered in the pre-configuration. :)

Azure offers some shared data for customers to try the search service.

https://docs.microsoft.com/en-us/azure/search/search-indexer-tutorial

Importing data and creating index are costy and usually user only needs it once so we don't want to run perf tests against them.

xiangyan99: Yes. This is covered in the pre-configuration. :)

Azure offers some shared data for customers to try the search service.

https://docs.microsoft.com/en-us/azure/search/search-indexer-tutorial

Importing data and creating index are costy and usually user only needs it once so we don't want to run perf tests against them.

annatisch: When you say it's covered in the pre-configuration - there's nothing in the readme about it that I can see?

sdk/search/azure-search-documents/tests/perfstress_tests/search_documents.py
```diff
@@ -0,0 +1,53 @@
...
+
+    async def close(self):
+        await self.async_service_client.close()
+        await super().close()
+
+    def run_sync(self):
+        if self.args.num_documents == -1:
+            results = len(self.service_client.search(search_text="luxury"))
+        else:
+            results = len(self.service_client.search(search_text="luxury", top=self.args.num_documents))
```
annatisch, line 43: Is this supposed to be `list()`?

sdk/search/azure-search-documents/tests/perfstress_tests/autocomplete.py
```diff
@@ -0,0 +1,32 @@
...
+
+    async def close(self):
+        await self.async_service_client.close()
+        await super().close()
+
+    def run_sync(self):
+        self.service_client.autocomplete(search_text="mot", suggester_name="sg")
+
+    async def run_async(self):
+        await self.async_service_client.autocomplete(search_text="mot", suggester_name="sg")
```
annatisch, line 32: Do we need to iterate the results for this one?

sdk/search/azure-search-documents/tests/perfstress_tests/suggest.py
```diff
@@ -0,0 +1,32 @@
...
+
+    async def run_async(self):
+        await self.async_service_client.suggest(search_text="mot", suggester_name="sg")
```
annatisch, line 32: Same question here - do we need to iterate over the results? Or does this API not return a pageable?

annatisch (approved):

Issue #16683, opened 2021-02-10
[formrecognizer] add support for identity documents


kristapratico:
This a new prebuilt model that will be added to FormRecognizerClient

- [x] sync impl - file stream/from_url
- [x] async impl - file stream/from_url
- [x] samples
- [x] tests
- [x] update aka link in docstrings with link pointing to available fields

catalinaperalta:
Need to go through and fix TODOs in urls and FIXMEs in tests.

catalinaperalta:
Should add a passport test when the docs are up.

PR #17128, opened 2021-03-05
[formrecognizer] Add Identity Documents APIs

catalinaperalta:
Initial work for #16683. 

Some commented test fields will be uncommented when the service docs are up and we can see the format we should be expecting. 

Samples coming in a separate PR.


sdk/formrecognizer/azure-ai-formrecognizer/azure/ai/formrecognizer/_form_recognizer_client.py
```diff
@@ -336,6 +336,123 @@ def begin_recognize_business_cards_from_url(self, business_card_url, **kwargs):
                 )
             raise e
 
+    @distributed_trace
+    def begin_recognize_id_documents(self, id_document, **kwargs):
+        # type: (Union[bytes, IO[bytes]], Any) -> LROPoller[List[RecognizedForm]]
+        """Extract field text and semantic values from a given ID document.
+        The input document must be of one of the supported content types - 'application/pdf',
+        'image/jpeg', 'image/png', 'image/tiff' or 'image/bmp'.
+        Alternatively, use 'application/json' type to specify the location (Uri) of the
```
kristapratico, line 345: You can remove the sentence starting with "Alternatively..." since this is the file stream method. edit: same with async method

kristapratico (approved):
LGTM! I would just add to Changelog :)

Issue #17093, opened 2021-03-04
Azure Data Tables: Batch processing throwing ResourceNotFound on aio libraries


verikono:
- **azure.data.tables.aio (azure-data-tables)**: 
- **12.0.0b4**: 
- **Linux/Ubuntu**:
- **3.8**: 

**Describe the bug**
Batch processing fails throwing a ResourceNotFoundError

**To Reproduce**
1. run the attached unittest. Python version seems to be unimportant but i have not tried in 3.9.

**Expected behavior**
3 rows should appear in the target datatable

[test_ms.zip](https://github.com/Azure/azure-sdk-for-python/files/6085983/test_ms.zip)

annatisch:
Thanks for the report @verikono! We'll investigate ASAP :)
@seankane-msft - could you please take a look?

seankane-msft:
Thanks for reporting this @verikono ! Currently, the batch operations cannot be done with SAS credentials. I am working on a fix for this, in the meantime if you need a work around you can use authentication with a key and batching operations should work fine. I will update this issue with more information.

seankane-msft:
Good news update, I was able to get a solution to your problem with #17133 . In short you'll need to wrap the token with `AzureSasCredential` class and that should solve your problem. This release will go out Tuesday morning and be available for download from pypi. Let me know if you have any further questions!

verikono:
Thanks Sean!

My manager is literally bewildered that we got a response and solution so fast - amazing! :)

seankane-msft:
@verikono The latest preview release of the `azure-data-tables` library was released to [pypi](https://pypi.org/project/azure-data-tables/) yesterday. You can install with `pip install --pre azure-data-tables` and you will have support for the `AzureSasCredential`. If you have any further questions please open an issue and feel free to tag me in it.

verikono:
Thanks @seankane-msft !

PR #17133, opened 2021-03-05
Sas batching

seankane-msft:
Fixes customer reported #17093 

I added two more tests, fixed sas tests to wrap with `AzureSasCredential`


kristapratico (approved):
🚀

swathipil (commented):
a little late but lgtm

PR #17136, opened 2021-03-05
Add any additional claims to AuthenticationRequiredError

chlowell:
This preserves additional claims passed to an interactive credential's `get_token()` on AuthenticationRequiredError, via a `claims` property, enabling an application that has disabled automatic authentication to request those claims in a subsequent `get_token()` call.


schaabs (approved):

mccoyp (approved):
Love how descriptive the docstring for AuthenticationRequiredError is!

Issue #16705, opened 2021-02-11
[EventHubs] Add push mode samples to timed samples section


swathipil:
To be completed after issue: #16617 

# Summary
Many EH samples run forever because of push mode. Currently, we choose to ignore these samples when testing in the live test pipeline. This is done by listing the samples in the `IGNORED_SAMPLES` of `scripts/devops_tasks/test_run_samples.py`. We want to allow these samples to exit() after a time limit.

# Scope of work
Add the following push-mode samples to the `TIMED_SAMPLES` section in `scripts/devops_tasks/test_run_samples.py`. Since this issue should be completed after #16617 , the `TIMED_SAMPLES` section should already be in place.

# Success Criteria
* Add EH push-mode samples to the `TIMED_SAMPLES` section.
* Remove these samples from the `IGNORED_SAMPLES` section

swathipil:
all push-mode samples:
```
SYNC
        "authenticate_with_sas_token.py",
        "connection_to_custom_endpoint_address.py",
        "proxy.py",
        "receive_batch_with_checkpoint.py",
        "recv.py",
        "recv_track_last_enqueued_event_prop.py",
        "recv_with_checkpoint_by_event_count.py",
        "recv_with_checkpoint_by_time_interval.py",
        "recv_with_checkpoint_store.py",
        "recv_with_custom_starting_position.py",
        "sample_code_eventhub.py",
        # checkpointstoreblob samples
        "receive_events_using_checkpoint_store.py",
        "receive_events_using_checkpoint_store_storage_api_version.py"

ASYNC
        "authenticate_with_sas_token_async.py",
        "connection_to_custom_endpoint_address_async.py",
        "iot_hub_connection_string_receive_async.py",
        "proxy_async.py",
        "receive_batch_with_checkpoint_async.py",
        "recv_async.py",
        "recv_track_last_enqueued_event_prop_async.py",
        "recv_with_checkpoint_by_event_count_async.py",
        "recv_with_checkpoint_by_time_interval_async.py",
        "recv_with_checkpoint_store_async.py",
        "recv_with_custom_starting_position_async.py",
        "sample_code_eventhub_async.py".
        # checkpointstoreblobaio samples
        "receive_events_using_checkpoint_store_async.py",
        "receive_events_using_checkpoint_store_storage_api_version_async.py"
```

swathipil:
as per @yunhaoling 's comments in #17139, update : "auto_lock_renew(_async), those two don't receive forever"

Issue #16642, opened 2021-02-09
[ServiceBus] Add push mode samples to timed samples section


swathipil:
To be completed after issue: #16617 

# Summary
A few of the SB samples run forever because of push mode. Currently, we choose to ignore these samples when testing in the live test pipeline. This is done by listing the samples in the `IGNORED_SAMPLES` of `scripts/devops_tasks/test_run_samples.py`. In order to test as many samples as possible, we should allow push-mode samples to run for a limited period of time then timeout.

# Scope of work
Add the following push-mode samples to the `TIMED_SAMPLES` section in `scripts/devops_tasks/test_run_samples.py`. Since this issue should be completed after #16617 , the `TIMED_SAMPLES` section should already be in place.

# Success Criteria
* Add SB push-mode samples to the TIMED_SAMPLES section.
* Remove those samples from IGNORED_SAMPLES section.

swathipil:
long-running samples to add to timed section:
```
SYNC
        "failure_and_recovery.py",
        "proxy.py",
        "receive_iterator_queue.py",
        "session_pool_receive.py",
ASYNC
        "proxy_async.py",
        "receive_iterator_queue_async.py",
        "session_pool_receive_async.py"
```

swathipil:
as per @yunhaoling 's comments in #17139, update : "auto_lock_renew(_async), those two don't receive forever"

PR #17139, opened 2021-03-05
add optional sample timeout to tox sample runner

kristapratico:


scripts/devops_tasks/test_run_samples.py
```diff
@@ -87,45 +109,112 @@
...
     samples_dir_path = os.path.abspath(os.path.join(targeted_package, "samples"))
     package_name = os.path.basename(targeted_package)
+    samples_need_timeout = TIMEOUT_SAMPLES.get(package_name, {})
+
+    # install extra dependencies for samples if needed
+    try:
+        with open(samples_dir_path + "/sample_dev_requirements.txt") as sample_dev_reqs:
+            for dep in sample_dev_reqs.readlines():
+                check_call([sys.executable, '-m', 'pip', 'install', dep])
+    except:
```
kristapratico, line 196: I think we can probably except FileNotFoundError here

swathipil: will do!

swathipil (approved):

rakshith91:
Do these samples run forever by design? Should we change the samples instead since we don't want users to run  for instance, an infinite loop.

yunhaoling:
@rakshith91 , the receiving forever sample is by design.

as far as I know customers usually use eventhub/servicebus as part of a micro server application which constantly pumps and handle messages from the service, and that's why we have the push receiving design in eventhub and the service bus iterator mode (also push-based receiving).

so I prefer to keeping most of the receiving sample as they are.

but you have brought a good point and I looked into samples which don't necessarily has to be receiving forever and could be moved into normal section:
- azure-eventhub
  - authenticate_with_sas_token(_async), we could change to use send as sample
- azure-servicebus
  - auto_lock_renew(_async), those two don't receive forever

@swathipil , let's update the tickets #16705 #16642 to address those two.

scripts/devops_tasks/test_run_samples.py
```diff
@@ -87,45 +109,112 @@
 }
 
 
+def run_check_call_with_timeout(
+    command_array,
+    working_directory,
+    acceptable_return_codes=[],
+    always_exit=False,
+    timeout=None,
+    pass_if_timeout=False
```
yunhaoling, line 118: depending on the fact that all the items in the `TIMEOUT_SAMPLES` are being set to `True` for `pass_if_timeout`,
should we consider changing the default value to `True`?
(if `TIMEOUT_SAMPLES` are for samples designed to running forever, then default value to True looks more intuitive)

rakshith91 (approved):
@yunhaoling  thanks for the explanation.

LGTM

scripts/devops_tasks/test_run_samples.py
```diff
@@ -87,45 +107,117 @@
...
     for path, subdirs, files in os.walk(samples_dir_path):
         for name in files:
-            if fnmatch(name, "*.py") and name not in IGNORED_SAMPLES.get(package_name, []):
+            if fnmatch(name, "*.py") and name in samples_need_timeout:
+                timeout = samples_need_timeout[name]
+                # timeout, pass_if_timeout is True by default if nothing passed in
+                if isinstance(timeout, tuple):
+                    timeout, pass_if_timeout = timeout
+                else:
+                    pass_if_timeout = True
```
swathipil, line 205: @kristapratico , added the default here in order to not change the flow of the current code too much. Made both `timeout` and `pass_if_timeout` required params in `run_check_call_with_timeout` since it seems that there would be no reason both of those wouldn't be passed in. Does that seem reasonable?

kristapratico: Sounds good to me!

kristapratico: Ah just realized I can't approve my own PR. I'm ready to merge when you are.

swathipil: sounds good!

Issue #17040, opened 2021-03-03
[translation] add test preparer and ARM template


kristapratico:
This template can be used as a start: https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/formrecognizer/test-resources.json

Example preparer: https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/formrecognizer/azure-ai-formrecognizer/tests/preparers.py#L18

Documentation describing the test setup: https://github.com/Azure/azure-sdk-for-python/blob/master/doc/dev/tests.md#write-your-tests

We will need to figure out the details of how to set up testing for live and recorded tests. One idea is to use an existing storage account with a container containing the source documents and then using the ARM template to create / tear down the target storage account / containers (in order to avoid file name conflicts when translating).

PR #17146, opened 2021-03-06
[translation] add ARM template and set up test preparers

kristapratico:
Resolves https://github.com/Azure/azure-sdk-for-python/issues/17040

Mostly just setting up the initial test infra and ARM template here. I expect we'll need to tweak it a little bit as we start writing tests and see how things work with storage.

Actual tests will be added in a later PR.


sdk/documenttranslation/azure-ai-documenttranslation/dev_requirements.txt
```diff
@@ -2,4 +2,5 @@
 -e ../../../tools/azure-devtools
 ../../core/azure-core
 -e ../../identity/azure-identity
+../../storage/azure-storage-blob
```
kristapratico, line 5: we'll re-evaluate after beta.1 if this is the right approach for testing

sdk/documenttranslation/azure-ai-documenttranslation/tests/testcase.py
```diff
@@ -19,4 +50,70 @@ class DocumentTranslationTest(AzureTestCase):
...
+        self.storage_key = os.getenv("DOCUMENTTRANSLATION_STORAGE_KEY")
+        self.scrubber.register_name_pair(
+            self.storage_endpoint, "https://redacted.blob.core.windows.net/"
+        )
+        self.scrubber.register_name_pair(
+            self.storage_name, "redacted"
+        )
+        self.scrubber.register_name_pair(
+            self.storage_key, "fakeZmFrZV9hY29jdW50X2tleQ=="
+        )
```
seankane-msft, line 65: If these tests use preparers, these are unnecessary. The preparer does this for you [here](https://github.com/Azure/azure-sdk-for-python/blob/master/tools/azure-sdk-tools/devtools_testutils/powershell_preparer.py#L74-L76)

sdk/documenttranslation/azure-ai-documenttranslation/tests/testcase.py
```diff
@@ -19,4 +50,70 @@ class DocumentTranslationTest(AzureTestCase):
...
+            container_name,
+            account_key=self.storage_key,
+            permission=permission,
+            expiry=datetime.datetime.utcnow() + datetime.timedelta(hours=2)
+        )
+        container_sas_url = self.storage_endpoint + container_name + "?" + sas_token
+        self.scrubber.register_name_pair(
+            sas_token, "container_sas_url"
+        )
+        return container_sas_url
```
seankane-msft, line 115: SAS urls are difficult to scrub especially when you have expiry added. Try using [this](https://github.com/Azure/azure-sdk-for-python/blob/master/tools/azure-sdk-tools/devtools_testutils/azure_testcase.py#L357-L370) method in `AzureTestCase` here.

sdk/documenttranslation/azure-ai-documenttranslation/tests/preparer.py
```diff
@@ -0,0 +1,49 @@
...
+class DocumentTranslationClientPreparer(AzureMgmtPreparer):
+    def __init__(self, client_cls, client_kwargs={}, **kwargs):
+        super(DocumentTranslationClientPreparer, self).__init__(
+            name_prefix='',
+            random_name_length=42
+        )
+        self.client_kwargs = client_kwargs
+        self.client_cls = client_cls
+
+    def create_resource(self, name, **kwargs):
```
seankane-msft, line 35: I did something similar for [appconfig here](https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/appconfiguration/azure-appconfiguration/tests/wrapper.py). This can replace the `if self.is_live` portion by using the values the preparer gives you.

seankane-msft (approved):

sdk/documenttranslation/azure-ai-documenttranslation/tests/testcase.py
kristapratico: This was paranoia on my part because the storage endpoint was not getting scrubbed in the recording. Now that I'm building the storage endpoint in the code (since ARM template returned it with weird concatenation), I think I do need to have it scrubbed here?

sdk/documenttranslation/azure-ai-documenttranslation/tests/testcase.py
kristapratico: thanks, will update

sdk/documenttranslation/azure-ai-documenttranslation/tests/testcase.py
seankane-msft: Yes you are correct, I read this too quickly. This approach works, you can add this to your preparer if you want.

sdk/documenttranslation/azure-ai-documenttranslation/tests/preparer.py
kristapratico: smart!

sdk/documenttranslation/azure-ai-documenttranslation/tests/preparer.py
```diff
@@ -0,0 +1,46 @@
...
+from devtools_testutils import PowerShellPreparer, AzureMgmtPreparer
+from azure.core.credentials import AzureKeyCredential
+
+
+DocumentTranslationPreparer = functools.partial(
+    PowerShellPreparer,
+    'documenttranslation',
+    documenttranslation_test_endpoint="https://redacted.cognitiveservices.azure.com/",
+    documenttranslation_test_api_key="fakeZmFrZV9hY29jdW50X2tleQ==",
+    documenttranslation_name="redacted",
```
maririos, line 20: is this the name of the document?

kristapratico: no, that's here: https://github.com/Azure/azure-sdk-for-python/blob/7770b87a13d71f73db33281d01257235f1abab66/sdk/documenttranslation/azure-ai-documenttranslation/tests/testcase.py#L91

maririos: I see, so what is this name?

kristapratico: It ends up being something like "ccda6168-42dd-4239-80f2-0aecf398e94f.txt" if you don't add a prefix

maririos (approved):

PR #17156, opened 2021-03-08
[EventHub] EventHub T2 perf test

yunhaoling:
The PR is to implement a minimal version of the perf test for the track2 EH library.

This PR **DOES NOT** include: Track1 perf test scripts

T1 perf test will be added in another PR


sdk/eventhub/azure-eventhub/tests/perf_test/README.md
```diff
@@ -0,0 +1,58 @@
...
+(env) ~/azure-eventhub/tests> perfstress
+```
+Using the `perfstress` command alone will list the available perf tests found.
+
+### Common perf command line options
+These options are available for all perf tests:
+- `--duration=10` Number of seconds to run as many operations (the "run" function) as possible. Default is 10.
+- `--iterations=1` Number of test iterations to run. Default is 1.
+- `--parallel=1` Number of tests to run in parallel. Default is 1.
+- `--warm-up=5` Number of seconds to spend warming up the connection before measuing begins. Default is 5.
```
swathipil, line 36: might be a typo?
```suggestion
- `--warm-up=5` Number of seconds to spend warming up the connection before measuring begins. Default is 5.
```

sdk/eventhub/azure-eventhub/tests/perfstress_tests/README.md
```diff
@@ -0,0 +1,58 @@
...
+(env) ~/azure-eventhub> pip install .
+```
+
+## Test commands
+
+When `azure-devtools` is installed, you will have access to the `perfstress` command line tool, which will scan the current module for runable perf tests. Only a specific test can be run at a time (i.e. there is no "run all" feature).
+
+```cmd
+(env) ~/azure-eventhub> cd tests
+(env) ~/azure-eventhub/tests> perfstress
```
rakshith91, line 27: i think it's better to give a complete example here including a file name 
or perhaps something  like
``` perfstress <ClassName> --<args>```

File: sdk/eventhub/azure-eventhub/tests/perfstress_tests/send_event_batch.py [hunk 3]
```diff
@@ -0,0 +1,40 @@
...
+
+
+class SendEventBatchTest(_SendTest):
+    def __init__(self, arguments):
+        super().__init__(arguments)
+        self.data = get_random_bytes(self.args.event_size)
+
+    def run_sync(self):
+        batch = self.producer.create_batch()
+        for i in range(self.args.num_events):
```
rakshith91, line 20: use _ here since i is not being used
```suggestion
        for _ in range(self.args.num_events):
```

sdk/eventhub/azure-eventhub/tests/perfstress_tests/send_event_batch.py
```diff
@@ -0,0 +1,40 @@
...
+                batch.add(EventData(self.data))
+            except ValueError:
+                # Batch full
+                self.producer.send_batch(batch)
+                batch = self.producer.create_batch()
+                batch.add(EventData(self.data))
+        self.producer.send_batch(batch)
+
+    async def run_async(self):
+        batch = await self.async_producer.create_batch()
```
rakshith91, line 31: is this something that can go in setup?
the run_** methods should have a single API ideally

sdk/eventhub/azure-eventhub/tests/perfstress_tests/send_event_batch.py
```diff
@@ -0,0 +1,40 @@
...
+        for i in range(self.args.num_events):
+            try:
+                batch.add(EventData(self.data))
+            except ValueError:
+                # Batch full
+                await self.async_producer.send_batch(batch)
+                batch = await self.async_producer.create_atch()
```
rakshith91, line 38: create_batch?

sdk/eventhub/azure-eventhub/tests/perfstress_tests/send_event_batch.py
rakshith91: please keep the run methods as simple as possible so that we can only test the api call performance - i would split create_batch and send_batch into two different perf tests

sdk/eventhub/azure-eventhub/tests/perfstress_tests/_test_base.py
```diff
@@ -0,0 +1,75 @@
...
+        eventhub_name = self.get_from_env("AZURE_EVENTHUB_NAME")
+        self.async_producer = AsyncEventHubProducerClient.from_connection_string(connection_string, eventhub_name=eventhub_name)
+        self.consumer = EventHubConsumerClient.from_connection_string(connection_string, _EventHubTest.consumer_group, eventhub_name=eventhub_name)
+        self.async_consumer = AsyncEventHubConsumerClient.from_connection_string(connection_string, _EventHubTest.consumer_group, eventhub_name=eventhub_name)
+
+    async def _preload_eventhub(self):
+        data = get_random_bytes(self.args.event_size)
+        async with self.async_producer as producer:
+            batch = await producer.create_batch()
+            for i in range(self.args.preload):
```
rakshith91, line 51:
```suggestion
            for _ in range(self.args.preload):
```

sdk/eventhub/azure-eventhub/tests/perfstress_tests/send_event_batch.py
yunhaoling: updated :)

sdk/eventhub/azure-eventhub/tests/perfstress_tests/send_event_batch.py
yunhaoling: thanks for catching the typo, updated

sdk/eventhub/azure-eventhub/tests/perfstress_tests/send_event_batch.py
yunhaoling: `create_batch` is majorly a local operation, I don't think we need a perf tests for that :P
but to simplify the `run` and `run_async`, I have taken your suggestion and moved event data batch creation into `__init__`.

sdk/eventhub/azure-eventhub/tests/perfstress_tests/README.md
yunhaoling: Sample command line is put at the bottom of the readme:

```
(env) ~/azure-eventhub/tests> perfstress ReceiveEventBatchTest --parallel=2 --event-size=1024 --num-events=100 --duration=100
```

sdk/eventhub/azure-eventhub/tests/perfstress_tests/README.md
```diff
@@ -0,0 +1,58 @@
+# EventHub Performance Tests
+
+In order to run the performance tests, the `azure-devtools` package must be installed. This is done as part of the `dev_requirements`.
+Start be creating a new virtual environment for your perf tests. This will need to be a Python 3 environment, preferably >=3.7.
```
swathipil, line 4:
```suggestion
Start by creating a new virtual environment for your perf tests. This will need to be a Python 3 environment, preferably >=3.7.
```

sdk/eventhub/azure-eventhub/tests/perfstress_tests/_test_base.py
```diff
@@ -0,0 +1,75 @@
...
+        await self.async_producer.close()
+        await super().close()
+
+
+class _ReceiveTest(_EventHubTest):
+    def __init__(self, arguments):
+        super().__init__(arguments)
+        connection_string = self.get_from_env("AZURE_EVENTHUB_CONNECTION_STRING")
+        eventhub_name = self.get_from_env("AZURE_EVENTHUB_NAME")
+        self.async_producer = AsyncEventHubProducerClient.from_connection_string(connection_string, eventhub_name=eventhub_name)
```
swathipil, line 43: qq: I'm assuming you're only using the async producer, rather than sync too, to send events since the consumers are what is being tested here. If you used only the sync producer, would it accomplish the same thing? (for my own curiosity)

yunhaoling: yes! the producer is just for preload the eventhub with events, so either sync or async producer would work.

sdk/eventhub/azure-eventhub/tests/perfstress_tests/receive_event.py
```diff
@@ -0,0 +1,39 @@
...
+        if self._total_received_cnt > self.args.num_events:
+            self.consumer.close()
+
+    async def _on_event_received_async(self, partition_context, event):
+        self._async_total_received_cnt += 1
+        if self._async_total_received_cnt > self.args.num_events:
+            await self.async_consumer.close()
+
+    def run_sync(self):
+        self._total_received_cnt = 0
```
rakshith91, line 26: this can move to setup - no?

sdk/eventhub/azure-eventhub/tests/perfstress_tests/receive_event.py
```diff
@@ -0,0 +1,39 @@
...
+        self.consumer.receive(
+            on_event=self._on_event_received,
+            max_wait_time=self.args.max_wait_time,
+            starting_position="-1",
+        )
+
+    async def run_async(self):
+        self._async_total_received_cnt = 0
```
rakshith91, line 34: this can move to setup - no?

sdk/eventhub/azure-eventhub/tests/perfstress_tests/receive_event.py
yunhaoling: My understanding is that each time after we finish the `run` method, the status related to each running shall be reset -- in the EH case, the received cnt shall be cleaned up.
well, this is some overhead introduced by the push callback mode  -- another choice is to use a nested method for callback so that we could keep every status inside the method scope, but it further complicates the run method, so I'd rather to keep it as it is for now to make it work first and improve in the future.

sdk/eventhub/azure-eventhub/tests/perfstress_tests/receive_event.py
yunhaoling: same as above, any thoughts?

sdk/eventhub/azure-eventhub/tests/perfstress_tests/_test_base.py
annatisch: Maybe we could name it accordingly - to avoid confusing it with the testable clients. Something like `self.event_preloader`

sdk/eventhub/azure-eventhub/tests/perfstress_tests/receive_event.py
```diff
@@ -0,0 +1,39 @@
...
+
+class ReceiveEventTest(_ReceiveTest):
+    def __init__(self, arguments):
+        super().__init__(arguments)
+        self._total_received_cnt = 0
+        self._async_total_received_cnt = 0
+
+    def _on_event_received(self, partition_context, event):
+        self._total_received_cnt += 1
+        if self._total_received_cnt > self.args.num_events:
```
annatisch, line 17: `>=`?
Not that I suppose one event makes much difference :D

sdk/eventhub/azure-eventhub/tests/perfstress_tests/receive_event.py
annatisch: Yes I think this needs to be here. Otherwise the counts will continue to accumulate between runs.

sdk/eventhub/azure-eventhub/tests/perfstress_tests/send_event_batch.py
```diff
@@ -0,0 +1,38 @@
...
+        batch = self.producer.create_batch()
+        for _ in range(self.args.num_events):
+            try:
+                batch.add(EventData(self.data))
+            except ValueError:
+                # Batch full
+                self.to_send_event_data_batches.append(batch)
+                batch = self.producer.create_batch()
+                batch.add(EventData(self.data))
+        self.to_send_event_data_batches.append(batch)
```
annatisch, line 30: We should make this part of the actual run. It's part of the code that a user would have to write, and they would be unlikely to separate out batch creation from sending like this - so our test should reflect that.

We may need to do a `globa_setup` that "starts" the connection so that when the first `create_batch` method is called we have already got the supported batch size.

sdk/eventhub/azure-eventhub/tests/perfstress_tests/receive_event.py
yunhaoling: you're right! if it's an empty eh, then even one event might keeps the test from ending itself!
thanks to calling it out, updated!

sdk/eventhub/azure-eventhub/tests/perfstress_tests/_test_base.py
```diff
@@ -0,0 +1,75 @@
...
+
+    async def close(self):
+        self.consumer.close()
+        await self.async_consumer.close()
+        await super().close()
+
+    @staticmethod
+    def add_arguments(parser):
+        super(_ReceiveTest, _ReceiveTest).add_arguments(parser)
+        parser.add_argument('--max-wait-time', nargs='?', type=int, help='Max time to wait for events before closing. Defaults to 0.', default=0)
```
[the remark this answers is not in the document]
yunhaoling: yea, definitely make sense to have a timeout instead of infinite waiting

sdk/eventhub/azure-eventhub/tests/perfstress_tests/send_event_batch.py
yunhaoling: updated code to have an explicit call `creat_batch` in `gloab_setup` which will communicate with the service and cache the max allowed link frame size, also updated sending part.

sdk/eventhub/azure-eventhub/tests/perfstress_tests/receive_event.py
yunhaoling: (see comment above)

annatisch (approved):

PR #17162, opened 2021-03-08
[Communication] - Phone Numbers - Changed User-Agent in request

jbeauregardb:


sdk/communication/azure-communication-phonenumbers/test/_shared/testcase.py
```diff
@@ -25,14 +25,14 @@ def process_response(self, response):
                     if isinstance(item, str):
                         body["phoneNumbers"] = [self._replacement]
                         break
-                    elif "phoneNumber" in item:
+                    if "phoneNumber" in item:
                         item['phoneNumber'] = self._replacement
-                    elif "id" in item:
```
sacheu, line 30: why we change from elif to if here?

sdk/communication/azure-communication-phonenumbers/test/_shared/testcase.py
```diff
@@ -25,14 +25,14 @@ def process_response(self, response):
                     if isinstance(item, str):
                         body["phoneNumbers"] = [self._replacement]
                         break
-                    elif "phoneNumber" in item:
```
sacheu, line 28: why we change from elif to if here?

sdk/communication/azure-communication-phonenumbers/test/_shared/testcase.py
```diff
@@ -25,14 +25,14 @@ def process_response(self, response):
...
+                    if "phoneNumber" in item:
                         item['phoneNumber'] = self._replacement
-                    elif "id" in item:
+                    if "id" in item:
                         item['id'] = self._replacement
             response['body']['string'] = json.dumps(body)
             response['url'] = self._replacement
             return response
-        except (KeyError, ValueError):
+        except (KeyError, ValueError, TypeError):
```
sacheu, line 35: why we need this change for the User-agent change?

sacheu (commented):
i left a few comments.

sdk/communication/azure-communication-phonenumbers/test/_shared/testcase.py
jbeauregardb: This is a fix for the type error issue we have in the search available phone numbers test. We weren't catching that exception before. It's totally unrelated to the user agent issue, though.

sdk/communication/azure-communication-phonenumbers/test/_shared/testcase.py
jbeauregardb: I wanted to sanitize both values. Having an elif only sanitized the phone number or the id, but not both at the same time.

sacheu (approved):
Thanks. Looks good.

PR #17168, opened 2021-03-08
Rsarkar/identity accesstoken fix

sarkar-rajarshi:
- Initialize CommunicationTokenCredential with token and not with CommunicationTokenRefreshOptions
- Convert AccessToken.expires_on to int as compared to datetime


sdk/communication/azure-communication-chat/azure/communication/chat/_shared/user_credential_async.py
```diff
@@ -22,8 +23,9 @@ class CommunicationTokenCredential(object):
     ON_DEMAND_REFRESHING_INTERVAL_MINUTES = 2
 
     def __init__(self,
-            communication_token_refresh_options
+            token # type: str
```
RezaJooyandeh, line 26: `CommunicationTokenRefreshOptions` has two arguments: `tokenRefresher` and `token`. Let's make sure we pass them both here as well

RezaJooyandeh: You can also remove the `CommunicationTokenRefreshOptions` type altogether and move `_token` and `_token_refresher` to `CommunicationTokenCredential`

sdk/communication/azure-communication-chat/azure/communication/chat/_shared/user_credential.py
```diff
@@ -21,8 +22,9 @@ class CommunicationTokenCredential(object):
     ON_DEMAND_REFRESHING_INTERVAL_MINUTES = 2
 
     def __init__(self,
-            communication_token_refresh_options
+            token # type: str
```
annatisch, line 25: Do we need to add `**kwargs` here to support passing in the "token refresher"?
Or is this not really needed for now?

sdk/communication/azure-communication-identity/azure/communication/identity/_shared/utils.py
```diff
@@ -65,10 +65,16 @@ def create_access_token(token):
     try:
         padded_base64_payload = base64.b64decode(parts[1] + "==").decode('ascii')
         payload = json.loads(padded_base64_payload)
-        return AccessToken(token, datetime.fromtimestamp(payload['exp']).replace(tzinfo=TZ_UTC))
+        return AccessToken(token,
+            _convert_expires_on_datetime_to_utc_int(datetime.fromtimestamp(payload['exp']).replace(tzinfo=TZ_UTC)))
```
annatisch, line 69: What was this extra conversion for? Looks like we're going `str` -> `datetime` -> `int`?
Is there a reason we do the `int` conversion here, rather than when this `AccessToken` is consumed in the pipeline? Returning an int here makes it non-human-readable if a user is expected to interact with it.

sdk/communication/azure-communication-chat/azure/communication/chat/_shared/user_credential.py
sarkar-rajarshi: @annatisch The token refresher has just the two fields curently. Don't think we need **kwargs. What is the recommended approach though?

sdk/communication/azure-communication-chat/azure/communication/chat/_shared/user_credential_async.py
sarkar-rajarshi: @RezaJooyandeh Haven't removed the CommunicationTokenRefreshOptions type completely. Added the 2 fields in CommunicationTokenCredential

sarkar-rajarshi:
abandoning this since this change is already merged in #17178

PR #17180, opened 2021-03-09
Initial ModelsRepositoryClient

cartertinney:
Initial ModelsRepositoryClient

Unittests + some integration tests will be added after API review signoff


sdk/iot/azure-iot-modelsrepository/azure/iot/modelsrepository/__init__.py
```diff
@@ -0,0 +1,18 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+# --------------------------------------------------------------------------
+
+# Main Client
+from .client import ModelsRepositoryClient
+
```
digimaun, line 9: Given the recent pivot in the C# SDK to provide utility functions, it would be good to expose`is_valid_dtmi()` and `get_model_path()` functionality. Not directly part of the client.

sdk/iot/azure-iot-modelsrepository/azure/iot/modelsrepository/client.py
```diff
@@ -0,0 +1,173 @@
...
+        :param list[str]: The DTMIs for the models you wish to retrieve
+        :param str dependency_resolution : Dependency resolution mode. Possible values:
+            - "disabled": Do not resolve model dependencies
+            - "enabled": Resolve model dependencies from the repository
+            - "tryFromExpanded": Attempt to resolve model and dependencies from an expanded DTDL
+                    document in the repository. If this is not successful, will fall back on
+                    manually resolving dependencies in the repository
+
+        :raises: ValueError if given an invalid dependency resolution mode
+        :raises: ResolverError if there is an error retreiving a model
```
digimaun, line 76: Small typo `retrieving`.

With respect to service operations errors (after getting through argument validation errors where `ValueError` is being thrown) we need to ensure we are raising the proper exceptions with respect to Track 2/azure-core expectations. C# wise `RequestFailedException` (an Azure.Core type) is thrown when run into issues in either remote or local flows.

sdk/iot/azure-iot-modelsrepository/azure/iot/modelsrepository/client.py
```diff
@@ -0,0 +1,173 @@
...
+        else:
+            raise ValueError("Invalid dependency resolution mode: {}".format(dependency_resolution))
+        return model_map
+
+
+class ModelsRepositoryClientConfiguration(Configuration):
+    """ModelsRepositoryClient-specific variant of the Azure Core Configuration for Pipelines"""
+
+    def __init__(self, **kwargs):
+        super(ModelsRepositoryClientConfiguration, self).__init__(**kwargs)
```
digimaun, line 106: Would be good to itemize in docstrings kwargs that can be passed in.

cartertinney: That was my first thought, however it seems fairly impractical given how kwargs work in the Track 2 context - all kwargs are passed to all internal objects, so kwargs are the superset of all options related to not just the internal pipeline, but all policies, etc. I suspect this is why no other python clients seem to be listing them either. This is something I will raise during review

sdk/iot/azure-iot-modelsrepository/azure/iot/modelsrepository/client.py
cartertinney: Another issue I will raise during review is how to present a uniform exception surface across both local and remote repositories, given that only one of them uses an azure.core Pipeline.

sdk/iot/azure-iot-modelsrepository/azure/iot/modelsrepository/pseudo_parser.py
```diff
@@ -0,0 +1,58 @@
...
+        ]
+
+        if dependencies_to_resolve:
+            resolved_dependency_map = self.resolver.resolve(dependencies_to_resolve)
+            model_map.update(resolved_dependency_map)
+            for dependency_model in resolved_dependency_map.items():
+                self._expand(dependency_model, model_map)
+
+
+def get_dependency_list(model):
```
digimaun, line 41: Once you onboard the model resolution tests and evaluate dependencies, the output of the tests would inform changes here.

For example in the context of json, extends can also be an inline object, or an array of mixed strings and objects. Component schemas are similar but more limited - the value can be a string, object, an array with a single string, an array with a single object (but the single element in an array constraint is not that important at our layer).

Please use this [implementation](https://github.com/Azure/azure-sdk-for-net/blob/master/sdk/modelsrepository/Azure.Iot.ModelsRepository/src/ModelQuery.cs) as an example in conjunction with supporting the same kinds of tests.

sdk/iot/azure-iot-modelsrepository/azure/iot/modelsrepository/client.py
digimaun: Sounds good

sdk/iot/azure-iot-modelsrepository/azure/iot/modelsrepository/resolver.py
```diff
@@ -0,0 +1,148 @@
...
+# --------------------------------------------------------------------------
+import logging
+import six
+import json
+import abc
+import re
+import os
+from .chainable_exception import ChainableException
+
+logger = logging.getLogger(__name__)
```
digimaun, line 14: Planning to add more logging plumbing throughout?  I want to ensure we are applying our [best effort](https://azure.github.io/azure-sdk/python_implementation.html#python-logging-usage) with respect to logging events at the right levels in key operations so our users can quickly diagnose a problem (outside of logging we get for free).

sdk/iot/azure-iot-modelsrepository/azure/iot/modelsrepository/resolver.py
```diff
@@ -0,0 +1,148 @@
...
+            try:
+                dtdl = self.fetcher.fetch(dtdl_path)
+            except FetcherError as e:
+                raise ResolverError("Failed to resolve dtmi: {}".format(dtmi), cause=e)
+
+            if expanded_model:
+                for model in dtdl:
+                    model_map[model["@id"]] = model
+            else:
+                model_map[dtmi] = dtdl
```
digimaun, line 56: Dtmis are case sensitive. When not grabbing models from the expanded model file, the client needs to ensure the dtmi extracted from a fetched model matches in case with what was requested/expected.

sdk/iot/azure-iot-modelsrepository/azure/iot/modelsrepository/client.py
```diff
@@ -0,0 +1,173 @@
...
+
+        # NOTE: depending on how this class develops over time, may need to adjust relationship
+        # between some of these objects
+        self.fetcher = _create_fetcher(
+            location=repository_location, api_version=api_version, **kwargs
+        )
+        self.resolver = resolver.DtmiResolver(self.fetcher)
+        self.pseudo_parser = pseudo_parser.PseudoParser(self.resolver)
+
+    def get_models(self, dtmis, dependency_resolution=DEPENDENCY_MODE_DISABLED):
```
digimaun, line 64: Would be good to support [distributed tracing](https://azure.github.io/azure-sdk/python_implementation.html#distributed-tracing) for client service methods (which only includes `get_models` in this first iteration goal) and apply the patterns where we need to.

sdk/iot/azure-iot-modelsrepository/azure/iot/modelsrepository/client.py
```diff
@@ -0,0 +1,173 @@
...
+# Convention-private constants
+_DEFAULT_LOCATION = "https://devicemodels.azure.com"
+_DEFAULT_API_VERSION = "2021-02-11"
+_REMOTE_PROTOCOLS = ["http", "https"]
+
+
+class ModelsRepositoryClient(object):
+    """Client providing APIs for Models Repository operations"""
+
+    def __init__(self, repository_location=None, api_version=None, **kwargs):
```
digimaun, line 41: Something to consider when no `repository_location` is given.  In this case we default to the global IoT models repository endpoint and here we can guarantee an expanded model file for any given model is available.

The C# implementation supports an optional model dependency resolution option through its initialization/ctor's client options, but if you don't provide one and are pointing to the global endpoint the default model dependency resolution becomes the try from expanded mode.

sdk/iot/azure-iot-modelsrepository/azure/iot/modelsrepository/client.py
digimaun, line 64: I think most of the time in our use cases, if you are looking up a model with dependencies you will want model dependencies in your result payload. Also in the Models Repository model ingestion workflow, the user is expected to fork, clone and iterate against their local copy of the repository. This client would be used to validate and test against it. If I am creating a new model with dependencies, I want to make sure the model and its dependencies are all available and in the right places.

sdk/iot/azure-iot-modelsrepository/azure/iot/modelsrepository/client.py
```diff
@@ -0,0 +1,173 @@
...
+        self.fetcher = _create_fetcher(
+            location=repository_location, api_version=api_version, **kwargs
+        )
+        self.resolver = resolver.DtmiResolver(self.fetcher)
+        self.pseudo_parser = pseudo_parser.PseudoParser(self.resolver)
+
+    def get_models(self, dtmis, dependency_resolution=DEPENDENCY_MODE_DISABLED):
+        """Retrieve a model from the Models Repository.
+
+        :param list[str]: The DTMIs for the models you wish to retrieve
```
digimaun, line 67: What about annotating `dtmis` as a [union type](https://docs.python.org/3/library/typing.html#typing.Union) so the user could pass in a single dtmi string or a list of dtmi strings?

cartertinney: We cannot support union types as it was only introduced in Python 3.7

sdk/iot/azure-iot-modelsrepository/azure/iot/modelsrepository/client.py
cartertinney: Are you saying you want me to change the default resolution option to DEPENDENCY_MODE_ENABLED?

sdk/iot/azure-iot-modelsrepository/azure/iot/modelsrepository/resolver.py
cartertinney: I will add judicious and consistent logging once the design is finalized - I want to make sure that logging is equally applied throughout and nothing slips through.

sdk/iot/azure-iot-modelsrepository/azure/iot/modelsrepository/resolver.py
cartertinney: It was my understanding as per our previous conversations in meetings that we are not validating any model contents beyond what was strictly necessary in order to find dependencies.

If this is considered necessary, I can make the change, however what is the expected behavior if there is a mismatch?

sdk/iot/azure-iot-modelsrepository/azure/iot/modelsrepository/client.py
digimaun: Yea I agree with you on the not great semantics. This was an idea to optimize devEx so you don't have to pass a list in for one model.

sdk/iot/azure-iot-modelsrepository/azure/iot/modelsrepository/client.py
cartertinney: To be clear, the desired design is to have a dependency mode tied to the instance of the client, that can be overridden when using the `.get_models()` API by passing in an optional argument?

I do think this relationship (especially when factoring in the elsewhere-mentioned dynamic default dependency mode based upon target URI) could be somewhat confusing and difficult to capture in documentation, but I can implement that if it's the requirement.

digimaun: I was addressing the other https://github.com/Azure/azure-sdk-for-python/pull/17180#discussion_r591987457 around this topic but will continue here and what you said is the gist of it.

1. Have an optional dependency resolution mode in client initialization
2. If no repository location is chosen, we default to the global endpoint. If we are pointing to the global endpoint and no dependency resolution is given on init, we default to try from expanded mode because the service will have these types of files.
3. If any other repository location is given the default dependency resolution mode is enabled.
4. Like you said `.get_models()` has an optional parameter which will override the client instance dep mode.

The samples here tries to explain the dependency resolution when pointing to the global endpoint https://github.com/Azure/azure-sdk-for-net/tree/master/sdk/modelsrepository/Azure.Iot.ModelsRepository/samples#initializing-the-models-repository-client

sdk/iot/azure-iot-modelsrepository/azure/iot/modelsrepository/client.py
digimaun: Please see this https://github.com/Azure/azure-sdk-for-python/pull/17180#discussion_r592671524 which tries to summarize

sdk/iot/azure-iot-modelsrepository/azure/iot/modelsrepository/resolver.py
digimaun: This is necessary. The service convention lowercases all model file paths. The client that fetches models needs to make sure the fetched models definition has exactly the requested dtmi. If there is a mismatch the client needs to raise the idiomatic service exception. In C# we are raising the `RequestFailedException`. At this point in time state of the code, the  ResolverError would fit.

sdk/iot/azure-iot-modelsrepository/azure/iot/modelsrepository/_constants.py
```diff
@@ -0,0 +1,12 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+# --------------------------------------------------------------------------
+import platform
+
+VERSION = "0.0.0-preview"
```
digimaun, line 8: This will not be a compliant version string. We should go with `1.0.0b1` or `1.0.0a1` which are compatible with pep440 and sdk guidelines. When we iterate on the preview SDK we can increment the pre-release segment. This is also aligned with SDKs for Java and dotnet.

sdk/iot/azure-iot-modelsrepository/README.md
```diff
@@ -0,0 +1,16 @@
...
+This package is not yet available on pip. Please install locally from source:
+
+```Shell
+python -m pip install -e <path to azure-iot-modelsrepository>
+```
+
+## Features
+
+* ### ModelsRepositoryClient
+    * Allows retrieval of model DTDLs from remote URLs or local filesystems
```
digimaun, line 16: I think "Allows the retrieval of model definitions..." or "Allows the retrieval of digital twin model definitions..." would be more clear.

sdk/iot/azure-iot-modelsrepository/README.md
```diff
@@ -0,0 +1,16 @@
+# Azure IoT Models Repository Library
+
+The Azure IoT Models Repository Library for Python provides functionality for working with the Azure IoT Models Repository
+
+## Installation
+
+This package is not yet available on pip. Please install locally from source:
```
digimaun, line 7: Are you planning to update the installation instructions since we aim to release a package on PyPI?

sdk/iot/azure-iot-modelsrepository/azure/iot/modelsrepository/_client.py
```diff
@@ -0,0 +1,234 @@
...
+            Possible values:
+                - "disabled": Do not resolve model dependencies
+                - "enabled": Resolve model dependencies from the repository
+                - "tryFromExpanded": Attempt to resolve model and dependencies from an expanded DTDL
+                        document in the repository. If this is not successful, will fall back on
+                        manually resolving dependencies in the repository
+
+        :raises: ValueError if given an invalid dependency resolution mode
+        :raises: ModelError if there is an error parsing the retrieved model(s)
+        :raises: ???? if there is an error retrieving a model
```
digimaun, line 127: Placeholder to update `????`

sdk/iot/azure-iot-modelsrepository/azure/iot/modelsrepository/_client.py
```diff
@@ -0,0 +1,234 @@
...
+        """
+        if isinstance(dtmis, str):
+            dtmis = [dtmis]
+
+        dependency_resolution = kwargs.get("dependency_resolution", self.resolution_mode)
+
+        if dependency_resolution == DEPENDENCY_MODE_DISABLED:
+            # Simply retrieve the model(s)
+            _LOGGER.debug("Getting models w/ dependency resolution mode: disabled")
+            _LOGGER.debug("Retreiving model(s): %s...", dtmis)
```
digimaun, line 140: Small typo on `Retrieving`

sdk/iot/azure-iot-modelsrepository/azure/iot/modelsrepository/_constants.py
cartertinney: Python will automatically convert the string "preview" into "rc" which is PEP440 compliant. So this semantic version would be 1.0.0-preview.1 (i.e. 1.0.0-rc1)

"rc" indicates "release candidate". Do you have a preference between the semantics of "alpha"/"beta"/"release candidate"?

sdk/iot/azure-iot-modelsrepository/README.md
cartertinney: I had intended to update this after the package was released, but I can go ahead and do it now.

sdk/iot/azure-iot-modelsrepository/azure/iot/modelsrepository/_client.py
cartertinney: Fixed

sdk/iot/azure-iot-modelsrepository/azure/iot/modelsrepository/_client.py
cartertinney: Fixed

sdk/iot/azure-iot-modelsrepository/azure/iot/modelsrepository/_constants.py
digimaun: Rather than rely on packaging to convert a non-pep440 compliant pre-release segment to a compliant one, it would be better to commit a compliant one in source code.

pep440 has a section on [semantic versioning](https://www.python.org/dev/peps/pep-0440/#semantic-versioning) which states "Semantic versions containing a hyphen (pre-releases - clause 10) or a plus sign (builds - clause 11) are not compatible with this PEP and are not permitted in the public version field."

The azure SDK guidelines also state:

image

Furthermore we can compare with prior-art with respect to pre-release packages in the track 2 mono repo. The preview azure quantum mgmt SDK release is versioned as [`1.0.0b1`](https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/quantum/azure-mgmt-quantum/azure/mgmt/quantum/_version.py)

Another example with the preview azure container registry package is versioned as [`1.0.0b1`](https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/containerregistry/azure-containerregistry/azure/containerregistry/_version.py)

Given the above reasons I would prefer an explicit `1.0.0b1`, `1.0.0a1`, or `1.0.0rc1` with my most preferred being `1.0.0b1`.

sdk/iot/azure-iot-modelsrepository/azure/iot/modelsrepository/dtmi_conventions.py
```diff
@@ -0,0 +1,65 @@
...
+
+    :raises ValueError if DTMI is invalid
+
+    :returns: Relative path of the model in a Models Repository
+    :rtype: str
+    """
+    pattern = re.compile(
+        "^dtmi:[A-Za-z](?:[A-Za-z0-9_]*[A-Za-z0-9])?(?::[A-Za-z](?:[A-Za-z0-9_]*[A-Za-z0-9])?)*;[1-9][0-9]{0,8}$"
+    )
+    if not pattern.match(dtmi):
```
digimaun, line 60: We could call `is_valid_dtmi()` here and reduce duplication.

sdk/iot/azure-iot-modelsrepository/samples/get_models_sample.py
```diff
@@ -0,0 +1,58 @@
...
+    # i.e. https://devicemodels.azure.com/dtmi/com/example/temperaturecontroller-1.expanded.json
+    with ModelsRepositoryClient() as client:
+        model_map = client.get_models(
+            dtmis=[dtmi], dependency_resolution=DEPENDENCY_MODE_TRY_FROM_EXPANDED
+        )
+        pprint.pprint(model_map)
+
+
+def get_model_and_dependencies():
+    # This API call will return a dictionary mapping the specified DTMI to it's corresponding model,
```
digimaun, line 52: Think you don't need to apostrophe here for `its`

sdk/iot/azure-iot-modelsrepository/azure/iot/modelsrepository/_client.py
```diff
@@ -0,0 +1,236 @@
...
+            DEPENDENCY_MODE_ENABLED,
+            DEPENDENCY_MODE_DISABLED,
+            DEPENDENCY_MODE_TRY_FROM_EXPANDED,
+        ]:
+            raise ValueError("Invalid dependency resolution mode: {}".format(self.resolution_mode))
+        _LOGGER.debug("Client configured for dependency mode %s", self.resolution_mode)
+
+        # NOTE: depending on how this class develops over time, may need to adjust relationship
+        # between some of these objects
+        self.fetcher = _create_fetcher(location=repository_location, **kwargs)
```
digimaun, line 90: Should the instance bound fetcher and resolver also have an underscore prefix/private convention like `_psuedo_parser`?

sdk/iot/azure-iot-modelsrepository/azure/iot/modelsrepository/_client.py
```diff
@@ -0,0 +1,236 @@
...
+        self.resolver = _resolver.DtmiResolver(self.fetcher)
+        self._psuedo_parser = _pseudo_parser.PseudoParser(self.resolver)
```
digimaun, line 92: spelling...`self._pseudo_parser`

sdk/iot/azure-iot-modelsrepository/samples/README.md
```diff
@@ -0,0 +1,12 @@
+# Azure IoT Models Repository Library Samples
+
+This directory contains samples showing how to use the features of the Azure IoT Models Repository Library.
+
+The pre-configured endpoints and DTMIs within the sampmles refer to example DTDL documents that can be found on [devicemodels.azure.com](https://devicemodels.azure.com/). These values can be replaced to reflect the locations of your own DTDLs, wherever they may be.
```
digimaun, line 5: spelling `samples`

sdk/iot/azure-iot-modelsrepository/azure/iot/modelsrepository/_client.py
cartertinney: I only made attributes convention-prviate for the client itself, for the inner classes I did not add convention-private attributes as they are not part of the customer-facing surface, and the SDK guidelines do not require me to do so, nor is it particularly common to rigidly enforce this convention on attributes.

I did choose to use convention-private methods on inner classes, because I find those helpful during development. 

If you feel particularly strongly on this point, I can go make everything convention private.

sdk/iot/azure-iot-modelsrepository/samples/README.md
```diff
@@ -0,0 +1,12 @@
...
+
+## Resolver Samples
```
digimaun, line 7: Should be "Models Repository client Samples" or something to that effect.

sdk/iot/azure-iot-modelsrepository/samples/README.md
digimaun: "These values can be replaced to reflect the locations of your own **DTDL models**, wherever they may be."

sdk/iot/azure-iot-modelsrepository/azure/iot/modelsrepository/_client.py
digimaun: I don't. We can consider this resolved.

digimaun:
Can we append at least you and me from the IoT D&D team as code owners for this project?

sdk/iot/tests.yml
```diff
@@ -0,0 +1,10 @@
+trigger: none
+
+stages:
+  - template: ../../eng/pipelines/templates/stages/archetype-sdk-tests.yml
+    parameters:
+      AllocateResourceGroup: 'false'
+      BuildTargetingString: $(BuildTargetingString)
+      ServiceDirectory: iot
+      EnvVars:
+        TEST_MODE: 'RunLiveNoRecord'
```
digimaun, line 10: Normally we'd want to run recordings in the pipeline. The live tests happen locally and test recordings would be checked-in.

sdk/iot/azure-iot-modelsrepository/README.md
```diff
@@ -0,0 +1,57 @@
...
+Install the Azure IoT Models Repository library for Python with [pip][pip]:
+
+```Shell
+pip install azure-iot-modelsrepository
+```
+
+### Prerequisites
+* A models repository following [repo_conventions][Azure IoT conventions]
+    * The models repository can be hosted on the local filesystem or hosted on a webserver
+    * Azure IoT hosts the global [][Azure IoT Models Repository] which the client will use if no custom location is provided
```
digimaun, line 18: missing value [][Azure IoT Models Repository]

sdk/iot/azure-iot-modelsrepository/tests/test_dtmi_conventions.py
```diff
@@ -0,0 +1,139 @@
...
+                "http://myrepository/dtmi/com/somedomain/example/foodtdl-1.json",
+                id="HTTP repository URI",
+            ),
+            pytest.param(
+                "dtmi:com:somedomain:example:FooDTDL;1",
+                "file:///myrepository/",
+                "file:///myrepository/dtmi/com/somedomain/example/foodtdl-1.json",
+                id="Filesystem URI",
+            ),
+            pytest.param(
```
digimaun, line 62: Would be good to add a test case for basic windows UNC path...as an example getting the model uri for "dtmi:com:example:Thermostat;1" in the repository location of "\\\\server\\\\repository" should net the uri of "file://server/repository/dtmi/com/example/thermostat-1.json"

scbedd:
@cartertinney The failure for `verifyChangeLog` is just because there is no changelog. I recommend adding a bare one with content of:

```
# Release History

## 1.0.0b1 (Unreleased)
```

To unblock the check. Prior to your actual release, you'll need to update the date, something like.

```
## 1.0.0b1 (2021-05-01)

* Initial Release
* Feature X/Y/Z supported

```

Or the like.

sdk/iot/ci.yml
```diff
@@ -30,5 +30,7 @@ extends:
   parameters:
     ServiceDirectory: iot
```
digimaun, line 31: I'm not sure the models repository assets should be rooted in `/sdk/iot/`. In both dotnet and java its rooted in `/sdk/modelsrepository/` and this is what was approved by the architecture board.

sdk/iot/azure-iot-modelsrepository/setup.py
```diff
@@ -0,0 +1,82 @@
...
+    version=VERSION,
+    description="Microsoft Azure IoT Models Repository Library",
+    license="MIT License",
+    author="Microsoft Corporation",
+    author_email="azpysdkhelp@microsoft.com",
+    url="https://github.com/Azure/azure-sdk-for-python",
+    long_description=_long_description,
+    long_description_content_type="text/markdown",
+    classifiers=[
+        "Development Status :: 3 - Alpha",
```
digimaun, line 50: We should update the development status classifier to 4 - beta....`Development Status :: 4 - Beta` to reflect the updated package version and pre-release segment.

sdk/iot/azure-iot-modelsrepository/setup.py
```diff
@@ -0,0 +1,82 @@
...
+            "tests",
+            "tests.*",
+            "samples",
+            "samples.*",
+            # Exclude packages that will be covered by PEP420 or nspkg
+            "azure",
+            "azure.iot",
+        ]
+    ),
+    zip_safe=False,
```
digimaun, line 81: What about adding something like the following entry for `project_urls`
```
project_urls={
    "Bug Reports": "https://github.com/Azure/azure-sdk-for-python/issues",
    "Source": "https://github.com/Azure/azure-sdk-python",
}
```

sdk/iot/tests.yml
cartertinney: Recordings have not yet been generated due to a what appears to be a bug in the testing infrastructure. This will be changed once recordings can be generated.

sdk/iot/ci.yml
cartertinney: We all approved this namespace. If you want to change it now, this is a different discussion to be had a later date.

sdk/iot/azure-iot-modelsrepository/tests/test_dtmi_conventions.py
cartertinney: The function `.get_model_uri()` has been scoped to require the base path to already be provided in URI format (as opposed to the ModelsRepositoryClient itself, which does support non-URI filepaths). If you wish to expand it's functionality to support absolute paths being converted to URI, this would be an additional feature that can be added at a later date.

sdk/iot/ci.yml
digimaun: I'm not referring to the namespace, I am referring to where in the sdk mono repo this SDK should exist.

According to the [Azure SDK repo directory structure](https://github.com/Azure/azure-sdk/blob/master/docs/policies/repostructure.md#sdk-directory-layout) the pattern is `sdk/<service name>/<package name>/*` where `<service name>` - should be the short name for the azure service.

This is the models repository service and therefore `sdk/modelsrepository/*` as the service root makes sense. And dotnet and java share `/sdk/modelsrepository/*`

sdk/iot/azure-iot-modelsrepository/azure/iot/modelsrepository/_client.py
```diff
@@ -0,0 +1,236 @@
...
+                - "tryFromExpanded": Attempt to resolve model and dependencies from an expanded
+                        model DTDL document in the repository. If this is not successful, will fall
+                        back on manually resolving dependencies in the repository
+
+        :raises: ValueError if given an invalid dependency resolution mode
+        :raises: ModelError if there is an error parsing the retrieved model(s)
+        :raises: ResourceNotFoundError if the model(s) cannot be found in the repository
+        :raises: ServiceRequestError if there is an error sending a request for the model(s)
+        :raises: ServiceResponseError if the model(s) cannot be retrieved
+        :raises: HttpResponseError if a failure respoonse is received
```
digimaun, line 128: Typo `respoonse`

sdk/iot/azure-iot-modelsrepository/samples/get_models_sample.py
```diff
@@ -0,0 +1,58 @@
...
+import pprint
+
+dtmi = "dtmi:com:example:TemperatureController;1"
+dtmi2 = "dtmi:com:example:TemperatureController;2"
+
+# By default the clients in this sample will use the Azure Device Models Repository endpoint
+# i.e. https://devicemodels.azure.com/
+# See client_configuration_sample.py for examples of alternate configurations
+
+
```
digimaun, line 20: Can we add a couple samples here which show users examples of error handling with respect to `get_models`?

sdk/iot/ci.yml
cartertinney: As this is a change with both infrastructure and semantic implications, I would much prefer it be done in a separate PR after a discussion involving all relevant parties.

digimaun: Ok. We could punt to a subsequent PR.

But note the repo directory structure spec for the `<service>` part also states "These names should match across all the different language repos..." and we've already committed to `modelsrepository` for 2 other languages.

@johanste what do you think about refactoring placement of the Python models repository SDK in `sdk/modelsrepository/` as its committed in dotnet and Java, versus the current placement in `sdk/iot/`?

digimaun: @annatisch , @lmazuel for more perspectives.

cartertinney: FWIW I would say that I do completely agree that all languages should eventually be organized the same way. The question is which is most correct? - the majority of clients are organized according to their <group> namespace. There are exceptions here and there, but I don't know what goes into that decision exactly, and if we don't consider this part of the `iot` folder, I would have to wonder why we're in the `iot` namespace at all.

drwill-ms: According to the Azure SDK guidelines, the root folder name should be the name of the service.

From Laurent Mazuel:
    
> <https://github.com/Azure/azure-sdk/blob/master/docs/policies/repostructure.md#sdk-directory-layout>
> This is the Azure SDK parent repository and mostly contains documentation around guidelines and policies as well as the releases for the various languages supported by the Azure SDK.
> The folder name should be the service name, not the group name
​> So according to our own guidelines, "deviceupdate" is the right folder name to me

digimaun (approved):
We have a number of follow up topics but I think we are at a solid baseline to build on.

sdk/modelsrepository/tests.yml
```diff
@@ -0,0 +1,10 @@
+trigger: none
+
+stages:
+  - template: ../../eng/pipelines/templates/stages/archetype-sdk-tests.yml
+    parameters:
+      AllocateResourceGroup: 'false'
+      BuildTargetingString: $(BuildTargetingString)
+      ServiceDirectory: modelsrepository
+      EnvVars:
+        TEST_MODE: 'RunLiveNoRecord'
```
seankane-msft, line 10:
```suggestion
      EnvVars:
        TEST_MODE: 'RunLiveNoRecord'
        AZURE_SKIP_LIVE_RECORDING: 'True'
        AZURE_TEST_RUN_LIVE: 'true'
```

scbedd:
Overriding check enforcer, this is the initial cut at the new SDK, and no build definition yet exists. @cartertinney will submit PRs to deal with any failures AFTER the build definition exists.

PR #17188, opened 2021-03-09
Fixed an issue when missing padding in base64 decode

craigktreasure:
- Apparently JWT doesn't use base64 padding and the Python base64 decoder requires it. So, we add the max padding and the decoder trims it properly.

### DIFF
--- a/sdk/mixedreality/azure-mixedreality-authentication/azure/mixedreality/authentication/_utils.py
+++ b/sdk/mixedreality/azure-mixedreality-authentication/azure/mixedreality/authentication/_utils.py
@@ -42,7 +42,11 @@ def retrieve_jwt_expiration_timestamp(jwt_value):
         raise ValueError("Invalid JWT structure. Expected a JWS Compact Serialization formatted value.")
 
     try:
-        padded_base64_payload = base64.b64decode(parts[1]).decode('utf-8')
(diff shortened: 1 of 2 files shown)


sdk/mixedreality/azure-mixedreality-authentication/azure/mixedreality/authentication/_utils.py
```diff
@@ -42,7 +42,7 @@ def retrieve_jwt_expiration_timestamp(jwt_value):
         raise ValueError("Invalid JWT structure. Expected a JWS Compact Serialization formatted value.")
 
     try:
-        padded_base64_payload = base64.b64decode(parts[1]).decode('utf-8')
+        padded_base64_payload = base64.b64decode(parts[1] + "===").decode('utf-8')
```
rikogeln, line 45: Could you add a comment what this does/ why it's needed? Point to https://tools.ietf.org/id/draft-jones-json-web-token-02.html#base64urlnotes  

I would be surprised/confused what this does

rikogeln (changes requested):

sdk/mixedreality/azure-mixedreality-authentication/azure/mixedreality/authentication/_utils.py
craigktreasure: Good call. Will do.

rikogeln (approved):

PR #17189, opened 2021-03-09
update api ref links

swathipil:


eng/ignore-links.txt
```diff
@@ -3,3 +3,7 @@ https://github.com/Azure/azure-digital-twins/blob/private-preview/Documentation/
 https://pypi.org/project/azure-digitaltwins-core
 https://docs.microsoft.com/azure/digital-twins/how-to-query-graph#query-limitations
 https://docs.microsoft.com/samples/azure-samples/azure-samples-python-management/{{package_doc_id}}
+https://azuresdkdocs.blob.core.windows.net/$web/python/azure-eventhub/5.3.1/azure.eventhub.html
```
weshaggard, line 6: In general it is a bad idea to add links to the ignore list as that implies that we have broken links, even if they are supposed to only be temporarily broken. Instead we should try to have links in our README files that aren't in this broken state.

swathipil: that makes sense, thanks for the feedback! for context, we were releasing 2 versions of EH. The first, version `5.4.0b1`, was on a feature branch and the second, version `5.3.1`, was on the `master` branch. We were trying to work around having our API ref documentation links on `master` point to `.../latest/azure.eventhub.html` since that would reference version `5.4.0b1` rather than `5.3.1`. In order to make the PR pass, we wanted to ignore it and remove these ignored links once they became valid. Since this would be a temporary fix (which we should try to avoid), would you know of a way to tackle this kind of situation in the future?

weshaggard: Please refer to some of the guidelines around links at https://github.com/Azure/azure-sdk/blob/master/docs/policies/README-TEMPLATE.md#link-guidelines

swathipil: will do in the future, thanks!

PR #17200, opened 2021-03-09
Update Artifact Upload to prevent failures on retry of Build Stage

scbedd:
We use `PublishPipelineArtifact` to pass along the artifacts to the test runs.  Unfortunately, if there is another failure, we _still_ upload them. This causes an issue if there is a failure in `build` stage. As now we _literally_ cannot get pasted it as any further attempts at the job will fail due to the publish pipeline failure. [Here's an example of this at play.](https://dev.azure.com/azure-sdk/public/_build/results?buildId=771657&view=results)

The best solution I can see here is to only upload after we know we're good to go.

EDIT: I noticed that this would affect builds that had failed due to `create-apireview` failures. I had a chat with @praveenkuttappan and also included updating the _location_ of `create-apireview` such that repeated failures could be successfully retried.


benbp (approved):

eng/pipelines/templates/steps/build-artifacts.yml
```diff
@@ -83,13 +90,42 @@ steps:
...
+      if ($docParam) {
+        # if we set it to false handle it here
+        $shouldPublishDocs = 	[System.Convert]::ToBoolean($docParam)
+      }
+      
+      $packagePublishes = ((Get-ChildItem '$(Build.ArtifactStagingDirectory)') | Measure-Object).Count
+      
+      if ($packagePublishes -eq 0) {
+          Write-Host "We are expecting to publish package artifacts, but no artifacts were found."
+          Write-Host (Get-ChildItem '$(Build.ArtifactStagingDirectory)')
```
weshaggard, line 106: Wouldn't this always output nothing if nothing was found?

eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -209,7 +206,7 @@ jobs:
         ToxEnvParallel: ${{ parameters.ToxEnvParallel }}
         InjectedPackages: $(InjectedPackages)
         BeforeTestSteps:
-          - task: DownloadPipelineArtifact@0
+          - task: DownloadBuildArtifacts@0
```
weshaggard, line 209: Why the switch from pipeline to build artifacts? @mitchdenny can correct me but I thought pipeline was the newer one which most folks should be using.

eng/pipelines/templates/steps/build-artifacts.yml
```diff
@@ -83,13 +90,42 @@ steps:
...
+        if ($docsPublishes -eq 0) {
+          Write-Host "We are expecting to publish documentation artifacts, but no documentation artifacts were found."
+          Write-Host (Get-ChildItem '$(Build.SourcesDirectory)/_docs')
+          exit 1
+        }
+      }
+    displayName: 'Verify Presence of Artifacts'
 
   - task: PublishBuildArtifacts@1
-    condition: and(succeededOrFailed(), ${{parameters.BuildDocs}})
```
weshaggard, line 91: Wouldn't you still want a condition here?

eng/pipelines/templates/steps/build-artifacts.yml
```diff
@@ -83,13 +90,42 @@ steps:
...
-    displayName: 'Publish Artifacts'
+    displayName: 'Publish Package Artifacts'
+    inputs:
+      artifactName: 'packages'
+      pathtoPublish: '$(Build.ArtifactStagingDirectory)'
+
+  - task: PublishBuildArtifacts@1
+    condition: and(succeeded(), ${{ parameters.BuildDocs }})
```
weshaggard, line 127: Do you want to skip this on all failures are only if there are no doc files?

eng/pipelines/templates/jobs/archetype-sdk-client.yml
scbedd: We were already publishing BuildArtifacts. I simply updated everything to align so we weren't getting warnings for differing artifact types.

eng/pipelines/templates/steps/build-artifacts.yml
scbedd: We only want to skip if we deliberately shut off docs. Otherwise, having no docs is a failure.

eng/pipelines/templates/steps/build-artifacts.yml
scbedd: Outside of weird powershell wonkiness, you're right.

eng/pipelines/templates/jobs/archetype-sdk-client.yml
scbedd: In a followup PR I'll update everything to PipelineArtifacts (publish side as well).

eng/pipelines/templates/steps/build-artifacts.yml
scbedd: No. In the previous step we verify that both have their contents as expected. If we skip docs, we won't publish docs below. Otherwise, we only want to publish after a successful build.

This breaks the test pipelines (as they won't be able to grab the artifact), but it's a much better experience than needing to entirely requeue the build if the `build` stage fails.

eng/pipelines/templates/jobs/archetype-sdk-client.yml
scbedd: K I've got a[ commit ready to go]()https://github.com/scbedd/azure-sdk-for-python/commit/f752a336d4f610fb8c300cb3b9fdecaf614e868b in that cleans that up + also removes the pending cleanup items.

eng/pipelines/templates/steps/build-artifacts.yml
scbedd: @weshaggard I would love to override the artifacts on retry, but that's not really a thing with azure DevOps unfortunately.

scbedd: I think your continueOnFailure is really the best option.

eng/pipelines/templates/jobs/archetype-sdk-client.yml
scbedd: I just merged that commit. Will do another test run over the weekend to confirm.

eng/pipelines/templates/steps/build-artifacts.yml
scbedd: Went with resolving the artifact name at runtime. If it's a failed build, we set the artifact name to something with the build number+iteration.

eng/common/pipelines/templates/steps/resolve-artifact-name.yml
```diff
@@ -0,0 +1,19 @@
...
+# will note REMAIN failing after the appropriate updates have been made to the source.
+# ArtifactNameVariable - The name of the DevOps Variable that will be consumed in the PublishPipelineArtifact steps
+# DefaultArtifactName  - The name of the artifact in the "successful" case.
+
+parameters:
+  ArtifactNameVariable: 'Publishing.Artifact.Name'
+  DefaultArtifactName: 'packaging'
+
+steps:
+  - pwsh: |
```
weshaggard, line 11: @mitchdenny and I were chatting about this and one option we came up with that would accomplish the same thing while avoiding the variable name dancing would be to just have to publish calls one for success and one for failure. Something like:

```
  - task: PublishPipelineArtifact@1
    condition: succeeded()
    displayName: 'Publish ${{ parameters.ArtifactName }} Artifacts'
    inputs:
      artifactName: ${{ parameters.ArtifactName }}
      pathtoPublish: ${{ parameters.ArtifactPath }}
  - task: PublishPipelineArtifact@1
    condition: failed()
    displayName: 'Publish failed ${{ parameters.ArtifactName }} Artifacts'
    inputs:
      artifactName: '${{ parameters.ArtifactName }}-FailedAttempt$(System.JobAttempt)'
      pathtoPublish: ${{ parameters.ArtifactPath }}
```

That would allow us to have a generic publish-artifact.yml template we could use in all cases.

scbedd: Done. Testing.

weshaggard (approved):
Changes look good. I leave it to you if you want to get your publish-artifact.yml into eng/common first or just commit it somewhere else in the python repo for now.

PR #17201, opened 2021-03-09
adding a keyvault cert

seankane-msft:


seankane-msft:
This keyvault certificate is not a secret (confirmed with @mccoyp), however it is flagged by CredScan as one. In the meantime we are working on a solution to redact this from recordings.

PR #17220, opened 2021-03-10
appropriately update the updatedocsciFn in Language-Settings. It woul…

scbedd:
…d only update an EXISTING package w/ the appropriate version target for Preview.


weshaggard (approved):
I'm don't have all the context but the change in isolation looks reasonable.

scbedd:
@weshaggard Essentially, I'm adding the logic to handle when we're not updating an existing package FROM preview TO GA. The logic as it would always create a "new" entry _without_ version targeting. AKA, only onboarding as "GA".

It'll install just fine, but it's not appropriate.

PR #17224, opened 2021-03-10
[Key Vault] Label unreleased certificates and secrets versions as beta

mccoyp:
This will (theoretically) resolve nightly build failures (see [example](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=773462&view=results)) caused by a lack of APIView approvals on unreleased packages. These checks shouldn't fail the pipeline if the package version checked into master is a beta version.

### DIFF
--- a/sdk/keyvault/azure-keyvault-certificates/azure/keyvault/certificates/_version.py
+++ b/sdk/keyvault/azure-keyvault-certificates/azure/keyvault/certificates/_version.py
@@ -3,4 +3,4 @@
 # Licensed under the MIT License.
 # ------------------------------------
 
-VERSION = "4.2.2b1"
+VERSION = "4.3.0b1"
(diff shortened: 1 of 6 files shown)


sdk/keyvault/azure-keyvault-certificates/azure/keyvault/certificates/_version.py
```diff
@@ -3,4 +3,4 @@
 # Licensed under the MIT License.
 # ------------------------------------
 
-VERSION = "4.2.2"
+VERSION = "4.2.2b1"
```
chlowell, line 6: These packages have new API, they should be betas of the next major version:
```suggestion
VERSION = "4.3.0b1"
```

chlowell (approved):

PR #17235, opened 2021-03-10
integrated sdk with generated client + models wrapping

mohamedshabanofficial:
@kristapratico


sdk/documenttranslation/azure-ai-documenttranslation/azure/ai/documenttranslation/_models.py
```diff
@@ -143,6 +206,23 @@ def __init__(
...
         self.total_characters_charged = kwargs.get('total_characters_charged', None)
 
+    @classmethod
+    def _from_generated(cls, batch_status_details):
+        return cls(
+            id = batch_status_details.id,
+            created_on = batch_status_details.created_date_time_utc,
+            last_updated_on = batch_status_details.last_action_date_time_utc,
+            status = batch_status_details.status,
+            error = DocumentTranslationError._from_generated(batch_status_details.error),
```
kristapratico, line 216: `error` is not a required parameter so the service might not always return it. I would add a conditional either here or in the helper method in the case that `batch_status_details.error` == None so we don't get an AttributeError when trying to wrap the model 
edit: same comment for DocumentStatusDetail.error

sdk/documenttranslation/azure-ai-documenttranslation/azure/ai/documenttranslation/_models.py
```diff
@@ -233,3 +336,16 @@ def __init__(
...
+        return cls(
+            format=file_format.format,
+            file_extensions=file_format.file_extensions,
+            content_types=file_format.content_types,
+            versions=file_format.versions
+        )
+
+    @classmethod
+    def _from_generated_list(cls, file_formats):
+        return list( [ FileFormat._from_generated(file_formats) for file_formats in file_formats ] )
```
kristapratico, line 351: Believe the generated code returns a model which we need to .value into to get the iterable

```suggestion
        return [FileFormat._from_generated(file_formats) for file_formats in file_formats.value ]
```

sdk/documenttranslation/azure-ai-documenttranslation/azure/ai/documenttranslation/_client.py
```diff
@@ -129,7 +129,7 @@ def callback(raw_response):
             deserialization_callback=callback,
             polling_method=LROBasePolling(
                 timeout=30,
-                lro_algorithms=StatusPollingGETRequest(),
+                lro_algorithms=TranslationPolling(),
```
kristapratico, line 131: LROBasePolling is expecting a list here, so we should pass `lro_algorithms=[TranslationPolling()]`

sdk/documenttranslation/azure-ai-documenttranslation/azure/ai/documenttranslation/_polling.py
```diff
@@ -0,0 +1,64 @@
...
+from azure.core.polling.base_polling import (
+    LongRunningOperation,
+    _is_empty,
+    _as_json,
+    BadResponse,
+    OperationFailed
+)
+
+
+class TranslationPolling(LongRunningOperation):
```
kristapratico, line 12: This impl looks good to me. Since we are inheriting from the LongRunningOperation abstract class, we just need to define a `get_final_get_url` method. Since the operation doesn't need to do a final get, we can just return None like this:

```
    def get_final_get_url(self, pipeline_response):
        return None
```

mohamedshabanofficial: yeah sorry, forgot it's an abstract method!

sdk/documenttranslation/azure-ai-documenttranslation/azure/ai/documenttranslation/_client.py
```diff
@@ -106,7 +143,21 @@ def list_submitted_jobs(self, **kwargs):
         :keyword int skip:
         :rtype: ~azure.core.polling.ItemPaged[JobStatusDetail]
         """
-        return self._client.document_translation.get_operations(**kwargs)
+
+        skip = kwargs.pop('skip', None)
+        top = kwargs.pop('top', None)
```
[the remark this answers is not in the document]
mohamedshabanofficial: by "this", do you mean list_submitted_jobs?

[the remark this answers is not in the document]
kristapratico: I meant the `top` keyword argument should be renamed `results_per_page`

sdk/documenttranslation/azure-ai-documenttranslation/azure/ai/documenttranslation/_models.py
```diff
@@ -32,6 +39,25 @@ def __init__(
         self.format_version = kwargs.get("format_version", None)
         self.storage_source = kwargs.get("storage_source", None)
 
+    def to_generated(self):
```
kristapratico, line 42: I know this is misleading because the linter checks will tell you _not_ to access a method prefixed with underscore, but we don't want our users to access these model methods. So we can keep them "private" and leave the underscore prefix, but add a linter ignore comment to any of the lines that call them: `# pylint: disable=protected-access`

sdk/documenttranslation/azure-ai-documenttranslation/azure/ai/documenttranslation/_client.py
```diff
@@ -120,7 +171,22 @@ def list_documents_statuses(self, job_id, **kwargs):
         :rtype: ~azure.core.paging.ItemPaged[DocumentStatusDetail]
         """
 
-        return self._client.document_translation.get_operation_documents_status(job_id, **kwargs)
+        skip = kwargs.pop('skip', None)
+        top = kwargs.pop('top', None)
```
kristapratico, line 175: we can rename `top` to `results_per_page` here as well

kristapratico (commented):
Looking good :) After resolving the merge conflict, let's address any other linting checks that come up in the CI pipeline and then merge!

kristapratico:
There we go, now the CI pipeline is running on this PR. Linter check errors can be found here: https://dev.azure.com/azure-sdk/public/_build/results?buildId=786387&view=logs&j=b70e5e73-bbb6-5567-0939-8415943fadb9&t=56d4e2e6-c43b-527c-bad7-234ebe7429dd

sdk/documenttranslation/azure-ai-documenttranslation/azure/ai/documenttranslation/_client.py
```diff
@@ -95,7 +115,28 @@ def wait_until_done(self, job_id, **kwargs):
         :return: JobStatusDetail
         :rtype: JobStatusDetail
         """
-        pass  # pylint: disable=unnecessary-pass
+
+        pipeline_response = self.get_job_status(
```
kristapratico, line 119: just realized, now that we've wrapped the `get_job_status` method to return the exposed model it won't work here. We should call the equivalent generated code method instead:
```suggestion
        pipeline_response = self._client.document_translation.get_operation_status(
```

sdk/documenttranslation/azure-ai-documenttranslation/azure/ai/documenttranslation/_models.py
```diff
@@ -60,8 +90,22 @@ def __init__(
         self.glossaries = kwargs.get("glossaries", None)
         self.storage_source = kwargs.get("storage_source", None)
 
+    def _to_generated(self):
+        return _TargetInput(
+            target_url=self.target_url,
+            category=self.category_id,
+            language=self.language,
+            storage_source=self.storage_source,
+            glossaries=TranslationGlossary._to_generated_list(self.glossaries)  # pylint: disable=protected-access
```
kristapratico, line 99: For the case when the user does not pass glossaries

```suggestion
            glossaries=TranslationGlossary._to_generated_list(self.glossaries) if self.glossaries else None  # pylint: disable=protected-access
```

kristapratico (approved):
🚀

PR #17257, opened 2021-03-10
Reorder Python versions in matrix

mccoyp:
Context: azure-keyvault-keys tests are consistently failing on Windows with Python 3.7 (example [here](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=776772&view=results)). Live tests with 3.7 on Windows pass locally, so this is an attempt to see if the failures are particular to how Windows and/or 3.7 tests are run in the pipeline.


mccoyp:
This would make more sense to do in a separate branch, so I'll point this change away from master

PR #17270, opened 2021-03-11
[Network] increase timeout limite for docs generate of pipeline

msyyc:


00Kai0 (approved):

eng/pipelines/generate-all-docs.yml
```diff
@@ -9,7 +9,7 @@ jobs:
     variables:
       skipComponentGovernanceDetection: true
     
-    timeoutInMinutes: 120
+    timeoutInMinutes: 1200
```
weshaggard, line 12: Why is this timeout so large? Do we really need such a large number?

jsntcy: @msyyc is trying to test if increasing the timeout can resolve the timeout issue when generating Network SDK docs.

Issue #16890, opened 2021-02-23
[Datalake]Optimize memory usage for upload_data for large files


xiafu-msft:
**Is your feature request related to a problem? Please describe.**
currently if each chunk is 4GM it seems the whole chunk will be keep in memory first.....

**Describe the solution you'd like**
mimic what blob does for large file upload

tasherif-msft:
Hi Emily, could you elaborate? I tried testing our current workflow for both blob uploads (using `upload_blob`) and file uploads in datalake (`upload_data`). The `tracemalloc` shows that memory usage is essentially the same (if you set the datalake `chunk_size = 4*1024*1024` to mimic blob chunk size when uploading)

Also by 4GM do you mean 4MB?
Could you give a code snippet of the use-case you're hitting? :)

PR #17272, opened 2021-03-11
Optimized memory usage for datalake file uploads large chunks

tasherif-msft:
This resolves https://github.com/Azure/azure-sdk-for-python/issues/16890

### DIFF
--- a/sdk/storage/azure-storage-blob/azure/storage/blob/_shared/uploads_async.py
+++ b/sdk/storage/azure-storage-blob/azure/storage/blob/_shared/uploads_async.py
@@ -124,7 +124,8 @@ async def upload_substream_blocks(
         range_ids = []
         for block in uploader.get_substream_blocks():
             range_ids.append(await uploader.process_substream_block(block))
-    return sorted(range_ids)
+    if any(range_ids):
+        return sorted(range_ids)
 
 
 class _ChunkUploader(object):  # pylint: disable=too-many-instance-attributes
--- a/sdk/storage/azure-storage-blob/azure/storage/blob/_shared/uploads_async.py
+++ b/sdk/storage/azure-storage-blob/azure/storage/blob/_shared/uploads_async.py
@@ -224,16 +224,16 @@ class _ChunkUploader(object):  # pylint: disable=too-many-instance-attributes
         for i in range(blocks):
             index = i * self.chunk_size
             length = last_block_size if i == blocks - 1 else self.chunk_size
-            yield ('BlockId{}'.format("%05d" % i), SubStream(self.stream, index, length, lock))
+            yield index, SubStream(self.stream, index, length, lock)
 
     async def process_substream_block(self, block_data):
         return await self._upload_substream_block_with_progress(block_data[0], block_data[1])
 
-    async def _upload_substream_block(self, block_id, block_stream):
+    async def _upload_substream_block(self, index, block_stream):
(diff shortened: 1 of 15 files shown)


sdk/storage/azure-storage-blob/azure/storage/blob/_shared/uploads_async.py
```diff
@@ -224,16 +226,16 @@ def get_substream_blocks(self):
         for i in range(blocks):
             index = i * self.chunk_size
             length = last_block_size if i == blocks - 1 else self.chunk_size
-            yield ('BlockId{}'.format("%05d" % i), SubStream(self.stream, index, length, lock))
+            yield index, SubStream(self.stream, index, length, lock)
```
xiafu-msft, line 229: if I remember properly blockid should have the same length, if we change it to index I'm worried something bad will happen.

tasherif-msft: I did not use the index, I still converted it to a proper block id here https://github.com/Azure/azure-sdk-for-python/pull/17272/files#diff-fd90426e10c45bf19310c14aef222c23145b307e761fef45eac642e704562103R269
I just moved it so this method can be used by datalake as well as blob

xiafu-msft: found the reference here https://docs.microsoft.com/en-us/rest/api/storageservices/put-block#uri-parameters

sdk/storage/azure-storage-blob/azure/storage/blob/_shared/uploads_async.py
```diff
@@ -256,14 +258,15 @@ async def _upload_chunk(self, chunk_offset, chunk_data):
...
+            body=chunk_data,
             data_stream_total=self.total_size,
             upload_stream_current=self.progress_total,
             **self.request_options)
         return index, block_id
 
-    async def _upload_substream_block(self, block_id, block_stream):
+    async def _upload_substream_block(self, index, block_stream):
         try:
+            block_id = 'BlockId{}'.format("%05d" % (index/self.chunk_size))
```
xiafu-msft, line 269: oh you formatted the blockid here

sdk/storage/azure-storage-blob/azure/storage/blob/_shared/uploads_async.py
xiafu-msft: nvm! I found you formated the blockid in another place!

sdk/storage/azure-storage-blob/azure/storage/blob/_shared/uploads_async.py
xiafu-msft: what will this generate?

xiafu-msft: hahaha it's index! nvm

tasherif-msft: why? index is i * self.chunk_size https://github.com/Azure/azure-sdk-for-python/pull/17272/files#diff-fd90426e10c45bf19310c14aef222c23145b307e761fef45eac642e704562103L224-L225

So (index/self.chunk_size) would be 0, 1, 2 etc

xiafu-msft (approved):

PR #17277, opened 2021-03-11
Increase gen doc task timeout.

mitchdenny:
This PR increases the build job to 3 hours, and the generate docs job to 2 hours. This is just an experiment to get a measure of how long the generate docs step is now taking for the network management APIs.

### DIFF
--- a/eng/pipelines/templates/jobs/archetype-sdk-client.yml
+++ b/eng/pipelines/templates/jobs/archetype-sdk-client.yml
@@ -86,7 +86,7 @@ parameters:
 
 jobs:
   - job: 'Build'
-    timeoutInMinutes: 90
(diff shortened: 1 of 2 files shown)


scbedd:
@mitchdenny it's still not enough 👍

mitchdenny:
> @mitchdenny it's still not enough 👍

Yep ... that is why I didn't just race in with that as the solution :)

weshaggard:
Is https://github.com/Azure/azure-sdk-for-python/pull/17270 related to this change as well?

jsntcy:
> Is #17270 related to this change as well?

Yes, try to resolve the timeout issue when generating Network SDK docs: #17179

Issue #17247, opened 2021-03-10
Cache Sphinx Results to Enable Build of Large Packages


scbedd:
[Link to example build](https://dev.azure.com/azure-sdk/public/_build/results?buildId=772952&view=logs&j=3dc8fd7e-4368-5a92-293e-d53cefc8c4b3&t=651a14cc-88df-515d-1311-e0e8b23b578a)

scbedd:
My original thought was that [this PR](https://github.com/Azure/azure-sdk-for-python/pull/17200)

introduced an issue that lengthens the generation time. Now I'm thinking that there was a PR merged that just crossed a threshold.

Evidence on [this PR](https://github.com/Azure/azure-sdk-for-python/pull/17286) can be found on [this manually triggered internal build of azure-mgmt-network](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=778056&view=logs&j=3dc8fd7e-4368-5a92-293e-d53cefc8c4b3&t=651a14cc-88df-515d-1311-e0e8b23b578a)

PR #17286, opened 2021-03-11
Revert Management Document Changes

scbedd:
To unblock release of `azure-mgmt-network`

While we dig into #17247 

While this will unblock release I have a very strong suspicion that the documentation that is being generated is omitting a lot of code. 

Need to compare the two generated sphinx sources prior to running `sphinx-build` and see what documents are being omitted from azure-mgmt-network.

### DIFF
--- a/doc/sphinx/package_service_mapping.json
+++ b/doc/sphinx/package_service_mapping.json
@@ -1,0 +1,1084 @@
+{
+  "azure-applicationinsights": {
+		"service_name": "Application Insights",
+		"category": "Client",
+    "namespaces": [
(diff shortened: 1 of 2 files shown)


jsntcy:
@scbedd, do we need apply this revert to release/v3 branch too?

mitchdenny (approved):

scbedd:
@jsntcy I'm doing this to unblock release, give me a day or so to try to figure out WHY it's such a difference. 

If you have a release to unblock from the feature branch in the next couple days, please merge it, yeah.

PR #17299, opened 2021-03-11
[formrecognizer] Test fixes

catalinaperalta:
Get the poller result and assert instead of just calling poller.wait()


catalinaperalta:
NOTE: did not add recordings for currently skipped tests

kristapratico (approved):
thank you for doing this!

PR #17318, opened 2021-03-12
fix running samples script

kristapratico:
1) unblocks running samples with python 2.7
2) continues to run samples even if one fails

### DIFF
--- a/scripts/devops_tasks/test_run_samples.py
+++ b/scripts/devops_tasks/test_run_samples.py
@@ -11,7 +11,11 @@ import sys
 import os
 import logging
 from fnmatch import fnmatch
-from subprocess import check_call, CalledProcessError, TimeoutExpired
+from subprocess import check_call, CalledProcessError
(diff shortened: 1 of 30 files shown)


rakshith91 (approved):

scripts/devops_tasks/test_run_samples.py
```diff
@@ -191,7 +195,7 @@ def run_samples(targeted_package):
         with open(samples_dir_path + "/sample_dev_requirements.txt") as sample_dev_reqs:
             for dep in sample_dev_reqs.readlines():
                 check_call([sys.executable, '-m', 'pip', 'install', dep])
```
swathipil, line 197: one of the dependencies for EH is async and can't be installed in python 2.7. would you be able to change it to the following?
```suggestion
                try:
                    check_call([sys.executable, '-m', 'pip', 'install', dep])
                except:
                    pass
```

PR #17322, opened 2021-03-12
Optimize Sphinx Build Settings

scbedd:
The more I look at this the more I think the best way to do this would be to pull down the latest existing _generated output_ from our blob storage (where we publish github.io docs) and use that as the basis for our _current_ build.

That would swap our sphinx builds to `incremental` which may dig us out of the quagmire that is generating azure-mgmt-network fresh.


scbedd:
Closing this to get it off the PR list.

PR #17326, opened 2021-03-13
[Key Vault] Add key creation helpers in tests to avoid throttling

mccoyp:
Because software key creation is rate limited at 10 keys per 10 seconds, `azure-keyvault-keys` tests are beginning to get consistently throttled during live test pipelines. This adds helpers to tests that create the most keys, similar to .NET's [DelayCreateKeyInterceptor](https://github.com/Azure/azure-sdk-for-net/blob/master/sdk/keyvault/Azure.Security.KeyVault.Keys/tests/DelayCreateKeyInterceptor.cs), that wait for 2 seconds before creating a key during live tests.


chlowell (commented):
This change warrants a live test run but I wouldn't expect new recordings. It would be a red flag if adding some sleep caused tests to fail with the current recordings, and I would expect the new recordings to make an uninteresting diff, i.e. no request added or removed (except in polling), no structural changes. It appears you did a lot more polling this time around, that's kind of interesting.

mccoyp:
The previous recordings fail because `match_body` defaults to true and there are extra new keyword arguments being added to most requests since keyword arguments are being passed, though they're usually using the default values anyway. The extra polling is interesting -- I think it might come from the fact that I was using `python-xdist` when making the recordings (I'll try recording in serial if we still create new ones)

chlowell:
Ah, I see. That makes me a little nervous; have we lost coverage of e.g. `client.create_rsa_key("name")`?

mccoyp:
That's a good point -- we still have coverage of the default creation method for RSA keys in `test_crypto_client.py`, but EC keys have a curve parameter defaulted by the helper. I'll tweak this a bit to make the helper more bare-bones

chlowell (approved):

PR #17340, opened 2021-03-15
add custom policy

xiangyan99:


xiangyan99:
#16519

johanste:
Let's align with the other languages on this. Both .NET/C# and Java uses the perCall or perRetry terminology rather than first and last. I have not checked the javascript implementation.

sdk/core/azure-core/azure/core/_pipeline_client_async.py
```diff
@@ -101,16 +112,34 @@ def _build_pipeline(self, config, **kwargs): # pylint: disable=no-self-use
...
             policies = [
                 RequestIdPolicy(**kwargs),
                 config.headers_policy,
                 config.user_agent_policy,
                 config.proxy_policy,
-                ContentDecodePolicy(**kwargs),
+                ContentDecodePolicy(**kwargs)
+            ]
+            if per_call_policies:
+                if isinstance(per_call_policies, (AsyncHTTPPolicy, SansIOHTTPPolicy)):
```
johanste, line 125: Isn't per_call_policies and Iterable?

sdk/core/azure-core/azure/core/_pipeline_client_async.py
```diff
@@ -101,16 +112,34 @@ def _build_pipeline(self, config, **kwargs): # pylint: disable=no-self-use
         policies = kwargs.get('policies')
 
         if policies is None:  # [] is a valid policy list
+            per_call_policies = kwargs.get('per_call_policies')
```
johanste, line 115: You can default these to `[]`.

sdk/core/azure-core/azure/core/_pipeline_client_async.py
johanste: Never mind, missed the single-or-list. I would have switched the test around and checked if it is an Iterable...

sdk/core/azure-core/tests/async_tests/test_pipeline_async.py
```diff
@@ -219,4 +219,42 @@ def send(*args):
...
+
+    boo_policy = BooPolicy()
+    foo_policy = FooPolicy()
+    client = AsyncPipelineClient(base_url="test", per_call_policies=boo_policy)
+    policies = client._pipeline._impl_policies
+    assert boo_policy in policies
+
+    client = AsyncPipelineClient(base_url="test", per_call_policies=[boo_policy])
+    policies = client._pipeline._impl_policies
+    assert boo_policy in policies
```
annatisch, line 242: We never assert the location/ordering of the custom policies in the pipeline?

xiangyan99: Yes we can. But hard-coding the location will make our tests fragile (adding policy into our default pipeline may break the test). I will re-think about it. :)

annatisch: We just need to test that the index is < or > than the index of the Retry policy right? The exact index shouldn't matter.

xiangyan99: Yes. This is exactly I am trying to do. :)

xiangyan99: Added.

annatisch (approved):

sdk/core/azure-core/azure/core/_pipeline_client.py
```diff
@@ -102,20 +113,34 @@ def _build_pipeline(self, config, **kwargs): # pylint: disable=no-self-use
...
-                config.authentication_policy,
-                config.custom_hook_policy,
-                config.logging_policy,
-                DistributedTracingPolicy(**kwargs),
-                config.http_logging_policy or HttpLoggingPolicy(**kwargs)
+                ContentDecodePolicy(**kwargs)
             ]
+            if isinstance(per_call_policies, Iterable):
+                for policy in per_call_policies:
+                    policies.append(policy)
```
jiasli, line 127: No need to explicitly iterate over `per_call_policies`. [`list.extend(iterable)`](https://docs.python.org/3/tutorial/datastructures.html#more-on-lists) does the same thing.

sdk/core/azure-core/azure/core/_pipeline_client.py
```diff
@@ -102,20 +113,34 @@ def _build_pipeline(self, config, **kwargs): # pylint: disable=no-self-use
...
+            else:
+                policies.append(per_call_policies)
+
+            policies = policies + [config.redirect_policy,
+                               config.retry_policy,
+                               config.authentication_policy,
+                               config.custom_hook_policy]
```
jiasli, line 134: This causes a new `list` object to be created and assigned to `policies`. [`list.extend(iterable)`](https://docs.python.org/3/tutorial/datastructures.html#more-on-lists) is a better choice as it modifies `policies` directly.

jiasli:
Fix #16519: Support adding custom policy with `per_call_policies`, `per_call_policies`

PR #17386, opened 2021-03-17
Create an example file showing how to stream data from and to a blob.

ali-elgabri:
This file contains an example of how to stream data from and to a blob. There are no docs and there are no examples on how to do this, and it can be quite confusing; it's a consequential thing to know, so thought I would help get that started.

If any maintainer wishes to make changes, please go ahead and make them.


tasherif-msft:
Hey @ali-elgabri I have created a PR here https://github.com/Azure/azure-sdk-for-python/pull/17435
Let me know what you think.

xiangyan99:
Given it is added in #17435, close the PR.

PR #17396, opened 2021-03-17
Storage devtool migration

xiafu-msft:


sdk/storage/azure-storage-file-datalake/tests/conftest.py
```diff
@@ -0,0 +1,33 @@
...
+# FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+# IN THE SOFTWARE.
+#
+# --------------------------------------------------------------------------
+import sys
+
+# fixture needs to be visible from conftest
+
+# Ignore async tests for Python < 3.5
+collect_ignore_glob = []
```
tasherif-msft, line 31: Out of curiosity, how did we specify this prior to this migration?

sdk/storage/tests.yml
```diff
@@ -18,10 +18,9 @@ stages:
         AllocateResourceGroup: 'false'
         BuildTargetingString: ${{ service }}
         ${{ if eq(service, 'azure-storage-file-datalake') }}:
-          MaxParallel: 1
-        ${{ if ne(service, 'azure-storage-file-datalake') }}:
+          DeployArmTemplate: true
```
tasherif-msft, line 21: Where are we specifying where to fetch the ARM template from?

sdk/storage/azure-storage-file-datalake/tests/testcase.py
```diff
@@ -5,6 +5,8 @@
 # license information.
 # --------------------------------------------------------------------------
 from __future__ import division
+
+import functools
```
tasherif-msft, line 9: There's a bunch of imports that we can remove now that we're migrating

sdk/storage/azure-storage-file-datalake/tests/testcase.py
```diff
@@ -71,37 +64,31 @@ def get_token(self, *args):
...
-        # example of qualified test name:
-        # test_mgmt_network.test_public_ip_addresses
-        _, filename = os.path.split(inspect.getsourcefile(type(self)))
-        name, _ = os.path.splitext(filename)
-        self.qualified_test_name = '{0}.{1}'.format(
-            name,
-            self._testMethodName,
-        )
 
+class StorageTestCase(AzureTestCase):
```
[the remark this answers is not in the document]
xiafu-msft: StorageTestCase and AzureMgmtTestcase are both inheriting AzureTestcase. We don't need anything specific from either StorageTestCase/AzureMgmtTestcase, so there's no need to use them.
For example people who are using AzureMgmtTestcase is mainly because they need to create_mgmt_client to do some mgmt plane operation. For StorageTestCase we were using that to create storage account, while for datalake the storage account is created through ARM Template deployment so we don't need StorageTestCase either.
I know blob is using AzureMgmtTestcase probably when it was added we were using something?

sdk/storage/tests.yml
xiafu-msft: this script will look for test-resources.json and deploy
https://github.com/Azure/azure-sdk-for-python/blob/f3760fc8d7ea1b46b0def0628579d36abe75976f/eng/common/TestResources/New-TestResources.ps1

sdk/storage/azure-storage-file-datalake/tests/conftest.py
xiafu-msft: good question we specified here https://github.com/Azure/azure-sdk-for-python/blob/638969c03e19894ad23bbfc545783d71e777f540/sdk/storage/azure-storage-file-datalake/conftest.py
let me see how I can modify this

sdk/storage/azure-storage-file-datalake/tests/testcase.py
```diff
@@ -172,73 +146,22 @@ def _set_test_proxy(service, settings):
                 settings.PROXY_PASSWORD,
             )
 
-    def _get_shared_key_credential(self):
+    def _get_shared_key_credential(self, account_name, account_key):
```
xiafu-msft, line 149: it's a type of credential we can pass in DataLakeServiceClient etc. 
It seems currently we are not using this credential, an we passed account key as credential.
While it would be better if we can use _shared_key_credential in dict format especially for customized url. Since with a pure account key we cannot parse out the account name for user.

sdk/storage/azure-storage-file-datalake/tests/testcase.py
```diff
@@ -172,73 +146,22 @@ def _set_test_proxy(service, settings):
...
         return {
-            "account_name": self.settings.STORAGE_DATA_LAKE_ACCOUNT_NAME,
-            "account_key": self.settings.STORAGE_DATA_LAKE_ACCOUNT_KEY
+            "account_name": account_name,
```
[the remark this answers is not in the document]
xiafu-msft: it's a type of credential we can pass in DataLakeServiceClient etc.
It seems currently we are not using this _get_shared_key_credential, an we passed account key as credential.
While it would be better if we can use _get_shared_key_credential so the credential is in dict format, this helps especially for customized url. Since with a pure account key we cannot parse out the account name for user.

sdk/storage/azure-storage-file-datalake/tests/testcase.py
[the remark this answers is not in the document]
tasherif-msft: Ah ok that explains it - so blobs we have https://github.com/Azure/azure-sdk-for-python/blob/77feaf14471eba6642f5c7ae2f3f06981ff361d7/sdk/storage/azure-storage-blob/tests/_shared/testcase.py#L382 when we're creating the account, whereas in Datalake we're directly passing account name and key using the ARM template to deploy these resources (storage accounts)?

sdk/storage/azure-storage-file-datalake/tests/testcase.py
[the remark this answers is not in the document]
tasherif-msft: that makes sense!

tasherif-msft (approved):

PR #17406, opened 2021-03-17
[formrecognizer] Update readme to include preview3 features

### LINKED ISSUE #17142: [formrecognizer] update readme with preview.3 features
Need links to new samples and update of the example code and API mapping version table. We also need to re-assess which prebuilt examples we will have in the README since that section is getting a bit lengthy.

catalinaperalta:
Fixes #17142


kristapratico:
This got closed because I merged the feature branch, sorry! we can re-open against master when you get back.

PR #17422, opened 2021-03-18
running black code formatter on appconfig

seankane-msft:
Runnning `black` code formatter

### DIFF
--- a/sdk/appconfiguration/azure-appconfiguration/azure/appconfiguration/_sync_token.py
+++ b/sdk/appconfiguration/azure-appconfiguration/azure/appconfiguration/_sync_token.py
@@ -26,9 +26,10 @@
 from azure.core.pipeline import PipelineRequest, PipelineResponse
 from azure.core.pipeline.policies import SansIOHTTPPolicy
 
+
 class SyncToken(object):
-    """The sync token structure
-    """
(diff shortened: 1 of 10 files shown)


sdk/appconfiguration/azure-appconfiguration/azure/appconfiguration/_sync_token.py
```diff
@@ -40,27 +41,31 @@ def __str__(self):
...
     """A simple policy that enable the given callback
     with the response.
 
     :keyword callback raw_response_hook: Callback function. Will be invoked on response.
     """
-    def __init__(self, **kwargs): # pylint: disable=unused-argument,super-init-not-called
+
+    def __init__(
+        self, **kwargs
+    ):  # pylint: disable=unused-argument,super-init-not-called
```
mccoyp, line 64: This one's interesting -- I don't know if pylint would raise an error if the comment makes the line too long, so I'm not sure this is necessary. It looks kind of awkward but it's also not a biggie

seankane-msft: One of the disables was unnecessary, removing it makes it look cleaner.

mccoyp (approved):

PR #17424, opened 2021-03-18
Make NetworkTraceLoggingPolicy show the auth token in plain text

xiangyan99:


sdk/core/azure-core/azure/core/pipeline/policies/_universal.py
```diff
@@ -280,8 +280,6 @@ def on_request(self, request):
                 _LOGGER.debug("Request method: %r", http_request.method)
                 _LOGGER.debug("Request headers:")
                 for header, value in http_request.headers.items():
-                    if header.lower() == 'authorization':
-                        value = '*****'
```
annatisch, line 284: This seems like a pretty fundamental change... or this is okay because this is only for debugging?

annatisch: Resolved as per email discussion :)

annatisch (approved):

Issue #16933, opened 2021-02-25
Error while Forwarding messages from Subscription to a Queue on ServiceBus sdk


eibrunorodrigues:
- **Package Name**: azure-servicebus
- **Package Version**: 7.0.1
- **Operating System**: Ubuntu
- **Python Version**: 3.8

**Describe the bug**
Having some issues creating subscriptions on a topic with the flag ForwardTo pointing to a queue.

**To Reproduce**
```
from azure.servicebus import ServiceBusClient, ServiceBusReceivedMessage
from azure.servicebus.management import ServiceBusAdministrationClient
from azure.core.exceptions import ResourceExistsError, ResourceNotFoundError
import os


servicebus_client = ServiceBusAdministrationClient.from_connection_string(
    conn_str=os.getenv("CONNECTION_STRING"))

queue_name = "client_example"
topic_name = "topic/client_example.main"

try:
    servicebus_client.create_queue("client_example")
except ResourceExistsError:
    pass

try:
    servicebus_client.create_topic("topic/client_example.main")
except ResourceExistsError:
    pass

try:
    servicebus_client.create_subscription(
        topic_name=topic_name,
        subscription_name=queue_name,
        forward_to=queue_name
    )
except ResourceExistsError:
    pass
```

**Expected behavior**
I'm creating a subscription that forwards all messages to a queue, this code works on C# sdk. 

**Screenshots**
![image](https://user-images.githubusercontent.com/39570923/109157679-1dbe5000-7751-11eb-9b67-6887f2ff6680.png)

**Additional context**
Error is thrown while attempting to forward-to messages to a queue:
`Exception has occurred: HttpResponseError       (note: full exception trace is shown but execution is paused at: _run_module_as_main)
BR0012The specified value is in an invalid format.
Parameter name: ForwardTo. TrackingId:614b32a2-b307-4e2e-9688-7b797f0cc32a_G19, SystemTracker:odin-hml:Topic:topic/client_example.main, Timestamp:2021-02-25T13:01:55`

eibrunorodrigues:
I was able to use ForwardTo using the sb protocol: sb://MYSB.servicebus.windows.net/{QUEUE_NAME}. 
So the question is, on the C# SDK we don't need to do that, just the destination for the messages. 

Should this be a requirement on Python SDK?

xiangyan99:
Thanks for the feedback, we’ll investigate asap.

yunhaoling:
[...]

PR #17430, opened 2021-03-18
[SB] accept only destination for forward_to 

swathipil:
addressing: #16933

### DIFF
--- a/sdk/servicebus/azure-servicebus/CHANGELOG.md
+++ b/sdk/servicebus/azure-servicebus/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 **New Features**
 
-* Updated `forward_to` parameter in `create_queue` and `create_subscription` methods on `ServiceBusAdministrationClient` to accept entities as well, rather than only full paths. In the case that an entity is passed in, it is assumed that the entity exists within the namespace specified by the `fully_qualified_namespace` property of the `ServiceBusAdministrationClient`.
+* Updated `forward_to` and `forward_dead_lettered_messages_to` parameters in `create_queue`, `update_queue`, `create_subscription`, and `update_subscription` methods on sync and async `ServiceBusAdministrationClient` to accept entities as well, rather than only full paths. In the case that an entity is passed in, it is assumed that the entity exists within the same namespace used for constructing the `ServiceBusAdministrationClient`.
 
 ## 7.1.0 (2021-03-09)
 
--- a/sdk/servicebus/azure-servicebus/CHANGELOG.md
+++ b/sdk/servicebus/azure-servicebus/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## 7.1.1 (Unreleased)
 
+**New Features**
+
+* Updated `forward_to` parameter in `create_queue` and `create_subscription` methods on `ServiceBusAdministrationClient` to accept entities as well, rather than only full paths. In the case that an entity is passed in, it is assumed that the entity exists within the namespace specified by the `fully_qualified_namespace` property of the `ServiceBusAdministrationClient`.
 
 ## 7.1.0 (2021-03-09)
 
--- a/sdk/servicebus/azure-servicebus/azure/servicebus/management/_management_client.py
+++ b/sdk/servicebus/azure-servicebus/azure/servicebus/management/_management_client.py
@@ -47,7 +47,7 @@ from ._utils import (
     serialize_rule_key_values,
     extract_rule_data_template,
     create_properties_from_dict_if_needed,
-    _prepend_fqn_to_forward_to_if_needed,
+    _normalize_forward_to_address,
     _validate_entity_name_type,
     _validate_topic_and_subscription_types,
     _validate_topic_subscription_and_rule_types,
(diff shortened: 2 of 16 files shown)


sdk/servicebus/azure-servicebus/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## 7.1.1 (Unreleased)
 
+**New Features**
+
+* Updated `forward_to` parameter in `create_queue` and `create_subscription` methods on `ServiceBusAdministrationClient` to accept entities as well, rather than only full paths. In the case that an entity is passed in, it is assumed that the entity exists within the namespace specified by the `fully_qualified_namespace` property of the `ServiceBusAdministrationClient`.
```
yunhaoling, line 7: nit: what you described is technically correct, but the namespace is specified when creating the `ServiceBusAdministrationClient`, the property is populated during the creation process. so what I'm thinking of is like the following:

also I see you updated `forward_dead_lettered_messages_to` but it's not mentioned here
```suggestion
* Updated `forward_to` and `forward_dead_lettered_messages_to` parameters in `create_queue` and `create_subscription` methods on `ServiceBusAdministrationClient` to accept entities as well, rather than only full paths. In the case that an entity is passed in, it is assumed that the entity exists within the same namespace used for constructing the `ServiceBusAdministrationClient`.
```

sdk/servicebus/azure-servicebus/azure/servicebus/management/_utils.py
```diff
@@ -320,6 +330,15 @@ def _validate_topic_subscription_and_rule_types(
             )
         )
 
+
+def _normalize_forward_to_address(forward_to, fully_qualified_namespace):
+    # type: (str, str) -> str
+    if forward_to:  # for python 2.7, cannot urlparse None
+        parsed = urlparse.urlparse(forward_to)
+        forward_to = "sb://" + fully_qualified_namespace + "/" + forward_to if not parsed.netloc else forward_to
+    return forward_to
```
yunhaoling, line 339: I would probably rename the method to be more general and tweak the logic a bit so that checking None is not for python 2.7 only.
```suggestion
def _normalize_entity_path_to_full_path_if_needed(entity_path, fully_qualified_namespace):
    # type: (str, str) -> str
    if not entity_path:
        return entity_path
    parsed = urlparse.urlparse(entity_path)
    forward_to = "sb://" + fully_qualified_namespace + "/" + entity_pathif not parsed.netloc else entity_path
    return entity_path
```

sdk/servicebus/azure-servicebus/azure/servicebus/management/_management_client.py
```diff
@@ -335,6 +336,13 @@ def create_queue(self, queue_name, **kwargs):
 
         :rtype: ~azure.servicebus.management.QueueProperties
         """
+        forward_to = _normalize_forward_to_address(
+            kwargs.pop("forward_to", None), self.fully_qualified_namespace
+        )
+        forward_dead_lettered_messages_to = _normalize_forward_to_address(
+            kwargs.pop("forward_dead_lettered_messages_to", None),
+            self.fully_qualified_namespace,
+        )
```
yunhaoling, line 345: could you help check whether `update_*` methods should also apply the change?

yunhaoling (commented):
left some comments, in general I'm good with it!

(I noticed that we're missing async changes and tests, but that should be easy to add)

sdk/servicebus/azure-servicebus/azure/servicebus/management/_management_client.py
swathipil: looks like JS does it: https://github.com/Azure/azure-sdk-for-js/blob/c81faf961d6eedf409470407e0f3f6b4b1621ea2/sdk/servicebus/service-bus/test/public/atomManagement.spec.ts#L2211. 
Java checks before setting forward to in all CRUD operations: https://github.com/Azure/azure-sdk-for-java/blob/8be0607215eea08527382d3f665f02b80af2572a/sdk/servicebus/microsoft-azure-servicebus/src/main/java/com/microsoft/azure/servicebus/management/SubscriptionDescription.java#L289
.NET does the check when normalizing the description, which can happen after create/update: https://github.com/Azure/azure-sdk-for-net/blob/8b0b10d6e0380d6f23941b0e864c1cf6d0657f02/sdk/servicebus/Microsoft.Azure.ServiceBus/src/Management/SubscriptionDescriptionExtensions.cs#L13
Will be updating to match.

sdk/servicebus/azure-servicebus/tests/mgmt_tests/test_mgmt_queues.py
```diff
@@ -302,6 +311,15 @@ def test_mgmt_queue_update_success(self, servicebus_namespace_connection_string,
...
             assert queue_description.lock_duration == datetime.timedelta(minutes=2)
 
+            # Update forwarding settings with entity name.
+            queue_description.forward_to = topic_name
+            queue_description.forward_dead_lettered_messages_to = topic_name
+            mgmt_service.update_queue(queue_description)
+
+            queue_description = mgmt_service.get_queue(queue_name)
+            assert queue_description.forward_dead_lettered_messages_to.endswith(".servicebus.windows.net/{}".format(topic_name))
+            assert queue_description.forward_to.endswith(".servicebus.windows.net/{}".format(topic_name))
```
yunhaoling, line 321: could we also test updating `forward_to` and `forward_dead_lettered_messages_to` to be None for the update operation?
```
            queue_description.forward_to = None
            queue_description.forward_dead_lettered_messages_to = None
            mgmt_service.update_queue(queue_description)

            queue_description = mgmt_service.get_queue(queue_name)
            assert not queue_description.forward_dead_lettered_messages_to
            assert not queue_description.forward_to
```

swathipil: meaning, setting it back to None after it has been set to a path?

yunhaoling (approved):
LGTM

PR #17442, opened 2021-03-18
Add HTTP responses to authentication exceptions

### LINKED ISSUE #16906: [Identity] Azure Identity should expose the original MSAL error

When `AuthenticationRequiredError` is raised, the original MSAL `result` (error dict) is discarded:

https://github.com/Azure/azure-sdk-for-python/blob/ef46a5c14a3c49c040a124d9c626c77cf2964cdc/sdk/identity/azure-identity/azure/identity/_internal/interactive.py#L198-L202

This makes `aad_exception_handler` (https://github.com/Azure/azure-cli/pull/17072) impossible to handle MSAL error directly.

`AuthenticationRequiredError` should have an attribute referring to the original MSAL `result`.

chlowell:
My goal here is to add Azure AD's response to each exception raised due to an Azure AD authentication error. That's easy where an exception is raised by code that already has the response. Less so for a credential implemented with MSAL, because MSAL returns only the response's deserialized body to the credential. So, this PR has MsalClient hang on to the last auth error response it saw, and adds a method credentials can call to get that response.

This closes #16906 by giving applications all the information AAD provided when authentication failed, for example:
```py
try:
    client.operation()
except AuthenticationRequiredError as ex:
    if ex.response:  # can be None, a response isn't always available
        content = json.loads(ex.response.text())
        ...
```

This PR removes the redundant `AuthenticationRequiredError.error_details`, whose value was simply `error_details` from AAD's response. That value is usually incorporated into an exception's message, and its raw value is available in the response.

### DIFF
--- a/sdk/identity/azure-identity/CHANGELOG.md
+++ b/sdk/identity/azure-identity/CHANGELOG.md
@@ -1,6 +1,11 @@
 # Release History
 
 ## 1.6.0b3 (Unreleased)
+### Breaking Changes
+> These changes do not impact the API of stable versions such as 1.5.0.
+> Only code written against a beta version such as 1.6.0b1 may be affected.
+- Removed property `AuthenticationRequiredError.error_details` 
+
 ### Fixed
 - Credentials consistently retry token requests after connection failures, or
   when instructed to by a Retry-After header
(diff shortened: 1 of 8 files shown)


mccoyp (approved):
This is an interesting solution to the issue! Do you think it would be worth adding a test that verifies the value of `AuthenticationRequiredError.response`?

chlowell:
I'm uncertain. A mock test of that scenario (specifically, "credential elicits an AAD error by presenting a refresh token") would be complex, fragile, and not representative because MSAL is responsible for all the HTTP requests in the real world. I actually wrote a recorded test for it, using CAE to invalidate a credential's refresh token. Having done that, I'm conflicted about what it means to "verify" the value of the response. It seems insufficient to simply assert that the exception carries a response, or that it carries any error response, but doing more requires having an opinion about the content of a "correct" response, which is really up to Azure AD to decide 😩

PR #17451, opened 2021-03-19
[Communication-Chat]Updated async credentials

annatisch:


RezaJooyandeh (approved):
Thank you so much for the quick fix. Looks great to me.

PR #17458, opened 2021-03-19
[Key Vault] Test keys library against managed HSM

### LINKED ISSUE #13716: Update KeyVault.Keys test suite to target MHSM
When MHSM becomes available as a live test resource, update the Keys live test suite to target a MHSM resource and mitigate any discovered issues.

mccoyp:
Resolves #13716.

This uses parameterized tests to run each applicable keys test against both Key Vault and Managed HSM, creating a recording for each. It also adds a platform to pipeline keys tests with `enableHsm = $true`, resulting in a single test run against MHSM.


sdk/keyvault/azure-keyvault-keys/tests/test_crypto_client.py
```diff
@@ -234,10 +291,13 @@ def test_wrap_and_unwrap(self, azure_keyvault_url, **kwargs):
         result = crypto_client.unwrap_key(result.algorithm, result.encrypted_key)
         self.assertEqual(key_bytes, result.key)
 
+    @parameterized.expand([param(is_hsm=True)], name_func=suffixed_test_name)
```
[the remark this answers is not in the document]
mccoyp: Hm... probably not. I had done it for consistency's sake more than anything, but if the test is never going to run against a vault then it's probably more clear to just avoid parameterizing.

sdk/keyvault/azure-keyvault-keys/tests/test_keys_async.py
```diff
@@ -258,9 +263,14 @@ async def test_backup_restore(self, azure_keyvault_url, **kwargs):
         restored_key = await self._poll_until_no_exception(restore_function, expected_exception=ResourceExistsError)
         self._assert_key_attributes_equal(created_bundle.properties, restored_key.properties)
 
+    @parameterized.expand([param(is_hsm=b) for b in [True, False]], name_func=suffixed_test_name)
     @KeyVaultPreparer()
     async def test_key_list(self, azure_keyvault_url, **kwargs):
-        client = self.create_client(azure_keyvault_url)
+        is_hsm = kwargs.pop("is_hsm")
+        self._skip_if_not_configured(is_hsm)
+        azure_keyvault_url = self.managed_hsm_url if is_hsm else azure_keyvault_url
```
chlowell, line 271: This method demands an argument but throws it away half the time because it knows better than the caller what its value should be. Could the caller instead determine the correct value?

mccoyp: The `azure_keyvault_url` parameter is automatically passed in as a positional parameter by the PowerShellPreparer, and the name matches the vault URL environment variable. I tried a couple different ways of fetching the MHSM URL through the preparer and passing that in as well, but wasn't able to get it working unfortunately

sdk/keyvault/azure-keyvault-keys/tests/_test_case_async.py
```diff
@@ -0,0 +1,52 @@
...
+
+from _shared.test_case_async import KeyVaultTestCase
+
+
+def suffixed_test_name(testcase_func, param_num, param):
+    suffix = "mhsm" if param.kwargs.get("is_hsm") else "vault"
+    return "{}_{}".format(testcase_func.__name__, parameterized.to_safe_name(suffix))
+
+
+class KeysTestCase(KeyVaultTestCase):
```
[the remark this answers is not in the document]
mccoyp: It would definitely be easy to create a mixin for these classes that contains `create_key_client`, `create_crypto_client`, and `_skip_if_not_configured`. Creating a base class that handles `setUp` and `tearDown` is probably doable as well, with these classes then inheriting from the base class and their respective KeyVaultTestCase. The latter would share code more efficiently but would be a bit more confusing -- is there a different setup that you had in mind, like a mixin for KeyVaultTestCase?

sdk/keyvault/azure-keyvault-keys/tests/test_keys_async.py
chlowell: Perhaps the (another?) problem then is that PowerShellPreparer requires the test to accept an argument it doesn't need? I don't have a solution in mind, I'm just poking at this confusion around who's responsible for figuring out client parameters👉 👉

sdk/keyvault/azure-keyvault-keys/tests/_test_case_async.py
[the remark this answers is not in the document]
chlowell: I was thinking that the async test case classes could inherit KeysTestCase and mix in the async methods. I was neglecting the create_client methods though. They make it look more complex than it's worth but maybe I'm just impatient or shortsighted. Anyway, not a high priority.

sdk/keyvault/azure-keyvault-keys/tests/test_keys_async.py
mccoyp: Yeah, it's not the best setup. I'd definitely like to dig into this when possible, but for now I don't see it as a blocker per se

sdk/keyvault/azure-keyvault-keys/tests/_test_case_async.py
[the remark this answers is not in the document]
mccoyp: I just pushed a base class implementation that shares all the code, with some (but I think acceptable levels of) complexity in the client creation methods

sdk/keyvault/azure-keyvault-keys/tests/test_keys_async.py
chlowell: Getting this right is worth the effort. I think that's appropriate effort for this PR because it's the first to confront this inflexibility of the current code. If you kick the can today, you can expect that someone else, probably future you, will have to choose between taking the extra time to refactor this kludge out of existence or adding another quick kludge on top of it. And in the meantime this kludge is just hanging out looking awkward :flushed:

sdk/keyvault/azure-keyvault-keys/tests/_test_case.py
```diff
@@ -0,0 +1,11 @@
+# ------------------------------------
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+# ------------------------------------
+from _test_case_base import KeysTestCaseBase
+from _shared.test_case import KeyVaultTestCase
+
+
+class KeysTestCase(KeysTestCaseBase, KeyVaultTestCase):
```
chlowell, line 9: Personally, I would just write e.g.
```py
class TestCaseWithTests(KeysTestCaseBase, KeyVaultTestCase)
```
everywhere instead of adding this empty class.

sdk/keyvault/azure-keyvault-keys/tests/test_keys_async.py
mccoyp: You're making points, and I know my future self appreciates it 😞

mccoyp: After brainstorming and messing with the preparer, I think the best approach here might actually be to not use the PowerShellPreparer for these tests. The main purpose of it is to unify/simplify fetching environment variables, but we already have to subvert that by manually getting the HSM URL and registering it with the scrubber -- which we have to do, because the PowerShellPreparer can't handle getting environment variables for two different directories (Key Vault and Managed HSM) simultaneously. 

Avoiding the preparer for these tests would unify the variable fetching approach, and it would also leave endpoint setting entirely to the test method. I think that makes sense, since the method is the only actor that's aware of whether or not it should target an HSM (via `is_hsm`). This would require re-implementing the preparer's [`_set_mgmt_settings_real_values` method](https://github.com/Azure/azure-sdk-for-python/blob/master/tools/azure-sdk-tools/devtools_testutils/powershell_preparer.py#L52-L62), but that could be done in the new KeysTestCase class.

chlowell (approved):

PR #17461, opened 2021-03-19
Fix timing issue in caching test

chlowell:
MSAL calculates the cached expires_on value of a token by adding its expires_in value to a timestamp, optionally provided at the time of caching. Async/ManagedIdentityClient does provide a timestamp, acquired before it requests a token. The cached expires_on value in these tests would therefore be later than expected if the system time rolls over to the next second before MSAL caches the token. That happened this morning, so here I am fixing it by ensuring the client gets a timestamp such that expires_in + timestamp = the expected expires_on. I almost fixed this by instead writing `assert token.expires_on - expected_expires_on < 2`  but felt it would then be necessary to add an explanatory paragraph like the one you're reading to the tests 😒


mccoyp (approved):
😵

PR #17487, opened 2021-03-22
Communication - Phone Numbers - Corrected LRO status

jbeauregardb:


malavika1 (approved):

sacheu (approved):

sdk/communication/azure-communication-phonenumbers/azure/communication/phonenumbers/__init__.py
```diff
@@ -15,6 +15,7 @@
     PhoneNumberAssignmentType,
     PhoneNumberCapabilityType,
     PhoneNumberType,
+    PhoneNumberOperationStatus
```
annatisch, line 18: Why is this being made public?

jbeauregardb: We're doing it to compare the result from the LRO directly from the PhoneNumberOperationStatus enum and to make sure the status that comes from the service stays consistent with the one we have registered.

annatisch: We should not make this part of the public API just to make the tests run....
We can import from the internal `_generated` namespace in the tests.

juancamilor: @annatisch  / @jbeauregardb could we address this one as a separate thing ? It is blocking an unrelated project (chat) and we are stuck since Thursday here.

PR #17496, opened 2021-03-22
[core] rest in azure core

iscai-msft:


sdk/core/azure-core/azure/core/pipeline/transport/_base.py
```diff
@@ -185,6 +189,22 @@ def close(self):
     def sleep(self, duration):  # pylint: disable=no-self-use
         time.sleep(duration)
 
+    @property
```
iscai-msft, line 192: These are the properties I've added to transport, to be able to support multiple formats

sdk/core/azure-core/azure/core/rest/__init__.py
```diff
@@ -0,0 +1,56 @@
+# --------------------------------------------------------------------------
```
iscai-msft, line 1: In looking up the provisional package, from what I understand of [PEP 411](https://www.python.org/dev/peps/pep-0411/#marking-a-package-provisional) is that it's only for entire packages. Is this correct?

If it is correct, would we want a preview module? https://www.python.org/dev/peps/pep-0408/

sdk/core/azure-core/tests/async_tests/test_authentication_async.py
```diff
@@ -7,17 +7,26 @@
 import time
 from unittest.mock import Mock
 
+from requests.models import RequestEncodingMixin
```
iscai-msft, line 10: why are there so many changes to the existing test files?

Reason is:
1. We mock transports a lot, and now that I've added these new properties to the transport, sometimes need to mock the properties
2. I've parameterized p much every test (besides `test_basic_transport.py`). Each test now first runs pipeline transport, then rest

sdk/core/azure-core/tests/async_tests/test_authentication_async.py
```diff
@@ -35,33 +44,45 @@ async def get_token(_):
...
+@pytest.mark.parametrize("request_type", [PipelineTransportHttpRequest, RestHttpRequest])
+async def test_bearer_policy_send(request_type, add_properties_to_transport):
     """The bearer token policy should invoke the next policy's send method and return the result"""
-    expected_request = HttpRequest("GET", "https://spam.eggs")
+    expected_request = request_type("GET", "https://spam.eggs")
     expected_response = Mock()
 
     async def verify_request(request):
-        assert request.http_request is expected_request
+        if hasattr(request_type, "content"):
```
iscai-msft, line 65: I use the attribute check of "content" to determine whether we're using pipline transport stuff, or rest stuff

sdk/core/azure-core/azure/core/rest/__init__.py
iscai-msft: Instead, we should just document each thing we're adding as azure.core.rest as provisional

PR #17503, opened 2021-03-23
Use LLC in EG

rakshith91:
Fixes https://github.com/Azure/azure-sdk-for-python/issues/17267


sdk/eventgrid/azure-eventgrid/azure/eventgrid/_helpers.py
```diff
@@ -134,3 +136,26 @@ def _cloud_event_to_generated(cloud_event, **kwargs):
...
+        header_parameters['Content-Length'] = str(len(data))
+
+    request = HttpRequest(
+        method="POST",
+        url=endpoint,
+        headers=header_parameters,
+        data=data
+    )
+    request.format_parameters(query_parameters)
+    return request
```
rakshith91, line 161:
```suggestion
    return request

```

sdk/eventgrid/azure-eventgrid/azure/eventgrid/_publisher_client.py
```diff
@@ -185,15 +187,13 @@ def send(self, events, **kwargs):
                 ]
             except AttributeError:
                 pass  # means it's a dictionary
-            kwargs.setdefault(
-                "content_type", "application/cloudevents-batch+json; charset=utf-8"
-            )
+            content_type = "application/cloudevents-batch+json; charset=utf-8"
```
iscai-msft, line 190: do you still want users to be able to set `content_type`?

sdk/eventgrid/azure-eventgrid/azure/eventgrid/_helpers.py
```diff
@@ -134,3 +136,27 @@ def _cloud_event_to_generated(cloud_event, **kwargs):
         additional_properties=cloud_event.extensions,
         **kwargs
     )
+
+def _build_request(endpoint, content_type, events, client):
+    serialize = client._serialize # pylint: disable=protected-access
```
iscai-msft, line 141: you don't actually really needs this. For the meantime, you can directly import `Serializer` from msrest and set `serialize = Serializer()`.

This will get rido f the need to pass the `client` parameter to `_build_request`

sdk/eventgrid/azure-eventgrid/azure/eventgrid/_helpers.py
```diff
@@ -134,3 +136,27 @@ def _cloud_event_to_generated(cloud_event, **kwargs):
...
+    header_parameters = {}  # type: Dict[str, Any]
+    header_parameters['Content-Type'] = serialize.header("content_type", content_type, 'str')
+
+    query_parameters = {}  # type: Dict[str, Any]
+    query_parameters['api-version'] = serialize.query("api_version", "2018-01-01", 'str')
+
+    body = serialize.body(events, '[object]')
```
iscai-msft, line 148: I believe you can just directly call `json.dumps(events)` instead of first serializing the body with the msrest serializer

sdk/eventgrid/azure-eventgrid/azure/eventgrid/_publisher_client.py
rakshith91: yes - updated :P

sdk/eventgrid/azure-eventgrid/azure/eventgrid/_helpers.py
rakshith91: single events are not JSON serializable - so need this

sdk/eventgrid/azure-eventgrid/azure/eventgrid/_helpers.py
```diff
@@ -134,3 +137,27 @@ def _cloud_event_to_generated(cloud_event, **kwargs):
         additional_properties=cloud_event.extensions,
         **kwargs
     )
+
+def _build_request(endpoint, content_type, events, client):
```
rakshith91, line 141:
```suggestion
def _build_request(endpoint, content_type, events):
```

sdk/eventgrid/azure-eventgrid/azure/eventgrid/_publisher_client.py
iscai-msft: I think you're overriding the content type you got from users here

iscai-msft (approved):

PR #17505, opened 2021-03-23
Add repr to Eg event

rakshith91:
Fixes https://github.com/Azure/azure-sdk-for-python/issues/17283

### DIFF
--- a/sdk/eventgrid/azure-eventgrid/azure/eventgrid/_models.py
+++ b/sdk/eventgrid/azure-eventgrid/azure/eventgrid/_models.py
@@ -91,7 +91,7 @@ class EventGridEvent(InternalEventGridEvent):
         kwargs.setdefault("data_version", data_version)
 
         super(EventGridEvent, self).__init__(**kwargs)
-    
+
     def __repr__(self):
         return "EventGridEvent(subject={}, event_type={}, id={}, event_time={})".format(
             self.subject, self.event_type, self.id, self.event_time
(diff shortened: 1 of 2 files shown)


sdk/eventgrid/azure-eventgrid/azure/eventgrid/_models.py
```diff
@@ -91,3 +91,8 @@ def __init__(self, subject, event_type, data, data_version, **kwargs):
         kwargs.setdefault("data_version", data_version)
 
         super(EventGridEvent, self).__init__(**kwargs)
+    
+    def __repr__(self):
+        return "EventGridEvent(subject={}, event_type={}, id={}, event_time={})".format(
+            self.subject, self.event_type, self.id, self.event_time
+        )[:1024]
```
yunhaoling, line 98: dumb question: only those four properties are needed for `__repr__`?
what about the other ivars like `data`, `data_version` etc.

rakshith91: The purpose of __repr__  is to give a small gist of what the object is with important information - and trimmed to 1024 characters instead of some random hexcode. It's NOT to give the complete info.
data and data_version aren't the most important aspects of an event

swathipil: out of curiosity, a quick follow-up: why is `data` not considered an important aspect of an event?

rakshith91: just to be clear - it's TYPICALLY not THE most important aspect in an event when representing the event object in the sense that having data doesn't provide as much value as providing subject and id given we only want 1024 characters in the repr

swathipil: consistency with CloudEvent repr makes sense!

swathipil (approved):

PR #17512, opened 2021-03-23
[Communication] - Phone Numbers - Removed PhoneNumberOperationStatus from public exposure

jbeauregardb:
Follow up PR for removing the PhoneNumberOperationStatus enum from the public init file as @annatisch suggested.


annatisch (approved):

RezaJooyandeh:
Was this type ever released? If yes, we should update the changelog

jbeauregardb:
@RezaJooyandeh No, this was never released. The fix I made yesterday for the LRO status issue we had on the chat live tests exposed this enum temporarily. @annatisch made the comment that this should not be exposed, so I changed it in this follow up PR 👍🏻

RezaJooyandeh (approved):

PR #17523, opened 2021-03-24
[translation] surface inner error level code/message to users

kristapratico:


mohamedshabanofficial (approved):

sdk/documenttranslation/azure-ai-documenttranslation/azure/ai/documenttranslation/_models.py
```diff
@@ -307,11 +307,18 @@ def __init__(
...
         # type: (**Any) -> None
         self.code = kwargs.get('code', None)
-        self.message = None
-        self.target = None
+        self.message = kwargs.get('message', None)
+        self.target = kwargs.get('target', None)
 
     @classmethod
     def _from_generated(cls, error):
+        if error.inner_error:
```
maririos, line 315: this assumes there is only one inner error per error, right?

kristapratico: Right, just like we do in TA. Alex confirmed that they are only sending one. If they ever decide to send nested inner errors in the future that can be additive. For now, we'll bubble up the most useful error code/message to the user.

maririos (approved):

Issue #17368, opened 2021-03-16
Azure.Core: AzureNamedKeyCredential 


jsquire:
# Summary

When using a shared key for authorization, some services require that the name of the key be used as a component of the algorithm to form the authorization token.  For these cases, `AzureKeyCredential` and `AzureSasCredential` aren't adequate as they track only a single value.  A new credential type, `AzureNamedKeyCredential` has been approved to address this gap.

# Goals

- Offer a token credential suitable for working with any service needing a named key that follows patterns established by the other credential types offered in `Azure.Core` and `Azure.Identity`.

- Developers should be able to construct the credential by providing a string-based name and key.

- Developers should be able to update the name and key after construction to support rolling credentials or using a new access policy.  Updating credential information should be done in-place and not require creating a new credential instance.

- Developers should be able to read the name of they key, in order to support caching within an application so that it can be easily identified when updates are needed.

- The Azure SDKs should be able to read the name and key, ensuring they reflect the most current values.

- Support updates and reads concurrently with consistent behavior.  Developers should be able to update the credential at any time without concern for whether the SDK is reading from it.

# Non-Goals

- Providing a friendly and discoverable means for developers to read the key from the credential; it is preferable to hide the key or discourage use.

- Allowing the name or key to be updated independently; both elements will be required when updating a credential.

# Success Criteria

- An `AzureNamedKeyCredential` type has been implemented, following language guidelines and idioms.

- The tests necessary for its validation have been created and pass reliably.

- The existing test suite continues to produce deterministic results and pass reliably.

# Related

- [.NET AzureNamedKeyCredential Design Options](https://gist.github.com/jsquire/861fb365fa672dd62530500ad2bc0db9)

PR #17548, opened 2021-03-24
Add AzureNamedKeyCredential

rakshith91:
Fixes https://github.com/Azure/azure-sdk-for-python/issues/17368

# Summary

When using a shared key for authorization, some services require that the name of the key be used as a component of the algorithm to form the authorization token.  For these cases, `AzureKeyCredential` and `AzureSasCredential` aren't adequate as they track only a single value.  A new credential type, `AzureNamedKeyCredential` has been approved to address this gap.

# Goals

- Offer a token credential suitable for working with any service needing a named key that follows patterns established by the other credential types offered in `Azure.Core` and `Azure.Identity`.

- Developers should be able to construct the credential by providing a string-based name and key.

- Developers should be able to update the name and key after construction to support rolling credentials or using a new access policy.  Updating credential information should be done in-place and not require creating a new credential instance.

- Developers should be able to read the name of they key, in order to support caching within an application so that it can be easily identified when updates are needed.

- The Azure SDKs should be able to read the name and key, ensuring they reflect the most current values.

- Support updates and reads concurrently with consistent behavior.  Developers should be able to update the credential at any time without concern for whether the SDK is reading from it.

# Non-Goals

- Providing a friendly and discoverable means for developers to read the key from the credential; it is preferable to hide the key or discourage use.

- Allowing the name or key to be updated independently; both elements will be required when updating a credential.

# Success Criteria

- An `AzureNamedKeyCredential` type has been implemented, following language guidelines and idioms.

- The tests necessary for its validation have been created and pass reliably.

- The existing test suite continues to produce deterministic results and pass reliably.


xiangyan99:
The code looks good. 

Please talk to Johan & Anna for the architect perspective feedback.

sdk/core/azure-core/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### Features
 
+- Added `azure.core.credentials.AzureNamedKeyCredential` credential #17548.
```
swathipil, line 7: maybe mention policy?
```suggestion
- Added `azure.core.credentials.AzureNamedKeyCredential` credential and its respective policy #17548.
```

swathipil (commented):
LGTM

sdk/core/azure-core/azure/core/credentials.py
```diff
@@ -111,3 +111,49 @@ def update(self, signature):
...
+    @property
+    def key(self):
+        # type () -> str
+        """The value of the configured key.
+
+        :rtype: str
+        """
+        return self._key
+
+    def update(self, name, key):
```
johanste, line 149: Do we have scenarios where you want to update/change the name of the key?

sdk/core/azure-core/azure/core/credentials.py
```diff
@@ -111,3 +111,49 @@ def update(self, signature):
...
+        # type: (str, str) -> None
+        """Update the named key credential.
+
+        Both name and key must be provided in order to update the named key credential.
+        Individual attributes cannot be updated.
+        """
+        if not isinstance(name, six.string_types) or not isinstance(key, six.string_types):
+            raise TypeError("Both name and key must be Strings.")
+        self._name = name
```
johanste, line 158: The reason for a single update method rather than two read/write properties is (per my understanding) to make the assignment atomic. This is not an atomic assignment. Is there a specific reason we have an non-atomic `update` method?

sdk/core/azure-core/azure/core/pipeline/policies/_authentication.py
```diff
@@ -145,3 +151,20 @@ def on_request(self, request):
...
+    """Adds a key header for the provided credential.
+
+    :param credential: The credential used to authenticate requests.
+    :type credential: ~azure.core.credentials.AzureNamedKeyCredential
+    :raises: ValueError or TypeError
+    """
+    def __init__(self, credential, **kwargs):  # pylint: disable=unused-argument
+        # type: (AzureNamedKeyCredential, **Any) -> None
+        super(AzureNamedKeyCredentialPolicy, self).__init__()
+        self._key = credential.key
```
johanste, line 166: What happens when a client is created with a given `NamedKeyCredential` and later call `update` on the NamedKeyCredential? The whole reason for the update method is to be able to roll the key.

sdk/core/azure-core/azure/core/credentials.py
xiangyan99: How common user creates an AzureNamedKeyCredential then modify the name?

It seems to me maybe

```py
 def update(self, key, **kwarg):
```
makes more sense?

@johanste

sdk/core/azure-core/azure/core/credentials.py
yunhaoling: I think one scenario could be that users want to use a new share access policy which has new name and key.

sdk/core/azure-core/azure/core/credentials.py
rakshith91: the `update` method is to 
1. be consistent with other creds in python (aka AzureKeyCredential)
2. be consistent with other languages when it comes to `AzureNamedKeyCredential`

sdk/core/azure-core/azure/core/credentials.py
```diff
@@ -111,3 +112,38 @@ def update(self, signature):
...
+
+    def update(self, name, key):
+        # type: (str, str) -> None
+        """Update the named key credential.
+
+        Both name and key must be provided in order to update the named key credential.
+        Individual attributes cannot be updated.
+        """
+        if not isinstance(name, six.string_types) or not isinstance(key, six.string_types):
+            raise TypeError("Both name and key must be Strings.")
```
johanste, line 148: why Strings (with capital S)?

sdk/core/azure-core/azure/core/credentials.py
```diff
@@ -111,3 +112,38 @@ def update(self, signature):
...
+    :raises: TypeError
+    """
+    def __init__(self, name, key):
+        # type: (str, str) -> None
+        if not isinstance(name, six.string_types) or not isinstance(key, six.string_types):
+            raise TypeError("Both name and key must be Strings.")
+        self._credential = AzureNamedKey(name, key)
+
+    @property
+    def credential(self):
```
johanste, line 132: It feels odd to have a credential property on a credential object. Is this the right name?

sdk/core/azure-core/azure/core/pipeline/policies/_authentication.py
```diff
@@ -145,3 +151,20 @@ def on_request(self, request):
...
+
+
+class AzureNamedKeyCredentialPolicy(SansIOHTTPPolicy):
+    """Adds a key header for the provided credential.
+
+    :param credential: The credential used to authenticate requests.
+    :type credential: ~azure.core.credentials.AzureNamedKeyCredential
+    :raises: ValueError or TypeError
+    """
+    def __init__(self, credential, name, **kwargs):  # pylint: disable=unused-argument
```
johanste, line 163: Who is consuming the name of the `AzureNamedKeyCredential`?

sdk/core/azure-core/azure/core/pipeline/policies/_authentication.py
```diff
@@ -17,7 +17,12 @@
 if TYPE_CHECKING:
     # pylint:disable=unused-import
     from typing import Any, Dict, Optional
-    from azure.core.credentials import AccessToken, TokenCredential, AzureKeyCredential, AzureSasCredential
+    from azure.core.credentials import (
+        AccessToken,
+        TokenCredential,
+        AzureKeyCredential,
+        AzureSasCredential,
+    )
```
rakshith91, line 25:
```suggestion
    from azure.core.credentials import AccessToken, TokenCredential, AzureKeyCredential, AzureSasCredential
```

sdk/core/azure-core/azure/core/pipeline/policies/__init__.py
```diff
@@ -25,7 +25,10 @@
 # --------------------------------------------------------------------------
 
 from ._base import HTTPPolicy, SansIOHTTPPolicy, RequestHistory
-from ._authentication import BearerTokenCredentialPolicy, AzureKeyCredentialPolicy, AzureSasCredentialPolicy
+from ._authentication import (
+    BearerTokenCredentialPolicy, AzureKeyCredentialPolicy, AzureSasCredentialPolicy,
+    AzureNamedKeyCredentialPolicy
+    )
```
rakshith91, line 31:
```suggestion
from ._authentication import BearerTokenCredentialPolicy, AzureKeyCredentialPolicy, AzureSasCredentialPolicy
```

sdk/core/azure-core/azure/core/pipeline/policies/__init__.py
```diff
@@ -45,6 +45,7 @@
     'SansIOHTTPPolicy',
     'BearerTokenCredentialPolicy',
     'AzureKeyCredentialPolicy',
+    'AzureNamedKeyCredentialPolicy',
```
rakshith91, line 48:
```suggestion
```

sdk/core/azure-core/CHANGELOG.md
```diff
@@ -1,7 +1,10 @@
 # Release History
 
-## 1.13.1 (Unreleased)
+## 1.14.0 (Unreleased)
 
+### New Features
+
+- Added `azure.core.credentials.AzureNamedKeyCredential` credential and its respective policy #17548.
```
xiangyan99, line 7: I don't see the policy in your PR?

sdk/core/azure-core/azure/core/credentials.py
xiangyan99: Can user update key only?

sdk/core/azure-core/azure/core/credentials.py
johanste: No. The reason for the update method is to be atomic. You cannot do that unless you expect callers to lock the instance when reading values.

sdk/core/azure-core/azure/core/credentials.py
rakshith91: technically - they are allowed to provide the same name - but they must provide a name

sdk/core/azure-core/CHANGELOG.md
rakshith91: updted the change log

sdk/core/azure-core/azure/core/credentials.py
xiangyan99: I think the most common scenario is the name is fixed after creation, but user may modify the key value?

sdk/core/azure-core/azure/core/credentials.py
rakshith91: changed to `named_key`

sdk/core/azure-core/azure/core/credentials.py
rakshith91: @yunhaoling  any thoughts on this?

If we know that users may normally not modify the name, I'll incline towards making name kwarg only

rakshith91: actually on second thought, i think we should let this be - as much as i like kwargs, it doesn't seem like they'll grow later in this case.

rakshith91: also, if the user thinks they wont need to change name for their use case, AzureKeyCredential serves their purpose

johanste: ...unless they need to initially specify the name?

sdk/core/azure-core/tests/test_authentication.py
```diff
@@ -230,3 +233,25 @@ def test_azure_sas_credential_policy_raises():
...
+
+def test_azure_named_key_credential():
+    cred = AzureNamedKeyCredential("sample_name", "samplekey")
+
+    assert cred.named_key.name == "sample_name"
+    assert cred.named_key.key == "samplekey"
+
+    cred.update("newname", "newkey")
+    assert cred.named_key.name == "newname"
+    assert cred.named_key.key == "newkey"
```
johanste, line 245: We should have a test to make sure that we are returning a tuple as well...

johanste (approved):

sdk/core/azure-core/azure/core/credentials.py
rakshith91: hmm, right - still doesn't change tht kwargs would not grow in this case - so will stick to it

sdk/core/azure-core/tests/test_authentication.py
rakshith91: added

Issue #17418, opened 2021-03-18
[azure-storage-blob] Blob Upload never finishes if a generator is passed


mth-cbc:
- **Package Name**: azure-storage-blob
- **Package Version**: 12.6.0
- **Operating System**: Ubuntu 18.04 in WSL on Windows 10 64-Bit
- **Python Version**:  3.8.0

**Describe the bug**
I want to pass a generator of the form:
```python
def data_generator():
    # do something
    yield data
```
to the function `upload_blob()` to avoid the need to hold all of my (potentially huge) data in memory.

The upload, however, never finishes even after the generator is exhausted.

**To Reproduce**
1. Get a `BlobClient`
2. Implement a generator which in total produces more data than your chunk size
3. Call `blob_client.upload_blob(data=<generator>)`
4. Observe that the operation never finishes

Here is a minimal script with which I could tirgger the behaviour every time:
```python
from azure.storage.blob import BlobClient


# Size of raw_string combined with "range" in `test_generator` needs
# to be bigger than the chunk size (otherwise everything is put in one
# go, instead of a "stream" case)
raw_string = "a" * 1024 * 1024 * 3


def test_generator():
    for it in range(0, 2):
        print(it)
        yield raw_string
    print("Exhausted generator")


test: BlobClient = BlobClient.from_connection_string(
    conn_str="<conn_str>",
    container_name="ttp",
    blob_name="generator_test.txt",
)

test.upload_blob(data=test_generator(), overwrite=True)
print("Finished")
```

The corresponding output is this:
```
0
1
Exhausted generator

```
Even after waiting several minutes the program does not finish.

**Expected behavior**
The upload should succeed normally and the blob should contain all the data produced by the generator.

**Additional context**
I wasn't completly sure if the `upload_blob` function was supposed to work as I expected. So I investigated a bit further and I might have found the underlying problem. This [`IterStreamer`](https://github.com/Azure/azure-sdk-for-python/blob/af066bb6990279e5868407918d9cd82ad159e7e7/sdk/storage/azure-storage-blob/azure/storage/blob/_shared/uploads.py#L503) does not behave as I would expect. Here some script which visualises my observations:

```python
import six


# Copy of the IterStreamer Class from
# https://github.com/Azure/azure-sdk-for-python/blob/af066bb6990279e5868407918d9cd82ad159e7e7/sdk/storage/azure-storage-blob/azure/storage/blob/_shared/uploads.py#L503
class IterStreamer(object):
    """
[...]

PR #17551, opened 2021-03-25
[Blob] Fixed the ability to upload from a generator

tasherif-msft:
This resolves https://github.com/Azure/azure-sdk-for-python/issues/17418

### DIFF
--- a/sdk/storage/azure-storage-blob/azure/storage/blob/_shared/uploads.py
+++ b/sdk/storage/azure-storage-blob/azure/storage/blob/_shared/uploads.py
@@ -541,10 +541,15 @@ class IterStreamer(object):
                     chunk = chunk.encode(self.encoding)
                 data += chunk
                 count += len(chunk)
+        # This means count < size and what's leftover will be returned in this call.
         except StopIteration:
+            self.leftover = b""
             pass
 
         if count > size:
             self.leftover = data[size:]
+
+        if count == size:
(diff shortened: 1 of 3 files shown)


varun-shahdadpuri:
Can we please expedite this PR? We have been facing issues while trying to upload block blob using a generator.

sdk/storage/azure-storage-blob/azure/storage/blob/_shared/uploads.py
```diff
@@ -541,10 +541,14 @@ def read(self, size):
                     chunk = chunk.encode(self.encoding)
                 data += chunk
                 count += len(chunk)
+        # This means count < size and what's leftover will be returned in this call.
         except StopIteration:
-            pass
+            self.leftover = b""
 
         if count > size:
```
xiafu-msft, line 548:
```suggestion
        if count >= size:
```

sdk/storage/azure-storage-blob/azure/storage/blob/_shared/uploads.py
```diff
@@ -541,10 +541,14 @@ def read(self, size):
...
             self.leftover = data[size:]
 
+        if count == size:
+            self.leftover = b""
+
```
xiafu-msft, line 553:
```suggestion
```

xiafu-msft (approved):

Issue #15648, opened 2020-12-04
First chunk of StorageStreamDownloader.chunks() is always 32MB big


mth-cbc:
- **Package**: azure-storage-blob
- **Package Version**: 12.6.0
- **Operating System**: Ubuntu 18.04
- **Python Version**: 3.8.0

**Describe the bug**
When using the `StorageStreamDownloader`s chunk interator to download/iterator offer a blob the first chunk is always 32\*1024\*1024 bytes big, regardless of the setting for `max_chunk_get_size`.

**To Reproduce**
1. instantiate a `BlobClient` for a file bigger than 32 MB
2. get a `StorageStreamDownloader` for the blob: `stream = blob_client.download_blob()`
3. iterate over the blob using the `chunks()` function and print the chunk size:
```
for chunk in stream.chunks():
    print(len(chunk))
```

**Expected behavior**
Each chunk, apart from the last one, should have the size given by the `max_chunk_get_size` parameter, so by default 4\*1024\*1024 bytes.

xiafu-msft:
Hi @mth-cbc

Thanks so much for reporting this issue, it's an SDK bug. 
Currently you can use this workaround by setting max_chunk_get_size=`your expected chunk size`, then the first chunk size should be changed.

Let me know if you have any concern! Sorry about the inconvenience.

mth-cbc:
Hi @xiafu-msft 

Thanks for your answer, but I observe the same behaviour when specifing an explicit chunk size via `max_chunk_get_size`.

xiafu-msft:
Sorry I had a typo there! I mean max_single_get_size, updated the previous comment

mth-cbc:
Thanks for the workaround, now it works as expected.

tasherif-msft:
This pr https://github.com/Azure/azure-sdk-for-python/pull/17559 will resolve this issue in the future :) 
I will go ahead and close this issue!

PR #17559, opened 2021-03-25
[Blob][Fix]fix downloader.chunks() return chunks in different size

xiafu-msft:
when iterate through downloader.chunks() the chunk size is different. The first chunk size is always equals to max_single_get_size while it should equal to max_chunk_get_size 
This will fix https://github.com/Azure/azure-sdk-for-python/issues/9419 and https://github.com/Azure/azure-sdk-for-python/issues/15648


sdk/storage/azure-storage-blob/azure/storage/blob/_download.py
```diff
@@ -231,21 +233,37 @@ def __next__(self):
...
-            self._current_content = self._iter_downloader.yield_chunk(chunk)
+            self._current_content += self._iter_downloader.yield_chunk(chunk)
+        except StopIteration as e:
+            self._complete = True
+            if self._current_content:
+                return self._current_content
+            raise e
 
-        return self._current_content
+        return self._get_chunk_data()
```
tasherif-msft, line 258: What's the reason for calling `self._get_chunk_data()` in this case? as far as I understand if we hit this case (we're done iterating the first single get call and now iterating through our download chunks like we usually do) then current content should always be the same as the chunk size so we can just return `self.current_content`. It seems from your `try-except` that there is an edge case where there will be excess content - could you just put a small comment on that in the code - its not obvious to me haha

EDIT: I just realized the case is when the current content from the first get is still there but smaller than chunk size therefore we want to make sure its also included. I think adding a comment in the code would be very helpful here!
Good work thinking of all the cases

tasherif-msft (approved):
Great work and test LGTM :) **please** add the comment I mentioned in my review for that case as it will be helpful to understand the code for future devs.

PR #17566, opened 2021-03-25
adding logging information to tags on resource group preparer

seankane-msft:


benbp (approved):

tools/azure-sdk-tools/devtools_testutils/resource_testcase.py
```diff
@@ -65,22 +69,29 @@ def create_resource(self, name, **kwargs):
             parameters = {"location": self.location}
             if self.delete_after_tag_timedelta:
                 expiry = datetime.datetime.utcnow() + self.delete_after_tag_timedelta
-                parameters["tags"] = {
-                    "DeleteAfter": expiry.replace(microsecond=0).isoformat()
-                }
+                parameters["tags"] = {"DeleteAfter": expiry.replace(microsecond=0).isoformat()}
```
weshaggard, line 72: This looks like it ends up being optional which might explain some of the RG's that don't have a DeleteAfter tag. Can we instead ensure we have a default time delta so that this DeleteAfter always gets added.

weshaggard: Also it looks like above the value is set to 1 day we might want to lower that to 8 hours.

tools/azure-sdk-tools/devtools_testutils/resource_testcase.py
```diff
@@ -97,18 +108,12 @@ def remove_resource(self, name, **kwargs):
                     azure_poller.wait(kwargs.get("wait_timeout"))
                     if azure_poller.done():
                         return
-                    raise AzureTestError(
-                        "Timed out waiting for resource group to be deleted."
-                    )
+                    raise AzureTestError("Timed out waiting for resource group to be deleted.")
                 else:
                     self.client.resource_groups.delete(name, polling=False)
```
weshaggard, line 113: Do we know how reliable this runs? Is there any fallback that helps ensure this delete runs?

tools/azure-sdk-tools/devtools_testutils/resource_testcase.py
```diff
@@ -65,22 +69,29 @@ def create_resource(self, name, **kwargs):
...
+                parameters["tags"]["BuildReason"] = os.environ["BUILD_REASON"]
+            except KeyError:
+                pass
             try:
-                self.resource = self.client.resource_groups.create_or_update(
-                    name, parameters
+                logging.info(
+                    "Attempting to create a Resource Group with name {} and parameters {}".format(name, parameters)
                 )
+                self.resource = self.client.resource_groups.create_or_update(name, parameters)
```
weshaggard, line 86: Would it be possible to add a common prefix, something like "rgp-" to all the names? So we know that the resource groups are coming from here.

tools/azure-sdk-tools/devtools_testutils/resource_testcase.py
seankane-msft: It's not optional, the default is 1 day, I will lower it to 8 hours.

weshaggard: Why is there a check for the value? If it isn't optional we should always just add it.

seankane-msft: Good point, if a user is setting `delete_after_tag_timedelta=None` this will be caught and we can investigate that situation. I'll make the fix.

PR #17573, opened 2021-03-25
Resource Group Preparer Logging

seankane-msft:


tools/azure-sdk-tools/devtools_testutils/resource_testcase.py
```diff
@@ -83,6 +82,10 @@ def create_resource(self, name, **kwargs):
                 logging.info(
                     "Attempting to create a Resource Group with name {} and parameters {}".format(name, parameters)
                 )
+                # Prefixing all RGs created here with 'rgpy-' for tracing purposes
+                name = u"rgpy-" + name
+                if len(name) > 90:
```
weshaggard, line 87: Is 90 the max name size?

tools/azure-sdk-tools/devtools_testutils/resource_testcase.py
```diff
@@ -104,14 +107,16 @@ def remove_resource(self, name, **kwargs):
         if self.is_live and self._need_creation:
             try:
                 if "wait_timeout" in kwargs:
-                    azure_poller = self.client.resource_groups.delete(name)
+                    azure_poller = self.client.resource_groups.begin_delete(name)
```
weshaggard, line 110: So were all the calls to delete failing all the time before?

tools/azure-sdk-tools/devtools_testutils/resource_testcase.py
seankane-msft: Yes, verified with the portal.

tools/azure-sdk-tools/devtools_testutils/resource_testcase.py
```diff
@@ -67,12 +69,9 @@ def create_resource(self, name, **kwargs):
         if self.is_live and self._need_creation:
             self.client = self.create_mgmt_client(ResourceManagementClient)
             parameters = {"location": self.location}
-            if self.delete_after_tag_timedelta:
-                expiry = datetime.datetime.utcnow() + self.delete_after_tag_timedelta
-                parameters["tags"] = {"DeleteAfter": expiry.replace(microsecond=0).isoformat()}
+            expiry = datetime.datetime.utcnow() + self.delete_after_tag_timedelta
+            parameters["tags"] = {"DeleteAfter": expiry.replace(microsecond=0).isoformat()}
```
weshaggard, line 73: Should we add these to the try block below with the other tags?

tools/azure-sdk-tools/devtools_testutils/resource_testcase.py
seankane-msft: Most likely yes. Any exception was being caught so there was no way to notify that it was failing.

tools/azure-sdk-tools/devtools_testutils/resource_testcase.py
seankane-msft: This one doesn't need to be, the others do because when running locally those environment variables wont be set.

weshaggard: Ok so the there is no try get env variable in python to allow for these to be empty if they aren't found?

weshaggard (approved):

PR #17586, opened 2021-03-26
frontdoor sdk generate

RAY-316:
frontdoor sdk generate


sdk/network/azure-mgmt-frontdoor/azure/mgmt/frontdoor/_metadata.json
```diff
@@ -1,72 +0,0 @@
-{
-    "chosen_version": "",
-    "total_api_version_list": ["2019-11-01", "2020-05-01", "2020-11-01"],
-    "client": {
```
jsntcy, line 4: We don't need this metadata file now?

PR #11284, opened 2020-05-06
check transport kwargs

xiangyan99:

PR #17610, opened 2021-03-26
Fix RetryPolicy setting extremely long connection timeout

### LINKED ISSUE #17481: RetryPolicy sets an extreme default connection timeout
When a caller provides no value for `connection_timeout` in a call to `Pipeline.run`, RetryPolicy sets `connection_timeout` to a default value:

https://github.com/Azure/azure-sdk-for-python/blob/58c46eed125a0c5ff9b64c9a479062dce5919080/sdk/core/azure-core/azure/core/pipeline/policies/_retry.py#L299-L304

`absolute_timeout` here is either the value of the `timeout` kwarg to `Pipeline.run` or 7 days. So in the typical case of a caller not passing a value for `connection_timeout` or `timeout`, RetryPolicy sets the request's `connection_timeout` to 7 days, overriding the transport's configuration (which has a default 300 seconds).

chlowell:
This fixes #17481 by ensuring RetryPolicy sets `connection_timeout` on a request only when it determines doing so will prevent a connection timeout from exceeding the policy's overall timeout.

### DIFF
--- a/sdk/core/azure-core/azure/core/pipeline/policies/_retry.py
+++ b/sdk/core/azure-core/azure/core/pipeline/policies/_retry.py
@@ -303,11 +303,15 @@ class RetryPolicyBase(object):
             request.context.options["connection_timeout"] = min(connection_timeout, absolute_timeout)
 
         # otherwise, try to ensure the transport's configured connection_timeout doesn't exceed absolute_timeout
-        # ("connection_config" isn't defined on HttpTransport but all implementations in this library have it)
+        # ("connection_config" isn't defined on Async/HttpTransport but all implementations in this library have it)
         elif hasattr(request.context.transport, "connection_config"):
             default_timeout = getattr(request.context.transport.connection_config, "timeout", absolute_timeout)
-            if absolute_timeout < default_timeout:
-                request.context.options["connection_timeout"] = absolute_timeout
+            try:
+                if absolute_timeout < default_timeout:
+                    request.context.options["connection_timeout"] = absolute_timeout
+            except TypeError:
+                # transport.connection_config.timeout is something unexpected (not a number)
+                pass
 
     def _configure_positions(self, request, retry_settings):
         body_position = None
--- a/sdk/core/azure-core/azure/core/pipeline/policies/_retry.py
+++ b/sdk/core/azure-core/azure/core/pipeline/policies/_retry.py
@@ -296,12 +296,18 @@ class RetryPolicyBase(object):
             if is_response_error:
                 raise ServiceResponseTimeoutError('Response timeout')
             raise ServiceRequestTimeoutError('Request timeout')
+
+        # if connection_timeout is already set, ensure it doesn't exceed absolute_timeout
         connection_timeout = request.context.options.get('connection_timeout')
         if connection_timeout:
-            req_timeout = min(connection_timeout, absolute_timeout)
-        else:
-            req_timeout = absolute_timeout
-        request.context.options['connection_timeout'] = req_timeout
+            request.context.options["connection_timeout"] = min(connection_timeout, absolute_timeout)
+
+        # otherwise, try to ensure the transport's configured connection_timeout doesn't exceed absolute_timeout
+        # ("connection_config" isn't defined on HttpTransport but all implementations in this library have it)
+        elif hasattr(request.context.transport, "connection_config"):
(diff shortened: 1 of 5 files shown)


sdk/core/azure-core/azure/core/pipeline/policies/_retry.py
```diff
@@ -296,12 +296,18 @@ def _configure_timeout(self, request, absolute_timeout, is_response_error):
...
         if connection_timeout:
-            req_timeout = min(connection_timeout, absolute_timeout)
-        else:
-            req_timeout = absolute_timeout
-        request.context.options['connection_timeout'] = req_timeout
+            request.context.options["connection_timeout"] = min(connection_timeout, absolute_timeout)
+
+        # otherwise, try to ensure the transport's configured connection_timeout doesn't exceed absolute_timeout
+        # ("connection_config" isn't defined on HttpTransport but all implementations in this library have it)
+        elif hasattr(request.context.transport, "connection_config"):
```
xiangyan99, line 307: So if there is no request.context.options.connection_timeout nor transport.connection_config, timeout will not take effect?

chlowell: If we've exceeded "timeout", `absolute_timeout <= 0` and this function raised an error above. Here it's just checking whether the transport's configured connection_timeout exceeds absolute_timeout.

xiangyan99: The case is user-specified timeout setting is less than our default one (5 min), in this case, timeout will not take effect?

xiangyan99: User calls the method some_method(timeout=100)

chlowell: Assuming the transport is one of ours and configured with the default connection_timeout, when a user calls  some_method(timeout=100), this code will set connection_timeout=100 for the request.

xiangyan99: Yes. But if transport.connection_config does not exist (they may be not using our transport), we still want timeout setting to work, right?

chlowell: It does: RetryPolicy won't send another retry after absolute_timeout reaches 0. We make an effort to fail as soon as possible after that but don't promise to fail the moment it happens. We can't adjust connection_timeout in the case you describe, because implementing connection_timeout and any default value for it is the transport's responsibility.

xiangyan99: Why we cannot adjust connection_timeout?

If I am a customer, I would prefer I don't need to specify connection_timeout or read_timeout, instead, I only have timeout and SDK take care of the rest for me.

chlowell: Because implementing connection_timeout is the transport's responsibility. In your scenario, the pipeline has a custom transport that does not expose its configuration as our in-box transports do. RetryPolicy therefore can't get a connection_timeout value from the transport to compare to absolute_timeout. It can't even tell whether this transport will observe connection_timeout.

If you want the SDK to take care of these things for you, you should not use a custom transport, or your custom transport should follow our conventions 😉

xiangyan99: connection_timeout will be part of the contract that if users want to use their custom transport, they have to support it.

We added connection_timeout in https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/core/azure-core/README.md#configurations.

And we are working on https://github.com/Azure/azure-sdk-for-python/pull/11284.

chlowell: That doesn't force a transport to observe connection_timeout. But I don't think whether it does is crucial for the code we're commenting on. The point I'm trying to make is that if a transport implements connection_config as our transports do, then this code will set connection_timeout to fall within absolute_timeout, as you want. However, connection_config is not part of Async/HttpTransport, so a transport may not implement it. If it does not, RetryPolicy can't determine the effective connection_timeout. I think RetryPolicy should set connection_timeout only when confident that's the right thing to do.

xiangyan99 (approved):

PR #17614, opened 2021-03-27
[Datalake] HNS Soft Delete

tasherif-msft:


xiafu-msft:
just want to make sure the bugfix for private drop is also included right

xiafu-msft (approved):

tasherif-msft:
> just want to make sure the bugfix for private drop is also included right

yep!

PR #17640, opened 2021-03-29
[core] connection string parser only

swathipil:
This is an attempt to create a generic connection string parsing function that will be used across SDKs to minimize redundant logic and validation, addressing #16782. 

Split from PR: #17049 

This parsed connection string will be returned as a dict of connection string keys (ie. `Endpoint`, `SharedAccessKey`, etc.) and values, with an option for allowing case insensitive keys. This option is useful when checking a connection string with keys that don't match expected case, without looping through all the keys. For example:
```
conn_str = "Endpoint=XXXXENDPOINT;ACcEsSKEY=XXXXKEY"
parsed = parse_connection_string(conn_str, case_sensitive_keys=False)
endpoint = parsed.get("endpoint")   # value is XXXXENDPOINT
accesskey = parsed.get("accesskey")  # value is XXXXKEY
```

gist: https://gist.github.com/swathipil/0640998161b2b9fb41dee053e16a770c


sdk/core/azure-core/CHANGELOG.md
```diff
@@ -23,6 +24,7 @@ This version will be the last version to officially support Python 3.5, future v
 ### Bug Fixes
 
 - Disable retry in stream downloading. (thanks to @jochen-ott-by @hoffmann for the contribution)  #16723
+>>>>>>> master
```
xiangyan99, line 27: ?

swathipil: fixing now

sdk/core/azure-core/azure/core/utils/__init__.py
```diff
@@ -0,0 +1,30 @@
...
+#
+# THE SOFTWARE IS PROVIDED *AS IS*, WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+# FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+# IN THE SOFTWARE.
+#
+# --------------------------------------------------------------------------
```
annatisch, line 25: Could we add a module-level docstring indicating that this module is intended for developers building on top of `azure-core`? If we are explicit about who we expect to use this module, we will hopefully avoid dumping any random functions in here.

sdk/core/azure-core/azure/core/utils/_connection_string_parser.py
```diff
@@ -0,0 +1,45 @@
...
+    if not case_sensitive_keys:
+        # if duplicate case insensitive keys are passed in, raise error
+        duplicate_keys = set()
+        for key in args_dict.keys():
+            if key.lower() in duplicate_keys:
+                raise ValueError(
+                    "Duplicate key in connection string: {}".format(key.lower())
+                )
+            duplicate_keys.add(key.lower())
+        args_dict = _case_insensitive_dict(**args_dict)
```
annatisch, line 43: An easy, transport and py2/3 independent way to get case insensitivity here is to simply return a lowercase-key dict - and document that behaviour so that consumers of this method know to retrieve keys as lower case:
```
if not case_sensitive_keys:
        # if duplicate case insensitive keys are passed in, raise error
        new_args_dict = {}
        duplicate_keys = set()
        for key in args_dict.keys():
            new_key = key.lower()
            if new_key in new_args_dict:
                raise ValueError(
                    "Duplicate key in connection string: {}".format(new_key)
                )
            new_args_dict[new_key] = args_dict[key]
       
        return new_args_dict
```
This isn't as fancy as the requests implementation - but at least will be consistent across all environments

sdk/core/azure-core/azure/core/utils/__init__.py
swathipil: yes!

sdk/core/azure-core/azure/core/utils/_connection_string_parser.py
swathipil: I was raising an error for duplicate keys earlier so that CIMultiDict and CaseInsensitiveDict would behave the same way, rather than one throwing error and the other replacing the key. In this case, would it be better to just allow replacement?

sdk/core/azure-core/azure/core/utils/_connection_string_parser.py
```diff
@@ -0,0 +1,47 @@
...
+def parse_connection_string(conn_str, case_sensitive_keys=False):
+    # type: (str, bool) -> Mapping[str, str]
+    """Parses the connection string into a dict of its component parts, with the option of preserving case
+    of keys, and validates that each key in the connection string has a provided value. If case of keys
+    is not preserved (ie. `case_sensitive_keys=False`), then a dict with LOWERCASE KEYS will be returned.
+    :param conn_str: String with connection details provided by Azure services.
+    :type conn_str: str
+    :param case_sensitive_keys: Indicates whether returned object should retrieve
+        keys with case sensitive checking. True means that case of keys will be preserved.
+        Default is False.
```
annatisch, line 18: I would rephrase this:
```
Indicates whether the casing of the keys will be preserved. When `False` (the default), all keys will be lower-cased. If set to `True`, the original casing of the keys will be preserved.
```

sdk/core/azure-core/azure/core/utils/_connection_string_parser.py
```diff
@@ -0,0 +1,47 @@
...
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+# --------------------------------------------------------------------------
+from typing import Mapping
+
+
+def parse_connection_string(conn_str, case_sensitive_keys=False):
+    # type: (str, bool) -> Mapping[str, str]
+    """Parses the connection string into a dict of its component parts, with the option of preserving case
```
annatisch, line 11: For some reason the documentation for this doesn't seem to be rendering?

swathipil: 🤔 looking into it!

swathipil: I copied the `_connection_string_parser.py` file to `azure/core` and imported `parse_connection_string` in `azure/core/__init__.py`. The change showed up in the docs, so I don't think `_connection_string_parser.py` has errors that would cause rendering failure.

I also copied the `azure/core/tracing` folder to a `azure/core/tracing_copy` folder and assumed that if `tracing` is working, then `tracing_copy` should also render. This didn't work either, so I believe this might not be a problem with this code. Thoughts, @annatisch ?

Issue #12652, opened 2020-07-21
sending amqp annotated message of different AMQP Body types design and implementation


KieranBrantnerMagee:
API Design:

https://gist.github.com/yunhaoling/8f5a632c4ba4c105b86b255360bb6387#gistcomment-3678799

See history [here](https://github.com/Azure/azure-sdk-for-net/issues/11143) for .net's version of this discussion

We have the same consideration in terms of how to incorporate AMQP body type variants into our design.

@yunhaoling kindly wrote up proposals for our options, to be found [here](https://gist.github.com/yunhaoling/8f5a632c4ba4c105b86b255360bb6387).

This issue tracks work to align on a proposal/get architectural buyin.

yunhaoling:
following up item:

check with the JS team to see if we want to utilize the internal class `AmqpAnnotatedMessage` for sending amqp message with different amqp types

yunhaoling:
according to the amqp spec:
- The body consists of either: one or more data sections, one or more amqp-sequence sections,
or a single amqp-value section.
![image](https://user-images.githubusercontent.com/47871814/112016752-3caec880-8aea-11eb-9915-e1ad67a17a0f.png)

- A sequence section contains an arbitrary number of structured data elements.
![image](https://user-images.githubusercontent.com/47871814/112016671-299bf880-8aea-11eb-946b-74db7c6b078a.png)

question: depending on our current uamqp implementation, wouldn't introducing sequence be a breaking change?
https://github.com/Azure/azure-uamqp-python/blob/master/uamqp/message.py#L101-L111

PR #17650, opened 2021-03-29
[ServiceBus] Support for sending AMQP Annotated Message

yunhaoling:
addressing issue: https://github.com/Azure/azure-sdk-for-python/issues/12652

pending uamqp 1.3.0 release


sdk/servicebus/azure-servicebus/azure/servicebus/_common/message.py
```diff
@@ -860,12 +882,148 @@ def locked_until_utc(self):
...
+        if len(validation) != 1:
+            raise ValueError(
+                "There should be one and only one of either data_body, sequence_body "
+                "or value_body being set as the body of the AMQPAnnotatedMessage."
+            )
+
+        header = kwargs.get("header")
+        footer = kwargs.get("footer")
+        properties = kwargs.get("properties")
+        application_properties = kwargs.pop("application_properties")
```
swathipil, line 962: was this meant to be a `get`? otherwise, I believe this will result in an error if `application_properties` isn't included

sdk/servicebus/azure-servicebus/tests/test_queues.py
```diff
@@ -2374,4 +2376,90 @@ def test_queue_send_dict_messages_scheduled_error_badly_formatted_dicts(self, se
...
+            with pytest.raises(ValueError):
+                AMQPAnnotatedMessage()
+
+            recv_data_msg = recv_sequence_msg = recv_value_msg = 0
+            with sb_client.get_queue_receiver(servicebus_queue.name, max_wait_time=10) as receiver:
+                with sb_client.get_queue_sender(servicebus_queue.name) as sender:
+                    batch = sender.create_message_batch()
+                    batch.add_message(data_message)
+                    batch.add_message(value_message)
+                    batch.add_message(sequence_message)
```
swathipil, line 2434: since the `create_from_messages_if_dict` now also accepts `messages: List[Union[Mapping, ServiceBusMessage, AMQPAnnotatedMessage]]`, maybe you could add a `dict`/`ServiceBusMessage` to check that a mixed list sends fine here and in async?
(may not really be needed, but just putting it out there)

swathipil (commented):
left a couple small comments

sdk/servicebus/azure-servicebus/tests/test_queues.py
yunhaoling: that's a good point, I'll cover it

sdk/servicebus/azure-servicebus/azure/servicebus/_common/message.py
yunhaoling: thanks! fixed

sdk/servicebus/azure-servicebus/azure/servicebus/_common/message.py
```diff
@@ -315,13 +317,31 @@ def scheduled_enqueue_time_utc(self, value):
...
-
-        :rtype: bytes or Iterable[bytes]
+        # type: () -> Optional[Union[bytes, Iterable[bytes], List, Iterable[List], Any]]
+        """The body of the Message. The format may vary depending
+        on the body type:
+        For ~azure.servicebus.AMQPMessageBodyType.DATA, the body could be bytes or Iterable[bytes]
+        For ~azure.servicebus.AMQPMessageBodyType.SEQUENCE, the body could be List or Iterable[List]
+        For ~azure.servicebus.AMQPMessageBodyType.VALUE, the body could be any type.
+
+        :rtype: Optional[Union[bytes, Iterable[bytes], List, Iterable[List], Any]]
```
rakshith91, line 327: will this generate properly in docstring?
afaik, it should be 
```suggestion
        :rtype: Optional[Union[bytes, Iterable[bytes], List, Iterable[List], Any]]
        :rtype: bytes or Iterable[bytes] or list or Iterable[list]
```

also, why everything else if there's Any in rtype

annatisch: hmm I do wonder if we should consider this breaking....
Maybe not? I mean, if the sender started sending a different body type that the receiver wasn't prepared for - the receiving code would likely be broken anyway....

sdk/servicebus/azure-servicebus/azure/servicebus/_common/message.py
```diff
@@ -559,31 +579,33 @@ def size_in_bytes(self):
...
         message = create_messages_from_dicts_if_needed(add_message, ServiceBusMessage)
         message = transform_messages_to_sendable_if_needed(message)
         message = cast(ServiceBusMessage, message)
         trace_message(
             message, parent_span
         )  # parent_span is e.g. if built as part of a send operation.
-        message_size = message.message.get_message_encoded_size()
+        message_size = (
+            message.message.get_message_encoded_size()
+        )  # pylint: disable=protected-access
```
annatisch, line 608: Why was 'disable=protected-access' needed here?

sdk/servicebus/azure-servicebus/azure/servicebus/_common/utils.py
```diff
@@ -215,18 +224,22 @@ def transform_messages_to_sendable_if_needed(messages):
...
 
-def _single_message_from_dict(message, message_type):
-    # type: (Union[ServiceBusMessage, Mapping[str, Any]], Type[ServiceBusMessage]) -> ServiceBusMessage
+def _convert_to_single_service_bus_message(message, message_type):
+    # type: (SingleMessageType, Type[ServiceBusMessage]) -> ServiceBusMessage
+    try:
+        # Handle AMQPAnnotatedMessage
+        message = message._to_service_bus_message()  # pylint: disable=protected-access
+    except AttributeError:
+        message = message
```
rakshith91, line 233:
```suggestion
        pass
```

sdk/servicebus/azure-servicebus/azure/servicebus/_common/message.py
```diff
@@ -860,12 +882,148 @@ def locked_until_utc(self):
...
+    access to low-level AMQP message sections.
+    Please refer to the AMQP spec:
+    http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-messaging-v1.0-os.html#section-message-format
+    for more information on the message format.
+
+    :keyword data_body: The body consists of one or more data sections and each section contains opaque binary data.
+    :paramtype data_body: Union[str, bytes, List[Union[str, bytes]]]
+    :keyword sequence_body: The body consists of one or more sequence sections and
+     each section contains an arbitrary number of structured data elements.
+    :paramtype sequence_body: Union[List[Any], List[List[Any]]]]
```
annatisch, line 895: Technically the 2nd of these types is already included in the first.

sdk/servicebus/azure-servicebus/samples/async_samples/send_and_receive_amqp_annotated_message_async.py
```diff
@@ -0,0 +1,96 @@
...
+from azure.servicebus.aio import ServiceBusClient
+
+CONNECTION_STR = os.environ['SERVICE_BUS_CONNECTION_STR']
+QUEUE_NAME = os.environ["SERVICE_BUS_QUEUE_NAME"]
+
+
+async def send_data_message(sender):
+    data_body = [b'aa', b'bb', b'cc']
+    data_app_prop = {"body_type": "data"}
+    del_anno = {"delann_key": "delann_value"}
```
annatisch, line 26: Can we just name it "delivery_annotations" 
`delivery_annotations = {"annotation key": "value"}`

sdk/servicebus/azure-servicebus/azure/servicebus/_common/message.py
yunhaoling: this is a good question as we're expanding the returning types.

I agree, it would be broken anyway... -- I'm thinking of other options like raising error if body is not of data body type and introducing new apis for getting sequence and value, but still being breaking, users have to update their receiving part to handle different body types.

torn on this one 😣

let's keep the design for now and iterate over it in next sprint? or do you think it's risky to have this shipped?

rakshith91: i'm fine with iterating for the next sprint

sdk/servicebus/azure-servicebus/azure/servicebus/_common/utils.py
yunhaoling: updated, thanks!

sdk/servicebus/azure-servicebus/azure/servicebus/_common/message.py
yunhaoling: updated :)

sdk/servicebus/azure-servicebus/samples/async_samples/send_and_receive_amqp_annotated_message_async.py
yunhaoling: fixed!

sdk/servicebus/azure-servicebus/azure/servicebus/_common/message.py
yunhaoling: it's no needed, this was previously me trying to add some new apis.
fixed

sdk/servicebus/azure-servicebus/azure/servicebus/_common/message.py
yunhaoling: updated the return type to be Any and created an issue to keep the discussion: https://github.com/Azure/azure-sdk-for-python/issues/17786

sdk/servicebus/tests.yml
```diff
@@ -4,6 +4,7 @@ stages:
   - template: ../../eng/pipelines/templates/stages/archetype-sdk-tests.yml
     parameters:
       ServiceDirectory: servicebus
+      InjectedPackages: $(WheelLocationsList)
```
yunhaoling, line 7:
```suggestion
```

annatisch (approved):

PR #17667, opened 2021-03-30
 Use base platform matrix for CI tests

benbp:
This PR updates the CI build definition to use the shared [platform matrix config](https://github.com/Azure/azure-sdk-for-python/blob/df8b520cca63946bae18b4a2494848b3d57044c7/eng/pipelines/templates/stages/platform-matrix.json), along with using the dynamically generated matrix approach for ci tests. The CI and live tests are now driven off the same base test matrix.

The overall change is minimal, but looks larger since I deleted some relics and renamed+moved the ci test job out into its own file (since it gets included as a template multiple times by the generate job, which we don't want to do for build/analyze).

The new common CI test matrix is slightly different from the current one, the only major change being I changed 3.5 to 3.6:

```
- Ubuntu 3.7
- Ubuntu 3.8
- Ubuntu 3.9
- Ubuntu pypy3
- MacOS 2.7
- Windows 3.6 + code coverage
```


benbp:
@lmazuel @johanste it was recommended I add you as you may have an interest in test matrix related topics.

scbedd (approved):
LGTM. I don't have an issues with the changes. We don't negatively impact our coverage.

benbp:
@scbedd I'm dealing with some edge cases with the one-off cosmos CI pipeline, so I need to reconcile that before merge.

eng/pipelines/templates/stages/platform-matrix.json
```diff
@@ -1,20 +1,31 @@
...
-          "ubuntu-18.04": { "OSVmImage": "MMSUbuntu18.04", "Pool": "azsdk-pool-mms-ubuntu-1804-general" }
+        "macOS-10.15": { "OSVmImage": "macOS-10.15", "Pool": "Azure Pipelines" }
       },
-      "PythonVersion": "3.9",
-      "CoverageArg": ""
+      "PythonVersion": "2.7",
+      "CoverageArg": "--disablecov"
+    },
+    {
+      "CoverageConfig": {
```
weshaggard, line 21: Correct me if I"m wrong but why are we limiting the windows to only the code coverage configuration?

eng/pipelines/templates/stages/pr.platform-matrix.json
```diff
@@ -0,0 +1,10 @@
+{
+  "matrix": {
+      "$IMPORT": "eng/pipelines/templates/stages/platform-matrix.json"
+  },
+  "exclude": [
```
weshaggard, line 5: Is the goal to only run on the latest 3.7, 3.8, and 3.9 versions of python here? If so it might be worth naming the file as such or adding a comment to that affect.

eng/pipelines/templates/stages/platform-matrix.json
```diff
@@ -1,20 +1,31 @@
...
   "matrix": {
     "Agent": {
-      "ubuntu-18.04": { "OSVmImage": "MMSUbuntu18.04", "Pool": "azsdk-pool-mms-ubuntu-1804-general" },
-      "windows-2019": { "OSVmImage": "MMS2019", "Pool": "azsdk-pool-mms-win-2019-general" },
-      "macOS-10.15": { "OSVmImage": "macOS-10.15", "Pool": "Azure Pipelines" }
+      "ubuntu-18.04": { "OSVmImage": "MMSUbuntu18.04", "Pool": "azsdk-pool-mms-ubuntu-1804-general" }
     },
-    "PythonVersion": [ "3.5", "3.7", "2.7", "pypy3" ],
-    "CoverageArg": [ "--disablecov" ]
+    "PythonVersion": [ "3.7", "3.8", "3.9", "pypy3" ],
```
weshaggard, line 9: We don't care about linux 2.7 support?

weshaggard: It kind of feels like you are trying to manually create a sparse matrix.

sdk/identity/ci.yml
```diff
@@ -31,16 +31,14 @@ extends:
...
-        CoverageArg: ''
-        RunForPR: false
-        InjectedPackages: 'git+https://github.com/AzureAD/microsoft-authentication-library-for-python@dev'
-
 
     Artifacts:
     - name: azure-identity
       safeName: azureidentity
+
+    ${{ if ne(variables['Build.Reason'], 'PullRequest') }}:
```
weshaggard, line 39: Does this mean we also don't want to be able to trigger this job via a comment?

eng/pipelines/templates/stages/archetype-sdk-client.yml
```diff
@@ -23,23 +23,40 @@ parameters:
...
   type: string
   default: azure-*
+- name: MatrixConfigs
+  type: object
+  default:
+    - Name: Python_ci_test_base
+      Path: eng/pipelines/templates/stages/platform-matrix.json
+      Selection: sparse
+      GenerateVMJobs: true
+- name: PRMatrixConfigs
```
weshaggard, line 39: The entire reason they had the PRMatrix configs before was to have a way to create a sparse matrix for PR validation. Given we now can do that with the new format do we really need this?

eng/pipelines/templates/stages/platform-matrix.json
benbp: With the way things have been set up in all the configs across the repos, code coverage appears on a single platform. I'm not sure what the underlying reason for this is, perhaps @seankane-msft has some context here.

eng/pipelines/templates/stages/platform-matrix.json
benbp: I'll add 2.7 linux back in. The way python had things set up was testing the majority of versions against Linux, and then adding one case each to cover windows and mac, so that's how I have the matrix config set up here.

sdk/identity/ci.yml
benbp: Correct, this matrix entry is currently set up in master to not run for PR.

eng/pipelines/templates/stages/archetype-sdk-client.yml
benbp: The current setup isn't a sparse matrix as we treat it, because sparse matrix means "make sure every parameter in the config appears at least once" whereas the behavior in master is to skip testing half the python versions.

eng/pipelines/templates/stages/platform-matrix.json
seankane-msft: Code coverage can only happen once or it will overwrite itself each time, we chose the Python 3.9 tests to run code coverage.

eng/pipelines/templates/stages/pr.platform-matrix.json
benbp: I changed it slightly to more closely mirror the current configuration. My assumption of the intent is just to run fewer versions in PR, so now it will run 2.7, 3.8, 3.9 on linux and 3.6+coverage on windows.

eng/pipelines/templates/stages/platform-matrix.json
benbp: This switches coverage to be in windows 3.6, do you foresee any issues with that?

eng/pipelines/templates/stages/archetype-sdk-client.yml
weshaggard: I still kind of feel this pattern is going against what we have in the other languages. If we feel we want a different matrix for PR vs not we should make that a common pattern across the languages. Currently if we can avoid it I'd like to not have a unique configuration for PR but instead just use a sparse version of the full matrix.

benbp: I think this approach represents a sparse version of the full matrix achieved with the tools we have available (sparseness operating at the parameter key boundary). The alternatives I can see are defining an independent matrix, defining matrix filter defaults in the yaml, or expanding the version list to use groupings containing a `RunForPR` variable. I think having a generic way to say "pick X sparse subset of Y key" is going to be problematic, because in most scenarios there will be a priority list of which X values are most important (whereas the current sparseness always gives an "everything is covered" guarantee). This method is essentially "pick X values from a shared base config."

Another alternative is to remove the PR vs. master split and just run all the tests on PR.

benbp: All that said, we could introduce a `limit` option to the sparse generator, which would extract N elements from an already generated sparse matrix (with indexes spread out across the length of that array).

eng/pipelines/templates/stages/platform-matrix.json
```diff
@@ -1,20 +1,31 @@
 {
+  "displayNames": {
+    "--disablecov": ""
+  },
   "matrix": {
     "Agent": {
-      "ubuntu-18.04": { "OSVmImage": "MMSUbuntu18.04", "Pool": "azsdk-pool-mms-ubuntu-1804-general" },
```
weshaggard, line 4: Based on our discussion yesterday I assume you still plan to add back the other OS's here to the base matrix as well, correct?

PR #17671, opened 2021-03-30
[DataLake][Fix]only set recursive=True for delete directory

xiafu-msft:
this PR is ensuring that recursive setting only applies to directories rather than directories and files


tasherif-msft:
So for future reference - this PR is ensuring that `recursive` setting only applies to directories rather than directories and files.

tasherif-msft (approved):

PR #17680, opened 2021-03-30
[Communications-Chat] Refactor identifiers

annatisch:


sdk/communication/azure-communication-chat/azure/communication/chat/_communication_identifier_serializer.py
```diff
@@ -3,130 +3,103 @@
...
-        if has_attributes(communicationIdentifier, ["raw_id"]):
-            return _IdentifierType.UNKNOWN_IDENTIFIER
+        return CommunicationIdentifierModel(
+            raw_id=identifier.id
+        )
+    except AttributeError:
+        raise TypeError("Unsupported identifier type " + identifier.__class__.__name__)
+
+
+def deserialize_identifier(identifier_model):
```
juancamilor, line 71: In other sdks (.NET i.e), the serialization/deserialization logic is auto generated from the swagger (autorest). 
I guess this needs to be hand-crafted in python ? 
Would not make sense to have this in a common project so other modalities ca reuse it ?

juancamilor (approved):

sdk/communication/azure-communication-chat/azure/communication/chat/_communication_identifier_serializer.py
annatisch: I didn't change this - just refactored it a bit.
Currently the response deserialization _is_ taking place in the autorest generated code - however the object returned (as represented by the swagger) is not the object we want users to interact with (as it's somewhat unwieldy) - so we then further deserialize that single object into these discrete objects.

sdk/communication/azure-communication-chat/azure/communication/chat/_shared/models.py
```diff
@@ -6,210 +6,150 @@
...
-    _validation = {
-        'error': {'required': True},
-    }
 
-    _attribute_map = {
-        'error': {'key': 'error', 'type': 'CommunicationError'},
-    }
+CommunicationUserProperties = TypedDict(
+    'CommunicationUserProperties',
+    identifier=str
```
DominikMe, line 53: We can name this property `id` now once we rename the other one to `raw_id`

DominikMe (commented):
Thanks so much for helping with this Anna!

First batch of comments, I'll keep reviewing.

sdk/communication/azure-communication-chat/azure/communication/chat/_chat_thread_client.py
```diff
@@ -547,14 +545,14 @@ def add_participants(
     @distributed_trace
     def remove_participant(
         self,
-        user,  # type: CommunicationUserIdentifier
+        user,  # type: CommunicationIdentifier
```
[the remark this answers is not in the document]
annatisch: In this case we eventually want to support a union of type here - either an Identifier, or a ChatParticipant.

I would prefer that we go with `participant`, but then I would also need to add the support for taking `ChatParticipant` as input here. Which I'm happy to do - because this is what I wanted.
But would need your teams approval ;)

sdk/communication/azure-communication-chat/azure/communication/chat/_shared/models.py
```diff
@@ -6,210 +6,150 @@
...
-
-    All required parameters must be populated in order to send to Azure.
-
-    :param error: Required. The Communication Services error.
-    :type error: ~communication.models.CommunicationError
+class CommunicationIdentifier(Protocol):
     """
+    Communication Identifier.
+
+    :ivar str id: The identifier.
```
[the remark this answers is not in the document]
annatisch: I've renamed to raw_id.
I would prefer to keep it on the Protocol - as it's the fundamental mechanism for these objects being somewhat "future-proof". It also seems that it would be necessary on `UnknownIdentifier`....

sdk/communication/azure-communication-chat/azure/communication/chat/_chat_thread_client.py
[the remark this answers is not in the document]
juancamilor: I would prefer if we don't diverge from the interface we agreed upon across sdks. It feels also like creating 2 nested objects just to be able to remove the user is a bit cumbersome.

sdk/communication/azure-communication-chat/azure/communication/chat/_communication_identifier_serializer.py
```diff
@@ -3,130 +3,102 @@
...
+    if not raw_id:
+        raise ValueError("Identifier must have a valid ID")
+
+    _assert_maximum_one_nested_model(identifier_model)
+
+    if identifier_model.communication_user:
+        return CommunicationUserIdentifier(raw_id, raw_id=raw_id)
+    if identifier_model.phone_number:
+        if not identifier_model.phone_number:
+            raise ValueError("PhoneNumberIdentifier must have a valid attribute - phone_number")
```
johanste, line 89: This is only called as part of deserialization of responses, correct? If so, we do not want to validate service responses. Service responses.

sdk/communication/azure-communication-chat/azure/communication/chat/_shared/models.py
```diff
@@ -6,210 +6,151 @@
...
-
-class CommunicationIdentifierKind(with_metaclass(_CaseInsensitiveEnumMeta, str, Enum)):
-    """Communication Identifier Kind.
-    """
-    Unknown = "UNKNOWN"
-    CommunicationUser = "COMMUNICATIONUSER"
-    PhoneNumber = "PHONENUMBER"
-    CallingApplication = "CALLINGAPPLICATION"
-    MicrosoftTeamsUser = "MICROSOFTTEAMSUSER"
+MicrosoftTeamsUserProperties = TypedDict(
```
johanste, line 120: Should this have `total=False`?

annatisch: I believe all the properties in the MicrosoftTeamsUserIdentifier are required - can you confirm @DominikMe?

sdk/communication/azure-communication-chat/azure/communication/chat/_communication_identifier_serializer.py
annatisch: Yeah I was thinking this while I was refactoring the code.... maybe I will take it out while I'm removing things :)

sdk/communication/azure-communication-chat/azure/communication/chat/_shared/models.py
johanste: Both as inputs and outputs?

sdk/communication/azure-communication-chat/azure/communication/chat/_chat_thread_client.py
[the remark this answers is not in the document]
DominikMe: @juancamilor So you agree that we should rename to `identifier` here? That's what the APIViews you linked to are doing (correctly).

sdk/communication/azure-communication-chat/azure/communication/chat/_communication_identifier_serializer.py
DominikMe: We have this validation in the other languages. Where would you want the code to throw if we're not validating the service response here? I get the sentiment, but I like that our SDKs actually help validating the service in live tests. This is how we were able to discover multiple bugs during development.

@RezaJooyandeh

sdk/communication/azure-communication-chat/azure/communication/chat/_shared/models.py
DominikMe: They all have defined default values which get set inside the `MicrosoftTeamsUserIdentifier` constructor and the user doesn't instantiate the `MicrosoftTeamsUserProperties` dict themselves.

So they are optional on the identifier but then filled in the properties dictionary, this design seems reasonable to me.

If the service starts adding additional properties they will be ignored by the client on deserialization, which is good. If the service removes properties, we should throw because that's breaking. (see other thread about service response validation)

sdk/communication/azure-communication-chat/azure/communication/chat/aio/_chat_thread_client_async.py
```diff
@@ -534,13 +531,13 @@ async def add_participants(
     @distributed_trace_async
     async def remove_participant(
         self,
-        user: CommunicationUserIdentifier,
+        user: CommunicationIdentifier,
```
DominikMe, line 534: This argument also needs to be renamed to `identifier` @juancamilor @sarkar-rajarshi

sdk/communication/azure-communication-identity/azure/communication/identity/_communication_identity_client.py
```diff
@@ -13,6 +15,12 @@
 from ._shared.models import CommunicationUserIdentifier
```
DominikMe, line 15: Can `import CommunicationUserIdentifier` be removed now?

sdk/communication/azure-communication-identity/azure/communication/identity/_communication_identity_client.py
```diff
@@ -78,75 +85,73 @@ def from_connection_string(
...
+            scopes, # type: List[Union[str, CommunicationTokenScope]]
             **kwargs # type: Any
         ):
-        # type: (...) -> Tuple[CommunicationUserIdentifier, AccessToken]
-        """create a single Communication user with an identity token.
-        :param scopes:
-            List of scopes to be added to the token.
-        :type scopes: list[str or ~azure.communication.identity.models.CommunicationTokenScope]
-        :return: A tuple of a CommunicationUserIdentifier and a AccessToken.
+        # type: (...) -> Tuple[CommunicationIdentifier, AccessToken]
```
DominikMe, line 104: This will always return a `CommunicationUserIdentifier`. Are we generalizing to `CommunicationIdentifier` here for cross-library is instance checks?

sdk/communication/azure-communication-chat/azure/communication/chat/_communication_identifier_serializer.py
```diff
@@ -3,130 +3,59 @@
...
+    :type identifier_model: CommunicationIdentifierModel
+    :return: CommunicationIdentifier
+    """
+    raw_id = identifier_model.raw_id
+
+    if identifier_model.communication_user:
+        return CommunicationUserIdentifier(raw_id, raw_id=raw_id)
+    if identifier_model.phone_number:
+        return PhoneNumberIdentifier(identifier_model.phone_number.value, raw_id=raw_id)
+    if identifier_model.microsoft_teams_user is not None:
```
annatisch, line 54: Oops - I can also get rid of the `is not None`...

sdk/communication/azure-communication-identity/azure/communication/identity/_communication_identity_client.py
```diff
@@ -78,75 +85,73 @@ def from_connection_string(
...
+            cls=lambda pr, u, e: (CommunicationUserIdentifier(u.identity.id, raw_id=u.identity.id),
                 AccessToken(u.access_token.token, u.access_token.expires_on)),
             create_token_with_scopes=scopes,
             **kwargs)
 
     @distributed_trace
     def delete_user(
             self,
-            communication_user, # type: CommunicationUserIdentifier
+            user, # type: CommunicationUserIdentifier
```
DominikMe, line 122: Now that we're only generalizing the create return type, will this prevent me from calling `create_user` and then passing the returned value into `delete_user` or `get_token`?

sdk/communication/azure-communication-identity/azure/communication/identity/aio/_communication_identity_client_async.py
```diff
@@ -78,87 +80,82 @@ def from_connection_string(
...
     @distributed_trace_async
     async def create_user_and_token(
             self,
-            scopes, # type: List[Union[str, "_model.CommunicationTokenScope"]]
-            **kwargs # type: Any
-        ):
-        # type: (...) -> Tuple[CommunicationUserIdentifier, AccessToken]
+            scopes: List[Union[str, 'CommunicationTokenScope']],
+            **kwargs
+        ) -> Tuple['CommunicationIdentifier', AccessToken]:
```
DominikMe, line 98: Same questions around the contravariant change as in the sync client

sdk/communication/azure-communication-identity/azure/communication/identity/_communication_identity_client.py
annatisch: We are generalizing here to ensure future compatibility if this library ever expands to support other Identifier types. If we decide it never will, we can always make the return type narrower later :)

sdk/communication/azure-communication-identity/azure/communication/identity/_communication_identity_client.py
annatisch: Nope - it would be non-breaking to expand an input to accept multiple types, but it _would_ be breaking to expand an output type to be multiple types. So we can update the input type if these APIs evolve in future.

DominikMe: That makes sense. What I was wondering though is why I can pass a `CommunicationIdentifier` protocol for a `CommunicationUserIdentifier` argument. Isn't the user type narrower than the identifier protocol and would reject it?

annatisch: Right now you cannot pass a CommunicationIdentifier protocol as an argument - the implementation is written to only work with CommunicationUserIdentifiers. So currently the input type is accurate - only that specific type is supported. I'm assuming that if the service starts supporting other Identifier types in this library it will require an SDK release - to support that, at which point we would evaluate the type hints.

As an aside - none of this prevents runtime usage. These typehints are just there for the IDE experience and static analysis.

sdk/communication/azure-communication-identity/azure/communication/identity/_communication_identity_client.py
RezaJooyandeh: This method will always return a `CommunicationUserIdentifier` that is why we made it more specific and called it `create_user`. Future potential types would be added as separate methods. In other languages also for this method we return `CommunicationUser`.

DominikMe (approved):

PR #17696, opened 2021-03-31
[Identity] Fix `padding_needed` for base64-decoding claims

jiasli:
When `len(encoded_claims)` is a multiple of 4, this line returns 4, thus appending extra `====` to the string.

This fix is copied from MSAL (+@rayluo for awareness)

https://github.com/AzureAD/microsoft-authentication-library-for-python/blob/1.10.0/msal/oauth2cli/oidc.py#L23

```py
raw += '=' * (-len(raw) % 4)  # https://stackoverflow.com/a/32517907/728675
```


rayluo (approved):
LGTM

chlowell (approved):

PR #17707, opened 2021-03-31
[KeyVault] tests-mgmt.yml

swathipil:
addressing: #17249


sdk/keyvault/tests-mgmt.yml
```diff
@@ -0,0 +1,15 @@
+trigger: none
+
+stages:
+  - template: ../../eng/pipelines/templates/stages/archetype-sdk-tests.yml
```
benbp, line 4: I'm trying to have use use consistent absolute paths for template includes in our configs, if you don't mind updating to:

```
```

benbp:
For the weekly schedule, we currently have the concept of a `tests-weekly` pipeline. Right now it's only rolled out for java and net, but I can create an instance of one for these as well, if that would work for your needs.

sdk/keyvault/tests-mgmt.yml
swathipil: will do!

benbp (approved):

PR #17709, opened 2021-03-31
[formrecognizer] update test-resources.json

kristapratico:
By the time our live tests get to the sample tests, these sas tokens are sometimes expired. Adding an hour to the signed expiry will help us see more green runs.


catalinaperalta (approved):
LGTM! Not sure how many times it goes over 2 hours, but should help in those cases! :)

Issue #17516, opened 2021-03-23
Test Our Libraries WITHOUT aiohttp


scbedd:
All of the environments as they are now either 

A) Install your package. Install `eng/test_tools.txt` Install `dev_requirements.txt`. Run tests.
B) Install your package. Install `eng/test_tools.txt`. Install a specific version of each `install_requires` from your package.

The issue here. Within `eng/test_tools.txt`, we _always_ install `aiohttp`. This is can _hide issues_ with our import logic.

We need to run tests in an environment _without_ AIOHTTP and ensure that `depends` and `whl`/`sdist` all pass without `aiohttp`.

seankane-msft:
I think we can close this now. We can re-open if I am mistaken.

PR #17718, opened 2021-03-31
Add Test Environment that runs tests w/out aiohttp

scbedd:
@seankane-msft  

This is related to [this issue](https://github.com/Azure/azure-sdk-for-python/issues/17516).

Can you poke holes in my logic? We don't really have anything to key off to identify one of these "optional" requirements that may or may not be present. So it's not really a case of extending `create_package_and_install.py` to identify additional requirements. 

To solve this generically without messing with anyone's day-to-day development, I created a new tox environment `whl_no_aio`. This is essentially a retread of `whl` (aka the default) tox environment. The only difference is that have an error-ignored line (notice the leading `-`) that attempts uninstall of aiohttp prior to invoking any of the install or test commands.


eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -178,7 +178,7 @@ jobs:
     - pwsh: |
         $toxenvvar = "whl,sdist"
         if ('$(System.TeamProject)' -eq 'internal') {
-          $toxenvvar = "whl,sdist,depends,latestdependency,mindependency"
+          $toxenvvar = "whl,sdist,depends,latestdependency,mindependency,whlnoaiohttp"
```
scbedd, line 181:
```suggestion
          $toxenvvar = "whl,sdist,depends,latestdependency,mindependency,whl_no_aio"
```

eng/tox/tox.ini
```diff
@@ -79,6 +79,23 @@ commands =
...
+changedir = {toxinidir}
+deps =
+  {[base]deps}
+commands =
+    - {envbindir}/python -m pip uninstall aiohttp --yes
+    {envbindir}/python {toxinidir}/../../../eng/tox/create_package_and_install.py -d {envtmpdir} -p {toxinidir} -w {envtmpdir}
+    {envbindir}/python -m pip freeze
+    pytest \
+      {[testenv]default_pytest_params} \
+      --ignore-glob='*_async.py' \
```
scbedd, line 94: Not having this means that async tests implode with 

image

seankane-msft:
Within the `create_package_and_install.py` file, can we add a flag for this toxenv where we add a clause like the following to verify that the uninstallation was successful:
```python
try:
    import aiohttp
    raise Exception("aiohttp was not successfully installed") 
except ImportError:
    pass
```

scbedd:
I see what you're going after here. I'll add something like it.

scbedd:
@seankane-msft  added!

eng/tox/try_import.py
```diff
@@ -0,0 +1,52 @@
...
+# This script is used to execute bandit within a tox environment. Depending on which package is being executed against,
+# a failure may be suppressed.
+
+
+from subprocess import check_call, CalledProcessError
+import argparse
+import os
+import logging
+import sys
+import pdb
```
seankane-msft, line 17:
```suggestion
import argparse
import logging
```

seankane-msft: Nit: you don't need all these imports. Otherwise this is perfect.

seankane-msft (approved):
:shipit:

eng/tox/try_import.py
scbedd: Big oof. Totally agree. Cleaned up comment, applied `black`, didn't clean up imports. /facepalm

Issue #17043, opened 2021-03-03
[translation] Resolve namespace


kristapratico:
currently using azure.ai.documenttranslation

should be azure.ai.documenttranslation?
azure.ai.documenttranslator?

Additionally need to update the user-agent string in _user_agent.py once we settle on the namespace.

PR #17736, opened 2021-04-01
[translator] rename package/namespace

kristapratico:
Resolves https://github.com/Azure/azure-sdk-for-python/issues/17043

This does the rename from azure-ai-documenttranslation to azure-ai-translator

(rename pending on more discussion from architects)


maririos (approved):

maririos:
can you trigger the live tests too just in case?

kristapratico:
> can you trigger the live tests too just in case?

Yes, was planning to :)

kristapratico:
Decided not to go with this naming after architect discussion. New namespace/package naming in https://github.com/Azure/azure-sdk-for-python/pull/17792

PR #17720, opened 2021-03-31
set-test-pipeline-version.yml failing due to lack of setuptools

scbedd:
What I cannot understand is why this has been succeeding up till now. Agent change maybe?

PR #17737, opened 2021-04-01
Setuptools is no longer present in the default python install

scbedd:
This PR is quite similar to #17720 

[Example Failure](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=815176&view=logs&j=b6f7b02f-4c06-5d94-ea42-ae5d966a2c2f&t=e1a5ddc8-f799-5d0b-74cd-349aec78d3e2)

This should have been caught in #17720, but I did not actually trigger the release, as the issue I was resolving was in the `build` stage. Dumb, to say the least.

### DIFF
--- a/eng/scripts/Language-Settings.ps1
+++ b/eng/scripts/Language-Settings.ps1
@@ -18,7 +18,7 @@ function Get-AllPackageInfoFromRepo ($serviceDirectory)
   try
   {
     Push-Location $RepoRoot
-    pip install packaging==20.4 -q -I
+    pip install packaging==20.4 setuptools==44.1.1 -q -I
     $allPkgPropLines = python (Join-path eng scripts get_package_properties.py) -s $searchPath
   }
   catch


weshaggard (approved):

eng/scripts/Language-Settings.ps1
```diff
@@ -18,7 +18,7 @@ function Get-AllPackageInfoFromRepo ($serviceDirectory)
   try
   {
     Push-Location $RepoRoot
-    pip install packaging==20.4 -q -I
+    pip install packaging==20.4 setuptools==44.1.1 -q -I
```
weshaggard, line 21: We might want to start considering a requirements file for this if we start getting a longer list.

scbedd: Agree.

scbedd:
[Have a green template release](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=818445&view=results)

(the one red check is because it can't create against master for a fake commit :))

PR #17761, opened 2021-04-02
Adapt SB to latest OT changes

rakshith91:
part 1 of https://github.com/Azure/azure-sdk-for-python/issues/17739


sdk/servicebus/azure-servicebus/azure/servicebus/_common/utils.py
```diff
@@ -33,7 +33,7 @@
 from uamqp import authentication, types
 
 from azure.core.settings import settings
-from azure.core.tracing import SpanKind
+from azure.core.tracing import SpanKind, Link
```
yunhaoling, line 36: do we need to update the azure-core dependency?

yunhaoling (approved):

PR #17762, opened 2021-04-02
[Key Vault] Add local crypto support for AES-CBCPAD

### LINKED ISSUE #17425: [Key Vault] Add local crypto capabilities for new algorithms
Add local crypto capabilities for as many algorithms added in `azure-keyvault-keys`'s February release as possible.

mccoyp:
Resolves #17425, at least to achieve parity with Javascript.

This adds local crypto support for encryption and decryption with AES-CBCPAD algorithms. AES-CBC support is on hold until we find a way to perform zero-padding -- the closest thing in `cryptography` is [ANSIX923](https://cryptography.io/en/latest/hazmat/primitives/padding.html#cryptography.hazmat.primitives.padding.ANSIX923) padding, but the final byte in that padding is nonzero.

This also fixes a bug that I had accidentally introduced in async tests by removing preparers: async test cases weren't being awaited due to the preparer's absence, and green test runs greatly exaggerated their success. This decorates affected tests with a dummy `PowerShellPreparer` that gets them noticed by the test runner without affecting execution otherwise.

### DIFF
--- a/sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/crypto/_internal/algorithms/aes_cbc.py
+++ b/sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/crypto/_internal/algorithms/aes_cbc.py
@@ -10,6 +10,8 @@ from ..algorithm import SymmetricEncryptionAlgorithm
 from ..transform import BlockCryptoTransform
 
 
+# pylint: disable=W0223
+
 _CBC_BLOCK_SIZE = 128
 
 
@@ -20,16 +22,16 @@ class _AesCbcCryptoTransform(BlockCryptoTransform):
 
     def transform(self, data):
         return self.update(data) + self.finalize()
-    
+
     def block_size(self):
         return _CBC_BLOCK_SIZE
 
 
 class _AesCbcDecryptor(_AesCbcCryptoTransform):
-    def __init__(self, key, iv, padding):
+    def __init__(self, key, iv, padding_mode):
         super(_AesCbcDecryptor, self).__init__(key, iv)
         self._ctx = self._cipher.decryptor()
-        self._padder = padding.unpadder()
+        self._padder = padding_mode.unpadder()
 
     def update(self, data):
         decrypted = self._ctx.update(data) + self._ctx.finalize()
@@ -40,10 +42,10 @@ class _AesCbcDecryptor(_AesCbcCryptoTransform):
 
 
 class _AesCbcEncryptor(_AesCbcCryptoTransform):
-    def __init__(self, key, iv, padding):
+    def __init__(self, key, iv, padding_mode):
         super(_AesCbcEncryptor, self).__init__(key, iv)
         self._ctx = self._cipher.encryptor()
-        self._padder = padding.padder()
+        self._padder = padding_mode.padder()
 
     def update(self, data):
         padded = self._padder.update(data) + self._padder.finalize()
--- a/sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/crypto/_internal/algorithms/aes_cbc.py
+++ b/sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/crypto/_internal/algorithms/aes_cbc.py
@@ -26,24 +26,10 @@ class _AesCbcCryptoTransform(BlockCryptoTransform):
 
 
 class _AesCbcDecryptor(_AesCbcCryptoTransform):
-    def __init__(self, key, iv):
+    def __init__(self, key, iv, padding):
         super(_AesCbcDecryptor, self).__init__(key, iv)
         self._ctx = self._cipher.decryptor()
-        self._padder = padding.ANSIX923(_CBC_BLOCK_SIZE).unpadder()  # TODO: Update this to zero-padding
-
-    def update(self, data):
-        decrypted = self._ctx.update(data) + self._ctx.finalize()
-        return self._padder.update(decrypted)
-
-    def finalize(self):
-        return self._padder.finalize()
-
-
-class _AesCbcPadDecryptor(_AesCbcCryptoTransform):
-    def __init__(self, key, iv):
-        super(_AesCbcPadDecryptor, self).__init__(key, iv)
-        self._ctx = self._cipher.decryptor()
-        self._padder = padding.PKCS7(_CBC_BLOCK_SIZE).unpadder()
+        self._padder = padding.unpadder()
 
     def update(self, data):
         decrypted = self._ctx.update(data) + self._ctx.finalize()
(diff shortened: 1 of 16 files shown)


sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/crypto/_internal/symmetric_key.py
```diff
@@ -94,15 +101,17 @@ def default_encryption_algorithm(self):
     def default_key_wrap_algorithm(self):
         return _default_kw_alg_by_size[len(self._key)]
 
-    def encrypt(self, plain_text, iv, **kwargs):  # pylint:disable=arguments-differ
+    def encrypt(self, plain_text, **kwargs):  # pylint:disable=arguments-differ
         algorithm = self._get_algorithm("encrypt", **kwargs)
-        encryptor = algorithm.create_encryptor(key=self._key, iv=iv)
-        return encryptor.transform(plain_text, **kwargs)
+        raise_if_incorrect_key_size(algorithm, len(self._key))
+        encryptor = algorithm.create_encryptor(key=self._key, iv=kwargs.pop("iv", None))
```
chlowell, line 107: Won't this just raise when `iv` is None?

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/crypto/aio/_client.py
```diff
@@ -181,7 +181,7 @@ async def encrypt(self, algorithm: "EncryptionAlgorithm", plaintext: bytes, **kw
         if self._local_provider.supports(KeyOperation.encrypt, algorithm):
             raise_if_time_invalid(self._not_before, self._expires_on)
             try:
-                return self._local_provider.encrypt(algorithm, plaintext)
+                return self._local_provider.encrypt(algorithm, plaintext, iv=iv, aad=aad)
```
chlowell, line 184: If we need  `aad` only for AES-GCM, which we won't do locally, why pass it here?

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/crypto/_internal/algorithms/aes_cbc.py
```diff
@@ -12,60 +10,80 @@
...
-        pass
+        return _CBC_BLOCK_SIZE
 
 
 class _AesCbcDecryptor(_AesCbcCryptoTransform):
     def __init__(self, key, iv):
         super(_AesCbcDecryptor, self).__init__(key, iv)
         self._ctx = self._cipher.decryptor()
-        self._padder = padding.PKCS7(self.block_size).unpadder()
+        self._padder = padding.ANSIX923(_CBC_BLOCK_SIZE).unpadder()  # TODO: Update this to zero-padding
```
chlowell, line 32: If the current code doesn't do the right thing, and this change doesn't fix that, I wouldn't make the change

sdk/keyvault/azure-keyvault-keys/tests/test_crypto_client.py
```diff
@@ -237,6 +244,7 @@ def test_wrap_and_unwrap(self, **kwargs):
         self.assertEqual(key_bytes, result.key)
 
     def test_symmetric_encrypt_and_decrypt_mhsm(self, **kwargs):
+        """Encrypt and decrypt with the service"""
         is_hsm = True
         self._skip_if_not_configured(is_hsm)
         endpoint_url = self.managed_hsm_url if is_hsm else self.vault_url
```
chlowell, line 250: Looks like I missed this in an earlier PR 😿 
```suggestion
        self._skip_if_not_configured(True)
        key_client = self.create_key_client(self.managed_hsm_url)
```

sdk/keyvault/azure-keyvault-keys/tests/test_crypto_client.py
```diff
@@ -335,6 +350,71 @@ def test_encrypt_local_from_jwk(self, **kwargs):
...
             result = crypto_client.decrypt(result.algorithm, result.ciphertext)
             self.assertEqual(result.plaintext, self.plaintext)
+    
+    def test_symmetric_encrypt_local_mhsm(self, **kwargs):
+        """Encrypt locally, decrypt with the service"""
+        is_hsm = True
+        self._skip_if_not_configured(is_hsm)
+        endpoint_url = self.managed_hsm_url if is_hsm else self.vault_url
+
+        key_client = self.create_key_client(endpoint_url)
```
chlowell, line 360:
```suggestion
        self._skip_if_not_configured(True)
        key_client = self.create_key_client(self.managed_hsm_url)
```

sdk/keyvault/azure-keyvault-keys/tests/test_crypto_client.py
mccoyp: I also just noticed this while making these changes and forgot to fix it -- thanks for catching it!

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/crypto/_providers/local_provider.py
```diff
@@ -60,16 +60,17 @@ def _raise_if_unsupported(self, operation, algorithm):
...
             raise AzureError('This key does not allow the "{}" operation'.format(operation))
 
-    def encrypt(self, algorithm, plaintext):
-        # type: (EncryptionAlgorithm, bytes) -> EncryptResult
+    def encrypt(self, algorithm, plaintext, **kwargs):
+        # type: (EncryptionAlgorithm, bytes, **Any) -> EncryptResult
         self._raise_if_unsupported(KeyOperation.encrypt, algorithm)
-        ciphertext = self._internal_key.encrypt(plaintext, algorithm=algorithm.value)
-        return EncryptResult(key_id=self._key.kid, algorithm=algorithm, ciphertext=ciphertext)
+        iv = kwargs.pop("iv", None)
```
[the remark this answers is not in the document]
mccoyp: Good point! I had decided against doing so since it's only necessary for some symmetric algorithms, but after a second look it does make sense to make it an optional parameter.

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/crypto/aio/_client.py
mccoyp: I decided to pass it since we'll ideally use it later, but it does make sense to only pass what's needed for now

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/crypto/_internal/symmetric_key.py
mccoyp: An error will raise from a check in `aes_cbc.py`, yes. I didn't enforce that check here partially because there are some symmetric algorithms (AES-GCM) that don't accept an IV for encryption. This could be a case of planning for functionality we don't yet have, but I think it makes sense to validate that while validating the key and IV size. Is that what you were asking about?

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/crypto/_internal/symmetric_key.py
```diff
@@ -59,13 +66,13 @@ def __init__(self, kid=None, key_bytes=None, key_size=None):
...
+            supported_encryption_algorithms.append(Aes128CbcPad.name())
             supported_key_wrap_algorithms.append(AesKw128.name())
         if key_size >= key_size_192:
-            supported_encryption_algorithms.append(Aes192Cbc.name())
+            supported_encryption_algorithms.append(Aes192CbcPad.name())
             supported_key_wrap_algorithms.append(AesKw192.name())
         if key_size >= key_size_256:
-            supported_encryption_algorithms.append(Aes256Cbc.name())
+            supported_encryption_algorithms.append(Aes256CbcPad.name())
             supported_encryption_algorithms.append(Aes128CbcHmacSha256.name())
```
chlowell, line 76: This makes me nervous: we don't intend to support HMAC algorithms locally but may now do so in an undocumented way.

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/crypto/_internal/symmetric_key.py
chlowell: I thought it was odd to use a default value that always gets an error. If AES-GCM is the only such algorithm supported by Key Vault, I'd forego the flexibility because we won't ever do AES-GCM locally.

chlowell (approved):

Issue #16453, opened 2021-02-01
[servicebus] add repr() to ServiceBusMessage/ServiceBusReceivedMessage/ServiceBusBatchMessage


rakshith91:
latest updates:

- we decided to pause the support for `body_as_str` method until there're customer requests.
- The ServiceBusMessage (and other message model) should implement the `repr` method as per our guideline here: https://azure.github.io/azure-sdk/python_design.html#python-models-repr

------

right now, `msg.body` returns a generator since it's directly returning the data object from the uamqp message object.

It means to extract contents as a string I have to do `next(msg..body).decode('utf-8')` 

body_as_Str would be a nice convenience to add ; there's on in EH too

yunhaoling:
hey @rakshith91 , I had a brief discussion with anna on this topic, some feedback:

1. in python `str(message) ` could be used and is the preferred way in python to get a string of the object.
  - call `str()` on a ServiceBusMessage will get the str.
2. the value that the method `body_as_str` could provide is taking the `encoding` parameter to decode the bytes.

Question for you:

do you know if there's any user case for SB or EG that has the need to decode a ServiceBusMessage with an encoding different than `utf-8`?
If the built-in `str()` method could satisfy your need, then we could push this request until there're customer requests.

yunhaoling:
updates:

- we decided to pause the support for `body_as_str` method until there're customer requests.
- The ServiceBusMessage model should implement the `repr` method as per our guideline here: https://azure.github.io/azure-sdk/python_design.html#python-models-repr

PR #17808, opened 2021-04-05
Add repr to SB messages

rakshith91:
Fixes https://github.com/Azure/azure-sdk-for-python/issues/16453


sdk/servicebus/azure-servicebus/azure/servicebus/_common/message.py
```diff
@@ -128,6 +128,13 @@ def __init__(self, body, **kwargs):
     def __str__(self):
         return str(self.message)
 
+    def __repr__(self):
+        # type: () -> str
+        rec_repr = "body={}".format(
+            self.message._body # pylint: disable=protected-access
```
yunhaoling, line 134: self.body should be enough?
```suggestion
            self.body
```

sdk/servicebus/azure-servicebus/azure/servicebus/_common/message.py
```diff
@@ -128,6 +128,13 @@ def __init__(self, body, **kwargs):
     def __str__(self):
         return str(self.message)
 
+    def __repr__(self):
+        # type: () -> str
+        rec_repr = "body={}".format(
+            self.body
```
yunhaoling, line 134: also wondering if other public properties should be included, it seems like EH is exposing all the public properties:

https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/eventhub/azure-eventhub/azure/eventhub/_common.py#L104-L132

sdk/servicebus/azure-servicebus/azure/servicebus/_common/message.py
```diff
@@ -128,6 +128,54 @@ def __init__(self, body, **kwargs):
     def __str__(self):
         return str(self.message)
 
+    def __repr__(self):
+        # type: () -> str
+        # pylint: disable=bare-except
+        message_repr = "body={}".format(
+            str(self.body)
```
yunhaoling, line 135: I remember `self.body` is returning a generator -- `str(self.body)` would print `<generator object DataBody.data at 0x000001F967066A48>`, `self.message` should be good
```suggestion
            str(self.message)
```

sdk/servicebus/azure-servicebus/azure/servicebus/_common/message.py
```diff
@@ -128,6 +128,54 @@ def __init__(self, body, **kwargs):
...
+        except:
+            pass
+        try:
+            message_repr += ", subject={}".format(self.subject)
+        except:
+            pass
+        try:
+            message_repr += ", scheduled_enqueue_time_utc={}".format(self.scheduled_enqueue_time_utc)
+        except:
+            pass
```
yunhaoling, line 176: in eventhub we have something like `event_repr += ", properties=<read-error>"` in the except clause, I think we should align with that?

e.g. https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/eventhub/azure-eventhub/azure/eventhub/_common.py#L115

sdk/servicebus/azure-servicebus/azure/servicebus/_common/message.py
```diff
@@ -691,6 +739,58 @@ def _to_outgoing_message(self):
             to=self.to,
         )
 
+    def __repr__(self):
```
yunhaoling, line 742: there're some other public properties on the ServiceBusReceivedMessage.
```python
dead_letter_error_description
dead_letter_reason
dead_letter_source
delivery_count
enqueued_sequence_number
enqueued_time_utc
expires_at_utc
sequence_number
lock_token
locked_until_utc
```

sdk/servicebus/azure-servicebus/azure/servicebus/_common/message.py
```diff
@@ -691,6 +739,58 @@ def _to_outgoing_message(self):
...
+        except:
+            pass
+        try:
+            message_repr += ", scheduled_enqueue_time_utc={}".format(self.scheduled_enqueue_time_utc)
+        except:
+            pass
+        try:
+            message_repr += ", auto_renew_error={}".format(self.auto_renew_error)
+        except:
+            pass
```
yunhaoling, line 791: could we use the `__repr__` of parent class ServiceBusMessage to avoid duplicate?

sdk/servicebus/azure-servicebus/azure/servicebus/_common/message.py
```diff
@@ -691,6 +739,58 @@ def _to_outgoing_message(self):
             to=self.to,
         )
 
+    def __repr__(self):
+        # type: () -> str
+        # pylint: disable=bare-except
+        message_repr = "body={}".format(
+            str(self.body)
```
yunhaoling, line 746: same body comment

sdk/servicebus/azure-servicebus/azure/servicebus/_common/message.py
```diff
@@ -691,6 +739,58 @@ def _to_outgoing_message(self):
...
+        )
+        try:
+            message_repr += ", application_properties={}".format(self.application_properties)
+        except:
```
yunhaoling, line 750: same except comment

sdk/servicebus/azure-servicebus/tests/test_repr.py
```diff
@@ -0,0 +1,27 @@
...
+from azure.servicebus._common.constants import (
+    _X_OPT_PARTITION_KEY,
+    _X_OPT_VIA_PARTITION_KEY,
+    _X_OPT_SCHEDULED_ENQUEUE_TIME
+)
+
+def test_servicebus_message_repr():
+    message = ServiceBusMessage("hello")
+    assert "application_properties=None, session_id=None," in message.__repr__()
+    assert "content_type=None, correlation_id=None, to=None, reply_to=None, reply_to_session_id=None, subject=None, scheduled_enqueue_time_utc" in message.__repr__()
```
yunhaoling, line 13: could you add another test which has those properties being set?

sdk/servicebus/azure-servicebus/tests/test_repr.py
```diff
@@ -0,0 +1,27 @@
...
+        annotations={
+            _X_OPT_PARTITION_KEY: b'r_key',
+            _X_OPT_VIA_PARTITION_KEY: b'r_via_key',
+            _X_OPT_SCHEDULED_ENQUEUE_TIME: 123424566,
+        },
+        properties=uamqp.message.MessageProperties()
+    )
+    received_message = ServiceBusReceivedMessage(uamqp_received_message, receiver=None)
+    assert "application_properties=None, session_id=None" in received_message.__repr__()
+    assert "content_type=None, correlation_id=None, to=None, reply_to=None, reply_to_session_id=None, subject=None, scheduled_enqueue_time_utc" in received_message.__repr__()
```
yunhaoling, line 27: I think we also need to verify in the assertion that `__repr__` contains the `annotations`.
```
        annotations={
            _X_OPT_PARTITION_KEY: b'r_key',
            _X_OPT_VIA_PARTITION_KEY: b'r_via_key',
            _X_OPT_SCHEDULED_ENQUEUE_TIME: 123424566,
        },
```

sdk/servicebus/azure-servicebus/tests/test_repr.py
```diff
@@ -0,0 +1,27 @@
...
+    uamqp_received_message = uamqp.message.Message(
+        body=b'data',
+        annotations={
+            _X_OPT_PARTITION_KEY: b'r_key',
+            _X_OPT_VIA_PARTITION_KEY: b'r_via_key',
+            _X_OPT_SCHEDULED_ENQUEUE_TIME: 123424566,
+        },
+        properties=uamqp.message.MessageProperties()
+    )
+    received_message = ServiceBusReceivedMessage(uamqp_received_message, receiver=None)
```
yunhaoling, line 25: another test with properties being set?

sdk/servicebus/azure-servicebus/azure/servicebus/_common/message.py
rakshith91: I really don't think we should align with it - given repr is 1024 chars long only - i don't think <read-error> gives out any  information important enough

sdk/servicebus/azure-servicebus/azure/servicebus/_common/message.py
rakshith91: I really don't think we should align with it - given repr is 1024 chars long only - i don't think gives out any information important enough

sdk/servicebus/azure-servicebus/azure/servicebus/_common/message.py
rakshith91: I really don't think we should align with it - given repr is 1024 chars long only - i don't think gives out any information important enough

sdk/servicebus/azure-servicebus/azure/servicebus/_common/message.py
rakshith91: the super class returns "ServiceBusMessage ......." as opposed to "ServiceBusReceivedMessage ....."

sdk/servicebus/azure-servicebus/tests/test_repr.py
rakshith91: added

sdk/servicebus/azure-servicebus/tests/test_repr.py
rakshith91: done

sdk/servicebus/azure-servicebus/azure/servicebus/_common/message.py
```diff
@@ -703,6 +751,98 @@ def _to_outgoing_message(self):
             to=self.to,
         )
 
+    def __repr__(self):
```
rakshith91, line 754:
```suggestion
    def __repr__(self): # pylint: disable=too-many-branches,too-many-statements
```

sdk/servicebus/azure-servicebus/tests/test_repr.py
yunhaoling: I need to apologize that it's my fault for not making it clear -- I see you exposing the `annotations` in the `__repr__`.
the `annotations` is not a public ivar/property, so I think we should not expose it.

I intended to check those values within the `annotations` are included in the repr as each of them you set is related to a public property.
like:
```
assert "partition_key=r_key, scheduled_enqueue_time_utc=123424566" in received_message.__repr__()
```

----

`via_partition_key` is not a public property on `ServiceBusMessage` by design as it's part of the transaction feature which we are missing in current uamqp.

sdk/servicebus/azure-servicebus/azure/servicebus/_common/message.py
```diff
@@ -703,6 +751,98 @@ def _to_outgoing_message(self):
...
+        except:
+            pass
+        try:
+            message_repr += ", locked_until_utc={}".format(self.locked_until_utc)
+        except:
+            pass
+        try:
+            message_repr += ", annotations={}".format(self.message.annotations)
+        except:
+            pass
```
yunhaoling, line 843: we should no expose this.. this is amqp details

sdk/servicebus/azure-servicebus/azure/servicebus/_common/message.py
yunhaoling: so I was checking the [python official doc for `__repr__`](https://docs.python.org/3/reference/datamodel.html#object.__repr__) which mentioned that

> This is typically used for debugging, so it is important that the representation is information-rich and unambiguous.

So I thought `properties=<read-error>` is kinda important information for debugging process.

but I won't be bikeshedding on this, we could always add later

rakshith91: ```This is typically used for debugging, so it is important that the representation is information-rich and unambiguous.```

hmm, in this case it makes sense - adding them

sdk/servicebus/azure-servicebus/azure/servicebus/_common/message.py
```diff
@@ -703,6 +751,98 @@ def _to_outgoing_message(self):
...
+            message_repr += ", reply_to=<read-error>"
+        try:
+            message_repr += ", reply_to_session_id={}".format(self.reply_to_session_id)
+        except:
+            message_repr += ", reply_to_session_id=<read-error>"
+        try:
+            message_repr += ", subject={}".format(self.subject)
+        except:
+            message_repr += ", subject=<read-error>"
+        try:
```
yunhaoling, line 796: `time_to_live` missing here

sdk/servicebus/azure-servicebus/azure/servicebus/_common/message.py
```diff
@@ -131,6 +131,54 @@ def __init__(self, body, **kwargs):
...
+            message_repr += ", scheduled_enqueue_time_utc={}".format(self.scheduled_enqueue_time_utc)
+        except:
+            message_repr += ", scheduled_enqueue_time_utc=<read-error>"
+        return "ServiceBusMessage({})".format(message_repr)[:1024]
```
yunhaoling, line 180: oops, looks like there're two missing here:
```
time_to_live
partition_key
```

sdk/servicebus/azure-servicebus/azure/servicebus/_common/message.py
```diff
@@ -703,6 +751,98 @@ def _to_outgoing_message(self):
...
+            message_repr += ", dead_letter_source=<read-error>"
+        try:
+            message_repr += ", delivery_count={}".format(self.delivery_count)
+        except:
+            message_repr += ", delivery_count=<read-error>"
+        try:
+            message_repr += ", enqueued_sequence_number={}".format(self.enqueued_sequence_number)
+        except:
+            message_repr += ", enqueued_sequence_number=<read-error>"
+        try:
```
swathipil, line 828: I think `enqueued_time_utc` is also missing?

rakshith91: added

sdk/servicebus/azure-servicebus/azure/servicebus/_common/message.py
rakshith91: added

yunhaoling (approved):
LGTM

PR #17809, opened 2021-04-05
[translation] add nspkg as a release artifact

kristapratico:


maririos (approved):
Also, more comfortable if someone from Python also takes a look

iscai-msft (approved):

PR #17832, opened 2021-04-06
Avoid running non-2.7 and non-3.6 python versions for windows/macos CI tests

benbp:
from @chlowell: "Our shared requirement for cryptography is >=2.1.4. For 2.1.4, PyPI has Windows wheels only up to Python 3.6. Now that CI runs 3.9 on Windows, mindepends must build cryptography 2.1.4 from source. This fails because the VM doesn't have the necessary build tools (for example)."

This also affects macOS.

This PR tunes the base test matrix a little bit to make sure that the python versions that land on windows are only 2.7 and 3.6. There are some overrides in the keyvault and identity CI matrices to also swap the macOS version to 3.6 and re-add the then-missing 3.7 test.


scbedd (approved):

eng/pipelines/templates/stages/platform-matrix.json
```diff
@@ -8,15 +8,15 @@
       "windows-2019": { "OSVmImage": "MMS2019", "Pool": "azsdk-pool-mms-win-2019-general" },
       "macOS-10.15": { "OSVmImage": "macOS-10.15", "Pool": "Azure Pipelines" }
     },
-    "PythonVersion": [ "pypy3", "2.7", "3.6", "3.7", "3.8" ],
+    "PythonVersion": [ "pypy3", "3.6", "3.7", "3.8", "2.7" ],
```
weshaggard, line 11: Lets not rely on our sparse ordering to workaround an issue just for identity and keyvault. Lets be explicit and deterministic and do the exclude/include in those 2 pipelines like you did for macos.

chlowell:
There are more affected libraries than identity and keyvault-keys:
```
❯ rg -tpy "cryptography>=2.1.4"
eventhub\azure-eventhub-checkpointstoreblob\setup.py
73:        "cryptography>=2.1.4",

eventhub\azure-eventhub-checkpointstoreblob-aio\setup.py
72:        "cryptography>=2.1.4",

identity\azure-identity\setup.py
76:        "cryptography>=2.1.4",

keyvault\azure-keyvault-keys\setup.py
85:        "cryptography>=2.1.4",

storage\azure-storage-blob\setup.py
98:        "cryptography>=2.1.4"

storage\azure-storage-file-share\setup.py
83:        "cryptography>=2.1.4"

storage\azure-storage-queue\setup.py
86:        "cryptography>=2.1.4"
```

weshaggard (approved):

scbedd:
@benbp the pypy3 failure for keyvault was a DNS resolution failure. Intermittent and **not a blocker** imo.

benbp:
Yep, generally things are looking pretty good, especially all the cases this PR is intended to fix.

PR #17852, opened 2021-04-06
CertificateCredential requires either certificate_path or certificate_data

chlowell:
... but giving it both is a ValueError.


sdk/identity/azure-identity/azure/identity/_credentials/certificate.py
```diff
@@ -72,6 +72,8 @@ def get_client_credential(certificate_path, password=None, certificate_data=None
     """Load a certificate from a filesystem path or bytes, return it as a dict suitable for msal.ClientApplication"""
 
     if certificate_path:
+        if certificate_data:
+            raise ValueError('Please specify either "certificate_path" or "certificate_data"')
```
mccoyp, line 76: The language technically already implies this, but I think it might help to be explicit about only accepting one or the other. I wouldn't die on this hill though
```suggestion
            raise ValueError('Please specify exactly one of either "certificate_path" or "certificate_data"')
```

mccoyp (approved):

PR #17856, opened 2021-04-06
Fix Issue with "Remove-TestResources" in smoke-tests

scbedd:
@benbp when we default the basename we're looking for, the [logic has issues.](https://github.com/Azure/azure-sdk-for-python/blob/4d50ff1f7b589256c85860226409d2f52846d206/eng/common/TestResources/Remove-TestResources.ps1#L132).

The problem is that we pass in a bunch of folders with `/` in it. We don't actually NEED the full service directory, as it doesn't get used. Updating this so that we can have green cleanup.

### DIFF
--- a/eng/pipelines/templates/jobs/smoke-test.yml
+++ b/eng/pipelines/templates/jobs/smoke-test.yml
@@ -214,7 +214,7 @@ jobs:
 
       - template: /eng/common/TestResources/deploy-test-resources.yml
         parameters:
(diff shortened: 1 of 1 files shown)


weshaggard (approved):

weshaggard:
We can close this as I'm fixing it in the tools repo https://github.com/Azure/azure-sdk-for-python/pull/17856

PR #17874, opened 2021-04-07
updating generated code

seankane-msft:


xiangyan99:
Did you generate the code with credential support?

seankane-msft:
> Did you generate the code with credential support?

No, we should discuss this later. Does the appconfig generated code need to be updated? I only discovered the discrepancies while working on [this](https://github.com/Azure/azure-sdk-for-python/pull/16667) PR.

xiangyan99:
try add --add-credentials in your autorest command and let's see the difference. :)

xiangyan99 (approved):

PR #17890, opened 2021-04-08
[Communication]: Migrated python identity tests to use ARM template

lsundaralingam:


sdk/communication/azure-communication-identity/tests/_shared/helper.py
```diff
@@ -10,6 +10,7 @@ class URIIdentityReplacer(RecordingProcessor):
     def process_request(self, request):
         import re
         request.uri = re.sub('/identities/([^/?]+)', '/identities/sanitized', request.uri)
+        request.uri = re.sub('^(.*?)\\.communication.azure.com', 'https://sanitized.communication.azure.com', request.uri)
```
jbeauregardb, line 13: Is there a way to just replace the regex instead of the whole string? The thing is that we may want to know if were tested against an INT resource or not even when the url was sanitized. As an example an INT URL may be something like: https://resourcename.int.communication.azure.com

sdk/communication/azure-communication-identity/tests/recordings/test_communication_identity_client_async.test_get_token.yaml
```diff
@@ -41,26 +42,27 @@ interactions:
...
+      request-context: appId=
       strict-transport-security: max-age=2592000
       transfer-encoding: chunked
-      x-processing-time: 215ms
+      x-processing-time: 146ms
     status:
       code: 200
       message: OK
-    url: https://communicationdo2ohmwsm52.communication.azure.com/identities/sanitized/:issueAccessToken?api-version=2021-03-07
+    url: https://lakshmanscommunication-communication.communication.azure.com/identities/sanitized/:issueAccessToken?api-version=2021-03-07
```
jbeauregardb, line 67: I know this was not sanitized before, but shouldn't this also be sanitized?

sdk/communication/azure-communication-identity/tests/_shared/helper.py
sacheu: Yes, we need to consider the scenario for testing in INT environment.

sacheu (approved):

jbeauregardb (approved):
Looks great! Just left a comment regarding INT testing

sdk/communication/azure-communication-identity/tests/_shared/helper.py
lsundaralingam: Good catch, I updated this to only sanitize the resource.

PR #17909, opened 2021-04-08
[Storage]fix retry on large block upload

xiafu-msft:


tasherif-msft:
Just to confirm this fix is related to https://github.com/Azure/azure-sdk-for-python/pull/17078/files?

xiafu-msft:
@tasherif-msft good suggestion! Also I will make this pr into two since these two commits are not related, then I will quote your description there

sdk/storage/azure-storage-file-share/azure/storage/fileshare/_shared/uploads.py
```diff
@@ -477,6 +477,13 @@ def read(self, size=None):
                             raise IOError("Stream failed to seek to the desired location.")
```
tasherif-msft, line 477: Why do we fail fast in the case of a lock instead of resetting to absolute position like what you've added? I'm guessing this could potentially mess with where other threads should reposition?

sdk/storage/azure-storage-blob/azure/storage/blob/_shared/policies_async.py
```diff
@@ -54,7 +54,6 @@ async def send(self, request):
             request.context.options.pop('raw_response_hook', self._response_callback)
 
         response = await self.next.send(request)
-        await response.http_response.load_body()
```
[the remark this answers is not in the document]
xiafu-msft: this is to load all data in stream, even we've already specifies stream=True.

sdk/storage/azure-storage-blob/azure/storage/blob/_download.py
```diff
@@ -357,29 +358,47 @@ def _initial_request(self):
...
+
+                try:
+                    self._current_content = process_content(
+                        response,
+                        self._initial_offset[0],
+                        self._initial_offset[1],
+                        self._encryption_options
+                    )
+                    retry_active = False
+                except (requests.exceptions.ChunkedEncodingError, requests.exceptions.ConnectionError) as error:
```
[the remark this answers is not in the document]
xiafu-msft: For example if you get a partial gzip file and try to unzip you will get ContentDecodingError, we want to throw immediately since retry wouldn't help.

sdk/storage/azure-storage-file-share/azure/storage/fileshare/_shared/uploads.py
xiafu-msft: we've already seeked and cannot find the absolute position
```
                        self._wrapped_stream.seek(absolute_position, SEEK_SET)
                        # If we can't seek to the right location, our read will be corrupted so fail fast.
                        if self._wrapped_stream.tell() != absolute_position:
                            raise IOError("Stream failed to seek to the desired location.")
```

sdk/storage/azure-storage-blob/azure/storage/blob/_shared/uploads.py
```diff
@@ -477,6 +477,13 @@ def read(self, size=None):
                             raise IOError("Stream failed to seek to the desired location.")
                         buffer_from_stream = self._wrapped_stream.read(current_max_buffer_size)
                 else:
+                    absolute_position = self._stream_begin_index + self._position
+                    # It's possible that there's connection problem during data transfer,
+                    # so when we retry we don't want to read from current position of wrapped stream,
+                    # instead we should seek to where we want to read from.
+                    if self._wrapped_stream.tell() != absolute_position:
+                        self._wrapped_stream.seek(absolute_position, SEEK_SET)
```
tasherif-msft, line 485: note to self: we could also always seek but this is bad because it wouldn't be as efficient (as Emily explained offline)

tasherif-msft (approved):
LGTM :)

Issue #17145, opened 2021-03-06
azure core needs to expose a way to return raw data in streaming


xiangyan99:
1. Now we are calling into requests library and the default behavior is we de-compress the data if there is a "transfer-encoding' header or "content-encoding" header. 
2. In storage SDK, if user wants to download a big file from blob, storage sdk splits it into multiple 4m chunks. 

Here is the broken case:
User uploads a big (>4m) gzipped file into blob and sets the content-encoding to gzip then he/she calls storage sdk to download the file.

Storage SDK splits it into multiple requests, for the non-first request, azure-core recognizes the "content-encoding" header and tries to decompress the stream --  fail

xiangyan99:
#17920

xiangyan99:
Fixed.

PR #17920, opened 2021-04-08
Core raw streaming

xiangyan99:
https://github.com/Azure/azure-sdk-for-python/issues/17145


sdk/core/azure-core/azure/core/pipeline/transport/_aiohttp.py
```diff
@@ -197,32 +197,55 @@ class AioHttpStreamDownloadGenerator(AsyncIterator):
...
-        self.downloaded = 0
 
     def __len__(self):
         return self.content_length
 
     async def __anext__(self):
+        if not(self._decompress) and self.pipeline:
+            try:
+                auto_decompress = self.pipeline.transport.session.auto_decompress
+                self.pipeline.transport.session.auto_decompress = False
```
johanste, line 218: We should not mutate the session since the session can be shared. Which means that we shouldn't have to make the transport public either.

sdk/core/azure-core/azure/core/pipeline/transport/_base.py
```diff
@@ -580,8 +580,8 @@ def __repr__(self):
 
 
 class HttpResponse(_HttpResponseBase):  # pylint: disable=abstract-method
-    def stream_download(self, pipeline):
-        # type: (PipelineType) -> Iterator[bytes]
+    def stream_download(self, pipeline, decompress=True):
```
johanste, line 583: `decompress` should be a keyword-only argument.

sdk/core/azure-core/azure/core/pipeline/transport/_base.py
```diff
@@ -580,8 +580,8 @@ def __repr__(self):
...
+        # type: (PipelineType, bool) -> Iterator[bytes]
```
johanste, line 584: The type is `Optional[bool]`

sdk/core/azure-core/azure/core/pipeline/transport/_aiohttp.py
xiangyan99: Agree. But the auto_decompress in aiohttp is a per-session setting. If we don't support session mutation, we have to make it as a per-session setting as well. Your idea?

johanste: Just throwing it out as an option; we can do the decompression outside of aiohttp and disable the decompression for the sessions we create. It may be tricky to do in a backwards compatible way, however.

sdk/core/azure-core/CLIENT_LIBRARY_DEVELOPER.md
```diff
@@ -269,7 +269,7 @@ class HttpResponse(object):
     def text(self, encoding=None):
         """Return the whole body as a string."""
 
-    def stream_download(self, chunk_size=None, callback=None):
+    def stream_download(self, pipeline, **kwargs):
```
johanste, line 272: QQ: what was the callback parameter? I'm pretty sure this was a doc bug, correct?

sdk/core/azure-core/azure/core/pipeline/transport/_aiohttp.py
```diff
@@ -216,6 +217,28 @@ async def __anext__(self):
             chunk = await self.response.internal_response.content.read(self.block_size)
             if not chunk:
                 raise _ResponseStopIteration()
+            if not self._decompress:
+                return chunk
+            enc = self.response.internal_response.headers.get('Content-Encoding')
+            if not enc:
+                return chunk
+            enc = enc.lower()
+            if enc in ("gzip", "deflate", "br"):
```
johanste, line 226: I think I'd prefer for a user to be able to register content encodings they can handle. And us not advertising support for brotli encodings. I am not entirely comfortable taking  a "hidden" dependency on `brotlipy`.

sdk/core/azure-core/azure/core/pipeline/transport/_requests_basic.py
```diff
@@ -98,13 +120,19 @@ class StreamDownloadGenerator(object):
...
+        on the ‘content-encoding’ header.
     """
-    def __init__(self, pipeline, response):
+    def __init__(self, pipeline, response, **kwargs):
         self.pipeline = pipeline
         self.request = response.request
         self.response = response
         self.block_size = response.block_size
-        self.iter_content_func = self.response.internal_response.iter_content(self.block_size)
+        decompress = kwargs.get("decompress", True)
```
johanste, line 131: We should `pop` the `decompress` `kwarg` and raise an error if we get unknown `kwarg`s (since we are not pushing them through anywhere else, I believe?)

sdk/core/azure-core/CLIENT_LIBRARY_DEVELOPER.md
xiangyan99: Yes. It was a doc bug.

sdk/core/azure-core/azure/core/pipeline/transport/_aiohttp.py
xiangyan99: This is the behavior in aiohttp. I think we don't want to change the behavior, right?

sdk/core/azure-core/azure/core/pipeline/transport/_requests_basic.py
xiangyan99: Yes. passed-in kwargs will be thrown away after we read the decompress information.

sdk/core/azure-core/azure/core/pipeline/transport/_aiohttp.py
```diff
@@ -197,16 +198,18 @@ class AioHttpStreamDownloadGenerator(AsyncIterator):
...
     """
-    def __init__(self, pipeline: Pipeline, response: AsyncHttpResponse) -> None:
+    def __init__(self, pipeline: Pipeline, response: AsyncHttpResponse, **kwargs) -> None:
         self.pipeline = pipeline
         self.request = response.request
         self.response = response
         self.block_size = response.block_size
+        self._decompress = kwargs.pop("decompress", True)
+        if len(kwargs) > 0:
+            raise ValueError("Unknown parameters!")
```
johanste, line 211: `TypeError` + include the first incorrect parameter name.

xiangyan99: Sorry, why TypeError? 

We expect the kwargs to be a single key dict by user passes in a multi-key dict, right?

sdk/core/azure-core/azure/core/pipeline/transport/_aiohttp.py
```diff
@@ -197,16 +198,18 @@ class AioHttpStreamDownloadGenerator(AsyncIterator):
...
+    :keyword bool decompress: If True which is default, will attempt to decode the body based
+        on the ‘content-encoding’ header.
     """
-    def __init__(self, pipeline: Pipeline, response: AsyncHttpResponse) -> None:
+    def __init__(self, pipeline: Pipeline, response: AsyncHttpResponse, **kwargs) -> None:
         self.pipeline = pipeline
         self.request = response.request
         self.response = response
         self.block_size = response.block_size
+        self._decompress = kwargs.pop("decompress", True)
```
johanste, line 209: I suspect that we can use the status code (206 - Partial content) + the Content-Range header to determine if we can decompress by default. And have that as the default (e.g. the user didn't provide the `decompress` parameter). Did you/we consider that?

sdk/core/azure-core/azure/core/pipeline/transport/_aiohttp.py
johanste: Thinking. It's not that I want to change the behavior (in fact, I very much don't want to do that). But the combination we are talking about here is range requests + auto decoding + brotli. Which is rare. And the user does have the work-around for that (raw stream + decompress themselves). I want to do a bit more due diligence before *we* directly recommend brotlipy to our users (there is a subtle difference)

xiangyan99: I think the scenario is non-range requests + auto decoding + 'br' content-encoding.

Currently, it succeeds and user gets decoded stream.

If we remove 'br' here, user will get encoded stream which is a behavior change, right?

xiangyan99: And brotlipy is what aiohttp uses. aiohttp also "hiddenly" import brotlipy. So I think it is fair to have it in our aiohttp transport.

xiangyan99: And this is aiohttp's code:

https://github.com/aio-libs/aiohttp/blob/master/aiohttp/http_parser.py#L44

https://github.com/aio-libs/aiohttp/blob/master/aiohttp/http_parser.py#L846

sdk/core/azure-core/azure/core/pipeline/transport/_aiohttp.py
johanste: `kwargs` is not being passed in as a dict. Try the following:

```python
def tryit(a=None, b=None):
    print(a, b)

tryit(c=7)
```

xiangyan99: Updated.

sdk/core/azure-core/azure/core/pipeline/transport/_aiohttp.py
johanste: Ok, a couple of things to point out -  aiohttp depends on [`Brotli`](https://pypi.org/project/Brotli/), not [`brotlipy`](https://pypi.org/project/brotlipy/). Which makes a pretty big difference in my mind. Secondly, aiohttp has made a decision to include a set of things as extras. Which we really should do as well if we want to advertise support for plug-ins like is the case here. 

If it wasn't for the potential of breaking existing working workloads, I would have strongly suggested a different pattern here. As it stands, I'm not sure what to do that isn't breaking unless we mimic what aiohttp does.

johanste: ...but, we should absolutely make sure that we suggest that users install the right dependency in the right way!

johanste: ...and we should add tests for brotli encoded responses.

xiangyan99: Took offline.

sdk/core/azure-core/azure/core/pipeline/transport/_requests_asyncio.py
```diff
@@ -138,17 +138,22 @@ class AsyncioStreamDownloadGenerator(AsyncIterator):
...
-    def __init__(self, pipeline: Pipeline, response: AsyncHttpResponse) -> None:
+    def __init__(self, pipeline: Pipeline, response: AsyncHttpResponse, **kwargs) -> None:
         self.pipeline = pipeline
         self.request = response.request
         self.response = response
         self.block_size = response.block_size
-        self.iter_content_func = self.response.internal_response.iter_content(self.block_size)
+        decompress = kwargs.pop("decompress", True)
+        if len(kwargs) > 0:
+            raise TypeError("Got an unexpected keyword argument: {}".format(list(kwargs.keys())[0]))
```
annatisch, line 151: Do we think adding this error might suddenly break anything that was previously 'silently ignore' unused kwargs?

annatisch (approved):
LGTM :)

sdk/core/azure-core/azure/core/pipeline/transport/_requests_asyncio.py
annatisch: Ah - nevermind - kwargs weren't even there before.

PR #17924, opened 2021-04-08
move apireview step into the analyze stage!

scbedd:
To avoid the weirdness with failing to upload due to failed apireview

### DIFF
--- a/eng/pipelines/templates/steps/analyze.yml
+++ b/eng/pipelines/templates/steps/analyze.yml
@@ -119,4 +119,8 @@ steps:
       TestMarkArgument: ${{ parameters.TestMarkArgument }}
       AdditionalTestArgs: ${{parameters.AdditionalTestArgs}}
 
+  - template: /eng/common/pipelines/templates/steps/create-apireview.yml
+    parameters:
+      Artifacts: ${{ parameters.Artifacts }}
+
   - template: /eng/common/pipelines/templates/steps/eng-common-workflow-enforcer.yml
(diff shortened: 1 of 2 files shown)


eng/pipelines/templates/steps/analyze.yml
```diff
@@ -119,4 +119,8 @@ steps:
       TestMarkArgument: ${{ parameters.TestMarkArgument }}
       AdditionalTestArgs: ${{parameters.AdditionalTestArgs}}
 
+  - template: /eng/common/pipelines/templates/steps/create-apireview.yml
```
weshaggard, line 122: Do you know if this will be able to find the correct path based one where they were downloaded?

weshaggard:
Doesn't look like the apireview runs in ci so you will need to do a manual run to verify.

weshaggard (approved):
Looks good as long as the paths work correctly.

eng/pipelines/templates/steps/analyze.yml
scbedd: It's not working properly right now. I'm going to be updating it to download the artifact, then passing that artifact directory to it.

scbedd: Passes tests, but it's wrong.

scbedd:
@weshaggard  they won't. am fixing. Thanks for the review.

eng/pipelines/templates/steps/analyze.yml
```diff
@@ -93,7 +93,7 @@ steps:
       TestMarkArgument: ${{ parameters.TestMarkArgument }}
```
weshaggard, line 93: We also need to be sure to move the artifact publishing to the end of this job and make sure it uses our new upload template so that we can retry this job in the case of failure.

scbedd:
image

Looks like it works! Thanks @weshaggard

PR #17960, opened 2021-04-12
Enable retain runs.

mitchdenny:
Enables retention of run after tagging occurs.


mitchdenny:
Related: https://github.com/Azure/azure-sdk/issues/1059

weshaggard (approved):

PR #17989, opened 2021-04-12
Update document translation sample slug

kristapratico:


heaths:
Should they maybe all be updated?

heaths (approved):
Approving to unblock, but

1. Should probably fix 'em all.
2. Make sure the Translator service team agrees (I'll update my PR accordingly, since this is probably a better choice you have here).

kristapratico:
> Should they maybe all be updated?

I opened issues for all other services that had invalid slugs.

> Make sure the Translator service team agrees (I'll update my PR accordingly, since this is probably a better choice you have here

Yeah going to hold on merging until we hear back from @maririos

PR #17999, opened 2021-04-13
add samples readme slug analysis

scbedd:


eng/pipelines/templates/steps/analyze.yml
```diff
@@ -59,6 +59,11 @@ steps:
     condition: and(succeededOrFailed(), ne(variables['Build.Reason'],'PullRequest'))
     displayName: 'Component Detection'
 
+  - pwsh: |
+      Get-ChildItem $(Build.SourcesDirectory)/sdk/${{parameters.ServiceDirectory}} -Filter *.md -Recurse `
+        | eng/scripts/Find-InvalidSlug.ps1 -IncludeParents
```
heaths, line 64: This should go into eng/common if we're going to use it across language repos. The step should too.

heaths (changes requested):

eng/scripts/Find-InvalidSlug.ps1
```diff
@@ -0,0 +1,451 @@
...
+        foreach ($product in $products) {
+            if ($slugs -notcontains $product) {
+                $invalidProducts += $product
+                Write-Error "File '$($file.FullName)' contains invalid slug: $product"
+            }
+        }
+
+        if ($invalidProducts) {
+            $file | Add-Member -Type NoteProperty -Name 'InvalidSlugs' -Value $invalidProducts -PassThru
+        }
```
heaths, line 53: Since you've changed `Write-Warning` to `Write-Error`, remove this block. I wrote the original script to be more PowerShelly, i.e. not doing a monolithic thing, but instead returning results on which you could act on further.

eng/pipelines/templates/steps/analyze.yml
heaths: I updated https://gist.github.com/heaths/08a395af1800a2826be8b8759e735394 with some improvements (no change in desired behavior), including Write-Error instead of Write-Warning. File info is only passed back now if you pass `-PassThru` - not exactly recommended with the `Find` standard verb, but 🤷‍♂️.

eng/scripts/Find-InvalidSlug.ps1
heaths: I fixed this in my updates (see other comment).

eng/pipelines/templates/steps/analyze.yml
heaths: I'll start a PR. If we're discussing adding more validation, I'm going to rename this to be less-specific to slugs and more general, e.g. `Test-SampleMetadata.ps1`.

heaths: See https://github.com/Azure/azure-sdk-tools/pull/1555

heaths: Azure/azure-sdk-tools#1555 has been merged.

PR #18013, opened 2021-04-13
allow filtering live test sets based on package name

GoWang:
#### What
We want to have the ability to trigger live tests separately for different packages 
#### How
Set runtime parameters ```Services``` at the test YAML file.
Based on [recommendations ](https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/keyvault/tests.yml)from sys engineering team, add in test filtering logic based on package name so that developers can override [ run parameters from the pipeline ](https://docs.microsoft.com/en-us/azure/devops/pipelines/process/runtime-parameters?view=azure-devops&tabs=script). By default, all packages will be built and tested.
### Test: 
* Run [all packages](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=842788&view=results)
* Run [a selected package](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=842906&view=results)

### DIFF
--- a/sdk/communication/tests.yml
+++ b/sdk/communication/tests.yml
@@ -21,6 +21,7 @@ stages:
           BuildTargetingString: 'azure-mgmt-communication'
         JobName: ${{ package }}
         ServiceDirectory: communication
+        DeployArmTemplate: true
         EnvVars:
           AZURE_TEST_RUN_LIVE: 'true'
           AZURE_TENANT_ID: $(aad-azure-sdk-test-tenant-id)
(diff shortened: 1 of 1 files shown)


beltr0n (approved):
lgtm, but i'd probably wait for someone from sys engineering to chime in as well

lsundaralingam (approved):
Will this mean that if I have a PR with just SMS changes that only the SMS tests run?

GoWang:
> Will this mean that if I have a PR with just SMS changes that only the SMS tests run?

you cannot trigger it directly from pr using /azp because azure cli doesn't support this, but you can always trigger it from UI 
Screenshot 2021-04-14 172726

the above example is for phone numbers and the same apply to java and .net.

benbp (approved):

PR #18014, opened 2021-04-13
fix search fields

xiangyan99:
convert search fields from list of string to comma-separated list of field names.


kristapratico (approved):
LGTM

PR #18023, opened 2021-04-13
Create the Python Track 2 SDK for the Microsoft Azure Attestation Service.

LarryOsterman:
Create initial implementation of a Track 2 SDK For Python for the MAA Service.


sdk/attestation/azure-security-attestation/tests/test_attestation_token.py
```diff
@@ -0,0 +1,95 @@
...
+        one_day = datetime.timedelta(1, 0, 0)
+        builder = builder.not_valid_before(datetime.datetime.today() - one_day)
+        builder = builder.not_valid_after(datetime.datetime.today() + (one_day * 30))
+        builder = builder.serial_number(x509.random_serial_number())        
+        builder = builder.public_key(eckey.public_key())
+        builder = builder.add_extension(x509.SubjectAlternativeName([x509.DNSName(u'attestation.test')]), critical=False)
+        builder = builder.add_extension(x509.BasicConstraints(ca=False, path_length=None), critical=True)
+        certificate = builder.sign(private_key=eckey, algorithm=hashes.SHA256())
+        signer = SigningKey(eckey, certificate)
+        assert  signer.certificate.subject==x509.Name([x509.NameAttribute(NameOID.COMMON_NAME, u'attestation.test')])
```
seankane-msft, line 95: Are these tests making HTTP calls? It looks like they are testing the models and not the clients.

sdk/attestation/azure-security-attestation/tests/test_azure_attestation.py
```diff
@@ -64,209 +57,151 @@ class AzureAttestationTest(AzureTestCase):
...
             super(AzureAttestationTest, self).setUp()
 
-    @pytest.mark.live_test_only
-    def test_shared_getopenidmetadata(self):
-        attest_client = self.shared_client()
-        open_id_metadata = attest_client.metadata_configuration.get()
+    @AttestationPreparer()
+    def test_shared_getopenidmetadata(self, attestation_location_short_name):
+        attest_client = self.shared_client(attestation_location_short_name)
+        open_id_metadata = attest_client.get_openidmetadata()
```
seankane-msft, line 63: I looked at the call for `get_openidmetadata`, we always remove this request from the recording to avoid leaks, any tests that run this call will have to be marked with the `@pytest.mark.live_test_only`, in the meantime we can discuss removing this with  the Python crew.

sdk/attestation/azure-security-attestation/tests/test_azure_attestation_async.py
```diff
@@ -0,0 +1,75 @@
...
+    def __init__(self, *args, **kwargs):
+        super(AsyncAzureAttestationTest, self).__init__(*args, **kwargs)
+
+    def setUp(self):
+        super(AsyncAzureAttestationTest, self).setUp()
+
+    @AttestationPreparer()
+    async def test_shared_getopenidmetadataasync(self, attestation_location_short_name):
+        attest_client = self.shared_client(attestation_location_short_name)
+        open_id_metadata = (await attest_client.get_openidmetadata())
```
seankane-msft, line 51: Same goes for the async `get_openidmetadata` calls.

sdk/attestation/tests.yml
```diff
@@ -34,3 +33,5 @@ jobs:
        ATTESTATION_SERIALIZED_POLICY_SIGNING_KEY2: "$(ATTESTATION_SERIALIZED_POLICY_SIGNING_KEY2)"
        ATTESTATION_SERIALIZED_POLICY_SIGNING_KEY0: "$(ATTESTATION_SERIALIZED_POLICY_SIGNING_KEY0)"
        ATTESTATION_SERIALIZED_POLICY_SIGNING_KEY1: "$(ATTESTATION_SERIALIZED_POLICY_SIGNING_KEY1)"
+
```
seankane-msft, line 36:
```suggestion
      AllocateResourceGroup: false
      BuildTargetingString: azure-security-attestation
      ServiceDirectory: attestation
      TestSamples: false
      DeployArmTemplate: true
```

sdk/attestation/azure-security-attestation/tests/test_attestation_token.py
seankane-msft: If they are not making HTTP calls, they should be split off into a different test class that does not inherit from `AzureTestCase`, reason being is the recording infrastructure acts up when there are tests that do not generate traffic. [Here](https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/tables/azure-data-tables/tests/test_table.py#L522-L562) is an example of this split with the `tables` library.

LarryOsterman: Ok, silly question: What should I derive the class from if NOT AzureTestCase. What is the minimal infrastructure I need to be considered a "test case"?

seankane-msft: You can inherit from `object`, just make sure that the class name starts with `Test` or pytest won't recognize it.

sdk/attestation/tests.yml
LarryOsterman: What does this do? I believe we need the tests to create the template in an RG, does this disable this?

seankane-msft: These are for when your tests run live in the pipelines, I know we're not worried about them now but better to be proactive. The last one will deploy the ARM template. I don't believe you need a resource group set-up outside of the one that will be deployed in the ARM template (disabled with `AllocateResourceGroup`), and for now won't test your samples live (we can enable later, but one thing at a time for now).

sdk/attestation/azure-security-attestation/azure/security/attestation/_models.py
```diff
@@ -0,0 +1,228 @@
...
+from json import JSONDecoder, JSONEncoder
+from datetime import datetime
+
+T = TypeVar('T')
+
+class AttestationSigner(object):
+    """ Represents a signing certificate returned by the Attestation Service.
+
+    """
+    def __init__(self, certificates: List[Certificate], key_id : str, **kwargs):
```
seankane-msft, line 18:
```suggestion
    def __init__(self, certificates, key_id, **kwargs):
    # type: (List[Certificate], str) -> None
```

sdk/attestation/azure-security-attestation/azure/security/attestation/_models.py
```diff
@@ -0,0 +1,228 @@
...
+
+    Typically the signing key used by the service consists of two components: An RSA or ECDS private key and an X.509 Certificate wrapped around
+    the public key portion of the private key.
+
+    :var signing_key: The RSA or ECDS signing key to sign the token supplied to the customer.
+    :vartype signing_key: (rsa.RSAPrivateKey | ElilipticCurvePrivateKey)
+    :var certificate: An X.509 Certificate whose public key matches the signing_key's public key.
+    :vartype certificate: Certificate
+    """
+    def __init__(self, signing_key, certificate: Certificate ):
```
seankane-msft, line 47:
```suggestion
    def __init__(self, signing_key, certificate):
    # type: (Union[rsa.RSAPrivateKey, ElilipticCurvePrivateKey], Certificate) -> None
```

sdk/attestation/azure-security-attestation/azure/security/attestation/_models.py
```diff
@@ -0,0 +1,228 @@
...
+    def __str__(self):
+        return self._token
+
+    def serialize(self):
+        return self._token
+
+
+    """ Validate the attestation token based on the options specified in the TokenValidationOptions
+    """
+    def validate_token(self, options: TokenValidationOptions, signing_certificates: List[AttestationSigner]) -> bool:
```
seankane-msft, line 182:
```suggestion
    def validate_token(self, options, signing_certificates):
    # type: (TokenValidationOptions, List[AttestationSigner]) -> bool
```

sdk/attestation/azure-security-attestation/azure/security/attestation/_models.py
```diff
@@ -0,0 +1,228 @@
...
+        if self.algorithm != 'none' and options.validate_signature:
+            # validate the signature for the token.
+            pass
+
+        if (options.validation_callback is not None):
+            return options.validation_callback(self, None)
+        return True
+
+
+    def get_body(self) -> T:
```
seankane-msft, line 197:
```suggestion
    def get_body(self):
    # type: () -> T
```

sdk/attestation/azure-security-attestation/azure/security/attestation/_models.py
```diff
@@ -0,0 +1,228 @@
...
+        return self._body
+
+    @staticmethod
+    def _create_unsecured_jwt(body: Any) -> str:
```
seankane-msft, line 201:
```suggestion
    def _create_unsecured_jwt(body):
    # type: (Any) -> str
```

sdk/attestation/azure-security-attestation/tests/test_azure_attestation_async.py
```diff
@@ -0,0 +1,76 @@
...
+            attestation_isolated_signing_certificate='xxxx',
+            attestation_service_management_url='https://management.core.windows.net/',
+            attestation_location_short_name='xxxx',
+            attestation_client_id='xxxx',
+            attestation_client_secret='secret',
+            attestation_tenant_id='tenant',
+            attestation_isolated_url='https://fakeresource.wus.attest.azure.net',
+            attestation_aad_url='https://fakeresource.wus.attest.azure.net',
+#            attestation_resource_manager_url='https://resourcemanager/zzz'
+        )
```
seankane-msft, line 38: I would recommend breaking this out into a separate file and importing it in each test file, reducing code 😃

seankane-msft (commented):
Some small nits and clean up things

sdk/attestation/azure-security-attestation/SWAGGER.md
```diff
@@ -0,0 +1,44 @@
...
+tag: package-2020-10-01
+package-version: 1.0.0b2
+enable-xml: false
+clear-output-folder: true
+python: true
+v3: true
+add-credentials: true
+azure-arm: false
+payload-flattening-threshold: 2
+package-name: azure-attestation
```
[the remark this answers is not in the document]
LarryOsterman: Does this apply for the generated code (this is generated only, not public).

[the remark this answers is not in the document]
seankane-msft: Yes

sdk/attestation/azure-security-attestation/azure/security/attestation/_administration_client.py
```diff
@@ -0,0 +1,126 @@
...
+        token = AttestationToken[PolicyResult](token=policyResult.token, body_type=PolicyResult)
+        token_body = token.get_body()
+        stored_policy = AttestationToken[StoredAttestationPolicy](token=token_body.policy, body_type=StoredAttestationPolicy)
+
+        actual_policy = stored_policy.get_body().attestation_policy #type: bytes
+
+        if self._config.token_validation_options.validate_token:
+            token.validate_token(self._config.token_validation_options, self._get_signers(**kwargs))
+
+        return AttestationResponse[str](token, actual_policy.decode('utf-8'))
```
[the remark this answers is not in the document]
LarryOsterman: Interesting - FWIW, the CI pipeline doesn't complain about this, and I would have thought that this would be caught in the CI when it tested on 2.7. Or do we need to support versions of Python older than the versions in the CI pipeline?

sdk/attestation/azure-security-attestation/tests/test_attestation_token.py
```diff
@@ -0,0 +1,97 @@
...
+    def test_create_unsecured_token(self):
+        token = AttestationToken(body={"val1":[1,2,3]})
+        assert token.get_body() == {"val1":[1,2,3]}
+
+    def test_create_signing_key(self):
+        key = self._create_rsa_key()
+        cert = self._create_x509_certificate(key, u'test certificate')
+
+        signer = SigningKey(key, cert)
+        self.assertEqual(signer.certificate.subject, cert.subject)
```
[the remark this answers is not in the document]
LarryOsterman: Why switch from the pytest variant (self.assertEqual) to assert? I would have thought that pytest's assertions were preferable.

sdk/attestation/azure-security-attestation/tests/test_attestation_token.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def test_create_signing_wrong_key(self):
+        """ SigningKey should throw if the key and certificate don't match.
+        """
+        key1 = self._create_rsa_key()
+        cert = self._create_x509_certificate(key1, u'test certificate')
+        key2 = self._create_rsa_key()
+
+        # This should throw an exception, fail if the exception isn't thrown.
+        with self.assertRaises(Exception):
```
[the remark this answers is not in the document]
LarryOsterman: That *is* specific to the type of Exception I am expecting. My code throws Exception because I can't find an Exception type which is more accurate.

sdk/attestation/azure-security-attestation/tests/test_azure_attestation_async.py
LarryOsterman: Tried, I'm getting "attempted relative import with no known parent package".

sdk/attestation/azure-security-attestation/azure/security/attestation/_configuration.py
```diff
@@ -6,17 +6,17 @@
 # Changes may cause incorrect behavior and will be lost if the code is regenerated.
 # --------------------------------------------------------------------------
```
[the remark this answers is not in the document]
LarryOsterman: This was copied from the _generated directory. Since my only "option" is the token validation options, this is mostly boilerplate.

sdk/attestation/azure-security-attestation/azure/security/attestation/_models.py
```diff
@@ -0,0 +1,505 @@
...
+        # Ensure that the public key in the certificate matches the public key of the key.
+        cert_public_key = certificate.public_key().public_bytes(
+            Encoding.PEM, PublicFormat.SubjectPublicKeyInfo)
+        key_public_key = signing_key.public_key().public_bytes(
+            Encoding.PEM, PublicFormat.SubjectPublicKeyInfo)
+        if cert_public_key != key_public_key:
+            raise Exception("Signing key must match certificate public key")
+
+
+class AttestationToken(Generic[T]):
```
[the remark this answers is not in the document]
LarryOsterman: All of the samples I've seen for using Generic[T] have indicated that the base type should be Generic[T], is that not the case?

sdk/attestation/azure-security-attestation/azure/security/attestation/_models.py
```diff
@@ -0,0 +1,505 @@
...
+        return_value += '.'
+        return_value += Base64Url.encode(signature)
+        return return_value
+
+
+class AttestationResponse(Generic[T]):
+    def __init__(self, token, value):
+        # type (AttestationToken, T) -> None
+        self.token = token #type: AttestationToken
+        self.value = value #type: T
```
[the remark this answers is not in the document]
LarryOsterman: The Generic specialization for AttestationResult.

sdk/attestation/azure-security-attestation/azure/security/attestation/aio/__init__.py
```diff
@@ -6,5 +6,19 @@
 # Changes may cause incorrect behavior and will be lost if the code is regenerated.
```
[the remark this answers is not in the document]
LarryOsterman: It was once upon a type.

sdk/attestation/azure-security-attestation/azure/security/attestation/aio/_client_async.py
```diff
@@ -0,0 +1,90 @@
+# coding=utf-8
```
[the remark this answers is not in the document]
LarryOsterman: I thought the # coding=utf-8 was a python-ism indicating that the source should be rendered as UTF-8, is that not the case?

sdk/attestation/azure-security-attestation/azure/security/attestation/_administration_client.py
```diff
@@ -0,0 +1,126 @@
...
+    @distributed_trace
+    def get_policy(self, attestation_type, **kwargs): 
+        #type(AttestationType) -> AttestationResult[str]:
+        """ Retrieves the attestation policy for a specified attestation type.
+        :param attestation_type - The attestation parameter type.
+        :type attestation_type: AttestationType
+        """
+        
+        policyResult = self._client.policy.get(attestation_type, **kwargs)
+        token = AttestationToken[PolicyResult](token=policyResult.token, body_type=PolicyResult)
```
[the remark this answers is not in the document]
LarryOsterman: Where would I want to use the `_from_generated` method? I'm not sure I understand how it's being used in AppConfig.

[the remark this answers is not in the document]
seankane-msft: The `_from_generated` methods turns the generated model into the model that is exported to the user.

sdk/attestation/azure-security-attestation/azure/security/attestation/aio/_client_async.py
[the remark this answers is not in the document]
seankane-msft: The encoding is fine, but this file looks like generated code, are you using this for your convenience layer?

sdk/attestation/azure-security-attestation/tests/test_attestation_token.py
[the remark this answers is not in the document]
seankane-msft: `self.assertEqual` is from the unittest inheritance (which is within the `AzureTestCase` inheritance . We try to not use `unittest`, hoping to remove it entirely in the near future.

sdk/attestation/azure-security-attestation/tests/test_azure_attestation_async.py
seankane-msft: An [example](https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/tables/azure-data-tables/tests/test_retry.py#L34) in tables SDK. You don't need to do a `from .preparers import AttestationPreparer`, you can drop the '.'

sdk/attestation/azure-security-attestation/azure/security/attestation/_administration_client.py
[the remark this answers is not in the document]
LarryOsterman: I must be impossibly dense. Why wouldn't I simply return the generated model (with some tiny cleanups done in the autorest processing) to the caller? I guess I could just always have a _from_generated and _to_generated method which does nothing but return its input, but that seems like a strange development paradigm to me.

sdk/attestation/azure-security-attestation/azure/security/attestation/aio/_client_async.py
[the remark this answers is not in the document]
LarryOsterman: Yes - because the convenience layer started with the generated code (because I trust the generated code to create all the python idiosyncrasies properly, and I have no faith in my  ability to do the same).

sdk/attestation/azure-security-attestation/tests/test_attestation_token.py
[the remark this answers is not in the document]
LarryOsterman: In this case, self.assertEqual came from pytest.UnitTest. If you are planning on moving away from pytest, what is the plan for testing moving forward?

sdk/attestation/azure-security-attestation/azure/security/attestation/_administration_client.py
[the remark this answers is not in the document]
LarryOsterman: I think I figured this out - specifically, except for enumerations, the generated model isn't suitable to be published, so building a convenience wrapper for the generated types make sense.

I'll tackle that in a subsequent pull request.

sdk/attestation/azure-security-attestation/azure/security/attestation/_models.py
```diff
@@ -0,0 +1,505 @@
...
+
+
+class SigningKey(object):
+    """ Represents a signing key used by the attestation service.
+
+    Typically the signing key used by the service consists of two components: An RSA or ECDS private key and an X.509 Certificate wrapped around
+    the public key portion of the private key.
+
+    :var signing_key: The RSA or ECDS signing key to sign the token supplied to the customer.
+    :vartype signing_key: (rsa.RSAPrivateKey | EllipticCurvePrivateKey)
```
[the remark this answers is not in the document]
LarryOsterman: Changed the type on signing_key to bytes to remove crypto dependencies in public API surface.

sdk/attestation/azure-security-attestation/azure/security/attestation/_models.py
```diff
@@ -0,0 +1,536 @@
...
+            return True
+
+        signer = None
+        if self.algorithm != 'none' and options.validate_signature:
+            # validate the signature for the token.
+            candidate_certificates = self._get_candidate_signing_certificates(
+                signers)
+            signer = self._validate_signature(candidate_certificates)
+            if (signer is None):
+                raise Exception(
```
seankane-msft, line 337: `ValueError`

sdk/attestation/azure-security-attestation/azure/security/attestation/_models.py
```diff
@@ -0,0 +1,536 @@
...
+
+    def _validate_static_properties(self, options):
+        # type:(TokenValidationOptions) -> bool
+        """ Validate the static properties in the attestation token.
+        """
+        if options.validate_expiration and self.expiration_time is not None:
+            if (datetime.now() > self.expiration_time):
+                delta = datetime.now() - self.expiration_time
+                if delta.total_seconds > options.validation_slack:
+                    raise Exception(u'Token is expired.')
```
seankane-msft, line 433: `ValueError`

sdk/attestation/azure-security-attestation/azure/security/attestation/aio/__init__.py
```diff
@@ -2,9 +2,21 @@
...
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License. See License.txt in the project root for license information.
-# Code generated by Microsoft (R) AutoRest Code Generator.
-# Changes may cause incorrect behavior and will be lost if the code is regenerated.
 # --------------------------------------------------------------------------
 
-from ._attestation_client import AttestationClient
-__all__ = ['AttestationClient']
+from ._client_async import AttestationClient
+#from .._administration_client import AttestationAdministrationClient
```
seankane-msft, line 8: Did this file get removed? If so, you can remove and why was it removed? Otherwise we should include

sdk/attestation/azure-security-attestation/azure/security/attestation/aio/_client_async.py
```diff
@@ -0,0 +1,87 @@
...
+    """
+
+    def __init__(
+        self,
+        credential,  # type: "TokenCredential"
+        instance_url,  # type: str
+        **kwargs  # type: Any
+    ):
+        # type: (str, Any, dict) -> None
+        base_url = '{instanceUrl}'
```
seankane-msft, line 46:
```suggestion
        self.base_url = instance_url
```

sdk/attestation/azure-security-attestation/tests/test_attestation.py
```diff
@@ -0,0 +1,386 @@
...
+    "F5MkxzellrTWY2SjhualZBaWJ2aE0iLAogICAgICAgICAgICAgICAgInkiOiJjVjRkU" +
+    "zRVYUxNZ1BfNGZZNGo4aXI3Y2wxVFhsRmRBZ2N4NTVvN1RrY1NBIgogICAgICAgICAg" +
+    "ICB9CiAgICAgICAgfQogICAgICAgIA")
+
+
+class AttestationTest(AzureTestCase):
+
+    def setUp(self):
+        super(AttestationTest, self).setUp()
+
```
seankane-msft, line 152: This can be removed

sdk/attestation/azure-security-attestation/tests/test_attestation.py
```diff
@@ -0,0 +1,386 @@
...
+
+    # The caching infrastructure won't cache .well-known/openid_metadata responses so
```
seankane-msft, line 154: We can fix this in a separate PR if you want

sdk/attestation/azure-security-attestation/tests/test_attestation.py
```diff
@@ -0,0 +1,386 @@
...
+    # mark the metadata related tests as live-only.
+    @AttestationPreparer()
+    @pytest.mark.live_test_only
+    def test_shared_getopenidmetadata(self, attestation_location_short_name):
+        attest_client = self.shared_client(attestation_location_short_name)
+        open_id_metadata = attest_client.get_openidmetadata()
+        print ('{}'.format(open_id_metadata))
+        assert open_id_metadata["response_types_supported"] is not None
+        if self.is_live:
```
seankane-msft, line 163: Because this is live, you don't need the `if self.is_live` in the tests. The print statements should also be removed at some point

sdk/attestation/azure-security-attestation/tests/test_attestation_token.py
```diff
@@ -0,0 +1,145 @@
...
+
+    def test_create_signing_wrong_key(self):
+        """ SigningKey should throw if the key and certificate don't match.
+        """
+        key1 = self._create_rsa_key()
+        cert = self._create_x509_certificate(key1, u'test certificate')
+        key2 = self._create_rsa_key()
+
+        # This should throw an exception, fail if the exception isn't thrown.
+        with pytest.raises(Exception):
```
seankane-msft, line 46: Generally we want to include the exact Exception so we can be sure of the error raised and not some other error is happening.

sdk/attestation/azure-security-attestation/azure/security/attestation/__init__.py
```diff
@@ -2,15 +2,33 @@
...
 # Licensed under the MIT License. See License.txt in the project root for license information.
-# Code generated by Microsoft (R) AutoRest Code Generator.
-# Changes may cause incorrect behavior and will be lost if the code is regenerated.
 # --------------------------------------------------------------------------
 
-from ._attestation_client import AttestationClient
+from ._client import AttestationClient
+from ._administration_client import AttestationAdministrationClient
+from ._models import AttestationResponse, AttestationSigner, AttestationToken, AttestationSigningKey, AttestationData, PolicyResult, AttestationResult, TpmAttestationResponse, TpmAttestationRequest
+from ._generated.models import AttestationType, StoredAttestationPolicy, CertificateModification
```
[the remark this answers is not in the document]
LarryOsterman: yeah, I'm working on cleaning that up one type at a time. But does importing the types into __init__.py cause them to be exported?

I thought it was just the types in the __all__ global attribute

sdk/attestation/azure-security-attestation/azure/security/attestation/_administration_client.py
```diff
@@ -0,0 +1,212 @@
...
+        policy_token = AttestationToken[StoredAttestationPolicy](
+            body=StoredAttestationPolicy(attestation_policy = attestation_policy.encode('ascii')),
+            signer=signing_key,
+            body_type=StoredAttestationPolicy)
+        policyResult = self._client.policy.set(attestation_type=attestation_type, new_attestation_policy=policy_token.serialize(), **kwargs)
+        token = AttestationToken[PolicyResult](token=policyResult.token,
+            body_type=PolicyResult)
+        if self._config.token_validation_options.validate_token:
+            if not token.validate_token(self._config.token_validation_options, self._get_signers(**kwargs)):
+                raise Exception("Token Validation of PolicySet API failed.")
```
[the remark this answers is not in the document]
LarryOsterman: This one is tough. There are three possible causes of this particular error:

1.  The customers validation callback returned False.
2.  The token returned from the caller is expired (or otherwise doesn't meet the criteria (too early, too late)).
3. Something went horribly wrong with the service and the token's signature is invalid.

For  .Net we defined a custom "token validation failed exception", do you recommend something similar for Python?

sdk/attestation/azure-security-attestation/azure/security/attestation/__init__.py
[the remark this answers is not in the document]
seankane-msft: It is everything in `__all__`, but I would recommend remove those for cleaner code.

sdk/attestation/azure-security-attestation/azure/security/attestation/_administration_client.py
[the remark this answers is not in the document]
seankane-msft: I think that is a good solution, and passing in a message about what the issue was would be awesome for the user.

sdk/attestation/azure-security-attestation/tests/test_attestation_token.py
LarryOsterman: This is the Exception you mentioned above - the exact Exception *is* Exception currently :).

sdk/attestation/azure-security-attestation/azure/security/attestation/__init__.py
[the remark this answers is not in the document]
LarryOsterman: Makes sense. I still have work to remove more generated types from exported types, but this does make things simpler.

sdk/attestation/azure-security-attestation/azure/security/attestation/_administration_client.py
```diff
@@ -0,0 +1,212 @@
...
+        token = AttestationToken[PolicyCertificatesModificationResult](token=cert_response.token,
+            body_type=PolicyCertificatesModificationResult)
+        if self._config.token_validation_options.validate_token:
+            if not token.validate_token(self._config.token_validation_options, self._get_signers(**kwargs)):
+                raise Exception("Token Validation of PolicyCertificate Add API failed.")
+        return AttestationResponse[PolicyCertificatesModificationResult](token, token.get_body())
+
+    @distributed_trace
+    def remove_policy_management_certificate(self, certificate_to_add, signing_key, **kwargs):
+        #type:(bytes, AttestationSigningKey, Any)-> AttestationResponse[PolicyCertificatesModificationResult]
```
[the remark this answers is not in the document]
LarryOsterman: Actually, I think this is handled.

sdk/attestation/azure-security-attestation/azure/security/attestation/_client.py
```diff
@@ -0,0 +1,216 @@
...
+
+        :return List[AttestationSigner]: A list of :class:`AttestationSigner` objects.
+
+        For additional request configuration options, please see https://aka.ms/azsdk/python/options.
+
+        """
+        signing_certificates = self._client.signing_certificates.get(**kwargs)
+        signers = []
+        for key in signing_certificates.keys:
+            assert key.x5_c is not None
```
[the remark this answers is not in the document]
LarryOsterman: You're right, in this case, it's an abend eitherway.

sdk/attestation/azure-security-attestation/azure/security/attestation/_client.py
```diff
@@ -0,0 +1,216 @@
...
+        response = self._client.attestation.attest_tpm(request.data)
+        return TpmAttestationResponse(response.data)
+
+    def _get_signers(self, **kwargs):
+        # type:(Any) -> List[AttestationSigner]
+        """ Returns the set of signing certificates used to sign attestation tokens.
+        """
+
+        with self._statelock:
+            if (self._signing_certificates == None):
```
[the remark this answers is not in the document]
LarryOsterman: Makes sense, but I think it should be `if self._signing_certificates is None:`

sdk/attestation/azure-security-attestation/azure/security/attestation/_client.py
```diff
@@ -0,0 +1,216 @@
...
+                signing_certificates = self._client.signing_certificates.get(**kwargs)
+                self._signing_certificates = []
+                for key in signing_certificates.keys:
+                    # Convert the returned certificate chain into an array of X.509 Certificates.
+                    certificates = []
+                    for x5c in key.x5_c:
+                        der_cert = base64.b64decode(x5c)
+                        certificates.append(der_cert)
+                    self._signing_certificates.append(AttestationSigner(certificates, key.kid))
```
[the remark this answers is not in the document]
LarryOsterman: Woah, that's *nice*.

sdk/attestation/azure-security-attestation/azure/security/attestation/_models.py
```diff
@@ -0,0 +1,536 @@
...
+class AttestationSigner(object):
+    """ Represents a signing certificate returned by the Attestation Service.
+
+    :param list[str] certificates: A list of Base64 encoded X.509 Certificates which may be used
+        to sign an :class:`AttestationToken`. 
+    :param str key_id: A string which identifies a signing key, See `RFC 7517 Section 4.5<https://tools.ietf.org/html/rfc7517#section-4.5>`:
+
+    """
+    def __init__(self, certificates, key_id, **kwargs):
+        # type: (list[bytes], str, Any) -> None
```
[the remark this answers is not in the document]
LarryOsterman: type hint is correct, it's a DER encoded list of bytes.

sdk/attestation/azure-security-attestation/azure/security/attestation/_models.py
```diff
@@ -0,0 +1,536 @@
...
+        if is_json is not None:
+            self._is_json = is_json
+        else:
+            # The caller didn't say if the parameter is JSON or not, try parsing it,
+            # and if it parses, assume it's JSON.
+            try:
+                json.loads(data)
+                self._is_json = True
+            except Exception as e:
+                print("exception ", e)
```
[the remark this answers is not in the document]
LarryOsterman: Nah - this was debugging code I added.

sdk/attestation/azure-security-attestation/azure/security/attestation/aio/__init__.py
LarryOsterman: The aio is a bare-bones implementation which only does one API. Completing the async API is on my long list of things to do.

sdk/attestation/azure-security-attestation/azure/security/attestation/_administration_client.py
```diff
@@ -0,0 +1,212 @@
...
+            signer=signing_key,
+            body_type=AttestationCertificateManagementBody)
+
+        cert_response = self._client.policy_certificates.add(cert_add_token.serialize(), **kwargs)
+        token = AttestationToken[PolicyCertificatesModificationResult](token=cert_response.token,
+            body_type=PolicyCertificatesModificationResult)
+        if self._config.token_validation_options.validate_token:
+            if not token.validate_token(self._config.token_validation_options, self._get_signers(**kwargs)):
+                raise Exception("Token Validation of PolicyCertificate Add API failed.")
+        return AttestationResponse[PolicyCertificatesModificationResult](token, token.get_body())
```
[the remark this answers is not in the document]
seankane-msft: Knowing that the token doesn't always come from the `get_body()` method, I think this is a fine implementation and we can work on something simpler for a later preview. This is all in the convenience wrapper so there should be no difference to the end user.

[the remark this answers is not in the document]
LarryOsterman: Thanks. I started working through this for a while and realized that because I was hiding the generated types, the situation where get_body() wasn't used was becoming increasingly rare, especially since I'm going to be working on hiding the generated types completely in the convenience layer.

sdk/attestation/azure-security-attestation/azure/security/attestation/_models.py
```diff
@@ -0,0 +1,536 @@
...
+        if (not isinstance(signing_key, RSAPrivateKey) and not isinstance(signing_key, EllipticCurvePrivateKey)):
+            raise Exception("Signing keys must be either ECDS or RSA keys.")
+
+        # Ensure that the public key in the certificate matches the public key of the key.
+        cert_public_key = certificate.public_key().public_bytes(
+            Encoding.PEM, PublicFormat.SubjectPublicKeyInfo)
+        key_public_key = signing_key.public_key().public_bytes(
+            Encoding.PEM, PublicFormat.SubjectPublicKeyInfo)
+        if cert_public_key != key_public_key:
+            raise Exception("Signing key must match certificate public key")
```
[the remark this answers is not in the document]
LarryOsterman: Done (FWIW, I'd actually made that change before I saw this comment :)).

sdk/attestation/azure-security-attestation/README.md
```diff
@@ -1,49 +1,361 @@
...
+## Getting started
+
+### Prerequisites
+
+* An Azure subscription.  To use Azure services, including the Microsoft Azure Attestation service, you'll need a subscription.  If you do not have an existing Azure account, you may sign up for a [free trial][azure_sub] or use your [Visual Studio Subscription](https://visualstudio.microsoft.com/subscriptions/) benefits when you [create an account](https://account.windowsazure.com/Home/Index).
+* An existing Azure Attestation Instance, or you can use the "shared provider" available in each Azure region. If you need to create an Azure Attestation service instance, you can use the Azure Portal or [Azure CLI][azure_cli].
+
+### Install the package
+
+Install the Microsoft Azure Attestation client library for .NET with [NuGet][nuget]:
```
lmazuel, line 28: Aaaaaaaaargh :p

LarryOsterman: @#$@#$. I thought I'd caught all of those.

sdk/attestation/azure-security-attestation/README.md
```diff
@@ -1,49 +1,361 @@
...
+* [Get attestation policy](#get-attestation-policy)
+* [Retrieve token validation certificates](#retrieve-token-certificates)
+* [Create an attestation client instance](#create-client-instance)
+
+### Create client instance
+
+Creates an instance of the Attestation Client at uri `endpoint`.
+
+```python
+    def create_client(self, base_uri):
```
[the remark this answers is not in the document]
LarryOsterman: Fair enough. Would it also make sense to add some other variants, like adding token validation options?

sdk/attestation/azure-security-attestation/azure/security/attestation/_administration_client.py
```diff
@@ -0,0 +1,243 @@
...
+        self._base_url = instance_url
+        if not credential:
+            raise ValueError("Missing credential.")
+        self._config = AttestationClientConfiguration(credential, instance_url, **kwargs)
+        self._client = AzureAttestationRestClient(credential, instance_url, **kwargs)
+        self._statelock = Lock()
+        self._signing_certificates = None
+
+    @property
+    def base_url(self):
```
lmazuel, line 59: It's unusual to have that, can we remove it or you need it for other scenarios?

sdk/attestation/azure-security-attestation/azure/security/attestation/_administration_client.py
```diff
@@ -0,0 +1,243 @@
...
+            body_type=PolicyCertificatesResult)
+        if self._config.token_validation_options.validate_token:
+            if not token.validate_token(self._config.token_validation_options, self._get_signers(**kwargs)):
+                raise Exception("Token Validation of PolicyCertificates API failed.")
+        certificates = []
+
+        cert_list = token.get_body()
+
+        for key in cert_list.policy_certificates.keys:
+            key_certs = list()
```
lmazuel, line 139: nit: `key_certs = []`

sdk/attestation/azure-security-attestation/azure/security/attestation/_administration_client.py
```diff
@@ -0,0 +1,243 @@
...
+            for cert in key.x5_c:
+                key_certs.append(base64.b64decode(cert))
+            certificates.append(key_certs)
+        return AttestationResponse[list](token, certificates)
```
lmazuel, line 143: I'm impressed you could get Python 2 CI job green with a line like that :)

sdk/attestation/azure-security-attestation/azure/security/attestation/_administration_client.py
LarryOsterman: I was under the impression that there was a pattern of exposing a `base_uri` attribute on all client classes.

I had a base_uri attribute previously, but PyLance complained that it wasn't being used.

If there is no expectation of exposing a `base_url` attribute, I'll happily remove it.

sdk/attestation/azure-security-attestation/azure/security/attestation/_administration_client.py
LarryOsterman: Fixing that and other cases (there are 4 others).

sdk/attestation/azure-security-attestation/azure/security/attestation/_administration_client.py
LarryOsterman: Removing the unnecessary `[list]`, I don't think it gives any value.

sdk/attestation/azure-security-attestation/azure/security/attestation/_administration_client.py
LarryOsterman: There was also a significant opportunity to clean up that for loop using `[base64.b64decode(cert) for cert in x5clist]` (major props to @seankane-msft for pointing that one out).

sdk/attestation/azure-security-attestation/azure/security/attestation/_administration_client.py
lmazuel: Yes, no expectation on the contrary :)

sdk/attestation/azure-security-attestation/README.md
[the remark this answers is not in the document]
lmazuel: If you believe it brings value since it's a common scenario, yes. Finding the right granularity between verbosity and missing things is tough :)

lmazuel (approved):
Good to preview to me

PR #18024, opened 2021-04-13
[Key Vault] Add public_exponent option to create_key

### LINKED ISSUE #18016: [Key Vault] Add public_exponent option to create_key
The `create_key` method in `azure-keyvault-keys` should accept a `public_exponent` keyword argument. This option specifies the public exponent for a created RSA-HSM key -- documentation should make it clear that (as of API version 7.2) this only works for RSA keys created in a Managed HSM.

mccoyp:
Resolves #18016.

This also tweaks pipeline test configuration, since local testing has shown that we need a longer delay before assigning crypto roles (otherwise we get an error). `platform-matrix.json` has been changed to more closely resemble [JS's](https://github.com/Azure/azure-sdk-for-js/blob/master/sdk/keyvault/keyvault-keys/platform-matrix.json) because HSMs aren't being properly deployed for pipeline tests (see example [here](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=839819&view=logs&j=d2931fad-161b-5770-82cb-ab9682370e2d&t=f8caccc0-eeff-5bcd-6c2a-15a3d1a89be2&l=1812)).

### DIFF
--- a/sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/aio/_client.py
+++ b/sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/aio/_client.py
@@ -58,6 +58,7 @@ class KeyClient(AsyncKeyVaultClientBase):
         :keyword curve: Elliptic curve name. Applies only to elliptic curve keys. Defaults to the NIST P-256
          elliptic curve. To create an elliptic curve key, consider using :func:`create_ec_key` instead.
         :paramtype curve: ~azure.keyvault.keys.KeyCurveName or str
+        :keyword int public_exponent: The RSA public exponent to use. Applies only to RSA keys created in a Managed HSM.
         :keyword key_operations: Allowed key operations
         :paramtype key_operations: list[~azure.keyvault.keys.KeyOperation or str]
         :keyword bool enabled: Whether the key is enabled for use.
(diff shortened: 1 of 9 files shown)


mccoyp:
@benbp Does this [`platform-matrix.json`](https://github.com/Azure/azure-sdk-for-python/blob/6065f08a05cb5037236cb331add6fc4d1824580e/sdk/keyvault/azure-keyvault-keys/platform-matrix.json) and [`tests.yml`](https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/keyvault/tests.yml) setup look like it will correctly enable HSM deployment?

chlowell:
Will `public_exponent` be in the stable 7.2?

mccoyp:
.NET still has this checked in so I would think it's staying, but we won't know for sure until the stable 7.2 swagger is finished up (expected by 4/23)

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/aio/_client.py
```diff
@@ -58,6 +58,7 @@ async def create_key(self, name: str, key_type: "Union[str, KeyType]", **kwargs:
         :keyword curve: Elliptic curve name. Applies only to elliptic curve keys. Defaults to the NIST P-256
          elliptic curve. To create an elliptic curve key, consider using :func:`create_ec_key` instead.
         :paramtype curve: ~azure.keyvault.keys.KeyCurveName or str
+        :keyword int public_exponent: The RSA public exponent to use. Applies only to RSA keys created in a Managed HSM.
```
chlowell, line 61: We should document this on `create_rsa_key` as well, no?

mccoyp: We should! I remember making that update in a dream last night, but I evidently didn't remember to translate that into the material world

chlowell (approved):

PR #18025, opened 2021-04-13
[Communication]: Added exception to begin_update_phone_number_capabilities

lsundaralingam:
This is to address the case when the service returns a response with the status code 200 and an error in the body.

### DIFF
--- a/sdk/communication/azure-communication-phonenumbers/azure/communication/phonenumbers/_phone_numbers_client.py
+++ b/sdk/communication/azure-communication-phonenumbers/azure/communication/phonenumbers/_phone_numbers_client.py
@@ -6,6 +6,7 @@
 
 from typing import TYPE_CHECKING
 from azure.core.tracing.decorator import distributed_trace
+from azure.core.exceptions import HttpResponseError
 from ._generated._phone_numbers_client import PhoneNumbersClient as PhoneNumbersClientGen
 from ._generated.models import PhoneNumberSearchRequest
 from ._shared.utils import parse_connection_str, get_authentication_policy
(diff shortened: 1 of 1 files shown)


sdk/communication/azure-communication-phonenumbers/azure/communication/phonenumbers/_phone_numbers_client.py
```diff
@@ -189,13 +190,19 @@ def begin_update_phone_number_capabilities(
...
-        return self._phone_number_client.phone_numbers.begin_update_capabilities(
+        poller = self._phone_number_client.phone_numbers.begin_update_capabilities(
             phone_number,
             calling=calling,
             sms=sms,
             **kwargs
         )
 
+        result_properties = poller.result().additional_properties
+        if 'status' in result_properties and result_properties['status'] == 'failed':
```
minnieliu, line 201: Is this check case sensitive?

jbeauregardb (approved):
Thank you for doing this!

sdk/communication/azure-communication-phonenumbers/azure/communication/phonenumbers/_phone_numbers_client.py
lsundaralingam: Yes, this check is case sensitive.

lsundaralingam: Discussed this offline. The status value isn't always lower case, so this check should handle that.

beltr0n (approved):

PR #18026, opened 2021-04-14
Correct expiration epoch time

Leoaqr:


sdk/communication/azure-communication-chat/azure/communication/chat/_shared/utils.py
```diff
@@ -17,8 +17,8 @@
 
 
 def _convert_datetime_to_utc_int(expires_on):
-    epoch = time.mktime(datetime(1970, 1, 1).timetuple())
-    return epoch-time.mktime(expires_on.timetuple())
+    epoch = time.localtime().tm_gmtoff
```
sarkar-rajarshi, line 20: Since its not `epoch`, could we rename the variables?
```suggestion
    offset_from_utc = time.localtime().tm_gmtoff
    return time.mktime(input_datetime.timetuple()) + offset_from_utc
```

sdk/communication/azure-communication-chat/azure/communication/chat/_shared/utils.py
```diff
@@ -17,8 +17,8 @@
```
sarkar-rajarshi, line 18: Could we add a docstring to this function as well?
```suggestion
"""
Converts datetime to seconds since the Epoch in UTC.

:param input_datetime: Input datetime
:type input_datetime: datetime
:return: Integer
:rtype: int
"""
```

sdk/communication/azure-communication-chat/azure/communication/chat/_shared/utils.py
```diff
@@ -17,8 +17,8 @@
 
 
 def _convert_datetime_to_utc_int(expires_on):
```
sarkar-rajarshi, line 19:
```suggestion
def _convert_datetime_to_utc_int(input_datetime):
```

sarkar-rajarshi:
Could you also add unit tests to cover this? Also, should this change be propagated across the other packages as well?

sdk/communication/azure-communication-chat/azure/communication/chat/_shared/utils.py
Leoaqr: I would prefer to keep it expires_on, which aligns with our naming in the AccessToken constructor.

annatisch:
Agree with @sarkar-rajarshi - this should have some unit tests including both naïve and TZ-aware inputs.
For reference, this is how these are handled in `msrest` serialization:
https://github.com/Azure/msrest-for-python/blob/master/msrest/serialization.py#L1136-L1152

With validation:

sdk/communication/azure-communication-chat/tests/_shared/test_utils.py
```diff
@@ -0,0 +1,26 @@
+import unittest
```
RezaJooyandeh, line 1: @annatisch where do you think we should put the tests for the shared code?

Leoaqr: I discussed this one with Anna offline before. Since the issue was raised in Chat, so we decide put it in Chat.

sdk/communication/azure-communication-chat/azure/communication/chat/_shared/utils.py
```diff
@@ -6,19 +6,25 @@
 
 import base64
 import json
-import time
 from typing import (  # pylint: disable=unused-import
     cast,
     Tuple,
 )
 from datetime import datetime
+import calendar
```
annatisch, line 14: This file will need to be updated across all the communications libraries (which all have the _shared module)

sdk/communication/azure-communication-chat/tests/_shared/test_utils.py
```diff
@@ -0,0 +1,26 @@
...
+from datetime import datetime
+import dateutil.tz
```
annatisch, line 3: dateutil is not part of the standard Python library - you will need to add this as a dependency in `dev_requirements.txt`.

sdk/communication/azure-communication-chat/tests/_shared/test_utils.py
annatisch, line 1: This file should be moved out of a `_shared` directory into just the `tests` directory. Otherwise I don't think it will be run.

sdk/communication/azure-communication-chat/tests/_shared/test_utils.py
Leoaqr: Updated

sdk/communication/azure-communication-chat/azure/communication/chat/_shared/utils.py
Leoaqr: I have had imported

sdk/communication/azure-communication-chat/tests/_shared/test_utils.py
Leoaqr: There is another set of test_communication_identifier_serializer UT in _shared folder. I validated it locally, it's included in pytest.

annatisch (approved):

Issue #17525, opened 2021-03-24
Parametrized query example


EvgenSk:
[Enter feedback here]
It would be really helpful if the page had an example of parametrized query.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: a472053b-358e-bc0e-4d91-367b5e039642
* Version Independent ID: 7a50ad47-3697-e29e-3f3e-afbead68aaec
* Content: [azure.servicebus.management.SqlRuleFilter class](https://docs.microsoft.com/en-us/python/api/azure-servicebus/azure.servicebus.management.sqlrulefilter?view=azure-python)
* Content Source: [docs-ref-autogen/azure-servicebus/azure.servicebus.management.SqlRuleFilter.yml](https://github.com/MicrosoftDocs/azure-docs-sdk-python/blob/master/docs-ref-autogen/azure-servicebus/azure.servicebus.management.SqlRuleFilter.yml)
* GitHub Login: @lmazuel
* Microsoft Alias: **lmazuel**

chlowell:
Thank you for the feedback, someone from the team will take a look as soon as possible.

yunhaoling:
hey @EvgenSk, thanks for the feedback.

If you're looking for some sample code, we have a test that shows how to use parametrized sql rule filer which could be found [here](https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/servicebus/azure-servicebus/tests/mgmt_tests/test_mgmt_rules.py#L51-L53).

there's also an official service doc on sql rule filter usage: https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-messaging-sql-filter

and we'll consider adding detailed sample in the docstring and mgmt samples.

PR #18042, opened 2021-04-14
[ServiceBus] update samples, docs and tests for sql rule filter

yunhaoling:
Closing issue: https://github.com/Azure/azure-sdk-for-python/issues/17525

### DIFF
--- a/sdk/servicebus/azure-servicebus/tests/async_tests/mgmt_tests/test_mgmt_rules_async.py
+++ b/sdk/servicebus/azure-servicebus/tests/async_tests/mgmt_tests/test_mgmt_rules_async.py
@@ -50,7 +50,7 @@ class ServiceBusAdministrationClientRuleAsyncTests(AzureMgmtTestCase):
             "@param": datetime(2020, 7, 5, 11, 12, 13),
         })
 
-        sql_filter = SqlRuleFilter("Priority = @param1 and Level = @param2", parameters={
+        sql_filter = SqlRuleFilter("Priority = @param1 AND Level = @param2", parameters={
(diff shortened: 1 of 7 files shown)


sdk/servicebus/azure-servicebus/tests/async_tests/mgmt_tests/test_mgmt_rules_async.py
```diff
@@ -50,8 +50,9 @@ async def test_async_mgmt_rule_create(self, servicebus_namespace_connection_stri
             "@param": datetime(2020, 7, 5, 11, 12, 13),
         })
 
-        sql_filter = SqlRuleFilter("Priority = @param1", parameters={
+        sql_filter = SqlRuleFilter("Priority = @param1 and Level = @param2", parameters={
```
swathipil, line 53: does and vs. AND matter?

swathipil (commented):
I think you may need to add the recordings? but o/w LGTM

swathipil (approved):

sdk/servicebus/azure-servicebus/tests/async_tests/mgmt_tests/test_mgmt_rules_async.py
yunhaoling: as far as I know sql syntax should be case-insensitive when working with databases.

but I checked the [service doc](https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-messaging-sql-filter)
 and found that capital words are used for sql syntax so I will align with the doc.

thanks for the heads up on this!

PR #18055, opened 2021-04-15
[Key Vault] Implement 7.3-preview with secure key release

### LINKED ISSUE #13562: Support key export from Key Vault
For Key Vault 7.2-preview, we need to add support to export keys. This would be a method like `KeyClient.ExportKey` that takes the key name, key version (both required), and a body with the `KeyExportParameters` model found in the link below. This only has a single required property named `environment` (model property: `env`), so we should take this as a parameter. This returns a `KeyVaultKey` (model `KeyBundle`).

This requires a number of other changes:

- [ ] Add optional `Exportable` boolean property to `KeyProperties`.
- [ ] Add `ReleasePolicy` property to `KeyProperties` of type `KeyReleasePolicy`.
- [ ] Add `PublicExponent` integer property to `CreateKeyOptions`.
- [ ] Add "export" to `KeyOperation` enum. This will correctly impact the `CreateKeyOptions` class and the `KeyClient.UpdateKey` method..
- [ ] Add `ReleasePolicy` property to `CreateKeyOptions`.
- [ ] Add `ReleasePolicy` property to `ImportKeyOptions`.

See the [service version 7.1 -> 7.2-preview diff](https://portal.azure-devex-tools.com/tools/diff/eyJzb3VyY2VVcmwiOiJodHRwczovL2dpdGh1Yi5jb20vQXp1cmUvYXp1cmUtcmVzdC1hcGktc3BlY3MvYmxvYi9tYXN0ZXIvc3BlY2lmaWNhdGlvbi9rZXl2YXVsdC9kYXRhLXBsYW5lL3JlYWRtZS5tZCIsInRhcmdldFVybCI6Imh0dHBzOi8vZ2l0aHViLmNvbS9BenVyZS9henVyZS1yZXN0LWFwaS1zcGVjcy9ibG9iL21hc3Rlci9zcGVjaWZpY2F0aW9uL2tleXZhdWx0L2RhdGEtcGxhbmUvcmVhZG1lLm1kIiwic291cmNlVGFnIjoicGFja2FnZS03LjEiLCJ0YXJnZXRUYWciOiJwYWNrYWdlLTcuMi1wcmV2aWV3In0=) 🔒 for details.

mccoyp:
Resolves #13562.

This implements the changes for service version 7.3-preview and makes this API version the default. Functional tests for the added key export and key release operations are pending service support and required resource configuration, respectively. Due to a behavior change with key import on vaults, tests that import keys are temporarily pinned to 7.2-preview.


sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_client.py
```diff
@@ -563,3 +592,52 @@ def import_key(self, name, key, **kwargs):
...
+    def export_key(self, name, version, parameters, **kwargs):
+        # type: (str, str, KeyExportParameters, **Any) -> KeyVaultKey
+        """Exports a key.
+
+        The export key operation is applicable to all key types. The target key must be marked
+        exportable. This operation requires the keys/export permission.
+
+        :param str name: The name of the key to export.
+        :param str version: A specific version of the key to export.
+        :param parameters: The parameters for the export operation.
```
heaths, line 606: We might want to include the name and version in the parameters, rather than all three. In fact, in .NET we'd have to change the order anyway, since `version` is optional and we can't have required parameters after optional parameters. But looking at similar cases, we tend to just consolidate into an "options" or "parameters" (in this case, since it's required, probably the latter) parameter.

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_client.py
```diff
@@ -563,3 +592,52 @@ def import_key(self, name, key, **kwargs):
...
+        :raises: :class:`~azure.core.exceptions.HttpResponseError`
+        """
+        key_parameters = self._models.KeyExportParameters(
+            kek=parameters.key._to_generated_model(), enc=parameters.algorithm
+        )
+        bundle = self._client.export(self.vault_url, name, version, key_parameters, **kwargs)
+        return KeyVaultKey._from_key_bundle(bundle)
+
+    @distributed_trace
+    def release_key(self, name, version, parameters, **kwargs):
```
heaths, line 619: Same here.

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_models.py
```diff
@@ -61,6 +61,23 @@ def _to_generated_model(self):
...
+
+    This includes a key and algorithm that are used for encrypting the exported key material.
+
+    :param key: The export encryption key. This must be an RSA key that supports encryption.
+    :type key: ~azure.keyvault.keys.JsonWebKey
+    :keyword str algorithm: The encryption algorithm to use to protected the exported key material. Possible values
+     include: "CKM_RSA_AES_KEY_WRAP", "RSA_AES_KEY_WRAP_256", "RSA_AES_KEY_WRAP_384".
+    """
+
+    def __init__(self, key, **kwargs):
```
heaths, line 75: As mentioned above, we probably want to add `name` and the optional `version` to this.

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_models.py
```diff
@@ -211,6 +237,59 @@ def managed(self):
...
+class KeyReleaseParameters(object):
+    """The parameters for key release.
+
+    :param str environment: The target environment assertion.
+    :keyword str algorithm: The encryption algorithm to use to protect the released key material. Possible values
+     include: "CKM_RSA_AES_KEY_WRAP", "RSA_AES_KEY_WRAP_256", "RSA_AES_KEY_WRAP_384".
+    :keyword str nonce: A client-provided nonce for freshness.
+    """
+
+    def __init__(self, environment, **kwargs):
```
heaths, line 249: Same here.

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_models.py
```diff
@@ -211,6 +237,59 @@ def managed(self):
...
+        # type: (str, **Any) -> None
+        self.environment = environment
+        self.algorithm = kwargs.get("algorithm", None)
+        self.nonce = kwargs.get("nonce", None)
```
heaths, line 253: If not passed, we probably should generate one.

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_models.py
```diff
@@ -211,6 +237,59 @@ def managed(self):
...
+    :type data: bytes
+    :keyword content_type: Content type and version of key release policy. Defaults to 
+     'application/json; charset=utf-8; version=1.0'.
+    :type content_type: str
+    """
+
+    def __init__(self, data, **kwargs):
+        # type: (bytes, **Any) -> None
+        self.data = data
+        self.content_type = kwargs.get("content_type", None) or "application/json; charset=utf-8; version=1.0"
```
heaths, line 269: "; version=1.0" should be removed (recent PR that hasn't merged yet). Though, we really shouldn't code defaults from the swagger. The service should use its defaults if not serialized and sent to the service. As written now, this creates a tight coupling we want to avoid.

heaths (approved):
I left a few comments on the 32199ca280f6e2e052fb1e21f0605a7c152c012a commit. We should change the shape a little based on those comments, but overall it looks good. I think this is fine as-is for now to unblock the CLI/PS.

chlowell (approved):
Looks good to me. I left a few initial comments we can revisit later, none of them block a private preview.

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_client.py
mccoyp: The remaining parameters are all optional according to the latest [swagger changes](https://github.com/Azure/azure-rest-api-specs/pull/13913) so I think it makes the most sense to make them kwargs here and drop the class. For other languages that use a parameter object pattern, it might make more sense to move the key name/version into `KeyExportParameters`, like you said.

heaths: We actually hoist up some optional parameters like `version` because they are important and common. We do that in other APIs. Maybe not `parameters`, but it's something to discuss further in the channel. For now, not blocking.

Issue #16725, opened 2021-02-12
[EventHubs] combine various connection string parsing logic


swathipil:
similar to : #13683

[`_connection_string_parser.py`](https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/eventhub/azure-eventhub/azure/eventhub/_connection_string_parser.py) holds redundant/duplicate parsing logic, and should be combined with [`_client_base.py`](https://github.com/Azure/azure-sdk-for-python/blob/891d7aaef5b2d61735c3521ee294f1582824e6e2/sdk/eventhub/azure-eventhub/azure/eventhub/_client_base.py#L46) logic for parsing connection strings. Should be very similar to SB PR here: #16464 

Doing so will help improve maintainability.

PR #18059, opened 2021-04-15
[Event Hubs] combine conn str parsing logic

swathipil:
fixes: #16725

### DIFF
--- a/sdk/eventhub/azure-eventhub/azure/eventhub/_client_base.py
+++ b/sdk/eventhub/azure-eventhub/azure/eventhub/_client_base.py
@@ -106,7 +106,9 @@ def _parse_conn_str(conn_str, **kwargs):
             "Invalid connection string. Should be in the format: "
             "Endpoint=sb://<FQDN>/;SharedAccessKeyName=<KeyName>;SharedAccessKey=<KeyValue>"
         )
-    if shared_access_signature and shared_access_key:
+    # Only connection string parser should check that only one of sas and shared access
+    # key exists. For backwards compatibility, client construction should not have this check.
+    if check_case and shared_access_signature and shared_access_key:
         raise ValueError(
             "Only one of the SharedAccessKey or SharedAccessSignature must be present."
         )
--- a/sdk/eventhub/azure-eventhub/azure/eventhub/_client_base.py
+++ b/sdk/eventhub/azure-eventhub/azure/eventhub/_client_base.py
@@ -20,6 +20,7 @@ except ImportError:
 
 from uamqp import AMQPClient, Message, authentication, constants, errors, compat, utils
 import six
+from azure.core.utils import parse_connection_string as core_parse_connection_string
 from azure.core.credentials import AccessToken, AzureSasCredential
 
 from .exceptions import _handle_exception, ClientClosedError, ConnectError
@@ -43,7 +44,7 @@ _Address = collections.namedtuple("Address", "hostname path")
 _AccessToken = collections.namedtuple("AccessToken", "token expires_on")
 
 
-def _parse_conn_str(conn_str, kwargs):
+def _parse_conn_str(conn_str, **kwargs):
     # type: (str, Dict[str, Any]) -> Tuple[str, Optional[str], Optional[str], str, Optional[str], Optional[int]]
     endpoint = None
     shared_access_key_name = None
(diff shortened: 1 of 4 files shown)


sdk/eventhub/azure-eventhub/azure/eventhub/_client_base.py
```diff
@@ -20,6 +20,7 @@
 
 from uamqp import AMQPClient, Message, authentication, constants, errors, compat, utils
 import six
+from azure.core.utils import parse_connection_string as core_parse_connection_string
```
yunhaoling, line 23: can we double check the azure-core dependency version is right (>= the version we introduced this `parse_connection_string` method)?

swathipil: yep! checked and it is

sdk/eventhub/azure-eventhub/azure/eventhub/_client_base.py
```diff
@@ -52,38 +53,66 @@ def _parse_conn_str(conn_str, kwargs):
...
+    if check_case:
+        shared_access_key = conn_settings.get("SharedAccessKey")
+        shared_access_key_name = conn_settings.get("SharedAccessKeyName")
+        endpoint = conn_settings.get("Endpoint")
+        entity_path = conn_settings.get("EntityPath")
+    
+        # non case sensitive check when parsing connection string for internal use
+        for key, value in conn_settings.items():
+            # only sas check is non case sensitive for both conn str properties and internal use
+            if key.lower() == "sharedaccesssignature":
```
yunhaoling, line 67: if conn_settings is dict, then we don't need to iterate every key-value pair in the items?
```python
signature = conn_settings.get("sharedaccesssignature") or conn_settings.get("SharedAccessSignature")
```
but if users pass weird case like `shareDacceSSSignaturE`, then we're not able to parse it.. just thinking out loud.

sdk/eventhub/azure-eventhub/azure/eventhub/_client_base.py
```diff
@@ -52,38 +53,66 @@ def _parse_conn_str(conn_str, kwargs):
...
+                    # An arbitrary, absurdly large number, since you can't renew.
+                    shared_access_signature_expiry = int(time.time() * 2)
+
+    if not check_case:
+        endpoint = conn_settings.get("endpoint") or conn_settings.get("hostname")
+        if endpoint:
+            endpoint = endpoint.rstrip("/")
+        shared_access_key_name = conn_settings.get("sharedaccesskeyname")
+        shared_access_key = conn_settings.get("sharedaccesskey")
+        entity_path = conn_settings.get('entitypath')
```
yunhaoling, line 88: nit: ' -> " for consistency
```suggestion
        entity_path = conn_settings.get("entitypath")
```

sdk/eventhub/azure-eventhub/azure/eventhub/_client_base.py
```diff
@@ -52,38 +53,66 @@ def _parse_conn_str(conn_str, kwargs):
...
+                except (
+                    IndexError,
+                    TypeError,
+                    ValueError,
+                ):  # Fallback since technically expiry is optional.
+                    # An arbitrary, absurdly large number, since you can't renew.
+                    shared_access_signature_expiry = int(time.time() * 2)
+
+    if not check_case:
+        endpoint = conn_settings.get("endpoint") or conn_settings.get("hostname")
```
yunhaoling, line 83: do you know if there any cases when both `endpoint` and `hostname` are provided?

sdk/eventhub/azure-eventhub/azure/eventhub/_client_base.py
```diff
@@ -52,38 +53,66 @@ def _parse_conn_str(conn_str, kwargs):
...
+        raise ValueError("Invalid Endpoint on the Connection String.")
+    host = cast(str, parsed.netloc.strip())
+
+    if any([shared_access_key, shared_access_key_name]) and not all(
+        [shared_access_key, shared_access_key_name]
+    ):
         raise ValueError(
-            "Invalid connection string. Should be in the format: "
-            "Endpoint=sb://<FQDN>/;SharedAccessKeyName=<KeyName>;SharedAccessKey=<KeyValue>"
+            "Connection string must have both SharedAccessKeyName and SharedAccessKey."
```
yunhaoling, line 105: may I ask why we're change the information in the error here? -- the previous one has some value in telling the right format about the connection string.

sdk/eventhub/azure-eventhub/azure/eventhub/_client_base.py
swathipil: I don't know of any cases where both would appear. I think the logic here is consistent with the previous logic, since it was checking if `endpoint` was in the connection string, else checking for `hostname`.

sdk/eventhub/azure-eventhub/azure/eventhub/_client_base.py
swathipil: Yeah, so I was trying to conform with the [error message in SB](https://github.com/Azure/azure-sdk-for-python/blob/b682f616902ec1ab969de3b47f41ccfa5055adad/sdk/servicebus/azure-servicebus/azure/servicebus/_base_handler.py#L113).
My logic then was that the `Should be in format` error message seemed to be saying that the connection string must include `SharedAccessKeyName` and `SharedAccessKey` in any case, even though they are not required if `SharedAccessSignature` is included. This error message seemed more clear that if one is included, both should be included. I'll try to combine the two to provide more info.

sdk/eventhub/azure-eventhub/azure/eventhub/_client_base.py
swathipil: potentially we could call the core parser again and get the sas? IE)
```conn_settings = core_parse_connection_string(conn_str, case_sensitive_keys=False)
  shared_access_signature = conn_settings.get("sharedaccesssignature")
```
although that's not really any more efficient than looping through...

sdk/eventhub/azure-eventhub/azure/eventhub/_client_base.py
```diff
@@ -43,47 +44,77 @@
...
-    else:
-        host = str(endpoint)
+    if shared_access_signature and shared_access_key:
+        raise ValueError(
+            "Only one of the SharedAccessKey or SharedAccessSignature must be present."
+        )
+    if not shared_access_signature and not shared_access_key:
+        raise ValueError(
+            "At least one of the SharedAccessKey or SharedAccessSignature must be present."
+        )
```
yunhaoling, line 116: If I'm understanding correctly, `def _parse_conn_str(conn_str, **kwargs):` is now the shared method used by 
1. connection string parser
2. client construction
while previously we had two separate part.

but this part of logic seems to only lies in the connection string parser before, so I think is this actually a breaking change here to the user?

example:
let's say user has a connection string:
`Endpoint=....;SharedAccessKeyName=...;SharedAccessKey=...;SharedAccessSignature=;`
it would work in `EventHubProducerClient.from_connection_string(...)` previously while after the change user would get ValueError.

swathipil: yes, I think that's true for line 109, but line 113 previously raised ValueError too (in the case that neither sas nor key was included). Since only the connection string parser logic requires the check for line 109 and it checks case, does this make sense to you for line 109?
`if check_case and shared_access_signature and shared_access_key:`

swathipil: or I can move the check from 109 to 77.

yunhaoling: I'm good with `if check_case and shared_access_signature and shared_access_key` as long as we don't break users. ( I don't think moving from 109 to 77 should matter?)

probably adding a comment saying that connection string parser (in which case check_case is True) needs checking shared_access_signature and shared_access_key while client construction should not have this for backward compatibility.

yunhaoling (approved):

PR #18075, opened 2021-04-15
[Tables] Test fixes

seankane-msft:
Removes tests that are supported in tables but not cosmos.


sdk/tables/azure-data-tables/azure/data/tables/_table_client.py
```diff
@@ -455,7 +455,7 @@ def list_entities(
         """
         user_select = kwargs.pop("select", None)
         if user_select and not isinstance(user_select, str):
-            user_select = ", ".join(user_select)
+            user_select = ",".join(user_select)
```
seankane-msft, line 458: Removed the space because cosmos will keep the space there on the returned object, which makes the deserialized object have a parameter with a preceding space. ie. ` value` instead of `value`.

annatisch (approved):

PR #18103, opened 2021-04-16
changing table_name to name

seankane-msft:
#18101


annatisch (approved):
Changes look good - though needs to be added to the changelog as a breaking change.

PR #18107, opened 2021-04-16
[Container Registry] Acr token cache

seankane-msft:


sdk/containerregistry/azure-containerregistry/azure/containerregistry/_exchange_client.py
```diff
@@ -47,29 +48,38 @@ def __init__(self, endpoint, credential, **kwargs):
...
         parsed_challenge = _parse_challenge(challenge)
-        refresh_token = self.exchange_aad_token_for_refresh_token(service=parsed_challenge["service"])
+        refresh_token = self.get_refresh_token(parsed_challenge["service"], **kwargs)
         return self.exchange_refresh_token_for_access_token(
             refresh_token, service=parsed_challenge["service"], scope=parsed_challenge["scope"], **kwargs
         )
 
+    def get_refresh_token(self, service, **kwargs):
+        # type: (str, **Any) -> str
+        if not self._refresh_token or datetime.now() - self._refresh_expiry > timedelta(minutes=5):
```
chlowell, line 74: `datetime.now()` returns a naive datetime (having no timezone information). This calculation therefore doesn't consider daylight saving time, causing this method to do the wrong thing when called on both sides of a DST transition. I recommend `time.time()` instead--you don't appear to need a `datetime` here anyway.

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_exchange_client.py
```diff
@@ -47,29 +48,38 @@ def __init__(self, endpoint, credential, **kwargs):
...
         self._endpoint = endpoint
-        self._credential_scopes = "https://management.core.windows.net/.default"
+        self._credential_scope = "https://management.core.windows.net/.default"
         self._client = ContainerRegistry(
             credential=credential,
             url=endpoint,
             sdk_moniker=USER_AGENT,
             authentication_policy=ExchangeClientAuthenticationPolicy(),
-            credential_scopes=kwargs.pop("credential_scopes", self._credential_scopes),
+            credential_scopes=kwargs.pop("credential_scopes", self._credential_scope),
```
chlowell, line 57: If "credential_scopes" is supposed to be public, should it be "credential_scope"?

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_exchange_client.py
```diff
@@ -47,29 +48,38 @@ def __init__(self, endpoint, credential, **kwargs):
...
+        refresh_token = self.get_refresh_token(parsed_challenge["service"], **kwargs)
         return self.exchange_refresh_token_for_access_token(
             refresh_token, service=parsed_challenge["service"], scope=parsed_challenge["scope"], **kwargs
         )
 
+    def get_refresh_token(self, service, **kwargs):
+        # type: (str, **Any) -> str
+        if not self._refresh_token or time.time() - self._refresh_expiry > 300:
+            self._refresh_token = self.exchange_aad_token_for_refresh_token(service, **kwargs)
+            self._refresh_expiry = time.time()
```
chlowell, line 76:
```suggestion
            self._last_refresh_time = time.time()
```

sdk/containerregistry/azure-containerregistry/azure/containerregistry/aio/_async_exchange_client.py
```diff
@@ -47,28 +48,37 @@ def __init__(self, endpoint, credential, **kwargs):
...
+            credential_scopes=kwargs.pop("credential_scopes", self._credential_scope),
             **kwargs
         )
         self._credential = credential
+        self._refresh_token = None
+        self._refresh_expiry = None
 
     async def get_acr_access_token(self, challenge, **kwargs):
-        # type: (str) -> str
+        # type: (str, Dict[str, Any]) -> str
```
chlowell, line 65: Is there a reason to use `Dict` here for kwargs instead of `**Any`? Also, definitely Python 3 here, you could use the newer syntax.

sdk/containerregistry/azure-containerregistry/azure/containerregistry/aio/_async_exchange_client.py
```diff
@@ -47,28 +48,37 @@ def __init__(self, endpoint, credential, **kwargs):
...
         parsed_challenge = _parse_challenge(challenge)
-        refresh_token = await self.exchange_aad_token_for_refresh_token(service=parsed_challenge["service"])
+        refresh_token = await self.get_refresh_token(parsed_challenge["service"], **kwargs)
         return await self.exchange_refresh_token_for_access_token(
             refresh_token, service=parsed_challenge["service"], scope=parsed_challenge["scope"], **kwargs
         )
 
+    async def get_refresh_token(self, service, **kwargs):
+        # type: (str, **Any) -> str
+        if not self._refresh_token or datetime.now() - self._refresh_expiry > timedelta(minutes=5):
```
chlowell, line 74: Same DST problem exists here

pallavit (approved):
:shipit:

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_exchange_client.py
```diff
@@ -47,29 +48,38 @@ def __init__(self, endpoint, credential, **kwargs):
...
             url=endpoint,
             sdk_moniker=USER_AGENT,
             authentication_policy=ExchangeClientAuthenticationPolicy(),
-            credential_scopes=kwargs.pop("credential_scopes", self._credential_scopes),
+            credential_scopes=kwargs.pop("credential_scopes", self.credential_scope),
             **kwargs
         )
         self._credential = credential
+        self._refresh_token = None
+        self._last_refresh_time = None
```
chlowell, line 62: Nitpicking, I don't think a TypeError is possible but I'd still write this:
```suggestion
        self._last_refresh_time = 0
```

chlowell (approved):

sdk/containerregistry/azure-containerregistry/azure/containerregistry/aio/_async_exchange_client.py
```diff
@@ -42,40 +40,48 @@ class ACRExchangeClient(object):
...
             url=endpoint,
             sdk_moniker=USER_AGENT,
             authentication_policy=ExchangeClientAuthenticationPolicy(),
-            credential_scopes=kwargs.pop("credential_scopes", self._credential_scopes),
+            credential_scopes=kwargs.pop("credential_scopes", self._credential_scope),
             **kwargs
         )
         self._credential = credential
+        self._refresh_token = None
+        self._last_refresh_time = None
```
chlowell, line 60: :suspect:

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_exchange_client.py
```diff
@@ -47,29 +48,38 @@ def __init__(self, endpoint, credential, **kwargs):
...
         parsed_challenge = _parse_challenge(challenge)
-        refresh_token = self.exchange_aad_token_for_refresh_token(service=parsed_challenge["service"])
+        refresh_token = self.get_refresh_token(parsed_challenge["service"], **kwargs)
         return self.exchange_refresh_token_for_access_token(
             refresh_token, service=parsed_challenge["service"], scope=parsed_challenge["scope"], **kwargs
         )
 
+    def get_refresh_token(self, service, **kwargs):
+        # type: (str, **Any) -> str
+        if not self._refresh_token or time.time() - self._last_refresh_time > 300:
```
jeremymeng, line 74: I think it's fair but want to confirm: we are assuming that `service` would be the same for a container registry client thus caching is fine?

seankane-msft: @jeremymeng yes that is the assumption

Issue #17954, opened 2021-04-10
[translation] add headers/query params to logging allowlist


kristapratico:
Here is a good example of what's getting redacted: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=832300&view=logs&j=6a36fa0f-00d1-5871-8873-802f6b762b9b&t=aa427338-b554-530f-3345-9ed388620bbb

Stuff like apim-request-id, etc. are useful in debugging and should be un-redacted.

PR #18117, opened 2021-04-16
[translation] unredact headers and query params for logging

kristapratico:
Resolves https://github.com/Azure/azure-sdk-for-python/issues/17954

Makes sure we do not redact non-sensitive information that could be helpful when debugging.

### DIFF
--- a/sdk/translation/azure-ai-translation-document/azure/ai/translation/document/_client.py
+++ b/sdk/translation/azure-ai-translation-document/azure/ai/translation/document/_client.py
@@ -20,6 +20,7 @@ from ._models import (
 )
 from ._user_agent import USER_AGENT
 from ._polling import TranslationPolling
+from ._helpers import get_http_logging_policy
 if TYPE_CHECKING:
     from azure.core.paging import ItemPaged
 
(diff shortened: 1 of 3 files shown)


sdk/translation/azure-ai-translation-document/azure/ai/translation/document/_client.py
```diff
@@ -62,12 +63,14 @@ def __init__(self, endpoint, credential, **kwargs):
...
             name=COGNITIVE_KEY_HEADER, credential=credential
         )
+
         self._client = _BatchDocumentTranslationClient(
             endpoint=endpoint,
             credential=credential,  # type: ignore
             api_version=self._api_version,
             sdk_moniker=USER_AGENT,
             authentication_policy=authentication_policy,
+            http_logging_policy=get_http_logging_policy(),
```
maririos, line 73: asking as I don't know how this works on Python... if the service introduces new headers/query parameters, this needs to be updated? or will the user be able to add more logging policies if they desire? -> it might not be needed if the list is always updated though

kristapratico: There is a list of allowed headers in the policy, but it's general and doesn't include service-specific headers or query params. It can always be updated if the service starts sending anything new.

maririos (approved):

PR #18119, opened 2021-04-16
Setup INT live tests

sacheu:


sdk/communication/test-resources.json
```diff
@@ -13,12 +13,34 @@
...
                 "description": "The client OID to grant access to test resources."
             }
+        },
+        "tenantId": {
+            "type": "String",
+            "metadata": {
+                "description": "The tenant id to which the application and resources belong."
         }
+      },
+      "testApplicationId": {
```
benbp, line 32: You have invalid json nesting here. These keys are not located within the `parameters` object.

benbp (changes requested):

sdk/communication/tests.yml
```diff
@@ -16,7 +16,9 @@ stages:
       parameters:
         CloudConfig:
           Public:
-            SubscriptionConfiguration: $(sub-config-communication-services-cloud-test-resources)
+            SubscriptionConfiguration:
```
benbp, line 19: This needs to be SubscriptionConfigurations

```suggestion
            SubscriptionConfigurations:
```

benbp (changes requested):

sdk/communication/tests.yml
benbp: Also, once we have all your PRs up, we should do the switch to use `sub-config-communication-services-cloud-test-resources` instead of `sub-config-communication-services-cloud-test-resources-test` before merging.

Issue #17899, opened 2021-04-08
[Service Bus] Sample needed for Dead Letter Queues


ramya-rao-a:
The doc [Overview of dead letter queues](https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-dead-letter-queues) links to a sample that uses the older .NET package WindowsAzure.ServiceBus

Ideally, there should be links to samples that use packages from all languages. This issue is to add a corresponding sample that uses the `azure-servicebus` package so that it can be linked appropriately.

Reference: [DeadletterQueue](https://github.com/Azure/azure-service-bus/tree/master/samples/DotNet/Microsoft.ServiceBus.Messaging/DeadletterQueue)

Adding this sample will also be useful for us to re-direct readers who land at the above sample.

cc @spelluru

PR #18123, opened 2021-04-18
[Service Bus] create dead-letter queue sample

swathipil:
fixes: #17899


sdk/servicebus/azure-servicebus/samples/async_samples/deadletter_queue_async.py
```diff
@@ -0,0 +1,116 @@
+#!/usr/bin/env python
+
+# --------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for license information.
+# --------------------------------------------------------------------------------------------
+
+"""
+Example to show receiving dead-lettered messages from a Service Bus Queue asynchronously.
```
yunhaoling, line 9: counterpart in sync?

yunhaoling, line 9: also we already had [receive deadletter sample](https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/servicebus/azure-servicebus/samples/sync_samples/receive_deadlettered_messages.py)

is there any difference between those two?
- could we merge two?
- if we could not merge, maybe we should rename the filename and description to reflect the different purpose?

sdk/servicebus/azure-servicebus/samples/async_samples/deadletter_queue_async.py
```diff
@@ -0,0 +1,116 @@
...
+            for i in range(num_messages)
+        ]
+        await sender.send_messages(msg)
+        print("Messages sent")
+
+async def exceed_max_delivery(servicebus_client):
+    receiver = servicebus_client.get_queue_receiver(queue_name=QUEUE_NAME)
+    dlq_receiver = servicebus_client.get_queue_receiver(queue_name=QUEUE_NAME, 
+                                                        sub_queue=ServiceBusSubQueue.DEAD_LETTER,
+                                                        prefetch_count=10)
```
yunhaoling, line 40: same `prefetch_count` comment

sdk/servicebus/azure-servicebus/samples/async_samples/deadletter_queue_async.py
```diff
@@ -0,0 +1,116 @@
...
+            if msg.subject and msg.subject == "Good":
+                print("Received fixed message: Body={}, Subject={}".format(next(msg.body), msg.subject))
+                await receiver.complete_message(msg)
+
+async def main():
+    servicebus_client = ServiceBusClient.from_connection_string(conn_str=CONNECTION_STR)
+    receiver = servicebus_client.get_queue_receiver(queue_name=QUEUE_NAME)
+    dlq_receiver = servicebus_client.get_queue_receiver(queue_name=QUEUE_NAME, 
+                                                        sub_queue=ServiceBusSubQueue.DEAD_LETTER,
+                                                        prefetch_count=100)
```
yunhaoling, line 101: same `prefetch_count` comment

sdk/servicebus/azure-servicebus/samples/async_samples/deadletter_queue_async.py
```diff
@@ -0,0 +1,116 @@
...
+                                                        sub_queue=ServiceBusSubQueue.DEAD_LETTER,
+                                                        prefetch_count=10)
+    async with receiver:
+        peek_msgs = await receiver.peek_messages()
+        while len(peek_msgs) > 0:
+            received_msgs = await receiver.receive_messages()
+            for msg in received_msgs:
+                print("Picked up message: delivery_count {}".format(msg.delivery_count))
+                await receiver.abandon_message(msg)
+            peek_msgs = await receiver.peek_messages()
```
yunhaoling, line 48: interesting.. so we're using peek message as a condition to drain message in the queue.
this is migrated from the .net sample?

sdk/servicebus/azure-servicebus/samples/async_samples/deadletter_queue_async.py
```diff
@@ -0,0 +1,116 @@
...
+    receiver = servicebus_client.get_queue_receiver(queue_name=QUEUE_NAME)
+    dlq_receiver = servicebus_client.get_queue_receiver(queue_name=QUEUE_NAME, 
+                                                        sub_queue=ServiceBusSubQueue.DEAD_LETTER,
+                                                        prefetch_count=10)
+    msgs_to_send = []
+    async with dlq_receiver:
+        received_dlq_msgs = await dlq_receiver.receive_messages(max_message_count=10)
+        for msg in received_dlq_msgs:
+            if msg.subject and msg.subject == "Bad":
+                msg_copy = ServiceBusMessage(next(msg.body), subject="Good")
```
yunhaoling, line 82: You could use str(msg.body) which returns the string representation of the body
```suggestion
                msg_copy = ServiceBusMessage(str(msg.body), subject="Good")
```

sdk/servicebus/azure-servicebus/samples/async_samples/deadletter_queue_async.py
swathipil: I assumed since this was a standalone sample that's needed for docs + re-directing when looking at the .NET sample, it wouldn't require a counterpart. But will add 👍

swathipil: It looks like Josh asked the same question, and Ramya said that it would be better to have this be a standalone sample separate from the other dead-letter one [here](https://github.com/Azure/azure-sdk-for-net/issues/20212#issuecomment-820803527). 

I'm also not sure if merging the two would be the best idea, only because the other deadletter sample is very simple and direct. I think merging it with this more comprehensive one would make it harder for users who just want to copy and paste to make this work.

I can add a link and more detailed description in the docstring here to show what this sample is for, so that users don't get confused as to why there are two.

sdk/servicebus/azure-servicebus/samples/async_samples/deadletter_queue_async.py
```diff
@@ -0,0 +1,116 @@
...
+                    reason="ProcessingError",
+                    error_description="Don't know what to do with this message.",
+                )
+
+async def fix_deadletters(servicebus_client):
+    sender = servicebus_client.get_queue_sender(queue_name=QUEUE_NAME)
+    receiver = servicebus_client.get_queue_receiver(queue_name=QUEUE_NAME)
+    dlq_receiver = servicebus_client.get_queue_receiver(queue_name=QUEUE_NAME, 
+                                                        sub_queue=ServiceBusSubQueue.DEAD_LETTER,
+                                                        prefetch_count=10)
```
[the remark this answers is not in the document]
swathipil: sounds good, will remove prefetch_count

sdk/servicebus/azure-servicebus/samples/async_samples/deadletter_queue_async.py
swathipil: The .net sample has a [`while True: receive if message not null; else break`](https://github.com/Azure/azure-service-bus/blob/c595bf8de42d41208a76adbe7329a8366466ed86/samples/DotNet/Microsoft.ServiceBus.Messaging/DeadletterQueue/Program.cs#L102) pattern, which isn't our pattern.
I used `len(peek_msgs)` here to check if messages were in the queue, rather than `len(received_msgs)` because `await receiver.receive_messages` was waiting forever. But I just realized that was because I didn't set `max_wait_time`. 🤦

sdk/servicebus/azure-servicebus/samples/async_samples/deadletter_queue_async.py
swathipil: `str(msg.body)` returns the string `b'<generator object DataBody.data at 0x0000020682C79040>'`, but str(msg) works

yunhaoling: yeah.. msg.body is a generator object, str(message) should be good

sdk/servicebus/azure-servicebus/samples/async_samples/deadletter_queue_async.py
yunhaoling: I'm good with keeping them separate and thanks for the description -- very illustrative.

The only small concern I have is for the file name -- `deadletter_queue<_async>.py` feels like no difference compared to our existing one `receive_deadlettered_messages<_async>.py`.

something like this <this_is_a_very_very_comprehensive_deadletter_queue_sample.py>
(just kidding 😄, use your imagination)

yunhaoling:
don't forget to run the livetest pipeline 😉

yunhaoling (approved):
LGTM (assuming livetest passing)

PR #18128, opened 2021-04-19
[Service Bus] add rule filters sample

swathipil:
fixes: #17898

### DIFF
--- a/sdk/servicebus/azure-servicebus/samples/async_samples/topic_subscription_with_rule_operations.py
+++ b/sdk/servicebus/azure-servicebus/samples/async_samples/topic_subscription_with_rule_operations.py
@@ -1,0 +1,159 @@
+#!/usr/bin/env python
+
+# --------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for license information.
+# --------------------------------------------------------------------------------------------
+
+"""
+Example to show receiving batch messages from a Service Bus Subscription under specific Topic asynchronously.
+"""
+
+# pylint: disable=C0111
+
+import os
+import time
+import asyncio
+from azure.core.exceptions import ResourceNotFoundError, ResourceExistsError
+from azure.servicebus.aio.management import ServiceBusAdministrationClient
+from azure.servicebus.management import (
+    TrueRuleFilter,
+    SqlRuleFilter,
+    SqlRuleAction,
+    CorrelationRuleFilter
+)
+from azure.servicebus.aio import ServiceBusClient
+from azure.servicebus import ServiceBusMessage
+
+CONNECTION_STR = os.environ['SERVICE_BUS_CONNECTION_STR']
+TOPIC_NAME = os.environ["SERVICE_BUS_TOPIC_NAME"]
+ALL_MSGS_SUBSCRIPTION_NAME = os.environ["SERVICE_BUS_ALL_MSGS_SUB_NAME"]
+SQL_FILTER_ONLY_SUBSCRIPTION_NAME = os.environ["SERVICE_BUS_SQL_FILTER_ONLY_SUB_NAME"]
+SQL_FILTER_WITH_ACTION_SUBSCRIPTION_NAME = os.environ["SERVICE_BUS_SQL_FILTER_ACTION_SUB_NAME"]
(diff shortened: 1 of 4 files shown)


sdk/servicebus/azure-servicebus/samples/async_samples/topic_subscription_with_rule_operations.py
```diff
@@ -0,0 +1,159 @@
+#!/usr/bin/env python
+
+# --------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for license information.
+# --------------------------------------------------------------------------------------------
+
+"""
+Example to show receiving batch messages from a Service Bus Subscription under specific Topic asynchronously.
```
yunhaoling, line 9: nit: async sample should have "_async" suffix in the file name.

also counterpart in sync sample?

sdk/servicebus/azure-servicebus/samples/async_samples/topic_subscription_with_rule_operations.py
```diff
@@ -0,0 +1,159 @@
...
+)
+from azure.servicebus.aio import ServiceBusClient
+from azure.servicebus import ServiceBusMessage
+
+CONNECTION_STR = os.environ['SERVICE_BUS_CONNECTION_STR']
+TOPIC_NAME = os.environ["SERVICE_BUS_TOPIC_NAME"]
+ALL_MSGS_SUBSCRIPTION_NAME = os.environ["SERVICE_BUS_ALL_MSGS_SUB_NAME"]
+SQL_FILTER_ONLY_SUBSCRIPTION_NAME = os.environ["SERVICE_BUS_SQL_FILTER_ONLY_SUB_NAME"]
+SQL_FILTER_WITH_ACTION_SUBSCRIPTION_NAME = os.environ["SERVICE_BUS_SQL_FILTER_ACTION_SUB_NAME"]
+CORRELATION_FILTER_SUBSCRIPTION_NAME = os.environ["SERVICE_BUS_CORR_FILTER_SUB_NAME"]
```
yunhaoling, line 33: great work! but I'm wondering what's the proper way to approach this (sample requests).
The major concern I have is that in the long term, when we find more samples are needed, and we choose to configure "test-resources.json" to include more envs, then our "test-resources.json" would expand and become difficult to maintain as there would be many many variables for different samples.

do you think it would be better to dynamically create those subscriptions in the code by using the `ServiceBusAdministrationClient` instead of configuring our `test-resources.json` each time?

swathipil: that's a really good point, and makes a lot more sense! will change 👍

sdk/servicebus/azure-servicebus/samples/async_samples/topic_subscription_with_rule_operations_async.py
```diff
@@ -0,0 +1,186 @@
...
+)
+from azure.servicebus.aio import ServiceBusClient
+from azure.servicebus import ServiceBusMessage
+
+CONNECTION_STR = os.environ['SERVICE_BUS_CONNECTION_STR']
+TOPIC_NAME = os.environ["SERVICE_BUS_TOPIC_NAME"]
+ALL_MSGS_SUBSCRIPTION_NAME = os.environ["SERVICE_BUS_ALL_MSGS_SUB_NAME"]
+SQL_FILTER_ONLY_SUBSCRIPTION_NAME = os.environ["SERVICE_BUS_SQL_FILTER_ONLY_SUB_NAME"]
+SQL_FILTER_WITH_ACTION_SUBSCRIPTION_NAME = os.environ["SERVICE_BUS_SQL_FILTER_ACTION_SUB_NAME"]
+CORRELATION_FILTER_SUBSCRIPTION_NAME = os.environ["SERVICE_BUS_CORR_FILTER_SUB_NAME"]
```
yunhaoling, line 33: I think those should be constant strings instead of os envs as we removed them from the `test-resources.json`, otherwise wouldn't we get error in the sample test running?

yunhaoling:
looks good to me overall (apart from reading sub name from os env), let's kick off the livetest for this

yunhaoling (approved):

PR #18130, opened 2021-04-19
Simplify regex for parsing `www-authenticate` header

jiasli:


sdk/core/azure-mgmt-core/azure/mgmt/core/policies/_authentication.py
```diff
@@ -83,8 +83,8 @@ def _parse_claims_challenge(challenge):
 
 # these expressions are for challenges with comma delimited parameters having quoted values, e.g.
 # Bearer authorization="https://login.microsoftonline.com/", resource="https://vault.azure.net"
-_AUTHENTICATION_CHALLENGE = re.compile(r'(?:(\w+) ((?:\w+=".*?"(?:, )?)+)(?:, )?)')
-_CHALLENGE_PARAMETER = re.compile(r'(?:(\w+)="([^"]*)")+')
+_AUTHENTICATION_CHALLENGE = re.compile(r'(\w+) ((?:\w+="[^"]*"(?:, )?)+)')
```
jiasli, line 86:
```
(?:(\w+) ((?:\w+=".*?"(?:, )?)+)(?:, )?)
^1                ^2            ^3
```


1. The surrounding non-capturing group `(?:)` is unnecessary.
2. Non-greedy match `.*?` is slower than greedy match `[^"]*` as it causes recursion.
3. The last `(?:, )?` is already captured by the first `(?:, )?`.

Note: This regex is used for capturing `www-authenticate` which may appear multiple times. Per https://github.com/psf/requests/issues/1415, the values are joined by `, `

https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2

> Multiple message-header fields with the same field-name MAY be present in a message if and only if the entire field-value for that header field is defined as a **comma-separated list**.

However, the format of `WWW-Authenticate` contradicts RFC 2616 (it is not a **comma-separated list**).

https://docs.microsoft.com/en-us/azure/active-directory/develop/app-resilience-continuous-access-evaluation#the-code

```
Bearer
  authorization_uri="https://login.windows.net/common/oauth2/authorize",
  error="insufficient_claims",
  claims="eyJhY2Nlc3NfdG9rZW4iOnsibmJmIjp7ImVzc2VudGlhbCI6dHJ1ZSwgInZhbHVlIjoiMTYwNDEwNjY1MSJ9fX0="
```

sdk/core/azure-mgmt-core/azure/mgmt/core/policies/_authentication.py
```diff
@@ -83,8 +83,8 @@ def _parse_claims_challenge(challenge):
...
+_CHALLENGE_PARAMETER = re.compile(r'(\w+)="([^"]*)"')
```
jiasli, line 87: The surrounding `(?:)+` is unnecessary.

sdk/core/azure-mgmt-core/azure/mgmt/core/policies/_authentication.py
jiasli: The public doc [How to use Continuous Access Evaluation enabled APIs in your applications](https://docs.microsoft.com/en-us/azure/active-directory/develop/app-resilience-continuous-access-evaluation#the-code) reads:

> With CAE, APIs will return a 401 status and *a* WWW-Authenticate header when the access token has been revoked or the API detects a change in IP address used.

The internal doc [Claims Challenge.docx](https://microsoft.sharepoint.com/teams/aad/protocols/Shared%20Documents/Claims%20Challenge.docx?web=1) reads:

> The 401 response MAY contain *more than one* www-authenticate header. 

They contradicts each other.

chlowell:
I wrote this regex for a broader scenario than the one we ultimately decided to support. It could definitely be simpler, even simpler than this simplification. I expect it's reasonable to parse the challenge without a regex because we can assume the challenge is from ARM and want only its "claims" parameter. I intended to revisit that ahead of releasing a final azure-mgmt-core beta next month. @jiasli, do you want to have a go at simplifying it even further in this PR?

chlowell:
Closing this because 1.3.0b3 is out now, with simpler challenge parsing ([no regex](https://github.com/Azure/azure-sdk-for-python/blob/azure-mgmt-core_1.3.0b3/sdk/core/azure-mgmt-core/azure/mgmt/core/policies/_authentication.py#L65)).

PR #18141, opened 2021-04-19
Add environment.yml to unify conda versioning

scbedd:


eng/environment.yml
```diff
@@ -0,0 +1,6 @@
+channels:
```
weshaggard, line 1: Any reason to not add this into the existing globals.yml file?

weshaggard, line 1: What is a channels? Is that an variable?

scbedd: Negative. Channels is **supposed** to be a way that you can define a "pull from this location before going to the rest"

The problem I'm running into is that `conda` apparently ignores the setting. I could completely remove it and it would make no difference.

Example.

```
conda env create --name "test123" --file <pathto>/environment.yml
source activate test123
conda install msrest>=2021.05.01
```

Will FAIL because it doesn't use the channels in the environment config for some reason. You have to manually either run 

```
conda config --add channels https://azuresdkconda.blob.core.windows.net/channel1/
```

Before it will actually use it. This is really really lame. I think adding it to globals is a good suggestion. Will do so.

scbedd: Nope! Going to use it. Aside from this, the `environment.yml` is a conda specific thing.

eng/pipelines/templates/steps/build-conda-artifacts.yml
```diff
@@ -53,21 +53,28 @@ steps:
...
+          -c "$(Build.SourcesDirectory)/sdk/${{ parameters.ServiceDirectory }}/ci.yml"
 
     - bash: |
         echo "##vso[task.prependpath]$CONDA/bin"
       displayName: 'Prepend PATH with Conda and INIT'
 
     - bash: |
-        conda create --yes --quiet --name ${{ artifact.name }}
-        source activate ${{ artifact.name }}
+        conda env create --name ${{ artifact.name }} --file $(Build.SourcesDirectory)/eng/environment.yml
```
weshaggard, line 71: Do you expect to add other variables into environment.yml? If not should we just pass through the version variable instead of needing to parse the yml file itself? Or is the yml file for other scenarios as well?

scbedd: Eventually, yes I do expect to add other variables/configurations. Aside from that, this method is the method recommended by `conda` documentation. Normally I'd change in response to your feedback, but I'd really rather keep this guy.

weshaggard: Sounds good I just wanted to make sure we weren't adding the extra complications without some reasons. The only other question I'd have is should we name the file conda_env.yml or something that indicates it should be specific to conda.

weshaggard (approved):
One suggestion both otherwise looks reasonable.

eng/pipelines/templates/steps/build-conda-artifacts.yml
scbedd: Honestly. I was kinda leaning that way :) 

I'll do that before merging this.

doc/dev/conda-builds.md
```diff
@@ -16,6 +16,20 @@ A Conda Artifact defines:
 
 ## How to Build an Azure SDK Conda Package Locally
 
+### Set up your conda environment
+
+You will notice that all the azure-sdk conda distributions have the **same** version number and requirement set. This is due to the fact that the azure-sdk team pushes our conda packages out in waves. To support this, all versions are set via a common environment variable `AZURESDK_CONDA_VERSION`.
+
+We keep this environment variable set properly across all our builds by using a common `environment.yml` when creating our build environment. This environment definition ensures that
```
weshaggard, line 23: Looks like a couple renames missed here.

scbedd: ctrl-f-ed to find the rest. Should have em all now. Thanks.

PR #18146, opened 2021-04-19
[Communication]: Added phone number exception tests

lsundaralingam:


sdk/communication/azure-communication-phonenumbers/test/test_phone_number_administration_client_async.py
```diff
@@ -181,3 +181,53 @@ async def test_purchase_phone_numbers(self):
...
+                    capabilities,
+                    polling = True
+                )
+    
+    @AsyncCommunicationTestCase.await_prepared_test
+    async def test_update_phone_number_capabilities_with_invalid_phone_number(self):
+        if self.is_playback():
+            phone_number = "sanitized"
+        else:
+            phone_number = "+14255550123"
```
jbeauregardb, line 221: I remember @beltr0n saying something regarding using fake numbers and using a specific area code or something like that. If this already has this, disregard this comment 😄

jbeauregardb (approved):
Approving. Just waiting on a response from @beltr0n to the comment I made. May probably be nothing though

sdk/communication/azure-communication-phonenumbers/test/test_phone_number_administration_client_async.py
lsundaralingam: I got this number from @beltr0n. Its a valid phone number that's not owned by the ACS resource and is the number used in the SMS tests. 

https://github.com/Azure/azure-sdk-for-js/blob/master/sdk/communication/communication-sms/test/public/suites/smsClient.send.ts#L88

beltr0n: it's "valid" in the sense the format is correct, this number follows the guidelines of being a fake number

beltr0n (approved):

Issue #18129, opened 2021-04-19
AttributeError: 'dict' object has no attribute 'iter' in 


cbare:
- **Package Name**: azure-storage-blob
- **Package Version**: 12.8.0
- **Operating System**: Windows 10 21H1
- **Python Version**: 3.8.5

I got the following error when trying to work with a blob container using the Azure SDK for Python:

```
  File "[...]/python3.8/site-packages/azure/core/paging.py", line 129, in __next__
    return next(self._page_iterator)
  File "[...]/python3.8/site-packages/azure/core/paging.py", line 76, in __next__
    self._response = self._get_next(self.continuation_token)
  File "[...]/python3.8/site-packages/azure/storage/blob/_list_blobs_helper.py", line
79, in _get_next_cb
    process_storage_error(error)
  File "[...]/python3.8/site-packages/azure/storage/blob/_shared/response_handlers.py", line 98, in process_storage_error
    for info in error_body.iter():
AttributeError: 'dict' object has no attribute 'iter'
```

The underlying cause was a mistaken account_url "dfs.core.windows.net" where it should have been "blob.core.windows.net", something like this:

``` Python
        service = BlobServiceClient(
            account_url="https://mydatalake.dfs.core.windows.net/",
            credential=credential
        )
```

The expected behavior would be to report the error message. But instead, an error happens in line 97 of `_shared/response_handlers.py` in the function `process_storage_error`. The relevant lines are:

``` python
[95]       error_body = ContentDecodePolicy.deserialize_from_http_generics(storage_error.response)
[96]        if error_body:
[97]            for info in error_body.iter():
```

Stepping in with pdb, I see the following:

```
[...]/venv/lib/python3.8/site-packages/azure/storage/blob/_shared/response_handlers.py(97)process_storage_error()
-> if error_body:
(Pdb) type(error_body)
<class 'dict'>
(Pdb) print(error_body)
{'error': {'code': 'InvalidQueryParameterValue', 'message': 'Value for one of the query parameters specified in the request URI is invalid.\nRequestId:82bff81c-501f-0053-39b5-344d92000000\nTime:2021-04-19T00:44:42.6291960Z'}}
```

So, it looks like under at least some circumstances, `ContentDecodePolicy.deserialize_from_http_generics` returns a dict rather than whatever type is expected. A look at the source of `deserialize_from_http_generics` seems to support that.

Seems to be related to this bug filed on the cli: https://github.com/Azure/azure-cli/issues/16217

[...]

Issue #15956, opened 2020-12-31
AttributeError: 'dict' object has no attribute 'iter'


stevenylai:
- **Package Name**: azure-storage-file-datalake
- **Package Version**: 12.2.0
- **Operating System**: Linux
- **Python Version**: 3.5.2 (from HDInsight cluster with ESP)

**Describe the bug**
When listing an Azure directory which doesn't have proper permission, we get AttributeError. Traceback:
```
Traceback (most recent call last):
  File "azure.whl/azure/storage/filedatalake/_list_paths_helper.py", line 56, in _get_next_cb
  File "azure.whl/azure/storage/filedatalake/_generated/operations/_file_system_operations.py", line 442, in list_paths
azure.storage.filedatalake._generated.models._models_py3.StorageErrorException: (AuthorizationPermissionMismatch) This request is not authorized to perform this operation using this permission.
RequestId:e6038639-e01f-007e-5013-df0d71000000
Time:2020-12-31T01:21:21.9445915Z

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "test_client.py", line 18, in <module>
    for run_result in fs_client.get_paths(path='rtgasi/negotiation/hdfs_rfe_v1/results/205ba9c9-8cdb-4f77-887f-c561e2c8edc3', recursive=False):
  File "azure.whl/azure/core/paging.py", line 129, in __next__
  File "azure.whl/azure/core/paging.py", line 76, in __next__
  File "azure.whl/azure/storage/filedatalake/_list_paths_helper.py", line 58, in _get_next_cb
  File "azure.whl/azure/storage/filedatalake/_shared/response_handlers.py", line 94, in process_storage_error
AttributeError: 'dict' object has no attribute 'iter'
```

**To Reproduce**
1. Just run the following code where `some/path` does not have the proper permission:
```python
credential = identity.ClientSecretCredential('xx', 'xx', 'xx')
service_client = filedatalake.DataLakeServiceClient('https://xx.dfs.core.windows.net', credential=credential)
fs_client = service_client.get_file_system_client(file_system='xx')
for result in fs_client.get_paths(path='some/path', recursive=False):
  pass
```

**Expected behavior**
Should raise `StorageErrorException` or other more relevant errors

From the look of the exception, we just need to remove `.iter()` part and it should work for both Python 2 and 3:
[...]

PR #18147, opened 2021-04-19
Raise appropriate exception when error_body is dict | Fix exception raised during handling of another exception

tasherif-msft:
This resolves https://github.com/Azure/azure-sdk-for-python/issues/18129, https://github.com/Azure/azure-cli/issues/16217 and https://github.com/Azure/azure-sdk-for-python/issues/15956

This PR will improve error handling in the following ways:
- The error traceback will be more thorough showing the error code, message, additional info consistently
- No more `error occurred during handling of the above exception`
- No more bug related to JSON responses
- All errors thrown should be consistent in their properties


cbare:
Hi Tamer,

Thanks for having a look at this bug. I hope you don't mind, but I went down a bit of rabbit-hole on this one.

What I imagine is happening is that Azure's endpoints return either XML or JSON depending on the Accept header in the request, and that applies to error responses as well as 200s.

So, the `error_body` we get in this method could be the result of deserializing an XML response into xml.etree.ElementTree objects or a JSON response into a dict.  Here's some toy code to play around with.

```
import xml.etree.ElementTree as ET
import json

xml_txt = """
<?xml version="1.0"?>
<error>
  <code>1234</code>
  <message>Flaming disaster!!</message>
  <asdf>quux</asdf>
</error>
""".strip()

err1 = ET.fromstring(xml_txt)

json_txt = """
{
    "error": {
        "code": "InvalidQueryParameterValue",
        "message": "Value for one of the query parameters specified in the request URI is invalid...",
        "xyz": "blah"
    }
}
"""
err2 = json.loads(json_txt)
```

We could tweak the existing code to handle either scenario, plus have a half-way decent fall-back if some other type, like plain text, comes back.

```
# try out the XML case
error_body = err1

# try out the dict case
# error_body = err2

additional_data = {}
if isinstance(error_body, ET.Element):
    for child in error_body:
        if child.tag.lower() == 'code':
            error_code = child.text
        elif child.tag.lower() == 'message':
            error_message = child.text
        else:
            additional_data[child.tag] = child.text
elif isinstance(error_body, dict):
    for key, value in error_body['error'].items():
        if key.lower() == 'code':
            error_code = value

        elif key.lower() == 'message':
            error_message = value
        else:
            additional_data[key] = value
else:
   #  _LOGGER.warning(f'Unexpected return type {type(error_body)} from ContentDecodePolicy.deserialize_from_http_generics.')
    error_message = str(error_body)

print(f'code: {error_code}')
print(f'message: {error_message}')
print(f'additional data: {additional_data}')
```

The code above is repetitive and still has some problems, like what if our message doesn't have "code" and "message" fields? Then `error_code` and `error_message` are undefined.

One idea would be to coerce the one case into the other:

```
if isinstance(error_body, ET.Element):
    error_dict = {
        child.tag.lower():child.text
        for child in error_body
    }
elif isinstance(error_body, dict):
    error_dict = error_body.get('error', {})
else:
    # _LOGGER.warning(f'Unexpected return type {type(error_body)} from ContentDecodePolicy.deserialize_from_http_generics.')
    error_dict = {'message': str(error_body)}

error_code = error_dict.get('code')
error_message = error_dict.get('message')
additional_data = {k:v for k,v in error_dict.items() if k not in {'code', 'message'}}

print(f'code: {error_code}')
print(f'message: {error_message}')
print(f'additional data: {additional_data}')
```

I hope this helps,
Cheers!

tasherif-msft:
Hi @cbare I've spent some time to dig deeper into our error handling.
As you mentioned in most cases we are deserializing the xml responses. In some cases we will not have a response body that contains error codes/messages and instead we will just fetch them from the response headers. Which means this check will be avoided (since the error code and message would just be set [here](https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/storage/azure-storage-blob/azure/storage/blob/_shared/response_handlers.py#L91-L92)). 

This is also why I don't think there is a case where we would encounter an undefined error code/message since the three cases (error in headers and not body, error in xml body, error in JSON body) should cover everything.

I do like your second recommendation, it's very concise. I will make some few tweaks to account for the case I mentioned earlier (error code and message in headers + empty body)

Thank you very much @cbare :)

xiafu-msft (approved):

Issue #12376, opened 2020-07-06
[keyvault] improve multi api test coverage


iscai-msft:
- [ ] add calls for each instantiation of a client with a different API version and ensure the correct API version is passed through
- [ ] test getting new features on older API versions to ensure that it's handled correctly
- [ ] potentially run all tests against every API version like .NET (link [here](https://teams.microsoft.com/l/message/19:7b87fb348f224b37b6206fa9d89a105b@thread.skype/1594056349625?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=3e17dcb0-4257-4a30-b843-77f47f1d4121&parentMessageId=1594056349625&teamName=Azure%20SDK&channelName=Language%20-%20DotNet&createdTime=1594056349625))

PR #18149, opened 2021-04-20
[Key Vault] Target multiple API versions with tests (keys)

mccoyp:
Part of #12376. Same methodology as #16979, with the exception that keys tests are also parameterized for resource type (vault vs. HSM). Tests against Managed HSM only use the default API version.


sdk/keyvault/azure-keyvault-keys/tests/_test_case.py
```diff
@@ -5,15 +5,33 @@
...
+    api_version = param.kwargs.get("api_version")
     suffix = "mhsm" if param.kwargs.get("is_hsm") else "vault"
-    return "{}_{}".format(testcase_func.__name__, parameterized.to_safe_name(suffix))
+    return "{}_{}_{}".format(
+        testcase_func.__name__, parameterized.to_safe_name(api_version), parameterized.to_safe_name(suffix)
+    )
+
+
+# parameters for test case parameterization, where [x, y] = [api_version, is_hsm]
+PARAMETER_COMBINATIONS = _get_test_parameters()
```
chlowell, line 34: Looks like you use this only to generate many copies of the same list. Could you do something like this instead?
```py
PARAMS = [param(api_version=p[0], is_hsm=p[1])) for p in _get_test_parameters()]

# then import PARAMS in test modules
@parameterized.expand(PARAMS, name_func=suffixed_test_name)
def test_...
```
Or like this, better yet and no need for `PARAMS`:
```py
test_all_versions = functools.partial(parameterized.expand, PARAMS, name_func=suffixed_test_name)

@test_all_versions()
def test_...
```

sdk/keyvault/azure-keyvault-keys/tests/test_key_client.py
```diff
@@ -134,13 +135,15 @@ def _to_bytes(hex):
...
+        [(param(api_version=p[0], is_hsm=p[1])) for p in PARAMETER_COMBINATIONS], name_func=suffixed_test_name
+    )
     def test_key_crud_operations(self, **kwargs):
         is_hsm = kwargs.pop("is_hsm")
-        self._skip_if_not_configured(is_hsm)
+        self._skip_if_not_configured(kwargs.get("api_version"), is_hsm)
         endpoint_url = self.managed_hsm_url if is_hsm else self.vault_url
 
-        client = self.create_key_client(endpoint_url)
+        client = self.create_key_client(endpoint_url, **kwargs)
```
chlowell, line 146: When I look at this I still see boilerplate crying out for a decorator. Maybe I just don't see the blocker, does something prevent consolidating client creation into a decorator?

mccoyp: `parameterized.expand` sends the `api_version` and `is_hsm` kwargs directly to the test method, so I don't think a decorator could take these into account when creating a client

sdk/keyvault/azure-keyvault-keys/tests/_test_case.py
mccoyp: Great ideas! I'll try them out and push an update

sdk/keyvault/azure-keyvault-keys/tests/test_key_client.py
chlowell: You mean, it sends those arguments directly to the function it decorates 😉 

```py
import functools
from parameterized import parameterized, param

def boilerplate(fn):
    @functools.wraps(fn)
    def wrapper(test_class_instance, api_version, is_hsm, **kwargs):
        test_class_instance._skip_if_not_configured(api_version, is_hsm)
        endpoint_url = test_class_instance.managed_hsm_url if is_hsm else test_class_instance.vault_url
        client = test_class_instance.create_key_client(endpoint_url, **kwargs)
        fn(client)
    return wrapper

class Tests(KeysTestCase, KeyVaultTestCase):
    @parameterized.expand([param(api_version=42, is_hsm=b) for b in (True, False)])
    @boilerplate
    def test_foo(self, client):
        ...
```

mccoyp: 😳

sdk/keyvault/azure-keyvault-keys/tests/_test_case.py
```diff
@@ -2,18 +2,60 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
 # ------------------------------------
+import asyncio
```
chlowell, line 5: This will raise an ImportError on 2.7. If you want to combine a/sync in one module, you need to do [something like this](https://github.com/Azure/azure-sdk-for-python/blob/5ec46673404af7aa30fe6e4161b7687d1521bdd2/tools/azure-devtools/src/azure_devtools/scenario_tests/preparers.py#L146-L155).

sdk/keyvault/azure-keyvault-keys/tests/_test_case.py
```diff
@@ -2,18 +2,61 @@
...
+        endpoint_url = test_class_instance.managed_hsm_url if is_hsm else test_class_instance.vault_url
+        client = test_class_instance.create_key_client(endpoint_url, api_version=api_version, **kwargs)
+
+        if kwargs.get("is_async"):
+            import asyncio
+
+            coroutine = asyncio.coroutine(testcase_func)
+            future = coroutine(test_class_instance, client, is_hsm=is_hsm)
+            loop = asyncio.get_event_loop()
+            loop.run_until_complete(future)
```
chlowell, line 30: If `testcase_func` is an `async` function, it returns a `coroutine`:
```py
>>> async def foo(): pass;
>>> foo()
<coroutine object foo at 0x7f4ce239ac50>
```
So you don't need to create one:
```suggestion
            coroutine = testcase_func(test_class_instance, client, is_hsm=is_hsm)
            loop = asyncio.get_event_loop()
            loop.run_until_complete(coroutine)
```

sdk/keyvault/azure-keyvault-keys/tests/test_crypto_client_async.py
```diff
@@ -134,15 +137,10 @@ async def _import_symmetric_test_key(self, client, name):
         assert key_vault_key.key.kid == imported_key.id == key_vault_key.id
         return key_vault_key
 
-    @parameterized.expand([param(is_hsm=b) for b in [True, False]], name_func=suffixed_test_name)
-    @PowerShellPreparer("keyvault")
-    async def test_ec_key_id(self, **kwargs):
+    @all_api_versions()
+    @client_setup
+    async def test_ec_key_id(self, key_client, is_hsm, **kwargs):
```
chlowell, line 142: `client_setup` passes `is_hsm` by keyword and the test doesn't use it, may as well leave it in kwargs:
```suggestion
    async def test_ec_key_id(self, key_client, **kwargs):
```

chlowell (commented):
So much tidier 🎁

sdk/keyvault/azure-keyvault-keys/tests/test_crypto_client_async.py
mccoyp: `is_hsm` is used here as the value of `hardware_protected` when creating a key. I did make sure to leave out the positional parameter in tests that don't use it to keep these nice and tidy 🤓

chlowell (approved):

Issue #13056, opened 2020-08-12
[ServiceBus] Cut off livetest time


yunhaoling:
- [x] mgmt livetest use recording, but have some nightly scheduled test running using real resource
- [x] decrease lock duration time in some test case, so that we could sleep less, as "lock_duration" is configurable now

pre cost: ~2 hours 45-50 minutes (according to nightly run)
cur cost:

PR #18176, opened 2021-04-20
[Service Bus] reduce livetest time

swathipil:
fixes: #13056


yunhaoling:
@swathipil , this is great! let's kick off the livetest pipeline to see how much time you have saved 😉

sdk/servicebus/azure-servicebus/tests/async_tests/mgmt_tests/test_mgmt_queues_async.py
```diff
@@ -579,7 +579,6 @@ async def test_mgmt_queue_async_update_dict_success(self, servicebus_namespace_c
         finally:
             await mgmt_service.delete_queue(queue_name)
     
-    @pytest.mark.liveTest
```
yunhaoling, line 582: out of curiosity, do you know how annotation work? and by removing this, the livetest pipeline triggered by PR will not run mgmt tests in live mode while nightly running still runs in live mode?

sdk/servicebus/azure-servicebus/tests/async_tests/test_queues_async.py
```diff
@@ -231,7 +231,7 @@ async def test_async_queue_by_queue_client_conn_str_receive_handler_receiveandde
 
             assert not receiver._running
             assert len(messages) == 10
-            time.sleep(30)
+            time.sleep(10)
```
rakshith91, line 234: can we ensure this will not increase the number of transient failures? same question for all similar changes

swathipil: 🤔 hmm, that's a good point. I'm not sure that this won't increase transient failures. Do you know what the best way to check that would be?

sdk/servicebus/azure-servicebus/tests/async_tests/mgmt_tests/test_mgmt_queues_async.py
swathipil: 🤔 hmm, I'm not quite sure. I'll do some more investigation and get back to you!

sdk/servicebus/azure-servicebus/tests/async_tests/test_queues_async.py
yunhaoling: @rakshith91 thanks for being thoughtful!

as far as I could remember, the transient failures are usually caused by tests including receiving which doesn't wait long enough.

I think we could merge the PR first and see how it goes, if we keep observing failures, then we fix them :P

yunhaoling (approved):

sdk/servicebus/azure-servicebus/tests/async_tests/test_queues_async.py
swathipil: sounds good to me!

PR #18181, opened 2021-04-20
[Tables] on a 413 throw a RequestEntity Too Large Error

seankane-msft:
#18172


sdk/tables/azure-data-tables/azure/data/tables/__init__.py
```diff
@@ -6,6 +6,7 @@
 from azure.data.tables._models import TableServiceStats
 
 from ._entity import TableEntity, EntityProperty, EdmType
+from ._error import RequestEntityTooLargeError
```
annatisch, line 9: I would prefer `RequestTooLargeError`
`Entity` already means something different in the Tables context :)

annatisch (approved):
LGTM - Changelog update? ;)

PR #18183, opened 2021-04-20
[Key Vault] Target multiple API versions with tests (secrets)

mccoyp:
Part of #12376 -- same methodology as #16979, with added decorator work from #18149.


sdk/keyvault/azure-keyvault-secrets/tests/_test_case.py
```diff
@@ -0,0 +1,63 @@
...
+
+def client_setup(testcase_func):
+    """decorator that creates a client to be passed in to a test method"""
+    @PowerShellPreparer("keyvault", azure_keyvault_url="https://vaultname.vault.azure.net")
+    @functools.wraps(testcase_func)
+    def wrapper(test_class_instance, azure_keyvault_url, api_version, **kwargs):
+        test_class_instance._skip_if_not_configured(api_version)
+        client = test_class_instance.create_client(azure_keyvault_url, api_version=api_version, **kwargs)
+
+        if kwargs.get("is_async"):
```
chlowell, line 23: Do you need `is_async` for anything else? If not, consider this: https://github.com/Azure/azure-sdk-for-python/blob/2794bfcc88aad415920c234a848701f2bcfeeb88/tools/azure-devtools/src/azure_devtools/scenario_tests/preparers.py#L145-155

chlowell (approved):

sdk/keyvault/azure-keyvault-secrets/tests/_test_case.py
mccoyp: `is_async` is also provided to create_client, but I'll keep this in mind!

PR #18186, opened 2021-04-20
Clean Up PR Triggers

scbedd:
To reduce random build spam from tools/ and eng/ changes.

### DIFF
--- a/sdk/appconfiguration/ci.yml
+++ b/sdk/appconfiguration/ci.yml
@@ -28,8 +28,6 @@ pr:
     include:
     - sdk/appconfiguration/
     - sdk/core/
-    - tools/
-    - eng/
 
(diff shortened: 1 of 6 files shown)


scbedd:
@seankane-msft reducing build spam on anything other than `core` changes. Leaving `core` triggering on eng/ and tools/ changes.

seankane-msft (approved):
LGTM, one question. Will we be able to test pipeline fixes on libraries still? Or does this just reduce the number of libraries that will trigger on eng/tools changes?

scbedd:
Both, unfortunately. This is to reduce build spam from our eng/common changes. It's hitting the build pools pretty hard. We will still trigger core build on tools/eng changes though.

PR #18197, opened 2021-04-21
[AutoRelease] t2-signalr-2021-04-21-60081(Do not merge)

azclibot:
Please Add the link issue


RAY-316:
new package will contains this change

PR #18199, opened 2021-04-21
Azure webpubsub service initial client library version

johanste:


sdk/signalr/azure-messaging-webpubsubservice/azure/messaging/webpubsubservice/__init__.py
```diff
@@ -0,0 +1,188 @@
...
+from ._policies import JwtCredentialPolicy
+from ._utils import UTC as _UTC
+
+if TYPE_CHECKING:
+    from azure.core.pipeline.policies import HTTPPolicy, SansIOHTTPPolicy
+    from typing import Any, List, cast, Type, TypeVar
+    ClientType = TypeVar('ClientType', bound='WebPubSubServiceClient')
+
+
+def build_authentication_token(endpoint, hub, key, **kwargs):
```
terencefan, line 37: We also need a `from_connection_string` version of this function.
Do you have any comments on this?

sdk/signalr/azure-messaging-webpubsubservice/setup.py
```diff
@@ -0,0 +1,76 @@
...
+            "azure.messaging",
+            "tests",
+            "examples",
+        ]
+    ),
+    install_requires=[
+        "azure-core<2.0.0,>=1.10.0",
+        "msrest>=0.6.18",
+        "cryptography>=2.1.4",
+        "pyjwt>=1.7.1",
```
terencefan, line 69: We need to upgrade the version of `pyjwt` to 2.0.x

zackliu: Related issue https://github.com/jpadilla/pyjwt/issues/505

sdk/signalr/azure-messaging-webpubsubservice/README.md
```diff
@@ -0,0 +1,99 @@
...
+>>> with open('file.json', 'r') as f:
+>>>    request = build_send_to_all_request('ahub', content=f, content_type='application/json')
+>>>    response = client.send_request(request)
+>>> print(response)
+<RequestsTransportResponse: 202 Accepted>
+```
+
+## Key concepts
+
+### Hubs, groups, connections and users
```
juniwang, line 54: missing contents?

sdk/signalr/azure-messaging-webpubsubservice/setup.py
johanste: Upgrading to pyjwt 2.0.x drops python 2.7 support. Looking at the issue referenced, I'm not sure why we need it?

zackliu: The current version return `byte` instead of `string` in `jwt.encode()` in python3, which makes the generated token incorrect. If you don't want to upgrade jwt version, you need to fix it here with a `decode('UTF-8')`.
https://github.com/Azure/azure-sdk-for-python/pull/18199/files#diff-418e1fbe16ab098ec7fe81f33d656dbca5ce789bd35f01dcca74754bae5b1215R89

sdk/signalr/azure-messaging-webpubsubservice/azure/messaging/webpubsubservice/_utils.py
```diff
@@ -0,0 +1,44 @@
...
+        return self.__class__.ZERO
+
+    def tzname(self, dt):
+        return "UTC"
+
+    def dst(self, dt):
+        return self.__class__.ZERO
+
+
+UTC = _UTC_TZ()
```
annatisch, line 44: We should probably support the stdlib type where available:
```python
try:
    from datetime import timezone
    UTC = timezone.utc
except ImportError:
    UTC = _UTC_TZ()
```

sdk/signalr/azure-messaging-webpubsubservice/azure/messaging/webpubsubservice/__init__.py
```diff
@@ -0,0 +1,188 @@
...
+
+        kwargs['credential'] = corecredentials.AzureKeyCredential(kwargs.pop('accesskey'))
+        return cls(**kwargs)
+
+    def __repr__(self):
+        return "<WebPubSubServiceClient> endpoint:'{}'".format(self.endpoint)
+
+    def _format_url(self, url):
+        # type: (str) -> str
+        assert self.endpoint[-1] != "/", "My endpoint should not have a trailing slash"
```
annatisch, line 165: Would be kind of strange to encounter an `AssertError` when constructing the client?

johanste: In that scenario, this is truly an assert (we are stripping the trailing slash before calling the method). However, when looking at the code, the endpoint is a read/write attribute (which it probably shouldn't be) and the user could have set it to a value that does not conform to our assumptions. I may want to change self.endpoint to a read-only property...

sdk/signalr/azure-messaging-webpubsubservice/azure/messaging/webpubsubservice/core/rest.py
```diff
@@ -0,0 +1,275 @@
...
+            self._internal_request.set_json_body(json)
+            if not self._internal_request.headers.get("Content-Type"):
+                self._internal_request.headers["Content-Type"] = "application/json"
+        if files is not None:
+            self._internal_request.set_formdata_body(files)
+
+        if not self._internal_request.headers.get("Content-Length"):
+            try:
+                # set content length header if possible
+                self._internal_request.headers["Content-Length"] = str(len(self._internal_request.data))  # type: ignore
```
annatisch, line 115: Is this going to be reliable?

sdk/signalr/azure-messaging-webpubsubservice/azure/messaging/webpubsubservice/core/rest.py
```diff
@@ -0,0 +1,275 @@
...
+    def encoding(self, value):
+        # type: (str) -> None
+        """Sets the response encoding"""
+        self._encoding = value
+
+    @property
+    def text(self):
+        # type: (...) -> str
+        """Returns the response body as a string"""
+        return self._internal_response.text(self.encoding)
```
annatisch, line 239: The behaves correctly when passing an empty string for encoding?

sdk/signalr/azure-messaging-webpubsubservice/azure/messaging/webpubsubservice/core/rest.py
```diff
@@ -0,0 +1,275 @@
...
+    def content(self):
+        # type: (...) -> bytes
+        return self._internal_response.body()
+
+
+class AsyncHttpResponse(_HttpResponseBase):
+    @property
+    def content(self):
+        # type: (...) -> bytes
+        return self._internal_response.body()
```
annatisch, line 275: Isn't this a coroutine?

sdk/signalr/azure-messaging-webpubsubservice/azure/messaging/webpubsubservice/rest.py
```diff
@@ -0,0 +1,969 @@
...
+from typing import TYPE_CHECKING
+from msrest import Serializer
+from azure.core.pipeline.transport._base import _format_url_section
+from azure.messaging.webpubsubservice.core.rest import HttpRequest
+
+if TYPE_CHECKING:
+    # pylint: disable=unused-import,ungrouped-imports
+    from typing import Any, IO, List, Optional, Union, Dict
+    from typing_extensions import Literal
+    Permissions = Union[Literal['joinLeaveGroup'], Literal['sendToGroup']]
```
annatisch, line 38: Not for the generator - I believe `Literal` supports multiple options: `Literal['joinLeaveGroup', 'sendToGroup']`

sdk/signalr/azure-messaging-webpubsubservice/README.md
johanste: Updated. Thanks.

sdk/signalr/azure-messaging-webpubsubservice/azure/messaging/webpubsubservice/rest.py
johanste: You are correct. Today I learned something new :).

sdk/signalr/azure-messaging-webpubsubservice/examples/send_messages.py
```diff
@@ -0,0 +1,67 @@
...
+
+endpoint = os.environ['ENDPOINT']
+api_key = os.environ['API_KEY']
+
+LOG = logging.basicConfig(level=logging.DEBUG)
+
+client = WebPubSubServiceClient(endpoint, credential=AzureKeyCredential(api_key), tracing_enabled=True)
+
+# Send message to everybody on the given hub...
+request = build_send_to_all_request('ahub', json={ 'Hello': 'all!' })
```
annatisch, line 17: If I pass in a string to the `json` parameter - will that just become the payload?

sdk/signalr/azure-messaging-webpubsubservice/azure/messaging/webpubsubservice/core/rest.py
annatisch: Oh true!

sdk/signalr/azure-messaging-webpubsubservice/examples/send_messages.py
```diff
@@ -0,0 +1,67 @@
...
+    print('Failed to send message: {}'.format(response))
+
+
+request = build_send_to_all_request('ahub', content='hello, text!', content_type='text/plain')
+print(request.headers.items())
+response = client.send_request(request)
+print(response)
+
+
+request = build_send_to_all_request('ahub', content=io.BytesIO(b'{ "hello": "world" }'), content_type='application/json')
```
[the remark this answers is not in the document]
johanste: `content_type` is special. Both in our handling as well as in openapi. But, yeah, other headers specified in swagger will become keyword-only arguments.

sdk/signalr/azure-messaging-webpubsubservice/azure/messaging/webpubsubservice/__init__.py
```diff
@@ -165,24 +174,45 @@ def _format_url(self, url):
...
 
-        :param request: The network request you want to make. Required.
-        :type request: ~corerest.HttpRequest
-        :keyword bool stream: Whether the response payload will be streamed. Defaults to False
-        :return: The response of your network call.
-        :rtype: ~corerest.HttpResponse
+        We have helper methods to create requests specific to this service in `azure.messaging.webpubsub.rest`.
+        Use these helper methods to create the request you pass to this method. See our example below:
+
+        >>> from azure.messaging.webpubsub.rest import build_healthapi_get_health_status_request
```
annatisch, line 184: I think you may want to wrap this in something like:
```
.. code-block:: python
```

sdk/signalr/azure-messaging-webpubsubservice/azure/messaging/webpubsubservice/aio.py
```diff
@@ -0,0 +1,108 @@
...
+
+if TYPE_CHECKING:
+    import azure.core.credentials as corecredentials
+    from azure.core.pipeline.policies import HTTPPolicy, SansIOHTTPPolicy
+    from typing import Any, List, cast
+
+
+class WebPubSubServiceClient(object):
+    def __init__(self, endpoint, credential, **kwargs):
+        # type: (str, corecredentials.AzureKeyCredential, Any) -> None
```
annatisch, line 32: No Python 3 typehint syntax? ;)

sdk/signalr/azure-messaging-webpubsubservice/azure/messaging/webpubsubservice/__init__.py
johanste: Good suggestion. @iscai-msft , we should make sure that the comments we generate render well in our docs. @annatisch , the format here is actually doctest compatible. Not sure if wrapping `.. code-block:: python` retains that.

sdk/signalr/azure-messaging-webpubsubservice/azure/messaging/webpubsubservice/aio.py
```diff
@@ -0,0 +1,108 @@
...
+            **kwargs
+        )
+        policies = [
+            corepolicies.HeadersPolicy(**kwargs),
+            corepolicies.UserAgentPolicy(**kwargs),
+            corepolicies.AsyncRetryPolicy(**kwargs),
+            corepolicies.ProxyPolicy(**kwargs),
+            corepolicies.CustomHookPolicy(**kwargs),
+            corepolicies.AsyncRedirectPolicy(**kwargs),
+            JwtCredentialPolicy(credential, kwargs.get("user", None)),
```
annatisch, line 55: You added `corepolicies.ContentDecodePolicy(**kwargs),` to the sync pipeline so I guess it should be here too?

sdk/signalr/azure-messaging-webpubsubservice/azure/messaging/webpubsubservice/aio.py
annatisch: Oh nevermind - you used them further down, so I guess that's a passing grade :P

sdk/signalr/azure-messaging-webpubsubservice/azure/messaging/webpubsubservice/aio.py
johanste: For this service, we actually don't need it. Will remove instead...

sdk/signalr/azure-messaging-webpubsubservice/examples/send_messages.py
```diff
@@ -62,6 +62,14 @@
     print('The group exists!')
 
 
+request = build_grant_permission_request('skalman', 'sendToGroup', '7')
+response = client.send_request(request)
+response.read()
+print(response.content)
```
annatisch, line 68: Does the `read()` method not also return the content?

sdk/signalr/azure-messaging-webpubsubservice/azure/messaging/webpubsubservice/core/rest/_rest_py3.py
```diff
@@ -0,0 +1,716 @@
...
+
+    def close(self):
+        self.response.close()
+
+class _AsyncStreamContextManager(_StreamContextManagerBase):
+    async def __aenter__(self) -> "AsyncHttpResponse":
+        """Actually make the call only when we enter. For async stream_response calls."""
+        if not isinstance(self.client, _AsyncPipelineClient):
+            raise TypeError(
+                "Only sync calls should enter here. If you mean to do a sync call, "
```
annatisch, line 263: I think this should be "Only _async_ calls..."?

sdk/signalr/azure-messaging-webpubsubservice/setup.py
johanste: six to the rescue. Fixed.

sdk/signalr/azure-messaging-webpubsubservice/azure/messaging/webpubsubservice/core/rest/_rest_py3.py
```diff
@@ -0,0 +1,716 @@
...
+            self._num_bytes_downloaded += len(raw_bytes)
+            yield raw_bytes
+
+        self._close_stream()
+
+
+class AsyncHttpResponse(_HttpResponseBase):
+
+    @property
+    def content(self) -> bytes:
```
annatisch, line 623: Seems like this property could live on `_HttpResponseBase`?

sdk/signalr/azure-messaging-webpubsubservice/azure/messaging/webpubsubservice/core/rest/_rest_py3.py
```diff
@@ -0,0 +1,716 @@
...
+
+        await self._close_stream()
+
+    async def close(self) -> None:
+        self.is_closed = True
+        self._internal_response.internal_response.close()
+        await asyncio.sleep(0)
+
+    async def __aexit__(self, *args) -> None:
+        await self._internal_response.internal_response.__aexit__(*args)
```
annatisch, line 689: Should we also be calling `self.close()` here?

sdk/signalr/azure-messaging-webpubsubservice/README.md
```diff
@@ -0,0 +1,142 @@
...
+
+```bash
+python -m pip install azure-messaging-webpubsubservice
+```
+
+#### Prequisites
+
+- Python 2.7, or 3.6 or later is required to use this package.
+- You need an [Azure subscription][azure_sub], and a [Azure WebPubSub service instance][webpubsubservice_docs] to use this package.
+- An existing Azure Web PubSub service instance.
```
annatisch, line 42: This seems like a duplicate of the line above.

sdk/signalr/azure-messaging-webpubsubservice/azure/messaging/webpubsubservice/__init__.py
annatisch: To me that would be weird for doctest to not test sections marked as code... but I could see nothing in the doctest docs that would indicate there might be issues.

sdk/signalr/azure-messaging-webpubsubservice/examples/send_messages.py
annatisch: After reading the rest of the code I see that it does - in which case I think it would look more natural in the samples to simply print the returned value?

iscai-msft: response.read reads the response into memory. but since this isn't a streamed response, it's not needed in the samples any more

sdk/signalr/azure-messaging-webpubsubservice/azure/messaging/webpubsubservice/__init__.py
iscai-msft: sounds good, making a note about this

sdk/signalr/azure-messaging-webpubsubservice/azure/messaging/webpubsubservice/core/rest/_rest_py3.py
iscai-msft: yep good point. it used to be different across sync and async, but no longer needed. Not going to update here though (it's fine to keep it like this), but will update in my core pr. thanks @annatisch !

sdk/signalr/azure-messaging-webpubsubservice/azure/messaging/webpubsubservice/core/rest/_rest_py3.py
iscai-msft: that's a good point, let me call close instead to handle stream closing. Not sure if we want to update this in this pr @johanste , but will udpate in my core pr

sdk/signalr/azure-messaging-webpubsubservice/azure/messaging/webpubsubservice/core/rest/_rest_py3.py
iscai-msft: Good point

sdk/signalr/azure-messaging-webpubsubservice/README.md
```diff
@@ -0,0 +1,148 @@
...
+This project has adopted the
+[Microsoft Open Source Code of Conduct][code_of_conduct]. For more information,
+see the Code of Conduct FAQ or contact opencode@microsoft.com with any
+additional questions or comments.
+
+<!-- LINKS -->
+[webpubsubservice_docs]: https://aka.ms/awps/doc
+[azure_cli]: https://docs.microsoft.com/cli/azure
+[azure_sub]: https://azure.microsoft.com/free/
+[webpubsubservice_client_class]: https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/signalr/azure-messaging-webpubsubservice/azure/messaging/webpubsubservice/__init__.py
```
vicancy, line 142: move the `azure-messaging-webpubsubservice` into a new `webpubsub` folder instead of using `signalr`?

iscai-msft:
@johanste did a quick look through the failing sdist error, it seems to be mad that your examples folder is not called `samples`, and that `samples` is not included in `MANIFEST.in`

sdk/signalr/azure-messaging-webpubsubservice/azure/messaging/webpubsubservice/rest.py
```diff
@@ -0,0 +1,942 @@
...
+    'build_grant_permission_request',
+    'build_healthapi_get_health_status_request',
+    'build_remove_connection_from_group_request',
+    'build_remove_user_from_all_groups_request',
+    'build_remove_user_from_group_request',
+    'build_revoke_permission_request',
+    'build_send_to_all_request',
+    'build_send_to_connection_request',
+    'build_send_to_group_request',
+    'build_send_to_user_request'
```
annatisch, line 28: I'm guessing there are no operations groups for this API?
I saw there's been a shift to include operation groups in the namespace rather than the function name - but it seems that change has no impact on this SDK?

annatisch (approved):

sdk/signalr/azure-messaging-webpubsubservice/azure/messaging/webpubsubservice/rest.py
iscai-msft: @annatisch @johanste has already done some manual-manipulation of the code, we completely got rid of the operation groups

PR #18214, opened 2021-04-22
[formrecognizer] add testcases for labeled tables - fixed/variable rows

kristapratico:


iscai-msft (approved):

catalinaperalta (approved):
LGTM! My only nit would be that `formrecognizer_label_table_variable_rows_storage_container_sas_url` is super long, but I dont see how to shorten it unless we use shorter versions of the words themselves. It's still fine if you merge it as is though.

kristapratico:
> LGTM! My only nit would be that formrecognizer_label_table_variable_rows_storage_container_sas_url is super long, but I dont see how to shorten it unless we use shorter versions of the words themselves. It's still fine if you merge it as is though.

I could drop the "label" and "storage" part? formrecognizer_table_variable_rows_container_sas_url

catalinaperalta:
> I could drop the "label" and "storage" part? formrecognizer_table_variable_rows_container_sas_url
Yeah that works for me! The name still seems pretty clear to me without those words. I like it 👍

maririos (approved):
Thank you!

PR #18229, opened 2021-04-22
show detailed error

xiangyan99:


xiangyan99:
#14909

sdk/core/azure-core/azure/core/exceptions.py
```diff
@@ -160,13 +160,14 @@ def error(self):
...
-        return "({}) {}".format(self.code, self.message)
+        return self.message_details()
 
     def message_details(self):
         """Return a detailled string of the error.
         """
         # () -> str
-        error_str = "Code: {}".format(self.code)
+        error_str = "({}) {}".format(self.code, self.message)
+        error_str += "\nCode: {}".format(self.code)
```
msyyc, line 170: To make sure the first line includes all info, it seems that 'error_str' is duplicated for 'self.code' and 'self.message'?

xiangyan99: We duplicate the information because 

error_str = "({}) {}".format(self.code, self.message)

is the old error message and we want to try our best to be backward compatible. (If customer checks the first line of new message, they can get same information as old one).

sdk/core/azure-core/azure/core/exceptions.py
```diff
@@ -107,6 +107,27 @@ class ODataV4Format(object):
...
+                },
+                {
+                    "code": "ValidationError",
+                    "target": "representation",
+                    "message": "Parsing error(s): The input OpenAPI file is not valid for the OpenAPI specificate
+                    https: //github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md
+                    (schema https://github.com/OAI/OpenAPI-Specification/blob/master/schemas/v2.0/schema.json)."
+                }
+            ]
+        }
```
annatisch, line 129: Would be worth checking how this renders in sphinx (if it's actually in the public docs).
It may need additional `code-block` tags for the formatting to work.

sdk/core/azure-core/azure/core/exceptions.py
```diff
@@ -160,13 +181,14 @@ def error(self):
         return self
 
     def __str__(self):
-        return "({}) {}".format(self.code, self.message)
+        return self.message_details()
```
annatisch, line 184: Is `str()` also supposed to have a character limit? Or is that just `repr()`?

sdk/core/azure-core/azure/core/exceptions.py
annatisch: hmm but would this be breaking for anyone who was parsing `error.message_details()`? That was already public right?

xiangyan99:
Technically, yes. But given the issue is to update exception error message. I think it is acceptable?

annatisch:
You could refactor it like this to make it otherwise unchanged:
```python
def _format_message_details(self):
    """
    Original message_details method unchanged - just made private.
    """
    error_str = "Code: {}".format(self.code)
    error_str += "\nMessage: {}".format(self.message)
    ...

def __str__(self):
        return "({}) {}\n{}".format(
            self.code,
            self.message
            self._format_message_details()
        )

def message_details(self):
    """
    Preserve the original formatted message details.
    """
    return self._format_message_details()

```

xiangyan99:
I thought about it :). But I felt a little weird if we show slightly different messages between str() & message_details(). I was concerned It would cause confusions.

I will update my PR.

sdk/core/azure-core/azure/core/exceptions.py
xiangyan99: From my experiment, .. code-block:: tag will make the json invisible in sphinx

sdk/core/azure-core/azure/core/exceptions.py
xiangyan99: I don't find there is a character limit in str()?

annatisch: It's in our guidelines - but looks to only be specified for `repr`:
https://azure.github.io/azure-sdk/python_implementation.html#python-models-repr-length

So str might be okay?

annatisch (approved):

PR #18242, opened 2021-04-22
[Blob]Add test upload_blob_from_url with encryption_scope

xiafu-msft:


sdk/storage/azure-storage-blob/tests/test_block_blob_async.py
```diff
@@ -150,6 +150,27 @@ async def test_upload_blob_with_and_without_overwrite(
...
         self.assertEqual(new_blob_content, b'source blob data')
 
+    @pytest.mark.playback_test_only
+    @GlobalStorageAccountPreparer()
+    async def test_upload_blob_from_url_with_encryption_scope(
+            self, resource_group, location, storage_account, storage_account_key):
+        await self._setup(storage_account, storage_account_key)
+        blob = await self._create_blob(data=b"source blob data", overwrite=True, encryption_scope="test")
+        # Act
+        sas = "se=2021-04-22T23%3A27%3A18Z&sp=rt&sv=2020-06-12&sr=b&sig=IqnyM6/Um7Fgt3VubBdAN8qvTV2roX4wjxvsPhF%2BSpA%3D"
```
tasherif-msft, line 160: why do we have a sas here?

tasherif-msft: Lets make the container public instead and make the test live as well

xf58:
@tasherif-msft it looks like rebase could fix the problem, it looks like the failed test is not related to this pr

Issue #17443, opened 2021-03-19
[Key Vault] Redact test values that trigger CredScan warnings


mccoyp:
Some values in Key Vault test recordings, which aren't sensitive per their test resource status, trigger warnings during credential scans. [Here](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=765215&view=logs&j=0874ffeb-2919-5b4b-20de-16a97970b94c&t=439e917e-962a-53e0-8a4b-b8bb4f8ed372&l=66) is an example of a warning generated because of a test certificate's value.

These certificate contents have been labeled as permissible since they're not sensitive, but we should still redact these contents from recordings where possible.

mccoyp:
Moving into a milestone where we re-record with the new test proxy system.

mccoyp:
Moving to backlog for more investigation at a later time. Redacting certificate values would be straightforward with body key sanitizers that sanitize `cer`, `csr`, and `value` fields in `azure-keyvault-certificates` recordings (which currently contain test certificate content). Trying this caused failures in playback tests, though:
```
msrest.exceptions.DeserializationError: Unable to deserialize response data. Data: certificate-content, bytearray, Error: Incorrect padding
```
Certificate operations need to be able to be decoded from responses into a `CertificateOperation` object, and this seems to require correct certificate information in recordings. For the time being, I can't think of a straightforward way to sanitize our recordings in a functional way.

mccoyp:
Closing since this issue hasn't had any progress and isn't pressing.

PR #18245, opened 2021-04-22
Add Key Vault test certificate suppression

mccoyp:
This suppresses warnings for a test certificate used by Key Vault. I have an [issue](https://github.com/Azure/azure-sdk-for-python/issues/17443) open to redact these values from recordings as a more long-term solution, but this will help in the short-term.


weshaggard (approved):
I'm Ok with this short term but I'm happy to see you have an issue for the long term fix which is to redact these instead.

PR #18249, opened 2021-04-23
Add missing dependency on azure-mgmt-core

scbedd:
This is to silence the nightly error.


sdk/regionmove/azure-mgmt-regionmove/setup.py
```diff
@@ -83,6 +83,7 @@
         'msrest>=0.5.0',
         'msrestazure>=0.4.32,<2.0.0',
```
jsntcy, line 84:
```suggestion
```

jsntcy:
`regionmove` may probably be replaced by `resourcemover`, let me confirm with service team, if yes, we can totally delete `regionmove` folder.

msyyc (approved):

PR #18260, opened 2021-04-23
[Tables] Use lists for batching

annatisch:


sdk/tables/azure-data-tables/azure/data/tables/aio/_table_batch_async.py
```diff
@@ -182,11 +182,11 @@ def _batch_create_entity(
...
         )
-        self._requests.append(request)
+        self.requests.append(request)
 
     _batch_create_entity.metadata = {"url": "/{table}"}  # type: ignore
 
-    def update_entity(
+    def update(
         self,
         entity,  # type: Union[TableEntity, Dict[str,str]]
```
seankane-msft, line 191: Can we update the type hints on entity here to use the `EntityType` defined previously?

sdk/tables/azure-data-tables/azure/data/tables/aio/_table_client_async.py
```diff
@@ -37,19 +30,19 @@
...
         self,
-        account_url,  # type: str
-        table_name,  # type: str
-        credential=None,  # type: str
-        **kwargs  # type: Any
-    ):
-        # type: (...) -> None
+        account_url: str,
+        table_name: str,
+        credential: Optional[Union[AzureSasCredential]] = None,
```
seankane-msft, line 43: `AzureSasCredential` or `NamedKeyCredential` right?

annatisch: Right - but the work to support NamedKeyCredential is outside the scope of this PR - so just leaving the typehint for what's supported today for now :)

seankane-msft (approved):
LGTM :shipit:

Issue #18259, opened 2021-04-23
[ServiceBus] GA AzureSasCredential


yunhaoling:
extract the code related to AzureSasCredential to the master branch from the feature branch
https://github.com/Azure/azure-sdk-for-python/tree/feature/servicebus%2F7.2.0b1/sdk/servicebus/azure-servicebus

PR #18264, opened 2021-04-23
[ServiceBus] GA Azure SAS Credential Support

yunhaoling:
addressing issue: https://github.com/Azure/azure-sdk-for-python/issues/18259


swathipil (approved):
LGTM

sdk/servicebus/azure-servicebus/azure/servicebus/_common/utils.py
```diff
@@ -330,3 +333,14 @@ def get_receive_links(messages):
     except AttributeError:
         pass
     return links
+
+
+def parse_sas_credential(credential):
+    # type: (AzureSasCredential) -> Tuple
+    sas = credential.signature
```
annatisch, line 340: Maybe `credential.signature.lower()` just to be safe?

sdk/servicebus/azure-servicebus/azure/servicebus/_base_handler.py
```diff
@@ -196,10 +196,28 @@ def get_token(self, *scopes, **kwargs):  # pylint:disable=unused-argument
             raise ValueError("No token scope provided.")
         return _generate_sas_token(scopes[0], self.policy, self.key)
 
+class AzureSasTokenCredential(object):
```
annatisch, line 199: Let's give this a more specific name to avoid any developer confusion between this and the Core credential object.
Something like `ServiceBusSasToken`

annatisch (approved):

sdk/servicebus/azure-servicebus/azure/servicebus/_common/utils.py
yunhaoling: as far as I know, signature is case-sensitive -- if we call lower(), then the signature would become invalid.
is there anything else I'm unaware of?

sdk/servicebus/azure-servicebus/azure/servicebus/_base_handler.py
yunhaoling: renamed to ServiceBusAzureSasTokenCredential (ServiceBusSasTokenCredential is already taken)

PR #18267, opened 2021-04-23
[Tables] remove client side validation of entity

seankane-msft:


sdk/tables/azure-data-tables/azure/data/tables/_table_client.py
```diff
@@ -353,6 +354,16 @@ def create_entity(
             return _trim_service_metadata(metadata)
         except ResourceNotFoundError as error:
             _process_table_error(error)
+        except HttpResponseError as error:
```
annatisch, line 357: Actually it looks like there might have been a bug here... line 355 should not be catching `ResourceNotFoundError`, _all_ HttpResponseErrors need to go through decoding in `_process_table_error` - as you should not need to do that awkward code below in order to get the error code.

sdk/tables/azure-data-tables/azure/data/tables/_table_client.py
```diff
@@ -351,7 +348,15 @@ def create_entity(
...
-        except ResourceNotFoundError as error:
+        except HttpResponseError as error:
+            try:
+                if error.model.additional_properties["odata.error"]["code"] == "PropertiesNeedValue":
+                    if entity.get("PartitionKey") is None:
+                        raise ValueError("PartitionKey must be present in an entity")
+                    if entity.get("RowKey") is None:
+                        raise ValueError("RowKey must be present in an entity")
+            except AttributeError:
+                raise error
```
annatisch, line 359: Let's wait till my batch PR is merged - then you'll be able to replace this code with:
```
from ._error import _decode_error, _reraise_error
...
    decoded = _decode_error(error)
    if decoded.error_code == "PropertiesNeedValue":
        ....
    _reraise_error(decoded)
```

annatisch: You didn't want to update to use the decoded error responses?

PR #18277, opened 2021-04-24
Remove implicit dependency on successful completion of verify-links

scbedd:
It's an eng/common/scripts which uses a specific function from within `LanguageSettings.ps1`.

That specific function actually needs `packaging` to be installed. This happens in `analyze_deps.py`, so we simply move this below it without burning any more time on installation.

### DIFF
--- a/eng/pipelines/templates/steps/analyze.yml
+++ b/eng/pipelines/templates/steps/analyze.yml
@@ -9,6 +9,16 @@ parameters:
 steps:
   - template: /eng/pipelines/templates/steps/analyze_dependency.yml
 
+  - template: /eng/common/pipelines/templates/steps/verify-links.yml
+    parameters:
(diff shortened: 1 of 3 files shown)


eng/pipelines/templates/steps/analyze.yml
```diff
@@ -9,6 +9,16 @@ parameters:
 steps:
   - template: /eng/pipelines/templates/steps/analyze_dependency.yml
 
+  - template: /eng/common/pipelines/templates/steps/verify-links.yml
```
weshaggard, line 12: I don't have a problem with moving this but I do don't like doing it based on a hidden dependency.  Is this just an optimization or is it a functional issue? I'd like to have this script work in isolation and not depend on analayze_dependency doing setup work for it.

seankane-msft (approved):

PR #18301, opened 2021-04-26
Refine `ARMPipelineClient.__init__`

jiasli:
Refine #18238's changes on `ARMPipelineClient.__init__`

### DIFF
--- a/sdk/core/azure-mgmt-core/azure/mgmt/core/_pipeline_client.py
+++ b/sdk/core/azure-mgmt-core/azure/mgmt/core/_pipeline_client.py
@@ -52,13 +52,13 @@ class ARMPipelineClient(PipelineClient):
                 )
             per_call_policies = kwargs.get('per_call_policies', [])
             if isinstance(per_call_policies, Iterable):
-                per_call_policies.append(ARMAutoResourceProviderRegistrationPolicy())
+                per_call_policies = list(per_call_policies).append(ARMAutoResourceProviderRegistrationPolicy())
             else:
                 per_call_policies = [per_call_policies,
                                      ARMAutoResourceProviderRegistrationPolicy()]
             kwargs["per_call_policies"] = per_call_policies
             config = kwargs.get('config')
             if not config.http_logging_policy:
-                config.http_logging_policy = kwargs.get('http_logging_policy', ARMHttpLoggingPolicy(**kwargs))
+                config.http_logging_policy = ARMHttpLoggingPolicy(**kwargs)
             kwargs["config"] = config
         super(ARMPipelineClient, self).__init__(base_url, **kwargs)


sdk/core/azure-mgmt-core/azure/mgmt/core/_pipeline_client.py
```diff
@@ -52,13 +52,13 @@ def __init__(self, base_url, **kwargs):
                 )
             per_call_policies = kwargs.get('per_call_policies', [])
             if isinstance(per_call_policies, Iterable):
-                per_call_policies.append(ARMAutoResourceProviderRegistrationPolicy())
+                per_call_policies = list(per_call_policies).append(ARMAutoResourceProviderRegistrationPolicy())
```
jiasli, line 55: Fix https://github.com/Azure/azure-sdk-for-python/pull/18238#discussion_r620131467

`append` will fail if `per_call_policies` is a `tuple`, instead of a `list`:

> AttributeError: 'tuple' object has no attribute 'append'

Convert to `per_call_policies` to a `list` first.

sdk/core/azure-mgmt-core/azure/mgmt/core/_pipeline_client.py
```diff
@@ -52,13 +52,13 @@ def __init__(self, base_url, **kwargs):
...
             else:
                 per_call_policies = [per_call_policies,
                                      ARMAutoResourceProviderRegistrationPolicy()]
             kwargs["per_call_policies"] = per_call_policies
             config = kwargs.get('config')
             if not config.http_logging_policy:
-                config.http_logging_policy = kwargs.get('http_logging_policy', ARMHttpLoggingPolicy(**kwargs))
+                config.http_logging_policy = ARMHttpLoggingPolicy(**kwargs)
```
jiasli, line 62: Fix https://github.com/Azure/azure-sdk-for-python/pull/18238#discussion_r619072800

We should forbid the usage of setting `http_logging_policy` with `kwargs` here, because it

- causes duplication with SDK's code https://github.com/Azure/azure-sdk-for-python/blob/28545a0bfd5586dcf2de8162cd4af9bfaa72f7f6/sdk/resources/azure-mgmt-resource/azure/mgmt/resource/subscriptions/_configuration.py#L59
- allows 2 ways to set `http_logging_policy` (with either `kwargs` or `config`), which can easily leads to inconsistency

xiangyan99: We have a clear priority order which is config > kwargs > default one.

I would suggest we keeping it because we want to align the priority with others.

jiasli: However, in specific client like `SubscriptionClient`, `http_logging_policy` in `kwargs` is firstly used to set `self._config.http_logging_policy`:

https://github.com/Azure/azure-sdk-for-python/blob/28545a0bfd5586dcf2de8162cd4af9bfaa72f7f6/sdk/resources/azure-mgmt-resource/azure/mgmt/resource/subscriptions/_subscription_client.py#L76

https://github.com/Azure/azure-sdk-for-python/blob/28545a0bfd5586dcf2de8162cd4af9bfaa72f7f6/sdk/resources/azure-mgmt-resource/azure/mgmt/resource/subscriptions/_configuration.py#L59

Then it is used as the `http_logging_policy` attribute of the `PipelineClient`:

https://github.com/Azure/azure-sdk-for-python/blob/227c700c3bae19a8f17327a3901ebd6cd5ac59f6/sdk/core/azure-mgmt-core/azure/mgmt/core/_pipeline_client.py#L61-L62

So the priority/flow here is more like kwargs > config > default one. 

Also, since `PipelineClient` doesn't take any policies from `kwargs`, why don't we remove the support for passing a `http_logging_policy` via `kwargs` to `ARMPipelineClient` as well, just to keep `ARMPipelineClient` simple (only `config` is supported)? 🤔

xiangyan99: In auto generated code (90% of the case), we will generate config which accept http_logging_policy. So http_logging_policy  in kwargs works in most of the case except the one in azure-core.

PR #18303, opened 2021-04-26
I add DataLakeAnalyticsScopeActivity model.

pelanzag:
At the moment, the only way of submitting Scope jobs programmatically is through the Rest API or UI. This change adds this functionality to the Python SDK.


kalyanrk61:
I don't think this is right way to do it as Scope is MSFT Internal only activity, its not visible for external customers. I think this will create security concerns for MSFT.

kalyanrk61:
@pelanzag please don't merge this PR, this needs to be packaged separately for internal customers.

pelanzag:
@kalyanrk61 who can I contact to get this packaged separately for internal customers? Can you point me to some docs or guidelines, please?

Issue #18287, opened 2021-04-25
Hardcoding prefixname in create_resource causes ResoruceNotFound when deleting resource group


00Kai0:
Hi @seankane-msft , can you remove this prefix in [create_resource](https://github.com/Azure/azure-sdk-for-python/blob/master/tools/azure-sdk-tools/devtools_testutils/resource_testcase.py).

It causes a problem that our testcase couldn't remove test resource group correctly.
I think we don't need hardcode this prefix of group name, it generated by testcase and test script name.
If it's needed, we could rename our test script name for tracking purposes.

cc: @jsntcy @msyyc 

Log:
```
__init__.py                235 INFO     <Request (DELETE) https://management.azure.com/subscriptions/00000000-0000-0000-0000-000000000000/resourcegroups/test_cli_mgmt_automation_test_automation53db109a?api-version=2020-10-01> not in cassette, sending to real server
cassette.py                232 INFO     Appending request <Request (DELETE) https://management.azure.com/subscriptions/00000000-0000-0000-0000-000000000000/resourcegroups/test_cli_mgmt_automation_test_automation53db109a?api-version=2020-10-01> and response {'status': {'code': 404, 'message': 'Not Found'}, 'headers': {'x-ms-failure-cause': ['gateway'], 'Content-Type': ['application/json; charset=utf-8'], 'x-ms-ratelimit-remaining-subscription-deletes': ['14999'], 'x-ms-correlation-request-id': ['84b14f55-1442-422a-842d-d9c3873d99dd'], 'Cache-Control': ['no-cache'], 'Pragma': ['no-cache'], 'Date': ['Sun, 25 Apr 2021 05:45:51 GMT'], 'Expires': ['-1'], 'x-ms-request-id': ['84b14f55-1442-422a-842d-d9c3873d99dd'], 'Strict-Transport-Security': ['max-age=31536000; includeSubDomains'], 'Content-Length': ['140'], 'X-Content-Type-Options': ['nosniff'], 'x-ms-routing-request-id': ['SOUTHEASTASIA:20210425T054551Z:84b14f55-1442-422a-842d-d9c3873d99dd']}, 'body': {'string': b'{"error":{"code":"ResourceGroupNotFound","message":"Resource group \'test_cli_mgmt_automation_test_automation53db109a\' could not be found."}}'}}
_universal.py              452 INFO     Response status: 404
_universal.py              453 INFO     Response headers:
[...]

PR #18304, opened 2021-04-26
[EngSys] updating the delete methods

seankane-msft:
#18287

### DIFF
--- a/tools/azure-sdk-tools/devtools_testutils/resource_testcase.py
+++ b/tools/azure-sdk-tools/devtools_testutils/resource_testcase.py
@@ -63,6 +63,12 @@ class ResourceGroupPreparer(AzureMgmtPreparer):
         self.set_cache(use_cache, parameter_name, name_prefix)
         self.delete_after_tag_timedelta = delete_after_tag_timedelta
 
+    def _prefix_name(self, name):
+        name = u"rgpy-" + name
+        if len(name) > 90:
+            name = name[:90]
+        return name
+
     def create_resource(self, name, **kwargs):
         if self.is_live and self._need_creation:
             self.client = self.create_mgmt_client(ResourceManagementClient)
@@ -76,9 +82,7 @@ class ResourceGroupPreparer(AzureMgmtPreparer):
             parameters["tags"]["BuildReason"] = os.environ.get("BUILD_REASON", "local")
             try:
                 # Prefixing all RGs created here with 'rgpy-' for tracing purposes
-                name = u"rgpy-" + name
-                if len(name) > 90:
-                    name = name[:90]
+                name = self._prefix_name(name)
                 logging.info(
                     "Attempting to create a Resource Group with name {} and parameters {}".format(name, parameters)
                 )
@@ -101,9 +105,7 @@ class ResourceGroupPreparer(AzureMgmtPreparer):
 
     def remove_resource(self, name, **kwargs):
         if self.is_live and self._need_creation:
-            name = u"rgpy-" + name
-            if len(name) > 90:
-                name = name[:90]
+            name = self._prefix_name(name)
             try:
                 if "wait_timeout" in kwargs:
                     azure_poller = self.client.resource_groups.begin_delete(name)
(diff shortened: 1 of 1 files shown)


tools/azure-sdk-tools/devtools_testutils/resource_testcase.py
```diff
@@ -101,6 +101,9 @@ def create_resource(self, name, **kwargs):
 
     def remove_resource(self, name, **kwargs):
         if self.is_live and self._need_creation:
+            name = u"rgpy-" + name
```
weshaggard, line 104: Oh this must have broken ever since you added the prefix. I was under an incorrect assumption when I responded to issue I thought they were hardcoding something in the test code.

weshaggard (approved):

tools/azure-sdk-tools/devtools_testutils/resource_testcase.py
jsntcy: I don't think adding this hardcode tricky logic in both `create_resource` and `remove_resource` is a good way. Maybe you can consider adding prefix for the `name` before you pass `name` parameter into `create_resource` and `remove_resource`.

jsntcy: Also the log shows the wrong name (the name without prefix), actually we create resource with prefix.
```Python
                logging.info(
                    "Attempting to create a Resource Group with name {} and parameters {}".format(name, parameters)
                )
                # Prefixing all RGs created here with 'rgpy-' for tracing purposes
                name = u"rgpy-" + name
                if len(name) > 90:
                    name = name[:90]
                self.resource = self.client.resource_groups.create_or_update(name, parameters)
```

jsntcy (changes requested):
Please take a look at the comments.

tools/azure-sdk-tools/devtools_testutils/resource_testcase.py
seankane-msft: > I don't think adding this hardcode tricky logic in both `create_resource` and `remove_resource` is a good way. Maybe you can consider adding prefix for the name before you pass name parameter into `create_resource` and `remove_resource`.

@jsntcy I don't understand what you mean by this, can you elaborate or provide a code snippet?

weshaggard: I'm not sure we would want to update all the callers of create/remove resources to have them pass in the name with the prefix but I do think it is a good idea to factor the name updates into a common function that we call at the beginning of both create and remove resources. 

Also from a quick look it seems as though there is already a name_prefix value in this class which serves a slight different but similar purpose. We might want to reconcile the concepts and perhaps have a rgname_prefix or something along those lines.

jsntcy: Yes, agree with @weshaggard, since both `create_resource` and `remove_resource` are in class `ResourceGroupPreparer`, you can define a variable in class which can be shared among all functions in class `ResourceGroupPreparer`.

Not sure if we can reuse `self.name_prefix` which already exists in class `ResourceGroupPreparer`.
https://github.com/Azure/azure-sdk-for-python/blob/master/tools/azure-sdk-tools/devtools_testutils/resource_testcase.py#L62

seankane-msft: @jsntcy a few weeks ago our Subscription ID hit the maximum number of resource groups allowed, and we had no information on hundreds of RGs. Wes and I added the prefix so we could easily determine where at  least some of the generated RGs were coming from.

tools/azure-sdk-tools/devtools_testutils/resource_testcase.py
```diff
@@ -75,13 +75,13 @@ def create_resource(self, name, **kwargs):
             parameters["tags"]["BuildNumber"] = os.environ.get("BUILD_BUILDNUMBER", "local")
             parameters["tags"]["BuildReason"] = os.environ.get("BUILD_REASON", "local")
             try:
-                logging.info(
-                    "Attempting to create a Resource Group with name {} and parameters {}".format(name, parameters)
-                )
                 # Prefixing all RGs created here with 'rgpy-' for tracing purposes
                 name = u"rgpy-" + name
                 if len(name) > 90:
                     name = name[:90]
```
seankane-msft, line 81: split this duplicated code into a method

lmazuel (approved):

PR #18306, opened 2021-04-26
[EngSys] adding nspkgs to same directory

seankane-msft:
#18248


scbedd (changes requested):
@seankane-msft do you want to to include `azure` and `azure-mgmt` nspkgs?

They're in the `core/` folder if yes. 

Also note that that with this change, please check `dev_requirements.txt` for any references to the relative paths to these npkgs. I would not be surprised if they existed around various packages in the repo.

seankane-msft:
> @seankane-msft do you want to to include `azure` and `azure-mgmt` nspkgs?
Yes, I wanted to start the CI and see how it fails. Goal is to put all `nspkg` in this folder

> Also note that that with this change, please check `dev_requirements.txt` for any references to the relative paths to these npkgs. I would not be surprised if they existed around various packages in the repo.
I will make these changes too.

kristapratico:
@seankane-msft  is this only for track 2 nspkg or should track 1 be included too?

seankane-msft:
> @seankane-msft is this only for track 2 nspkg or should track 1 be included too?

@kristapratico I believe both track1 and track2

kristapratico:
> > @seankane-msft is this only for track 2 nspkg or should track 1 be included too?
> 
> @kristapratico I believe both track1 and track2

Okay, just checking. There are several track 1 nspkg under the cognitiveservices directory as well.

scbedd:
@seankane-msft for the failing readme analysis failures. Instead of updating the urls to each readme, I recommend just excluding the folder entirely and removing all the individual exclusions.

```
omitted_paths:
 - <other paths>
 - sdk/nspkg/*
```

Should work just fine.

scbedd (approved):

Issue #14336, opened 2020-10-07
[ServiceBus] Ensure autolockreceiver halts session renewal on volntary close, determine if desired for messages.


KieranBrantnerMagee:
AutoLockRenew doesn't appear to halt itself when a session is closed voluntarily.

This probably applies to normal receivers via message renewal as well? But what about our 'fallback' mechanism to allow settlement even after close.  Determine if we want this as well, but the former item is certainly desired.

Normal process, implement, tests, document etc.

ramya-rao-a:
@rakshith91 Please note that we are moving this out of the November milestone since it can be added post GA without it being a breaking change. If you were planning to start on this, please pick the next item for Service Bus in the November milestone instead.

I discussed this with @KieranBrantnerMagee over chat

rakshith91:
Unable to reproduce this - however this would need a validation.
Having a full integration test for "renewer halts properly when session receiver shuts down" may be worthwhile.

yunhaoling:
writing a test case for this

swathipil:
if potential fix needs to be added to set `ALR._shutdown.is_set()` to True when receiver/session is closed voluntarily (not just a test case), it will probably be around here: https://github.com/Azure/azure-sdk-for-python/blob/97978c41e2939aa73fb6e7fec905896c296f9114/sdk/servicebus/azure-servicebus/azure/servicebus/_base_handler.py#L479

PR #18315, opened 2021-04-26
[Service Bus] add test to check halt of ALR on session voluntary close

swathipil:
fixes: #14336

### DIFF
--- a/sdk/servicebus/azure-servicebus/tests/test_sessions.py
+++ b/sdk/servicebus/azure-servicebus/tests/test_sessions.py
@@ -714,22 +714,24 @@ class ServiceBusSessionTests(AzureMgmtTestCase):
                                 assert isinstance(e.inner_exception, AutoLockRenewTimeout)
                             messages.append(message)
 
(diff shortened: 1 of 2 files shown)


sdk/servicebus/azure-servicebus/tests/test_sessions.py
```diff
@@ -731,6 +731,27 @@ def lock_lost_callback(renewable, error):
             renewer.close()
             assert len(messages) == 2
 
+        # test voluntary halt of auto lock renewer when session is closed
```
swathipil, line 734: hey @rakshith91, does this test look similar to what you did when you were [trying to reproduce](https://github.com/Azure/azure-sdk-for-python/issues/14336#issuecomment-721246436)?

rakshith91: yes - this seems right to me

rakshith91 (approved):

sdk/servicebus/azure-servicebus/tests/test_sessions.py
yunhaoling: looks good to me -- we should cover async well.

PR #18317, opened 2021-04-26
[Identity] CredScan suppressions for identity

seankane-msft:
adding to identity scrubber to store values in a file, adding the file to the cred scan suppression

### DIFF
--- a/sdk/identity/azure-identity/tests/recording_processors.py
+++ b/sdk/identity/azure-identity/tests/recording_processors.py
@@ -25,7 +25,7 @@ SECRETS = frozenset({
 })
 
 
-MGMT_SETTINGS_REAL = "_mgmt_settings_real_identity.py"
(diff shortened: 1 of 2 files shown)


sdk/identity/azure-identity/tests/recording_processors.py
```diff
@@ -40,12 +43,19 @@ def process_response(self, response):
         except (KeyError, ValueError):
             return response
 
-        for field in body:
-            if field in SECRETS:
-                # record a hash of the secret instead of a simple replacement like "redacted"
-                # because some tests (e.g. for CAE) require unique, consistent values
-                digest = hashlib.sha256(six.ensure_binary(body[field])).digest()
-                body[field] = six.ensure_str(binascii.hexlify(digest))
+        with open(MGMT_SETTINGS_FAKE_IDENTITY, "w+") as outfile:
```
weshaggard, line 46: What should we generate this each time? What happens if this file ends up getting checked in?

PR #18327, opened 2021-04-26
add py.typed

xiangyan99:


sdk/appconfiguration/azure-appconfiguration/setup.py
```diff
@@ -64,6 +64,7 @@
 setup(
     name=PACKAGE_NAME,
     version=version,
+    include_package_data=True,
```
kristapratico, line 67: reading this makes me think that just adding to MANIFEST.in and providing this option should suffice: https://setuptools.readthedocs.io/en/latest/userguide/datafiles.html#data-files-support

If we include `package_data`, I think the key needs to be the name of the package.

kristapratico (approved):

PR #18333, opened 2021-04-27
Add Conda Testing

scbedd:
Due to a [known issue](https://github.com/tox-dev/tox-conda/issues/50) with `tox-conda`, I'm not going to be utilizing it at this juncture.

Instead, going to allow the user to create a conda environment and invoke the test themselves.


eng/pipelines/templates/stages/archetype-conda-release.yml
```diff
@@ -32,3 +32,26 @@ stages:
                         Get-ChildItem -Recurse $(Pipeline.Workspace)/${{parameters.ArtifactName}}/${{artifact.name}}
                       workingDirectory: $(Pipeline.Workspace)
                       displayName: Output Visible Conda Artifacts
+      - stage: Test_${{ replace(artifact.name, '-', '_') }}_Distribution
+        displayName: 'Conda Release: ${{artifact.name}}'
+        dependsOn: Release_${{ replace(artifact.name, '-', '_') }}_To_Blob
```
weshaggard, line 37: Given you look to depend on the previous stage is there a reason to break this up into a stage as opposed to just another job in the previous stage?

eng/pipelines/templates/steps/test-conda.yml
```diff
@@ -0,0 +1,95 @@
...
+          -d "$(conda.output)"
+          -b "$(conda.build)"
+          -m "$(Build.SourcesDirectory)/sdk/${{ parameters.ServiceDirectory }}/${{ artifact.meta_source }}"
+          -r "${{ artifact.common_root }}"
+          -n "${{ artifact.name }}"
+          -s "${{ parameters.ServiceDirectory }}"
+          -e "$(Build.SourcesDirectory)/eng/conda_env.yml"
+          -c "$(Build.SourcesDirectory)/sdk/${{ parameters.ServiceDirectory }}/ci.yml"
+
+    - bash: |
```
weshaggard, line 67: Is thee a reason you are using bash here in this template? It will end up limiting its usage to only linux. Do we need to ever run these tests on other platforms?

eng/pipelines/templates/stages/archetype-conda-release.yml
scbedd: Nope. I'm ripping that out.

eng/pipelines/templates/steps/test-conda.yml
scbedd: Totally ack. It's going to have to change. I'm just trying to get to the place where it's running -> get it working cross plat.

scbedd:
Closing PR because I need to make changes within the `sdk/core` directory and don't want to spam the build pool while I do so.

PR #18348, opened 2021-04-27
adding container registry

seankane-msft:
Was having issues pushing directly to your branch


sdk/containerregistry/azure-containerregistry/setup.py
```diff
@@ -58,6 +59,9 @@
             "azure",
         ]
     ),
+    package_data={
```
xiangyan99, line 62: Seems like we don't need this.

MANIFEST.in + include_package_data=True, is good enough.

xiangyan99 (approved):

PR #18392, opened 2021-04-28
[Container Registry] Update API for Beta 2

seankane-msft:


sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_registry_client.py
```diff
@@ -165,18 +166,30 @@ def get_next(next_link=None):
...
 
     @distributed_trace
-    def get_repository_client(self, repository, **kwargs):
-        # type: (str, Dict[str, Any]) -> ContainerRepositoryClient
-        """Get a repository client
+    def get_repository(self, repository, **kwargs):
+        # type: (str, Dict[str, Any]) -> ContainerRepository
+        """Get a Container Repository object
 
         :param str repository: The repository to create a client for
```
annatisch, line 173: This is documented as a `str`, though type-hint says it can also be a `dict` - is that why it wasn't changed to `repository_name` like the others?

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_repository.py
```diff
@@ -0,0 +1,223 @@
...
+        :type repository: str
+        :param credential: The credential with which to authenticate
+        :type credential: :class:`~azure.core.credentials.TokenCredential`
+        :returns: None
+        :raises: None
+        """
+        if not endpoint.startswith("https://") and not endpoint.startswith("http://"):
+            endpoint = "https://" + endpoint
+        self._endpoint = endpoint
+        self.repository = repository
```
annatisch, line 51: Should this just be `self.name`?

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_repository.py
```diff
@@ -0,0 +1,223 @@
...
+)
+from ._registry_artifact import RegistryArtifact
+
+if TYPE_CHECKING:
+    from typing import Any, Dict
+    from azure.core.credentials import TokenCredential
+    from ._models import ContentProperties
+
+
+class ContainerRepository(ContainerRegistryBaseClient):
```
annatisch, line 34: Why is this no longer a client? Is it because a user cannot instantiate it themselves?

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_models.py
```diff
@@ -156,11 +155,21 @@ def _from_generated(cls, generated):
...
+            created_on=self.created_on,
+            last_updated_on=self.last_updated_on,
+            manifest_count=self.manifest_count,
+            tag_count=self.tag_count,
+            writeable_properties=self.writeable_properties._to_generated(),  # pylint: disable=protected-access
         )
 
 
-class RegistryArtifactOrderBy(str, Enum):
+class ManifestOrderBy(str, Enum):
```
annatisch, line 172: I feel like this should just be called `ManifestOrder` or `ManifestOrdering`.
Then it reads better: `order_by=ManifestOrder.LAST_UPDATE_TIME_DESCENDING`

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_registry_artifact.py
```diff
@@ -30,15 +28,17 @@
     from ._models import ContentProperties
 
 
-class ContainerRepositoryClient(ContainerRegistryBaseClient):
-    def __init__(self, endpoint, repository, credential, **kwargs):
-        # type: (str, str, TokenCredential, Dict[str, Any]) -> None
-        """Create a ContainerRepositoryClient from an endpoint, repository name, and credential
+class RegistryArtifact(ContainerRegistryBaseClient):
+    def __init__(self, endpoint, repository, tag_or_digest, credential, **kwargs):
```
annatisch, line 32: It looks like this is also kind of a client - but it didn't have the pipeline/transport wrapping in the methods that construct it.

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_registry_client.py
seankane-msft: This is an inconsistency on my part. The param should be `repository_name` and it will be a str. The type-hint is meant to be: `def get_repository(self, repository: str, **kwargs: Any) -> ContainerRepository:

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_repository.py
seankane-msft: Yes! Thank you for the catch.

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_repository.py
seankane-msft: A user could instantiate it themselves but it would look a little awkward. We received feedback from the service team that the two clients made it confusing on where a user should start. Krystof and Anne Thompson came up with a design where both `ContainerRepository` and `RegistryArtifact` have client-like properties (being able to make service calls) but are more like models. This produces a more intuitive starting point with `ContainerRegistryClient`.

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_registry_artifact.py
seankane-msft: The reasoning is the same as for `ContainerRegistry`, but I do need to add the pipeline/transport wrapping constructs.

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_repository.py
```diff
@@ -0,0 +1,228 @@
...
+        :returns: :class:`~azure.containerregistry.RepositoryProperties`
+        :raises: :class:`~azure.core.exceptions.ResourceNotFoundError`
+        """
+        return RepositoryProperties._from_generated(  # pylint: disable=protected-access
+            self._client.container_registry.get_properties(self.name, **kwargs)
+        )
+
+    @distributed_trace
+    def list_manifests(self, **kwargs):
+        # type: (Dict[str, Any]) -> ItemPaged[ArtifactManifestProperties]
```
annatisch, line 81: Are the listed items here the same data that I would get from a `get_manifest_properties` call?
We're using the same response object for both - so just checking that they both include the same information.

sdk/containerregistry/azure-containerregistry/azure/containerregistry/aio/_async_registry_artifact.py
```diff
@@ -357,7 +181,7 @@ def list_tags(self, **kwargs: Dict[str, Any]) -> AsyncItemPaged[ArtifactTagPrope
             call will return values after last lexically
         :paramtype last: str
         :keyword order_by: Query parameter for ordering by time ascending or descending
-        :paramtype order_by: :class:`~azure.containerregistry.TagOrderBy`
+        :paramtype order_by: :class:`~azure.containerregistry.TagOrder`
```
annatisch, line 184: The param documentation should include `str` as a valid input where ever we also use enums :)
Though it doesn't need to be fixed for this preview.

sdk/containerregistry/azure-containerregistry/azure/containerregistry/__init__.py
```diff
@@ -7,28 +7,30 @@
...
 __version__ = VERSION
 
 __all__ = [
     "ContainerRegistryClient",
-    "ContainerRepositoryClient",
-    "DeleteRepositoryResult",
+    "ContainerRepository",
     "ContentProperties",
-    "ManifestOrderBy",
+    "DeleteRepositoryResult",
```
annatisch, line 29: What information do we get back from a delete operation?

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_models.py
```diff
@@ -132,7 +132,6 @@ class RepositoryProperties(object):
     :vartype last_updated_on: :class:`datetime.datetime`
     :ivar int manifest_count: Number of manifest in the repository
     :ivar str name: Name of the repository
-    :ivar str registry: Registry the repository belongs to
```
annatisch, line 135: We don't think this would be useful info?

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_repository.py
```diff
@@ -0,0 +1,228 @@
...
+        :param str name: The name of a repository
+        :param credential: The credential with which to authenticate
+        :type credential: :class:`~azure.core.credentials.TokenCredential`
+        :returns: None
+        :raises: None
+        """
+        if not endpoint.startswith("https://") and not endpoint.startswith("http://"):
+            endpoint = "https://" + endpoint
+        self._endpoint = endpoint
+        self.name = name
```
annatisch, line 50: Should this also have something like the property on `RegistryArtifact`:
```python
self.fully_qualified_name = self._endpoint + self.repository
```

annatisch (approved):

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_repository.py
seankane-msft: Yes they are, both calls return the same data.

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_models.py
seankane-msft: This was an error on my part, the registry is not included in the response object.

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_repository.py
seankane-msft: Yes adding it now.

PR #18397, opened 2021-04-28
[EngSys] added to python script to not fail if there is an existing PR

seankane-msft:
This PR adds to the `verify_autorest.py` file to only fail if the autorest is not the same AND there is not an existing PR out for autorest generation.

### DIFF
--- a/scripts/devops_tasks/verify_autorest.py
+++ b/scripts/devops_tasks/verify_autorest.py
@@ -6,9 +6,10 @@
 # --------------------------------------------------------------------------------------------
 
 import argparse
-import os
 import logging
-import sys
+import json
+import os
+import requests
 
 from common_tasks import run_check_call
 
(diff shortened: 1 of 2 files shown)


seankane-msft:
#18393

scripts/devops_tasks/verify_autorest.py
```diff
@@ -54,7 +55,31 @@ def find_swagger_folders(directory):
     return ret
 
 
-def check_diff(folder):
+def check_for_open_pr(service_directory):
```
weshaggard, line 58: You shouldn't need to do this if your branch name is consistent. As the create-pull-request template will not create an additional PR against a given branch if one already exists. That is how we generally handle this case in other places.

scripts/devops_tasks/verify_autorest.py
```diff
@@ -54,7 +55,31 @@ def find_swagger_folders(directory):
...
+        for link in body:
+            if link["title"] == pr_title:
+                return True
+
+        next_page = response.links["next"]["url"]
+        response = s.get(next_page)
+    return False
+
+
+def check_diff(folder, service_dir):
```
weshaggard, line 82: Similar you shouldn't need to manually do a git diff.  See https://github.com/Azure/azure-sdk-for-python/blob/master/eng/common/pipelines/templates/steps/git-push-changes.yml#L18

scripts/devops_tasks/verify_autorest.py
seankane-msft: That is great to know, I can close this PR entirely then

scripts/devops_tasks/verify_autorest.py
seankane-msft: The git diff is for users to be able to inspect which files are different in the pipeline on a failure.

weshaggard: OK if that is interesting for other cases that is fine. However I'd encourage you to do those type of command line calls inline script blocks in the yml file, unless you feel folks would want to run them locally as well. 

This is perhaps my preference for using a shell scripting language (script/bash/pwsh) when running command line tools as opposed to wrapping them in process calls in a higher level language like python. Just something to keep in mind not something that is necessarily actionable. :)

PR #18400, opened 2021-04-28
[Container Registry] Acr enums

seankane-msft:
Dependent on #18397


seankane-msft:
Dependent on #18392

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_models.py
```diff
@@ -78,22 +79,24 @@ def _from_generated(cls, gen):
 class ArtifactManifestProperties(object):
     """Represents properties of a registry artifact
 
-    :ivar str cpu_architecture: CPU Architecture of an artifact
+    :ivar architecture: CPU Architecture of an artifact
+    :vartype architecture: ~azure.containerregistry.ArtifactArchitecture
     :ivar created_on: Time and date an artifact was created
-    :vartype created_on: :class:`~datetime.datetime`
+    :vartype created_on: datetime.datetime
```
kristapratico, line 85: I think the `~` is required to get the link in documentation

```suggestion
    :vartype created_on: ~datetime.datetime
```

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_models.py
```diff
@@ -78,22 +79,24 @@ def _from_generated(cls, gen):
...
     :ivar str digest: Digest for the artifact
     :ivar last_updated_on: Time and date an artifact was last updated
-    :vartype last_updated_on: :class:`~datetime.datetime`
-    :ivar str operating_system: Operating system for the artifact
+    :vartype last_updated_on: datetime.datetime
+    :ivar operating_system: Operating system for the artifact
+    :vartype writeable_properties: ~azure.containerregistry.ArtifactOperatingSystem
```
kristapratico, line 90: It looks like this should be?

```suggestion
    :vartype operating_system: ~azure.containerregistry.ArtifactOperatingSystem
```

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_models.py
```diff
@@ -105,6 +108,16 @@ def __init__(self, **kwargs):
         if self.writeable_properties:
             self.writeable_properties = ContentProperties._from_generated(self.writeable_properties)
 
+        for data in ArtifactArchitecture:
+            if data.value == self.architecture:
+                self.architecture = data
+                break
```
kristapratico, line 114: You should be able to do `ArtifactArchitecture(self.architecture)` if you want to convert the string to the enum type. Is this an input or result model? I feel like python users prefer to work with strings but ignore me if this design decision was already made.

kristapratico (approved):

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_models.py
seankane-msft: This is a result model. The `ArtifactArchitecture` inherits from str and Enum, users should be able to work with strings with this model.

Issue #18047, opened 2021-04-14
[formrecognizer] sample/docs feedback


kristapratico:
1) Explain structure of result objects / where you can find tables, lines, etc. on the top level object so users can understand how to navigate the data. E.g. on RecognizedForm, in the model description docstring point out where tables can be found.

2) For custom form samples that use sample forms that have tables, demonstrate iterating through tables and printing data. Probably only include cell location and `text` to not overwhelm the output.

3) Create samples that show how to iterate over labeled tables - fixed and variable rows

PR #18409, opened 2021-04-29
[formrecognizer] adds samples for labeled tables and addresses sample/doc feedback

kristapratico:
Resolves https://github.com/Azure/azure-sdk-for-python/issues/18047

This also adds the training docs for the fixed/dynamic tables samples. We got feedback from the docs team that it was really helpful to have this for users.


sdk/formrecognizer/azure-ai-formrecognizer/samples/async_samples/sample_recognize_custom_forms_async.py
```diff
@@ -71,9 +71,18 @@ async def recognize_custom_forms(self):
                     print("...Label '{}' has value '{}' with a confidence score of {}".format(
                         field.label_data.text if field.label_data else name, field.value, field.confidence
                     ))
+                # [END recognize_custom_forms_async]
```
catalinaperalta, line 74: Why move this up?

catalinaperalta (commented):
Adding all the OCR files is mainly for customers to be able to see what's needed for training labeled tables?

sdk/formrecognizer/azure-ai-formrecognizer/samples/async_samples/sample_differentiate_output_labeled_tables_async.py
```diff
@@ -0,0 +1,138 @@
...
+    table tags don't have to replicate the structure of every table found in your form. Tables extracted
+    automatically by Form Recognizer will be included in the `tables` property under `RecognizedForm.pages`.
+
+    A conceptual explanation of using table tags to train your custom form model can be found in the
+    service documentation: https://docs.microsoft.com/azure/cognitive-services/form-recognizer/supervised-table-tags
+
+USAGE:
+    python sample_differentiate_output_labeled_tables_async.py
+
+    Set the environment dynamics with your own values before running the sample:
```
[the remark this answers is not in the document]
kristapratico: 😄 good catch, should be more careful with global replace

sdk/formrecognizer/azure-ai-formrecognizer/samples/async_samples/sample_recognize_custom_forms_async.py
kristapratico: I think my thought was that this sample snippet showcases custom form fields and I didn't want to detract from that with the tables code. But actually I agree, I'll move it back so it captures the whole snippet.

kristapratico:
> Adding all the OCR files is mainly for customers to be able to see what's needed for training labeled tables?

Yes, and they can download the files and put them in a container to train on. Let's them skip the whole labeling process and run the samples more easily.

catalinaperalta (approved):
🚢

sdk/formrecognizer/azure-ai-formrecognizer/samples/async_samples/sample_differentiate_output_labeled_tables_async.py
```diff
@@ -0,0 +1,138 @@
...
+FILE: sample_differentiate_output_labeled_tables_async.py
+
+DESCRIPTION:
+    This sample demonstrates the differences in output that arise when begin_recognize_custom_forms
+    is called with custom models trained with fixed vs. dynamic table tags.
+    The models used in this sample can be created in the sample_train_model_with_labels.py using the
+    training files in
+    https://github.com/Azure/azure-sdk-for-python/tree/master/sdk/formrecognizer/azure-ai-formrecognizer/samples/sample_forms/labeled_tables
+
+    Note that Form Recognizer automatically finds and extracts all tables in your documents whether the tables
```
maririos, line 19: I'm not sure I follow the purpose of this text. Could you expand on the reason why we need this?

sdk/formrecognizer/azure-ai-formrecognizer/samples/async_samples/sample_differentiate_output_labeled_tables_async.py
```diff
@@ -0,0 +1,138 @@
...
+"""
+
+import os
+import asyncio
+
+
+class TestDifferentiateOutputLabeledTablesAsync(object):
+
+    async def test_recognize_tables_fixed_rows_async(self):
+        from azure.core.credentials import AzureKeyCredential
```
maririos, line 44: Is this sample runnable?

sdk/formrecognizer/azure-ai-formrecognizer/samples/async_samples/sample_differentiate_output_labeled_tables_async.py
```diff
@@ -0,0 +1,138 @@
...
+        print("\n--------Recognizing labeled table with fixed rows--------\n")
+        for form in result:
+            for name, field in form.fields.items():
+                # substitute "table" for the label given to the table tag during training
+                # (if different than sample training docs)
+                if name == "table":
+                    for row_name, column in field.value.items():
+                        print("Row '{}' has columns:".format(row_name))
+                        for column_name, column_value in column.value.items():
+                            print("...Column '{}' with value '{}' and a confidence score of {}".format(
```
maririos, line 76: if `column_value.value` is None then would it print `Columbn with value None` ? or nothing

sdk/formrecognizer/azure-ai-formrecognizer/samples/async_samples/sample_recognize_custom_forms_async.py
```diff
@@ -71,6 +71,16 @@ async def recognize_custom_forms(self):
...
+
+                # iterate over tables on each page
+                for page in form.pages:
```
maririos, line 77: hmm why only tables? should we show selection mark, lines, etc? or not include it at all

sdk/formrecognizer/azure-ai-formrecognizer/samples/async_samples/sample_differentiate_output_labeled_tables_async.py
kristapratico: We've gotten feedback that our design makes the OCR tables hard to find. Didn't want to give the impression with this sample that you must label tables to get tables in the output. The service team noted this in the service docs and I thought it would be important to underscore that here as well.

sdk/formrecognizer/azure-ai-formrecognizer/samples/async_samples/sample_differentiate_output_labeled_tables_async.py
kristapratico: yes

sdk/formrecognizer/azure-ai-formrecognizer/samples/async_samples/sample_differentiate_output_labeled_tables_async.py
kristapratico: It will print "column with value None"

sdk/formrecognizer/azure-ai-formrecognizer/samples/async_samples/sample_recognize_custom_forms_async.py
kristapratico: docs team's feedback was specifically on highlighting tables. We already have a [sample ](https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/formrecognizer/azure-ai-formrecognizer/samples/sample_get_bounding_boxes.py) that goes into the lines, selection marks, words elements in more details.

sdk/formrecognizer/azure-ai-formrecognizer/samples/async_samples/sample_differentiate_output_labeled_tables_async.py
maririos: hmm are the model ids created every time the sample is run? or how does the magic work?
Not blocking your PR, just curious

sdk/formrecognizer/azure-ai-formrecognizer/samples/async_samples/sample_differentiate_output_labeled_tables_async.py
maririos: I see. I guess the part that is kind of confusing is 
>Therefore, you don't have to label every table from your form with a table tag and your
    table tags don't have to replicate the structure of every table found in your form.

Maybe something like:

..are tagged/labeled or not. Tables extracted automatically by Form Recognizer will be included in the `tables` property under `RecognizedForm.pages`. If this doesn't satisfy your needs, you can label the table.
A conceptual explanation ....

sdk/formrecognizer/azure-ai-formrecognizer/samples/async_samples/sample_recognize_custom_forms_async.py
maririos: But if someone only looks here, might think that the service only returns tables in the form type... so idk... maybe a comment after this that says there is also more information on pages? or at the beginning say some of the things that user can find in pages, and for example, here is the automatically recognized table: ...

sdk/formrecognizer/azure-ai-formrecognizer/samples/async_samples/sample_differentiate_output_labeled_tables_async.py
kristapratico: Got it - will update to your suggestion to make it less confusing.

sdk/formrecognizer/azure-ai-formrecognizer/samples/async_samples/sample_differentiate_output_labeled_tables_async.py
kristapratico: No, these are static model IDs that get passed in as environment variables. All samples are run from an existing FR resource, nothing created on the fly here.

sdk/formrecognizer/azure-ai-formrecognizer/samples/async_samples/sample_recognize_custom_forms_async.py
kristapratico: I think I was hoping to keep this sample more focused initially. But now that we are including tables here, I guess it would make sense to include lines and selection marks as well. will update

sdk/formrecognizer/azure-ai-formrecognizer/samples/async_samples/sample_differentiate_output_labeled_tables_async.py
maririos: I worry someone might delete the models and then all samples stop working, but it will be pretty clear, so easy to fix.
of course, if it ever becomes a problem the samples can be updated an in charge of creating the model too

maririos (approved):

PR #18430, opened 2021-04-29
[Container Registry] renaming TagProperties -> ArtifactTagProperties

seankane-msft:


seankane-msft:
Dependent on #18405

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_repository_client.py
```diff
@@ -3,7 +3,7 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
 # ------------------------------------
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any, Dict
```
iscai-msft, line 6: nit: there doesn't seem to be a reason to move this out of the TYPE_CHECKING block, at least looking at the changes you made in this pr

iscai-msft (approved):

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_repository_client.py
seankane-msft: Correct, I learned that all imports from typing can be grouped together so I made that change here while it was fresh in my mind.

PR #17489, opened 2021-03-22
Add ChallengeAuthenticationPolicy

chlowell:
Clients expecting authentication challenges should subclass ChallengeAuthenticationPolicy and override `on_challenge` to handle those challenges.

PR #18437, opened 2021-04-29
Extend BearerTokenCredentialPolicy to handle auth challenges

chlowell:
When its on_challenge method is not overridden, the ChallengeAuthenticationPolicy proposed in #17489 is functionally equivalent to the existing BearerTokenCredentialPolicy. It's reasonable then to wonder how the existing policy would look when extended to handle authentication challenges. Short answer: it would look something like this 😉

### DIFF
--- a/sdk/core/azure-core/azure/core/pipeline/policies/_authentication.py
+++ b/sdk/core/azure-core/azure/core/pipeline/policies/_authentication.py
@@ -71,7 +71,7 @@ class _BearerTokenCredentialPolicyBase(object):
         return not self._token or self._token.expires_on - time.time() < 300
 
 
-class BearerTokenCredentialPolicy(_BearerTokenCredentialPolicyBase, SansIOHTTPPolicy, HTTPPolicy):
+class BearerTokenCredentialPolicy(_BearerTokenCredentialPolicyBase, HTTPPolicy):
     """Adds a bearer token Authorization header to requests.
 
     :param credential: The credential.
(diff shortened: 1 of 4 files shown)


sdk/core/azure-core/azure/core/pipeline/policies/_authentication.py
```diff
@@ -71,7 +71,7 @@ def _need_new_token(self):
         return not self._token or self._token.expires_on - time.time() < 300
 
 
-class BearerTokenCredentialPolicy(_BearerTokenCredentialPolicyBase, SansIOHTTPPolicy):
+class BearerTokenCredentialPolicy(_BearerTokenCredentialPolicyBase, SansIOHTTPPolicy, HTTPPolicy):
```
xiangyan99, line 74: The the pipeline will call on_request or send or both? 

Do we have a design for this scenario? @annatisch

chlowell: After #16726, the pipeline will call this policy's send method (instead of _SansIOHTTPPolicyRunner.send). Another option is for this policy not to inherit SansIOHTTPPolicy.

johanste: Since this is an "AvecIOHTTPPolicy" (as opposed to Sans), I would remove the inheritance from `SansIOHTTPPolicy` and derive from the "correct" base class. I think that your implementation can call `self.on_request` and `self.on_response` to make sure they are called in a very similar fashion as before in order to handle any cases where someone has extended/derived from the policy, no?

chlowell: Have done. The extended policy follows _SansIOHTTPPolicyRunner's calling pattern such that it calls a subclass's `on_*` as the runner would.

sdk/core/azure-core/azure/core/pipeline/policies/_authentication.py
```diff
@@ -71,7 +71,7 @@ def _need_new_token(self):
         return not self._token or self._token.expires_on - time.time() < 300
 
 
-class BearerTokenCredentialPolicy(_BearerTokenCredentialPolicyBase, SansIOHTTPPolicy):
+class BearerTokenCredentialPolicy(_BearerTokenCredentialPolicyBase, HTTPPolicy):
```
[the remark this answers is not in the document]
chlowell: That's #16726. No conflict with this change, and this change doesn't require that PR.

sdk/core/azure-core/azure/core/pipeline/policies/_authentication.py
```diff
@@ -82,17 +82,94 @@ class BearerTokenCredentialPolicy(_BearerTokenCredentialPolicyBase, SansIOHTTPPo
...
-        """Adds a bearer token Authorization header to request and sends request to next policy.
+        """Called before the policy sends a request.
 
-        :param request: The pipeline request object
-        :type request: ~azure.core.pipeline.PipelineRequest
+        The base implementation authorizes the request with a bearer token.
+
+        :param ~azure.core.pipeline.PipelineRequest request: the request
         """
         self._enforce_https(request)
```
xiangyan99, line 91: How do you think about moving _enforce_https into shared utils?

xiangyan99 (approved):
LGTM

sdk/core/azure-core/azure/core/pipeline/policies/_authentication.py
chlowell: Makes sense the next time another policy needs it. Do we want the key and SAS policies to enforce HTTPS as well?

xiangyan99: I asked this because I saw ACR also used it. :)

Issue #18441, opened 2021-04-29
[Tables] Add support for AzureNamedKeyCredential


annatisch:
This should be in place of the current dict/str based implementation.
The only supported types for the `credential` parameter will now be `AzureNamedKeyCredential`, `AzureSasKeyCredential` or via connection string (with samples/readme etc updated to match).

The azure-core dependency version will also need to be updated to match.

seankane-msft:
@annatisch should this be `AzureKeyCredential` ?

annatisch:
No, `AzureNamedKeyCredential`
https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/core/azure-core/azure/core/credentials.py#L117

PR #18456, opened 2021-04-30
[Tables] Adds support for AzureNamedKeyCredential

seankane-msft:
#18441


sdk/tables/azure-data-tables/azure/data/tables/_authentication.py
```diff
@@ -32,6 +33,9 @@
     _wrap_exception,
 )
 
+if TYPE_CHECKING:
+    from typing import Union, Awaitable
```
annatisch, line 37: You can just import these from `typing` without needing to wrap it the `TYPE_CHECKING`.
You only need to do this for stuff imported from external libraries, like core, or other modules within this SDK.

sdk/tables/azure-data-tables/azure/data/tables/_authentication.py
```diff
@@ -82,10 +83,10 @@ def _get_canonicalized_resource(self, request):
                 )
             ):
                 uri_path = URL(uri_path)
-                return "/" + self.account_name + str(uri_path)
+                return "/" + self._credential.name + str(uri_path)
```
annatisch, line 86: Why is this one `.name`, but below it's `.named_key.name`?

sdk/tables/azure-data-tables/azure/data/tables/_shared_access_signature.py
```diff
@@ -32,13 +32,13 @@ def __init__(self, account_name, account_key, x_ms_version=DEFAULT_X_MS_VERSION)
         """
         :param str account_name:
             The storage account name used to generate the shared access signatures.
-        :param str account_key:
+        :param AzureNamedKeyCredential account_key:
```
annatisch, line 35: hmm not sure this is needed for SAS generation. Or if we think it is - maybe we would also remove the `account_name` parameter, as that's also included in the credential.

sdk/tables/azure-data-tables/samples/async_samples/sample_insert_delete_entities_async.py
```diff
@@ -75,7 +77,7 @@ async def delete_entity(self):
         from azure.core.exceptions import ResourceNotFoundError, ResourceExistsError
         from azure.core import MatchConditions
 
-        table_client = TableClient(account_url=self.account_url, credential=self.access_key, table_name=self.table_name)
+        table_client = TableClient(account_url=self.account_url, credential=AzureNamedKeyCredential(self.account_name, self.access_key), table_name=self.table_name)
```
annatisch, line 80: For the samples, can we move the instantiation of the credential out onto a separate line?
Helps with readability.

sdk/tables/azure-data-tables/samples/sample_create_client.py
```diff
@@ -54,7 +56,7 @@ def create_table_service_client(self):
         # Instantiate a TableServiceClient using a shared access key
         # [START create_table_service_client]
         from azure.data.tables import TableServiceClient
-        with TableServiceClient(account_url=self.account_url, credential=self.access_key) as table_service:
+        with TableServiceClient(account_url=self.account_url, credential=AzureNamedKeyCredential(self.account_name, self.access_key)) as table_service:
```
annatisch, line 59: Also for the samples, I think the import statement from `AzureNamedKeyCredential` should be moved into the `[START]` tag so that it shows up in the code snippets - otherwise people might not know where to find it.

annatisch (approved):

PR #18466, opened 2021-04-30
[Communication]: Updated test_search_available_phone_numbers_with_invalid_country_code async test to be consistent with sync test

lsundaralingam:


jbeauregardb (approved):
Can we delete the previous recording as well? Other than that, it looks good!

lsundaralingam:
> Can we delete the previous recording as well? Other than that, it looks good!

The previous request didn't have a recording since it failed before sending the request to the service.

PR #18468, opened 2021-04-30
Add aliases for redis list operations for backwards compatibility

TimLovellSmith:
That will compensate for aligning Redis operations in the open API specs with the standard operation name conventions. 

To unblock https://github.com/Azure/azure-rest-api-specs/pull/14023

(Looks like it was pretty well validated by CI.)


iscai-msft (approved):

lmazuel (approved):

msyyc:
I have approved the breaking change of https://github.com/Azure/azure-rest-api-specs/pull/14023#pullrequestreview-652995840. So we needn't the patch anymore.

TimLovellSmith:
@msycc So we don't need the Python SDK to keep backcompat across the versions in this particular case?

msyyc:
@iscai-msft @lmazuel I think both are ok. 
(1)But now that our CHANGELOG has indicated that there is breaking change. So it is better to let it happen. Otherwise customers will be confused and doubt the precision of our CHANGELOG.
(2)The patch may increase our work load of maintenance in the future. We should avoid manual change for mgmt SDK code ASAP.

TimLovellSmith:
@msycc Can we manually edit the CHANGELOG also in this PR? Or its too late because a release is already generated?

msyyc:
@TimLovellSmith CHANGELOG is generated by tools compared with last published package. It is better not to edit it manually since it may mislead customers.

PR #18469, opened 2021-04-30
[communication] Live Testing - Introduce CloudConfig into test.yml 

beltr0n:
This modifies how we interact with the build pipeline to populate env vars

### DIFF
--- a/sdk/communication/test-resources.json
+++ b/sdk/communication/test-resources.json
@@ -13,10 +13,36 @@
             "defaultValue": "communication",
             "type": "string"
         },
+        "communicationServicesEndpointSuffix": {
+            "defaultValue": ".communication.azure.com",
+            "type": "string"
+        },
+          "subscriptionId": {
+            "defaultValue": "[subscription().subscriptionId]",
+            "type": "string"
+          },
         "testApplicationOid": {
             "type": "string",
             "metadata": {
                 "description": "The client OID to grant access to test resources."
+            }
+        },
+        "tenantId": {
+            "type": "string",
+            "metadata": {
+                "description": "The tenant id to which the application and resources belong."
+            }
+        },
+        "testApplicationId": {
+            "type": "string",
+            "metadata": {
+                "description": "The application client id used to run tests."
+            }
+        },
+        "testApplicationSecret": {
+            "type": "string",
+            "metadata": {
+                "description": "The application client secret used to run tests."
             }
         }
     },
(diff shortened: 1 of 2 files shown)


sdk/communication/tests.yml
```diff
@@ -23,11 +23,10 @@ stages:
         ServiceDirectory: communication
         DeployArmTemplate: true
         TestSamples: true
-        EnvVars:
-          AZURE_TEST_RUN_LIVE: 'true'
```
minnieliu, line 27: Can you get rid of the `AZURE_TEST_RUN_LIVE` as well?

beltr0n: I'll add that to the python config since this is python only :)

beltr0n: also need to add AZURE_COMMUNICATION_SERVICE_CONNECTION_STRING

lsundaralingam (approved):
LGTM!

sdk/communication/test-resources.json
```diff
@@ -13,11 +13,37 @@
             "defaultValue": "communication",
             "type": "string"
         },
+        "communicationServicesEndpointSuffix": {
+            "defaultValue": ".communication.azure.com",
+            "type": "string"
+        },
+        "subscriptionId": {
```
sacheu, line 20: I think subscriptoin id is not needed. I have a PR which fixes this in .net.

sdk/communication/test-resources.json
```diff
@@ -46,9 +72,29 @@
         }
     ],
     "outputs": {
+        "AZURE_SUBSCRIPTION_ID": {
```
sacheu, line 75: Do we need this? I think .net work without it.

sdk/communication/test-resources.json
```diff
@@ -46,9 +72,29 @@
...
+        "AZURE_CLIENT_SECRET": {
+            "type": "string",
+            "value": "[parameters('testApplicationSecret')]"
+        },
         "COMMUNICATION_CONNECTION_STRING": {
             "type": "string",
             "value": "[listKeys(resourceId('Microsoft.Communication/CommunicationServices',variables('uniqueSubDomainName')), '2020-08-20-preview').primaryConnectionString]"
-        }
+        },
+        "COMMUNICATION_SERVICE_ENDPOINT": {
```
sacheu, line 95: Do we need this? I think .net work without it.

sdk/communication/test-resources.json
```diff
@@ -13,11 +13,37 @@
             "defaultValue": "communication",
             "type": "string"
         },
+        "communicationServicesEndpointSuffix": {
```
sacheu, line 16: Do we need this? I think .net work without it.

sdk/communication/tests.yml
sacheu, line 27: I think we need this, right?

sacheu (changes requested):

sdk/communication/test-resources.json
beltr0n: Well, Josh removed it last time from JS and it started failing. Let's sync up on your .net fix

sdk/communication/tests.yml
beltr0n: This has been put into the python subconfig

sacheu (approved):
Looks good. Thanks Bert

Issue #18072, opened 2021-04-15
[formrecognizer] update multipage receipt tests file


kristapratico:
service team recommends we recall our current file for multipage receipt tests and use this one instead: https://github.com/Azure-Samples/cognitive-services-REST-api-samples/blob/master/curl/form-recognizer/contoso-allinone.jpg

We can just create a document that has that receipt on two pages.

PR #18473, opened 2021-05-01
[formrecognizer] update multipage receipt sample for tests

kristapratico:
Resolves https://github.com/Azure/azure-sdk-for-python/issues/18072

Preparation for targeting/testing the 2.1 GA endpoints - unskips some tests and updates the multipage receipt sample per service team recommendation.


iscai-msft (approved):

catalinaperalta (approved):
LGTM! Just out of curiosity, why did the service team suggest the change?

kristapratico:
> LGTM! Just out of curiosity, why did the service team suggest the change?

The original sample was created by our team but this was before we knew what a multipage sample for receipts should look like. The old sample reads more like an invoice, whereas the "correct" type of sample should be multiple sales receipt scanned/copied into a PDF. We've seen "regressions" on these tests for almost every new service version because it's not really the right sample for this prebuilt model. So fixing it finally. :)

PR #18474, opened 2021-05-01
[formrecognizer] rename id documents -> identity documents

kristapratico:
Resolves https://github.com/Azure/azure-sdk-for-python/issues/18449


catalinaperalta (approved):
Looks great! A really small nit that could do in the future, I think since we changed everything to identity document, then we could also update the test file names to use the full `identity` word instead of just `id`. Also should we rename the directories holding the example identity documents? (the ones under `samples/sample_forms/` and later on `tests/sample_forms/`)

kristapratico:
> Looks great! A really small nit that could do in the future, I think since we changed everything to identity document, then we could also update the test file names to use the full identity word instead of just id. Also should we rename the directories holding the example identity documents? (the ones under samples/sample_forms/ and later on tests/sample_forms/)

Done!

PR #18478, opened 2021-05-02
Add Testing of Conda Artifacts

scbedd:
`azure-devops` and `conda` ...0/10 would not recommend.

Note that while I'll take feedback on the step `Evaluate OS Specific PATH and Parameters` in `test-conda.yml`, if it looks hacky, there is probably a reason for it. 

@benbp I have extended `archetype-sdk-tests-generate.yml` in this PR because I wanted a nice way to filter out running tests on certain python versions (we only need to support >=3.6 for conda). Rather than doing wonky conditions on every step, I merely extended our usage of `archetype-sdk-tests-generate` so that we invoke it twice. Once for when we do have conda artifacts to test, once for our regular tests..

Reason being, even if we used `parameter sets` functionality in Create-Job-Matrix.ps1, we still run into the issue where we still have to have a `condition` attached to every step within `test-conda` (wherever the steps live). This is due to the fact that the variables are essentially _only available at runtime_, which will means we can't use any compile-time `${{ if` to keep our logs clean.


eng/pipelines/templates/steps/test-conda.yml
```diff
@@ -0,0 +1,143 @@
...
+
+        echo "$(activate.method) index-env"
+        $(activate.method) index-env
+
+        echo "conda index $(Build.ArtifactStagingDirectory)/${{ artifact.name }}"
+        conda index $(Build.ArtifactStagingDirectory)/${{ artifact.name }}
+      displayName: 'Prepare Conda Environment for Testing ${{ artifact.name }}, Index the Target Local Artifact'
+
+    - script: |
+        echo "$(activate.method) ${{ artifact.name }}"
```
weshaggard, line 97: Given this is a script block instead of pwsh it will run on different shells on different OS's and echo is one of the commands that works different on linux vs windows.  There is a little more info on why at https://dev.azure.com/azure-sdk/internal/_wiki/wikis/internal.wiki/2/Azure-DevOps-Pipeline-Guidance?anchor=use-full-cmdlet-names-for-better-scripts.

scbedd: What does it do differently? It's doing what I expect it to AFAICT.

[Example](https://dev.azure.com/azure-sdk/public/_build/results?buildId=875540&view=logs&j=7fdb0983-a0fb-582a-6142-53ff13da6d21&t=2d4deb93-1e3c-5a2c-90ed-c2018537416f)

scbedd: An extra `"` around it isn't the end of the world IMO.

weshaggard: I agree the extra quote isn't the end of the world but that can start to fall apart when you are doing things like writing setvariable or other things that react different depending on the quoting around it. I know it bit me which is why I wrote about it in the wiki.

scbedd: Ok. I'm only using this to output the command that's being invoked for visibility. Thanks for a heads up on the setvariable stuff especially. Gonna look for any instances of that.

scbedd:
Closing to figure out these core/storage errors. Can't keep spamming the build pool.

mitchdenny (approved):

eng/pipelines/templates/stages/platform-matrix-conda-support.json
```diff
@@ -0,0 +1,10 @@
+{
+  "matrix": {
+    "Agent": {
+      "ubuntu-18.04": { "OSVmImage": "MMSUbuntu18.04", "Pool": "azsdk-pool-mms-ubuntu-1804-general" },
```
benbp, line 4: If you change these to something like `"ubuntu-18.04_conda"` then it will be easier to discern which jobs are which in the UI by the display name (which would then be `ubuntu_1804_conda_36`).

scbedd:
Now azure-storage has issues. I don't want to spam the build system and eat more minutes than I already have. Closing again temporarily.

Issue #13747, opened 2020-09-11
[ServiceBus] [(Potentially)Preparers] Address "Resource does not exist" preparer errors.


KieranBrantnerMagee:
Occasionally a Preparer will fail during livetest run because the resource it relies on does not appear to exist, despite that creation method having successfully returned.  See the failure below:

```
                    _logger.error(msg)
>                   raise Exception(msg)
E                   Exception: Preparer failure when creating resource ServiceBusQueueAuthorizationRulePreparer for test test_sb_client_incorrect_queue_conn_str (test_sb_client.ServiceBusClientTests): Operation returned an invalid status code 'Not Found'
E                   Detailed error message: The requested resource servicebustest_qtwofk7xh does not exist. CorrelationId: c8258c67-85f5-40d8-a5eb-e73d69aca731

.tox/whl/lib/python3.5/site-packages/azure_devtools/scenario_tests/preparers.py:82: Exception
------------------------------ Captured log call -------------------------------
ERROR    azure_devtools.scenario_tests.preparers:preparers.py:81 Preparer failure when creating resource ServiceBusQueueAuthorizationRulePreparer for test test_sb_client_incorrect_queue_conn_str (test_sb_client.ServiceBusClientTests): Operation returned an invalid status code 'Not Found'
Detailed error message: The requested resource servicebustest_qtwofk7xh does not exist. CorrelationId: c8258c67-85f5-40d8-a5eb-e73d69aca731
```

This has been observed at all levels, including RG, so may require centralized preparer fix, although lately has most been occurring as seen for AuthRules beneath ServiceBus, so focusing on that repro.

chlowell:
Prior art from `KeyVaultPreparer`, handling name conflict and resource group not found: https://github.com/Azure/azure-sdk-for-python/blob/e25a3539b293731b345e1306e4ec601c69e8df70/tools/azure-sdk-tools/devtools_testutils/keyvault_preparer.py#L102-L115

KieranBrantnerMagee:
Aha.  First; thanks for even noting this and thinking to add your breadcrumb, it's of great comfort to know that this isn't unprecedented.  Second, I'm curious if you have any thoughts on whether there would be a viable fix at the framework level? (If not, no worries, going to dig in myself for this work, just figured I'd ask in case there was any context you had encountered ala "don't waste your time because X" :P)

chlowell:
[...]

PR #18479, opened 2021-05-02
[Service Bus] readdress fix for preparer "resource does not exist" error (potentially)

swathipil:
fixes: #13747

### DIFF
--- a/sdk/servicebus/azure-servicebus/tests/servicebus_preparer.py
+++ b/sdk/servicebus/azure-servicebus/tests/servicebus_preparer.py
@@ -52,15 +52,25 @@ class ServiceBusNamespacePreparer(AzureMgmtPreparer):
         if self.is_live:
             self.client = self.create_mgmt_client(ServiceBusManagementClient)
             group = self._get_resource_group(**kwargs)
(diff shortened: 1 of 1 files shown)


yunhaoling (commented):
do we also need this part for: 
- topics.create_or_update
- queues.create_or_update
- namespaces.create_or_update

swathipil:
> do we also need this part for:
> 
> * topics.create_or_update
> * queues.create_or_update
> * namespaces.create_or_update

I decided not to add it for now, because the original issue only mentioned `SBNamespace` and `SBQueueAuth` as having these errors. If these errors show up for the topics/queues as well, I thought I could add the same fix later. Thoughts @yunhaoling ?

yunhaoling:
me personally prefixing fixing them in one shot as this seems be a general error which will occur when creating resources -- but you're correct as well.

I'll leave the judgement to you :P

yunhaoling (approved):

swathipil:
> me personally prefer fixing them in one shot as this seems be a general error which will occur when creating resources -- but you're correct as well.
> 
> I'll leave the judgement to you :P

makes sense, will do that!

PR #18485, opened 2021-05-03
swallow on 404 for delete operations

seankane-msft:


sdk/tables/azure-data-tables/azure/data/tables/_table_client.py
```diff
@@ -323,6 +325,8 @@ def delete_entity(
                 **kwargs
             )
         except HttpResponseError as error:
+            if error.status_code == 404 and error.reason == u"Not Found":
```
annatisch, line 328: I don't think you need the `error.reason` check - this is just a human readable version of the `status_code`.

annatisch: What we could check here is whether there's an error code in the response:
```python
error = _decode_error(error)
if error.status_code == 404  and error.error_code == 'EntityNotFound':
    # Not sure if this is an actual code value, or even if the service gives us a code at all
    return
_reraise_error(error)
```

sdk/tables/azure-data-tables/samples/async_samples/sample_create_delete_table_async.py
```diff
@@ -95,15 +95,15 @@ async def create_from_table_client(self):
...
+        from azure.core.exceptions import HttpResponseError
 
         # [START delete_from_table_client]
         async with TableClient.from_connection_string(conn_str=self.connection_string, table_name=self.table_name) as table_client:
             try:
                 await table_client.delete_table()
                 print("Deleted table {}!".format(self.table_name))
-            except ResourceNotFoundError:
-                print("Table could not be found")
+            except HttpResponseError:
```
annatisch, line 105: Wondering if we should just remove these exception wrappers from the sample... 
We technically wouldn't have been catching these before.

annatisch (commented):
Loving the test code simplication :)

sdk/tables/azure-data-tables/azure/data/tables/_table_client.py
seankane-msft: There's nothing in the error object, unfortunately the error object does not have an error_code parameter. I'll leave as 
```python
if error.status_code == 404:
    return
```

sdk/tables/azure-data-tables/samples/async_samples/sample_create_delete_table_async.py
seankane-msft: I agree, I removed them.

sdk/tables/azure-data-tables/samples/sample_create_delete_table.py
```diff
@@ -88,15 +85,15 @@ def create_from_table_client(self):
...
+        from azure.core.exceptions import HttpResponseError
 
         # [START delete_table_from_table_client]
         with TableClient.from_connection_string(conn_str=self.connection_string, table_name="myTable") as table_client:
             try:
                 table_client.delete_table()
                 print("Deleted table {}!".format(table_client.table_name))
-            except ResourceNotFoundError:
-                print("Table could not be found")
+            except HttpResponseError as error:
```
annatisch, line 95: This can be removed

sdk/tables/azure-data-tables/samples/async_samples/sample_create_delete_table_async.py
annatisch: This one didn't get removed ;)

sdk/tables/azure-data-tables/azure/data/tables/_table_client.py
```diff
@@ -272,6 +272,8 @@ def delete_table(
         try:
             self._client.table.delete(table=self.table_name, **kwargs)
         except HttpResponseError as error:
+            if error.status_code == 404:
```
annatisch, line 275: Could we add comments to all the docstrings that no error will be raised in the case of the target table/entity not being found?

annatisch (approved):
Thanks!

Issue #16645, opened 2021-02-09
[ServiceBus] Creating resources with mgmt randomly results in ResourceExistsError


swathipil:
- **Package Name**: [azure-servicebus]
- **Package Version**: 7.0.2
- **Operating System**: Windows
- **Python Version**: 3.9

**Describe the bug**
When creating resources by running the `mgmt_*.py` samples, they randomly fail with the following error: `azure.core.exceptions.ResourceExistsError: Operation returned an invalid status 'Conflict'`. This happens both locally and in the test pipeline. For example, if we were running a sample multiple times in rapid succession or in parallel that created a topic named `demo_topic` in the same resource group and SB namespace, it is possible that one sample would try to create a `demo_topic` before the other sample could delete the `demo_topic` that it created.

To ensure this was not happening, for each resource being created, a UUID was added to the name to ensure that multiple resources with the same name were not being created at the same time before they could be deleted. This results in the same error randomly, even when the resource does not already exist.

**To Reproduce**
1. Run any of the samples named `mgmt_*.py` (and add env variables as needed). Like [`mgmt_rule.py`](https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/servicebus/azure-servicebus/samples/sync_samples/mgmt_rule.py)
2. This will randomly result in the error.

**Expected behavior**
I expect the sample to create and delete the resource using the `ServiceBusAdministrationClient` without the ResourceExistsError each time it is run.

**Screenshots**
Sample Error Log:
```
-- List Rules
Rule Name: $Default
Rule Name: sb_mgmt_rule18ba42ea-215e-43b6-a958-bc66d10ecc1e    
Rule Name: sb_mgmt_rule88d8ab79-b463-4337-8d81-264b120c0342    
Rule Name: sb_mgmt_rule8e1c13a5-ac73-411c-b7a3-03160ce13814    
                                                               
-- Create Rule
Creating rule sb_mgmt_rulea8e2fce7-c833-4e27-9de8-b2bbf4a4ab4c.
Traceback (most recent call last):
  File "C:\Users\swathip\Documents\forks\azure-sdk-for-python\sdk\servicebus\azure-servicebus\samples\sync_samples\mgmt_rule.py", line 63, in <module>   
    create_rule(servicebus_mgmt_client)
[...]

PR #18491, opened 2021-05-03
[Service Bus] address mgmt samples ResourceExistsError

swathipil:
fixes: #16645

### DIFF
--- a/sdk/servicebus/azure-servicebus/samples/async_samples/mgmt_rule_async.py
+++ b/sdk/servicebus/azure-servicebus/samples/async_samples/mgmt_rule_async.py
@@ -27,7 +27,7 @@ CONNECTION_STR = os.environ['SERVICE_BUS_CONNECTION_STR']
 TOPIC_NAME = os.environ['SERVICE_BUS_TOPIC_NAME']
 SUBSCRIPTION_NAME = os.environ['SERVICE_BUS_SUBSCRIPTION_NAME']
 RULE_NAME = "sb_mgmt_rule" + str(uuid.uuid4())
-RULE_WITH_SQL_FILTER_NAME = "sb_mgmt_sql_filter_rule" + str(uuid.uuid4())[:16]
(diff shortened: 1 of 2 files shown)


sdk/servicebus/azure-servicebus/samples/async_samples/mgmt_queue_async.py
```diff
@@ -27,7 +28,10 @@
 
 async def create_queue(servicebus_mgmt_client):
     print("-- Create Queue")
-    await servicebus_mgmt_client.create_queue(QUEUE_NAME, max_delivery_count=10, dead_lettering_on_message_expiration=True)
+    try:
+        await servicebus_mgmt_client.create_queue(QUEUE_NAME, max_delivery_count=10, dead_lettering_on_message_expiration=True)
+    except ResourceExistsError:
+        pass
```
yunhaoling, line 34: my gut feeling is that samples should be kept simple and intuitive.

but through this way we're introducing new mental burden to users -- exception handling and `azure.core.exceptions.ResourceExistsError` just for fixing our livetest..

besides, I'm wondering whether the fix would work because based on the description in the issue:
> This results in the same error randomly, even when the resource does not already exist.

the resources not being created yet will also lead to this `ResourceExistsError` error. Will the update/delete operation fail because the entity doesn't exist?

yunhaoling: I'm wondering if it's just the mgmt_rule(_async) causing the random exist failure, because in rule sample we truncate uuid4()[0:16]. (there's limitation on entity name length).


probably we could try:
```
RULE_WITH_SQL_FILTER_NAME = "sql_rule" + str(uuid.uuid4())
```

swathipil: I did notice it was primarily, if not always, the mgmt_rule samples. I'll try that, thanks!

rakshith91 (approved):

PR #18492, opened 2021-05-03
[Container Registry] renamings

seankane-msft:
More re-namings for consistency with other languages.


sdk/containerregistry/azure-containerregistry/azure/containerregistry/_models.py
```diff
@@ -117,6 +117,7 @@ def _from_generated(cls, generated):
             size=generated.size,
             tags=generated.tags,
             content_permissions=generated.writeable_properties,
+            repository_name=kwargs.get("repository_name"),
```
swathipil, line 120: in line 101, you have `kwargs.get("repository_name", None)`. Will the generated object always have `repository_name`, so you don't have to return None otherwise?

swathipil (commented):
qq, but otherwise lgtm

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_models.py
seankane-msft: It should always have `repository_name` because I have to add that additionally to the returned object, however it's good practice to include the `None` default so I will add that.

sdk/containerregistry/azure-containerregistry/azure/containerregistry/aio/_async_container_repository_client.py
```diff
@@ -22,7 +22,7 @@
 from .._models import (
```
swathipil, line 22: weird nit: it seems that the convention for naming async client files is more like `_container_repository_client_async.py` from my experience (even though the user won't see this). Is there a reason you did `_async_container...`?

sdk/containerregistry/azure-containerregistry/azure/containerregistry/aio/_async_container_repository_client.py
```diff
@@ -104,19 +104,20 @@ async def get_properties(self, **kwargs: Dict[str, Any]) -> RepositoryProperties
     @distributed_trace_async
     async def get_registry_artifact_properties(
```
swathipil, line 105: out of curiosity, is there a reason you didn't rename the `registry_artifact` methods to `artifact_manifest`?

sdk/containerregistry/azure-containerregistry/azure/containerregistry/aio/_async_container_repository_client.py
seankane-msft: No reason, you're the first person to point this out too. I'll make a fix for that after this release, I think it will create a bunch of annoying merge conflicts if I do it now.

sdk/containerregistry/azure-containerregistry/azure/containerregistry/aio/_async_container_repository_client.py
seankane-msft: I left this alone because in a separate PR (#18392) I re-shape the entire surface of the API. This method goes away and is put on the a `RegistryArtifact` object and the `ContainerRepositoryClient` also is renamed to `ContainerRepository` with fewer methods that are specific to interactions with just a repository.

swathipil (approved):
lgtm

PR #18493, opened 2021-05-03
[AppConfiguration] Appconfig consistency

seankane-msft:
Test scenarios for https://github.com/Azure/azure-sdk/issues/2801

### DIFF
--- a/sdk/appconfiguration/azure-appconfiguration/azure/appconfiguration/_models.py
+++ b/sdk/appconfiguration/azure-appconfiguration/azure/appconfiguration/_models.py
@@ -61,21 +61,21 @@ class ConfigurationSetting(Model):
         if key_value is None:
             return None
         if key_value.content_type is not None:
-            if key_value.content_type.startswith(
-                FeatureFlagConfigurationSetting._feature_flag_content_type  # pylint:disable=protected-access
-            ) and key_value.key.startswith(FeatureFlagConfigurationSetting.key_prefix):
-                try:
+            try:
+                if key_value.content_type.startswith(
+                    FeatureFlagConfigurationSetting._feature_flag_content_type  # pylint:disable=protected-access
+                ) and key_value.key.startswith(FeatureFlagConfigurationSetting.key_prefix):
                     return FeatureFlagConfigurationSetting._from_generated(  # pylint: disable=protected-access
                         key_value
                     )
-                except (KeyError, AttributeError, TypeError):
-                    pass
-            if key_value.content_type.startswith(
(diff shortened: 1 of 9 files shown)


sdk/appconfiguration/azure-appconfiguration/azure/appconfiguration/_models.py
```diff
@@ -334,26 +331,23 @@ def _validate(self):
...
-            )
-        except (KeyError, AttributeError):
-            return ConfigurationSetting._from_generated(key_value)
+        if key_value is None:
+            return None
+        if key_value.value:
+            try:
+                key_value.value = json.loads(key_value.value)
+            except json.decoder.JSONDecodeError:
+                pass
```
xiangyan99, line 340: Why pass not raise ValueError?

sdk/appconfiguration/azure-appconfiguration/azure/appconfiguration/_models.py
```diff
@@ -143,7 +147,7 @@ class FeatureFlagConfigurationSetting(
     kind = "FeatureFlag"
 
     def __init__(self, feature_id, enabled, filters=[], **kwargs):  # pylint: disable=dangerous-default-value
-        # type: (str, bool, Optional[List[Dict[str, Any]]]) -> None
+        # type: (str, bool, Optional[List[Dict[str, Any]]], Any) -> None
```
xiangyan99, line 150: **Any

sdk/appconfiguration/azure-appconfiguration/azure/appconfiguration/_models.py
seankane-msft: When we try to do the indexing it will raise a `TypeError` that is caught, and a `ConfigurationSetting` is returned instead.

hinoryu (approved):

sdk/appconfiguration/azure-appconfiguration/azure/appconfiguration/_models.py
xiangyan99: My concern is if we pass here, will it get mad in key=key_value.key?

seankane-msft: This is called by the `ConfigurationSetting._from_generated` method so if it does raise an error on `key=key_value.key`, the ConfigurationSetting will catch the error and return a plain `ConfigurationSetting` object.

xiangyan99 (approved):

PR #18514, opened 2021-05-05
[Test Infra] adding operation-location to headers that are scrubbed

seankane-msft:
Found when working with @iscai-msft that some services provide the next URL in the `operation-location` header, which can leak accounts and specific job details. Adding this to the headers that are scrubbed the same way `location` is.


iscai-msft (approved):
thanks @seankane-msft !

PR #18521, opened 2021-05-05
Fix sample by truncating iot hub name to 25 chars

mtin:
I had trouble running the event hub sample for retrieving messages from the built-in endpoint of an IoT hub. An earlier version of the sdk had integrated support to create a receiver from an iot hub connection string (`EventHubClient.from_iothub_connection_string()`) but it seems like that has been removed?

The sample `sdk/eventhub/azure-eventhub/samples/async_samples/iot_hub_connection_string_receive_async.py` has methods to generate an event hub connection string from an iot hub connection string - but apparently there is a bug.

The `EntityPath` part of the connection string has to be truncated to 25 characters, otherwise the sample fails with the following messages:

```shell
Authentication Put-Token failed. Retries exhausted.
Authentication Put-Token failed. Retries exhausted.
Authentication Put-Token failed. Retries exhausted.
Authentication Put-Token failed. Retries exhausted.
EventProcessor instance '<snip>' of eventhub '<snip>-test-iothub' consumer group '$Default'. An error occurred while load-balancing and claiming ownership. The exception is AuthenticationError("The messaging entity 'sb://iothub-ns-<snip>.servicebus.windows.net/<snip>-test-iothub' could not be found. To know more visit https://aka.ms/sbResourceMgrExceptions.  TrackingId:<snip>, Timestamp:2021-05-05T22:15:29
CBS Token authentication failed.
Status code: 404
Description: The messaging entity 'sb://iothub-ns-<snip>.servicebus.windows.net/<snip>-test-iothub' could not be found. To know more visit https://aka.ms/sbResourceMgrExceptions.  TrackingId:<snip>, Timestamp:2021-05-05T22:15:29 occurred during the load balance process.
```

I validated the 25 char limit by manually retrieving the event hub compatible connection string from the iot hub page of the azure portal

### DIFF
--- a/sdk/eventhub/azure-eventhub/samples/async_samples/iot_hub_connection_string_receive_async.py
+++ b/sdk/eventhub/azure-eventhub/samples/async_samples/iot_hub_connection_string_receive_async.py
@@ -84,7 +84,7 @@ def convert_iothub_to_eventhub_conn_str(iothub_conn_str):
             fully_qualified_name,
             shared_access_key_name,
             shared_access_key,
-            iot_hub_name
+            iot_hub_name[:25]
         )
     except Exception as exp:
         raise ValueError(


swathipil:
@rakshith91 I believe this sample is in the list of ignored samples, so it should automatically pass the tests.

swathipil:
Hey @mtin, just wanted to give you an update! We're working on testing this sample and testing your fix right now. Thanks for your contribution!

swathipil:
Hey @mtin! After discussing this with the team, we decided not to merge this PR for the following reasons:

- The bug that you found was a result of the IoT Hub Service silently truncating the IoT Hub name to 25 characters without any clear documentation. This silent change of the information passed in by the user should not be happening. Since this is a service issue, we have contacted the IoT team to let them know about the bug, so that they can either fix it or document it.
- By updating the sample with your commit, we would be silently mutating the information sent in by a user. This is not something we would ever want to do, in any case.

However, we really appreciate you calling attention to this! We're hoping to have a response from the service team within the next month and have created an issue to track this.

Closing this PR. Thanks again!

Issue #18422, opened 2021-04-29
App configuration samples issues


v-xuto:
1.
Section [link1](https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/appconfiguration/azure-appconfiguration/samples/sync_token_samples.py#L26) and [link2](https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/appconfiguration/azure-appconfiguration/samples/sync_token_async_samples.py#L27):
![image](https://user-images.githubusercontent.com/57166602/116521204-3dffbc00-a906-11eb-9669-4065520b5616.png)

Reason:
with AzureAppConfigurationClient.from_connection_string(CONNECTION_STRING) as client:
AttributeError: __enter__

Suggestion:
Update to: 
`client = AzureAppConfigurationClient.from_connection_string(CONNECTION_STRING)`.

2.
Section [link](https://github.com/Azure/azure-sdk-for-python/tree/master/sdk/appconfiguration/azure-appconfiguration/samples):
![image](https://user-images.githubusercontent.com/57166602/116521271-5a035d80-a906-11eb-81eb-d25ce1db26a8.png)

Suggestion:
Should be consistent with the file name of the async samples of other services, for example: key vault, storage and so on. 
Update to:
```
conditional_operation_sample_async.py
hello_world_advanced_sample_async.py
hello_world_sample_async.py
list_revision_sample_async.py
read_only_sample_async.py
sync_token_samples_async.py
```

@jongio for notification.

PR #18542, opened 2021-05-06
[AppConfig] Fixing samples

seankane-msft:


seankane-msft:
#18422

sdk/appconfiguration/azure-appconfiguration/samples/sync_token_samples_async.py
```diff
@@ -24,16 +24,18 @@ async def handle_event_grid_notifications(event_grid_events):
 
     all_keys = []
 
-    async with AzureAppConfigurationClient.from_connection_string(CONNECTION_STRING) as client:
```
xiangyan99, line 27: This will cause connection unclose.

We recommend users to use context manager to close client.

seankane-msft: There is no `__enter__` (or `__aenter__` in async) on the clients. Adding this now.

xiangyan99 (approved):

PR #18547, opened 2021-05-06
Remove unused parameters in docs-metadata-release template

benbp:
These parameters are unused, and it's causing some template validation errors against https://github.com/Azure/azure-sdk-tools/pull/1585/, which uses a more strict parameter checking syntax.


eng/pipelines/templates/stages/archetype-python-release.yml
```diff
@@ -179,13 +179,11 @@ stages:
                           ReleaseSha: $(Build.SourceVersion)
                           RepoId: Azure/azure-sdk-for-python
                           WorkingDirectory: $(System.DefaultWorkingDirectory)
```
weshaggard, line 181: FYI @danieljurek this is just a small part of the clean-up we need to do for the docs processing once you refactor it for the daily packages as well.

weshaggard (approved):

Issue #18482, opened 2021-05-03
[formrecognizer] regenerate on swagger for field value type changes


kristapratico:
Affecting gender and country

Issue #17062, opened 2021-03-03
[formrecognizer] Remove 3.5 support


kristapratico:
- [x] We should GA version 3.1.0 with 3.5 support but include a message that it will not be supported moving forward:

> This version will be the last version to officially support Python 3.5, future versions will require Python 2.7 or Python 3.6+

- [ ] Any version after 3.1.0 (GA or preview) will not support 3.5.

> Please keep in mind that dropping support for a python version should indeed be a minor bump (not just patch).

kristapratico:
Issue moved to [Azure/azure-sdk-for-python-pr #768](https://github.com/Azure/azure-sdk-for-python-pr/issues/768) via [**ZenHub**](https://www.zenhub.com/)

Issue #15276, opened 2020-11-12
[formrecognizer] use value_selection_mark once service returns this value


kristapratico:
Known service bug causing `value_selection_mark` to be None on a FieldValue of type == "selectionMark". We are working around this by returning the `text` value in the SDK: https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/formrecognizer/azure-ai-formrecognizer/azure/ai/formrecognizer/_models.py#L67

ETA v2.1 GA for fix

kristapratico:
Updated to June milestone, fix will be in GA release

PR #18551, opened 2021-05-06
[formrecognizer] regenerates on v2.1

### LINKED ISSUE #18133: [formrecognizer] generate on new swagger
https://github.com/Azure/azure-rest-api-specs/pull/13997

kristapratico:
Resolves #18133 https://github.com/Azure/azure-sdk-for-python/issues/17062 https://github.com/Azure/azure-sdk-for-python/issues/15276 https://github.com/Azure/azure-sdk-for-python/issues/18482

Regenerates on v2.1 and sets as default. Note that AAD is not working yet.


sdk/formrecognizer/azure-ai-formrecognizer/azure/ai/formrecognizer/_generated/v2_0/aio/operations/_form_recognizer_client_operations.py
```diff
@@ -710,15 +710,15 @@ async def begin_analyze_receipt_async(
...
+        :keyword polling: By default, your polling method will be AsyncLROBasePolling.
+         Pass in False for this operation to not poll, or pass in your own initialized polling object for a personal polling strategy.
         :paramtype polling: bool or ~azure.core.polling.AsyncPollingMethod
         :keyword int polling_interval: Default waiting time between two polls for LRO operations if no Retry-After header is present.
         :return: An instance of AsyncLROPoller that returns either None or the result of cls(response)
         :rtype: ~azure.core.polling.AsyncLROPoller[None]
         :raises ~azure.core.exceptions.HttpResponseError:
         """
-        polling = kwargs.pop('polling', False)  # type: Union[bool, AsyncPollingMethod]
+        polling = kwargs.pop('polling', True)  # type: Union[bool, AsyncPollingMethod]
```
catalinaperalta, line 721: This is an interesting behavior change, I know it's coming from generated code, but just wondering if this will impact customers who were counting on the previous behavior? Mainly this is just curiosity on my end

sdk/formrecognizer/azure-ai-formrecognizer/tests/test_content_from_url_async.py
```diff
@@ -98,7 +98,7 @@ async def test_content_url_pdf(self, client):
         self.assertEqual(layout.page_number, 1)
         self.assertFormPagesHasValues(result)
         self.assertEqual(layout.tables[0].row_count, 3)
-        self.assertEqual(layout.tables[0].column_count, 6)
+        self.assertEqual(layout.tables[0].column_count, 5)
```
catalinaperalta, line 101: Why did this change?

sdk/formrecognizer/azure-ai-formrecognizer/samples/async_samples/sample_recognize_identity_documents_async.py
```diff
@@ -72,9 +72,9 @@ async def recognize_identity_documents(self):
                 address = id_document.fields.get("Address")
                 if address:
                     print("Address: {} has confidence: {}".format(address.value, address.confidence))
-                country = id_document.fields.get("Country")
+                country = id_document.fields.get("CountryRegion")
                 if country:
-                    print("Country: {} has confidence: {}".format(country.value, country.confidence))
+                    print("Country Region: {} has confidence: {}".format(country.value, country.confidence))
```
catalinaperalta, line 77: Do we need to add a slash here? Or join the words like the field name?

sdk/formrecognizer/azure-ai-formrecognizer/tests/test_invoice.py
```diff
@@ -320,7 +320,7 @@ def test_invoice_pdf_include_field_elements(self, client):
         self.assertEqual(invoice.fields.get("CustomerName").value, "Microsoft")
         self.assertEqual(invoice.fields.get("InvoiceId").value, '34278587')
         self.assertEqual(invoice.fields.get("InvoiceDate").value, date(2017, 6, 18))
-        self.assertEqual(invoice.fields.get("InvoiceTotal").value, 56651.49)
+        # self.assertEqual(invoice.fields.get("InvoiceTotal").value, 56651.49)  FIXME: not finding InvoiceTotal
```
catalinaperalta, line 323: Sad about this regression. :(

catalinaperalta: Also curious, are we using an invoice the service team recommended for this test?

sdk/formrecognizer/azure-ai-formrecognizer/azure/ai/formrecognizer/_generated/v2_0/aio/operations/_form_recognizer_client_operations.py
kristapratico: we always either passed True or the polling method ourselves so no breaking behavior here.

sdk/formrecognizer/azure-ai-formrecognizer/tests/test_content_from_url_async.py
kristapratico: Improvements in the layout model. It was finding the wrong amount of columns previously.

sdk/formrecognizer/azure-ai-formrecognizer/tests/test_invoice.py
kristapratico: I think this test is using a different invoice. Might be good to have one that does use the recommended one. https://github.com/Azure/azure-sdk-for-python/issues/18642

sdk/formrecognizer/azure-ai-formrecognizer/samples/async_samples/sample_recognize_identity_documents_async.py
catalinaperalta: I was wondering if `Country Region` should be `Country/Region` or `CountryRegion`? Simply because I skimmed through the docs shared from the service team and I know it is sensitive to how this information is shown. 

Ok just checked the doc...I think we should go with `Country/Region` to make sure we're displaying the information correctly (this is the example I see the guidance following the most)

catalinaperalta: Honestly this feels like a nit to me 😅 but just to be said since I'm not sure if the space implies something different than the slash...but I may be analyzing this too much.

catalinaperalta (approved):
Only left my country region nit, but aside from that this looks great. Thanks for doing this!!

PR #18554, opened 2021-05-06
Lower retry backoff factor for IMDS managed identity

chlowell:
IMDS has [particular retry guidance](https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/how-to-use-vm-token#retry-guidance) which ImdsCredential attempts to follow. However, its current backoff factor is too high. The guidance suggests retrying after approximately 2, 6, 14, 30 seconds but the credential waits 8, 16, 32, 60 seconds. The curve of our retry policy's backoff function doesn't match what IMDS has in mind, so we can't follow the guidance exactly, but we can do better. In particular, we shouldn't wait 8 seconds before the first retry. With this PR, the credential waits 2.2, 4.4, 8.8, 17.6 seconds.

### DIFF
--- a/sdk/identity/azure-identity/azure/identity/_credentials/imds.py
+++ b/sdk/identity/azure-identity/azure/identity/_credentials/imds.py
@@ -25,7 +25,7 @@ IMDS_URL = "http://169.254.169.254/metadata/identity/oauth2/token"
 
 PIPELINE_SETTINGS = {
     "connection_timeout": 2,
-    "retry_backoff_factor": 1.1,
+    "retry_backoff_factor": 2,
     "retry_backoff_max": 60,
     "retry_on_status_codes": [404, 429] + list(range(500, 600)),
     "retry_status": 5,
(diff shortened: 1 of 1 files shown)


mccoyp (commented):
This does seem better, but I wonder if this backoff curve is too flat. It's probably fine, but I'm curious about whether the priority here is on matching the curve or matching the delay of early retries (0.2 second difference compared to recommended for the first request, 1.6 s for the second, etc.).

If I understand this correctly (I might not), I think a `retry_backoff_factor` of 2 would retry at 4, 8, 16, and 32 seconds -- which is very close to the recommended curve. Is the issue there that the first retry has twice as long of a delay?

chlowell:
The backoff curve is flatter than recommended for IMDS and not configurable (making it configurable seems excessive to me, as does implementing a custom retry policy). My goal here is to reach a better compromise with the curve we have. I prioritized aligning the early delays believing that this will cause less wasted time on average, i.e. I gambled on more intermittent problems resolve earlier than later. You persuaded me though; with a backoff factor of 2 we spread the retries over the period of time the platform recommends.

mccoyp (approved):

PR #18576, opened 2021-05-07
[Key Vault] Add API version 7.2 for secrets

mccoyp:
This regenerates on commit [1e2c9f](https://github.com/Azure/azure-rest-api-specs/commit/1e2c9f3ec93078da8078389941531359e274f32a).


sdk/keyvault/azure-keyvault-secrets/azure/keyvault/secrets/_shared/client_base.py
```diff
@@ -71,7 +72,7 @@ def __init__(self, vault_url, credential, **kwargs):
                 **kwargs
             )
             self._models = _KeyVaultClient.models(api_version=self.api_version)
-        except NotImplementedError:
+        except ValueError:
             raise NotImplementedError(
```
mccoyp, line 76: Question here: the generated code has switched from raising a NotImplementedError to a ValueError if an unsupported API version is provided. I think either error is valid in this context, but I don't know if it's best to switch the error raised by the client for consistency, or to keep the NotImplementedError to avoid a (very small) breaking change. I marginally prefer the latter approach

chlowell: I agree. Either error seems reasonable to me, and consistency with prior versions of this library is more important to me than consistency with the generated code we wrap.

PR #18578, opened 2021-05-07
use non decompression in stream only

xiangyan99:


sdk/core/azure-core/azure/core/pipeline/transport/_aiohttp.py
```diff
@@ -89,10 +90,16 @@ async def open(self):
                 loop=self._loop,
                 trust_env=self._use_env_settings,
                 cookie_jar=jar,
+            )
+            self._non_decompress_session = aiohttp.ClientSession(
```
annatisch, line 94: We have to maintain separate sessions? Why did the tests not catch this earlier?

xiangyan99: We found the regression in live translation tests which server supports "transfer-encoding" and "content-encoding".

I checked the recorded tests. All recordings are decompressed strings.

PR #18581, opened 2021-05-07
Core decompress body

xiangyan99:


sdk/core/azure-core/azure/core/pipeline/transport/_aiohttp.py
```diff
@@ -274,10 +305,39 @@ def text(self, encoding: Optional[str] = None) -> str:
...
+        if not encoding:
+            if mimetype.type == "application" and (
+                    mimetype.subtype == "json" or mimetype.subtype == "rdap"
+            ):
+                # RFC 7159 states that the default encoding is UTF-8.
+                # RFC 7483 defines application/rdap+json
+                encoding = "utf-8"
+            elif body is None:
+                raise RuntimeError(
+                    "Cannot guess the encoding of " "a not yet read body"
```
annatisch, line 330: Not sure why this string is in two halves?

sdk/core/azure-core/azure/core/pipeline/transport/_aiohttp.py
```diff
@@ -274,10 +305,39 @@ def text(self, encoding: Optional[str] = None) -> str:
...
+                )
+            else:
+                encoding = chardet.detect(body)["encoding"]
```
annatisch, line 333: Should this be `chardet.detect(body).get("encoding")`?

sdk/core/azure-core/azure/core/pipeline/transport/_aiohttp.py
```diff
@@ -274,10 +305,39 @@ def text(self, encoding: Optional[str] = None) -> str:
...
         if not encoding:
```
annatisch, line 334: Based on the code above, I don't think `encoding` can still be falsey at this point?
Unless we want to add this default to the if-statement above in the case that `chardet` failed to detect?

sdk/core/azure-core/azure/core/pipeline/transport/_aiohttp.py
```diff
@@ -274,10 +305,39 @@ def text(self, encoding: Optional[str] = None) -> str:
...
-            encoding = self.internal_response.get_encoding()
+            encoding = "utf-8"
+
+        if encoding == "utf-8" or encoding is None:
```
annatisch, line 337: `encoding` wont be `None` any more

sdk/core/azure-core/tests/async_tests/test_universal_http_async.py
```diff
@@ -123,6 +123,28 @@ async def test_aiohttp_response_text():
...
+            b"\x1f\x8b\x08\x00\x00\x00\x00\x00\x04\x00\x8d\x8d\xb1n\xc30\x0cD"
+            b"\xff\x85s\x14HVlY\xda\x8av.\n4\x1d\x9a\x8d\xa1\xe5D\x80m\x01\x12="
+            b"\x14A\xfe\xbd\x92\x81d\xceB\x1c\xef\xf8\x8e7\x08\x038\xf0\xa67Fj+"
+            b"\x946\x9d8\x0c4\x08{\x96(\x94mzkh\x1cM/a\x07\x94<\xb2\x1f>\xca8\x86"
+            b"\xd9\xff0\x15\xb6\x91\x8d\x12\xb2\x15\xd2\x1c\x95q\xbau\xba\xdbk"
+            b"\xd5(\xd9\xb5\xa7\xc2L\x98\xf9\x8d8\xc4\xe5U\xccV,3\xf2\x9a\xcb\xddg"
+            b"\xe4o\xc6T\xdeVw\x9dgL\x7f\xe0n\xc0\x91q\x02'w0b\x98JZe^\x89|\xce\x9b"
+            b"\x0e\xcbW\x8a\x97\xf4X\x97\xc8\xbf\xfeYU\x1d\xc2\x85\xfc\xf4@\xb7\xbe"
+            b"\xf7+&$\xf6\xa9\x8a\xcb\x96\xdc\xef\xff\xaa\xa1\x1c\xf9$\x01\x00\x00",
+            {'Content-Type': 'text/plain', 'Content-Encoding':"gzip"}
```
annatisch, line 138: Can we add some negative tests for unsupported content-encoding?

sdk/core/azure-core/tests/async_tests/test_universal_http_async.py
```diff
@@ -123,6 +123,28 @@ async def test_aiohttp_response_text():
...
+        )
+        body = res.body()
+        expect = b'{"id":"e7877039-1376-4dcd-9b0a-192897cff780","createdDateTimeUtc":' \
+                 b'"2021-05-07T17:35:36.3121065Z","lastActionDateTimeUtc":' \
+                 b'"2021-05-07T17:35:36.3121069Z","status":"NotStarted",' \
+                 b'"summary":{"total":0,"failed":0,"success":0,"inProgress":0,' \
+                 b'"notYetStarted":0,"cancelled":0,"totalCharacterCharged":0}}'
+        assert res.body() == expect, "Decompression didn't work"
```
annatisch, line 146: Can we also test `res.text()` both positive and negative?

sdk/core/azure-core/azure/core/pipeline/transport/_aiohttp.py
xiangyan99: From the code chardet/universaldetector.py, it seems to me it is possible encoding is None.

sdk/core/azure-core/azure/core/pipeline/transport/_aiohttp.py
xiangyan99: This is the code from aiohttp. 

We want to keep it same because we want to have exactly same behavior of aiohttp. (I don't want users to notice we re-implement these logic in azure-core)

sdk/core/azure-core/azure/core/pipeline/transport/_aiohttp.py
```diff
@@ -145,6 +150,11 @@ async def send(self, request: HttpRequest, **config: Any) -> Optional[AsyncHttpR
         :keyword str proxy: will define the proxy to use all the time
         """
         await self.open()
+        try:
+            auto_decompress = self.session.auto_decompress
```
johanste, line 154: `self.session` is supposed to be an `aiohttp.ClientSession`, correct? If so, this will never raise an `AttributeError`.

johanste: Btw, not sure why we are missing type annotations for `self.session`. I would expect us to have that!

xiangyan99: This is to handle custom session. I suppose it is possible for a custom session that does not have attribute auto_decompress and I will regard custom session as auto_decompress enabled.

sdk/core/azure-core/azure/core/pipeline/transport/_aiohttp.py
```diff
@@ -250,21 +258,40 @@ class AioHttpTransportResponse(AsyncHttpResponse):
...
+                 aiohttp_response: aiohttp.ClientResponse,
+                 block_size=None, **kwargs) -> None:
         super(AioHttpTransportResponse, self).__init__(request, aiohttp_response, block_size=block_size)
         # https://aiohttp.readthedocs.io/en/stable/client_reference.html#aiohttp.ClientResponse
         self.status_code = aiohttp_response.status
         self.headers = CIMultiDict(aiohttp_response.headers)
         self.reason = aiohttp_response.reason
         self.content_type = aiohttp_response.headers.get('content-type')
         self._body = None
+        self._decompress = kwargs.pop("decompress", True)
```
johanste, line 274: You should be able to use an actual keyword-only argument for decompress.

sdk/core/azure-core/azure/core/pipeline/transport/_aiohttp.py
johanste: `chardet.detect` always returns an `"encoding": something` key/value - even though it may have a very low confidence in its detection, it always makes a best effort.

sdk/core/azure-core/tests/async_tests/test_streaming_async.py
```diff
@@ -0,0 +1,176 @@
...
+    pipeline_response = await client._pipeline.run(request, stream=True)
+    response = pipeline_response.http_response
+    data = response.stream_download(client._pipeline, decompress=True)
+    content = b""
+    async for d in data:
+        content += d
+    try:
+        decoded = content.decode('utf-8')
+        assert False
+    except UnicodeDecodeError:
```
annatisch, line 81: So this raises because we couldn't decompress because no header way found - is that right?

sdk/core/azure-core/tests/async_tests/test_streaming_async.py
```diff
@@ -0,0 +1,176 @@
...
+    # expect error
+    import zlib
+    account_name = "coretests"
+    account_url = "https://{}.blob.core.windows.net".format(account_name)
+    url = "https://{}.blob.core.windows.net/tests/test_with_header.txt".format(account_name)
+    client = AsyncPipelineClient(account_url)
+    request = client.get(url)
+    pipeline_response = await client._pipeline.run(request, stream=True)
+    response = pipeline_response.http_response
+    data = response.stream_download(client._pipeline, decompress=True)
```
annatisch, line 115: What's the header that being returned here?
Is it raising because the header says it's gzip, but the content itself doesn't match?

sdk/core/azure-core/tests/async_tests/test_streaming_async.py
xiangyan99: Because there is no encoding header, we will not try to decompress it. 

Here raises error because it fails to decode a compressed stream.

sdk/core/azure-core/tests/async_tests/test_streaming_async.py
xiangyan99: Right. 

In this scenario, there is an encoding header and we pass in decompress=True. We will try to decompress the stream which is not in correct format. So the decompression will fail.

annatisch: Is this failing because the decompression algorithm mismatches the header? Or because the content itself mismatches the header?
I'm wondering because it's a zlib error, but the test name indicates a 'plain' content header.

xiangyan99: With encoding header "gzip" we will try to use "gzip" algorithm to decompress the stream. 

But the content of the stream itself is not compressed (it is a plain text).

What happens here is we try to decompress an un-compressed stream so we fail to decompress.

annatisch: Gotcha - perfect!

annatisch (approved):

xiangyan99:
python - autorest - pr failure is known.

sdk/core/azure-core/azure/core/pipeline/transport/_aiohttp.py
```diff
@@ -145,6 +150,10 @@ async def send(self, request: HttpRequest, **config: Any) -> Optional[AsyncHttpR
         :keyword str proxy: will define the proxy to use all the time
         """
         await self.open()
+        try:
+            auto_decompress = self.session.auto_decompress  # type: ignore
+        except AttributeError:
+            auto_decompress = True
```
johanste, line 156: I would add a comment as to why we need this. I know I would be confused unless I knew the history...

xiangyan99: Added.

sdk/core/azure-core/azure/core/pipeline/transport/_aiohttp.py
```diff
@@ -250,21 +259,38 @@ class AioHttpTransportResponse(AsyncHttpResponse):
...
+        enc = self.headers.get('Content-Encoding')
+        if not enc:
+            return self._body
+        enc = enc.lower()
+        if enc in ("gzip", "deflate"):
+            import zlib
+            zlib_mode = 16 + zlib.MAX_WBITS if enc == "gzip" else zlib.MAX_WBITS
+            decompressor = zlib.decompressobj(wbits=zlib_mode)
+            body = decompressor.decompress(self._body)
+            return body
```
johanste, line 294: I do have some concerns about us not caching the decompressed body. Because we only need it once, right? Do we have any other access to `self._body` that requires us to keep the compressed data?

xiangyan99: I don't expect (as least I did not see) users need to get body twice.

If you want, we can update the code like:

            if self._decompressed_body:
                     return self._decompressed_body
            import zlib
            self._decompressed_body = decompressor.decompress(self._body)
            return self._decompressed_body
        return self._body

But to be honest, I don't see lots of value for this.

johanste: The don't need to get the body more than once. And it would not be clear to me that getting the body and then the text will decompress the body twice.

I don't think we need to keep the compressed data around once it has been decompressed, right?

xiangyan99: Sounds fair. Updated. :)

sdk/core/azure-core/azure/core/pipeline/transport/_aiohttp.py
```diff
@@ -250,21 +260,41 @@ class AioHttpTransportResponse(AsyncHttpResponse):
...
+            return self._body
+        enc = enc.lower()
+        if enc in ("gzip", "deflate"):
+            if self._decompressed_body:
+                return self._decompressed_body
+            import zlib
+            zlib_mode = 16 + zlib.MAX_WBITS if enc == "gzip" else zlib.MAX_WBITS
+            decompressor = zlib.decompressobj(wbits=zlib_mode)
+            self._decompressed_body = decompressor.decompress(self._body)
+            return self._decompressed_body
```
johanste, line 297: I would suggest keeping a single copy of the body around. Unless you still need the compressed version for some reason...

PR #18598, opened 2021-05-10
T2 cosmosdb 2021 04 previw

RAY-316:
https://github.com/Azure/sdk-release-request/issues/1489


msyyc (commented):
edit '_meta.json' with swagger repo commit id

Issue #17928, opened 2021-04-08
[translation] opt-in to mypy checks


kristapratico:
https://github.com/Azure/azure-sdk-for-python/blob/master/eng/tox/mypy_hard_failure_packages.py#L16

PR #18599, opened 2021-05-10
mypy linting

mohamedshabanofficial:
solves #17928

### DIFF
--- a/sdk/translation/azure-ai-translation-document/azure/ai/translation/document/_polling.py
+++ b/sdk/translation/azure-ai-translation-document/azure/ai/translation/document/_polling.py
@@ -4,8 +4,10 @@
 # Licensed under the MIT License.
 # ------------------------------------
 
+from typing import Optional
 from azure.core.polling.base_polling import (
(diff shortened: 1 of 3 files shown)


sdk/translation/azure-ai-translation-document/azure/ai/translation/document/_polling.py
```diff
@@ -4,8 +4,10 @@
 # Licensed under the MIT License.
 # ------------------------------------
 
+from typing import Optional
 from azure.core.polling.base_polling import (
     LongRunningOperation,
+    PipelineResponseType,
```
kristapratico, line 10: This can't be imported here because it's defined in a type checking block. You'll need to redefine it in this file.

kristapratico (commented):
Please also add `azure-ai-translation-document` to this [list](https://github.com/Azure/azure-sdk-for-python/blob/master/eng/tox/mypy_hard_failure_packages.py#L16) so we can opt-in to mypy checks in the CI pipeline.

PR #18612, opened 2021-05-10
Changed http to https for gutenberg sample

tasherif-msft:


sdk/storage/azure-storage-blob/tests/test_common_blob.py
```diff
@@ -324,7 +324,7 @@ def test_create_blob_with_requests(self, resource_group, location, storage_accou
         self._setup(storage_account, storage_account_key)
 
         # Act
-        uri = "http://www.gutenberg.org/files/59466/59466-0.txt"
+        uri = "https://www.gutenberg.org/files/59466/59466-0.txt"
```
xiafu-msft, line 327: we don't have to change it from http to https, since  requests.get supports website redirect

xiafu-msft (approved):

PR #18656, opened 2021-05-12
[Communication]: Cleaned up ACS samples

lsundaralingam:


jbeauregardb (approved):

beltr0n (approved):

beltr0n (changes requested):
Since we're in clean up, I don't think ACS account is the right term. It's part of the ACS resource in your Azure account. maybe change it to ACS resource?

beltr0n (approved):

PR #18690, opened 2021-05-13
Deprecated new_name in Undelete

tasherif-msft:
`new_name` is no longer supported on the service side for `undelete_container()` and `undelete_filesystem()`


amishra-dev (dismissed):
Can we update the API comments to indicate that the parameter does not work if the value is different than the original container?

tasherif-msft:
> Can we update the API comments to indicate that the parameter does not work if the value is different than the original container?

So we want to keep the parameter even though the only applicable value is if it matches the old one?
In this PR I just removed the parameter altogether and surfaced a deprecation warning if the user passes it - then we let service throw.

amishra-dev:
> > Can we update the API comments to indicate that the parameter does not work if the value is different than the original container?
> 
> So we want to keep the parameter even though the only applicable value is if it matches the old one?
> In this PR I just removed the parameter altogether and surfaced a deprecation warning if the user passes it - then we let service throw.

isnt' that a breaking change?

tasherif-msft:
> > > Can we update the API comments to indicate that the parameter does not work if the value is different than the original container?
> > 
> > 
> > So we want to keep the parameter even though the only applicable value is if it matches the old one?
> > In this PR I just removed the parameter altogether and surfaced a deprecation warning if the user passes it - then we let service throw.
> 
> isnt' that a breaking change?

Well technically in Python they can still pass it if they want to and currently we ARE parsing it like we did before - its just now we surface a warning (if the cx passes it) and let service throw accordingly. However since this kwarg is no longer supported thought it would be best to get removed from the docstrings as an accepted kwarg. Do you feel like it should be kept in the docstrings as an accepted kwarg and add a deprecation comment besides it?

sdk/storage/azure-storage-blob/azure/storage/blob/aio/_blob_service_client_async.py
```diff
@@ -568,14 +569,14 @@ async def undelete_container(self, deleted_container_name, deleted_container_ver
...
-        :keyword str new_name:
-            The new name for the deleted container to be restored to.
-            If not specified deleted_container_name will be used as the restored container name.
         :keyword int timeout:
             The timeout parameter is expressed in seconds.
         :rtype: ~azure.storage.blob.aio.ContainerClient
         """
         new_name = kwargs.pop('new_name', None)
+        if new_name:
+            warnings.simplefilter('always', DeprecationWarning)
```
xiafu-msft, line 578: Hi Anna! I think we should let the user to choose how they want to deal with the warning, what do you think?  @annatisch

xiafu-msft (approved):

sdk/storage/azure-storage-file-datalake/tests/test_file_system_async.py
```diff
@@ -235,52 +236,18 @@ async def test_undelete_file_system(self, datalake_storage_account_name, datalak
...
-            filesystem_list.append(fs)
-        self.assertTrue(len(filesystem_list) >= 1)
-
-        for filesystem in filesystem_list:
-            # find the deleted filesystem and restore it
-            if filesystem.deleted and filesystem.name == filesystem_client.file_system_name:
-                with self.assertRaises(HttpResponseError):
-                    await self.dsc.undelete_file_system(filesystem.name, filesystem.deleted_version,
-                                                        new_name=existing_filesystem_client.file_system_name)
-
```
xiafu-msft, line 281: I think we still need to test restoring to existing filesystem/container, while it can be added later and tracked in an issue
the steps are
1. use soft delete enabled account
2. create container A and soft delete it
3. recreate container A
4. restore deleted container and see if we can get HttpResponseError

xiafu-msft (approved):

PR #18692, opened 2021-05-13
[AutoRelease] t2-netapp-2021-05-13-22948

azclibot:
https://github.com/Azure/sdk-release-request/issues/1542


hildurhodd:
@00Kai0 @msyyc Can we please get this moving?

msyyc:
autorest version is old and it is better to rerun with new autorest

PR #18699, opened 2021-05-13
[AutoRelease] t2-imagebuilder-2021-05-13-12144(wave4)

azclibot:
a issue also relation this service


RAY-316:
new version will release

PR #18711, opened 2021-05-13
updating sample for beta2 api

### LINKED ISSUE #18657: Update sample_create_client.py for Beta 3 API
The [sample_create_client.py](https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/containerregistry/azure-containerregistry/samples/sample_create_client.py) feels like it fits to the shape of the original Beta 1 sample.  Please consider making the following updates:

- [x] Refer to `ContainerRepository` as something other than client [here](https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/containerregistry/azure-containerregistry/samples/sample_create_client.py#L40) and [here](https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/containerregistry/azure-containerregistry/samples/sample_create_client.py#L61)
- [x] Since `ContainerRepository` isn't a client, we shouldn't be able to create an instance of it directly as we do [here](https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/containerregistry/azure-containerregistry/samples/sample_create_client.py#L46), unless this is a Python-specific requirement.
- [x] In the proposed Beta3 API, we're not exposing functionality to list tags on a repository directly as we do [here](https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/containerregistry/azure-containerregistry/samples/sample_create_client.py#L65).
- [x] If you'd like to illustrate the use of `ContainerRepository` in this sample, consider also including an illustration of `ArtifactRegistry`, as it is another Helper type that makes service calls.

seankane-msft:
closes #18657

### DIFF
--- a/sdk/containerregistry/azure-containerregistry/samples/async_samples/sample_create_client_async.py
+++ b/sdk/containerregistry/azure-containerregistry/samples/async_samples/sample_create_client_async.py
@@ -27,53 +27,40 @@ import os
 class CreateClients(object):
     def __init__(self):
         load_dotenv(find_dotenv())
-        self.account_url = os.environ["CONTAINERREGISTRY_ENDPOINT"]
 
     async def create_registry_client(self):
         # Instantiate the ContainerRegistryClient
         # [START create_registry_client]
         from azure.containerregistry.aio import ContainerRegistryClient
         from azure.identity.aio import DefaultAzureCredential
+        account_url = os.environ["CONTAINERREGISTRY_ENDPOINT"]
 
-        client = ContainerRegistryClient(self.account_url, DefaultAzureCredential())
+        client = ContainerRegistryClient(account_url, DefaultAzureCredential())
         # [END create_registry_client]
-
-    async def create_repository_client(self):
-        # Instantiate the ContainerRegistryClient
-        # [START create_repository_client]
-        from azure.containerregistry.aio import ContainerRepository
-        from azure.identity.aio import DefaultAzureCredential
-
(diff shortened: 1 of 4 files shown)


annelo-msft (approved):
LGTM - thanks!  Possibly worth getting Python dev eyes on, per your judgment.

sdk/containerregistry/azure-containerregistry/samples/async_samples/sample_create_client_async.py
```diff
@@ -27,53 +27,40 @@
...
         # [START create_registry_client]
         from azure.containerregistry.aio import ContainerRegistryClient
         from azure.identity.aio import DefaultAzureCredential
+        account_url = os.environ["CONTAINERREGISTRY_ENDPOINT"]
 
-        client = ContainerRegistryClient(self.account_url, DefaultAzureCredential())
+        client = ContainerRegistryClient(account_url, DefaultAzureCredential())
         # [END create_registry_client]
 
-    async def create_repository_client(self):
```
swathipil, line 41: did you move the creating repository client sample elsewhere?

swathipil (approved):
one question but LGTM

sdk/containerregistry/azure-containerregistry/samples/async_samples/sample_create_client_async.py
seankane-msft: This object isn't meant to be instantiated directly so I removed the sample

PR #18715, opened 2021-05-13
update for aiohttp 3.6

xiangyan99:


kristapratico (approved):

xiangyan99:
in aiohttp 3.6 auto_compress property is not readable but when we set it in ctor, it takes effect.

PR #18716, opened 2021-05-13
moving some acr specific recording infra to the general replacers

seankane-msft:


tools/azure-devtools/src/azure_devtools/scenario_tests/recording_processors.py
```diff
@@ -220,6 +231,13 @@ def process_response(self, response):
             self.replace_header(response, 'location', old, new)
             self.replace_header(response, 'operation-location', old, new)
             self.replace_header(response, 'azure-asyncoperation', old, new)
+            self.replace_header(response, "www-authenticate", old, new)
+
+        try:
+            for old, new in self.names_name:
+                response["url"].replace(old, new)
+        except KeyError:
+            pass
```
seankane-msft, line 240: @chlowell would this solve your issue?

chlowell: It would 👍

sdk/containerregistry/azure-containerregistry/tests/testcase.py
```diff
@@ -191,18 +143,13 @@ def get_token(self, *args):
 
 class ContainerRegistryTestClass(AzureTestCase):
     def __init__(self, method_name):
-        super(ContainerRegistryTestClass, self).__init__(
-            method_name,
-            recording_processors=[
```
scbedd, line 196: For my own clarity. We are using the default set of recording_processors from `azure_testcase`, which would be just

```
    def _get_recording_processors(self):
        return [
            GeneralNameReplacer()
            AuthenticationMetadataFilter(),
            OAuthRequestResponsesFilter(),
            RequestUrlNormalizer(),
        ]
```

Seems fine to me!

scbedd (approved):
Seems reasonable. I kicked a test run of [storage](https://dev.azure.com/azure-sdk/public/_build/results?buildId=903606&view=results) just to double check.

Issue #18727, opened 2021-05-13
[EH/SB] re-add mgmt to ci yml


swathipil:
undo: #18724

PR #18738, opened 2021-05-13
[SB/EH] revert removing mgmt from ci

swathipil:
fixes: #18727


scbedd (approved):
Acknowledging that merging this PR breaks the build. Test re-record is necessary. Mgmt folks want to release track2, which means they'll be in this area getting that fixed very soon.

PR #18739, opened 2021-05-14
[Communication]: Added unit tests to check if idempotence parameters are being set

lsundaralingam:


beltr0n:
These are good, but is there a way to test the repeatabilityResponse coming back from the server as well? that's what we're testing in Java/.Net/JS

lsundaralingam:
> These are good, but is there a way to test the repeatabilityResponse coming back from the server as well? that's what we're testing in Java/.Net/JS

There doesn't seem to be way in Python to get the raw HTTP response back in the live tests like we do for the other languages, butI don't see anything that we're missing here in terms of test coverage. We have live tests that would fail if there's something wrong with the repeatabilityResponse and we have these tests to ensure that the parameters are being set.

beltr0n (approved):

jbeauregardb (approved):

PR #18766, opened 2021-05-15
[Key Vault] Align secrets with other languages

mccoyp:
Adds a `KeyVaultSecretIdentifier` (part of #16183).


chlowell (approved):

sdk/keyvault/azure-keyvault-secrets/azure/keyvault/secrets/_models.py
```diff
@@ -259,7 +259,7 @@ class KeyVaultSecretIdentifier(object):
             :dedent: 8
     """
 
-    def __init__(self, id):
+    def __init__(self, id):  # pylint: disable=W0622
```
chlowell, line 262: I would've gone with something like `secret_id`

mccoyp: I did this for alignment and because it didn't seem like a parameter name would be an issue (since we don't use the name as a variable), but I can ask about this

PR #18772, opened 2021-05-17
reset continuation token

xiangyan99:


kristapratico (commented):
Is there a test for this?

kristapratico (approved):

PR #18789, opened 2021-05-18
[formrecognizer] update samples to train a model where a model_id is needed

kristapratico:
Resolves https://github.com/Azure/azure-sdk-for-python/issues/18465

- Adds the option to provide a model_id or train a model for any sample that requires a model_id to run. 
- Improves docs to point to the sample training files location.
- Removes the last of our static environment variables so samples no longer need to rely on an existing FR resource. This means I don't have to maintain said resource and periodically delete models from it. :)

### DIFF
--- a/sdk/formrecognizer/azure-ai-formrecognizer/samples/async_samples/sample_copy_model_async.py
+++ b/sdk/formrecognizer/azure-ai-formrecognizer/samples/async_samples/sample_copy_model_async.py
@@ -28,7 +28,7 @@ USAGE:
     5) AZURE_SOURCE_MODEL_ID - the model ID from the source resource to be copied over to the target resource.
         - OR -
        CONTAINER_SAS_URL - The shared access signature (SAS) Url of your Azure Blob Storage container with your forms.
-       A model will be trained and used to to run the sample.
+       A model will be trained and used to run the sample.
     6) AZURE_FORM_RECOGNIZER_TARGET_REGION - the region the target resource was created in
     7) AZURE_FORM_RECOGNIZER_TARGET_RESOURCE_ID - the entire resource ID to the target resource
 """
@@ -47,7 +47,7 @@ class CopyModelSampleAsync(object):
         source_key = os.environ["AZURE_FORM_RECOGNIZER_KEY"]
         target_endpoint = os.environ["AZURE_FORM_RECOGNIZER_TARGET_ENDPOINT"]
         target_key = os.environ["AZURE_FORM_RECOGNIZER_TARGET_KEY"]
-        source_model_id = os.getenv("AZURE_SOURCE_MODEL_ID") or custom_model_id
+        source_model_id = os.getenv("AZURE_SOURCE_MODEL_ID", custom_model_id)
         target_region = os.environ["AZURE_FORM_RECOGNIZER_TARGET_REGION"]
         target_resource_id = os.environ["AZURE_FORM_RECOGNIZER_TARGET_RESOURCE_ID"]
 
--- a/sdk/formrecognizer/azure-ai-formrecognizer/samples/async_samples/sample_copy_model_async.py
+++ b/sdk/formrecognizer/azure-ai-formrecognizer/samples/async_samples/sample_copy_model_async.py
@@ -14,6 +14,9 @@ DESCRIPTION:
     to a target Form Recognizer resource. The resource id and the resource region can be found
     in the azure portal.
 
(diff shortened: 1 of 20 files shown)


sdk/formrecognizer/azure-ai-formrecognizer/samples/async_samples/sample_differentiate_output_models_trained_with_and_without_labels_async.py
```diff
@@ -13,6 +13,7 @@
     This sample demonstrates the differences in output that arise when begin_recognize_custom_forms
     is called with custom models trained with labels and without labels. The models used in this
     sample can be created in sample_train_model_with_labels_async.py and sample_train_model_without_labels_async.py
+    using the training files found here: https://aka.ms/azsdk/formrecognizer/sampletrainingfiles
```
maririos, line 16: smart!

sdk/formrecognizer/tests.yml
```diff
@@ -30,19 +30,3 @@ stages:
         TEST_MODE: 'RunLiveNoRecord'
         AZURE_SKIP_LIVE_RECORDING: 'True'
         AZURE_TEST_RUN_LIVE: 'true'
-        # EnvVars for samples to run. TODO: These use an existing FR resource. Fix samples so they don't have to
```
maririos, line 33: woohooooo

sdk/formrecognizer/azure-ai-formrecognizer/samples/async_samples/sample_copy_model_async.py
```diff
@@ -23,6 +26,9 @@
     3) AZURE_FORM_RECOGNIZER_TARGET_ENDPOINT - the endpoint to your target Form Recognizer resource.
     4) AZURE_FORM_RECOGNIZER_TARGET_KEY - your target Form Recognizer API key
     5) AZURE_SOURCE_MODEL_ID - the model ID from the source resource to be copied over to the target resource.
+        - OR -
+       CONTAINER_SAS_URL - The shared access signature (SAS) Url of your Azure Blob Storage container with your forms.
+       A model will be trained and used to to run the sample.
```
maririos, line 31: duplicate `to to` it is in other places too

maririos: And not related, but in my head, when I read it in Spanish it sounded like: https://www.youtube.com/watch?v=5AkDqm-cEgg

maririos (approved):

sdk/formrecognizer/azure-ai-formrecognizer/samples/async_samples/sample_copy_model_async.py
```diff
@@ -33,15 +39,15 @@
...
+    async def copy_model_async(self, custom_model_id):
         from azure.core.credentials import AzureKeyCredential
         from azure.ai.formrecognizer.aio import FormTrainingClient
 
         source_endpoint = os.environ["AZURE_FORM_RECOGNIZER_ENDPOINT"]
         source_key = os.environ["AZURE_FORM_RECOGNIZER_KEY"]
         target_endpoint = os.environ["AZURE_FORM_RECOGNIZER_TARGET_ENDPOINT"]
         target_key = os.environ["AZURE_FORM_RECOGNIZER_TARGET_KEY"]
-        source_model_id = os.environ["AZURE_SOURCE_MODEL_ID"]
+        source_model_id = os.getenv("AZURE_SOURCE_MODEL_ID") or custom_model_id
```
iscai-msft, line 50: nit: i think you can do os.getenv("AZURE_SOURCE_MODEL_ID", custom_model_id)

iscai-msft (approved):

catalinaperalta (approved):
Agree with the other nits. Aside from that this looks great! I think the samples will be so much easier to use now as well! 🎉

sdk/formrecognizer/azure-ai-formrecognizer/samples/async_samples/sample_copy_model_async.py
kristapratico: thanks for catching this! love the song :)

sdk/formrecognizer/azure-ai-formrecognizer/samples/async_samples/sample_copy_model_async.py
kristapratico: good catch! I was using os.environ before and then missed this when I switched to getenv :)

PR #18790, opened 2021-05-18
[Tables] Add tests for cosmos sas

### LINKED ISSUE #18444: [Tables] Add back tests for Cosmos + SAS credentials
We previously had tests for using SAS tokens against a Cosmos endpoint. These were removed as it was believe not to be a supported scenario - however _**Table-level**_ SAS tokens should be supported. So we should add the applicable tests back again.

I think this where they were removed:
https://github.com/Azure/azure-sdk-for-python/pull/18075

seankane-msft:
closes #18444


sdk/tables/azure-data-tables/tests/test_table.py
```diff
@@ -383,7 +383,6 @@ def test_set_table_acl_too_many_ids(self, tables_storage_account_name, tables_pr
         finally:
             ts.delete_table(table.table_name)
 
-    @pytest.mark.live_test_only
```
annatisch, line 386: These used to be live-only to prevent the sas tokens being in the recordings. Are they being substituted out of recordings?

sdk/tables/azure-data-tables/tests/test_table_batch_cosmos.py
```diff
@@ -474,6 +478,56 @@ def test_delete_batch_with_bad_kwarg(self, tables_cosmos_account_name, tables_pr
...
+            table = service.get_table_client(self.table_name)
+
+            entity = TableEntity()
+            entity['PartitionKey'] = 'batch_inserts'
+            entity['test'] = EntityProperty(True, EdmType.BOOLEAN)
+            entity['test2'] = 'value'
+            entity['test3'] = 3
+            entity['test4'] = EntityProperty(1234567890, EdmType.INT32)
+# 'https://seankaneprim.table.core.windows.net/$batch?st=2021-05-18T16%3A19%3A58Z&se=2021-05-18T17%3A20%3A58Z&sp=raud&sv=2019-02-02&tn=uttabled4f0e8f&sig=cROztTVf0vKXUE1IQyyJpnjA3Ua1a4NMYYF0VnCMt/4%3D'
+# 'https://seankaneprim.table.core.windows.net/$batch?st=2021-05-18T16%3A20%3A57Z&se=2021-05-18T17%3A21%3A57Z&sp=raud&sv=2019-02-02&tn=uttable81841182&sig=Czf9WOvwZT790ZaUqOPanvpa04mtBMQ28xvAr/l%2BPbM%3D'
```
annatisch, line 512: We should avoid putting the sas tokens in the tests as well ;)

sdk/tables/azure-data-tables/tests/test_table.py
seankane-msft: Yes they are now.

sdk/tables/azure-data-tables/tests/test_table_batch_cosmos.py
seankane-msft: Ahhh that was dumb, I was comparing when i had some failures and never removed them. Big mistake on my part

annatisch (approved):

PR #18806, opened 2021-05-19
[AutoRelease] t2-sqlvirtualmachine-2021-05-19-98245(wave4)

azclibot:
Please Add the link issue


RAY-316:
@autorest/python@5.6.6" should be 5.8.0

PR #18818, opened 2021-05-19
Updated docstr and typehints to reflect apiview.dev requirements

LarryOsterman:


sdk/attestation/azure-security-attestation/azure/security/attestation/_administration_client.py
```diff
@@ -44,9 +44,11 @@
 class AttestationAdministrationClient(object):
     """Provides administrative APIs for managing an instance of the Attestation Service.
 
-    :param str instance_url: base url of the service
+    :param instance_url: base url of the service
+    :paramtype instance_url: str
     :param credential: Credentials for the caller used to interact with the service.
-    :type credential: azure.core.credentials.TokenCredential
+    :paramtype credential: azure.core.credentials.TokenCredential
+
```
seankane-msft, line 51: If these work in apiview and sphinx then ignore the rest of this comment.

If not, [here](https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/formrecognizer/azure-ai-formrecognizer/azure/ai/formrecognizer/_form_training_client.py#L53-L63) is a great example in formrecognizer. `:paramtype` is used for keyword arguments, `:type` is used for parameters. You can do single line type hints for the basic types (bool, int, float, str, etc.), the two line type hint is preferred for any `azure` types.

Issue #18795, opened 2021-05-18
[ServiceBus] Receiver iterator exits the loop silently after 10 mins idle timeout


yunhaoling:
- **Package Name**: azure-servicebus
- **Package Version**: 7.2.0
- **Operating System**: Windows (should apply to all platforms)
- **Python Version**: 3.7

**Describe the bug**
service bus receiver iterator exit the loop silently after 10 mins idle timeout (force detached by the service)

**To Reproduce**
```python
import os
import time
import logging
logging.basicConfig(level=logging.INFO)
from azure.servicebus import ServiceBusClient

CONNECTION_STR = os.environ['SERVICE_BUS_CONNECTION_STR']
QUEUE_NAME = os.environ["SERVICE_BUS_QUEUE_NAME"]
servicebus_client = ServiceBusClient.from_connection_string(conn_str=CONNECTION_STR, logging_enable=True)

with servicebus_client:
    receiver = servicebus_client.get_queue_receiver(queue_name=QUEUE_NAME)
    with receiver:
        for msg in receiver:
            print(str(msg))
            receiver.complete_message(msg)
            print('start sleeping')
            time.sleep(11*60)
            print('finish sleeping')

print("Receive is done.")
```

**Expected behavior**
`for msg in receiver` should keep receiving the message after 10 minutes idle timeout.

**Additional context**
- link detach would result in a `ServiceBusConnectionError` which is retryable
- the receiver iterator exits the loop after the connection is reestablished.
- complete logging:
```
INFO:uamqp.connection:Connection b'SBReceiver-11eb87d7-4b78-4dad-964f-122b57a72b34' state changed from <ConnectionState.START: 0> to <ConnectionState.START: 0>
INFO:uamqp.connection:Connection b'SBReceiver-11eb87d7-4b78-4dad-964f-122b57a72b34' state changed from <ConnectionState.START: 0> to <ConnectionState.START: 0>
INFO:uamqp.c_uamqp:b'-> Header (AMQP 0.1.0.0)'
INFO:uamqp.connection:Connection b'SBReceiver-11eb87d7-4b78-4dad-964f-122b57a72b34' state changed from <ConnectionState.START: 0> to <ConnectionState.HDR_SENT: 2>
INFO:uamqp.c_uamqp:b'<- Header (AMQP 0.1.0.0)'
[...]

PR #18826, opened 2021-05-20
[ServiceBus] Fix receiver iterator stops after recovery from error

yunhaoling:
fixing issue: https://github.com/Azure/azure-sdk-for-python/issues/18795

once we encounter a connection error, we will shutdown the broken amqp handler and create a new one.
however, the internal service bus message iterator hooked with the amqp handler was not cleaned up when we were shutting down the broken amqp handler.

so to fix this, we just need to set the internal service bus message iterator to None when we shutdown the broken amqp handler.


swathipil (approved):
lgtm

PR #18834, opened 2021-05-20
ACR - Resolve MSDocs Errors

scbedd:
@seankane-msft I checked our custom pylint plugin and the reason this slipped through is because ACR is using `... codeblock` instead of `... literalinclude`. Our pylint plugin catches the necessary newline because it's not _looking_ at codeblock segments.

I want to submit a PR to add these to our existing pylint checks, but for now, this will getcha online.

Are you comfortable with publishing the current version of this package into our preview docs? (this would be a manually generated 1.0.0b3)


seankane-msft:
These changes need to be applied to #18667

PR #18839, opened 2021-05-20
[Communication]: Updated swagger paths

lsundaralingam:


JoshuaLai (approved):
The changes to the swaggers looks good. I am not sure why there is added casting to `Any` but I assume it is an autorest version change? The one thing that does look concerning is the polling from false to true. I would leave that to a Python expert to comment on.

jbeauregardb (approved):

jbeauregardb:
@JoshuaLai I don't know if that change was deliberate, but from a customer perspective I think polling being set to True as default is a better choice.

lsundaralingam:
> @JoshuaLai I don't know if that change was deliberate, but from a customer perspective I think polling being set to True as default is a better choice. @lsundaralingam If the polling kwarg is being set to True as default... Should we change our tests code to remove the kwarg from the function call? It shouldn't be required anymore. On the other side, I don't know if this change was deliberate and they may change it back.

I discussed with @jbeauregardb offline and we decided on keeping the kwarg since it's safer to always pass in the parameter.

JoshuaLai:
> > @JoshuaLai I don't know if that change was deliberate, but from a customer perspective I think polling being set to True as default is a better choice. @lsundaralingam If the polling kwarg is being set to True as default... Should we change our tests code to remove the kwarg from the function call? It shouldn't be required anymore. On the other side, I don't know if this change was deliberate and they may change it back.
> 
> I discussed with @jbeauregardb offline and we decided on keeping the kwarg since it's safer to always pass in the parameter.
Awesome! Does that mean this is safe to merge in now?

PR #18846, opened 2021-05-20
Read tenant and authority from VS Code user settings

### LINKED ISSUE #14808: Set authority from VS Code settings
Users can configure VisualStudioCodeCredential's authority with the `authority` keyword argument, and this is necessary in all clouds save Public Cloud. This could instead be optional, because the credential could determine the correct authority from the user's VS Code settings by mapping the configured cloud name to an authority's URL. Other SDKs do this, for example in Java: https://github.com/Azure/azure-sdk-for-java/blob/d1c4f95452b289bb04ff749652bb41278832e4b0/sdk/identity/azure-identity/src/main/java/com/azure/identity/VisualStudioCodeCredential.java#L47

chlowell:
Closes #14808


mccoyp (approved):
Looks good -- verified locally that this works on Mac (and tests bring up requests for Keychain access)

Issue #18778, opened 2021-05-17
[translation] update generated code to target v1.0


kristapratico:
https://github.com/Azure/azure-rest-api-specs/blob/master/specification/cognitiveservices/data-plane/TranslatorText/stable/v1.0/TranslatorBatch.json

There will need to be some updates to the mapping since I think some renames were done in the swagger for the LLC clients.

PR #18861, opened 2021-05-21
[translation] regenerate on v1.0

kristapratico:
Resolves https://github.com/Azure/azure-sdk-for-python/issues/18778

Regenerates on the v1.0 swagger and rerecords all tests.


sdk/translation/azure-ai-translation-document/azure/ai/translation/document/_client.py
```diff
@@ -11,7 +11,7 @@
 from azure.core.credentials import AzureKeyCredential
 from azure.core.pipeline.policies import AzureKeyCredentialPolicy
 from ._generated import BatchDocumentTranslationClient as _BatchDocumentTranslationClient
-from ._generated.models import BatchStatusDetail as _BatchStatusDetail
+from ._generated.models import TranslationStatus as _TranslationStatus
```
maririos, line 14: this is because of the swagger update, right? did you change any public api ? not sure how I can see that :(

sdk/translation/azure-ai-translation-document/tests/test_list_submitted_jobs.py
```diff
@@ -125,7 +125,7 @@ def test_list_submitted_jobs_filter_by_created_after(self, client):
         docs_per_job = 2
 
         # create some jobs
-        start = datetime.now()
+        start = datetime.utcnow()
```
maririos, line 128: was this a bug?

sdk/translation/azure-ai-translation-document/azure/ai/translation/document/_client.py
kristapratico: This is updated due to the renames in the swagger. No public API changes.

sdk/translation/azure-ai-translation-document/tests/test_list_submitted_jobs.py
kristapratico: This is in the tests, so not an SDK bug. Just an adjustment so that these tests pass for whatever timezone you are running it in.

maririos (approved):
Changelog update

PR #18863, opened 2021-05-21
Support BuildTargetingString queue time variable

scbedd:
This means we can queue and override it.

### DIFF
--- a/eng/pipelines/templates/stages/archetype-sdk-client.yml
+++ b/eng/pipelines/templates/stages/archetype-sdk-client.yml
@@ -64,6 +64,8 @@ parameters:
 stages:
   - stage: Build
     jobs:
+
+
     - template: /eng/pipelines/templates/jobs/ci.yml
       parameters:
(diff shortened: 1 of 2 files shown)


eng/pipelines/templates/stages/archetype-sdk-client.yml
```diff
@@ -63,6 +63,9 @@ parameters:
 
 stages:
   - stage: Build
+    variables:
+      - name: BuildTargetingStringValue
+        value: $[ coalesce(variables['BuildTargetingString'], '${{ parameters.BuildTargetingString }}') ]
```
weshaggard, line 68: This approach will work but only if BuildTargetingString isn't set as a variable anywhere else.

weshaggard (approved):

eng/pipelines/templates/stages/archetype-sdk-client.yml
scbedd: Yep.

PR #18864, opened 2021-05-21
Rename internal _identity_config -> identity_config

chlowell:
`identity_config` is part of ManagedIdentityCredential's public API now, so the indirection that prevents an application from passing it through to ManagedIdentityClient is unnecessary and just confusing.

### DIFF
--- a/sdk/identity/azure-identity/azure/identity/_credentials/app_service.py
+++ b/sdk/identity/azure-identity/azure/identity/_credentials/app_service.py
@@ -63,7 +63,7 @@ def _get_client_args(**kwargs):
     return dict(
         kwargs,
         _content_callback=_parse_app_service_expires_on,
-        _identity_config=identity_config,
+        identity_config=identity_config,
         base_headers={"secret": secret},
         request_factory=functools.partial(_get_request, url),
     )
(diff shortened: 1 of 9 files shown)


sdk/identity/azure-identity/azure/identity/_credentials/app_service.py
```diff
@@ -63,7 +63,7 @@ def _get_client_args(**kwargs):
     return dict(
         kwargs,
         _content_callback=_parse_app_service_expires_on,
-        _identity_config=identity_config,
+        identity_config=identity_config,
```
mccoyp, line 66: Is there a reason to pop this from `kwargs` and pass it here, rather than leaving it in like in the [Service Fabric credential](https://github.com/Azure/azure-sdk-for-python/pull/18864/files#diff-3754eb569e7416961c0972e95ceec52d6697189aed8b25df4192a7cc994bd041L52)?

mccoyp (approved):

sdk/identity/azure-identity/azure/identity/_credentials/app_service.py
chlowell: Because this version of App Service is unique in its "clientid" parameter (every other platform uses "client_id"), we use `identity_config` to pass along the caller's `client_id` argument, i.e. we may have `identity_config={"clientid": kwargs["client_id"]}` here. You're right to be suspicious, this is mysterious and deserves some comment at least.

PR #18886, opened 2021-05-24
[formrecognizer] add version info on models

catalinaperalta:


sdk/formrecognizer/azure-ai-formrecognizer/azure/ai/formrecognizer/_models.py
```diff
@@ -132,6 +132,9 @@ class Point(namedtuple("Point", "x y")):
 
     :ivar float x: x-coordinate
     :ivar float y: y-coordinate
+
```
kristapratico, line 135: I'm confused - I thought we already added this and it does show up in master? https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/formrecognizer/azure-ai-formrecognizer/azure/ai/formrecognizer/_models.py#L136

catalinaperalta: Oh you're right. I was checking and I didnt see if for some reason so I thought I never pushed it. 😮 Well then I'm just going to close this PR. Thanks for showing me that! Sorry about the confusion!

kristapratico: No problem! Just curious why the diff shows that it's not there. I wonder if it's because it's from your forked master -> master. Weird

catalinaperalta: Yeah the diff is what really threw me for a loop

PR #18904, opened 2021-05-25
[STG78] Added Quick Query Parquet Dialect

tasherif-msft:


sdk/storage/azure-storage-blob/tests/test_quick_query.py
```diff
@@ -936,4 +936,52 @@ def on_error(error):
...
+
+        blob_name = self._get_blob_reference()
+        blob_client = bsc.get_blob_client(self.container_name, blob_name)
+        parquet_path = os.path.abspath(os.path.join(os.path.abspath(__file__), "..", "./resources/parquet.parquet"))
+        with open(parquet_path, "rb") as parquet_data:
+            blob_client.upload_blob(parquet_data, overwrite=True)
+        errors = []
+
+        def on_error(error):
+            errors.append(error)
```
xiafu-msft, line 981: these seems not being used

shared_requirements.txt
```diff
@@ -156,7 +156,7 @@ pyjwt>=1.7.1
 #override azure-storage-file-share azure-core<2.0.0,>=1.10.0
 #override azure-storage-file-datalake azure-core<2.0.0,>=1.10.0
 #override azure-storage-file-datalake msrest>=0.6.21
-#override azure-storage-file-datalake azure-storage-blob<13.0.0,>=12.8.0b1
+#override azure-storage-file-datalake azure-storage-blob<13.0.0,>=12.9.0b1
```
tasherif-msft, line 159: Need to revert this since we no longer have a parquet dialect class

sdk/storage/azure-storage-file-datalake/tests/test_quick_query.py
```diff
@@ -866,4 +866,47 @@ def on_error(error):
...
+    @DataLakePreparer()
+    def test_quick_query_output_in_parquet_format(self, datalake_storage_account_name, datalake_storage_account_key):
+        # Arrange
+        self._setUp(datalake_storage_account_name, datalake_storage_account_key)
+        file_name = self._get_file_reference()
+        file_client = self.dsc.get_file_client(self.filesystem_name, file_name)
+        errors = []
+
+        def on_error(error):
+            errors.append(error)
```
xiafu-msft, line 900: these are not being used since there's no place checking if there were error during service processing the content

tasherif-msft: Yes my bad I forgot to remove these two lines! good catch

shared_requirements.txt
xiafu-msft, line 159: I forget if this worked previously, since the blob hasn't been released yet, is the pipeline smart enough to detect unreleased dependency? It seems what we did was releasing blob first then datalake while I cannot remember clearly.

tasherif-msft: I am reverting this change since I removed the ParquetDialect class. But usually yes we would need to do two releases as far as I remember - blob first then datalake. Though that might have changed

xiafu-msft (approved):

PR #18917, opened 2021-05-25
moving cog services nspkg to the nspkg folder

seankane-msft:


tjprescott (approved):
Thanks @seankane-msft!

PR #18919, opened 2021-05-25
NetAppFiles: Tests from track1 copied and updated and new recordings gathered

hildurhodd:
After NetAppFiles was migrated from track 1 to track 2 the tests were disabled and were left behind. Here are the tests up to date and new recordings.


hildurhodd:
@msyyc @RAY-316 can we get this moving? This is only tests and recordings that were skipped and left behind during the track 2 migration process

msyyc (approved):

PR #18929, opened 2021-05-25
[ACR] parse expiration time from refresh token

seankane-msft:


sdk/containerregistry/azure-containerregistry/azure/containerregistry/_helpers.py
```diff
@@ -84,3 +88,19 @@ def _enforce_https(request):
         raise ServiceRequestError(
             "Bearer token authentication is not permitted for non-TLS protected (non-https) URLs."
         )
+
+def _parse_exp_time(raw_token):
```
kristapratico, line 92: are there any tests for this?

seankane-msft: There were but I think they got removed in the merge conflict squashing. Thanks for catching that!

kristapratico (approved):

PR #18930, opened 2021-05-25
[Test EngSys] adding compression to scrubbers

### LINKED ISSUE #18812: [Test Infra] Scrub secrets from compressed bodies.

seankane-msft:
closes #18812


lmazuel (approved):

tools/azure-devtools/src/azure_devtools/scenario_tests/recording_processors.py
```diff
@@ -224,6 +216,7 @@ def _process_body_as_dict(self, body):
         return new_body
 
     def process_response(self, response):
+        response = _decompress_response_body(response)
```
chlowell, line 219: What do you think of doing this in `ReplayableTest._process_response_recording`? That is to say, decompressing the body before running any processors? No processor would need a change then, because they would all get the decompressed body they expect.

tools/azure-devtools/src/azure_devtools/scenario_tests/recording_processors.py
```diff
@@ -238,6 +229,8 @@ def process_response(self, response):
             self.replace_header(response, 'azure-asyncoperation', old, new)
             self.replace_header(response, "www-authenticate", old, new)
 
+        response["url"] = replace_subscription_id(response["url"])
```
chlowell, line 232: SubscriptionRecordingProcessor already does this, do we need to do it again here?

seankane-msft: `SubscriptionRecordingProcessor` is not used by default in the `AzureMgmtTestCase` class. It probably should be but I'd rather do that in a separate PR.

seankane-msft: #18981

PR #18935, opened 2021-05-25
resolve failing analyze steps + breaking azure-eventhub tests

scbedd:
This unblocks PR #18901


eng/tox/allowed_pylint_failures.py
```diff
@@ -51,4 +51,5 @@
     "azure-ai-anomalydetector",
     "azure-security-attestation",
     "azure-iot-deviceupdate",
+    "azure-eventhub"
```
mitchdenny, line 54: Does this mean that any PyLint failure is going to be ignored?

PR #18955, opened 2021-05-26
[Service Bus] Match the partition key with session id only when both are set

bishnu-shb:
The azure documentation says "if a message has the partition key property but not the session ID property set, then Service Bus uses the partition key property value as the partition key. If the message has both the session ID and the partition key properties set, both properties must be identical" [https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-partitioning#using-a-partition-key].


yunhaoling:
@bishnu-shb, thanks for your PR! we'll take a look ASAP

swathipil:
Hi @bishnu-shb,

I was in the middle of changing some failing tests to pass with your change. However, I accidentally pushed without pulling your changes and this closed the PR. I am no longer able to make changes to this PR. Would you be able to create a new PR with the same change (adding `and self.session_id` check to the ServiceBus `partition_key` property)?

Thank you for your patience!

bishnu-shb:
Created https://github.com/Azure/azure-sdk-for-python/pull/19233

PR #18964, opened 2021-05-26
[Tables] Various small fixes

annatisch:
* `TableClient.create_table` now returns an instance of `TableItem`.
* Any additional odata metadata is returned in entitys metadata.
* The timestamp in entity metadata is now deserialized to a timestamp.
* If the `prefer` header is added in the `create_entity` operation, the echo will be returned.
* Errors raised on a 412 if-not-match error will now be a specific `azure.core.exceptions.ResourceModifiedError`.
* `EdmType.DOUBLE` values are now explicitly typed in the request payload.


seankane-msft (approved):
small nits

PR #18965, opened 2021-05-26
Update test results titles to include service directory and cloud names

benbp:
When testing with multiple clouds, it can be hard to tell from the coverage view which failures come from which cloud and/or job.


mitchdenny (approved):
LGTM

weshaggard (approved):
Looks reasonable

scbedd (approved):

PR #18967, opened 2021-05-26
Add environment variable for redirecting IMDS token requests

chlowell:
Adding environment variable AZURE_POD_IDENTITY_TOKEN_URL (name suggestions welcome) to enable redirecting IMDS token requests. When the variable has a value, we assume the endpoint is available.


sdk/identity/azure-identity/tests/test_imds_credential.py
```diff
@@ -176,6 +177,45 @@ def test_identity_config():
...
 
+def test_imds_url_override():
+    url = "https://localhost/token"
+    expected_token = "***"
+    scope = "scope"
+    now = int(time.time())
+
+    transport = validating_transport(
+        requests=[
+            Request(base_url=url),
```
mccoyp, line 188: Would a request without required headers get sent in this case, since we assume that the endpoint is available?

chlowell: It does in this case because the test is wrong, thanks for pointing that out! 🤣

mccoyp (approved):

PR #18983, opened 2021-05-27
formatting tools code

seankane-msft:


scbedd (approved):
Gonna be honest I kinda skimmed because it looks like `black` did all of it :) 

Your checks are green. LGTM.

PR #18994, opened 2021-05-27
[text analytics] change return type of analyze actions to list of list

iscai-msft:
Instead of adding a new class `AnalyzeActionsResult`, we're going to reuse the current action result classes, i.e. `RecognizeEntitiesResult`.

Now, a call with documents ["Hello", "world"] and actions [`RecognizeEntitiesAction()`, `AnalyzeSentimentAction()`] will get you
[
    [`RecognizeEntitiesResult` of "Hello", `AnalyzeSentimentResult` of "Hello"],
    [`RecognizeEntitiesResult` of "world", `AnalyzeSentimentResult` of "world"],
]

This is more consistent with the return types of the normal operations, and allows us to use the models we currently have


sdk/textanalytics/azure-ai-textanalytics/CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 ## 5.1.0b8 (Unreleased)
 
+**Breaking Changes**
+- Changed the response structure of `being_analyze_actions`. Now, we return a list of list of an action result of documents
```
kristapratico, line 6: I think this should be "list of list**s**"? Also wondering if we can clarify the change a bit more - maybe "Now, we return a list of results, where each result is a list of the action results for the document, in the order the actions were passed". If that's more confusing then ignore me

sdk/textanalytics/azure-ai-textanalytics/tests/test_analyze.py
```diff
@@ -206,19 +206,20 @@ def test_all_successful_passing_text_document_input_entities_task(self, client):
...
+            assert len(document_results) == 1
+            document_result = document_results[0]
+            assert isinstance(document_result, RecognizeEntitiesResult)
+            assert len(document_result.entities) == 4
+            assert document_result.id is not None
+            for entity in document_result.entities:
+                assert entity.text is not None
+                assert entity.category is not None
+                assert entity.offset is not None
+                assert entity.confidence_score is not None
```
kristapratico, line 222: I suppose someday we should go through and convert all the unittest assert helpers to `assert` since we have a funny mix of both now. Not blocking on this PR, though

sdk/textanalytics/azure-ai-textanalytics/tests/test_analyze.py
```diff
@@ -602,41 +601,27 @@ def test_multiple_pages_of_results_returned_successfully(self, client):
...
+            _AnalyzeActionsType.RECOGNIZE_ENTITIES,
+            _AnalyzeActionsType.EXTRACT_KEY_PHRASES,
+            _AnalyzeActionsType.RECOGNIZE_PII_ENTITIES,
+            _AnalyzeActionsType.RECOGNIZE_LINKED_ENTITIES,
+            _AnalyzeActionsType.ANALYZE_SENTIMENT,
+        ]
+        action_type_to_document_results = defaultdict(list)
+
+        for doc_idx, document_result in enumerate(document_results):
+            for action_idx, document_result in enumerate(document_result):
```
kristapratico, line 616: we're redefining document_result in the loop here

sdk/textanalytics/azure-ai-textanalytics/tests/test_analyze_async.py
```diff
@@ -103,34 +106,35 @@ async def test_all_successful_passing_dict_sentiment_task(self, client):
...
-        self.assertEqual(len(action_result.document_results[1].sentences), 2)
-        self.assertEqual(action_result.document_results[1].sentences[0].text, "I did not like the hotel we stayed at.")
-        self.assertEqual(action_result.document_results[1].sentences[1].text, "It was too expensive.")
-        self.assertEqual(len(action_result.document_results[2].sentences), 2)
-        self.assertEqual(action_result.document_results[2].sentences[0].text, "The restaurant had really good food.")
-        self.assertEqual(action_result.document_results[2].sentences[1].text, "I recommend you try it.")
+            pages.append(p)
+
+        assert len(pages) == len(docs)
+        assert len(pages) == len(docs)
```
kristapratico, line 114: double checking? 😉

sdk/textanalytics/azure-ai-textanalytics/tests/testcase.py
```diff
@@ -5,6 +5,7 @@
 # Licensed under the MIT License. See License.txt in the project root for
 # license information.
 # --------------------------------------------------------------------------
+from re import L
```
kristapratico, line 8: do we need this?

kristapratico (approved):
nits

sdk/textanalytics/azure-ai-textanalytics/CHANGELOG.md
iscai-msft: I think i like that more, thank you @kristapratico !

sdk/textanalytics/azure-ai-textanalytics/tests/test_analyze_async.py
iscai-msft: lmao, will remove this

sdk/textanalytics/azure-ai-textanalytics/tests/testcase.py
iscai-msft: ooh, looks like pylance snuck in an autoimport here, will remove

sdk/textanalytics/azure-ai-textanalytics/README.md
```diff
@@ -532,72 +532,56 @@ poller = text_analytics_client.begin_analyze_actions(
...
-        print("...Data source language: {}".format(linked_entity.language))
-        print("...Data source entity ID: {}".format(linked_entity.data_source_entity_id))
-        print("...Data source URL: {}".format(linked_entity.url))
-        print("...Document matches:")
+pages = poller.result()
+
+for doc, document_results in zip(documents, pages):
+    print("\nDocument text: {}".format(doc))
+    recognize_entities_result = document_results[0]
+    assert not recognize_entities_result.is_error
```
johanste, line 540: This is probably not a good use of an `assert`. Any document could have failed for a transient reason, so you are not really in a position to say that this didn't fail. I would print a "failed" message instead...

johanste: Actually, scratch that, this is in the `readme.md`. It is a bit long for the readme already. I don't think we need to show all types of actions in the readme example inline. It definitely won't scale as new action types comes along. Can we limit this to two?

johanste (approved):

PR #18998, opened 2021-05-28
[STG78] Copy Source OAuth Token Support

tasherif-msft:


tasherif-msft:
rename to `source_authorization` and remove the "bearer" logic

xiafu-msft:
just a note
        :keyword str source_authorization:

xiafu-msft (approved):

Issue #18602, opened 2021-05-10
[EventGrid/Core] CloudEvent.from_dict fails to convert a datetime string to datetime object when microsecond exceeds the python limitation


yunhaoling:
- **Package Name**:  azure-eventgrid
- **Package Version**: 4.1.1
- **Operating System**: Windows
- **Python Version**: 3.7

**Describe the bug**
`CloudEvent.from_dict` fails to convert a datetime string to datetime object when the microsecond field exceeds the python limitation.

The time in the input dictionary is "2021-05-10T14:39:23.5811478Z" (This value is set by the event grid service).

```
Traceback (most recent call last):
  File "d:\projects\azure-sdk-for-python\sdk\core\azure-core\azure\core\_utils.py", line 65, in _convert_to_isoformat
    deserialized = datetime.datetime.strptime(timestamp, "%Y-%m-%dT%H:%M:%S.%f")
  File "C:\Users\yuling\Anaconda3\envs\sbenv37\lib\_strptime.py", line 577, in _strptime_datetime
    tt, fraction, gmtoff_fraction = _strptime(data_string, format)
  File "C:\Users\yuling\Anaconda3\envs\sbenv37\lib\_strptime.py", line 362, in _strptime
    data_string[found.end():])
ValueError: unconverted data remains: 8

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "D:/Projects/azure-sdk-for-python/sdk/eventhub/azure-eventhub/samples/eg.py", line 37, in <module>
    c = CloudEvent.from_dict(json_obj)
  File "d:\projects\azure-sdk-for-python\sdk\core\azure-core\azure\core\messaging.py", line 166, in from_dict
    time=_convert_to_isoformat(event.get("time")),
  File "d:\projects\azure-sdk-for-python\sdk\core\azure-core\azure\core\_utils.py", line 67, in _convert_to_isoformat
    deserialized = datetime.datetime.strptime(timestamp, "%Y-%m-%dT%H:%M:%S")
  File "C:\Users\yuling\Anaconda3\envs\sbenv37\lib\_strptime.py", line 577, in _strptime_datetime
    tt, fraction, gmtoff_fraction = _strptime(data_string, format)
  File "C:\Users\yuling\Anaconda3\envs\sbenv37\lib\_strptime.py", line 362, in _strptime
    data_string[found.end():])
ValueError: unconverted data remains: .5811478

Process finished with exit code 1
```

[...]

PR #19019, opened 2021-06-01
Handle 6 decimal places cloud event

rakshith91:
Fixes https://github.com/Azure/azure-sdk-for-python/issues/18602

### DIFF
--- a/sdk/core/azure-core/azure/core/_utils.py
+++ b/sdk/core/azure-core/azure/core/_utils.py
@@ -53,6 +53,17 @@ def _convert_to_isoformat(date_time):
         sign, offset = date_time[-6], date_time[-5:]
         delta = int(sign + offset[:1]) * 60 + int(sign + offset[-2:])
 
+    check_decimal = timestamp.split('.')
+    if len(check_decimal) > 1:
+        decimal_str = ""
+        for digit in check_decimal[1]:
+            if digit.isdigit():
+                decimal_str += digit
+            else:
+                break
+        if len(decimal_str) > 6:
+            timestamp = timestamp.replace(decimal_str, decimal_str[0:6])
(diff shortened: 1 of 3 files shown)


sdk/core/azure-core/azure/core/_utils.py
```diff
@@ -53,6 +53,17 @@ def _convert_to_isoformat(date_time):
...
+    check_decimal = timestamp.split('.')
+    if len(check_decimal) > 1:
+        decimal_str = ""
+        for digit in check_decimal[1]:
+            if digit.isdigit():
+                decimal_str += digit
+            else:
+                break
+        if len(decimal_str) > 6:
+            timestamp = timestamp.replace(decimal_str, decimal_str[0:6])
```
xiangyan99, line 65: I think we have this discussion. If the precision is more than 6, should we just drop the rest?

rakshith91: yes - that's what we have decided :)

xiangyan99: Do we want to log a warning message for precision loss?

rakshith91: i don't think it's necessary -  this is an event, there is no scenario where you would re-create a PUT request for ARM and expect precise value.

xiangyan99: and is it possible that timestamp.replace(decimal_str, decimal_str[0:6]) applies to some other places other than the last segment?

rakshith91: nope - it should not

xiangyan99 (approved):

PR #19031, opened 2021-06-02
Cosmos DB Python SDK updates

Rodrigossz:
New limitations, max_item_count, Integrated cache.


sdk/cosmos/azure-cosmos/README.md
```diff
@@ -136,6 +150,12 @@ Cosmos DB SQL language allows you to [get subitems by using the FROM clause](htt
 * For SQL queries using the `query_items` method, this SDK demands that you specify the `partition_key` or use the `enable_cross_partition_query` flag. 
 * If you are getting subitems and specifying the `partition_key`, please make sure that your partition key is included in the subitems, which is not true for most of the cases.
 
+## Max Item Count
+
+This is a parameter of the `query_items` method, an integer indicating the maximum number of items to be returned per page. The `-1` value can be specified to let the service determine the optimal item count. This is the recommended configuration value, and the default behavior of this SDK when it is not informed.
```
southpolesteve, line 155: This is confusing. The first paragraph makes it sound like -1 is preferred. Here Is what I would say:

> This is a parameter of the `query_items` method, an integer indicating the maximum number of items to be returned per page. The `None` value can be specified to let the service determine the optimal item count. This is the recommended configuration value, and the default behavior of this SDK when it is not set.

southpolesteve (approved):

PR #19039, opened 2021-06-02
[Storage][Test] Test multiple service versions

seankane-msft:


kasobol-msft (commented):
LGTM.
Could you pick 1 or 2 tests in each storage package and make them use this? So that folks have reference to look at how to use it as well as prove that this is working as expected?

tasherif-msft:
A working example would def be great here

sdk/storage/platform-matrix.json
```diff
@@ -0,0 +1,20 @@
...
+        "$IMPORT": "eng/pipelines/templates/stages/platform-matrix.json"
+    },
+    "exclude": [
+        { "OSVmImage": "macOS-10.15" }
+    ],
+    "include": [
+        {
+            "Agent": { "macOS-10.15": { "OSVmImage": "macOS-10.15", "Pool": "Azure Pipelines" } },
+            "JavaTestVersion": "1.8",
+            "AZURE_TEST_HTTP_CLIENTS": "netty"
```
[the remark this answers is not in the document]
seankane-msft: Good catch, I'll fix that

sdk/storage/tests.yml
```diff
@@ -26,6 +26,17 @@ stages:
...
           - Pool=(.*)-general/$1-storage
+        MatrixConfigs:
+          - ${{ if contains(variables['Build.DefinitionName'], 'tests-weekly') }}:
+            - Name: Storage_all_versions_live_test
+              Path: sdk/storage/platform-matrix-all-versions.json
+              Selection: sparse
+              GenerateVMJobs: true
+          - ${{ if not(contains(variables['Build.DefinitionName'], 'tests-weekly')) }}:
+            - Name: Storage_live_test
+              Path: sdk/storage/platform-matrix.json
```
benbp, line 37: Since storage uses the default matrix for python unlike java, I would delete `MatrixConfigs` and add the weekly config this way:

```
      ${{ if contains(variables['Build.DefinitionName'], 'tests-weekly') }}:
        AdditionalMatrixConfigs:
            - Name: Storage_all_versions_live_test
              Path: sdk/storage/platform-matrix-all-versions.json
              Selection: sparse
              GenerateVMJobs: true
```

benbp (changes requested):

sdk/storage/azure-storage-blob/tests/_shared/testcase.py
```diff
@@ -310,6 +313,32 @@ def generate_sas_token(self):
...
+class ApiVersionAssertPolicy(SansIOHTTPPolicy):
+    """
+    Assert the ApiVersion is set properly on the response
+    """
+
+    def __init__(self, api_version):
+        self.api_version = api_version
+
+    def on_request(self, request):
+        assert request.http_request.headers['x-ms-version'] == self.api_version
```
kasobol-msft, line 340: will this throw if env variable is empty? , we should either handle it here or not inject policy if there's no service version provided extrenally.

sdk/storage/azure-storage-queue/tests/_shared/testcase.py
```diff
@@ -310,13 +313,39 @@ def generate_sas_token(self):
     def generate_fake_token(self):
         return FakeTokenCredential()
 
+    def _get_service_version(self, **kwargs):
+        env_version = os.environ.get("AZURE_LIVE_TEST_SERVICE_VERSION", LATEST_SERVICE_VERSION)
+        return kwargs.pop("service_version", env_version)
```
kasobol-msft, line 318: Is there a way to not copy code between storage modules? "storage test shared" module would be great.
We have one in .NET, we started one in Java while doing similar changes. Maybe we could use this opportunity to start one in python?

seankane-msft: We can probably put this in `devtools_testutils`, I'm working on an improvement PR for this that includes code we have shared in tables and storage [here](https://github.com/Azure/azure-sdk-for-python/pull/19040)

sdk/storage/azure-storage-blob/tests/_shared/testcase.py
seankane-msft: No, if the env variable is empty it will default to `LATEST_SERVICE_VERSION`

kasobol-msft (approved):

sdk/storage/azure-storage-queue/tests/_shared/testcase.py
kasobol-msft: That'd be great. not a blocker here, but we should do something about all these clones in long term.

seankane-msft: After merging this, I can make changes for that

sdk/storage/tests.yml
```diff
@@ -26,6 +26,18 @@ stages:
...
+            AdditionalMatrixConfigs:
+              - Name: Storage_all_versions_live_test
+                Path: sdk/storage/platform-matrix-all-versions.json
+                Selection: sparse
+                GenerateVMJobs: true
+          - ${{ if not(contains(variables['Build.DefinitionName'], 'tests-weekly')) }}:
+            - Name: Storage_live_test
+              Path: sdk/storage/platform-matrix.json
+              Selection: sparse
+              GenerateVMJobs: true
```
seankane-msft, line 40:
```suggestion
```

sdk/storage/tests.yml
```diff
@@ -26,6 +26,12 @@ stages:
         MatrixReplace:
           # Use dedicated storage pool in canadacentral with higher memory capacity
           - Pool=(.*)-general/$1-storage
+        ${{ if contains(variables['Build.DefinitionName'], 'tests-weekly') }}:
+          AdditionalMatrixConfigs:
```
benbp:
```suggestion
          MatrixConfigs:
```

I forgot we want to override the base matrix, not add on to it.

benbp (dismissed):

kasobol-msft (approved):

PR #19049, opened 2021-06-02
[Key Vault] Revise administration models

### LINKED ISSUE #13575: Revise administration models
Decisions to make and implement:
- [ ] Prefer properties (`KeyVaultRoleAssignment`) or attributes (`_Operation`)?
    - properties have the benefit of easy immutability
- [ ] How to expose `_Operation.error`?
    - it matches the OData 4 error format, so we could raise it as [ODataV4Error](https://github.com/Azure/azure-sdk-for-python/blob/3808ad57daced04b57ff43018045b977105573ab/sdk/core/azure-core/azure/core/exceptions.py#L346)
- [x] ~Remove `SelectiveKeyRestoreOperation`? (it's identical in all but name to `RestoreOperation`)~
    - we decided to keep this as future-proofing
- [ ] `_Operation.start_time` and `.end_time` should be `datetime`

### LINKED ISSUE #18992: Successful restore operations may yield error
Managed HSM may return an "empty" error object in otherwise successful responses, e.g.:

```json
{
  "status": "complete",
  // ...
  "error": {
    "code": null,
    "innererror": null,
    "message": null
  }
}
```

This may yield an error depending on how the response is processed. To mitigate this possibility, check not only that the `error` object is null but that the [`error.code`](https://github.com/Azure/azure-rest-api-specs/blob/9875f8eebd8c1392a83c052bc50a79c8251b1174/specification/keyvault/data-plane/Microsoft.KeyVault/stable/7.2/common.json#L52-L72) is null before yielding an error.

mccoyp:
Resolves #13575 and resolves #18992.

### DIFF
--- a/sdk/keyvault/azure-keyvault-administration/azure/keyvault/administration/_models.py
+++ b/sdk/keyvault/azure-keyvault-administration/azure/keyvault/administration/_models.py
@@ -5,7 +5,7 @@
 from datetime import datetime
 from typing import TYPE_CHECKING
 
-from azure.core.exceptions import AzureError
+from azure.core.exceptions import ODataV4Format
 
 if TYPE_CHECKING:
     from typing import Any, Optional
@@ -256,7 +256,7 @@ class _Operation(object):
         if error is None or error.code is None:
             self._error = None
         else:
-            self._error = AzureError("Error: {}".format(error.message or self._status_details), error=error)
+            self._error = ODataV4Format(vars(error))
 
     @property
     def status(self):
@@ -272,7 +272,7 @@ class _Operation(object):
 
     @property
     def error(self):
-        # type: () -> Optional[AzureError]
+        # type: () -> Optional[ODataV4Format]
         """Error encountered, if any, during the operation"""
         return self._error
 
--- a/sdk/keyvault/azure-keyvault-administration/azure/keyvault/administration/_models.py
+++ b/sdk/keyvault/azure-keyvault-administration/azure/keyvault/administration/_models.py
@@ -2,10 +2,13 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
 # ------------------------------------
+from datetime import datetime
 from typing import TYPE_CHECKING
 
+from azure.core.exceptions import AzureError
+
 if TYPE_CHECKING:
-    from typing import Any
+    from typing import Any, Optional
 
 
 # pylint:disable=protected-access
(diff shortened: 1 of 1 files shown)


sdk/keyvault/azure-keyvault-administration/azure/keyvault/administration/_models.py
```diff
@@ -216,57 +245,77 @@ def _from_generated(cls, definition):
...
+        self._start_time = kwargs.get("start_time", None)
+        self._end_time = kwargs.get("end_time", None)
+        self._job_id = kwargs.get("job_id", None)
+
+        error = kwargs.get("error", None)
+        # responses can return an empty error object when there's no error
+        if error is None or error.code is None:
+            self._error = None
+        else:
+            self._error = AzureError("Error: {}".format(error.message or self._status_details), error=error)
```
chlowell, line 259: It's odd to use an exception type as a data object, and AzureError has an awkward API for this case. It won't document the structure of `error`, which it would expose as its `inner_exception`, which is misleading because `error` isn't an exception. Have you considered [ODataV4Format](https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/core/azure-core/azure/core/exceptions.py#L105)?

mccoyp: I did notice ODataV4Format, and I think that would be a better choice -- I just need to figure out how to provide the generated Error object we get back as a JSON object that ODataV4Format can understand

sdk/keyvault/azure-keyvault-administration/azure/keyvault/administration/_models.py
```diff
@@ -216,57 +245,77 @@ def _from_generated(cls, definition):
...
+        self._start_time = kwargs.get("start_time", None)
+        self._end_time = kwargs.get("end_time", None)
+        self._job_id = kwargs.get("job_id", None)
+
+        error = kwargs.get("error", None)
+        # responses can return an empty error object when there's no error
+        if error is None or error.code is None:
+            self._error = None
+        else:
+            self._error = ODataV4Format(vars(error))
```
mccoyp, line 259: Small but important detail here: ODataV4Format expects the inner error to be labeled as `innererror`, but the error we get back labels it as `inner_error` and doesn't get applied to the resulting ODataV4Format object. Does it make sense to edit the label to be `innererror`, or would that be forcing a square peg into a round hole?

chlowell: ODataV4Format expects JSON deserialized to a dict but the Error class generated by autorest has already deserialized the JSON and Pythonified "innererror" to "inner_error". What do you get from `error.as_dict()`? Also, `ODataV4Format.innererror` should be a JSON dict as well--it's JSON dicts all the way down.

mccoyp: `error.as_dict()` gives the same dictionary as `vars(dict)` -- we're still left with an `inner_error` key 😕

chlowell: Can you use the [`key_transformer` argument](https://github.com/Azure/msrest-for-python/blob/master/msrest/serialization.py#L278) to rename it?

mccoyp: Okay nice, that does work with something like:
```python
def my_key_transformer(key, attr_desc, value):
    if key == "inner_error":
        return ("innererror", value)
    return (key, value)
```
for errors with no inner error (`error.inner_error = None`) and errors with JSON dict inner errors (they get recursively renamed)

mccoyp: It looks like something like this would work as well:
```python
error.as_dict(key_transformer=lambda k, _, v: ("innererror", v) if k == "inner_error" else (k, v))
```

mccoyp:
Closing, since other changes that affect these are being done in #19099.

Issue #17914, opened 2021-04-08
[translation] bad target URL input is returning "Failed" instead of "ValidationFailed"


kristapratico:
This service inconsistency is failing our live tests. 

Service bug opened here: https://machinetranslation.visualstudio.com/MachineTranslation/_workitems/edit/120321

Discussion here: https://teams.microsoft.com/l/message/19:932da56130f3427abf190aa6585143f1@thread.skype/1617638094117?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=3e17dcb0-4257-4a30-b843-77f47f1d4121&parentMessageId=1617638094117&teamName=Azure%20SDK&channelName=Service%20-%20Language%20(Cognitive)&createdTime=1617638094117

PR #19090, opened 2021-06-03
[translation] unskip tests now that correct error status is returned

kristapratico:
Resolves https://github.com/Azure/azure-sdk-for-python/issues/17914

This was previously returning "Failed" for a bad target URL, but the service has fixed it to correctly return "ValidationFailed' in v1.0. Also looks like they updated the error code to be more explicit.


iscai-msft (approved):

maririos (approved):

maririos:
Thanks for verifying. I need to do the same in .NET

PR #19098, opened 2021-06-04
[Blob][STG78]Object Imutability Policy (Version Level Worm)

xiafu-msft:


tasherif-msft (approved):

gapra-msft (approved):
public API looks good. Please ask Kamil about the bundled vs unbundled immutabilitypolicy stuff

sdk/storage/azure-storage-blob/azure/storage/blob/_blob_client.py
```diff
@@ -1384,6 +1407,63 @@ def set_blob_metadata(self, metadata=None, **kwargs):
...
+        """The Set Immutability Policy operation sets the immutability policy on the blob.
+
+        .. versionadded:: 12.10.0
+            This operation was introduced in API version '2020-10-02'.
+
+        :keyword ~datetime.datetime immutability_policy_expiry_time:
+            Specifies the date time when the blobs immutability policy is set to expire.
+        :keyword immutability_policy_mode:
+            Specifies the immutability policy mode to set on the blob.
+        :paramtype immutability_policy_mode: ~azure.storage.blob.BlobImmutabilityPolicyMode or str
```
xiafu-msft, line 1422: Hi @kasobol-msft I think we need your thoughts here.
Gauri @gapra-msft and I were discussing about these two parameters. For Java and .Net these two parameters are grouped into a class called BlobImmutabilityPolicy, while in Python we don't have the overloading concern so usually we just list the parameters. Do you have any thoughts? Should we keep it consistent with Java and .Net or leave it like this?
Also Anna suggested to rename these two parameters to expiry and mode, and I put the reason why I haven't renamed them yet https://apiview.dev/Assemblies/Review/266426541eba45098d08bbddfb5ce8ac/7977909c50af4022868fd86d76bbc17c#azure.storage.blob.aio.BlobClient.set_immutability_policy:async.param(immutability_policy_mode

Thanks for your input!

kasobol-msft: I'd group them together in both places. (we seem to do this from time to time for inputs that belong together https://github.com/Azure/azure-sdk-for-python/blob/a65e8535376de8864b0c7b89f27b3da9d7e47182/sdk/storage/azure-storage-blob/azure/storage/blob/_blob_client.py#L538-L540 )

xiafu-msft: We can group them together just some params are not grouped together and they are all for conditional match
https://github.com/Azure/azure-sdk-for-python/blob/a65e8535376de8864b0c7b89f27b3da9d7e47182/sdk/storage/azure-storage-blob/azure/storage/blob/_blob_client.py#L514-L530

kasobol-msft (approved):

PR #19113, opened 2021-06-04
Fix directory walking for finding setup.py files.

weshaggard:
- Handle errors with parsing setup.py files by skipping
- Using os.walk because we still need to support py2.7
- Use abspath with os.walk to avoid depending on cwd
  as parse failures can leave the cwd in a different location
- Make sure we only try to parse setup.py files under `sdk/*/*/` paths


scbedd (approved):
Thanks @weshaggard

chidozieononiwu (approved):

PR #19117, opened 2021-06-04
[Tables] Make credential parameter keyword-only

annatisch:


sdk/tables/azure-data-tables/samples/async_samples/sample_insert_delete_entities_async.py
```diff
@@ -65,7 +65,7 @@ async def delete_entity(self):
         from azure.core.credentials import AzureNamedKeyCredential
 
         credential = AzureNamedKeyCredential(self.account_name, self.access_key)
-        table_client = TableClient(endpoint=self.endpoint, credential=credential, table_name=self.table_name)
+        table_client = TableClient(endpoint=self.endpoint, table_name=self.table_name, credential=credential)
```
seankane-msft, line 68: nit: does this have to be the last keyword?

annatisch: It doesn't, but given the `endpoint, table_name` are the two positional parameters - I thought for the samples it would be best to demonstrate the correct order. I didn't bother swapping them in the tests ;)

sdk/tables/azure-data-tables/tests/test_table.py
```diff
@@ -36,7 +36,7 @@ class StorageTableTest(AzureTestCase, TableTestCase):
     def test_create_properties(self, tables_storage_account_name, tables_primary_storage_account_key):
         # # Arrange
         account_url = self.account_url(tables_storage_account_name, "table")
-        ts = self.create_client_from_credential(TableServiceClient, tables_primary_storage_account_key, endpoint=account_url)
+        ts = TableServiceClient(credential=tables_primary_storage_account_key, endpoint=account_url)
```
seankane-msft, line 39: If this passes I'd keep it for now, but you can also change it to `credential=tables_primary_storage_account_key`. Small nit

seankane-msft (approved):
small nits

PR #19165, opened 2021-06-08
[EngSys] adding the word 'password' to suppressions

seankane-msft:


weshaggard (changes requested):
As mentioned offline lets try to avoid this if we can, but if there aren't any other good way we can consider do this after we get feedback from the credscan folks.

weshaggard (approved):
If this is suppressing the strings please go ahead with merging.

eng/CredScanSuppression.json
```diff
@@ -44,6 +44,12 @@
                 "tools/azure-sdk-tools/devtools_testutils/mgmt_settings_fake.py"
             ],
             "_justification": "File contains private key used by test code."
+        },
+        {
+            "hash":[
+                "sHdLtMfSDs4w/A83iF5BEf3+19x3is6z4yqFHtkxKLs="
+            ],
+            "_justification": "Password in docstrings"
```
weshaggard, line 52: For extra context can you add the string format for the docstrings that is being suppressed.

PR #19175, opened 2021-06-09
Added retry on Connection Errors

tasherif-msft:


sdk/storage/azure-storage-blob/azure/storage/blob/_shared/policies.py
```diff
@@ -501,6 +503,20 @@ def send(self, request):
...
+                    if retries_remaining:
+                        retry_hook(
+                            retry_settings,
+                            request=request.http_request,
+                            response=None,
+                            error=err)
+                        self.sleep(retry_settings, request.context.transport)
+                        continue
+                    raise err
             except AzureError as err:
```
xiafu-msft, line 520: Note: ServiceResponseError is an AzureError, so I think it is handled already. will sync offline about this

PR #19187, opened 2021-06-09
Enabling Before/After Test Customization for Standard CI Pipelines

scbedd:
This is to enable a feature that @iscai-msft is working on that requires a local `flask`-based test server to spin up.

[Example of this working in an internal build.](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=939566&view=logs&j=ae353b49-8f95-573e-27fe-def0c7b2e202&t=442859fe-9a61-56c2-8288-a0ea1a58f2cc)

And implementation:

image


iscai-msft (approved):
thanks so much @scbedd

PR #19233, opened 2021-06-13
[Service Bus]Match the partition key with session id only when both are set

bishnu-shb:
The azure documentation says "if a message has the partition key property but not the session ID property set, then Service Bus uses the partition key property value as the partition key. If the message has both the session ID and the partition key properties set, both properties must be identical" [https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-partitioning#using-a-partition-key]. 
[Re-raised #18955]

### DIFF
--- a/sdk/servicebus/azure-servicebus/CHANGELOG.md
+++ b/sdk/servicebus/azure-servicebus/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 **Bug Fixes**
 
-* Fixed a bug when setting the `partition_key` on `ServiceBusMessage` that should ensure that the session ID has been set before checking that the value passed in is equivalent to the session ID (PR #19233, thanks @bishnu-shb).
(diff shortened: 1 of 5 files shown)


swathipil:
Hey @bishnu-shb, do you mind signing the CLA above?

bishnu-shb:
CLA signed.

sdk/servicebus/azure-servicebus/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Release History
 
+## 7.3.1 (Unreleased)
+
+**Bug Fixes**
+
+* Fixed a bug when setting the `partition_key` on `ServiceBusMessage` that should ensure that the session ID has been set before checking that the value passed in is equivalent to the session ID (PR #19233, thanks @bishnu-shb).
+
```
yunhaoling, line 8: @swathipil , I would probably rephrase with something like:
```suggestion
* Fixed a bug on validation of setting `ServiceBusMessage.partition_key` that the value should be not validated against `session_id` of None (PR #19233, thanks @bishnu-shb).
```

swathipil (approved):

PR #19276, opened 2021-06-16
[Blob][Fileshare] Ensured that download fails if blob/file modified mid download

tasherif-msft:
Blobs:
Currently in blobs we lock on etag, and we try to let the user override this by passing `If-Match:'*'` but the logic was incorrect which just led to the ability to modify a blob during the second chunked call, otherwise it would still fail. I have removed this extra logic completely.

Fileshare:
Added locking on etag - previously there was no etag locking.


sdk/storage/azure-storage-file-share/azure/storage/fileshare/_download.py
```diff
@@ -162,6 +163,9 @@ def _download_chunk(self, chunk_start, chunk_end):
                 download_stream_current=self.progress_total,
                 **self.request_options
             )
+            if response.properties.etag != self.etag:
+                raise ValueError("The file has been modified while downloading.")
```
kasobol-msft, line 167: Please check with @annatisch if this is right exception type.

kasobol-msft (approved):
LGTM.

sdk/storage/azure-storage-file-share/azure/storage/fileshare/_download.py
kasobol-msft: In dotnet we throw [this](https://github.com/Azure/azure-sdk-for-net/blob/master/sdk/storage/Azure.Storage.Files.Shares/src/Models/ShareFileModifiedException.cs)  and provide more data than just message.

tasherif-msft: Confirmed with @xiafu-msft and we are now throwing `ResourceModifiedError` as per blob natively does in this case.

PR #19283, opened 2021-06-16
Reduce redundant logging during ChainedTokenCredential.get_token()

### LINKED ISSUE #18972: DefaultAzureCredential prints get_token failed even when it succeeds
- **Package Name**: azure-identity
- **Package Version**: 1.6.0
- **Operating System**: MacOS
- **Python Version**: 3.7.4

**Describe the bug**
When authenticating using the `DefaultAzureCredential` in a Jupyter Notebook, it will try several authentication mechanisms as described [here](https://azuresdkdocs.blob.core.windows.net/$web/python/azure-identity/1.6.0/azure.identity.html#azure.identity.DefaultAzureCredential), as expected.

But even if one of the authentication methods happen to succeed, you will see the messages for the authentication methods that didn't succeed before the final (successful) authentication method was attempted (which you don't get a success message).

So you may see a message like this, even if the authentication succeeded (in this case via Interactive Browser):
```
EnvironmentCredential.get_token failed: EnvironmentCredential authentication unavailable. Environment variables are not fully configured.
ManagedIdentityCredential.get_token failed: ManagedIdentityCredential authentication unavailable, no managed identity endpoint found.
SharedTokenCacheCredential.get_token failed: SharedTokenCacheCredential authentication unavailable. No accounts were found in the cache.
VisualStudioCodeCredential.get_token failed: Failed to get Azure user details from Visual Studio Code.
AzureCliCredential.get_token failed: Please run 'az login' to set up an account
```

This is **confusing** because the user may think that the authentication failed.

Ideally we should not show the failed attempts if one of them succeeded.
Or at least it should display which authentication method succeeded at the end...

**To Reproduce**
1. Authenticate with any Azure SDK client that accepts `DefaultAzureCredential` in a Jupyter Notebook
2. Don't pass environment variables for the `EnvironmentCredential` so that you see at least one `get_token failed` message even when one of the authentication methods succeeded.

**Expected behavior**
Don't show failed messages if one of the authentication methods succeeded.
Or show what was attempted but also show what succeeded.

**Related issues**
- azure-quantum: https://github.com/microsoft/qdk-python/issues/67

chlowell:
This PR closes #18972 by reducing redundant logging during `ChainedTokenCredential.get_token()` and, by inheritance, `DefaultAzureCredential.get_token()`. The rationale is that when `get_token()` fails for a credential chain, it logs an aggregate of the inner credentials' error messages, making those credentials' logging of same redundant. Similarly, when `get_token()` succeeds, users will seldom find error messages from unavailable inner credentials interesting. On the contrary, those messages can be confusing: "Why do I see half a dozen error messages when authentication succeeds?"

Here's the diff for DefaultAzureCredential failure given typical logging configuration (filtering `DEBUG` messages):
```diff
[INFO] No environment configuration found.
[INFO] ManagedIdentityCredential will use IMDS
-[WARNING] EnvironmentCredential.get_token failed: EnvironmentCredential authentication unavailable. Environment variables are not fully configured.
-[INFO] DefaultAzureCredential - EnvironmentCredential is unavailable
[INFO] No response from the IMDS endpoint.
-[WARNING] ImdsCredential.get_token failed: ManagedIdentityCredential authentication unavailable, no managed identity endpoint found.
-[WARNING] ManagedIdentityCredential.get_token failed: ManagedIdentityCredential authentication unavailable, no managed identity endpoint found.
-[INFO] DefaultAzureCredential - ManagedIdentityCredential is unavailable
-[WARNING] SharedTokenCacheCredential.get_token failed: SharedTokenCacheCredential authentication unavailable. No accounts were found in the cache.
-[INFO] DefaultAzureCredential - SharedTokenCacheCredential is unavailable
-[WARNING] VisualStudioCodeCredential.get_token failed: Failed to get Azure user details from Visual Studio Code.
-[INFO] DefaultAzureCredential - VisualStudioCodeCredential is unavailable
-[WARNING] AzureCliCredential.get_token failed: Please run 'az login' to set up an account
-[INFO] DefaultAzureCredential - AzureCliCredential is unavailable
-[WARNING] AzurePowerShellCredential.get_token failed: Please run "Connect-AzAccount" to set up account
-[INFO] DefaultAzureCredential - AzurePowerShellCredential is unavailable
[WARNING] DefaultAzureCredential failed to retrieve a token from the included credentials.
Attempted credentials:
        EnvironmentCredential: EnvironmentCredential authentication unavailable. Environment variables are not fully configured.
        ManagedIdentityCredential: ManagedIdentityCredential authentication unavailable, no managed identity endpoint found.
        SharedTokenCacheCredential: SharedTokenCacheCredential authentication unavailable. No accounts were found in the cache.
        VisualStudioCodeCredential: Failed to get Azure user details from Visual Studio Code.
        AzureCliCredential: Please run 'az login' to set up an account
        AzurePowerShellCredential: Please run "Connect-AzAccount" to set up account
```

Similarly, DefaultAzureCredential success:
```diff
[INFO] No environment configuration found.
[INFO] ManagedIdentityCredential will use IMDS
-[WARNING] EnvironmentCredential.get_token failed: EnvironmentCredential authentication unavailable. Environment variables are not fully configured.
-[INFO] DefaultAzureCredential - EnvironmentCredential is unavailable
[INFO] No response from the IMDS endpoint.
-[WARNING] ImdsCredential.get_token failed: ManagedIdentityCredential authentication unavailable, no managed identity endpoint found.
-[WARNING] ManagedIdentityCredential.get_token failed: ManagedIdentityCredential authentication unavailable, no managed identity endpoint found.
-[INFO] DefaultAzureCredential - ManagedIdentityCredential is unavailable
-[WARNING] SharedTokenCacheCredential.get_token failed: SharedTokenCacheCredential authentication unavailable. No accounts were found in the cache.
-[INFO] DefaultAzureCredential - SharedTokenCacheCredential is unavailable
-[INFO] DefaultAzureCredential - VisualStudioCodeCredential is unavailable
-[WARNING] AzureCliCredential.get_token failed: Please run 'az login' to set up an account
-[INFO] DefaultAzureCredential - AzureCliCredential is unavailable
-[INFO] AzurePowerShellCredential.get_token succeeded
[INFO] DefaultAzureCredential acquired a token from AzurePowerShellCredential
```

Inner credentials still log these messages, however they do so at `DEBUG` level when they're invoked by ChainedTokenCredential, using a ContextVar to determine whether that's the case. Unfortunately, ContextVar was added in Python 3.7. For Python 3.6, we have a [backport](https://pypi.org/project/contextvars/). The nearest equivalent available to 2.7 is `threading.local`, which isn't concurrency-safe. Another alternative that works everywhere is to pass around an internal keyword argument, but 🤮

In the end, because this isn't a functional or API change, I don't mind excluding 2.7. Similarly, if the backport is unacceptable, I observe that 3.6 will be EOL in 6 months. But that's just my opinion, please feel free to change my mind or propose alternative solutions 😄

### DIFF
--- a/sdk/identity/azure-identity/CHANGELOG.md
+++ b/sdk/identity/azure-identity/CHANGELOG.md
@@ -9,6 +9,10 @@
 ### Bugs Fixed
 
 ### Other Changes
+- Reduced redundant `ChainedTokenCredential` and `DefaultAzureCredential`
+  logging. On Python 3.7+, credentials invoked by these classes now log debug
+  rather than info messages.
+  ([#18972](https://github.com/Azure/azure-sdk-for-python/issues/18972))
 
 ## 1.7.0b2 (2021-07-08)
 ### Features Added
--- a/sdk/identity/azure-identity/azure/identity/_credentials/chained.py
+++ b/sdk/identity/azure-identity/azure/identity/_credentials/chained.py
@@ -71,16 +71,15 @@ class ChainedTokenCredential(object):
             except CredentialUnavailableError as ex:
                 # credential didn't attempt authentication because it lacks required data or state -> continue
                 history.append((credential, ex.message))
-                _LOGGER.info("%s - %s is unavailable", self.__class__.__name__, credential.__class__.__name__)
             except Exception as ex:  # pylint: disable=broad-except
                 # credential failed to authenticate, or something unexpectedly raised -> break
                 history.append((credential, str(ex)))
-                _LOGGER.warning(
+                _LOGGER.debug(
                     '%s.get_token failed: %s raised unexpected error "%s"',
                     self.__class__.__name__,
                     credential.__class__.__name__,
                     ex,
-                    exc_info=_LOGGER.isEnabledFor(logging.DEBUG),
+                    exc_info=True,
                 )
                 break
 
--- a/sdk/identity/azure-identity/azure/identity/_credentials/chained.py
+++ b/sdk/identity/azure-identity/azure/identity/_credentials/chained.py
@@ -7,6 +7,7 @@ import logging
 from azure.core.exceptions import ClientAuthenticationError
 
 from .. import CredentialUnavailableError
+from .._internal import within_credential_chain
 
 try:
     from typing import TYPE_CHECKING
@@ -61,6 +62,7 @@ class ChainedTokenCredential(object):
         :param str scopes: desired scopes for the access token. This method requires at least one scope.
         :raises ~azure.core.exceptions.ClientAuthenticationError: no credential in the chain provided a token
         """
+        within_credential_chain.set(True)
         history = []
         for credential in self.credentials:
             try:
@@ -83,6 +85,7 @@ class ChainedTokenCredential(object):
                 )
                 break
 
+        within_credential_chain.set(False)
         attempts = _get_error_message(history)
         message = self.__class__.__name__ + " failed to retrieve a token from the included credentials." + attempts
         _LOGGER.warning(message)
--- a/sdk/identity/azure-identity/azure/identity/_internal/__init__.py
+++ b/sdk/identity/azure-identity/azure/identity/_internal/__init__.py
@@ -14,6 +14,23 @@ from .._constants import EnvironmentVariables, KnownAuthorities
 if TYPE_CHECKING:
     from typing import Any, Optional
 
+try:
+    from contextvars import ContextVar
+
+    within_credential_chain = ContextVar("within_credential_chain", default=False)
+except ImportError:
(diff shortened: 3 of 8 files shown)


mccoyp (approved):
I think this is a reasonable solution -- at least, I can't think of a better one. I feel bad excluding 2.7, but I don't think the necessary work to support it would be justifiable when this is a cosmetic change, albeit a really nice one.

I do wonder if it would be good to include a note in the changelog that these changes are available on 3.6 if you pip install the backport

chlowell:
It turns out the backport doesn't work correctly in all cases and likely never will given the impending 3.6 EOL (see MagicStack/contextvars/issues/2). I'll stick with 3.7+ here and recommend upgrading Python to anyone running 3.6 who really wants this change.

Issue #18677, opened 2021-05-12
[EventHubs] Support for AMQPAnnotatedMessage


yunhaoling:
duplicate what we have for the AMQPAnnotatedMessage in service bus: https://github.com/Azure/azure-sdk-for-python/issues/18665, and introduce extra public fields on EventData

API Skeleton:

- new namespace as in the service bus
- new AMQPAnnotatedMessage class in the new namespace
- new body type enum/amqpheader/amqpproperties class in the new namespace
- send batch should support AMQPAnnotatedMessage
- EventData should have extra public properties:
  - Pull up ContentType, MesssageId and CorrelationId as top level properties

Action item:
- settle down on the service bus design
- prepare api proposal/gist and get arch approval
  - cross team discussion
    - whether we want a preview or just GA it
  - could be a simple one, as it's mostly the same as servicebus
- implement
  - copy&paste service bus impl and adapt to eventhub
  - add the required public properties to eventdata
- test & sample & doc

swathipil:
Have a draft PR out for this which includes the idempotent producer feature as well. However, it was decided in code complete week of June that this should not be a preview release. As a result, will need to de-couple the idempotent producer code from the AMQP Annotated Message code (which should be in the GA release). However, there is a discussion tomorrow about whether idempotent producer feature should be GA'ed. Will status after discussion.

PR #19287, opened 2021-06-17
[EventHub] only amqp annotated message

swathipil:
fixes: #18677

### DIFF
--- a/sdk/eventhub/azure-eventhub/CHANGELOG.md
+++ b/sdk/eventhub/azure-eventhub/CHANGELOG.md
@@ -6,15 +6,14 @@
 
 - Added support for sending AMQP annotated message which allows full access to the AMQP message fields.
   - Introduced new namespace `azure.eventhub.amqp`.
-  - `azure.eventhub.amqp.AMQPAnnotatedMessage` is now made public and could be instantiated for sending.
-- Added new enum class `azure.eventhub.amqp.AMQPMessageBodyType` to represent the body type of the message message which includes:
+- Added new enum class `azure.eventhub.amqp.AmqpMessageBodyType` to represent the body type of the message message which includes:
   - `DATA`: The body of message consists of one or more data sections and each section contains opaque binary data.
   - `SEQUENCE`: The body of message consists of one or more sequence sections and each section contains an arbitrary number of structured data elements.
   - `VALUE`: The body of message consists of one amqp-value section and the section contains a single AMQP value.
-- Introduced new class `azure.eventhub.amqp.AMQPAnnotatedMessage` for accessing low-level amqp message sections.
-- Introduced new classes `azure.eventhub.amqp.AMQPMessageHeader` and `azure.eventhub.amqp.AMQPMessageProperties` for accessing amqp header and properties.
-- Added new property `body_type` on `azure.eventhub.EventData` which returns `azure.eventhub.amqp.AMQPMessageBodyType`.
-- Added new read-only property `raw_amqp_message` on `azure.eventhub.EventData` which returns `azure.eventhub.amqp.AMQPAnnotatedMessage`.
+- Introduced new class `azure.eventhub.amqp.AmqpAnnotatedMessage` for accessing low-level amqp message sections which can be instantiated for sending.
+- Introduced new classes `azure.eventhub.amqp.AmqpMessageHeader` and `azure.eventhub.amqp.AmqpMessageProperties` for accessing amqp header and properties.
+- Added new property `body_type` on `azure.eventhub.EventData` which returns `azure.eventhub.amqp.AmqpMessageBodyType`.
+- Added new read-only property `raw_amqp_message` on `azure.eventhub.EventData` which returns `azure.eventhub.amqp.AmqpAnnotatedMessage`.
 
 ## 5.5.0 (2021-05-13)
 
--- a/sdk/eventhub/azure-eventhub/CHANGELOG.md
+++ b/sdk/eventhub/azure-eventhub/CHANGELOG.md
@@ -1,5 +1,21 @@
 # Release History
 
+## 5.6.0 (Unreleased)
+
+**New Features**
+
+- Added support for sending AMQP annotated message which allows full access to the AMQP message fields.
+  - Introduced new namespace `azure.eventhub.amqp`.
+  - `azure.eventhub.amqp.AMQPAnnotatedMessage` is now made public and could be instantiated for sending.
+- Added new enum class `azure.eventhub.amqp.AMQPMessageBodyType` to represent the body type of the message message which includes:
+  - `DATA`: The body of message consists of one or more data sections and each section contains opaque binary data.
+  - `SEQUENCE`: The body of message consists of one or more sequence sections and each section contains an arbitrary number of structured data elements.
+  - `VALUE`: The body of message consists of one amqp-value section and the section contains a single AMQP value.
+- Introduced new class `azure.eventhub.amqp.AMQPAnnotatedMessage` for accessing low-level amqp message sections.
+- Introduced new classes `azure.eventhub.amqp.AMQPMessageHeader` and `azure.eventhub.amqp.AMQPMessageProperties` for accessing amqp header and properties.
+- Added new property `body_type` on `azure.eventhub.EventData` which returns `azure.eventhub.amqp.AMQPMessageBodyType`.
+- Added new read-only property `raw_amqp_message` on `azure.eventhub.EventData` which returns `azure.eventhub.amqp.AMQPAnnotatedMessage`.
+
 ## 5.5.0 (2021-05-13)
 
 **New Features**
--- a/sdk/eventhub/azure-eventhub/CHANGELOG.md
+++ b/sdk/eventhub/azure-eventhub/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## 5.6.0 (Unreleased)
 
-**New Features**
+### Features Added
 
 - Added support for sending AMQP annotated message which allows full access to the AMQP message fields.
   - Introduced new namespace `azure.eventhub.amqp`.
--- a/sdk/eventhub/azure-eventhub/CHANGELOG.md
+++ b/sdk/eventhub/azure-eventhub/CHANGELOG.md
@@ -6,14 +6,14 @@
 
 - Added support for sending AMQP annotated message which allows full access to the AMQP message fields.
   - Introduced new namespace `azure.eventhub.amqp`.
-- Added new enum class `azure.eventhub.amqp.AmqpMessageBodyType` to represent the body type of the message message which includes:
-  - `DATA`: The body of message consists of one or more data sections and each section contains opaque binary data.
-  - `SEQUENCE`: The body of message consists of one or more sequence sections and each section contains an arbitrary number of structured data elements.
-  - `VALUE`: The body of message consists of one amqp-value section and the section contains a single AMQP value.
-- Introduced new class `azure.eventhub.amqp.AmqpAnnotatedMessage` for accessing low-level amqp message sections which can be instantiated for sending.
-- Introduced new classes `azure.eventhub.amqp.AmqpMessageHeader` and `azure.eventhub.amqp.AmqpMessageProperties` for accessing amqp header and properties.
-- Added new property `body_type` on `azure.eventhub.EventData` which returns `azure.eventhub.amqp.AmqpMessageBodyType`.
-- Added new read-only property `raw_amqp_message` on `azure.eventhub.EventData` which returns `azure.eventhub.amqp.AmqpAnnotatedMessage`.
+  - Added new enum class `azure.eventhub.amqp.AmqpMessageBodyType` to represent the body type of the message which includes:
+    - `DATA`: The body of message consists of one or more data sections and each section contains opaque binary data.
+    - `SEQUENCE`: The body of message consists of one or more sequence sections and each section contains an arbitrary number of structured data elements.
+    - `VALUE`: The body of message consists of one amqp-value section and the section contains a single AMQP value.
+  - Introduced new class `azure.eventhub.amqp.AmqpAnnotatedMessage` for accessing low-level amqp message sections which can be instantiated for sending.
+  - Introduced new classes `azure.eventhub.amqp.AmqpMessageHeader` and `azure.eventhub.amqp.AmqpMessageProperties` for accessing amqp header and properties.
+  - Added new property `body_type` on `azure.eventhub.EventData` which returns `azure.eventhub.amqp.AmqpMessageBodyType`.
+  - Added new read-only property `raw_amqp_message` on `azure.eventhub.EventData` which returns `azure.eventhub.amqp.AmqpAnnotatedMessage`.
 
 ## 5.5.0 (2021-05-13)
 
--- a/sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
+++ b/sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
@@ -193,14 +193,14 @@ class EventData(object):
     def _decode_non_data_body_as_str(self, encoding="UTF-8"):
         # type: (str) -> str
         # pylint: disable=protected-access
-        self.raw_amqp_message._message._body._encoding = encoding
         body = self.raw_amqp_message._message._body
         if self.body_type == AmqpMessageBodyType.VALUE:
             if not body.data:
                 return ""
             return str(decode_with_recurse(body.data, encoding))
 
-        return str(body)
+        seq_list = [d for seq_section in body.data for d in seq_section]
+        return str(decode_with_recurse(seq_list, encoding))
 
     def _to_outgoing_message(self):
         # type: () -> EventData
--- a/sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
+++ b/sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
@@ -30,7 +30,6 @@ from ._utils import (
     decode_with_recurse,
 )
 from ._constants import (
-    MESSAGE_PROPERTY_MAX_LENGTH,
     PROP_SEQ_NUMBER,
     PROP_OFFSET,
     PROP_PARTITION_KEY,
@@ -401,7 +400,6 @@ class EventData(object):
     @property
     def correlation_id(self):
         # type: () -> Optional[str]
-        # pylint: disable=line-too-long
         """The correlation identifier.
         Allows an application to specify a context for the message for the purposes of correlation, for example
         reflecting the MessageId of a message that is being replied to.
--- a/sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
+++ b/sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
@@ -43,6 +43,12 @@ from ._constants import (
     PROP_USER_ID,
     PROP_CREATION_TIME,
 )
+from .amqp import (
+    AmqpAnnotatedMessage,
+    AmqpMessageBodyType,
+    AmqpMessageHeader,
+    AmqpMessageProperties
+)
 
 if TYPE_CHECKING:
     import datetime
@@ -100,6 +106,14 @@ class EventData(object):
             self.message = Message(body)
         self.message.annotations = {}
         self.message.application_properties = {}
+
+        # Internal usage only for transforming AmqpAnnotatedMessage to outgoing ServiceBusMessage
+        self._raw_amqp_message = AmqpAnnotatedMessage(message=self.message)
+        self.message = self._raw_amqp_message._message  # pylint:disable=protected-access
+        self._raw_amqp_message.header = AmqpMessageHeader()
+        self._raw_amqp_message.properties = AmqpMessageProperties()
+        self._published_sequence_number = None
+        self._pending_published_sequence_number = None
 
     def __repr__(self):
         # type: () -> str
@@ -165,11 +179,25 @@ class EventData(object):
         """
         event_data = cls(body="")
         event_data.message = message
+        event_data._raw_amqp_message = AmqpAnnotatedMessage(message=message)
         return event_data
 
     def _encode_message(self):
         # type: () -> bytes
-        return self.message.encode_message()
+        # pylint: disable=protected-access
+        return self._raw_amqp_message._message.encode_message()
+
+    def _to_outgoing_message(self):
+        # type: () -> EventData
+        # pylint: disable=protected-access
+        self.message = self._raw_amqp_message._to_outgoing_amqp_message() # pylint:disable=protected-access
+        return self
+
+    @property
+    def raw_amqp_message(self):
+        # type: () -> AmqpAnnotatedMessage
+        """Advanced usage only. The internal AMQP message payload that is sent or received."""
+        return self._raw_amqp_message
 
     @property
     def sequence_number(self):
@@ -178,7 +206,18 @@ class EventData(object):
 
         :rtype: int
         """
-        return self.message.annotations.get(PROP_SEQ_NUMBER, None)
+        return self._raw_amqp_message.annotations.get(PROP_SEQ_NUMBER, None)
+
+    @property
+    def published_sequence_number(self):
+        # type: () -> Optional[int]
+        """
+        The publishing sequence number assigned to the event at the time it was successfully published.
+        If the producer was not configured to apply sequence numbering or if the event has not yet been successfully
+        published, the value will be None.
+        :rtype: int
+        """
+        return self._published_sequence_number
 
     @property
     def offset(self):
@@ -188,7 +227,7 @@ class EventData(object):
         :rtype: str
         """
         try:
-            return self.message.annotations[PROP_OFFSET].decode("UTF-8")
+            return self._raw_amqp_message.annotations[PROP_OFFSET].decode("UTF-8")
         except (KeyError, AttributeError):
             return None
 
@@ -199,7 +238,7 @@ class EventData(object):
 
         :rtype: datetime.datetime
         """
-        timestamp = self.message.annotations.get(PROP_TIMESTAMP, None)
+        timestamp = self._raw_amqp_message.annotations.get(PROP_TIMESTAMP, None)
         if timestamp:
             return utc_from_timestamp(float(timestamp) / 1000)
         return None
@@ -212,9 +251,9 @@ class EventData(object):
         :rtype: bytes
         """
         try:
-            return self.message.annotations[PROP_PARTITION_KEY_AMQP_SYMBOL]
+            return self._raw_amqp_message.annotations[PROP_PARTITION_KEY_AMQP_SYMBOL]
         except KeyError:
-            return self.message.annotations.get(PROP_PARTITION_KEY, None)
+            return self._raw_amqp_message.annotations.get(PROP_PARTITION_KEY, None)
 
     @property
     def properties(self):
@@ -223,7 +262,7 @@ class EventData(object):
 
         :rtype: dict
         """
-        return self.message.application_properties
+        return self._raw_amqp_message.application_properties
 
     @properties.setter
     def properties(self, value):
@@ -233,7 +272,7 @@ class EventData(object):
         :param dict value: The application properties for the EventData.
         """
         properties = None if value is None else dict(value)
-        self.message.application_properties = properties
+        self._raw_amqp_message.application_properties = properties
 
     @property
     def system_properties(self):
@@ -266,25 +305,25 @@ class EventData(object):
 
         if self._sys_properties is None:
             self._sys_properties = {}
-            if self.message.properties:
+            if self._raw_amqp_message.properties:
                 for key, prop_name in _SYS_PROP_KEYS_TO_MSG_PROPERTIES:
-                    value = getattr(self.message.properties, prop_name, None)
+                    value = getattr(self._raw_amqp_message.properties, prop_name, None)
                     if value:
                         self._sys_properties[key] = value
-            self._sys_properties.update(self.message.annotations)
+            self._sys_properties.update(self._raw_amqp_message.annotations)
         return self._sys_properties
 
     @property
     def body(self):
-        # type: () -> Union[bytes, Iterable[bytes]]
-        """The content of the event.
-
-        :rtype: bytes or Generator[bytes]
+        # type: () -> Any
+        """The body of the Message. The format may vary depending on the body type:
+        For ~azure.eventhub.AmqpMessageBodyType.DATA, the body could be bytes or Iterable[bytes]
+        For ~azure.eventhub.AmqpMessageBodyType.SEQUENCE, the body could be List or Iterable[List]
+        For ~azure.eventhub.AmqpMessageBodyType.VALUE, the body could be any type.
+        :rtype: Any
         """
-        try:
-            return self.message.get_data()
-        except TypeError:
-            raise ValueError("Event content empty.")
+
+        return self._raw_amqp_message.body
 
     def body_as_str(self, encoding="UTF-8"):
         # type: (str) -> str
@@ -392,9 +431,9 @@ class EventDataBatch(object):
             try:
                 self.add(event_data)
             except ValueError:
-                raise ValueError("The combined size of EventData collection exceeds the Event Hub frame size limit. "
-                                 "Please send a smaller collection of EventData, or use EventDataBatch, "
-                                 "which is guaranteed to be under the frame size limit")
+                raise ValueError("The combined size of EventData and AmqpAnnotatedMessage collection exceeds "
+                                 "the Event Hub frame size limit. Please send a smaller collection of EventData "
+                                 "or use EventDataBatch, which is guaranteed to be under the frame size limit")
 
     @property
     def size_in_bytes(self):
@@ -406,7 +445,7 @@ class EventDataBatch(object):
         return self._size
 
     def add(self, event_data):
-        # type: (EventData) -> None
+        # type: (Union[EventData, AmqpAnnotatedMessage]) -> None
         """Try to add an EventData to the batch.
 
         The total size of an added event is the sum of its body, properties, etc.
@@ -414,10 +453,16 @@ class EventDataBatch(object):
         be raised.
 
         :param event_data: The EventData to add to the batch.
-        :type event_data: ~azure.eventhub.EventData
+        :type event_data: Union[~azure.eventhub.EventData, ~azure.eventhub.amqp.AmqpAnnotatedMessage]
         :rtype: None
         :raise: :class:`ValueError`, when exceeding the size limit.
         """
+        # pylint:disable=protected-access
+        if isinstance(event_data, AmqpAnnotatedMessage):
+            event_data = EventData._from_message(event_data._to_outgoing_amqp_message())    # pylint: disable=protected-access
+        else:
+            event_data = event_data._to_outgoing_message()
+
         if self._partition_key:
             if (
                 event_data.partition_key
--- a/sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
+++ b/sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
@@ -28,6 +28,7 @@ from ._utils import (
     trace_message,
     utc_from_timestamp,
     transform_messages_if_needed,
+    transform_single_message_if_needed,
 )
 from ._constants import (
     MESSAGE_PROPERTY_MAX_LENGTH,
@@ -549,7 +550,7 @@ class EventDataBatch(object):
         :raise: :class:`ValueError`, when exceeding the size limit.
         """
 
-        outgoing_event_data = transform_messages_if_needed(event_data, EventData)
+        outgoing_event_data = transform_single_message_if_needed(event_data, EventData)
 
         if self._partition_key:
             if (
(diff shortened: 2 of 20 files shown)


sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
```diff
@@ -101,6 +107,14 @@ def __init__(self, body=None):
         self.message.annotations = {}
         self.message.application_properties = {}
 
+        # Internal usage only for transforming AmqpAnnotatedMessage to outgoing ServiceBusMessage
```
yunhaoling, line 110:
```suggestion
        # Internal usage only for transforming AmqpAnnotatedMessage to outgoing EventData
```

sdk/eventhub/azure-eventhub/CHANGELOG.md
```diff
@@ -1,5 +1,21 @@
...
+  - `azure.eventhub.amqp.AMQPAnnotatedMessage` is now made public and could be instantiated for sending.
+- Added new enum class `azure.eventhub.amqp.AMQPMessageBodyType` to represent the body type of the message message which includes:
+  - `DATA`: The body of message consists of one or more data sections and each section contains opaque binary data.
+  - `SEQUENCE`: The body of message consists of one or more sequence sections and each section contains an arbitrary number of structured data elements.
+  - `VALUE`: The body of message consists of one amqp-value section and the section contains a single AMQP value.
+- Introduced new class `azure.eventhub.amqp.AMQPAnnotatedMessage` for accessing low-level amqp message sections.
+- Introduced new classes `azure.eventhub.amqp.AMQPMessageHeader` and `azure.eventhub.amqp.AMQPMessageProperties` for accessing amqp header and properties.
+- Added new property `body_type` on `azure.eventhub.EventData` which returns `azure.eventhub.amqp.AMQPMessageBodyType`.
+- Added new read-only property `raw_amqp_message` on `azure.eventhub.EventData` which returns `azure.eventhub.amqp.AMQPAnnotatedMessage`.
```
yunhaoling, line 17: the "AMQP" in the class names should be "Amqp", e.g. `AMQPMessageBodyType` -> `AmqpMessageBodyType`

sdk/eventhub/azure-eventhub/CHANGELOG.md
```diff
@@ -1,5 +1,21 @@
 # Release History
 
+## 5.6.0 (Unreleased)
+
+**New Features**
+
+- Added support for sending AMQP annotated message which allows full access to the AMQP message fields.
+  - Introduced new namespace `azure.eventhub.amqp`.
+  - `azure.eventhub.amqp.AMQPAnnotatedMessage` is now made public and could be instantiated for sending.
```
yunhaoling, line 9: I don't think we need this line for EH? (IIRC there's no AMQPAnnotatedMessage in EH before)

sdk/eventhub/azure-eventhub/CHANGELOG.md
```diff
@@ -1,5 +1,21 @@
...
+- Added new enum class `azure.eventhub.amqp.AMQPMessageBodyType` to represent the body type of the message message which includes:
+  - `DATA`: The body of message consists of one or more data sections and each section contains opaque binary data.
+  - `SEQUENCE`: The body of message consists of one or more sequence sections and each section contains an arbitrary number of structured data elements.
+  - `VALUE`: The body of message consists of one amqp-value section and the section contains a single AMQP value.
+- Introduced new class `azure.eventhub.amqp.AMQPAnnotatedMessage` for accessing low-level amqp message sections.
```
yunhaoling, line 14: `could be instantiated for sending.` can be put here

sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
```diff
@@ -101,6 +107,14 @@ def __init__(self, body=None):
...
+        self._raw_amqp_message = AmqpAnnotatedMessage(message=self.message)
+        self.message = self._raw_amqp_message._message  # pylint:disable=protected-access
+        self._raw_amqp_message.header = AmqpMessageHeader()
+        self._raw_amqp_message.properties = AmqpMessageProperties()
+        self._published_sequence_number = None
+        self._pending_published_sequence_number = None
```
yunhaoling, line 116: those two are idempotent producer related, we can remove them
```suggestion
```

sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
```diff
@@ -178,7 +206,18 @@ def sequence_number(self):
...
+    @property
+    def published_sequence_number(self):
+        # type: () -> Optional[int]
+        """
+        The publishing sequence number assigned to the event at the time it was successfully published.
+        If the producer was not configured to apply sequence numbering or if the event has not yet been successfully
+        published, the value will be None.
+        :rtype: int
+        """
+        return self._published_sequence_number
```
yunhaoling, line 220: idempotent producer related, let's remove this
```suggestion
```

sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
```diff
@@ -266,25 +305,25 @@ def system_properties(self):
...
+        For ~azure.eventhub.AmqpMessageBodyType.SEQUENCE, the body could be List or Iterable[List]
+        For ~azure.eventhub.AmqpMessageBodyType.VALUE, the body could be any type.
+        :rtype: Any
         """
-        try:
-            return self.message.get_data()
-        except TypeError:
-            raise ValueError("Event content empty.")
+
+        return self._raw_amqp_message.body
```
yunhaoling, line 326: this seems like a breaking change -- previously error is raised but now the data would be passed to user directly, so their handling logic might fail.

we'd better check when `get_data` would raise TypeError (if in case of sequence/value data, or in the case `get_data` returns None etc.

then decide whether we could simply `return self._raw_amqp_message.body` or we want to keep the error behavior

sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
```diff
@@ -392,9 +431,9 @@ def _load_events(self, events):
             try:
                 self.add(event_data)
             except ValueError:
-                raise ValueError("The combined size of EventData collection exceeds the Event Hub frame size limit. "
-                                 "Please send a smaller collection of EventData, or use EventDataBatch, "
-                                 "which is guaranteed to be under the frame size limit")
+                raise ValueError("The combined size of EventData and AmqpAnnotatedMessage collection exceeds "
```
yunhaoling, line 434:
```suggestion
                raise ValueError("The combined size of EventData or AmqpAnnotatedMessage collection exceeds "
```

sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
```diff
@@ -406,18 +445,24 @@ def size_in_bytes(self):
...
 
         :param event_data: The EventData to add to the batch.
-        :type event_data: ~azure.eventhub.EventData
+        :type event_data: Union[~azure.eventhub.EventData, ~azure.eventhub.amqp.AmqpAnnotatedMessage]
         :rtype: None
         :raise: :class:`ValueError`, when exceeding the size limit.
         """
+        # pylint:disable=protected-access
+        if isinstance(event_data, AmqpAnnotatedMessage):
+            event_data = EventData._from_message(event_data._to_outgoing_amqp_message())    # pylint: disable=protected-access
```
yunhaoling, line 462: is it possible to add the `_to_outgoing_message` on `AmqpAnnotatedMessage` like what we have done in Service Bus?

https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/servicebus/azure-servicebus/azure/servicebus/amqp/_amqp_message.py#L296-L298

sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
```diff
@@ -101,6 +107,14 @@ def __init__(self, body=None):
         self.message.annotations = {}
         self.message.application_properties = {}
 
+        # Internal usage only for transforming AmqpAnnotatedMessage to outgoing ServiceBusMessage
+        self._raw_amqp_message = AmqpAnnotatedMessage(message=self.message)
+        self.message = self._raw_amqp_message._message  # pylint:disable=protected-access
+        self._raw_amqp_message.header = AmqpMessageHeader()
+        self._raw_amqp_message.properties = AmqpMessageProperties()
```
yunhaoling, line 114: we could rewrite the `__init__` to replace uamqp.message construction with directly the AmqpAnnotatedMessage construction -- AmqpAnnotatedMessage will read the message and construct again which wastes cycles.

yunhaoling (commented):
We also need to pull up ContentType, MesssageId and CorrelationId as top level properties on EventData

sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
swathipil: So I didn't add it because in the SB AmqpAnnotatedMessage `_to_outgoing_message`, you send in the `message` arg into `ServiceBusMessage`. However, `EventData` does not take keyword args, so we cannot pass in `message`. If we wanted to do this, we would need to allow for an optional `message` property on `EventData` and change the API. Thoughts?

sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
swathipil: looks like return type of `self._raw_amqp_message.body` is `Any`, so re-adding the exception

sdk/eventhub/azure-eventhub/azure/eventhub/_utils.py
```diff
@@ -254,3 +261,37 @@ def parse_sas_credential(credential):
...
+    except TypeError:
+        # AmqpAnnotatedMessage
+        # pylint: disable=protected-access
+        return message_type._from_message(message.to_outgoing_amqp_message())  # type: ignore
+    # best way to do below? is it needed?
+    #except:
+    #    raise TypeError(
+    #        "Only AmqpAnnotatedMessage or EventData instances are supported. "
+    #        "Received instead: {}".format(message.__class__.__name__)
+    #    )
```
yunhaoling, line 280: what's the current behavior if an object of non-EventData type is added to a batch?
wondering whether raising TypeError would be a breaking change.

sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
```diff
@@ -266,24 +292,26 @@ def system_properties(self):
...
+        :rtype: Any
         """
         try:
-            return self.message.get_data()
-        except TypeError:
+            return self._raw_amqp_message.body
+        except:
             raise ValueError("Event content empty.")
 
     def body_as_str(self, encoding="UTF-8"):
```
yunhaoling, line 317: could you check `body_as_str` and `body_as_json` behavior for body of data of multiple sections (e.g. sending a list of strings)/sequence/value?

sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
```diff
@@ -165,11 +178,24 @@ def _from_message(cls, message):
         """
         event_data = cls(body="")
         event_data.message = message
+        event_data._raw_amqp_message = AmqpAnnotatedMessage(message=message)
```
yunhaoling, line 181: My concern is that it waste cycles for constructing the ivar raw_amqp_message and message of EventData twice in the sending scenario.

I think we could update this method to take an optional raw_amqp_message kwargs so that in the sending case, we pass the both the  AMQPAnnoatedMessage and the AMQPAnnoatedMessage._to_outgoing_amqp_message

e.g. code in service bus.

https://github.com/Azure/azure-sdk-for-python/blob/3a471ada95c696fd2c3795f80a46cc381f977067/sdk/servicebus/azure-servicebus/azure/servicebus/amqp/_amqp_message.py#L296-L298

https://github.com/Azure/azure-sdk-for-python/blob/3a471ada95c696fd2c3795f80a46cc381f977067/sdk/servicebus/azure-servicebus/azure/servicebus/_common/message.py#L101-L104

sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
yunhaoling: I think we ultimately would need the kwargs in the constructor.

I'm checking the [design guideline](https://azure.github.io/azure-sdk/python_design.html#python-models-input-constructor)
> DO craft a constructor for models that are intended to be instantiated by a user (i.e. non-result types) with minimal required information and optional information as keyword-only arguments.

We're missing it in EventData (e.g. `properties` could be keyword argument in the constructor), but that's a different topic.

I'm good with leaving as it is now as it's an implementation detail, I feel adding kwargs to the constructor while users could not pass anything and there being no documentation is not friendly.

@annatisch , what about your opinions?

sdk/eventhub/azure-eventhub/azure/eventhub/_utils.py
swathipil: It results in an AttributeError. Since the exception above with AmqpAnnotatedMessage also uses `AttributeError`, would it be better to remove this? Or keep and `raise AttributeError("Only Amqp...")`?

sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
swathipil: If we do the above as you mentioned, then we would need to add kwargs to `EventData` constructor. Is this what we want to do @annatisch ? (I believe this also ties into the comment you made above:  https://github.com/Azure/azure-sdk-for-python/pull/19287#discussion_r656515493)

sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
```diff
@@ -296,6 +338,11 @@ def body_as_str(self, encoding="UTF-8"):
         """
         data = self.body
         try:
+            if self.body_type != AmqpMessageBodyType.DATA:
+                # pylint: disable=protected-access
+                self.raw_amqp_message._message._body._encoding = encoding
+                return str(self.raw_amqp_message._message._body)
```
swathipil, line 344: @Adam should I save the encoding here and restore it after calling str()?
```suggestion
                old_encoding = self.raw_amqp_message._message._body._encoding
                self.raw_amqp_message._message._body._encoding = encoding
                body_str = str(self.raw_amqp_message._message._body)
                self.raw_amqp_message._message._body._encoding = old_encoding
                return body_str
```

sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
swathipil: for now, just adding extra param to `EventData._from_message()`. Will think about `kwargs` later

sdk/eventhub/azure-eventhub/tests/unittest/test_event_data.py
```diff
@@ -90,18 +90,18 @@ def test_sys_properties():
 
 
 def test_event_data_batch():
-    batch = EventDataBatch(max_size_in_bytes=100, partition_key="par")
+    batch = EventDataBatch(max_size_in_bytes=110, partition_key="par")
```
swathipil, line 93: @yunhaoling the size of the batch needs to be updated. Since every EventData object now has `MessageProperties()` added in the constructor automatically (even if all properties on MessageProperties are `None`), it increases size of each EventData. Previously, `message.properties = None` not `MessageProperties`.

sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
```diff
@@ -154,22 +171,52 @@ def __str__(self):
...
         :rtype: ~azure.eventhub.EventData
         """
         event_data = cls(body="")
-        event_data.message = message
+        if raw_amqp_message and message:
+            event_data._raw_amqp_message = raw_amqp_message
+            event_data.message = message
+        elif message:
+            event_data._raw_amqp_message = AmqpAnnotatedMessage(message=message)
+            event_data.message = message
```
yunhaoling, line 190: nit: I think this could be improved as the following:
```suggestion
        event_data.message = message
        event_data._raw_amqp_message = raw_amqp_message if raw_amqp_message else AmqpAnnotatedMessage(message=message)
```

sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
```diff
@@ -154,22 +171,52 @@ def __str__(self):
...
-        return self.message.encode_message()
+        # pylint: disable=protected-access
+        return self._raw_amqp_message._message.encode_message()
+
+    def _decode_non_data_body_as_str(self, encoding="UTF-8"):
+        # type: (str) -> str
+        # pylint: disable=protected-access
+        self.raw_amqp_message._message._body._encoding = encoding
+        body = self.raw_amqp_message._message._body
+        if self.body_type == AmqpMessageBodyType.VALUE:
```
yunhaoling, line 203: what about the sequence type?

sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
```diff
@@ -154,22 +171,52 @@ def __str__(self):
...
     def _encode_message(self):
         # type: () -> bytes
-        return self.message.encode_message()
+        # pylint: disable=protected-access
+        return self._raw_amqp_message._message.encode_message()
+
+    def _decode_non_data_body_as_str(self, encoding="UTF-8"):
+        # type: (str) -> str
+        # pylint: disable=protected-access
+        self.raw_amqp_message._message._body._encoding = encoding
```
yunhaoling, line 201: I don't think we should mutate the _encoding attribute of the _body, it will affect the following up decode/encode operations on the message.

but I should ask first, why you want to update the `_encoding` attribute to the _body?

sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
```diff
@@ -16,14 +16,22 @@
     List,
     TYPE_CHECKING,
     cast,
+    Mapping,
```
yunhaoling, line 19: Is the import used?

sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
```diff
@@ -322,6 +381,82 @@ def body_as_json(self, encoding="UTF-8"):
...
+            return None
+        try:
+            return self._raw_amqp_message.properties.message_id.decode("UTF-8")
+        except (AttributeError, UnicodeDecodeError):
+            return self._raw_amqp_message.properties.message_id
+
+    @message_id.setter
+    def message_id(self, value):
+        # type: (str) -> None
+        if value and len(str(value)) > MESSAGE_PROPERTY_MAX_LENGTH:
```
yunhaoling, line 449: I guess the MESSAGE_PROPERTY_MAX_LENGTH is copied from the service bus, but I don't remember the eventhubs service has mentioned such restriction.

I would suggest testing setting message_id longer than MESSAGE_PROPERTY_MAX_LENGTH in event hubs, and drop the condition if it works with eh

sdk/eventhub/azure-eventhub/azure/eventhub/_producer.py
```diff
@@ -189,9 +190,10 @@ def _wrap_eventdata(
     ):
         # type: (...) -> Union[EventData, EventDataBatch]
         if isinstance(event_data, EventData):
```
yunhaoling, line 192: do we need to check if data is AmqpAnnotatedMessage here?
```suggestion
        if isinstance(event_data, EventData):
```

sdk/eventhub/azure-eventhub/azure/eventhub/_utils.py
```diff
@@ -245,12 +258,84 @@ def get_last_enqueued_event_properties(event_data):
...
+    parsed_sas = sas.split("&")
     expiry = None
     for item in parsed_sas:
-        if item.startswith('se='):
+        if item.startswith("se="):
             expiry = int(item[3:])
     return (sas, expiry)
+
+
+def transform_single_message_if_needed(message, message_type):
```
yunhaoling, line 273: I think we should probably rename this method just to `transform_outbound_single_message`.

`if_needed` doesn't reflect the thing we're doing here

sdk/eventhub/azure-eventhub/azure/eventhub/_utils.py
```diff
@@ -245,12 +258,84 @@ def get_last_enqueued_event_properties(event_data):
...
+    1. update the internal messages to reflect any updates to settable properties on EventData
+    2. transform the AmqpAnnotatedMessage to be EventData
+    :param MessagesType messages: A list or single instance of messages of type EventData 
+        or AmqpAnnotatedMessage.
+    :param Type[EventData] message_type: The class type to return the messages as.
+    :rtype: Union[EventData, List[EventData]]
+    """
+    if isinstance(messages, Iterable):
+        return [transform_single_message_if_needed(m, message_type) for m in messages]
+    return transform_single_message_if_needed(messages, message_type)
```
yunhaoling, line 310: do we need this method any more?

sdk/eventhub/azure-eventhub/tests/livetest/asynctests/test_send_async.py
```diff
@@ -12,8 +12,135 @@
...
+    AmqpMessageBodyType,
+    AmqpAnnotatedMessage,
+    AmqpMessageProperties,
+)
+
+#@pytest.mark.liveTest
+#@pytest.mark.asyncio
+#async def test_send_amqp_annotated_message(connstr_receivers):
+#    connection_str, receivers = connstr_receivers
+#    received_count = {}
```
yunhaoling, line 28: the test should be uncommented?

sdk/eventhub/azure-eventhub/azure/eventhub/_utils.py
```diff
@@ -245,12 +258,84 @@ def get_last_enqueued_event_properties(event_data):
...
+        for k,v in data.items():
+            decoded_key = decode_with_recurse(k, encoding)
+            decoded_val = decode_with_recurse(v, encoding)
+            decoded_mapping[decoded_key] = decoded_val
+        return decoded_mapping
+    if isinstance(data, Iterable):
+        decoded_list = []
+        for d in data:
+            decoded_list.append(decode_with_recurse(d, encoding))
+        return decoded_list
```
yunhaoling, line 339: I'm thinking of cases when data of type integer, float, boolean, etc. is passed but it looks they are not handled.
probably we just need to handle it like this?
```python
if isinstance(data, Mapping):
    #recursive
elif isinstance(data, Iterable):
    #recursive
elif if isinstance(data, str):

else:
    return data
```

yunhaoling (commented):
looks good to me overall, let some minor comments
don't forget to generate an apiview for our arch to review :)

sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
swathipil: sequence type returns the same for python 2.7 and 3+, so we don't need to modify the data

sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
swathipil: good point! I was thinking about whether to store the original encoding and reset it. The reason I update the encoding is because when we call `str(body)` for sequence body, it uses `body._encoding` to decode. Do you think this is needed @yunhaoling?

sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
swathipil: dropped

sdk/eventhub/azure-eventhub/azure/eventhub/_producer.py
swathipil: no, I don't think it's needed since this method is only called in `send` in the same file, which is called by `send_batch`, which transforms `AmqpAnnotatedMessage` to `EventData` by the call to `send`. Since customers are only directed to use `send_batch` to send, they shouldn't call `send`.

sdk/eventhub/azure-eventhub/azure/eventhub/_utils.py
swathipil: good catch, just removed

sdk/eventhub/azure-eventhub/azure/eventhub/_utils.py
swathipil: so for integer, float, boolean, we should just be returning them as is. the thing that you did above will result in recursion error b/c string is actually `Iterable`. If instead of what you said above, we did:
```
if isinstance(data, str):
    return data
elif isinstance(data, Mapping):
    #recursive
elif isinstance(data, Iterable):
    #recursive
elif if isinstance(data, six.binary_type):
    return data.decode(encoding)
return str(data)
```
This would result in the types that you mentioned being turned to string before returning. So passing in `"[1, True, {b'key': ['hi', 4.5]}]"` would result in `['1', 'True', {'key': ['hi', '4.5']}]`. This probably isn't what the user would be expecting. By returning `data` if it doesn't fall under any of the conditions, and calling `str(data)` on the returned object in `_decode_non_body_data()`, this would result in `"[1, True, {'key': ['hi', 4.5]}]"`.

sdk/eventhub/azure-eventhub/tests/livetest/asynctests/test_send_async.py
```diff
@@ -12,8 +12,135 @@
...
+#        on_event.received = []
+#        on_event.app_prop = None
+#        consumer = EventHubConsumerClient.from_connection_string(connection_str,
+#                                                            consumer_group='$default')
+#        async with consumer:
+#            task = asyncio.ensure_future(consumer.receive(on_event,
+#                                                        partition_id="0", starting_position="-1"))
+#            await asyncio.sleep(10)
+#            for event in on_event.received:
+#                received_count = check_values(event, received_count)
```
yunhaoling, line 135:
```suggestion
#                check_values(event)
```

sdk/eventhub/azure-eventhub/tests/livetest/asynctests/test_send_async.py
```diff
@@ -12,8 +12,135 @@
...
+#@pytest.mark.asyncio
+#async def test_send_amqp_annotated_message(connstr_receivers):
+#    connection_str, receivers = connstr_receivers
+#    received_count = {}
+#    received_count["recv_data_msg"] = 0
+#    received_count["recv_sequence_msg"] = 0
+#    received_count["recv_value_msg"] = 0
+#    received_count["normal_msg"] = 0
+#
+#    def check_values(event, received_count):
```
yunhaoling, line 34:
```suggestion
#    def check_values(event):
```

sdk/eventhub/azure-eventhub/tests/livetest/asynctests/test_send_async.py
```diff
@@ -12,8 +12,135 @@
...
+#            received_count["recv_sequence_msg"] += 1
+#        elif raw_amqp_message.body_type == AmqpMessageBodyType.VALUE:
+#            assert raw_amqp_message.body == value_body
+#            assert event.body_as_str() == "{'key': [-123, 'data', False]}"
+#            # uncomment below after discussing with service team
+#            #assert raw_amqp_message.header.priority == 10
+#            assert raw_amqp_message.annotations[b'ann_key'] == b'ann_value'
+#            assert raw_amqp_message.application_properties[b'body_type'] == b'value'
+#            received_count["recv_value_msg"] += 1
+#        return received_count
```
yunhaoling, line 67:
```suggestion
```

sdk/eventhub/azure-eventhub/tests/livetest/asynctests/test_send_async.py
```diff
@@ -12,8 +12,135 @@
...
+#        await producer.send_batch(batch)
+#        await producer.send_batch([data_message, value_message, sequence_message, event_data])
+#
+#        on_event.received = []
+#        on_event.app_prop = None
+#        consumer = EventHubConsumerClient.from_connection_string(connection_str,
+#                                                            consumer_group='$default')
+#        async with consumer:
+#            task = asyncio.ensure_future(consumer.receive(on_event,
+#                                                        partition_id="0", starting_position="-1"))
```
yunhaoling, line 132:
```suggestion
#            task = asyncio.ensure_future(consumer.receive(on_event,
#                                                        starting_position="-1"))
```

sdk/eventhub/azure-eventhub/tests/livetest/asynctests/test_send_async.py
```diff
@@ -12,8 +12,135 @@
...
+#            assert event.body_as_str() == "{'key': [-123, 'data', False]}"
+#            # uncomment below after discussing with service team
+#            #assert raw_amqp_message.header.priority == 10
+#            assert raw_amqp_message.annotations[b'ann_key'] == b'ann_value'
+#            assert raw_amqp_message.application_properties[b'body_type'] == b'value'
+#            received_count["recv_value_msg"] += 1
+#        return received_count
+#
+#    async def on_event(partition_context, event):
+#        await on_event.received.append(event)
```
yunhaoling, line 70:
```suggestion
#        on_event.received.append(event)
```

sdk/eventhub/azure-eventhub/tests/livetest/asynctests/test_send_async.py
```diff
@@ -12,8 +12,135 @@
...
+#        async with consumer:
+#            task = asyncio.ensure_future(consumer.receive(on_event,
+#                                                        partition_id="0", starting_position="-1"))
+#            await asyncio.sleep(10)
+#            for event in on_event.received:
+#                received_count = check_values(event, received_count)
+#
+#        await task
+#
+#        assert received_count["normal_msg"] == 2 # receiving 0?
```
yunhaoling, line 139:
```suggestion
#        assert len(on_event.received) == 8
#        assert received_count["normal_msg"] == 2 # receiving 0?
```

sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
swathipil: updated to take seq based on discussion: seq should return string of list, not just string of all concatenated values

sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
swathipil: updated to no longer mutate _encoding

sdk/eventhub/azure-eventhub/CHANGELOG.md
```diff
@@ -1,5 +1,20 @@
 # Release History
 
+## 5.6.0 (Unreleased)
+
+### Features Added
+
+- Added support for sending AMQP annotated message which allows full access to the AMQP message fields.
+  - Introduced new namespace `azure.eventhub.amqp`.
+- Added new enum class `azure.eventhub.amqp.AmqpMessageBodyType` to represent the body type of the message message which includes:
```
yunhaoling, line 9:
```suggestion
- Added new enum class `azure.eventhub.amqp.AmqpMessageBodyType` to represent the body type of the message which includes:
```

sdk/eventhub/azure-eventhub/CHANGELOG.md
```diff
@@ -1,5 +1,20 @@
...
+  - `DATA`: The body of message consists of one or more data sections and each section contains opaque binary data.
+  - `SEQUENCE`: The body of message consists of one or more sequence sections and each section contains an arbitrary number of structured data elements.
+  - `VALUE`: The body of message consists of one amqp-value section and the section contains a single AMQP value.
+- Introduced new class `azure.eventhub.amqp.AmqpAnnotatedMessage` for accessing low-level amqp message sections which can be instantiated for sending.
+- Introduced new classes `azure.eventhub.amqp.AmqpMessageHeader` and `azure.eventhub.amqp.AmqpMessageProperties` for accessing amqp header and properties.
+- Added new property `body_type` on `azure.eventhub.EventData` which returns `azure.eventhub.amqp.AmqpMessageBodyType`.
+- Added new read-only property `raw_amqp_message` on `azure.eventhub.EventData` which returns `azure.eventhub.amqp.AmqpAnnotatedMessage`.
```
yunhaoling, line 16: should those be indented one more level as the sub section?

sdk/eventhub/azure-eventhub/azure/eventhub/_constants.py
```diff
@@ -35,6 +35,7 @@
 TIMEOUT_SYMBOL = b"com.microsoft:timeout"
 RECEIVER_RUNTIME_METRIC_SYMBOL = b"com.microsoft:enable-receiver-runtime-metric"
 
+MESSAGE_PROPERTY_MAX_LENGTH = 128
```
yunhaoling, line 38: I think we could remove this

sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
```diff
@@ -22,8 +22,15 @@
...
-from ._utils import set_message_partition_key, trace_message, utc_from_timestamp
+from ._utils import (
+    set_message_partition_key,
+    trace_message,
+    utc_from_timestamp,
+    transform_outbound_single_message,
+    decode_with_recurse,
+)
 from ._constants import (
+    MESSAGE_PROPERTY_MAX_LENGTH,
```
yunhaoling, line 33: I think we could remove this?

sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
```diff
@@ -322,6 +376,74 @@ def body_as_json(self, encoding="UTF-8"):
...
+    def content_type(self, value):
+        # type: (str) -> None
+        if not self._raw_amqp_message.properties:
+            self._raw_amqp_message.properties = AmqpMessageProperties()
+        self._raw_amqp_message.properties.content_type = value
+
+    @property
+    def correlation_id(self):
+        # type: () -> Optional[str]
+        # pylint: disable=line-too-long
```
yunhaoling, line 404: nit: is the pylint needed here?
if it's for suppressing the doc, than we could update the comment to start from a new line

sdk/eventhub/azure-eventhub/samples/async_samples/send_and_receive_amqp_annotated_message_async.py
```diff
@@ -0,0 +1,118 @@
+#!/usr/bin/env python
+
+# --------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for license information.
+# --------------------------------------------------------------------------------------------
+
+"""
+Example to show sending, receiving and parsing amqp annotated message(s) to an Event Hub partition.
```
yunhaoling, line 9: nit: not to a partition as in the code we don't specify a partition to send/receive
```suggestion
Example to show sending, receiving and parsing amqp annotated message(s) to Event Hubs.
```

sdk/eventhub/azure-eventhub/samples/sync_samples/send_and_receive_amqp_annotated_message.py
```diff
@@ -0,0 +1,111 @@
+#!/usr/bin/env python
+
+# --------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for license information.
+# --------------------------------------------------------------------------------------------
+
+"""
+Example to show sending, receiving and parsing amqp annotated message(s) to an Event Hub partition.
```
yunhaoling, line 9:
```suggestion
Example to show sending, receiving and parsing amqp annotated message(s) to Event Hubs.
```

sdk/eventhub/azure-eventhub/tests/livetest/asynctests/test_send_async.py
```diff
@@ -12,8 +12,116 @@
...
+                                                        consumer_group='$default')
+    async with client:
+        task = asyncio.ensure_future(client.receive(on_event, starting_position="-1"))
+        await asyncio.sleep(15)
+        for event in on_event.received:
+            check_values(event)
+
+    await task
+
+    assert len(on_event.received) == 8
```
yunhaoling, line 123: sorry for not making it clear, I actually like the previous part where we have dictionary for checking count of different message types, it validates that we receive and parse the right type of messages? could we add it back :)

(also the sync one)

sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
swathipil: no just forgot to remove it, good catch

sdk/eventhub/azure-eventhub/samples/async_samples/send_and_receive_amqp_annotated_message_async.py
```diff
@@ -0,0 +1,118 @@
+#!/usr/bin/env python
+
+# --------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for license information.
+# --------------------------------------------------------------------------------------------
+
+"""
+Example to show sending, receiving and parsing amqp annotated message(s) to an Event Hubs.
```
yunhaoling, line 9: nit:
```suggestion
Example to show sending, receiving and parsing amqp annotated message(s) to Event Hubs.
```

yunhaoling (approved):
LGTM

yunhaoling (approved):

PR #19302, opened 2021-06-17
[Tables] Add __repr__ and __str__ methods to models

### LINKED ISSUE #19278: [Tables] Add __repr__ methods to custom models
This will help the debugging experience

seankane-msft:
closes #19278 

Will help make debugging easier for us and customers


sdk/tables/azure-data-tables/azure/data/tables/_models.py
```diff
@@ -191,6 +221,14 @@ def _from_generated(cls, generated, **kwargs):  # pylint: disable=unused-argumen
             days=generated.days,
         )
 
+    def __str__(self):
+        # type: () -> str
+        return u"enabled={}, days={}".format(self.enabled, self.days)
```
annatisch, line 226: `__str__` is not supposed to return unicode. Must be type `str`

annatisch: Not sure whether the same applied to `__repr__` - would need to check the Py2.7 stdlib ref docs.

sdk/tables/azure-data-tables/azure/data/tables/_models.py
```diff
@@ -617,8 +682,13 @@ def __init__(self, **kwargs):
         )
 
     def __str__(self):
+        # type: () -> str
         return self._str
 
+    def __repr__(self):
+        # type: () -> str
+        return u"AccountSasPermissions(_str={})".format(self._str)
```
annatisch, line 690: I wouldn't document `_str` - this isn't supposed to be public.
I would probably either document it as just the string:
`return "AccountSasPermissions({})".format(self._str)`
Or you could go fully verbose:
`return "AccountSasPermissions(read={}, write={}, ...)".format(self.read, self.write, ...)`

sdk/tables/azure-data-tables/azure/data/tables/_models.py
```diff
@@ -546,8 +606,13 @@ def __init__(self, **kwargs):
         self._str = ("s" if self.service else "") + ("o" if self.object else "")
 
     def __str__(self):
+        # type: () -> str
         return self._str
 
+    def __repr__(self):
+        # type: () -> str
+        return u"<ResourceTypes(_str={})".format(self.__str__())
```
annatisch, line 614: Sneaky angle bracket at the front there.

sdk/tables/azure-data-tables/azure/data/tables/_models.py
```diff
@@ -153,6 +173,16 @@ def _from_generated(cls, generated):
             )
         )
 
+    def __str__(self):
+        # type: () -> str
+        return "version={}, enabled={}, include_apis={}, retention_policy={}".format(
+            self.version, self.enabled, self.include_apis, self.retention_policy
+        )
```
annatisch, line 180: I'm thinking maybe we just don't define `__str__` for the time being. Formatting with an instance will default to using `__repr__` in the meantime anyway.

sdk/tables/azure-data-tables/azure/data/tables/_models.py
```diff
@@ -153,6 +173,16 @@ def _from_generated(cls, generated):
...
+
+    def __repr__(self):
+        # type: () -> str
+        return "TableMetrics({})".format(self.__str__())
```
annatisch, line 184: Even though it's pretty safe to assume we'll never hit it for these models, we tend to add `[:1024]` to the end.

sdk/tables/azure-data-tables/azure/data/tables/_models.py
seankane-msft: Went with the latter for consistency

sdk/tables/azure-data-tables/azure/data/tables/_models.py
seankane-msft: I went with this suggestion and removed `__str__` methods for now, we can always add them back later.

iscai-msft (approved):

PR #19311, opened 2021-06-17
[Key Vault] Update test-resources.json for new MHSM API

mccoyp:
Due to a service change, deployments with the current ARM template now fail. This updates the resource API version and adds new required fields.


chlowell (approved):
If you push this to a feature branch, you can run the pipeline on that before merging to master

Issue #19087, opened 2021-06-03
Cannot Create EH compatible endpoint if IoT Hub name is more than 25 chars


swathipil:
IoT Hub Service is silently truncating the IoT Hub name in the Event Hub compatible endpoint to 25 characters. This is not documented anywhere. 

This silent mutation of information passed in by the user should not be happening. Since this is a service issue, we have contacted the IoT team to let them know about the bug, so that they can either fix it or document it.

A PR was created to address this issue in EH: #18521 
We decided not to accept the PR since it would also mean silently changing information passed in by the user.

TODO:
- [x] poke the service team again about this issue

swathipil:
After discussion with Jesse + Sherry Shu, it was found that C# was no longer able to reproduce this error. In Python, realized that we need to get the IoT Hub name from the redirection address.

PR #19314, opened 2021-06-17
[EventHubs] Get IoT Hub Name from Redirect Address in sample

swathipil:
Fixes: #19087 
Follows[ C# sample,](https://github.com/Azure/azure-sdk-for-net/blob/350c23ea19ed76d74f7d97dfe3aec82671fdcc9d/samples/iothub-connect-to-eventhubs/IotHubConnection.cs#L97) which also gets IoT Hub name from redirect address.


jsquire (approved):
The approach looks correct to me.

yunhaoling (commented):
this looks amazing :)
could you add a short comment to help understand the expression?
also I believe it is tested? (e.g. names longer than 25 chars)

PR #19335, opened 2021-06-18
[Key Vault] Support for API version 7.3-preview

mccoyp:


scbedd:
@mccoyp  I think the triggers are just a bit delayed here. The `keyvault - ci` job itself is updated properly afaict.

scbedd:
Yep. It's going! You're g2g.

mccoyp:
Closing now that actual implementations of SKR are up (Python's: #20048).

PR #19401, opened 2021-06-22
[Blob][Sync]Let user opt in decompress or not

xiafu-msft:


sdk/storage/azure-storage-blob/tests/test_get_blob.py
```diff
@@ -144,6 +146,29 @@ def test_get_blob_to_bytes(self, resource_group, location, storage_account, stor
...
+        data = self.get_random_bytes(1024*1024*1024)
+
+        with gzip.open(file_path, 'wb') as stream2:
+            stream2.write(data)
+
+        with open(file_path, 'rb') as stream:
+            data = stream.read()
+            blob.upload_blob(data, content_settings=ContentSettings(content_encoding='gzip'), overwrite=True)
+
+        content = blob.download_blob(offset=1024, length=512, decompress=False).readall()
```
xiafu-msft, line 167: I will add doc about decompress later

sdk/storage/azure-storage-file-share/tests/test_get_file.py
```diff
@@ -135,6 +136,28 @@ def test_unicode_get_file_binary_data(self, resource_group, location, storage_ac
...
+    def test_get_file_in_raw_stream(self, resource_group, location, storage_account, storage_account_key):
+        # recording mechanism cannot detect the decompress option properly, it always try to decompress the body
+        # as long as the content-encoding header is in response, so mark this as live test only
+        self._setup(storage_account, storage_account_key)
+        file_name = self._get_file_reference()
+        file_client = self.fsc.get_share_client(self.share_name).get_file_client(file_name)
+        file_path = os.path.abspath(os.path.join(os.path.abspath(__file__), "..", "./resources/testgzip.txt.gz"))
+        data = self.get_random_bytes(1024 * 1024 * 1024)
+
+        with gzip.open(file_path, 'wb') as stream2:
```
tasherif-msft, line 150: just remove the random data since file has content

sdk/storage/azure-storage-blob/azure/storage/blob/_blob_client.py
```diff
@@ -805,6 +805,8 @@ def download_blob(self, offset=None, length=None, **kwargs):
             The number of parallel connections with which to download.
         :keyword str encoding:
             Encoding to decode the downloaded bytes. Default is None, i.e. no decoding.
+        :keyword bool decompress:
+            This is True by default. Set it to False to disable the blob content auto-decompress.
```
tasherif-msft, line 809: Let's improve the description by mentioning the terms "encoded/zipped files" or something along those lines

tasherif-msft (approved):

sdk/storage/azure-storage-blob/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## 12.9.0b1 (Unreleased)
 
+## 12.8.2 (2021-06-25)
+**Fixes**
+- Fixed the auto decompress response body bug, and let user opt in decompress or not.
```
kasobol-msft, line 7: is it a bug or new functionality?
it seems that decompress is true by default - should this be "Added ability to opt-out from decompression" ?

xiafu-msft: great question! I feel it's more like a bug though there's api change... because currently if the blob/file content is gzipped and content size is larger than 32MB, user are never be able to download the content back..... changing the api is a workaround to fix this bug, the ideal behavior is to download whatever uploaded (decompress is ideally set to False), while switching decompress to True will cause a breaking change, because for those blobs whose size is less or equal to 32MB are able to download the blob unzipped.
Let me know if we want to make it a feature.

xiafu-msft: regarding this `it seems that decompress is true by default - should this be "Added ability to opt-out from decompression" ?`
sounds great, I will rephrase it in the way you recommended

kasobol-msft: Given that extra context. I agree this is a bug.
I also suggest we internally force underlying http client to not decompress without adding this option to public surface.

xiafu-msft: Then It think it will be a breaking change for the users whose blob is less than 32MB and relying on the auto decompress feature.

kasobol-msft: That'd be users who accidentally discovered this behavior, i.e. users who uploaded already compressed bytes and set content-encoding header and then expecting uncompressed payload to get back from storage. This is quite a corner case - do we believe such user exist? I'd rather expect them to fill a bug rather than leverage this - they must be surprised as well.

kasobol-msft: @annatisch what do you think ?

kasobol-msft: Imo we should fix this trap and don't decompress by default. If by any chance users leveraging that exists we may consider having a flag for them and documenting the behavior change.

annatisch: Thanks @kasobol-msft 
We should keep the current behaviour and auto-decompress by default is the content-encoding header is present.
This is the standard behaviour across Python HTTP libraries, and would be the expectation of a typical Python customer.
Granted, this default is different to the other languages - but is idiomatic to Python.

kasobol-msft: @annatisch I'm not sure if this is expected/idiomatic. When I look at https://github.com/Azure/azure-storage-python/issues/509 and referenced issues it seems that users are surprised that high level API is attempting to decompress blob and are putting workaround in place.
They don't interact with low level HTTP libraries but our client API. Why low level component we use internally should dictate how our download api behaves? 
Besides it's kind of bad experience if roundtrip (client.upload -> client.download) just doesn't work if someone is setting content-encoding on blob.

amishra-dev: @annatisch I agree with Kamil here, this will make a fundamental difference in behavior between out Python and other libraries. I am ok with an option to download it and decompress it, we should not be trying to interpret what the user's intention is. The user might want to validate integrity before decompress (make sure someone has not replaced the binary by checking against a stored SHA). Lets sync up offline and we can post the resolution here.

xf58:
This one can be closed. as this one is the final version https://github.com/Azure/azure-sdk-for-python/pull/20078, I will put more comments on https://github.com/Azure/azure-sdk-for-python/pull/20078

PR #19403, opened 2021-06-22
Add AzureApplicationCredential

### LINKED ISSUE #18020: [FEATURE REQ] Create AzureApplicationCredential
This issue is to track the design and implementation of the `AzureApplicationCredential` for applications calling service APIs not suited for `DefaultAzureCredential`. Details can be found [here](https://gist.github.com/9793dbc036d16708fb7d8fe411fe0f1f)

### LINKED ISSUE #19309: Create AzureApplicationCredential
This issue is to track the design and implementation of the `AzureImplementationCredential` for applications calling service APIs not suited for `DefaultAzureCredential`. Details can be found [here](https://gist.github.com/9793dbc036d16708fb7d8fe411fe0f1f)

chlowell:
It's the first two links of DefaultAzureCredential, namely EnvironmentCredential and ManagedIdentityCredential, so a better default for an application deployed to Azure that doesn't want to authenticate as a user or through dev tools.

Closes #18020, closes #19309


mccoyp (approved):
Looks good to me -- my only question is about the open question that was posed in the credential proposal doc:

>OPEN QUESTION: Should the AzureApplicationCredential allow users to specify service principal credential details through code? This would be different from the DefaultAzureCredential where users can only provide service principal credential information via environment variables. However, since this new credential strictly authenticates applications perhaps also providing a means to configure via code makes sense.

Did everyone agree to stick to the DefaultAzureCredential-style flow for now? I assume that means that `AZURE_TENANT_ID`, etc. are only provided as environment variables and can't be provided as, say, kwargs

chlowell:
We haven't discussed it, so no consensus on that question as yet. I started without API for specifying a service principal because I think developers who want to do that should use Certificate- or ClientSecretCredential instead. What do you think?

mccoyp:
I think I agree. It's better to add that support later if it's desired than get stuck with a feature that's not useful, since I don't see much value in using the AzureApplicationCredential in that scenario. The only useful application I can think of off the top of my head would be if someone wanted to override (a) particular environment variable(s), but that seems like an edge case.

PR #19405, opened 2021-06-23
Raw stream private

xiafu-msft:


sdk/storage/azure-storage-blob/tests/_shared/settings_real.py
```diff
@@ -0,0 +1,17 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+# ----------------------------------------------------------------s----------
+
+STORAGE_ACCOUNT_NAME = "emilydevtest"
+STORAGE_ACCOUNT_KEY = ""
```
xiafu-msft, line 8: replace this with your storage account name/key (I can also send you the key to emilydevtest offline)

sdk/storage/azure-storage-blob/tests/_shared/settings_real.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+ACCOUNT_URL_SUFFIX = 'core.windows.net'
+RUN_IN_LIVE = "True"
+SKIP_LIVE_RECORDING = "False"
```
xiafu-msft, line 13: set this to True then test will pass.

sdk/storage/azure-storage-blob/tests/test_get_blob_async.py
```diff
@@ -161,11 +163,36 @@ async def test_get_blob_to_bytes_async(self, resource_group, location, storage_a
         blob = self.bsc.get_blob_client(self.container_name, self.byte_blob)
 
         # Act
-        content = await (await blob.download_blob(max_concurrency=2)).readall()
+        content = await (await blob.download_blob()).readall()
```
xiafu-msft, line 166: if set SKIP_LIVE_RECORDING=False then this test will fail (test_get_blob_to_bytes_async is the test name)

sdk/storage/azure-storage-blob/tests/test_get_blob_async.py
```diff
@@ -161,11 +163,36 @@ async def test_get_blob_to_bytes_async(self, resource_group, location, storage_a
...
+            stream2.write(data)
+
+        with open("C:/Users/xiafu/Downloads/testthrottle.txt.gz", 'rb') as stream:
+            data = stream.read()
+            await blob.upload_blob(data, content_settings=ContentSettings(content_encoding='gzip'), overwrite=True)
+
+        # Act
+        # with self.assertRaises()
+
+        content = await (await blob.download_blob(offset=1024, length=512, decompress=False)).readall()
```
xiafu-msft, line 191: this is a more specific case, we can focus on sync test_get_blob_in_raw_stream in test_get_blob.py first

sdk/storage/azure-storage-blob/tests/test_get_blob.py
```diff
@@ -144,6 +146,50 @@ def test_get_blob_to_bytes(self, resource_group, location, storage_account, stor
...
+            blob.upload_blob(data, content_settings=ContentSettings(content_encoding='gzip'), overwrite=True)
+
+        content = blob.download_blob(offset=1024, length=512, decompress=False).readall()
+
+        # Assert
+        self.assertEqual(data[1024: 1024 + 512], content)
+
+    @pytest.mark.live_test_only
+    @GlobalStorageAccountPreparer()
+    def test_get_blob_in_raw_stream(self, resource_group, location, storage_account, storage_account_key):
```
xiafu-msft, line 172: this would fail once we turn on recording because content_encoding header is set and recording mechanism always tries to decompress.

Issue #19091, opened 2021-06-03
CredentialUnavailableError raised from `ImdsCredential` with no real connection issue


bonsairobo:
- **Package Name**: azure.identity
- **Package Version**: master
- **Operating System**: Linux
- **Python Version**: 3

**Describe the bug**

Here: https://github.com/Azure/azure-sdk-for-python/blob/e918edd34ff64e235b990f54622510ce4c34de2a/sdk/identity/azure-identity/azure/identity/_credentials/imds.py#L60

The code assumes that if gets an `Exception` other than `HttpResponseError`, then it must be some connectivity issue. This is probably not valid, as we have tested our connection to 169.254.169.254 immediately after getting the "no managed identity endpoint found" error, and there was no connection problem.

The larger issue is that the original exception is dropped, so the user can't even know what went wrong. Rather, they get a useless `CredentialUnavailableError`. It would be much nicer if you wrapped the original error so the user knows what happened.

Furthermore, after the first presumed connection failure with IMDS, the `_request_token` method will set `_endpoint_available` to `False`, which makes it *always* raise `CredentialUnavailableError` on further calls; there is no hope for recovering.

**To Reproduce**

1. Create a `SecretClient` with a `ManagedIdentity` credential.
2. We use azure.keyvault.secrets.SecretClient.get_secret and occasionally we get:

```
azure.identity._exceptions.CredentialUnavailableError: ManagedIdentityCredential authentication unavailable, no managed identity endpoint found.
```
We retry on this error, but it never recovers, and we eventually timeout at a higher layer.

**Expected behavior**

I think you have a few options:

1. Actually test TCP connections with IMDS instead of relying on exception handling at the HTTP layer.
2. Wrap the exception into the `CredentialUnavailableError` instead of dropping it so users can see what happened.

chlowell:
Sure, we can find a way to expose the error. I'm also curious what exactly is happening here. Is your application deployed to a VM?

> after the first presumed connection failure with IMDS, the _request_token method will set _endpoint_available to False, which makes it always raise CredentialUnavailableError on further calls; there is no hope for recovering.

This is a deliberate part of the design. It's awkward in some cases (pod identity) and we've discussed alternatives but have kept this behavior because the application may be running on a platform without IMDS, in which case we don't want to keep timing out waiting for responses from it. This behavior is per instance though, so if it's problematic for your application, one (admittedly not great) workaround is to create another credential instance.

bonsairobo:
[...]

PR #19423, opened 2021-06-23
Make IMDS unavailability easier to debug

chlowell:
A few small changes to make it easier to understand why ImdsCredential decided IMDS is unavailable (#19091):
1. use `raise from` to expose the exceptions motivating authentication errors
2. remove redundant logging of IMDS unavailability (the resulting authentication exception is logged at a higher level elsewhere)
3. raise a consistent error message on subsequent `get_token` calls after deciding IMDS is unavailable


mccoyp (approved):
:shipit:

Issue #11279, opened 2020-05-06
Investigate httpx transport for azure-core


lmazuel:
https://github.com/encode/httpx

CC @brettcannon

PR #19424, opened 2021-06-24
add httpx support

xiangyan99:


michaeloliverx:
@xiangyan99 Was this abandoned?

xiangyan99:
In terms of this PR, yes.

We decided to prioritize adding httpx support for rest transports (https://github.com/Azure/azure-sdk-for-python/tree/main/sdk/core/azure-core/azure/core/rest) rather than the pipeline transports.

michaeloliverx:
Is there a ticket tracking this? I would like to subscribe to it.

xiangyan99:
Yes. We have https://github.com/Azure/azure-sdk-for-python/issues/11279.

But we cannot make it happen until httpx has a stable release.

Issue #17365, opened 2021-03-16
[Core] Odata error parser fails on null details


annatisch:
The code to parse the `details` field of an Odata error assumes the value will be an iterable, or the key not present at all.
Sometimes services can return this value as `null` - which then fails to parse.
We should update this section:
```python
for detail_node in json_object.get(cls.DETAILS_LABEL, []):
    ...
```
To something like this:
```python
for detail_node in json_object.get(cls.DETAILS_LABEL) or []:
    ...
```

Services that return null for the `details` field are in violation of Odata v4 - so we're already parsing this correctly. However it seems like a small change we can do to make this code a tiny bit more robust in case of an imperfect response body.

xiangyan99:
Fixed

PR #19430, opened 2021-06-24
handle details "null"

xiangyan99:
#17365


annatisch (approved):
Change looks good - was there a test that was breaking/fixed due to this?
(and if not - could we add one?)

PR #19433, opened 2021-06-24
raise decode error instead of ContentDecodingError

xiangyan99:


sdk/core/azure-core/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## 1.15.1 (Unreleased)
 
+### Breaking Changes
+
+- Sync stream downloading now raises `azure.core.exceptions.DecodeError` rather than `requests.exceptions.ContentDecodingError`
```
annatisch, line 7: Are we okay with this breaking change because no one should have written code to explicitly handle this error?

xiangyan99: Yes. Checked with @johanste and he agreed this was an acceptable breaking change.

sdk/core/azure-core/azure/core/pipeline/transport/_requests_basic.py
```diff
@@ -58,11 +59,11 @@ def _read_raw_stream(response, chunk_size=1):
             for chunk in response.raw.stream(chunk_size, decode_content=False):
                 yield chunk
         except ProtocolError as e:
-            raise requests.exceptions.ChunkedEncodingError(e)
-        except DecodeError as e:
-            raise requests.exceptions.ContentDecodingError(e)
+            raise ServiceResponseError(e, error=e)
+        except CoreDecodeError as e:
+            raise DecodeError(e, error=e)
```
annatisch, line 64: This scenario is being correctly handled in the other transports?

xiangyan99: Currently different transport raises different type of error. (That's the reason we need this change).

We want to uniform the errors so user can catch all by using except AzureError.

lmazuel (approved):

Issue #19297, opened 2021-06-17
[ServiceBus] time_to_live property triggering OverflowError: mktime argument out of range in core regression test


yunhaoling:
- **Package Name**: azure-servicebus
- **Package Version**: 7.3.0
- **Operating System**: Linux
- **Python Version**: 3.7.10

**Describe the bug**
in the internal pipeline `python - core` 20210616.1, test `test_servicebus_message_repr_with_props` consistently fails:
```
2021-06-17T07:21:57.2207974Z ___________________ test_servicebus_message_repr_with_props ____________________
2021-06-17T07:21:57.2212410Z 
2021-06-17T07:21:57.2213432Z     def test_servicebus_message_repr_with_props():
2021-06-17T07:21:57.2214516Z         message = ServiceBusMessage(
2021-06-17T07:21:57.2215552Z             body="hello",
2021-06-17T07:21:57.2216947Z             application_properties={'prop': 'test'},
2021-06-17T07:21:57.2230072Z             session_id="id_session",
2021-06-17T07:21:57.2231277Z             message_id="id_message",
2021-06-17T07:21:57.2232473Z             scheduled_enqueue_time_utc=datetime.now(),
2021-06-17T07:21:57.2233686Z             time_to_live=timedelta(seconds=30),
2021-06-17T07:21:57.2234921Z             content_type="content type",
2021-06-17T07:21:57.2236123Z             correlation_id="correlation",
2021-06-17T07:21:57.2237283Z             subject="github",
2021-06-17T07:21:57.2238427Z             partition_key="id_session",
2021-06-17T07:21:57.2239597Z             to="forward to",
2021-06-17T07:21:57.2240660Z             reply_to="reply to",
2021-06-17T07:21:57.2241659Z >           reply_to_session_id="reply to session"
2021-06-17T07:21:57.2242616Z             )
2021-06-17T07:21:57.2243542Z 
2021-06-17T07:21:57.2245062Z azure-sdk-for-python/sdk/servicebus/azure-servicebus/tests/test_message.py:37: 
2021-06-17T07:21:57.2246547Z _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
2021-06-17T07:21:57.2248760Z azure-sdk-for-python/sdk/servicebus/azure-servicebus/azure/servicebus/_common/message.py:123: in __init__
[...]

PR #19437, opened 2021-06-24
[ServiceBus] Fix mktime overflow and milliseconds miscalculation issue

yunhaoling:
fixing issue: https://github.com/Azure/azure-sdk-for-python/issues/19297

- on ubuntu 20.04, mktime(datetime.now(msrest.serialization.UTC()).timetuple()) would trigger Overflow. we should use msrest.serialization.TZ_UTC instead which will use the python built-in timezone info to avoid the issue.
- according to the AMQP spec, creation_time and absoulate_expiry_time should be calculated in the unit of millisecond.


swathipil (approved):
LG(great)TM

PR #19499, opened 2021-06-25
Resolve Checkout of `AutoRest` repo during `verify-autorest`

scbedd:
Still on master.


iscai-msft (approved):

scbedd:
Overriding due to the fact that this is a safe change.

Issue #19243, opened 2021-06-14
[ServiceBus] Stress test embracing azure application insights for metrics/traces


yunhaoling:
echoing issue: https://github.com/Azure/azure-sdk-for-python/issues/18792
check the code in azure-eventhub stress test for application insights usage

PR #19501, opened 2021-06-25
[ServiceBus] Stress test embracing appinsights

yunhaoling:
addressing issue: https://github.com/Azure/azure-sdk-for-python/issues/19243

- add async stress test code
- integration with appinsights


sdk/servicebus/azure-servicebus/tests/stress_tests/utilities.py
```diff
@@ -0,0 +1,49 @@
...
+    _logger.debug("Delivery count: {}".format(message.delivery_count))
+    try:
+        _logger.debug("Locked until: {}".format(message.locked_until_utc))
+        _logger.debug("Lock Token: {}".format(message.lock_token))
+    except (TypeError, AttributeError):
+        pass
+    _logger.debug("Enqueued time: {}".format(message.enqueued_time_utc))
+
+
+def sleep_until_expired(entity):
```
swathipil, line 48: do all the methods in utilities.py get used? this one looks like it doesn't

yunhaoling: thanks for calling it out! I find that the file is not needed at all.
the only method used is the `_build_logger` which could totally be replaced by the method in `logger.py`.

removed :)

swathipil (approved):
LGTM

PR #19502, opened 2021-06-25
[core] add provisional azure.core.rest

iscai-msft:


sdk/core/azure-core/azure/core/exceptions.py
```diff
@@ -433,3 +433,45 @@ def __str__(self):
         if self._error_format:
             return str(self._error_format)
         return super(ODataV4Error, self).__str__()
+
+class StreamConsumedError(Exception):
```
xiangyan99, line 437: Should it be a sub class of AzureError?

xiangyan99: And if the error type is provisional, should we host it in exceptions.py?

sdk/core/azure-core/azure/core/rest/_rest_py3.py
```diff
@@ -0,0 +1,522 @@
...
+                yield line
+
+    async def close(self) -> None:
+        """Close the response.
+
+        :return: None
+        :rtype: None
+        """
+        self.is_closed = True
+        self.internal_response.close()
```
annatisch, line 510: This is never an async operation?

sdk/core/azure-core/azure/core/rest/_requests_trio.py
```diff
@@ -0,0 +1,58 @@
...
+        """
+
+        async for part in iter_bytes_helper(TrioStreamDownloadGenerator, self):
+            yield part
+        await self.close()
+
+    async def close(self) -> None:
+        self.is_closed = True
+        self.internal_response.close()
+        await trio.sleep(0)
```
annatisch, line 58: Why is this the only transport that overrides `close`?

sdk/core/azure-core/azure/core/rest/_requests_basic.py
```diff
@@ -0,0 +1,132 @@
...
+    def __init__(self, **kwargs):
+        super(_RestRequestsTransportResponseBase, self).__init__(**kwargs)
+        self.status_code = self.internal_response.status_code
+        self.headers = self.internal_response.headers
+        self.reason = self.internal_response.reason
+        self.content_type = self.internal_response.headers.get('content-type')
+
+    def _get_content(self):
+        """Return the internal response's content"""
+        if not self.internal_response._content_consumed:  # pylint: disable=protected-access
```
xiangyan99, line 45: Why if content is not consumed, we return None?

sdk/core/azure-core/azure/core/rest/_requests_basic.py
```diff
@@ -0,0 +1,132 @@
...
+        if encoding:
+            if encoding == "utf-8":
+                encoding = "utf-8-sig"
+        self.internal_response.encoding = encoding
+
+    @property
+    def text(self):
+        if not self._has_content():
+            raise ResponseNotReadError()
+        return self.internal_response.text
```
xiangyan99, line 85: Does it work correctly in aiohttp compressed content?

sdk/core/azure-core/azure/core/rest/_requests_basic.py
```diff
@@ -0,0 +1,132 @@
...
+
+    def _set_content(self, val):
+        """Set the internal response's content"""
+        self.internal_response._content = val  # pylint: disable=protected-access
+
+    def _has_content(self):
+        return self._get_content() is not None
+
+    @_HttpResponseBase.encoding.setter  # type: ignore
+    def encoding(self, value):
```
xiangyan99, line 63: Does it work correctly in aiohttp compressed content?

xiangyan99: I mean if a service supports transfer-encoding and returns compressed body (not streaming case)

sdk/core/azure-core/azure/core/rest/_requests_trio.py
iscai-msft: because the default one uses asyncio. Had thought about passing in the transport to the response so we could use the transport's sleep method, but it seemed a little overkill, lmk your thoughts

sdk/core/azure-core/azure/core/rest/_rest_py3.py
iscai-msft: yeah, that's bc the internal responses's `close` s are not awaitable. You're comment does make me think that in the base class, it can be neater and we just call `await self.internal_response.close()`, and I'll move this onto the base classes themselves. Thanks!

sdk/core/azure-core/azure/core/_pipeline_client.py
```diff
@@ -170,3 +182,30 @@ def _build_pipeline(self, config, **kwargs): # pylint: disable=no-self-use
...
+        return_pipeline_response = kwargs.pop("_return_pipeline_response", False)
+        pipeline_response = self._pipeline.run(request_to_run, **kwargs)  # pylint: disable=protected-access
+        response = pipeline_response.http_response
+        if rest_request:
+            response = response._to_rest_response()  # pylint: disable=protected-access
+            if not kwargs.get("stream", False):
+                response.read()
+                response.close()
+        if return_pipeline_response:
+            pipeline_response.http_response = response
```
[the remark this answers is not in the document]
iscai-msft: oh that's a v good point, thanks!

sdk/core/azure-core/azure/core/rest/_requests_basic.py
iscai-msft: updated, please check. thanks!

sdk/core/azure-core/azure/core/rest/_requests_basic.py
iscai-msft: this is for requests, but have updated aiohttp, lmk your thoughts!

sdk/core/azure-core/azure/core/rest/_helpers_py3.py
```diff
@@ -0,0 +1,99 @@
...
+        response=response,
+        decompress=decompress,
+    )
+
+async def iter_bytes_helper(
+    stream_download_generator: Callable,
+    response,
+) -> AsyncIteratorType[bytes]:
+    if response._has_content():  # pylint: disable=protected-access
+        yield response._get_content()  # pylint: disable=protected-access
```
xiangyan99, line 79: If you use it in iter, you need a decompressor.

iscai-msft: I just updated the code to not use these getters, hopefully it makes more sense now. Lmk your thoughts, thanks!

sdk/core/azure-core/azure/core/rest/_requests_basic.py
iscai-msft: this code was kind of confusing, so I fixed it to be less confusing. thanks!

sdk/core/azure-core/azure/core/exceptions.py
iscai-msft: @johanste bringing this question to you

sdk/core/azure-core/azure/core/exceptions.py
```diff
@@ -433,3 +433,45 @@ def __str__(self):
...
+    """
+    def __init__(self):
+        message = (
+            "You are attempting to read or stream content that has already been streamed. "
+            "You have likely already consumed this stream, so it can not be accessed anymore."
+        )
+        super(StreamConsumedError, self).__init__(message)
+
+class StreamClosedError(Exception):
+    """:bolditalic:`Provisional` error thrown if you try to access the stream of a response once closed.
```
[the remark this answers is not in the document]
iscai-msft: open a bug for better repr for `AzureError`

sdk/core/azure-core/azure/core/pipeline/transport/_aiohttp.py
```diff
@@ -358,3 +359,9 @@ def __getstate__(self):
         state['internal_response'] = None  # aiohttp response are not pickable (see headers comments)
         state['headers'] = CIMultiDict(self.headers)  # MultiDictProxy is not pickable
         return state
+
+    def _to_rest_response(self):
```
iscai-msft, line 363: switch this to external mapping

sdk/core/azure-core/azure/core/pipeline/transport/_base.py
```diff
@@ -473,6 +473,15 @@ def serialize(self):
         """
         return _serialize_request(self)
 
+    def _to_rest_request(self):
```
iscai-msft, line 476: switch this to external function

sdk/core/azure-core/azure/core/pipeline/transport/_requests_basic.py
```diff
@@ -71,6 +72,9 @@ def _read_raw_stream(response, chunk_size=1):
                 break
             yield chunk
 
+    # following behavior from requests iter_content, we set content consumed to True
+    response._content_consumed = True  # pylint: disable=protected-access
```
iscai-msft, line 76: link the requests code

sdk/core/azure-core/azure/core/rest/__init__.py
```diff
@@ -0,0 +1,56 @@
...
+    from ._rest_py3 import (
+        HttpRequest,
+        HttpResponse,
+        _HttpResponseBase,
+    )
+except (SyntaxError, ImportError):
+    from ._rest import (  # type: ignore
+        HttpRequest,
+        HttpResponse,
+        _HttpResponseBase,
```
iscai-msft, line 36: get rid of _HttpResponseBase

sdk/core/azure-core/azure/core/rest/__init__.py
```diff
@@ -0,0 +1,56 @@
...
+__all__ = [
+    "HttpRequest",
+    "HttpResponse",
+    "_HttpResponseBase",
+]
+
+try:
+    from ._rest_py3 import (  # pylint: disable=unused-import
+        AsyncHttpResponse,
+        _AsyncContextManager,
```
iscai-msft, line 48: remove _AsyncContextManager

sdk/core/azure-core/azure/core/rest/_aiohttp.py
```diff
@@ -0,0 +1,145 @@
...
+
+class RestAioHttpTransportResponse(AsyncHttpResponse):
+    def __init__(
+        self,
+        *,
+        request: HttpRequest,
+        internal_response,
+    ):
+        super().__init__(request=request, internal_response=internal_response)
+        self.status_code = internal_response.status
```
iscai-msft, line 50: make `internal_response` property private

sdk/core/azure-core/azure/core/rest/_aiohttp.py
```diff
@@ -0,0 +1,145 @@
...
+    async def iter_raw(self) -> AsyncIterator[bytes]:
+        """Asynchronously iterates over the response's bytes. Will not decompress in the process
+        :return: An async iterator of bytes from the response
+        :rtype: AsyncIterator[bytes]
+        """
+        async for part in iter_raw_helper(AioHttpStreamDownloadGenerator, self):
+            yield part
+        await self.close()
+
+    async def iter_bytes(self) -> AsyncIterator[bytes]:
```
iscai-msft, line 117: add `type: ignore` here, so I don't have to have gross code to satiate mypy in base class

iscai-msft: open an issue, link it

sdk/core/azure-core/CHANGELOG.md
```diff
@@ -1,6 +1,14 @@
 # Release History
 
-## 1.15.1 (Unreleased)
+## 1.16.0 (Unreleased)
+
+### New Features
+
+- Add new ***provisional*** methods `send_request` onto the `azure.core.PipelineClient` and `azure.core.AsyncPipelineClient`. This method takes in
+requests and sends them through our pipelines.
+- Add new ***provisional*** module `azure.core.rest`. `azure.core.rest` is our new public simple HTTP library in `azure.core` that users will use to create requests, and consume responses. Can only be used with the provisional method `send_request` on our `PipelineClient`s
```
iscai-msft, line 9: get rid of `PipelineClient` `send_request` mention in `azure.core.rest` changelog entry.

Instead, in doc for azure.core.rest, tell them to pass it to their client's `send_request`

xiangyan99 (commented):
Should we update readme to introduce these new error types?

sdk/core/azure-core/azure/core/exceptions.py
iscai-msft: @xiangyan99 ^^

sdk/core/azure-core/azure/core/_pipeline_client_async.py
```diff
@@ -168,3 +172,46 @@ def _build_pipeline(self, config, **kwargs): # pylint: disable=no-self-use
...
+                rest_response._content = response.body()  # pylint: disable=protected-access
+                await rest_response.close()
+            response = rest_response
+        if return_pipeline_response:
+            pipeline_response.http_response = response
+            pipeline_response.http_request = request
+            return pipeline_response
+        return response
+
+    def send_request(
```
xiangyan99, line 198: Not async?

iscai-msft: yeah, it's a little confusing, but it returns an awaitable object. This has to do with the fact that we're returning a context manager that both has an `__await__` and a `__aenter__` etc. So this way, you can call it like this:
```python
response = await client.send_request(request)
```
and like this (mainly in the case of streaming)
```python
async with client.send_request(request) as response:
    print(response)
```

talked through this with @johanste and he's ok with not making it `async def`. Reason is once we mark it as `async def`, the compiler thinks we're returning a coroutine, which don't have attributes `__aenter__`, so we get an attribute error when trying to use it as a context managed response

sdk/core/azure-core/azure/core/pipeline/_tools_async.py
```diff
@@ -31,3 +32,36 @@ async def await_result(func, *args, **kwargs):
         # type ignore on await: https://github.com/python/mypy/issues/7587
         return await result  # type: ignore
     return result
+
+def to_rest_response(pipeline_transport_response):
+    response_type = None
+    try:
+        from .transport import AioHttpTransportResponse
+        from ..rest._aiohttp import RestAioHttpTransportResponse
```
xiangyan99, line 40: Do we support (or plan to support) custom transport?

sdk/core/azure-core/azure/core/pipeline/_tools_async.py
```diff
@@ -31,3 +32,36 @@ async def await_result(func, *args, **kwargs):
...
+        if isinstance(pipeline_transport_response, AioHttpTransportResponse):
+            response_type = RestAioHttpTransportResponse
+    except ImportError:
+        pass
+    try:
+        from .transport import AsyncioRequestsTransportResponse
```
xiangyan99, line 46: Do we need to import all of them? 

Not only needed in except ImportError?

sdk/core/azure-core/azure/core/pipeline/_tools_async.py
iscai-msft: so what I'm doing here is "if . However, I can't do a specific corresponding rest transport response for a custom transport since idk what that transport is, so I fall back to just creating the default `AsyncHttpResponse` out of it

Our goal in a later PR (hopefully in the next release cycle) is to fix the need to transform rest requests and responses, and somehow have our pipelines support both the old and new requests and responses. So this code is also temporary.

sdk/core/azure-core/azure/core/pipeline/_tools_async.py
iscai-msft: i can add a check for each of these to only try to import if response_type is still `None`. But basically I put them each into their own try except import error block bc, for example, if you have a trio transport and no aiohttp isntalled, importing aiohttp will cause an error for you, so I wanted to accomodate all of the different combos

sdk/core/azure-core/azure/core/rest/_aiohttp.py
```diff
@@ -0,0 +1,147 @@
...
+        enc = self.headers.get('Content-Encoding')
+        if not enc:
+            return self._content
+        enc = enc.lower()
+        if enc in ("gzip", "deflate"):
+            if self._decompressed_content:
+                return self._decompressed_content
+            import zlib
+            zlib_mode = 16 + zlib.MAX_WBITS if enc == "gzip" else zlib.MAX_WBITS
+            decompressor = zlib.decompressobj(wbits=zlib_mode)
```
xiangyan99, line 72: You need a class level decompressor.

sdk/core/azure-core/azure/core/pipeline/_tools_async.py
iscai-msft: ok, I now stop importing once I get a reseponse type. Thanks for the suggestion @xiangyan99 !

sdk/core/azure-core/azure/core/rest/_helpers.py
```diff
@@ -0,0 +1,279 @@
...
+    Tuple,
+    IO,
+    Any,
+    Dict,
+    Iterable,
+    Iterator,
+    cast,
+    Callable,
+)
+import xml.etree.ElementTree as ET
```
xiangyan99, line 46: If I remember correctly, xml.etree is very heavy. We can consider import it only needed.

sdk/core/azure-core/azure/core/_pipeline_client.py
```diff
@@ -170,3 +184,37 @@ def _build_pipeline(self, config, **kwargs): # pylint: disable=no-self-use
...
+        :rtype: ~azure.core.rest.HttpResponse
+        # """
+        rest_request, request_to_run = _prepare_request(request)
+        return_pipeline_response = kwargs.pop("_return_pipeline_response", False)
+        pipeline_response = self._pipeline.run(request_to_run, **kwargs)  # pylint: disable=protected-access
+        response = pipeline_response.http_response
+        if rest_request:
+            response = _to_rest_response(response)
+            if not kwargs.get("stream", False):
+                response.read()
```
johanste, line 214: Will we leak a stream if `read` fails?

sdk/core/azure-core/azure/core/_pipeline_client_async.py
```diff
@@ -168,3 +173,52 @@ def _build_pipeline(self, config, **kwargs): # pylint: disable=no-self-use
             transport = AioHttpTransport(**kwargs)
 
         return AsyncPipeline(transport, policies)
+
+    async def _make_pipeline_call(self, request, stream, **kwargs):
```
johanste, line 177: Nitpick; I'd make `stream` a keyword-only argument.

sdk/core/azure-core/azure/core/exceptions.py
```diff
@@ -433,3 +433,45 @@ def __str__(self):
...
+class ResponseNotReadError(AzureError):
+    """**Provisional** error thrown if you try to access a response's content without reading first.
+
+    This error is marked as **provisional**, meaning it may be changed in a future release. It is
+    thrown if you try to access an ~azure.core.rest.HttpResponse or
+    ~azure.core.rest.AsyncHttpResponse's content without first reading the response's bytes in first.
+    """
+
+    def __init__(self):
+        message = (
```
johanste, line 474: From a diagnostics perspective, it may be interesting to know for *which* request this failed. In other words, either provide the response in the constructor, or, at a minimum, take the method + path as arguments to the exception....

sdk/core/azure-core/azure/core/rest/_helpers.py
iscai-msft: ok sounds good

iscai-msft: actually @xiangyan99 , we end up having to use ET whenever people pass in `content` to the `HttpRequest` (and it has to be the first check bc it's the most specific), so I think I'm going to leave it at the top of the module, since I think it becomes more costly if we have to import it everytime someone passes in `content`. Thanks!

sdk/core/azure-core/azure/core/_pipeline_client_async.py
iscai-msft: sounds good. I've left it as a keyword only argument on send_request

sdk/core/azure-core/azure/core/exceptions.py
iscai-msft: good point, I'll do something like

"You have not read in the bytes for the response from request '<HttpRequest [GET], url: http://example.com>'. Call .read() on this response first."

sdk/core/azure-core/azure/core/rest/_aiohttp.py
```diff
@@ -0,0 +1,147 @@
...
+        return self._content
+
+    @property
+    def text(self) -> str:
+        content = self.content
+        encoding = self.encoding
+        ctype = self.headers.get(aiohttp.hdrs.CONTENT_TYPE, "").lower()
+        mimetype = aiohttp.helpers.parse_mimetype(ctype)
+
+        encoding = mimetype.parameters.get("charset")
```
xiangyan99, line 84: Given we set encoding here, why we need L80?

sdk/core/azure-core/azure/core/rest/_rest.py
```diff
@@ -0,0 +1,389 @@
...
+        """
+        raise NotImplementedError()
+
+    def iter_bytes(self):
+        # type: () -> Iterator[bytes]
+        """Iterate over the response bytes
+        """
+        raise NotImplementedError()
+
+    def iter_text(self):
```
xiangyan99, line 371: Do you override iter_text for aiohttp?

If not, how can we make sure self.encoding works fine?

sdk/core/azure-core/azure/core/rest/_aiohttp.py
iscai-msft: talked to @xiangyan99 , no need for this. Thanks!

sdk/core/azure-core/azure/core/exceptions.py
iscai-msft: fixed, thanks!

sdk/core/azure-core/azure/core/rest/_aiohttp.py
iscai-msft: I'm following the strucutre for the code in pipeline transport. But basically, it seems that if the chardet detecting fails, encoding still might be none, so we fall back to "utf-8-sig"

xiangyan99: I mean L84 will override L80, right?

iscai-msft: I cleaned this up a bit, lmk your thoughts, thanks!

sdk/core/azure-core/azure/core/rest/_rest.py
iscai-msft: I ended up moving parts of the encoding into the base class for encoding.
Basically, 1. if we have an application/json, following RFC 7159 and 7483, I default this to "utf-8". 2. if the response's content has been read in, I'll add an additional step of using chardet.detect to get the response's content. With this, the self.encoding should cover more corner cases with `iter_text`

sdk/core/azure-core/azure/core/_pipeline_client.py
iscai-msft: good point, I'm going to add a try except when I read the response, and close the response before throwing the error.  thanks!

sdk/core/azure-core/azure/core/rest/_rest.py
```diff
@@ -0,0 +1,376 @@
...
+    def text(self):
+        # type: (...) -> str
+        """Returns the response body as a string"""
+        encoding = self.encoding
+        if encoding == "utf-8" or encoding is None:
+            encoding = "utf-8-sig"
+        return self.content.decode(encoding)
+
+    @property
+    def num_bytes_downloaded(self):
```
xiangyan99, line 243: Please hide it. There is no reliable way to return the bytes.

sdk/core/azure-core/azure/core/pipeline/_tools.py
```diff
@@ -32,3 +32,40 @@ def await_result(func, *args, **kwargs):
...
+    from ..rest import HttpRequest as RestHttpRequest
+    return RestHttpRequest(
+        method=pipeline_transport_request.method,
+        url=pipeline_transport_request.url,
+        headers=pipeline_transport_request.headers,
+        files=pipeline_transport_request.files,
+        data=pipeline_transport_request.data
+    )
+
+def to_rest_response(pipeline_transport_response):
```
annatisch, line 46: What will this method do if users have created their client with a custom transport?
Does it matter here?

iscai-msft: if they create their client with a custom transport, since i don't have a corresponding rest transport response to change it to, I change it to the default `azure.core.rest.HttpResponse` or `azure.core.rest.AsyncHttpResponse`

sdk/core/azure-core/azure/core/rest/_rest.py
iscai-msft: All I do in num_bytes_downloaded is I get. the length of the chunk before I send it to users, I feel like that's p reliable, why are you thinking it's not reliable? sorry if I'm missing something

xiangyan99: in compression scenario. Which num is the correct? Before decompression or after?

sdk/core/azure-core/azure/core/rest/_rest.py
xiangyan99: Is it possible iter_text is called w/o .read()? 

In this case, content is empty?

sdk/core/azure-core/azure/core/rest/_rest.py
iscai-msft: you're right that seems fair, i can remove for now

sdk/core/azure-core/azure/core/rest/_rest.py
iscai-msft: yeah definitely. Basically for iter_bytes, iter_text, and iter_lines, if `content` exists, we'll yield the content to users. Otherwise, we'll call stream download generator and get the bytes from there.

This is the same behavior as httpx

xiangyan99: You calls iter_bytes or content?

It seems to me iter_bytes does not generate content?

iscai-msft: here's kind of the scenarios where you would use one vs. the other
1. I want the whole body on the response, and I want to access it all at once: Here you would call .content, similar to `.body()` on the old responses. You may need to call `.read()` first, to read in the response's bytes to content before
2. I want to iterate over the body: here, i call one of the `iter_` functions. This doesn't read the body into the response, but outputs it as an iterator. As more of an implementation detail: if the response's bytes are already read into `content`, we will iterate over the content in all of the `iter_` functions except `iter_raw` (consistent with httpx), and output them in chunks <= the block size we get from the transport

xiangyan99: I think we are on same page. So if the content is not read into the body, iter_text will not be able to detect the encoding from content, right?

iscai-msft: It still will be able. to do all of the checks, except the final fallback I added, which is the chardet.detect of the content

annatisch (approved):
LGTM - can't see anything that would break existing pipelines.
Haven't dug too deep into the new implementation of the new `rest` implementation, but API surface looks good and well documented as provisional :)

lmazuel (approved):
Did a short fast review, LGTM :)

PR #19547, opened 2021-06-28
[Tables] updating docstring

seankane-msft:


sdk/tables/azure-data-tables/azure/data/tables/aio/_base_client_async.py
```diff
@@ -35,17 +35,20 @@
...
     def __init__(  # pylint: disable=missing-client-constructor-parameter-credential
         self,
         endpoint: str,
         *,
-        credential: Optional[Union[AzureSasCredential, AzureNamedKeyCredential]] = None,
+        credential: Optional[Union[AzureSasCredential, AzureNamedKeyCredential, "AsyncTokenCredential"]] = None,
         **kwargs: Any
     ) -> None:
-        super(AsyncTablesBaseClient, self).__init__(endpoint, credential=credential, **kwargs)
+        super(AsyncTablesBaseClient, self).__init__(endpoint, credential=credential, **kwargs)  # type: ignore
```
annatisch, line 51: Is this a problem that we are passing an async credential policy into the parent base_client constructor?

sdk/tables/azure-data-tables/azure/data/tables/aio/_table_service_client_async.py
```diff
@@ -326,7 +326,7 @@ def get_table_client(self, table_name: str, **kwargs) -> TableClient:
         return TableClient(
             self.url,
             table_name=table_name,
-            credential=self.credential,
+            credential=self.credential,  # type: ignore
```
annatisch, line 329: Why do we have to suppress the type here?

sdk/tables/azure-data-tables/azure/data/tables/aio/_base_client_async.py
seankane-msft: Yes, this gets passed all the way down to `AccountHostsMixin` which expects a sync token credential.

sdk/tables/azure-data-tables/azure/data/tables/aio/_table_service_client_async.py
seankane-msft: The AccountHostsMixin expects a sync token credential.

annatisch (approved):

PR #19551, opened 2021-06-28
Patch for devtools_testutils

seankane-msft:


tools/azure-sdk-tools/devtools_testutils/storage_testcase.py
```diff
@@ -14,8 +14,11 @@
     pass
 
 from azure.mgmt.storage import StorageManagementClient
```
xiangyan99, line 16: Do we need to try this as well?

iscai-msft (approved):

xiafu-msft (approved):

PR #19581, opened 2021-06-29
Override mindependency for cryptography

scbedd:
From @chlowell 

> In a nutshell, we have `azure-identity -> cryptography` and `azure-identity -> msal -> PyJWT -> cryptography`. Min cryptography is `2.1.4 `for azure-identity but the latest PyJWT requires `3.3+` and will raise given < 3.2. So, mindependency can't pass. I think msal should drop PyJWT but in the meantime I need to pass mindependency. I could just raise azure-identity's min cryptography, that's ugly but I guess workable for a beta release


chlowell (approved):

scbedd:
Follow-up issue is #19582

PR #19595, opened 2021-06-30
Add custom JSONEncoder for model serialization

mccoyp:
This adds a custom JSONEncoder that can serialize `datetime` objects (`date`s, `time`s, `datetime`s, and `timedelta`s) in UTC ISO 8601 format, as well as `bytes` and `bytearray`s in base64 strings.

### DIFF
--- a/sdk/core/azure-core/azure/core/serialization.py
+++ b/sdk/core/azure-core/azure/core/serialization.py
@@ -5,10 +5,13 @@
 # license information.
 # --------------------------------------------------------------------------
 import base64
-import datetime
 from json import JSONEncoder
+from typing import TYPE_CHECKING
 
 from .utils._utils import _FixedOffset
+
+if TYPE_CHECKING:
+    from datetime import timedelta
 
 __all__ = ["NULL"]
 
@@ -30,7 +33,7 @@ with no data. This gets serialized to `null` on the wire.
 
 
 def timedelta_as_isostr(value):
-    # type: (datetime.timedelta) -> str
+    # type: (timedelta) -> str
     """Converts a datetime.timedelta object into an ISO 8601 formatted string, e.g. 'P4DT12H30M05S'
 
     Function adapted from the Tin Can Python project: https://github.com/RusticiSoftware/TinCanPython
@@ -66,16 +69,16 @@ def timedelta_as_isostr(value):
     # Seconds
     try:
         if seconds.is_integer():
-            seconds = "{:02}".format(int(seconds))
+            seconds_string = "{:02}".format(int(seconds))
         else:
             # 9 chars long w/ leading 0, 6 digits after decimal
-            seconds = "%09.6f" % seconds
+            seconds_string = "%09.6f" % seconds
             # Remove trailing zeros
-            seconds = seconds.rstrip("0")
+            seconds_string = seconds_string.rstrip("0")
     except AttributeError:  # int.is_integer() raises on Python 2.7
-        seconds = "{:02}".format(seconds)
+        seconds_string = "{:02}".format(seconds)
 
-    time += "{}S".format(seconds)
+    time += "{}S".format(seconds_string)
 
     return "P" + date + time
 
@@ -104,8 +107,7 @@ class ComplexEncoder(JSONEncoder):
                         return o.replace(tzinfo=TZ_UTC).isoformat()
                     return o.astimezone(TZ_UTC).isoformat()
                 # Next try datetime.date or datetime.time
-                else:
-                    return o.isoformat()
+                return o.isoformat()
             except AttributeError:
                 pass
             # Last, try datetime.timedelta
--- a/sdk/core/azure-core/azure/core/serialization.py
+++ b/sdk/core/azure-core/azure/core/serialization.py
@@ -29,13 +29,14 @@ with no data. This gets serialized to `null` on the wire.
 """
 
 
-def iso_timedelta(value):
-    """Represent a timedelta in ISO 8601 format.
+def timedelta_as_isostr(value):
+    # type: (datetime.timedelta) -> str
+    """Converts a datetime.timedelta object into an ISO 8601 formatted string, e.g. 'P4DT12H30M05S'
 
-    Function from the Tin Can Python project: https://github.com/RusticiSoftware/TinCanPython
+    Function adapted from the Tin Can Python project: https://github.com/RusticiSoftware/TinCanPython
     """
 
-    # split seconds to larger units
+    # Split seconds to larger units
     seconds = value.total_seconds()
     minutes, seconds = divmod(seconds, 60)
     hours, minutes = divmod(minutes, 60)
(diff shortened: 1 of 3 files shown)


sdk/core/azure-core/azure/core/serialization.py
```diff
@@ -4,20 +4,117 @@
...
+        seconds = '%09.6f' % seconds
+        # remove trailing zeros
+        seconds = seconds.rstrip('0')
+
+    time += '{}S'.format(seconds)
+
+    return 'P' + date + time
+
+
+class UTC(datetime.tzinfo):
```
xiangyan99, line 77: We have similar code https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/core/azure-core/azure/core/_utils.py#L10.

xiangyan99:
@mccoyp _utils.py has been moved into utils folder.

sdk/core/azure-core/azure/core/serialization.py
```diff
@@ -4,20 +4,103 @@
...
 A falsy sentinel object which is supposed to be used to specify attributes
 with no data. This gets serialized to `null` on the wire.
 """
+
+
+def iso_timedelta(value):
+    """Represent a timedelta in ISO 8601 format.
+
+    Function from the Tin Can Python project: https://github.com/RusticiSoftware/TinCanPython
+    """
```
annatisch, line 36: I would be a bit more explicit about declaring/documenting the function:
```
def timedelta_as_isostr(value):
    # type: (datatime.timedelta) -> str
    """
    Converts a datetime.timedelta object into an ISO 8601 formatted string, e.g. 'P3Y6M4DT12H30M5S'
    """
```

sdk/core/azure-core/azure/core/serialization.py
```diff
@@ -4,20 +4,103 @@
...
+
+
+class ComplexEncoder(JSONEncoder):
+    """A JSON encoder that's capable of serializing datetime objects and bytes."""
+
+    def default(self, o):  # pylint: disable=too-many-return-statements
+        try:
+            return super(ComplexEncoder, self).default(o)
+        except TypeError:
+            o_type = type(o)
```
annatisch, line 94: We should avoid using `type`
For example, in the Tables SDK, we use a custom datetime object that I think would be missed by this logic:
https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/tables/azure-data-tables/azure/data/tables/_deserialize.py#L26-L33

sdk/core/azure-core/dev_requirements.txt
```diff
@@ -1,5 +1,6 @@
 trio; python_version >= '3.5'
 aiohttp>=3.0; python_version >= '3.5'
+isodate>=0.6.0
```
annatisch, line 3: This is only a test requirement right? Core isn't taking a dependency on this library?

sdk/core/azure-core/azure/core/serialization.py
annatisch: Looking at the if-statement logic below - I think you could get away with no type checking for the datetime objects, and isinstance should work for bytes/bytearray.
Roughly something like:
```python
if isinstance(o, (bytes, bytearray)):
    return base64.b64encode(o).decode()
try:
    # First try datetime.datetime
    if not o.tzinfo:
        return o.replace(tzinfo=TZ_UTC).isoformat()
    return o.astimezone(TZ_UTC).isoformat()
except AttributeError:
    pass
try:
    # Next try datetime.date or datetime.time
    return o.isoformat()
except AttributeError:
    pass
try:
    # Last, try datetime.timedelta
    return iso_timedelta(o)
except AttributeError:
    # This will be raised when it hits value.total_seconds in the method above.
    pass
return super(ComplexEncoder, self).default(o)
```

annatisch: I would then add a test for serializing an instance of the Tables object I linked to above to make sure it works :)

sdk/core/azure-core/tests/test_serialization.py
```diff
@@ -2,10 +2,356 @@
...
+        "list_of_ints": [1, 2, 3],
+        "dictionary_of_number": {"pi": 3.14},
+    }
+    assert json.loads(json_dumps_with_encoder(expected.to_dict())) == expected_dict
+
+def test_dictionary_datetime(json_dumps_with_encoder):
+    test_obj = {
+        "timedelta": timedelta(1),
+        "date": date(2021, 5, 12),
+        "datetime": isodate.parse_datetime('2012-02-24T00:53:52.780Z'),
```
annatisch, line 114: When the conversation starts regarding deserialization - we will want to make sure we support weird .NET timestamps (with extra decimal places). When that conversation happens - what we do to serialize round-tripped values will be important as well.

sdk/core/azure-core/azure/core/serialization.py
```diff
@@ -4,20 +4,103 @@
 # Licensed under the MIT License. See License.txt in the project root for
 # license information.
 # --------------------------------------------------------------------------
+import base64
+import datetime
+from json import JSONEncoder
+
+from .utils._utils import _FixedOffset
```
johanste, line 11: Do we really need to import a private type from a private submodule?

xiangyan99: I think we should rename _FixedOffset to FixedOffset. But not sure if it is worth...

sdk/core/azure-core/azure/core/serialization.py
```diff
@@ -4,20 +4,103 @@
...
+    if bigger_exists:
+        time += "{:02}H".format(hours)
+
+    # minutes
+    bigger_exists = bigger_exists or minutes
+    if bigger_exists:
+        time += "{:02}M".format(minutes)
+
+    # seconds
+    if seconds.is_integer():
```
johanste, line 66: Can't seconds be an `int` (as opposed to a `float`) - in which case, there is no `is_integer` method in Python 2.7..

sdk/core/azure-core/dev_requirements.txt
mccoyp: Yep, this would only be for testing. I could also try to work `isodate` out of the tests -- it's convenient but not necessary

sdk/core/azure-core/azure/core/serialization.py
```diff
@@ -4,20 +4,116 @@
...
+    try:
+        if seconds.is_integer():
+            seconds_string = "{:02}".format(int(seconds))
+        else:
+            # 9 chars long w/ leading 0, 6 digits after decimal
+            seconds_string = "%09.6f" % seconds
+            # Remove trailing zeros
+            seconds_string = seconds_string.rstrip("0")
+    except AttributeError:  # int.is_integer() raises on Python 2.7
+        seconds_string = "{:02}".format(seconds)
```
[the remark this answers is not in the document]
mccoyp: I wouldn't think so, since this AttributeError should only come up if `seconds` is an integer. `seconds.is_integer()` works when `seconds` is a float on 2.7. 3.6, and 3.9 when I test locally

annatisch (approved):

PR #19602, opened 2021-06-30
Credentials accept tenant_id argument to get_token

### LINKED ISSUE #17979: Add support for tenant discovery in ClientCredentials
Currently, for services that return challenges which include context about the tenant that can service a failed service request, such as `authorization_url`,
 Azure.Identity does not utilize this information to request a new token with the correct tenantId.

 This is a proposal to enhance Azure.Identity to make it possible for service clients to interpret challenges to request tokens for the correct tenant Id, where possible.

The proposal details can be found in [this gist](https://gist.github.com/christothes/34b713a70aad4e94f228b5a9b64f92a4)

### LINKED ISSUE #19300: Add support for tenant Id Challenges / tenant discovery in ClientCredentials
Currently, for services that return challenges which include context about the tenant that can service a failed service request, such as `authorization_url`,
 Azure.Identity does not utilize this information to request a new token with the correct tenantId.

 This is a proposal to enhance Azure.Identity to make it possible for service clients to interpret challenges to request tokens for the correct tenant Id, where possible.

The proposal details can be found in [this gist](https://gist.github.com/christothes/34b713a70aad4e94f228b5a9b64f92a4)

Related: https://github.com/Azure/azure-sdk-for-net/issues/20302

chlowell:
This is the first step in enabling clients to implement tenant discovery via authentication challenges. Credentials which can in principle authenticate in multiple tenants (i.e. not ManagedIdentityCredential) receive a new parameter, `allow_multitenant_authentication`. When this is True, these credentials will honor a `tenant_id` keyword argument to `get_token`. When it's False, which is the default, `get_token` raises an exception when given a tenant that doesn't match the credential's configured tenant (i.e. the one set at instantiation), unless environment variable AZURE_IDENTITY_ENABLE_LEGACY_TENANT_SELECTION is set (because raising this exception may be a breaking change for some applications). When that environment variable is set `get_token` will use the credential's configured tenant, ignoring a `tenant_id` keyword argument.

Closes #17979, closes #19300


sdk/identity/azure-identity/azure/identity/aio/_credentials/azure_cli.py
```diff
@@ -31,8 +31,15 @@ class AzureCliCredential(AsyncContextManager):
...
 
     This requires previously logging in to Azure via "az login", and will use the CLI's currently logged in identity.
+
+    :keyword bool allow_multitenant_authentication: when True, enables the credential to acquire tokens from any tenant
+        the identity logged in to the Azure CLI is registered in. When False, which is the default, the credential will
+        acquire tokens only from the tenant of the Azure CLI's active subscription.
     """
 
+    def __init__(self, **kwargs):
+        self._allow_multitenant = kwargs.get("allow_multitenant_authentication", False)
```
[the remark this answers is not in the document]
chlowell: No need, `kwargs` isn't used again. I should use keyword-only syntax here though.

sdk/identity/azure-identity/azure/identity/_internal/__init__.py
```diff
@@ -43,6 +49,24 @@ def validate_tenant_id(tenant_id):
         )
 
 
+def resolve_tenant(default_tenant, allow_multitenant, tenant_id=None, **_):
```
chlowell, line 52: @christothes I believe this is equivalent to your `TenantIdResolver` but please take a look, maybe I misread that (C# pattern matching syntax is new to me).

sdk/identity/azure-identity/azure/identity/_internal/interactive.py
```diff
@@ -79,7 +84,7 @@ def _build_auth_record(response):
         six.raise_from(auth_error, ex)
 
 
-class InteractiveCredential(MsalCredential):
+class InteractiveCredential(MsalCredential, ABC):
```
mccoyp, line 87: What's the reasoning for making InteractiveCredential an ABC instead of MsalCredential? I'm not too familiar with the purpose of inheriting from ABC

chlowell: A class inheriting ABC can only be instantiated if it implements every method the ABC decorates with `abstractmethod`. That's useful here because it enables InteractiveCredential.get_token to consolidate common code and do most of the work for subclasses while ensuring you get a runtime error when a subclass doesn't implement the one method that differs between authentication flows. With this change MsalCredential loses its only abstract method and thus its reason for inheriting ABC. InteractiveCredential still has an abstract method but was inheriting ABC through MsalCredential. So, now it must inherit ABC directly.

mccoyp (approved):

PR #19604, opened 2021-06-30
add aad support

xiangyan99:


sdk/search/azure-search-documents/azure/search/documents/_utils.py
```diff
@@ -3,6 +3,14 @@
...
+from azure.core.pipeline import policies
+
+CREDENTIAL_SCOPES = ['https://search.azure.com/.default']
+
 def is_retryable_status_code(status_code):
     # type: (int) -> bool
     return status_code in [422, 409, 503]
+
+def get_authentication_policy(credential):
+    authentication_policy = policies.BearerTokenCredentialPolicy(credential, CREDENTIAL_SCOPES)
```
kristapratico, line 15: is it possible that a user might want to pass their own scopes? For different clouds, e.g. govt or mooncake? Wondering if scopes should be a kwarg

xiangyan99: For now (at least in this preview), no. It is hard-coded by the search team.

kristapratico (approved):

PR #19625, opened 2021-07-01
Update Validate-Formatting to use `black` output vs `git diff`

scbedd:
[Here is a failing build.](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=977321&view=logs&j=b70e5e73-bbb6-5567-0939-8415943fadb9&t=23d29da4-00cb-5783-b17c-96bda25d9c52) the reason this is happening is because the script currently will break on anything being different at all. Unfortunately the `template` pipeline actually modifies the `version` and the `changelog`, so this means that the `validate formatting` finds a git diff and explodes.

Do you have any issues with this @seankane-msft ?

### DIFF
--- a/scripts/devops_tasks/validate_formatting.py
+++ b/scripts/devops_tasks/validate_formatting.py
@@ -9,44 +9,35 @@ import argparse
 import os
 import logging
 import sys
-
-from common_tasks import run_check_call
+import subprocess
+import pdb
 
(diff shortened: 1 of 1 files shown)


seankane-msft (approved):
LGTM :shipit:

danieljurek (approved):

PR #19627, opened 2021-07-01
Update Test Timout from 0 to 60

scbedd:
This was fine on hosted agents, as the defaults when `0` is passed are much lower.

On self-hosted machines, it's forever. Can't have that!

@weshaggard I adjusted this thing down to 60 (where it was 120), but individually bumped `cognitiveservices` and `storage` to 120, as they DO regularly go above 60 for the nightly run.


weshaggard (approved):
We should consider 60 mins but otherwise looks good.

weshaggard (approved):

PR #19646, opened 2021-07-02
[textanalytics] move async file to aio

kristapratico:
Moves _response_handlers_async.py into the `aio` directory


deyaaeldeen (approved):
Not sure what that is about but I am approving anyway 😄

maririos (approved):

PR #19664, opened 2021-07-06
[ACR] Fixing credential_scopes kwarg

seankane-msft:


sdk/containerregistry/azure-containerregistry/tests/test_anon_access.py
```diff
@@ -139,6 +169,9 @@ def test_update_repository_properties(self, containerregistry_anonregistry_endpo
 
     @acr_preparer()
     def test_update_tag_properties(self, containerregistry_anonregistry_endpoint):
+        if not self.is_public_endpoint(containerregistry_anonregistry_endpoint):
```
iscai-msft, line 172: I think you can do a `pytest.mark.skipif` instead, even though you're looking at a `self` parameter

seankane-msft: I get a `NameError: name 'self' is not defined` if I do this, also the parameter is passed in by the preparer so it wouldn't be available either.

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_exchange_client.py
```diff
@@ -43,13 +43,12 @@ def __init__(self, endpoint, credential, **kwargs):
         if not endpoint.startswith("https://") and not endpoint.startswith("http://"):
             endpoint = "https://" + endpoint
         self._endpoint = endpoint
-        self.credential_scope = kwargs.get("authentication_scope", "https://management.core.windows.net/.default")
+        self.credential_scopes = kwargs.get("credential_scopes", ["https://management.core.windows.net/.default"])
```
kristapratico, line 46: should this be a private attr?

kristapratico (approved):

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_exchange_client.py
seankane-msft: It probably should, but the `ExchangeClient` models are not exposed publicly

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_exchange_client.py
```diff
@@ -74,7 +73,7 @@ def get_refresh_token(self, service, **kwargs):
     def exchange_aad_token_for_refresh_token(self, service=None, **kwargs):
         # type: (str, Dict[str, Any]) -> str
         refresh_token = self._client.authentication.exchange_aad_access_token_for_acr_refresh_token(
-            service=service, access_token=self._credential.get_token(self.credential_scope).token, **kwargs
+            service=service, access_token=self._credential.get_token(*self.credential_scopes).token, **kwargs
```
mccoyp, line 76: Is `credential_scopes` an iterable (the name implies that it is)? If so, does the keyword paramtype need to be updated from `str` to something like `List[str]`?

seankane-msft: Yes, it should be a `List[str]`.

swathipil (approved):
lgtm

mccoyp (approved):

PR #19677, opened 2021-07-06
[textanalytics] prepare release

kristapratico:
Fixes some docs, unexposes `to_generated` on *Action types


iscai-msft (approved):
great catch, thank you queen

PR #19689, opened 2021-07-07
Add check for tracing_enabled setting

helsaawy:
Updating `DistributedTracingPolicy` to check that tracing is enabled along with having an implementation type present before tracing a request


sdk/core/azure-core/azure/core/pipeline/policies/_distributed_tracing.py
```diff
@@ -85,7 +85,7 @@ def on_request(self, request):
         ctxt = request.context.options
         try:
             span_impl_type = settings.tracing_implementation()
-            if span_impl_type is None:
+            if (not settings.tracing_enabled()) or (span_impl_type is None):
```
rakshith91, line 88: would you be able to add a test?

rakshith91:
Also, tagging @johanste - should this be considered a breaking change?

PR #19700, opened 2021-07-07
Generated Queues for new Swagger

tasherif-msft:


sdk/storage/azure-storage-queue/azure/storage/queue/_generated/_azure_queue_storage.py
```diff
@@ -62,6 +64,24 @@ def __init__(
         self.message_id = MessageIdOperations(
             self._client, self._config, self._serialize, self._deserialize)
 
+    def _send_request(self, http_request, **kwargs):
```
xiafu-msft, line 67: where are we gonna call this from?

PR #19704, opened 2021-07-07
apply aiohttp work around for vcr

xiangyan99:
in vcr playback mode, response.content creates a new mock stream every time.

Apply a work around to only load it once.


iscai-msft (approved):

xiangyan99:
No need any more.

Issue #19552, opened 2021-06-28
[core] improve repr of AzureError


iscai-msft:
Currently we don't have a very nice repr for [`AzureError`](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/core/azure-core/azure/core/exceptions.py#L242), we just inherit from `Exception`'s repr. 

We should add a repr that takes advantage of `AzureError`'s properties, instead of just "message".

PR #19708, opened 2021-07-07
add repr for AzureError

xiangyan99:
#19552


sdk/core/azure-core/azure/core/exceptions.py
```diff
@@ -249,6 +249,13 @@ def raise_with_traceback(self):
             self.__traceback__ = self.exc_traceback
             raise self
 
+    def __repr__(self):
+        class_name = self.__class__.__name__
+
+        return "{}(message={})".format(
+            class_name, self.message
+        )[:1024]
```
annatisch, line 257: How does this compare the output of `__repr__` for the default Exception? How does it compare to the output of `str()`?
Just curious :)

PR #19710, opened 2021-07-07
[Storage][Logging]Let users opt in logging_request_body

xiafu-msft:


sdk/storage/azure-storage-blob/azure/storage/blob/_shared/policies.py
```diff
@@ -216,11 +218,11 @@ def on_request(self, request):
                     _LOGGER.debug("    %r: %r", header, value)
                 _LOGGER.debug("Request body:")
 
-                # We don't want to log the binary data of a file upload.
-                if isinstance(http_request.body, types.GeneratorType):
-                    _LOGGER.debug("File upload")
-                else:
+                if self.logging_request_body or options.pop("logging_request_body", False):
```
xiafu-msft, line 221: I didn't touch on_response because logging response body is handled pretty well.
For download blob, since we are doing streaming download, no actually body will be logged. For other responses logging response body is helpful, since the response body could contains raw error message or some xml elements we want to check.
Let me know if you think we should disable logging for any kinds of response body .

kasobol-msft: We should make it symetric and possibly consider single switch for both request and response like .net and java do.
.net https://github.com/Azure/azure-sdk-for-net/search?p=1&q=IsLoggingContentEnabled&type=
java https://github.com/Azure/azure-sdk-for-java/blob/3f31d68eed6fbe11516ca3afe3955c8840a6e974/sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpLogDetailLevel.java

kasobol-msft: besides some may want to disable body logging due to PII being present inside.

kasobol-msft (commented):
should we add test coverage ?

sdk/storage/azure-storage-blob/azure/storage/blob/_shared/policies.py
```diff
@@ -216,11 +219,11 @@ def on_request(self, request):
...
 
-                # We don't want to log the binary data of a file upload.
-                if isinstance(http_request.body, types.GeneratorType):
-                    _LOGGER.debug("File upload")
-                else:
+                if self.logging_body:
                     _LOGGER.debug(str(http_request.body))
+                else:
+                    # We don't want to log the binary data of a file upload.
+                    _LOGGER.debug("Hidden body, please use logging_body to show body")
```
kasobol-msft, line 226: I'd rather delete this else branch. It's going to produce lot of entries that are not very useful.
Instead consider adding logger.debug in ctor/ctors if logging is enabled but body logging is not. so that there's some hint about that option but is not rendered on each request.

sdk/storage/azure-storage-blob/azure/storage/blob/_shared/policies.py
```diff
@@ -240,19 +243,24 @@ def on_response(self, request, response):
...
 
                 if header and pattern.match(header):
                     filename = header.partition('=')[2]
                     _LOGGER.debug("File attachments: %s", filename)
-                elif response.http_response.headers.get("content-type", "").endswith("octet-stream"):
+                elif resp_content_type.endswith("octet-stream"):
                     _LOGGER.debug("Body contains binary data.")
-                elif response.http_response.headers.get("content-type", "").startswith("image"):
+                elif resp_content_type.startswith("image"):
                     _LOGGER.debug("Body contains image data.")
```
kasobol-msft, line 254: we could consider dumbing binary payloads in encoded (base64 for example) form. (not blocking this PR).

kasobol-msft (approved):

PR #19711, opened 2021-07-08
leave stream open in response

xiangyan99:


tools/vcrpy/vcr/stubs/aiohttp_stubs/__init__.py
```diff
@@ -71,7 +73,7 @@ def build_response(vcr_request, vcr_response, history):
     response._headers = CIMultiDictProxy(CIMultiDict(vcr_response["headers"]))
     response._history = tuple(history)
 
-    response.close()
+    # response.close()
```
annatisch, line 76: Can this just be removed? Or are we still testing?

xiangyan99: In fact, here we applied a patch to vcr 3.0.0. I keep it just to easier track our change. :)

xiangyan99: Removed.

annatisch (approved):

weshaggard:
@xiangyan99 the path filters end up blocking devops from triggering some of these other pipelines. You can work around this issue by tringgering the non-ci version of the pipeline example `/azp run python - keyvault`

PR #19731, opened 2021-07-08
Exclude Latest Version of azure-communication-phonenumbers from regression

scbedd:
Due to imcompatible tests. @xiangyan99 

Checking for others that have the same issue.

- [Run with just phonenumbers out](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=990350&view=results)

### DIFF
--- a/scripts/devops_tasks/git_helper.py
+++ b/scripts/devops_tasks/git_helper.py
@@ -27,7 +27,8 @@ EXCLUDED_PACKAGE_VERSIONS = {
     'azure-storage-blob-changefeed' : '12.0.0b2',
     'azure-storage-file-datalake': '12.2.0b1',
     'azure-communication-identity': '1.0.0',
-    'azure-communication-phonenumbers': '1.0.1'
+    'azure-communication-phonenumbers': '1.0.0',
+    'azure-communication-sms': '1.0.0'
 }
 
 # This method identifies release tag for latest or oldest released version of a given package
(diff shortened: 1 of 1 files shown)


xiangyan99 (approved):

scripts/devops_tasks/git_helper.py
```diff
@@ -26,7 +26,8 @@
     'azure-schemaregistry-avroserializer': '1.0.0b1',
     'azure-storage-blob-changefeed' : '12.0.0b2',
     'azure-storage-file-datalake': '12.2.0b1',
-    'azure-communication-identity': '1.0.0'
+    'azure-communication-identity': '1.0.0',
+    'azure-communication-phonenumbers': '1.0.1'
```
xiangyan99, line 30: Just double check, shouldn't it be 1.0.0?

scbedd: You asked me to exclude the current versions from regression, which does mean `1.0.1`. I'll reset it to `1.0.0` and see if it works.

scbedd: [Unfortunately looks like sms has the same issue](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=990350&view=logs&j=3fadc984-4acf-5cfe-b896-a35ffcc3efec&t=fda75cf5-3a75-5173-d794-9c1135621789&l=1031) so I'll do a pass of `1.0.0` on both. See what pans out.

xiangyan99: Given this is in their beartokenpolicy, I think all communication packages have the same issue.

iscai-msft (approved):

Issue #15222, opened 2020-11-11
UnboundLocalError when SansIOHTTPPolicy handles an exception


chlowell:
_SansIOHTTPPolicyRunner allows a SansIOHTTPPolicy to stop it propagating pipeline exceptions, by returning `True` from its [on_exception](https://github.com/Azure/azure-sdk-for-python/blob/82175dc661d41a2cd03bf1ac4d2519c04bf61a15/sdk/core/azure-core/azure/core/pipeline/policies/_base.py#L111-L116) method. However, when a policy does this, _SansIOHTTPPolicyRunner raises UnboundLocalError:

```py
from azure.core.pipeline import Pipeline
from azure.core.pipeline.transport import HttpRequest
from azure.core.pipeline.policies import HTTPPolicy, SansIOHTTPPolicy

class ReproPolicy(SansIOHTTPPolicy):
    def on_exception(self, request):
        return True

class ErrorPolicy(HTTPPolicy):
    def send(self, request):
        raise Exception("oops")

pipeline = Pipeline(transport=object(), policies=[ReproPolicy(), ErrorPolicy()])
pipeline.run(HttpRequest("GET", "..."))
```

```
Traceback (most recent call last):
  ...
  File "c:\code\azure-sdk-for-python\sdk\core\azure-core\azure\core\pipeline\_base.py", line 77, in send
    return response
UnboundLocalError: local variable 'response' referenced before assignment
```

... which makes sense because an exception prevents the assignment:

https://github.com/Azure/azure-sdk-for-python/blob/82175dc661d41a2cd03bf1ac4d2519c04bf61a15/sdk/core/azure-core/azure/core/pipeline/_base.py#L70-L77

I opened an issue rather than a PR because the fix requires some design. Should _SansIOHTTPPolicyRunner return a PipelineResponse with no `http_response` in this case?

chlowell:
A fix for this should apply to BearerTokenCredentialPolicy as well. That policy began as a subclass of SansIOHTTPPolicy and now follows its behavior for compatibility reasons.

PR #19744, opened 2021-07-09
fix UnboundLocalError

xiangyan99:
#15222


sdk/core/azure-core/azure/core/pipeline/_base.py
```diff
@@ -74,7 +75,7 @@ def send(self, request):
                 raise
         else:
             _await_result(self._policy.on_response, request, response)
-        return response
+        return response # type: ignore
```
chlowell, line 78: mypy is right to warn about this though. The annotation is wrong: this method now returns `Optional[PipelineResponse]`

chlowell (changes requested):
This prevents the UnboundLocalError by making a breaking change: a method that either returned a PipelineResponse instance or raised an exception may now return None instead. All policies and clients will need to consider that possibility. Autorest clients aren't ready for it:

https://github.com/Azure/azure-sdk-for-python/blob/863f6876452de4fdd5be962b11b679e9768d415c/sdk/keyvault/azure-keyvault-administration/azure/keyvault/administration/_generated/v7_2/operations/_role_assignments_operations.py#L94-L95

sdk/core/azure-core/azure/core/pipeline/_base.py
xiangyan99: Yes. But unfortunately, changing return type from PipelineResponse to Optional[PipelineResponse] is a breaking change.

It will break mypy for its consumers.

chlowell: The code change itself is breaking. Preventing mypy from warning people about it doesn't help.

xiangyan99: The behavior change is from raise -> return None. This only breaks non-working code. (in 99% of the case)

mypy breaking will be more impactful and breaks working code.

chlowell: It doesn't make that non-working code work though. It just replaces an UnboundLocalError in this method with an AttributeError somewhere else because none of our code expects Pipeline.run() to return None. Broken code stays broken while working code gains a new failure mode hidden from mypy.

PR #19746, opened 2021-07-09
Add context manager API to azure.identity credentials

### LINKED ISSUE #18798: Enable opening and closing synchronous credential transports
Sync credentials should have `__enter__`, `__exit__` and `close` methods which open/close their internal Pipelines.

chlowell:
Closes #18798


sdk/identity/azure-identity/tests/test_managed_identity.py
```diff
@@ -2,6 +2,7 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
 # ------------------------------------
+from azure.identity._credentials import managed_identity
```
mccoyp, line 5: Was this auto-imported by accident? I know VS Code has done this to me a few times

sdk/identity/azure-identity/tests/test_vscode_credential.py
```diff
@@ -84,6 +84,26 @@ def test_no_scopes():
...
+def test_context_manager():
+    transport = mock.MagicMock()
+    credential = get_credential(transport=transport)
+
+    with credential:
+        assert transport.__enter__.call_count == 1
+
+    assert transport.__enter__.call_count == 1
+    assert transport.__exit__.call_count == 1
+
```
mccoyp, line 105: It looks like a VisualStudioCodeCredential can be an a state where `self._client` is None and the `__enter__` and `__exit__` methods account for this. Should there be sync and async incomplete configuration tests, or is the idea with those tests to just test environment-based credentials in unsupported environments?

sdk/identity/azure-identity/tests/test_managed_identity.py
chlowell: I think so. It's not a line of code I would write :suspect:

sdk/identity/azure-identity/tests/test_vscode_credential.py
chlowell: There are two doomed configuration scenarios: ADFS and custom cloud with no authority. I see only the former has a test, so let me add one for the latter. Not sure I follow your question though. Do you mean, should we test the context management API as well in those scenarios?

mccoyp: I had meant that -- the scenario seemed similar to what's being tested in managed identity `incomplete_configuration` tests -- but if an exception is being raised in these scenarios then it doesn't actually make sense. We got more test coverage as a result of my mistaken thinking though, so that's cool 😎

mccoyp (approved):

sdk/identity/azure-identity/tests/test_managed_identity.py
```diff
@@ -21,6 +21,56 @@
...
+        EnvironmentVariables.TOKEN_FILE_PATH: __file__,
+    },
+    {},  # IMDS
+)
+
+
+@pytest.mark.parametrize("environ", ALL_ENVIRONMENTS)
+def test_close(environ):
+    transport = mock.MagicMock()
+    with mock.patch.dict("os.environ", environ, clear=True):
```
mccoyp, line 45: Did you want to patch `{Credential}.__module__ + ".os.environ"` across the board, or just in async EnvironmentCredential tests?

chlowell: I would like to use more narrow patches everywhere. It doesn't work here because the inner credentials need to observe the same environment variables. Since broad patches don't impact our tests as currently run, I prefer using one here over baking more knowledge of the implementation into this test or testing the inner credentials separately.

mccoyp (approved):

PR #19748, opened 2021-07-09
[EventHub] basic receive event scenario with pyamqp

swathipil:


sdk/eventhub/azure-eventhub/azure/eventhub/_client_base.py
```diff
@@ -306,7 +307,11 @@ def _create_auth(self):
                 port=self._config.connection_port,
                 verify=self._config.connection_verify
             )
-            auth.update_token()
+            # TODO: determine if below should move to `def update_token()` method
+            access_token = auth.get_token()
+            self.expires_at = access_token.expires_on
+            value = access_token.token
+            self.token = value.encode('utf-8') if isinstance(value, six.text_type) else value
```
yunhaoling, line 314: would the code work for authentication? I see we're setting `self.expires_at/self.token` and in which context `self` is `ClientBase`. and to your question, I think it would better be moved into `update_token` method.

btw, what I'm doing in my PR is just letting the EH Client to create the authentication object, the amqp library would handle the token for the eh client.
```
from .pyamqp.authentication import JWTTokenAuth as PyJWTTokenAuth
#...
if token_type == b"servicebus.windows.net:sastoken":
    return PyJWTTokenAuth(self._auth_uri, self._auth_uri, functools.partial(self._credential.get_token, self._auth_uri))
```

sdk/eventhub/azure-eventhub/azure/eventhub/_client_base.py
```diff
@@ -419,9 +437,11 @@ def _add_span_request_attributes(self, span):
     def _get_eventhub_properties(self):
         # type:() -> Dict[str, Any]
         mgmt_msg = Message(application_properties={"name": self.eventhub_name})
-        response = self._management_request(mgmt_msg, op_type=MGMT_OPERATION)
+        # TODO: remove mocked response
+        #response = self._management_request(mgmt_msg, op_type=MGMT_OPERATION)
         output = {}
-        eh_info = response.get_data()  # type: Dict[bytes, Any]
+        eh_info = {b"name": b"eventhub-test", b"created_at": 1608181432000, b"partition_ids": [b'0']}
```
yunhaoling, line 443: we could set it to the array as following to mock receiving from multiple partions:
```python
b"partition_ids": [b'0', b'1', b'2', b'3', b'4', b'5', b'6', b'7', b'8', b'9', b'10', b'11', b'12', b'13', b'14', b'15', b'16', b'17', b'18', b'19', b'20', b'21', b'22', b'23', b'24', b'25', b'26', b'27', b'28', b'29', b'30', b'31']
```

(32 is the max allowed partition count that could be created on azure portal)

sdk/eventhub/azure-eventhub/azure/eventhub/_consumer.py
```diff
@@ -128,37 +134,42 @@ def __init__(self, client, source, **kwargs):
...
     def _create_handler(self, auth):
         # type: (JWTTokenAuth) -> None
-        source = Source(self._source)
+        source = Source(address=self._source, filters={})
         if self._offset is not None:
-            source.set_filter(
-                event_position_selector(self._offset, self._offset_inclusive)
-            )
+            # set_filter - from which event should I start receiving?
+            # TODO: move set_filter stuff to separate method?
```
yunhaoling, line 140: does setting filter work now? if not then let's add a TODO

sdk/eventhub/azure-eventhub/azure/eventhub/_consumer.py
```diff
@@ -188,19 +199,26 @@ def _open(self):
...
+            hostname = self._client._address.hostname
+            uri = "sb://{}/{}".format(hostname, eventhub_name)
+            key = self._client._credential.key
+            policy = self._client._credential.policy
+            auth = SASTokenAuth(
+                uri=uri,
+                audience=uri,
+                username=policy,
+                password=key
             )
```
yunhaoling, line 214: we could reuse `self._create_auth` to create the authentication object that is required.

yunhaoling (commented):
awesome work -- left some minor comments
also don't forget to remove .c/.pyx

sdk/eventhub/azure-eventhub/azure/eventhub/_client_base.py
swathipil: ah maybe that's why using JWTTokenAuth wasn't working. I switched to SASTokenAuth since it wasn't working but will try this. Thanks for the snippet!

sdk/eventhub/azure-eventhub/azure/eventhub/_consumer.py
swathipil: for some reason JWTTokenAuth wasn't working, it was on my TODO list. however, I'll try your snippet above

yunhaoling (approved):

PR #19785, opened 2021-07-13
Use `pip==20.3`, Remove pinned min-dependency override on `cryptography`

scbedd:
Need to  double check the `storage` tests as well.

If this doesn't work, we're kinda out of luck for a simple fix. Will need to update `install_depend_packages` to handle transitive as well I think. 

We cannot go newer than `20.3` due to the fact that `21.0` drops support for `py2.7`. Need to  double check the `storage` tests as well.

[Identity](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=996808&view=logs&j=e2ec4050-08de-5f03-3408-7f9baa73d35a&t=b763cfd9-933a-5b3d-402a-3832ab180e1f)
[Storage](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=996806&view=results)
[Additional Storage Run](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=996993&view=results)

Looks like identity and storage are ok, but some core packages are having issues with the new resolver. Investigating.


scbedd:
Existing `appconfiguration` failure is already present.

scripts/devops_tasks/git_helper.py
```diff
@@ -28,7 +28,8 @@
     'azure-storage-file-datalake': '12.2.0b1',
     'azure-communication-identity': '1.0.0',
     'azure-communication-phonenumbers': '1.0.0',
-    'azure-communication-sms': '1.0.0'
+    'azure-communication-sms': '1.0.0',
+    'azure-search-documents': '11.2.0b3',
```
xiangyan99, line 32: To make sure we are on same page, we want to skip 11.0.0, 11.1.0 and all betas. We want to start with 11.2.0. Is this correct?

scbedd: @xiangyan99 that's exactly what this does. First allowed version is `11.2.0`.

seankane-msft (approved):
For ACR and Tables, LGTM :shipit:

xiangyan99:
app config failure is because of a known issue and Izzy has a PR to fix it.

https://github.com/Azure/azure-sdk-for-python/pull/19758

PR #19803, opened 2021-07-14
[Tables] updating entities in samples to use more edmtypes

### LINKED ISSUE #18732: [Tables] Update samples to include a wider range of EdmTypes
Currently the samples for creating and updating entities only use string properties in those entities.
We should update the samples to show the full range of EdmTypes in a Python context - both as input and deserialized output.

Specifically these samples:
https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/tables/azure-data-tables/samples/sample_insert_delete_entities.py
https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/tables/azure-data-tables/samples/sample_update_upsert_merge_entities.py

The readme already has a more expansive sample entity - maybe we can leverage this more widely.

seankane-msft:
closes #18732

### DIFF
--- a/sdk/tables/azure-data-tables/samples/async_samples/sample_insert_delete_entities_async.py
+++ b/sdk/tables/azure-data-tables/samples/async_samples/sample_insert_delete_entities_async.py
@@ -48,7 +48,7 @@ class InsertDeleteEntity(object):
             "last_updated": datetime.today(),
             "product_id": uuid4(),
             "inventory_count": 42,
-            "some_binary_value": b"somebinaryvalue"
+            "barcode": b"135aefg8oj0ld58"
         }
 
     async def create_entity(self):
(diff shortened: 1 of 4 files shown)


sdk/tables/azure-data-tables/samples/async_samples/sample_insert_delete_entities_async.py
```diff
@@ -37,7 +39,17 @@ def __init__(self):
...
+        self.entity = {
+            "PartitionKey": "color",
+            "RowKey": "brand",
+            "text": "Marker",
+            "color": "Purple",
+            "price": 4.99,
+            "last_updated": datetime.today(),
+            "product_id": uuid4(),
+            "inventory_count": 42,
+            "some_binary_value": b"somebinaryvalue"
```
annatisch, line 51: Haha - It looks like the marker merchandise theme petered out a bit :P 
How about:
```python
"barcode": b"somebinaryvalue"
```

annatisch (commented):
Thanks for doing this :)

sdk/tables/azure-data-tables/samples/async_samples/sample_insert_delete_entities_async.py
seankane-msft: Haha thank you I could not think of a way to use binary with the marker theme.

annatisch (approved):

PR #19808, opened 2021-07-15
aiohttp should not depend on requests

xiangyan99:


sdk/core/azure-core/azure/core/pipeline/transport/_aiohttp.py
```diff
@@ -247,8 +246,6 @@ async def __anext__(self):
         except _ResponseStopIteration:
             internal_response.close()
             raise StopAsyncIteration()
-        except StreamConsumedError:
-            raise
```
annatisch, line 251: Does this need to be replaced by anything?
....probably not because we were just reraising it, and a requests error never would have been raised by aiohttp anyway.... did aiohttp have an equivalent need to handle?

xiangyan99: Replacing (if needed) would be a potential breaking change. For this PR, I want to fix the for-sure bad code to unblock the work to cut dependency on requests library. Does this make sense?

annatisch (approved):

PR #19814, opened 2021-07-15
[POC] Blobs Partial list deserialization

annatisch:


sdk/storage/azure-storage-blob/azure/storage/blob/_shared/base_client.py
```diff
@@ -237,21 +253,22 @@ def _create_pipeline(self, credential, **kwargs):
...
             QueueMessagePolicy(),
+            config.headers_policy,
             config.proxy_policy,
             config.user_agent_policy,
             StorageContentValidation(),
-            ContentDecodePolicy(response_encoding="utf-8"),
             RedirectPolicy(**kwargs),
             StorageHosts(hosts=self._hosts, **kwargs),
             config.retry_policy,
-            config.headers_policy,
```
xiafu-msft, line 247: just a note: headers_policy cannot be put before retry, since we need to regenerate the timestamp in header

annatisch: Thanks @xiafu-msft - I wondered about this! Why was it different between the sync and async pipelines?

sdk/storage/azure-storage-blob/azure/storage/blob/_shared/xml_deserialization.py
```diff
@@ -0,0 +1,587 @@
...
+        :param str content_type: Swagger "produces" if available.
+        """
+        try:
+            return self(target_obj, data, content_type=content_type)
+        except:  # pylint: disable=bare-except
+            _LOGGER.warning(
+                "Ran into a deserialization error. Ignoring since this is failsafe deserialization",
+				exc_info=True
+            )
+            return None
```
xiafu-msft, line 479: It seems this not decoding the response body? ContentDecodingPolicy was turning body which describes the error into string, so probably we want to do the same thing?

sdk/storage/azure-storage-blob/azure/storage/blob/_shared/base_client.py
xiafu-msft: thanks Anna @annatisch if async is having different order then it's a bug...

sdk/storage/azure-storage-blob/azure/storage/blob/_shared/xml_deserialization.py
xiafu-msft: also some response body on error is in json format, not sure if it will be a problem

sdk/storage/azure-storage-blob/azure/storage/blob/_list_blobs_helper.py
```diff
@@ -7,14 +7,70 @@
...
+        blob.deleted = load_xml_bool(element, 'Deleted')
+    if 'snapshot' in select:
+        blob.snapshot = load_xml_string(element, 'Snapshot')
+    if 'version' in select:
+        blob.version_id = load_xml_string(element, 'VersionId')
+        blob.is_current_version = load_xml_bool(element, 'IsCurrentVersion')
+    return blob
+
+
+class BlobPropertiesPaged(PageIterator):  # pylint: disable=too-many-instance-attributes
```
annatisch, line 73: It looks like I updated the existing BlobPropertiesPaged - which means the perf of this model would be improved, however if we wanted to leave the original `list_blobs` API _completely_ untouched, we could revert the changes here and have the new `list_blob_names` API use it's own custom Paged object.

sdk/storage/azure-storage-blob/azure/storage/blob/_shared/base_client.py
```diff
@@ -74,6 +75,7 @@ def __init__(
         # type: (...) -> None
         self._location_mode = kwargs.get("_location_mode", LocationMode.PRIMARY)
         self._hosts = kwargs.get("_hosts")
+        self._msrest_xml = kwargs.get('msrest_xml', False)
```
annatisch, line 78: We'd want to remove these for now most likely.
Though we could in future look at doing some kind of 'opt-in' flag to opt out of msrest.
Might be best to leave that decision for now until the long-term deprecation story for msrest starts developing. Then we can start planning more realistically.

sdk/storage/azure-storage-blob/azure/storage/blob/_shared/base_client.py
```diff
@@ -237,21 +253,22 @@ def _create_pipeline(self, credential, **kwargs):
             config.transport = RequestsTransport(**kwargs)
         policies = [
             QueueMessagePolicy(),
+            config.headers_policy,
```
annatisch, line 256: Not sure what these changes in the pipeline policy order were all about. We should probably ignore this for now.
I remove the ContentDecode policy below because this was prematurely decoding the XML, however I don't _think_ messing around with the pipeline will be necessary for doing just a `list_blob_names` API.

sdk/storage/azure-storage-blob/azure/storage/blob/_generated/aio/operations/_container_operations.py
```diff
@@ -1452,12 +1452,12 @@ async def list_blob_flat_segment(
         response_headers['x-ms-request-id']=self._deserialize('str', response.headers.get('x-ms-request-id'))
         response_headers['x-ms-version']=self._deserialize('str', response.headers.get('x-ms-version'))
         response_headers['Date']=self._deserialize('rfc-1123', response.headers.get('Date'))
-        deserialized = self._deserialize('ListBlobsFlatSegmentResponse', pipeline_response)
+        #deserialized = self._deserialize('ListBlobsFlatSegmentResponse', pipeline_response)
```
annatisch, line 1455: This is one of the bigger challenges to figure out. Currently the deserialization process is out of our hands. We would need to add directives to the autorest code gen for the list_blobs_flat_segment API to _not_ deserialize the response payload. This should be possible by simply overwriting the output model to have no output. We then use the `cls` hook and do the deserialization ourselves.
In the case of the existing `list_blobs` API, this probably means manually using the existing msrest deserializer if we don't want to deal with the testing burden of validating the new deserializer for the old API.

sdk/storage/azure-storage-blob/azure/storage/blob/_blob_client.py
```diff
@@ -175,6 +176,8 @@ def __init__(
         self._query_str, credential = self._format_query_string(sas_token, credential, snapshot=self.snapshot)
         super(BlobClient, self).__init__(parsed_url, service='blob', credential=credential, **kwargs)
         self._client = AzureBlobStorage(self.url, pipeline=self._pipeline)
+        if not self._msrest_xml:
+            self._custom_xml_deserializer(generated_models)
```
annatisch, line 180: Removable. Same does for the all the clients.

sdk/storage/azure-storage-blob/azure/storage/blob/_list_blobs_helper.py
```diff
@@ -7,14 +7,70 @@
...
+        element = load_single_node(element, wrapper)
+    return list(element.findall(name))
+
+
+def blob_properties_from_xml(element, select, deserializer):
+    if not select:
+        generated = deserializer.deserialize_data(element, 'BlobItemInternal')
+        return get_blob_properties_from_generated_code(generated)
+    blob = BlobProperties()
+    if 'name' in select:
```
annatisch, line 61: I implemented this `select` logic in case we wanted to return more from the payload than just the name. However that seems unlikely - so we could refactor this out and simplify the logic a big here.

sdk/storage/azure-storage-blob/azure/storage/blob/_list_blobs_helper.py
```diff
@@ -7,14 +7,70 @@
...
+
+def load_many_nodes(element, name, wrapper=None):
+    if wrapper:
+        element = load_single_node(element, wrapper)
+    return list(element.findall(name))
+
+
+def blob_properties_from_xml(element, select, deserializer):
+    if not select:
+        generated = deserializer.deserialize_data(element, 'BlobItemInternal')
```
annatisch, line 58: This is using the old msrest deserializer - so once we've altered the generated layer to not deserialize for us - keeping this should mean that the existing `list_blobs` doesn't change.

sdk/storage/azure-storage-blob/azure/storage/blob/_list_blobs_helper.py
```diff
@@ -7,14 +7,70 @@
...
+from ._generated.models import FilterBlobItem
 from ._models import BlobProperties, FilteredBlob
 from ._shared.models import DictMixin
+from ._shared.xml_deserialization import unpack_xml_content
 from ._shared.response_handlers import return_context_and_deserialized, process_storage_error
 
 
-class BlobPropertiesPaged(PageIterator):
+def deserialize_list_result(pipeline_response, *_):
+    payload = unpack_xml_content(pipeline_response.http_response)
```
annatisch, line 20: I _believe_ this line here is replacing the ContentDecodePolicy that I removed from the pipeline. So this would already be unpacked if we put that policy back in.

sdk/storage/azure-storage-blob/azure/storage/blob/_list_blobs_helper.py
```diff
@@ -73,30 +133,29 @@ def _get_next_cb(self, continuation_token):
                 prefix=self.prefix,
                 marker=continuation_token or None,
                 maxresults=self.results_per_page,
-                cls=return_context_and_deserialized,
+                cls=deserialize_list_result,
```
annatisch, line 136: This is the `cls` parameter I mentioned that we would use to hook into the deserialization.

sdk/storage/azure-storage-blob/azure/storage/blob/_shared/xml_deserialization.py
```diff
@@ -0,0 +1,587 @@
...
+# THE SOFTWARE IS PROVIDED *AS IS*, WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+# FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+# IN THE SOFTWARE.
+#
+# --------------------------------------------------------------------------
+
```
annatisch, line 26: We can probably remove this for now. If we just want to add the `list_blob_names` API, then I think we can keep the existing `ContentDecodePolicy`, and the `list_blob_helper.py` file already has the extraction of data from the XML payload, so I think we could just store this away for a distant future if and when we want to decouple from msrest.

annatisch (commented):
Leaving some explanatory comments for @jalauzon-msft and @vincenttran-msft

Issue #19802, opened 2021-07-14
[translation] align on filtering naming/design


kristapratico:
.NET is here: https://apiview.dev/Assemblies/Review/46373cdd5bff4c86aeef03d3b8f79d50?diffRevisionId=bdba2e7c72b0480baec03266a26037bb&doc=False&diffOnly=False#Azure.AI.Translation.Document.DocumentFilter

Python added the filters kwargs a few releases ago. Check that naming across languages makes sense.

PR #19826, opened 2021-07-15
[translation] filters renames/alignment

kristapratico:
Resolves https://github.com/Azure/azure-sdk-for-python/issues/19802

- Renamed `translated_before` and `translated_after` kwargs to `created_before` and `created_after`, respectively,
for `list_all_document_statuses`. Reasoning is that documents which failed to translate will show up in the query and saying "translated_xx" is not totally correct in this case. Also aligns with .NET naming.
- Renamed `order_by` sorting query option `createdDateTimeUtc` to `created_on` for `list_all_translation_statuses` and
`list_all_document_statuses`. So people can map this sorting query option to the property on the result model.
- Updated tests and enabled the mixed filters tests to test when statuses == "Succeeded" since service fixed this bug recently.

### DIFF
--- a/sdk/translation/azure-ai-translation-document/azure/ai/translation/document/_helpers.py
+++ b/sdk/translation/azure-ai-translation-document/azure/ai/translation/document/_helpers.py
@@ -5,7 +5,7 @@
 # ------------------------------------
 
 import datetime
-from typing import Union
+from typing import Union, Optional
 import six
 from azure.core.credentials import AzureKeyCredential
 from azure.core.pipeline.policies import AzureKeyCredentialPolicy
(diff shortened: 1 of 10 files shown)


mohamedshabanofficial (approved):

sdk/translation/azure-ai-translation-document/azure/ai/translation/document/_helpers.py
```diff
@@ -123,3 +123,10 @@ def convert_datetime(date_time):
             except ValueError:
                 return datetime.datetime.strptime(date_time, "%Y-%m-%d %H:%M:%S")
     raise TypeError("Bad datetime type")
+
+
+def convert_order_by(order_by):
+    # type: (Optional[List[str]]) -> Optional[List[str]]
+    if order_by:
+        order_by = [order.replace("created_on", "createdDateTimeUtc") for order in order_by]
```
maririos, line 131: would this work if the user passes any comb of upper case and lower case for `created_on` like i.e. `Created_on`?
not sure if this is a scenario, but asking in case it is important in Python

maririos (approved):

sdk/translation/azure-ai-translation-document/azure/ai/translation/document/_helpers.py
kristapratico: Hmm this wouldn't be proper snake case so I wouldn't consider it a scenario. I think I'll let the service validate here, they do send back an error if the query param is not supported.

Issue #19798, opened 2021-07-14
[translation] support translation kwargs in single overloaded version of begin_translation


kristapratico:
So you can do something like this instead of needing to instantiate a `DocumentTranslationInput`:

`poller = client.begin_translation(source, target, "fr", storage_type="file")`

Need to add support to pass `storage_type`, `prefix`, `suffix`, `source_storage_source`, `target_storage_source`, `source_language_code`, `category_id`, and `glossaries` as keyword arguments into the single translation overloaded version of `begin_translation`.

The kwargs above account for both source and target options. They match the name of the properties found on `DocumentTranslationInput` except in the case of `source_storage_source` and `target_storage_source` where they are prefixed with "source" and "target" to differentiate the types of storage sources. These currently only support "AzureBlob" so might not be worth even exposing?

PR #19851, opened 2021-07-16
[translation] support translation kwargs for single translation overload

kristapratico:
Resolves https://github.com/Azure/azure-sdk-for-python/issues/19798

https://apiview.dev/Assemblies/Review/019355482c6245a5a404eef64740da1e?diffRevisionId=39de9ee01595481abdf9749b09966197&doc=False&diffOnly=True

### DIFF
--- a/sdk/translation/azure-ai-translation-document/azure/ai/translation/document/_client.py
+++ b/sdk/translation/azure-ai-translation-document/azure/ai/translation/document/_client.py
@@ -144,10 +144,6 @@ class DocumentTranslationClient(object):  # pylint: disable=r0205
         :keyword str category_id: Category / custom model ID for using custom translation.
         :keyword glossaries: Glossaries to apply to translation.
         :paramtype glossaries: list[~azure.ai.translation.document.TranslationGlossary]
-        :keyword str source_storage_source: Source storage source. Default value: "AzureBlob".
-            Currently only "AzureBlob" is supported.
-        :keyword str target_storage_source: Target storage source. Default value: "AzureBlob".
-            Currently only "AzureBlob" is supported.
         :return: An instance of a DocumentTranslationLROPoller. Call `result()` on the poller
             object to return a pageable of DocumentStatus. A DocumentStatus will be
             returned for each translation on a document.
--- a/sdk/translation/azure-ai-translation-document/azure/ai/translation/document/_client.py
+++ b/sdk/translation/azure-ai-translation-document/azure/ai/translation/document/_client.py
@@ -113,8 +113,8 @@ class DocumentTranslationClient(object):  # pylint: disable=r0205
     ):  # pylint: disable=client-method-missing-type-annotations
         """Begin translating the document(s) in your source container to your target container
         in the given language. To perform a single translation from source to target, pass the `source_url`,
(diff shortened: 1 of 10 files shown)


sdk/translation/azure-ai-translation-document/azure/ai/translation/document/_client.py
```diff
@@ -131,6 +131,23 @@ def begin_translation(
...
+            sub folders for translation.
+        :keyword str suffix: A case-sensitive suffix string to filter documents in the source path for
+            translation. This is most often use for file extensions.
+        :keyword storage_type: Storage type of the input documents source string. Possible values
+            include: "Folder", "File".
+        :paramtype storage_type: str or ~azure.ai.translation.document.StorageInputType
+        :keyword str category_id: Category / custom model ID for using custom translation.
+        :keyword glossaries: Glossaries to apply to translation.
+        :paramtype glossaries: list[~azure.ai.translation.document.TranslationGlossary]
+        :keyword str source_storage_source: Source storage source. Default value: "AzureBlob".
```
iscai-msft, line 147: Slightly weird how you have "Default value is "AzureBlob". Only "AzureBlob" supported"

I think for now you can just have the "Only AzureBlob supported" for now, and once more values get added, switch it to "Default value is AzureBlob"

sdk/translation/azure-ai-translation-document/azure/ai/translation/document/_helpers.py
```diff
@@ -44,14 +45,39 @@ def get_translation_input(args, kwargs, continuation_token):
             target_language_code = kwargs.pop("target_language_code", None)
             if not target_language_code:
                 target_language_code = args[2]
+
+            # Additional kwargs
+            source_language_code = kwargs.pop("source_language_code", None)
+            prefix = kwargs.pop("prefix", None)
+            suffix = kwargs.pop("suffix", None)
+            storage_type = kwargs.pop("storage_type", None)
+            source_storage_source = kwargs.pop("source_storage_source", None)
```
iscai-msft, line 54: just to make sure: are you getting the "AzureBlob" default from the generated code / service, and that's why you can default to None here?

iscai-msft (approved):

sdk/translation/azure-ai-translation-document/azure/ai/translation/document/_client.py
maririos: curious, why is Python exposing this property if the user can't set other values right now?
in .NET we decided to hide it, and whenever there are other options we will expose it.

maririos (approved):

sdk/translation/azure-ai-translation-document/azure/ai/translation/document/_client.py
kristapratico: I was on the fence about it and Johan didn't have a preference for exposing it. Seeing as .NET is hiding it, I will hide as well. Thanks!

sdk/translation/azure-ai-translation-document/azure/ai/translation/document/_helpers.py
kristapratico: yes service defaults to AzureBlob

PR #19872, opened 2021-07-20
Remove responses and return a list

rakshith91:


sdk/monitor/azure-monitor-query/tests/test_logs_client.py
```diff
@@ -58,7 +58,7 @@ def test_logs_server_timeout():
             "range x from 1 to 10000000000 step 1 | count",
             server_timeout=1,
         )
-        assert e.message.contains('Gateway timeout')
+    assert e.value.message.contains('Gateway timeout')
```
rakshith91, line 61:
```suggestion
    assert 'Gateway timeout' in e.value.message
```

sdk/monitor/azure-monitor-query/tests/test_logs_client.py
```diff
@@ -58,7 +58,7 @@ def test_logs_server_timeout():
             "range x from 1 to 10000000000 step 1 | count",
```
rakshith91, line 58:
```suggestion
            "range x from 1 to 1000000000000000 step 1 | count",
```

sdk/monitor/azure-monitor-query/azure/monitor/query/_log_query_client.py
```diff
@@ -131,7 +131,7 @@ def query(self, workspace_id, query, duration=None, **kwargs):
             process_error(e)
 
     def batch_query(self, queries, **kwargs):
```
xiangyan99, line 133: Shouldn't it return a pageable object?

sdk/monitor/azure-monitor-query/azure/monitor/query/_log_query_client.py
```diff
@@ -162,9 +162,11 @@ def batch_query(self, queries, **kwargs):
             request_order = [req['id'] for req in queries]
         batch = BatchRequest(requests=queries)
         generated = self._query_op.batch(batch, **kwargs)
-        return LogsBatchResults._from_generated( # pylint: disable=protected-access
-            generated, request_order
-            )
+        return order_results(
+            request_order,
```
xiangyan99, line 166: what if the id in request_order does not exist in results?

sdk/monitor/azure-monitor-query/azure/monitor/query/_log_query_client.py
rakshith91: no - service doesn't support paging

sdk/monitor/azure-monitor-query/azure/monitor/query/_log_query_client.py
rakshith91: it will always exist - it's a 1:1 mapping

sdk/monitor/azure-monitor-query/azure/monitor/query/_log_query_client.py
xiangyan99: I think our guideline is even service does not support page, we should wrap the return into pageable results?
@annatisch

rakshith91: we talked to the architects and they were fine in this case with not supporting both paging and LRO
as service is not planning to ever support it on these endpoints

rakshith91: i have created this issue for september milestone to follow up again https://github.com/Azure/azure-sdk-for-python/issues/19942

xiangyan99 (approved):

PR #19888, opened 2021-07-20
[Batch] Management plane SDK v16.0.0

cRui861:
Updates the SDK to use the 2021-06-01 API version.


dpwatrous:
The test failure here seems to be related to 'Azure SDK Bot' automatically removing 'azure_mgmt_core' from setup.py

dpwatrous:
Updating sdk_packaging.toml fix it. @msyyc @RAY-316 this PR should be good to merge now if it looks good to you.

dpwatrous:
Squashed some commits where we were trying to get the CI passing for easier review

dpwatrous:
Almost forgot: I noticed while I was updating the data plane to use the new version of azure-mgmt-batch that the client name has changed to `BatchManagement` when previously it was `BatchManagementClient`.

Opinions on if we can switch it back to `BatchManagementClient` as a breaking change? This will actually be *less* of a breaking change for customers that have not yet moved to our Track 2 SDKs.

dpwatrous:
Note the doc string changes in this PR are caused by a regression in autorest.python. See https://github.com/Azure/autorest.python/issues/978

Because this is out of our control, but we have important customers waiting on this management plane release, we plan to fix this in a point release when the issue is fixed in autorest.

msyyc:
@cRui861 For python mgmt. SDK, if you want to publish it, please make a release issue in https://portal.azure-devex-tools.com/tools/request-api-release. We do not permit people outside team to merge PR. If any question, please send the e-mail to notice us.

dpwatrous:
@msyyc For what it's worth we weren't looking to merge this ourselves, just front-load the work to help expedite the release. I've submitted a release request through OpenAPI Hub, and I'll follow up separately.

PR #19902, opened 2021-07-21
Token exchange support for ManagedIdentityCredential 

### LINKED ISSUE #19304: Support exchanging k8s token to AAD token
Tracking Issue for Better Support for Kubernetes service accounts.  See @chlowell

chlowell:
With this change, ManagedIdentityCredential authenticates via a token exchange when these variables are set:
- AZURE_CLIENT_ID
- AZURE_TENANT_ID
- TOKEN_FILE_PATH: path to a file containing the JWT to be exchanged
    - the credential caches this file's content for 5 minutes i.e. it reads the file only after 5 minutes have passed since the last time it read the file
- optional: AZURE_AUTHORITY_HOST

I implemented this atop a new credential class, ClientAssertionCredential, which authenticates with a JWT assertion. It's quite similar to CertificateCredential but instead of constructing a JWT for the "client_assertion" field, this new credential gets one from a callback. I kept ClientAssertionCredential internal because there's no need to make it public today; we may find one in the future.

Closes #19304


sdk/identity/azure-identity/azure/identity/aio/_credentials/token_exchange.py
```diff
@@ -0,0 +1,43 @@
...
+                **kwargs
+            )
+
+    async def __aenter__(self):
+        if self._credential:
+            await self._credential.__aenter__()
+        return self
+
+    async def close(self) -> None:
+        await self._credential.close()
```
mccoyp, line 38: Should we do a check for `self._credential` here (and if we shouldn't, should we do that check in the sync credential)? Also, is there a reason that async credentials follow a `close() -> self._x.close()` pattern instead of `close() -> self.__aexit__() -> self._x.__aexit__()` pattern like sync credentials do?

sdk/identity/azure-identity/azure/identity/_credentials/token_exchange.py
```diff
@@ -0,0 +1,68 @@
...
+            self._last_read_time = now
+        return self._jwt
+
+
+class TokenExchangeCredential(TokenFileMixin):
+    def __init__(self, **kwargs):
+        # type: (**Any) -> None
+        super(TokenExchangeCredential, self).__init__()
+        self._available = all(os.environ.get(var) for var in EnvironmentVariables.TOKEN_EXCHANGE_VARS)
+        if self._available:
```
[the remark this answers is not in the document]
chlowell: Summarizing offline conversation: it's moot because ManagedIdentityCredential is the public API of this credential and already logs the interesting bit (what platform it believes it's running on) and will only instantiate this credential when its required environment variables are set. TokenExchangeCredential can therefore assume its prerequisites are met, and stop trying to raise a useful error message when that isn't the case. Commit to that effect forthcoming.

sdk/identity/azure-identity/azure/identity/aio/_credentials/token_exchange.py
chlowell: We should check that, thanks for noticing! There's no good reason for the sync vs. async difference. I prefer close -> __exit__ today but it doesn't really matter.

sdk/identity/azure-identity/azure/identity/aio/_credentials/client_assertion.py
```diff
@@ -0,0 +1,50 @@
...
+        self._client = AadClient(tenant_id, client_id, **kwargs)
+        super().__init__(**kwargs)
+
+    async def __aenter__(self):
+        await self._client.__aenter__()
+        return self
+
+    async def close(self) -> None:
+        """Close the credential's transport session."""
+        await self._client.close()
```
mccoyp, line 42: Are you keeping the async context management API since it's more relevant for async credentials?

sdk/identity/azure-identity/azure/identity/_credentials/token_exchange.py
```diff
@@ -0,0 +1,38 @@
...
+        now = int(time.time())
+        if now - self._last_read_time > 300:
+            with open(self._token_file_path) as f:
+                self._jwt = f.read()
+            self._last_read_time = now
+        return self._jwt
+
+
+class TokenExchangeCredential(ClientAssertionCredential, TokenFileMixin):
+    def __init__(self, token_file_path, **kwargs):
```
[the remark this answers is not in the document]
chlowell: Sure, that will make it easier for future generations to understand how this works.

sdk/identity/azure-identity/azure/identity/aio/_credentials/client_assertion.py
chlowell: That and `self._client` here already implements the API. I removed it from the sync credential because I don't want to block this PR on #19746 or paste part of that PR into this one.

mccoyp (approved):

PR #19914, opened 2021-07-22
Refactor SharedTokenCacheCredential

chlowell:
SharedTokenCacheCredential takes one of two distinct code paths depending on whether it's initialized with an AuthenticationRecord. It's basically two credential types mashed together (my bad). That makes it difficult to understand and maintain, so here I am unmashing that which has been mashed. I think this will be a boon for future generations, and today it makes fixing #19746 easier 😊


mccoyp (approved):
Super nice change, thank you for noticing the opportunity to split these up!

Issue #19777, opened 2021-07-13
Memory leak in ServiceBusClient get_subscription_receiver


miroag:
- **Package Name**: azure-servicebus
- **Package Version**: azure-servicebus==7.3.1 && uamqp==1.4.1
- **Operating System**: Windows 10/WSL Ubuntu/Docker python 3.9
- **Python Version**: 3.9

**Describe the bug**
Memory leak in get_subscription_receiver

**To Reproduce**
Snippet below will result in memory leak of few Mb in 10 minutes
```
import os
import time


from azure.servicebus import ServiceBusClient, NEXT_AVAILABLE_SESSION

if __name__ == '__main__':
    sb_client = ServiceBusClient.from_connection_string(os.environ['SB_CONNECTION_STRING'])
    start = time.time()
    i=0
    while time.time() - start < 10 * 60:
        with sb_client.get_subscription_receiver(topic_name='sync', subscription_name='data',
                                                 idle_timeout=1, session_id=NEXT_AVAILABLE_SESSION) as receiver:
            i+=1
            print(f'Receiver #{i}')
```

**Expected behavior**
No memory leaks obviously. The code like in the snippet above results in 1Gb memory leak in 12 hrs in our production environment. Unless, I'm missing something I cannot reuse the receiver in the scenario when I need to work with sessions.
Effectively, this renders the use of ServiceBus very problematic.

**Screenshots**
![image](https://user-images.githubusercontent.com/10284158/125480818-b27bd267-9318-4074-8d64-68fa1ca37fef.png)

**Additional context**
This might be similar / identical to https://github.com/Azure/azure-sdk-for-python/issues/15747
But the above one is closed apparently without proper resolution.

Kind regards,
Alex

xiangyan99:
Thanks for the feedback, we’ll investigate asap.

yunhaoling:
hey @miroag, I'll take a look into the issue this week, thanks for bring it to my attention!

yunhaoling:
hey @miroag , confirmed that memory leak is happening on Windows, also did a thorough debugging and long running test to find the leaking places. I had a [PR](https://github.com/Azure/azure-uamqp-python/pull/258) out for the fix in the uamqp library, but am now waiting for the C team to verify my fix is the right approach.

However, I have run long running test on Linux as well as valgrind (c program memory leak debug tool) but didn't observe any memory leak. Could you help confirm whether there's memory leak on Linux or not?

miroag:
[...]

PR #19915, opened 2021-07-22
[ServiceBus] improve memory usage of service bus client

yunhaoling:
Potential improvement for issue: https://github.com/Azure/azure-sdk-for-python/issues/19777

automatically deregistering spawned sender and receiver from the client to save memory usage


sdk/servicebus/azure-servicebus/tests/async_tests/test_sb_client_async.py
```diff
@@ -174,6 +178,38 @@ async def test_async_sb_client_close_spawned_handlers(self, servicebus_namespace
...
+            pass
+        assert len(client._handlers) == 0
+
+        queue_sender = client.get_queue_sender(servicebus_queue.name)
+        queue_receiver = client.get_queue_receiver(servicebus_queue.name)
+        assert len(client._handlers) == 2
+        await queue_sender.close()
+        await queue_sender.close()
+        await queue_receiver.close()
+        await queue_receiver.close()
```
swathipil, line 194: is there a reason these are doubled?

sdk/servicebus/azure-servicebus/tests/async_tests/test_sb_client_async.py
```diff
@@ -174,6 +178,38 @@ async def test_async_sb_client_close_spawned_handlers(self, servicebus_namespace
...
+            pass
+        assert len(client._handlers) == 0
+
+        topic_sender = client.get_topic_sender(servicebus_topic.name)
+        subscription_receiver = client.get_subscription_receiver(servicebus_topic.name, servicebus_subscription.name)
+        assert len(client._handlers) == 2
+        await topic_sender.close()
+        await topic_sender.close()
+        await subscription_receiver.close()
+        await subscription_receiver.close()
```
swathipil, line 210: same question, reason for closing twice?

sdk/servicebus/azure-servicebus/tests/async_tests/test_sb_client_async.py
```diff
@@ -174,6 +178,38 @@ async def test_async_sb_client_close_spawned_handlers(self, servicebus_namespace
         assert not receiver._handler and not receiver._running
         assert len(client._handlers) == 0
 
+        queue_sender = client.get_queue_sender(servicebus_queue.name)
+        queue_receiver = client.get_queue_receiver(servicebus_queue.name)
+        assert len(client._handlers) == 2
+        async with queue_sender, queue_receiver:
+            pass
+        assert len(client._handlers) == 0
```
swathipil, line 186: so is it not possible to do, for example, the following?
```suggestion
        assert len(client._handlers) == 2
        async with queue_sender, queue_receiver:
            pass
         # what would this line be?
         # assert len(client._handlers) == ?
        async with queue_sender, queue_receiver:
            pass
        assert len(client._handlers) == 0
```

sdk/servicebus/azure-servicebus/tests/test_sb_client.py
```diff
@@ -238,6 +242,45 @@ def test_sb_client_close_spawned_handlers(self, servicebus_namespace_connection_
...
+            pass
+        assert len(client._handlers) == 0
+
+        queue_sender = client.get_queue_sender(servicebus_queue.name)
+        queue_receiver = client.get_queue_receiver(servicebus_queue.name)
+        assert len(client._handlers) == 2
+        queue_sender.close()
+        queue_sender.close()
+        queue_receiver.close()
+        queue_receiver.close()
```
swathipil, line 265: closed twice?

sdk/servicebus/azure-servicebus/tests/test_sb_client.py
```diff
@@ -238,6 +242,45 @@ def test_sb_client_close_spawned_handlers(self, servicebus_namespace_connection_
...
+            pass
+        assert len(client._handlers) == 0
+
+        topic_sender = client.get_topic_sender(servicebus_topic.name)
+        subscription_receiver = client.get_subscription_receiver(servicebus_topic.name, servicebus_subscription.name)
+        assert len(client._handlers) == 2
+        topic_sender.close()
+        topic_sender.close()
+        subscription_receiver.close()
+        subscription_receiver.close()
```
swathipil, line 281: closed twice?

sdk/servicebus/azure-servicebus/CHANGELOG.md
```diff
@@ -10,6 +10,8 @@
 
 ### Other Changes
 
+- Improved memory usage of `ServiceBusClient` to automatically de-registered spawned `ServiceBusSender` and `ServiceBusReceiver` which are closed.
```
swathipil, line 13: small nit, flow:
```suggestion
- Improved memory usage of `ServiceBusClient` by automatically de-registering spawned `ServiceBusSender` and `ServiceBusReceiver` that are closed.
```

sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_client.py
```diff
@@ -115,7 +115,7 @@ def close(self):
 
         :return: None
         """
-        for handler in self._handlers:
+        for handler in set(self._handlers):
```
swathipil, line 118: are there instances in which Sender/Receiver can be added multiple times to the same client?

sdk/servicebus/azure-servicebus/azure/servicebus/aio/_base_handler_async.py
```diff
@@ -385,3 +386,9 @@ async def close(self) -> None:
         """
         await self._close_handler()
         self._shutdown.set()
+        try:
+            self._service_bus_client._handlers.remove(self)  # pylint: disable=protected-access
```
swathipil, line 390: if using `set()` when iterating through handlers, does this mean that multiple/duplicates of the same handler can be added to the list? In that case, would you want to iterate through the list to remove all instances of `self` since `remove` only removes the first element/`self`?

(Same comment for sync)

swathipil (commented):
left a few small comments!

sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_client.py
yunhaoling: according to our current design, nope, the handlers are maintained internally only and users could not add existing handler to the client.

sdk/servicebus/azure-servicebus/tests/async_tests/test_sb_client_async.py
yunhaoling: it's basically for testing removing twice is handled properly, ValueError is caught -- `self._service_bus_client._handlers.remove(self)`

sdk/servicebus/azure-servicebus/azure/servicebus/aio/_base_handler_async.py
yunhaoling: duplicates of the same handler CAN NOT be add to the list. sorry for the confusion, the set is used by me for dynamic deleting, I'll remove that 😉

sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_client.py
yunhaoling: this is actually doing a copy of the handler array, otherwise the handler array would be changing while performing the `handler.close()` operation as the handler.close method would remove itself from the handlers of service bus client.

sdk/servicebus/azure-servicebus/azure/servicebus/aio/_base_handler_async.py
yunhaoling: same handlers could not be added, see my comment above.

sdk/servicebus/azure-servicebus/azure/servicebus/_base_handler.py
```diff
@@ -536,3 +537,9 @@ def close(self):
         """
         self._close_handler()
         self._shutdown.set()
+        try:
+            self._service_bus_client._handlers.remove(self)  # pylint: disable=protected-access
```
johanste, line 541: We really don't want any protected access here. It should be safe to assume that no-one outside of the current class touches an underscored name. We've been burn before when refactoring code. In this case, it is especially true because we are also catching `AttributeError`s that may happen if someone were to rename `_handlers` or similar.

swathipil (approved):
LGTM!

sdk/servicebus/azure-servicebus/azure/servicebus/_base_handler.py
yunhaoling: hey @johanste , any comments on the callback approach?

yunhaoling: per OOB discussion, use weakSet which could automatically discard elements for us

PR #19922, opened 2021-07-23
CloudEvents/EG Events must recognize magic events

rakshith91:
Fixes https://github.com/Azure/azure-sdk-for-python/issues/19247


sdk/eventgrid/azure-eventgrid/azure/eventgrid/_models.py
```diff
@@ -96,3 +98,8 @@ def __repr__(self):
         return "EventGridEvent(subject={}, event_type={}, id={}, event_time={})".format(
             self.subject, self.event_type, self.id, self.event_time
         )[:1024]
+    
+    @classmethod
+    def from_dict(cls, event):
+        event = _EventMixin._get_bytes(event)
```
rakshith91, line 104:
```suggestion
        event = EventGridEvent._get_bytes(event)
```

sdk/eventgrid/azure-eventgrid/azure/eventgrid/_models.py
```diff
@@ -3,16 +3,18 @@
 # Licensed under the MIT License. See License.txt in the project root for license information.
 # --------------------------------------------------------------------------------------------
 # pylint:disable=protected-access
+from _typeshed import Self
```
rakshith91, line 6:
```suggestion
```

sdk/eventgrid/azure-eventgrid/azure/eventgrid/_models.py
```diff
@@ -96,3 +98,8 @@ def __repr__(self):
         return "EventGridEvent(subject={}, event_type={}, id={}, event_time={})".format(
             self.subject, self.event_type, self.id, self.event_time
         )[:1024]
+    
+    @classmethod
+    def from_dict(cls, event):
+        event = EventGridEvent._get_bytes(event)
+        super(EventGridEvent).from_dict(event)
```
rakshith91, line 104:
```suggestion
        super(EventGridEvent, cls).from_dict(event)
```

sdk/eventgrid/azure-eventgrid/azure/eventgrid/_models.py
```diff
@@ -96,3 +97,36 @@ def __repr__(self):
...
             self.subject, self.event_type, self.id, self.event_time
         )[:1024]
+
+    @classmethod
+    def from_dict(cls, data, key_extractors=None, content_type=None):
+        event = EventGridEvent._get_bytes(data)
+        super(EventGridEvent, cls).from_dict(event, key_extractors, content_type)
+
+    @staticmethod
+    def _get_bytes(obj):
```
annatisch, line 107: Same comment re name and making a separate utility function.

sdk/core/azure-core/CHANGELOG.md
```diff
@@ -1,9 +1,11 @@
 # Release History
 
-## 1.16.1 (Unreleased)
+## 1.17.0 (Unreleased)
 
 ### Features Added
 
+- `CloudEvent`'s `from_dict` method now accepts objects from servicebus, eventhubs and storage directly.
```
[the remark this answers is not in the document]
annatisch: I guess if you stretch the definition then the payload is formatted as a dict....

sdk/core/azure-core/azure/core/messaging.py
```diff
@@ -181,3 +183,31 @@ def from_dict(cls, event):
...
+        except ValueError:
+            raise ValueError(
+                "Failed to retrieve content from the object. Make sure the "
+                 + "content follows the CloudEvent schema."
+                )
+        except AttributeError:
+            # eventhubs
+            try:
+                return json.loads(next(obj.body))[0]
+            except KeyError:
```
annatisch, line 204: Is it possible an IndexError could be raised here?

sdk/core/azure-core/azure/core/messaging.py
```diff
@@ -181,3 +183,31 @@ def from_dict(cls, event):
...
+        """Event mixin to have methods that are common to different Event types
+        like CloudEvent, EventGridEvent etc.
+        """
+        try:
+            # storage queue
+            return json.loads(obj.content)
+        except ValueError:
+            raise ValueError(
+                "Failed to retrieve content from the object. Make sure the "
+                 + "content follows the CloudEvent schema."
```
annatisch, line 198: Technically this error will be raised if the content isn't valid JSON, regardless of whether that JSON follows the schema?

sdk/core/azure-core/azure/core/messaging.py
annatisch: And in what case would it raise a KeyError?
Do we also need different handling for a StopIteration error? In that scenario, it meant that we had been passed the right object, but the payload was not valid - so I would expect that to then raise a ValueError, no?

sdk/core/azure-core/CHANGELOG.md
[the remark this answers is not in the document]
rakshith91: it's inspired from msrest's terminology i believe

sdk/core/azure-core/azure/core/messaging.py
rakshith91: That is correct -the wording is a bit intentional

sdk/core/azure-core/azure/core/messaging.py
```diff
@@ -181,3 +183,31 @@ def from_dict(cls, event):
                     " The `source` and `type` params are required."
                     )
         return event_obj
+
+    @staticmethod
+    def _get_bytes(obj):
```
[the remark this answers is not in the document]
rakshith91: renaming to `get_json_content`

sdk/core/azure-core/azure/core/messaging.py
rakshith91: ```Is it possible an IndexError could be raised here?```

We always pass our events as a list in the send method. Even if the user passes a single event, we internally convert it into a list. So, there shouldn't be an index error

```And in what case would it raise a KeyError?```

A keyerror is raised when we try to subscript a SB message with 0. essentially, service bus message is a dict. which is fetched by `json.loads(next(obj.body))`. So doing a `json.loads(next(obj.body))[0]` will give you a key error.

Comment (rakshith91) [reply sdk/core/azure-core/azure/core/messaging.py]: ```Do we also need different handling for a StopIteration error```
No - uamqp does't throw stop iteration https://github.com/Azure/azure-uamqp-python/blob/master/uamqp/message.py#L1159-L1162

File: sdk/eventgrid/azure-eventgrid/azure/eventgrid/_models.py (hunk 4 above)
Comment (annatisch) [reply sdk/eventgrid/azure-eventgrid/azure/eventgrid/_models.py:108]: Shouldn't this method now be removed?

Comment (rakshith91) [reply sdk/eventgrid/azure-eventgrid/azure/eventgrid/_models.py:108]: oops - removing

File: sdk/eventgrid/azure-eventgrid/azure/eventgrid/_models.py [hunk 9]
```diff
@@ -6,7 +6,9 @@
 from typing import Any
 import datetime as dt
 import uuid
+import json
```
rakshith91, line 9:
```suggestion
```

sdk/core/azure-core/azure/core/messaging.py
```diff
@@ -181,3 +181,16 @@ def from_dict(cls, event):
                     " The `source` and `type` params are required."
                     )
         return event_obj
+
+    @classmethod
+    def from_json(cls, json):
+        # type: (Any) -> CloudEvent
+        """
+        Returns the deserialized CloudEvent object when a json is provided.
```
annatisch, line 189: Nit: maybe "a json payload"? Or just "json content"?

sdk/core/azure-core/azure/core/messaging.py
```diff
@@ -181,3 +181,16 @@ def from_dict(cls, event):
...
+        :param json: The json string that should be converted into a CloudEvent. This can also be
+         a storage QueueMessage, eventhub's EventData or ServiceBusMessage
+        :type json: object
+        :rtype: CloudEvent
```
annatisch, line 193: Might be worth adding something like:
`:raises ValueError: If the provided JSON is invalid.`

sdk/core/azure-core/azure/core/utils/_utils.py
```diff
@@ -106,3 +107,30 @@ def _case_insensitive_dict(*args, **kwargs):
...
+
+def _get_json_content(obj):
+    """Event mixin to have methods that are common to different Event types
+    like CloudEvent, EventGridEvent etc.
+    """
+    try:
+        # storage queue
+        return json.loads(obj.content)
+    except ValueError:
+        raise ValueError(
```
annatisch, line 119: This should use `azure.core.exceptions.raise_with_traceback`

sdk/core/azure-core/azure/core/utils/_utils.py
```diff
@@ -106,3 +107,30 @@ def _case_insensitive_dict(*args, **kwargs):
...
+        except KeyError:
+            # servicebus
+            return json.loads(next(obj.body))
+        except ValueError:
+            raise ValueError(
+                "Failed to retrieve body from the object. Make sure the "
+                + "body follows the CloudEvent schema."
+                )
+        except: # pylint: disable=bare-except
+            return json.loads(obj)
```
annatisch, line 136: Shouldn't this have the same `ValueError` wrapper?

sdk/core/azure-core/tests/test_messaging_cloud_event.py
```diff
@@ -469,4 +550,179 @@ def test_wrong_schema_raises_no_type():
...
+    random_obj =  {
+        "id":"de0fd76c-4ef4-4dfb-ab3a-8f24a307e033",
+        "source":"https://egtest.dev/cloudcustomevent",
+        "data":{"team": "event grid squad"},
+        "type":"Azure.Sdk.Sample",
+        "time":"2020-08-07T02:06:08.11969Z",
+        "specversion":"1.0"
+    }
+
+    assert _get_json_content(json_str) == random_obj
```
annatisch, line 660: Would be worth trying this with the full constructor call to ensure that "time" is deserialized into the timestamp.

sdk/eventgrid/azure-eventgrid/samples/consume_samples/consume_cloud_events_from_eventhub.py
```diff
@@ -24,11 +24,9 @@
 
 CONNECTION_STR = os.environ["EVENT_HUB_CONN_STR"]
 EVENTHUB_NAME = os.environ["EVENT_HUB_NAME"]
-
-
 def on_event(partition_context, event):
-    dict_event = CloudEvent.from_dict(json.loads(event)[0])
-    print("data: {}\n".format(deserialized_event.data))
+    dict_event = CloudEvent.from_dict(event)
```
annatisch, line 28: Samples need updating :)

sdk/eventgrid/azure-eventgrid/azure/eventgrid/_helpers.py
```diff
@@ -165,3 +165,30 @@ def _build_request(endpoint, content_type, events):
...
+    """Event mixin to have methods that are common to different Event types
+    like CloudEvent, EventGridEvent etc.
+    """
+    try:
+        # storage queue
+        return json.loads(obj.content)
+    except ValueError:
+        raise ValueError(
+            "Failed to retrieve content from the object. Make sure the "
+                + "content follows the EventGridEvent schema."
```
annatisch, line 179: We can probably rephrase this now that it lives in a json-specific context - maybe something like:
"Failed to load JSON content from the object."

sdk/core/azure-core/azure/core/messaging.py
annatisch: Interesting - so if the Event body is an empty list - calling `next` on it doesn't raise? What does it do?

sdk/core/azure-core/tests/test_messaging_cloud_event.py
rakshith91: i've added these

sdk/core/azure-core/azure/core/messaging.py
rakshith91:
```
m = message.Message(body= b'', body_type=uamqp.MessageBodyType.Data)
for data in m.get_data():
  print(data)

>>> b''
```

annatisch (approved):

Issue #19834, opened 2021-07-15
raise ValueError for explicit loop parameter in Python 3.10(core)


xiangyan99:
The loop parameter has been removed from most of asyncio‘s high-level API following deprecation in Python 3.8. 

Raise ValueError to catch it.

PR #19927, opened 2021-07-23
deprecate loop parameter

xiangyan99:


xiangyan99:
#19834

sdk/core/azure-core/azure/core/pipeline/transport/_aiohttp.py
```diff
@@ -73,6 +73,8 @@ class AioHttpTransport(AsyncHttpTransport):
             :caption: Asynchronous transport with aiohttp.
     """
     def __init__(self, *, session: Optional[aiohttp.ClientSession] = None, loop=None, session_owner=True, **kwargs):
+        if loop and sys.version_info >= (3, 10):
+            raise ValueError("Starting Python 3.10, asyncio doesn’t support loop as a parameter anymore")
```
annatisch, line 77: Do we want to add a deprecation warning `if loop and sys.version_info < (3, 10)`?

annatisch: Grammar nitpick: Starting **with** Python 3.10, asyncio no longer supports the loop parameter.

annatisch (approved):

sdk/core/azure-core/azure/core/pipeline/transport/_aiohttp.py
xiangyan99: No. If user does not use Python >= 3.10, we want to behave exactly same. There should be no impact at all. (new warnings are still side effects.)

PR #19930, opened 2021-07-23
cut hard dependency on requests

xiangyan99:


sdk/core/azure-core/azure/core/pipeline/transport/__init__.py
```diff
@@ -26,82 +26,131 @@
...
+    from ._requests_basic import RequestsTransport, RequestsTransportResponse
     __all__.extend([
-        'AsyncHttpTransport',
-        'AsyncHttpResponse',
-        'AsyncioRequestsTransport',
-        'AsyncioRequestsTransportResponse'
+        'RequestsTransport',
+        'RequestsTransportResponse',
     ])
+    try:
```
annatisch, line 43: Could we add comments to all the the try/except blocks here? The logic is a little hard to follow :)

sdk/core/azure-core/azure/core/utils/_utils.py
```diff
@@ -99,6 +99,8 @@ def _case_insensitive_dict(*args, **kwargs):
         # multidict is installed by aiohttp
         from multidict import CIMultiDict
 
+        if len(kwargs) == 0 and len(args) == 1 and (not args[0]):
```
annatisch, line 102: What's `args[0]` in this case? Maybe add a small comment as to what this if-statement is addressing?

xiangyan99: This is to handle case_insensitive_dict(None). requests.structures.CaseInsensitiveDict returns empty dict while multidict.CIMultiDict raise error.

Here we want to return dict().

annatisch (approved):

PR #19932, opened 2021-07-24
[AutoRelease] t2-batch-2021-07-24-41488

azclibot:
https://github.com/Azure/sdk-release-request/issues/1800


msyyc (approved):

dpwatrous (changes requested):
We just need to regenerate the models based on the latest swagger, then this should be good to go.

Issue #19308, opened 2021-06-17
Feature: Add On-Behalf-Of (OBO) Auth Flow for the Microsoft Graph Team


joshfree:
The [OAuth 2.0 On-Behalf-Of flow (OBO)](https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-on-behalf-of-flow) serves the use case where an application invokes a service/web API, which in turn needs to call another service/web API. The idea is to propagate the delegated user identity and permissions through the request chain. For the middle-tier service to make authenticated requests to the downstream service, it needs to secure an access token from the Microsoft identity platform, on behalf of the user.

**Related Links:**
Prototype: https://github.com/jongio/azidext/pull/41
Documenation: https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-on-behalf-of-flow
.NET Design: https://gist.github.com/christothes/15b9903f7da0954aee166166a8fa1919
MS Graph Team's tracking issue: https://github.com/microsoftgraph/msgraph-sdk-java-core/issues/132

PR #19940, opened 2021-07-26
TokenCredential.supports_caching

chlowell:
Part of #19308. Today, BearerTokenCredentialPolicy caches the last access token it acquired and calls `get_token()` only when this cached token will soon expire. This isn't safe with on-behalf-of tokens because they contain a user assertion identifying the user on whose behalf the application accesses resources. When an application changes its intended user, a client must not continue using a token for the prior user. Our design for on-behalf-of authentication makes the credential responsible for tracking the application's user assertion changes, which is to say it requires BearerTokenCredentialPolicy to call `get_token()` every time it authorizes a request. So, this PR adds `TokenCredential.supports_caching()`. BearerTokenCredentialPolicy calls this method to learn whether a given TokenCredential maintains its own token cache. When this returns True, BearerTokenCredentialPolicy defers to the credential's cache, calling `get_token()` every time it wants to authorize a request. The policy doesn't expect all credentials to implement this method, and assumes it's safe to cache tokens from credentials which do not.

### DIFF
--- a/sdk/core/azure-core/azure/core/credentials.py
+++ b/sdk/core/azure-core/azure/core/credentials.py
@@ -24,6 +24,13 @@ if TYPE_CHECKING:
             # type: (*str, **Any) -> AccessToken
             pass
 
+        def supports_caching(self):
+            # type: () -> bool
+            """Whether this TokenCredential maintains its own token cache.
+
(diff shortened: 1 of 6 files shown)


chlowell:
Closing in favor of #20451

PR #19965, opened 2021-07-28
Allow Tox Environments to be Configured As Skippable via platform-matrix

scbedd:
This PR will allow us to resolve the nightly failure that identity is currently hitting every day.

It also:

- updates identity matrix to exclude `mindependency/latestdependency` checks for the MSAL run
- updates an inline pshell script to a python script

### DIFF
--- a/eng/pipelines/templates/jobs/ci.tests.yml
+++ b/eng/pipelines/templates/jobs/ci.tests.yml
@@ -23,6 +23,9 @@ parameters:
   - name: ToxEnvParallel
     type: string
     default: '--tenvparallel'
+  - name: UnsupportedToxEnvironments
+    type: string
+    default: ''
   - name: InjectedPackages
     type: string
     default: ''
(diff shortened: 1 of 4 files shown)


eng/pipelines/templates/jobs/ci.tests.yml
```diff
@@ -79,20 +82,14 @@ jobs:
...
-
-        # ensure that the variable is unset. if it isn't, use the value discovered there
-        if ('$(Run.ToxCustomEnvs)' -ne ('$' + '(Run.ToxCustomEnvs)'))
-        {
-          $toxenvvar = '$(Run.ToxCustomEnvs)'
-        }
-
-        echo "##vso[task.setvariable variable=toxenv]$toxenvvar"
-      displayName: "Set Tox Environment"
+    - task: PythonScript@0
```
mccoyp, line 85: Out of curiosity, why the "@0"?

mccoyp (approved):

eng/pipelines/templates/jobs/ci.tests.yml
semick-dev: It's the version of the job!  The way that you access native devops tasks is by the name + the version. [here is the index of native tasks](https://github.com/microsoft/azure-pipelines-tasks/tree/master/Tasks)

image

chlowell (approved):
👍

PR #19986, opened 2021-07-28
make core tests offline

xiangyan99:
convert online tests into test_server tests

### DIFF
--- a/sdk/core/azure-core/tests/async_tests/test_retry_policy_async.py
+++ b/sdk/core/azure-core/tests/async_tests/test_retry_policy_async.py
@@ -62,7 +62,7 @@ def test_retry_types():
 @pytest.mark.parametrize("retry_after_input", [('0'), ('800'), ('1000'), ('1200')])
 def test_retry_after(retry_after_input):
     retry_policy = AsyncRetryPolicy()
-    request = HttpRequest("GET", "https://bing.com")
+    request = HttpRequest("GET", "http://127.0.0.1")
     response = HttpResponse(request, None)
     response.headers["retry-after-ms"] = retry_after_input
     pipeline_response = PipelineResponse(request, response, None)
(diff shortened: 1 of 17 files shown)


sdk/core/azure-core/tests/async_tests/test_retry_policy_async.py
```diff
@@ -80,7 +80,7 @@ def test_retry_after(retry_after_input):
 @pytest.mark.parametrize("retry_after_input", [('0'), ('800'), ('1000'), ('1200')])
 def test_x_ms_retry_after(retry_after_input):
     retry_policy = AsyncRetryPolicy()
-    request = HttpRequest("GET", "https://bing.com")
+    request = HttpRequest("GET", "http://127.0.0.1")
```
iscai-msft, line 83: nit: i like using `localhost` more than `127.0.0.1`, i think it's more clear

sdk/core/azure-core/tests/test_basic_transport.py
```diff
@@ -96,27 +92,6 @@ def test_url_join():
     assert _urljoin('devstoreaccount1/', 'testdir/') == 'devstoreaccount1/testdir/'
 
 
-def test_http_client_response():
```
iscai-msft, line 99: i'm thinking that instead of migrating all of these tests into the little testserver_tests folder, why don't we get rid of the testserver_tests folder? What are your thoughts?

I think it's worth it to start now, esp if we're going to start moving more and more to testserver

xiangyan99: Hah. I was thinking about the opposite. :)

I noticed that some tests don't need test server at all. So in addition to keeping the testserver_tests folder, I was thinking moving some of your tests out of the folder to make them run faster.

Your ideas?

iscai-msft: Ah I see your point. I'm down with that, I don't feel too strongly!

sdk/core/azure-core/tests/conftest.py
```diff
@@ -39,3 +47,57 @@
...
+
+@pytest.fixture
+def port():
+    return os.environ["FLASK_PORT"]
+
+def start_testserver():
+    port = get_port()
+    os.environ["FLASK_APP"] = "coretestserver"
+    os.environ["FLASK_PORT"] = str(port)
+    # os.environ["FLASK_PORT"] = str(port)
```
iscai-msft, line 74: nit: unnecessary comment

iscai-msft (approved):

sdk/core/azure-core/tests/conftest.py
xiangyan99: Good catch. :)

PR #19992, opened 2021-07-29
[translation] update target container permissions

kristapratico:


maririos (approved):
Can u queue the live tests to verify that with the change they are all passing?

maririos:
>Can u queue the live tests to verify that with the change they are all passing?

Just saw u already did :P

kristapratico:
> Can u queue the live tests to verify that with the change they are all passing?

I did on the previous commit and they all passed. Most recent commit was just docs.

mohamedshabanofficial (approved):

PR #19998, opened 2021-07-30
[AutoRelease] t2-loganalytics-2021-07-30-40375(Do not merge)

azclibot:
123123


BigCat20196:
test function`add issue_link`

PR #20046, opened 2021-08-02
Synapse multiapi support for artifacts and vpn

xiangyan99:


sdk/synapse/azure-synapse-artifacts/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 - Updated API version to "2020-12-01"
 - Added `NotebookOperationResultOperations`, `OperationResultOperations`, `OperationStatusOperations`
+- Added API version "2021-06-01-preview" support
```
kristapratico, line 7: Since we're adding two API versions in this release, it might be nice to clarify which one the library will default to.

kristapratico (approved):

PR #20047, opened 2021-08-02
Close session of ChallengePolicyClient in context manager

xiangyan99:
#20000


seankane-msft (approved):
LGTM! Thanks @xiangyan99

PR #20048, opened 2021-08-03
[Key Vault] Implement secure key release

### LINKED ISSUE #18114: [Key Vault] Clean up SKR models/implementation
Refine models and implementation as other languages implement SKR. Specific items to consider below:

- [x] Add `ReleasePolicy` property to `KeyProperties` of type `KeyReleasePolicy`.
- [x] Add `release_policy` kwarg to `update_key_properties`

### LINKED ISSUE #19588: [Key Vault] Add support for Secure Key Release
To support Secure Key Release, implement the following changes:

## CreateKeyOptions & ImportKeyOptions:
- Add `exportable` - an optional boolean value
- Add `releasePolicy` an optional `KeyReleasePolicy` value (see below)

### KeyReleasePolicy 
an object / interface containing the following attributes:
- `contentType` - an optional string. When omitted, defaults to `application/json; charset=utf-8` by the service (not in client code)
- `data` - a byte array encoding the policy rules under which the key can be released.

## KeyClient.releaseKey

Used to actually release the key. In JS, we're currently implementing it as follows:

```ts
releaseKey(name: string, version: string, target: string, options?: ReleaseKeyOptions): Promise<ReleaseKeyResult>;
```
where target is the attestation assertion for the target of the key release.

See this PR for implementation: https://github.com/Azure/azure-sdk-for-js/pull/16150/files

Finally, we'll need a resource that can generate an attestation token and return the right keys to MHSM. I set up a test fixture under https://skrattestation.azurewebsites.net/ for now (tracking better implementation in #16314) that you can record against. See our [test setup](https://github.com/Azure/azure-sdk-for-js/pull/16150/files#diff-ce273ee0d8933251856ca06bca630733bbb89b0ecd4cae4b8d2f1ab035beab7aR77-R94) for what is needed there. The important parts are the claims that need to match because those are the claims the test fixture returns.

mccoyp:
Resolves #19588 and resolves #18114.


sdk/keyvault/azure-keyvault-keys/CHANGELOG.md
```diff
@@ -3,6 +3,11 @@
 ## 4.5.0b2 (Unreleased)
 
 ### Features Added
+- Added support for secure key release from a Managed HSM
+  - Added `release_key` method to `KeyClient` for releasing the private component of a key
+  - Added `exportable` and `release_policy` keyword-only arguments to key creation methods
```
heaths, line 8: "key create *and import* methods"?

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_client.py
```diff
@@ -8,7 +8,7 @@
 from ._shared import KeyVaultClientBase
 from ._shared.exceptions import error_map as _error_map
 from ._shared._polling import DeleteRecoverPollingMethod, KeyVaultOperationPoller
-from ._models import DeletedKey, KeyVaultKey, KeyProperties
+from ._models import DeletedKey, KeyVaultKey, KeyProperties, KeyReleaseResult
```
heaths, line 11: Maybe this gets renamed later, but this should be `ReleaseKeyResult` to match our naming convention for the method, `release_key`. Same with `ReleaseKeyOptions`, though I suspect you don't have that in favor of `kwargs`.

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_client.py
```diff
@@ -82,11 +88,18 @@ def create_key(self, name, key_type, **kwargs):
         enabled = kwargs.pop("enabled", None)
         not_before = kwargs.pop("not_before", None)
         expires_on = kwargs.pop("expires_on", None)
-        if enabled is not None or not_before is not None or expires_on is not None:
-            attributes = self._models.KeyAttributes(enabled=enabled, not_before=not_before, expires=expires_on)
+        exportable = kwargs.pop("exportable", None)
+
+        if enabled is not None or not_before is not None or expires_on is not None or exportable is not None:
```
heaths, line 93: Suggestion: refactor this into a method you can reuse elsewhere, if indeed it checked elsewhere. In .NET I have 2 or 3 places I need to check this, and keeping each location up to date is risky. Instead, refactoring into a method you can use on the object to de/serialize consolidates the logic.

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_client.py
```diff
@@ -82,11 +88,18 @@ def create_key(self, name, key_type, **kwargs):
...
+            attributes = self._models.KeyAttributes(
+                enabled=enabled, not_before=not_before, expires=expires_on, exportable=exportable
```
heaths, line 95: Why is the lvalue for `expires` not the same as `expires_on`. All the other properties match. Is this correct?

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_client.py
```diff
@@ -579,29 +623,42 @@ def import_key(self, name, key, **kwargs):
...
         :raises: :class:`~azure.core.exceptions.HttpResponseError`
         """
         enabled = kwargs.pop("enabled", None)
         not_before = kwargs.pop("not_before", None)
         expires_on = kwargs.pop("expires_on", None)
-        if enabled is not None or not_before is not None or expires_on is not None:
-            attributes = self._models.KeyAttributes(enabled=enabled, not_before=not_before, expires=expires_on)
+        exportable = kwargs.pop("exportable", None)
+
+        if enabled is not None or not_before is not None or expires_on is not None or exportable is not None:
```
heaths, line 646: Here's a second instance of the same logic.

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_client.py
```diff
@@ -613,12 +670,45 @@ def import_key(self, name, key, **kwargs):
         )
         return KeyVaultKey._from_key_bundle(bundle)
 
+    @distributed_trace
+    def release_key(self, name, target, version=None, **kwargs):
```
heaths, line 674: Maybe it's different for Python, but in past archboard meetings they wanted `name` and `version` to be kept in the same order. I realize Python doesn't support overloading, though, so maybe this is acceptable. I would talk with your architects if you're not sure.

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_models.py
```diff
@@ -210,6 +210,46 @@ def managed(self):
...
+    :paramtype content_type: str
+    """
+
+    def __init__(self, data, **kwargs):
+        # type: (bytes, **Any) -> None
+        self.data = data
+        self.content_type = kwargs.get("content_type", None)
+
+
+class KeyReleaseResult(object):
```
heaths, line 243: `ReleaseKeyResult` since it's associated with the `release_key` method.

heaths (changes requested):

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_client.py
```diff
@@ -613,12 +670,45 @@ def import_key(self, name, key, **kwargs):
...
+        :keyword str nonce: A client-provided nonce for freshness.
+
+        :return: The result of the key release.
+        :rtype: ~azure.keyvault.keys.KeyReleaseResult
+        :raises: :class:`~azure.core.exceptions.HttpResponseError`
+        """
+        result = self._client.release(
+            vault_base_url=self._vault_url,
+            key_name=name,
+            key_version=version or "",
```
chlowell, line 697: `None` isn't functionally the same?

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_client.py
```diff
@@ -125,6 +140,10 @@ def create_rsa_key(self, name, **kwargs):
         :paramtype tags: dict[str, str]
         :keyword ~datetime.datetime not_before: Not before date of the key in UTC
         :keyword ~datetime.datetime expires_on: Expiry date of the key in UTC
+        :keyword bool exportable: Whether the private key can be exported.
```
chlowell, line 143:
```suggestion
        :keyword bool exportable: Whether the key can be exported.
```
There's no public/private separation here, it's just a single key

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/aio/_client.py
```diff
@@ -470,6 +509,9 @@ async def update_key_properties(self, name: str, version: "Optional[str]" = None
         else:
             attributes = None
 
+        policy = kwargs.pop("release_policy", None)
+        if policy is not None:
+            policy = self._models.KeyReleasePolicy(content_type=policy.content_type, data=policy.data)
```
chlowell, line 514: Should this be passed to KeyUpdateParameters?

sdk/keyvault/azure-keyvault-keys/tests/test_key_client.py
```diff
@@ -427,6 +429,61 @@ def test_get_random_bytes(self, client, **kwargs):
...
+
+        release_result = client.release_key(imported_key_name, attestation)
+        assert release_result.value
+
+    @only_hsm_7_3_preview()
+    @client_setup
+    def test_exportable_without_release_policy_error(self, client, **kwargs):
+        """A key cannot be created as exportable without having a release policy"""
+        key_name = self.get_resource_name("key-name")
+        with pytest.raises(HttpResponseError):
```
chlowell, line 473: Why test the service's argument validation?

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_client.py
mccoyp: The generated KeyAttributes model accepts an `expires` kwarg, but we rename this as `expires_on` as a method kwarg and as a property of KeyProperties. This is correct; it just looks a little weird

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_client.py
mccoyp: Passing `None` instead of a string raises an error, presumably because `key_version` is technically marked as required in the swagger (even though the service doesn't treat it as required). `get_key` and `update_key_properties` use this same `version or ""` pattern for, I assume, a similar reason

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_client.py
mccoyp: Hm, would a more accurate description be whether the private component of the key can be exported? This wording was based on [JS's](https://github.com/Azure/azure-sdk-for-js/pull/16150/files#diff-bd87b9a2319e7bf24d7284d987cadf4a3d1e6fceb32aa96ef7ffc21f0452a692R219) and the generated model's property description: "Indicates if the private key can be exported."

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/aio/_client.py
mccoyp: It should be; thank you for noticing! I should probably add a test for this.

sdk/keyvault/azure-keyvault-keys/tests/test_key_client.py
mccoyp: There's not really a compelling reason to, other than that JS tests for this and I was curious about what behavior is allowed. You're right, though -- this doesn't have anything to do with SDK code and is superfluous. I'll use this real estate for more valuable tests

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_client.py
heaths: It's a subtlety for oct (AES) since it's a symmetric key, but I agree that Charles's suggestion is more accurate. When declared on a model - or on a class like .NET and Java have - we can only have a single comment and "private key" is more accurate for all 3 supported algorithms than "key" would be misleading.

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_client.py
mccoyp: That's good to know; I'll open an issue to discuss this before the next release. We can't have a required parameter after an optional one, which means we'd need to lead with `target` if we keep `name` and `version` adjacent -- but that feels like an odd ordering since the key is being acted on. @chlowell, do you think this is a good enough signature for the beta?

chlowell: I'm okay with this signature. We'll get more opinions and hear what the architects think before the stable release.

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_shared/client_base.py
```diff
@@ -79,6 +79,13 @@ def __init__(self, vault_url, credential, **kwargs):
                 + "Supported versions: {}".format(", ".join(v.value for v in ApiVersion))
             )
 
+    def _get_attributes(self, enabled, not_before, expires_on, exportable=None):
```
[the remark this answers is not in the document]
mccoyp: Right you are, my mistake. Will fix this in a second

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_client.py
```diff
@@ -95,6 +112,7 @@ def create_key(self, name, key_type, **kwargs):
             tags=kwargs.pop("tags", None),
             curve=kwargs.pop("curve", None),
             public_exponent=kwargs.pop("public_exponent", None),
+            release_policy=policy,
```
chlowell, line 115: Is there test coverage of `policy != None`?

mccoyp: Do you mean testing `create_key` with a non-`None` release policy?

chlowell: Yes, I didn't see/perhaps overlooked a test of creating a key with a release policy

mccoyp: Got it! That's covered in [this test](https://github.com/Azure/azure-sdk-for-python/pull/20048/files#diff-20c2e1ce15e9fdb884cf57de69b095ef3e6cb6683e74d84668f18c25789a496bR450), though it's a little obscured by the helper method

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_client.py
```diff
@@ -486,15 +528,16 @@ def update_key_properties(self, name, version=None, **kwargs):
         enabled = kwargs.pop("enabled", None)
         not_before = kwargs.pop("not_before", None)
         expires_on = kwargs.pop("expires_on", None)
-        if enabled is not None or not_before is not None or expires_on is not None:
-            attributes = self._models.KeyAttributes(enabled=enabled, not_before=not_before, expires=expires_on)
-        else:
-            attributes = None
+        attributes = self._get_attributes(enabled=enabled, not_before=not_before, expires_on=expires_on)
```
chlowell, line 531: Does the service allow setting `exportable` in an update?

sdk/keyvault/test-resources.json
```diff
@@ -187,7 +194,8 @@
                 "enablePurgeProtection": false,
                 "enableSoftDelete": true,
                 "publicNetworkAccess": "Enabled",
-                "networkAcls": "[variables('networkAcls')]"
+                "networkAcls": "[variables('networkAcls')]",
+                "softDeleteRetentionInDays": 7
```
chlowell, line 198: Not sure whether this is retention for the MHSM itself or the resources it contains, but I would think we want the smallest value here.

chlowell (approved):

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_client.py
maorleger: Hope you dont mind me chiming in but no - changing the exportable property on updates should not be allowed according to service team. You should be able to update the release policy though of an exportable key

sdk/keyvault/test-resources.json
mccoyp: According to [documentation](https://docs.microsoft.com/azure/key-vault/managed-hsm/best-practices#turn-on-recovery-options) it looks like we can only set this between 7 and 90 days unfortunately

maorleger (approved):

sdk/keyvault/test-resources.json
chlowell: See how I assumed the smallest value is less than 7 😄

heaths (approved):
Public API surface and shipping code LGTM

Issue #20059, opened 2021-08-03
Document Translation Samples issues


v-xuto:
1.
Section [link1](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/translation/azure-ai-translation-document/samples/async_samples/sample_list_document_statuses_with_filters_async.py#L63), [link2](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/translation/azure-ai-translation-document/samples/sample_list_document_statuses_with_filters.py#L60):
![image](https://user-images.githubusercontent.com/57166602/127979297-d5695f05-2a4c-455f-95cc-e5b75d64b0ec.png)

Reason:
TypeError: request() got an unexpected keyword argument 'translated_after'

Suggestion:
Update to
```
created_after=start,
created_before=end,
```

@jameszliao-msft , @lmazuel , @meeraharidasa , @mayurid , @kristapratico and @jongio for notification.

PR #20063, opened 2021-08-03
[translation] fix samples with renamed params

kristapratico:
Resolves https://github.com/Azure/azure-sdk-for-python/issues/20059

Missed updating these in the renaming PR.


iscai-msft (approved):
:shipit:

maririos (approved):
curious, how did you notice the difference?
is there a way the CI could check if the samples are not updated with the latest code changes?

kristapratico:
> curious, how did you notice the difference?
> is there a way the CI could check if the samples are not updated with the latest code changes?

Our vendor team opened an issue when testing samples. We do run all translation samples through CI _except_ these ones which will depend on user input to set the filters.

maririos:
>Our vendor team opened an issue when testing samples

Wow!!! Love to see that we have that extra validation :)\

Thanks for the context!

Issue #13292, opened 2020-08-24
[Schema Registry] Cached schema registry design


yunhaoling:
- Does the SRClient need cache?
  - built-in
  - or a separate CachedSRClient
  - or no cache, just let the *Serializer have the cache
    - if we don't cache for them, then they need to cache by themselves

see confluent design:

https://github.com/confluentinc/confluent-kafka-python/blob/5aa6439df081ead048932ee395085236ef0bca9d/confluent_kafka/schema_registry/schema_registry_client.py#L185

https://github.com/confluentinc/confluent-kafka-python/blob/master/confluent_kafka/avro/cached_schema_registry_client.py

swathipil:
for now, going with : 

Option 3 - Single client with caching
In this option, there's only one client that can be created by making caching optional - on by default. If user creates the client with caching enabled, APIs can return cached responses when available.

```
public final class ConfigurationClientBuilder {

    // Setup cache policy. If this is not set, default caching policy will be applied.
    public ConfigurationClientBuilder cachePolicy(CachePolicy cachePolicy);

    // build clients
    public ConfigurationClient buildClient();
    public ConfigurationAsyncClient buildAsyncClient();

    // other configurations
}

public final class CachePolicy {
    private CacheMode cacheMode;
    private long maxCacheEntrySize;
    private Duration cacheEntryTimeToLive;

    // getter and setters
}

public enum CacheMode {
    NONE, // this will disable the cache
    LEAST_RECENTLY_USED,
    LEAST_FREQUENTLY_USED;
}
```
One open question with such clients is how to implement the withResponse methods that return the HTTP response information like status code and headers along with the actual response.

Option 3c - Do not offer withResponse APIs for APIs that are eligible for caching
The client will only return T instead of Response<T> for all cache-eligible APIs.
```
public final class ConfigurationClient {
    public ConfigurationSetting getSetting(String key);

    // max overload that may return a cached response
    public ConfigurationSetting getSetting(SettingSelector selector);

    // clear cache
    public void clearCache();
}
```

swathipil:
questions for Anna:
- should CachePolicy be moved to azure-core, even though it's only used by the schema registry client?
- should the properties on CachePolicy have a default? Which defaults?
- should we make cache policy accessible/public to users with an instance var on client? (currently we don't do this, it's private)

PR #20065, opened 2021-08-03
[SchemaRegistry] move cache to client from serializer

swathipil:
addresses: #13292

### DIFF
--- a/sdk/schemaregistry/azure-schemaregistry/README.md
+++ b/sdk/schemaregistry/azure-schemaregistry/README.md
@@ -40,7 +40,7 @@ schema_registry_client = SchemaRegistryClient(endpoint, credential)
 
 - Schema: Schema is the organization or structure for data.
 
-- SchemaRegistryClient: `SchemaRegistryClient ` provides the API for storing and retrieving schemas in schema registry.
+- SchemaRegistryClient: `SchemaRegistryClient` provides the API for storing and retrieving schemas in schema registry.
 
 ## Examples
 
--- a/sdk/schemaregistry/azure-schemaregistry/azure/schemaregistry/_schema_registry_client.py
+++ b/sdk/schemaregistry/azure-schemaregistry/azure/schemaregistry/_schema_registry_client.py
@@ -32,6 +32,7 @@ from ._common._response_handlers import (
     _parse_response_schema_id,
 )
 from ._generated._azure_schema_registry import AzureSchemaRegistry
+from azure.core.exceptions import ResourceNotFoundError
 
 if TYPE_CHECKING:
     from azure.core.credentials import TokenCredential
@@ -73,7 +74,7 @@ class SchemaRegistryClient(object):
     def __exit__(self, *exc_details):
         # type: (Any) -> None
         self._generated_client.__exit__(*exc_details)
-
+    
     def close(self):
         # type: () -> None
         """This method is to close the sockets opened by the client.
@@ -194,9 +195,10 @@ class SchemaRegistryClient(object):
             pass
 
         try:
-            return self._description_to_properties[
+            properties = self._description_to_properties[
                 (schema_group, schema_name, serialization_type, schema_content)
             ]
+            return properties
         except KeyError:
             schema_properties = self._generated_client.schema.query_id_by_content(
                 group_name=schema_group,
@@ -209,4 +211,5 @@ class SchemaRegistryClient(object):
             self._description_to_properties[
                 (schema_group, schema_name, serialization_type, schema_content)
             ] = schema_properties
+            self._id_to_schema[schema_properties.schema_id] = Schema(schema_content, schema_properties)
             return schema_properties
--- a/sdk/schemaregistry/azure-schemaregistry/azure/schemaregistry/_schema_registry_client.py
+++ b/sdk/schemaregistry/azure-schemaregistry/azure/schemaregistry/_schema_registry_client.py
@@ -27,9 +27,11 @@ from typing import Any, TYPE_CHECKING, Union
 
 from ._common._constants import SerializationType
 from ._common._schema import Schema, SchemaProperties
-from ._common._response_handlers import _parse_response_schema, _parse_response_schema_id
+from ._common._response_handlers import (
+    _parse_response_schema,
+    _parse_response_schema_id,
+)
 from ._generated._azure_schema_registry import AzureSchemaRegistry
-from ._policies import CachePolicy
 
 if TYPE_CHECKING:
     from azure.core.credentials import TokenCredential
@@ -43,8 +45,7 @@ class SchemaRegistryClient(object):
     :param str endpoint: The Schema Registry service endpoint, for example my-namespace.servicebus.windows.net.
     :param credential: To authenticate to manage the entities of the SchemaRegistry namespace.
     :type credential: TokenCredential
-    :keyword CachePolicy cache_policy: Mode for caching policy, with "least" being least recently used, "most" being most recently
-        used, and default of None being no caching.
+
     .. admonition:: Example:
 
         .. literalinclude:: ../samples/sync_samples/sample_code_schemaregistry.py
@@ -55,16 +56,14 @@ class SchemaRegistryClient(object):
             :caption: Create a new instance of the SchemaRegistryClient.
 
     """
-    def __init__(
-        self,
-        endpoint,
-        credential,
-        **kwargs
-    ):
+
+    def __init__(self, endpoint, credential, **kwargs):
         # type: (str, TokenCredential, Any) -> None
-        self._generated_client = AzureSchemaRegistry(credential=credential, endpoint=endpoint, **kwargs)
+        self._generated_client = AzureSchemaRegistry(
+            credential=credential, endpoint=endpoint, **kwargs
+        )
+        self._description_to_properties = {}
         self._id_to_schema = {}
-        self._schema_to_id = {}
 
     def __enter__(self):
         # type: () -> SchemaRegistryClient
@@ -77,12 +76,14 @@ class SchemaRegistryClient(object):
 
     def close(self):
         # type: () -> None
-        """ This method is to close the sockets opened by the client.
+        """This method is to close the sockets opened by the client.
         It need not be used when using with a context manager.
         """
         self._generated_client.close()
 
-    def register_schema(self, schema_group, schema_name, serialization_type, schema_content, **kwargs):
+    def register_schema(
+        self, schema_group, schema_name, serialization_type, schema_content, **kwargs
+    ):
         # type: (str, str, Union[str, SerializationType], str, Any) -> SchemaProperties
         """
         Register new schema. If schema of specified name does not exist in specified group,
@@ -112,7 +113,7 @@ class SchemaRegistryClient(object):
         except AttributeError:
             pass
 
-        return self._generated_client.schema.register(
+        schema_properties = self._generated_client.schema.register(
             group_name=schema_group,
             schema_name=schema_name,
             schema_content=schema_content,
@@ -120,6 +121,18 @@ class SchemaRegistryClient(object):
             cls=_parse_response_schema_id,
             **kwargs
         )
+        schema_description = (
+            schema_group,
+            schema_name,
+            serialization_type,
+            schema_content,
+        )
+        self._id_to_schema[schema_properties.schema_id] = Schema(
+            schema_content, schema_properties
+        )
+        self._description_to_properties[schema_description] = schema_properties
+
+        return schema_properties
 
     def get_schema(self, schema_id, **kwargs):
         # type: (str, Any) -> Schema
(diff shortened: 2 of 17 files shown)


sdk/schemaregistry/azure-schemaregistry/azure/schemaregistry/_schema_registry_client.py
```diff
@@ -61,6 +63,8 @@ def __init__(
     ):
         # type: (str, TokenCredential, Any) -> None
         self._generated_client = AzureSchemaRegistry(credential=credential, endpoint=endpoint, **kwargs)
+        self._id_to_schema = {}
```
yunhaoling, line 66: FYI: the map would be like this:
key: schema_id which is string
value: Schema(schema_name, schema_content, serilization_type, schema_group)

sdk/schemaregistry/azure-schemaregistry/azure/schemaregistry/_schema_registry_client.py
```diff
@@ -136,13 +153,18 @@ def get_schema(self, schema_id, **kwargs):
...
-        )
-
-    def get_schema_id(self, schema_group, schema_name, serialization_type, schema_content, **kwargs):
+        try:
+            return self._id_to_schema[schema_id]
+        except KeyError:
+            schema = self._generated_client.schema.get_by_id(
+                schema_id, cls=_parse_response_schema, **kwargs
+            )
+            self._id_to_schema[schema_id] = schema
```
yunhaoling, line 162: I'm wondering if the `self._description_to_properties` cache could also be updated in the `get_schema`?
seems every piece of info needed for building an entry in the `self._description_to_properties` could be extracted from the schema.

and if doing such we could get back the `_add_cache` method?

sdk/schemaregistry/azure-schemaregistry/azure/schemaregistry/_schema_registry_client.py
```diff
@@ -171,11 +193,20 @@ def get_schema_id(self, schema_group, schema_name, serialization_type, schema_co
...
+                group_name=schema_group,
+                schema_name=schema_name,
+                schema_content=schema_content,
+                x_schema_type=serialization_type,
+                cls=_parse_response_schema_id,
+                **kwargs
+            )
+            self._description_to_properties[
+                (schema_group, schema_name, serialization_type, schema_content)
+            ] = schema_properties
```
yunhaoling, line 211: same comment as the `get_schema` method, we could also update `self._id_to_schema` here?

sdk/schemaregistry/azure-schemaregistry/azure/schemaregistry/_schema_registry_client.py
swathipil: I think in this case, the other `_description_to_properties` cache cannot be updated at the same time b/c the other info required to get description (schema_group, schema_name, serialization_type) are all not available here. The only pieces of info available are schema_content and schema_properties. I'm not sure whether it's possible to update that dict here, unless you have a different idea?

sdk/schemaregistry/azure-schemaregistry/azure/schemaregistry/_schema_registry_client.py
swathipil: in this case, I think it makes sense to also update the other dict!

sdk/schemaregistry/azure-schemaregistry/azure/schemaregistry/_schema_registry_client.py
swathipil: now that you mention it though, I think it does seem like the other cache needs to be updated somehow, to remain consistent with this one. How do you feel about a `_id_to_description` cache? (which can be used to update the `description_to_properties` cache)

yunhaoling: correct me if I'm wrong, the SchemaProperties is an ivar of the Schema object, so I think the schema_group and schema_name could be extracted from `schema.schema_properties`:

`schema.schema_properties.schema_group`,  `schema.schema_properties.schema_name`

swathipil: so those two are the only ivars that are not in SchemaProperties, unfortunately. Here is the list of ivars for SchemaProperties:
    :ivar schema_id: References specific schema in registry namespace.
    :type schema_id: str
    :ivar location: URL location of schema, identified by schema group, schema name, and version.
    :type location: str
    :ivar location_by_id: URL location of schema, identified by schema ID.
    :type location_by_id: str
    :ivar serialization_type: Serialization type for the schema being stored.
    :type serialization_type: str
    :ivar version: Version of the returned schema.
    :type version: int

yunhaoling: my bad for not checking the class

I'm good with an extra `_id_to_description` cache for now, but we definitely should revisit this later.
I don't feel good with the three asymmetry cache maps.

swathipil: agreed, will write add it to the design page

swathipil: final 'offline' conversation:
- decided to not add another asymmetric map, since it would definitely decrease comprehensibility
- for now, to avoid this, we can follow the JS approach, which was to:
1) update both maps for the [`getSchemaId` method](https://github.com/Azure/azure-sdk-for-js/blob/934968168f7464d5a5aa8e8717d1acb7cc77dcd0/sdk/schemaregistry/schema-registry/src/schemaRegistryClient.ts#L125
2) update only the `idToSchemaMap` in the [`getSchemaById` method](https://github.com/Azure/azure-sdk-for-js/blob/934968168f7464d5a5aa8e8717d1acb7cc77dcd0/sdk/schemaregistry/schema-registry/src/schemaRegistryClient.ts#L155)

sdk/schemaregistry/azure-schemaregistry/tests/async_tests/recordings/test_schema_registry_async.test_schema_basic_async.yaml
```diff
@@ -9,58 +9,30 @@ interactions:
...
   response:
     body:
-      string: '{"id":"a8e62268f67e49afb6bcd87537bf2557"}'
+      string: '{"id":"fcf0c4ae8130497aab2f29c493b980b6"}'
     headers:
       content-type: application/json
-      date: Thu, 12 Nov 2020 05:59:46 GMT
-      location: https://fake_resource.servicebus.windows.net:443/$schemagroups/fakegroup/schemas/test-schema-basic-asynce5e1482/versions/1?api-version=2017-04
+      date: Tue, 10 Aug 2021 23:09:09 GMT
+      location: https://swathip-test-eventhubs.servicebus.windows.net:443/$schemagroups/fakegroup/schemas/test-schema-basic-asynce5e1482/versions/1?api-version=2017-04
```
swathipil, line 23: I think this is mutating the url (getting sanitized here)?

sdk/schemaregistry/azure-schemaregistry/tests/async_tests/recordings/test_schema_registry_async.test_schema_basic_async.yaml
```diff
@@ -9,58 +9,30 @@ interactions:
       Content-Type:
       - application/json
       User-Agent:
-      - azsdk-python-azureschemaregistry/1.0.0b1 Python/3.9.0rc1 (Windows-10-10.0.19041-SP0)
+      - azsdk-python-azureschemaregistry/1.0.0b1 Python/3.9.0 (Windows-10-10.0.19041-SP0)
       X-Schema-Type:
       - Avro
     method: PUT
-    uri: https://fake_resource.servicebus.windows.net/$schemagroups/fakegroup/schemas/test-schema-basic-asynce5e1482?api-version=2017-04
+    uri: https://fake_resource.servicebus.windows.net$schemagroups/fakegroup/schemas/test-schema-basic-asynce5e1482?api-version=2017-04
```
kristapratico, line 16: it looks like we're losing the `/` here

sdk/schemaregistry/azure-schemaregistry/tests/test_schema_registry.py
```diff
@@ -41,7 +41,11 @@ def test_schema_basic(self, schemaregistry_endpoint, schemaregistry_group, **kwa
         schema_name = self.get_resource_name('test-schema-basic')
         schema_str = """{"namespace":"example.avro","type":"record","name":"User","fields":[{"name":"name","type":"string"},{"name":"favorite_number","type":["int","null"]},{"name":"favorite_color","type":["string","null"]}]}"""
         serialization_type = "Avro"
+        assert len(client._id_to_schema) == 0
```
yunhaoling, line 44: async tests?

sdk/schemaregistry/azure-schemaregistry/azure/schemaregistry/_schema_registry_client.py
```diff
@@ -171,11 +193,22 @@ def get_schema_id(self, schema_group, schema_name, serialization_type, schema_co
...
+            self._id_to_schema[schema_properties.schema_id] = Schema(schema_content, schema_properties)
```
yunhaoling, line 213: it is possible that schema_properties.schema_id is already in the `self._id_to_schame` such that we're overwriting the entry here?.

e.g. if a user calls `get_schema` first, then call `get_schema_id`

sdk/schemaregistry/azure-schemaregistry/azure/schemaregistry/aio/_schema_registry_client_async.py
```diff
@@ -181,11 +200,21 @@ async def get_schema_id(
...
+                schema_name=schema_name,
+                schema_content=schema_content,
+                x_schema_type=serialization_type,
+                cls=_parse_response_schema_id,
+                **kwargs
+            )
+            self._description_to_properties[
+                (schema_group, schema_name, serialization_type, schema_content)
+            ] = schema_properties
+            self._id_to_schema[schema_properties.schema_id] = Schema(schema_content, schema_properties)
```
yunhaoling, line 219: it is possible that schema_properties.schema_id is already in the `self._id_to_schame` such that we're overwriting the item here?.

e.g. if a user calls `get_schema` first, then call `get_schema_id`

sdk/schemaregistry/azure-schemaregistry/README.md
```diff
@@ -40,7 +40,7 @@ schema_registry_client = SchemaRegistryClient(endpoint, credential)
 
 - Schema: Schema is the organization or structure for data.
 
-- SchemaRegistryClient: `SchemaRegistryClient ` provides the API for storing and retrieving schemas in schema registry.
+- SchemaRegistryClient: `SchemaRegistryClient` provides the API for storing and retrieving schemas in schema registry.
```
yunhaoling, line 43: would you mind updating the changelog, version number (if needed)?

sdk/schemaregistry/azure-schemaregistry/azure/schemaregistry/_schema_registry_client.py
swathipil: that's a good question, I think it would overwrite the existing schema. We can do something like the following to ensure it's not overwritten:
```suggestion
            cached_schema = self._id_to_schema.get(schema_properties.schema_id)
            if cached_schema:
                self._id_to_schema[schema_properties.schema_id].schema_properties = schema_properties
            else:
                self._id_to_schema[schema_properties.schema_id] = Schema(schema_content, schema_properties)
```
thoughts?

sdk/schemaregistry/azure-schemaregistry/azure/schemaregistry/aio/_schema_registry_client_async.py
swathipil: same response as above

sdk/schemaregistry/azure-schemaregistry/azure/schemaregistry/_schema_registry_client.py
yunhaoling: I think we only need to `else` part that when it's not cached -- if it's already cached, then why would we need to update it?

swathipil: originally I was thinking that (when the schema is registered through the portal and not with `client.register_schema`), the call to the service to get schema_properties in `get_schema_id` would create a new `SchemaProperties` object that is different from the `SchemaProperties` object in `schema.schema_properties`. I wanted both the `schema_properties` object in the `_description_to_properties` cache and `schema.schema_properties` in `_id_to_schema` cache to point to be the same object.

yunhaoling (approved):
LGTM

Issue #19362, opened 2021-06-21
azure-servicebus.AutoLockRenewer change ThreadPool usage


michaelcredera:
**Is your feature request related to a problem? Please describe.**
The azure.servicebus.AutoLockRenewer is implicitly limited to renewing locks for at most `max_workers` number of messages. For an application that takes many messages off of the queue for parallel processing, this can present problems.

The azure.servicebus.AutoLockRenewer uses a single future inside of a ThreadPoolExecutor to monitor messages and renew locks. This means it's limited to monitoring at most `max_workers` messages at a time (since a thread pool only runs one future per-thread until the future completes).

**Describe the solution you'd like**
An alternative implementation could instead use `max_workers` futures monitoring a set of messages per future rather than a single message per future. 

**Describe alternatives you've considered**
An alternative solution would be to implement parallelism with each worker taking a message off of the queue for itself to process. This is certainly possible, but there are still cases where taking more than `max_workers` messages off of the queue could be desirable (e.g. prefetching messages).

yunhaoling:
hey @michaelcredera , thanks for your feedback, I'll take a look into it.

yunhaoling:
hey @michaelcredera , sorry we have limited resourcing currently. would you like to try making a PR for the change? I would be glad to review it😊 (It's completely fine if you don't have the time, just FYI we're an open-source library and we welcome user contributions)

michaelcredera:
Hey @yunhaoling - I can try to get a PR up this week

michaelcredera:
@yunhaoling Are there branching permissions on the repo? I didn't see anything in the contribution guide, but I'm getting a 403 error pushing my branch:
```
git push --set-upstream origin feature/19362-threadpool-improvements-for-servicebus
...
fatal: unable to access 'https://github.com/Azure/azure-sdk-for-python.git/': The requested URL returned error: 403
```

yunhaoling:
hey @michaelcredera , creating a branch directly in the azure-sdk-for-python repo is not allowed for external contributors.
instead, you could fork the azure-sdk-for-python repo and create a PR from your repo branch.

yunhaoling:
hey @michaelcredera , just friendly checking if you need any help on the PR?

yunhaoling:
hey @michaelcredera , we have improved the implementation in the release v7.3.2 so that the message count that auto lock renewer could handle is not limited to the max_worker value, sorry for keeping you waiting and please try it out.

[...]

PR #20071, opened 2021-08-03
[ServiceBus] Improve auto lock renewer implementation

yunhaoling:
addressing issue: https://github.com/Azure/azure-sdk-for-python/issues/19362

currently sync auto lock renewer is implemented the way that each thread in the ThreadPoolExecutor handles lock renewal for one message/session at a time.

This could bring trouble when the amount of messages needed to be auto lock renewed is greater than the max workers of the executor -- E.g., executor has 4 workers, user register 10 messages, only the first 4 messages will be auto lock renewed.

This PR has proposed a new approach to support messages amount > max workers amount which is described as following:

The overall design is to decouple the renew process via a queue so that worker threads could keep popping and handling renew tasks from the queue.

The algorithm is:

- a: If users pass the max worker
  - a-1: if max worker is 1, then we have a single thread, which keeps pop renew task from -> check -> renew if needed -> push back renew tasks
  - a-2: if max worker is None or > 1, then the model is we have 1 dispatching thread, and (max_worker - 1) renew worker threads. The dispatching thread keeps popping task meta info from the queue and submit them to the executor. The worker thread do the renew work if needed and push the task meta info back into the queue if the object is still renewable.
- b: If users pass the executor, there's no public api in python to tell us the max worker value of an executor. we have to infer whether max_worker > 1 and the dispatching model could be applied. A inferring process will be started at the very beginning.
  - the inferring process would submit two tasks to the executor. each tasks contains a while loop and will exit within the given time. The task is majorly setting a flag. **If max_workers > 1, then two tasks will be run in parallel, both of the flag could be set. If max_workers is 1, then only one task will be run, only one flag is set.** Based on the algorithm, we could tell whether the executor max_workers is > 1 or not.
  - with the inferred max worker info, we could go to part a.


sdk/servicebus/azure-servicebus/azure/servicebus/_common/auto_lock_renewer.py
```diff
@@ -91,23 +96,57 @@ def __init__(
...
+        self._infer_max_workers_flags = [0, 0]
 
     def __enter__(self):
         if self._shutdown.is_set():
             raise ServiceBusError(
                 "The AutoLockRenewer has already been shutdown. Please create a new instance for"
                 " auto lock renewing."
             )
+
+        self._init_workers()
```
annatisch, line 120: Will it be a problem if the user code isn't using a context manager?

sdk/servicebus/azure-servicebus/azure/servicebus/_common/auto_lock_renewer.py
```diff
@@ -91,23 +96,57 @@ def __init__(
...
+            self._executor.submit(self._infer_max_workers_value_worker, 1, end_time)
+            time.sleep(self._infer_max_workers_time)
+            # if max_workers > 1, then the two submitted work load would be in run in parallel
+            # sum(self._infer_max_workers_flags) == 2, otherwise the sum would just be 1
+            self._is_max_workers_greater_than_one = (sum(self._infer_max_workers_flags) == 2)
+
+    def _infer_max_workers_value_worker(self, flag_idx, stop_time):
+        while time.time() < stop_time:
+            self._infer_max_workers_flags[flag_idx] = 1
+            time.sleep(0.05)
```
annatisch, line 148: Rather than do a time-based approach - maybe we could force a deadlock in the case of a single thread?
There's some examples in the Python docs:
https://docs.python.org/3/library/concurrent.futures.html#threadpoolexecutor

```python
def wait_on_future():
    f = executor.submit(pow, 5, 2)
    # This will never complete because there is only one worker thread and
    # it is executing this function.
    print(f.result())

executor = ThreadPoolExecutor(max_workers=1)
dispatcher_check = executor.submit(wait_on_future)
try:
    dispatcher_check.result(timeout=self._dispatch_check_timeout)
    # All good - we have more than one thread
except TimeoutError:
    # Hit a deadlock - assume we have a single thread in the pool
```

sdk/servicebus/azure-servicebus/azure/servicebus/_common/auto_lock_renewer.py
```diff
@@ -120,7 +159,26 @@ def _renewable(self, renewable):
             return False
         return True
 
-    def _auto_lock_renew(
+    def _dispatch_worker(self):
+        self._last_activity_timestamp = time.time()
+        while not self._shutdown.is_set() and self._running.is_set():
```
annatisch, line 164: What's the distinction between the 2 flags here?

sdk/servicebus/azure-servicebus/azure/servicebus/_common/auto_lock_renewer.py
```diff
@@ -120,7 +159,26 @@ def _renewable(self, renewable):
...
+                else:
+                    self._auto_lock_renew_task(*renew_task)
+                self._renew_tasks.task_done()
+                self._last_activity_timestamp = time.time()
+            # If there's no activity in the past self._idle_timeout seconds, exit the method
+            # This ensures the dispatching thread could exit, not blocking the main python thread
+            # the main worker thread could be started again if new tasks get registered
+            if time.time() - self._last_activity_timestamp >= self._idle_timeout:
+                self._running.clear()
+                self._last_activity_timestamp = None
```
annatisch, line 178: This shuts down the renew thread after a period of inactivity? That seems like a big change in behaviour?

annatisch: Nevermind - I actually read the comment. 
Why would the dispatching thread be blocking the main Python thread?

sdk/servicebus/azure-servicebus/tests/test_queues.py
```diff
@@ -974,12 +975,12 @@ def test_queue_by_queue_client_conn_str_receive_handler_with_autolockrenew(self,
                     if not messages:
                         messages.append(message)
                         assert not message._lock_expired
-                        renewer.register(receiver, message, max_lock_renewal_duration=10)
+                        renewer.register(receiver, message, max_lock_renewal_duration=20)
```
annatisch, line 978: I'm curious why we needed to bump all the durations in the test - it wouldn't work with the existing values?

sdk/servicebus/azure-servicebus/azure/servicebus/_common/auto_lock_renewer.py
yunhaoling: the _init_workers is also called in the register method, making sure the non-context manager case is covered

sdk/servicebus/azure-servicebus/azure/servicebus/_common/auto_lock_renewer.py
```diff
@@ -91,23 +96,57 @@ def __init__(
...
         self._executor = executor or ThreadPoolExecutor(max_workers=max_workers)
+        # None indicates it's unknown whether the provided executor has max worker > 1
+        self._is_max_workers_greater_than_one = None if executor else (max_workers is None or max_workers > 1)
         self._shutdown = threading.Event()
-        self._sleep_time = 1
+        self._sleep_time = 0.5
         self._renew_period = 10
+        self._running = threading.Event()  # indicate whether the main worker thread is running
+        self._last_activity_timestamp = None  # the last timestamp when the main worker is active dealing with tasks
+        self._idle_timeout = 10  # the idle time that main worker thead should exist if there's no activity
```
swathipil, line 106: baby nit:
```suggestion
        self._idle_timeout = 10  # the idle time that main worker thread should exist if there's no activity
```

sdk/servicebus/azure-servicebus/azure/servicebus/_common/auto_lock_renewer.py
yunhaoling: this is much more neat, I like it, will try it out

sdk/servicebus/azure-servicebus/azure/servicebus/_common/auto_lock_renewer.py
```diff
@@ -120,7 +159,26 @@ def _renewable(self, renewable):
...
+    def _dispatch_worker(self):
+        self._last_activity_timestamp = time.time()
+        while not self._shutdown.is_set() and self._running.is_set():
+            while not self._renew_tasks.empty():
+                renew_task = self._renew_tasks.get()
+                if self._is_max_workers_greater_than_one:
+                    self._executor.submit(self._auto_lock_renew_task, *renew_task)
+                else:
+                    self._auto_lock_renew_task(*renew_task)
+                self._renew_tasks.task_done()
```
swathipil, line 171: should this be renew_task.done()? since `self._renew_tasks` is a queue, which doesn't have a `task_done()` method.

sdk/servicebus/azure-servicebus/azure/servicebus/_common/auto_lock_renewer.py
```diff
@@ -163,6 +218,18 @@ def _auto_lock_renew(
                         # Renewable is a message
                         receiver.renew_message_lock(renewable)  # type: ignore
                 time.sleep(self._sleep_time)
+            if self._renewable(renewable):
```
swathipil, line 221: is there a reason there are 2 `if self._renewable(renewable):` lines? (other at 195). can the below be moved into the block above?

sdk/servicebus/azure-servicebus/azure/servicebus/_common/auto_lock_renewer.py
```diff
@@ -91,23 +96,57 @@ def __init__(
...
-        self._sleep_time = 1
+        self._sleep_time = 0.5
         self._renew_period = 10
+        self._running = threading.Event()  # indicate whether the main worker thread is running
+        self._last_activity_timestamp = None  # the last timestamp when the main worker is active dealing with tasks
+        self._idle_timeout = 10  # the idle time that main worker thead should exist if there's no activity
         self._max_lock_renewal_duration = max_lock_renewal_duration
         self._on_lock_renew_failure = on_lock_renew_failure
+        self._renew_tasks = queue.Queue()  # type: ignore
+        self._infer_max_workers_time = 1
```
swathipil, line 110: out of curiosity, why choose 1 sec as time, rather than 0.5, etc?

sdk/servicebus/azure-servicebus/azure/servicebus/_common/auto_lock_renewer.py
```diff
@@ -91,23 +96,57 @@ def __init__(
...
         self.close()
 
+    def _init_workers(self):
+        self._infer_max_workers_greater_than_one_if_needed()
+        if not self._running.is_set():
+            self._running.set()
+            self._executor.submit(self._dispatch_worker)
+
+    def _infer_max_workers_greater_than_one_if_needed(self):
+        # infer max_workers value if executor is passed in
```
swathipil, line 133: would inferring max_workers by sleeping for 1 sec. (like below) increase the time it takes to run this? out of curiosity, is this insignificant enough of perf difference that we can ignore it?

also why do this rather than checking max workers with `executor._max_workers`? Is that bad by convention b/c we (as you mentioned in your description) only want to use public API?

sdk/servicebus/azure-servicebus/azure/servicebus/_common/auto_lock_renewer.py
```diff
@@ -31,7 +36,7 @@
 )
 
 
-class AutoLockRenewer(object):
+class AutoLockRenewer(object):  # pylint:disable=too-many-instance-attributes
```
swathipil, line 39: should async AutoLockRenewer be updated too?

yunhaoling: async is working differently than the sync one, it relies on the built-in `asyncio.ensure_future` which doesn't have the limitation as the sync ThreadPoolExecutor one

sdk/servicebus/azure-servicebus/azure/servicebus/_common/auto_lock_renewer.py
yunhaoling: re1: it's a one time cost during the very beginning when auto lock renewer just starts, once the max_worker is inferred, `_infer_max_workers_greater_than_one_if_needed` would just be a no-op.

re2: right, the idea is not to rely on private ivars (hacky), which is unstable and could be changed by the python team.

sdk/servicebus/azure-servicebus/azure/servicebus/_common/auto_lock_renewer.py
yunhaoling: sorry I didn't have a good reason for choosing 1 -- largely based on my instinct.

1 seconds doesn't waste too much time, while a too small time might not be enough to let the python scheduler run and execute the threads.

sdk/servicebus/azure-servicebus/azure/servicebus/_common/auto_lock_renewer.py
yunhaoling: yea, I think you're right, it should be moved into the block above

sdk/servicebus/azure-servicebus/azure/servicebus/_common/auto_lock_renewer.py
yunhaoling: we're using the `Queue.queue` class which is synchronized queue (thread safe) and has the task_done method.

reference could be found here: https://docs.python.org/3/library/queue.html#queue.Queue.task_done

sdk/servicebus/azure-servicebus/azure/servicebus/_common/auto_lock_renewer.py
yunhaoling: the first shutdown is for the whole life cycle control of the renewer, once shut down is set (close called or exit the context manager), it could no longer be used.

the second is for controling dispatch thread

sdk/servicebus/azure-servicebus/azure/servicebus/_common/auto_lock_renewer.py
yunhaoling: as oob discussion, this is to handle the case that error being raised in the main python thread -- in such case if errors are not caught, then code after error raised (e.g. shutdown the renewer) can not be reached, so that the dispatcher is already running.

The idle timeout gives a way to let the dispatcher thread exit the loop

sdk/servicebus/azure-servicebus/tests/test_queues.py
yunhaoling: I revert the values, we should be good :)

sdk/servicebus/azure-servicebus/azure/servicebus/_common/auto_lock_renewer.py
annatisch: Gotcha - maybe we could use a more descriptive name for the second? e.g. `self._running_dispatcher`

sdk/servicebus/azure-servicebus/azure/servicebus/_common/auto_lock_renewer.py
```diff
@@ -163,6 +215,18 @@ def _auto_lock_renew(
                         # Renewable is a message
                         receiver.renew_message_lock(renewable)  # type: ignore
                 time.sleep(self._sleep_time)
```
annatisch, line 217: I know this is not new - but why was the sleep necessary here?

yunhaoling: as oob discussion: leaving it here to save CPU cycles

sdk/servicebus/azure-servicebus/azure/servicebus/_common/auto_lock_renewer.py
```diff
@@ -91,23 +96,54 @@ def __init__(
...
             )
+
+        self._init_workers()
         return self
 
     def __exit__(self, *args):
         self.close()
 
+    def _init_workers(self):
+        self._infer_max_workers_greater_than_one_if_needed()
```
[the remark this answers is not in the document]
yunhaoling: yep, I think we could, just moved

sdk/servicebus/azure-servicebus/azure/servicebus/_common/auto_lock_renewer.py
```diff
@@ -91,23 +96,54 @@ def __init__(
         :type max_workers: Optional[int]
         """
         self._executor = executor or ThreadPoolExecutor(max_workers=max_workers)
+        # None indicates it's unknown whether the provided executor has max worker > 1
+        self._is_max_workers_greater_than_one = None if executor else (max_workers is None or max_workers > 1)
         self._shutdown = threading.Event()
-        self._sleep_time = 1
+        self._sleep_time = 0.5
         self._renew_period = 10
+        self._running = threading.Event()  # indicate whether the main worker thread is running
```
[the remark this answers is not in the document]
yunhaoling: updated

sdk/servicebus/azure-servicebus/azure/servicebus/_common/auto_lock_renewer.py
```diff
@@ -91,23 +96,54 @@ def __init__(
...
+        self._last_activity_timestamp = None  # the last timestamp when the main worker is active dealing with tasks
+        self._idle_timeout = 10  # the idle time that main worker thread should exist if there's no activity
```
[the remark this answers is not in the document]
yunhaoling: as oob discussion: not exposing it now to avoid confusion to customers, could expose if there're customer requests.

also update the timeout to 5 seconds.

sdk/servicebus/azure-servicebus/azure/servicebus/_common/auto_lock_renewer.py
yunhaoling: name updated

annatisch (approved):
LGTM :)

swathipil (approved):
lgtm!

PR #20097, opened 2021-08-05
[Key Vault] Revise KeyClient.get_random_bytes

### LINKED ISSUE #19895: [Key Vault] Revise get_random_bytes method
- [x] Enforce having to request at least 1 byte
- [x] Return a model instead of bytes directly

mccoyp:
Resolves #19895.

### DIFF
--- a/sdk/keyvault/azure-keyvault-keys/CHANGELOG.md
+++ b/sdk/keyvault/azure-keyvault-keys/CHANGELOG.md
@@ -5,11 +5,14 @@
 ### Features Added
 
 ### Breaking Changes
+> These changes do not impact the API of stable versions such as 4.4.0.
+> Only code written against a beta version such as 4.5.0b1 may be affected.
+- `KeyClient.get_random_bytes` now returns a `RandomBytes` model with bytes in a `value`
+  property, rather than returning the bytes directly
 
 ### Bugs Fixed
 
 ### Other Changes
-- `KeyClient.get_random_bytes` now returns a `RandomBytes` model
 
 ## 4.5.0b1 (2021-07-08)
 Beginning with this release, this library requires Python 2.7 or 3.6+.
--- a/sdk/keyvault/azure-keyvault-keys/CHANGELOG.md
+++ b/sdk/keyvault/azure-keyvault-keys/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Bugs Fixed
 
 ### Other Changes
+- `KeyClient.get_random_bytes` now returns a `RandomBytes` model
 
 ## 4.5.0b1 (2021-07-08)
 Beginning with this release, this library requires Python 2.7 or 3.6+.
--- a/sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_models.py
+++ b/sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_models.py
@@ -441,7 +441,7 @@ class DeletedKey(KeyVaultKey):
 
 
 class RandomBytes(object):
-    """Contains random bytes returned from :func:`~azure.keyvault.keys.KeyClient.get_random_bytes`.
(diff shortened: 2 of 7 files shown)


sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_models.py
```diff
@@ -438,3 +438,13 @@ def scheduled_purge_date(self):
         :rtype: ~datetime.datetime
         """
         return self._scheduled_purge_date
+
+
+class RandomBytes(object):
+    """Contains random bytes returned from :func:`~azure.keyvault.keys.KeyClient.get_random_bytes`.
```
chlowell, line 444: Having overthought this, I think we should either not have a link back to the client, or add a link to the async method:
- I expect most readers to come from this link i.e. reading the method docs, they want to see details of the model
- someone working on an async application may follow this to the sync client's docs
    - probably they will survive but there's a potential for confusion around async support

A second link would require fully qualified text. That isn't a guaranteed win and makes the whole thing harder to read. So, I guess I would lean toward replacing the link with "a Managed HSM", but really I'm just trying to provide food for thought here 😇

sdk/keyvault/azure-keyvault-keys/tests/test_key_client.py
```diff
@@ -422,12 +423,20 @@ def test_get_random_bytes(self, client, **kwargs):
 
         generated_random_bytes = []
         for i in range(5):
-            random_bytes = client.get_random_bytes(count=8)
+            random_bytes = client.get_random_bytes(count=8).value
```
chlowell, line 426: Consider reworking this test into an example we can pull into the method's docstring. Now that the method returns a model, I think we need to demonstrate getting the bytes off it.

sdk/keyvault/azure-keyvault-keys/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 ### Bugs Fixed
 
 ### Other Changes
+- `KeyClient.get_random_bytes` now returns a `RandomBytes` model
```
chlowell, line 12: It's a breaking change 💣

mccoyp: Ope, wrong header 😶

sdk/keyvault/azure-keyvault-keys/tests/test_keys_async.py
```diff
@@ -453,11 +453,14 @@ async def test_get_random_bytes(self, client, **kwargs):
...
-            assert len(bytes_value) == 8
-            assert all([bytes_value != rb] for rb in generated_random_bytes)
-            generated_random_bytes.append(bytes_value)
+            # [START get_random_bytes]
+            # get eight random bytes from a managed HSM
+            result = await client.get_random_bytes(count=8)
+            random_bytes = result.value
+            # [END get_random_bytes]
+            assert len(random_bytes) == 8
+            assert all([random_bytes != rb] for rb in generated_random_bytes)
```
chlowell, line 462: `all` isn't doing what you intend because you're giving it an expression that generates lists of bools:
```py
>>> all([s == 's'] for s in ['s'] * 2)
True
>>> all([s != 's'] for s in ['s'] * 2)
True
>>> list([s != 's'] for s in ['s'] * 2)
[[False], [False]]
```
(`[False]` is truthy because it's a non-empty list)

Easy fix--generate bools instead:
```suggestion
            assert all(random_bytes != rb for rb in generated_random_bytes)
```

chlowell (approved):

sdk/keyvault/azure-keyvault-keys/tests/test_keys_async.py
mccoyp: My mistake, thank you for catching this!

PR #20112, opened 2021-08-05
close in a reverse order of what we opened

xiangyan99:


seankane-msft (approved):
LGTM, but qq what is the scenario where this order matters? I understand best practice to close in reverse order but we won't lose references to the context managers so what is the difference?

xiangyan99:
> LGTM, but qq what is the scenario where this order matters? I understand best practice to close in reverse order but we won't lose references to the context managers so what is the difference?

I don't think it matters. Just to make it prettier. :)

PR #20136, opened 2021-08-06
[textanalytics] py27 sample encoding fix

kristapratico:
the service started returning a different output which had more surprise problematic characters for py2.7


deyaaeldeen (approved):
interesting

PR #20193, opened 2021-08-09
[packaging] Include license in PyPi packages

major:
The MIT license requires that a copy of the license is distributed with
the software. Add the license to the template and also add it to the
manifest file.

Fixes: azure/azure-sdk-for-python#20183

Signed-off-by: Major Hayden <major@mhtx.net>


major:
Trying this instead of fixing each package individually as mentioned by @lmazuel in #20183.

xiangyan99:
@scbedd could you help to take a look?

major:
Feel free to do whatever is needed! Thanks!

PR #20231, opened 2021-08-11
Avoid passing unexpected kwarg to EnvironmentCredential

chlowell:
Whoops, AzureApplicationCredential passes "managed_identity_client_id" to EnvironmentCredential. It works today but is something of a time bomb, better defused sooner than later.


mccoyp (approved):
Great catch :shipit:

PR #20233, opened 2021-08-11
Update to use timespan

rakshith91:


sdk/monitor/azure-monitor-query/azure/monitor/query/_helpers.py
```diff
@@ -49,26 +50,34 @@ def order_results(request_order, responses):
...
+    if not timespan:
+        return None
+    try:
+        start, end, duration = None, None, None
+        if isinstance(timespan[1], datetime):
+            start, end = timespan[0], timespan[1]
+        elif isinstance(timespan[1], timedelta):
+            start, duration = timespan[0], timespan[1]
+        else:
+            raise ValueError('Tuple must be a start datetime with a timedelta or an end datetime.')
```
xiangyan99, line 63: Could you add some comments to describe the situation? 
e.g. if isinstance(timespan[1], datetime): # if timespan is tuple[datetime, datetime], we treat it as [start time, end time].

sdk/monitor/azure-monitor-query/azure/monitor/query/_helpers.py
```diff
@@ -49,26 +50,34 @@ def order_results(request_order, responses):
...
+    except TypeError:
+        duration = timespan
     if duration is not None:
```
xiangyan99, line 66: We can just use if duration: :)

sdk/monitor/azure-monitor-query/azure/monitor/query/_helpers.py
```diff
@@ -49,26 +50,34 @@ def order_results(request_order, responses):
...
     if start is not None:
         start = Serializer.serialize_iso(start)
-        if end and duration:
-            raise ValueError("start_time can only be provided with duration or end_time, but not both.")
         if end is not None:
             end = Serializer.serialize_iso(end)
             iso_str = start + '/' + end
         elif duration is not None:
             iso_str = start + '/' + duration
         else:
```
xiangyan99, line 79: In which case will enter here?

sdk/monitor/azure-monitor-query/azure/monitor/query/_helpers.py
```diff
@@ -49,26 +50,34 @@ def order_results(request_order, responses):
...
             raise ValueError("Start time must be provided along with duration or end time.")
-    elif end is not None:
-        if not duration:
-            raise ValueError("End time must be provided along with duration or start time.")
-        end = Serializer.serialize_iso(end)
-        iso_str = duration + '/' + end
     else:
```
xiangyan99, line 81: Also, in which case will enter here?

sdk/monitor/azure-monitor-query/azure/monitor/query/_helpers.py
rakshith91: premeptively raising  in case start is provided with something other than a timedelta or a datetime

xiangyan99: It looks to me here is the case start is not None while end and duration are None?

rakshith91: improved the message to warn about none explicitly

xiangyan99 (approved):

PR #20234, opened 2021-08-11
[rest] unify response headers behaviors across transports

### LINKED ISSUE #20076: [rest] investigate the behavior of tuples in response headers
What happens when the service returns 2 headers with the same name. See what other networks do, and decide whether we want to return a list / what we want to return here.

iscai-msft:
fixes #20076

**NOTE**: I didn't fix the old responses to have this behavior, because I'm not sure if fixing it would be considered breaking, and from what I understand, we're not super investing time in them. LMK if we should fix though!

special shout out to @annatisch for helping me navigate and extend the aiohttp library here!


sdk/core/azure-core/azure/core/rest/_aiohttp.py
```diff
@@ -23,14 +23,62 @@
...
+        super().__init__(ref)
+        self._ref = ref
+
+    def __iter__(self):
+        for key, groups in groupby(self._ref.__iter__(), lambda x: x[0]):
+            yield tuple([key, ", ".join(group[1] for group in groups)])
+
+    def __contains__(self, item):
+        assert isinstance(item, (list, tuple))
+        assert len(item) == 2
```
annatisch, line 46: We shouldn't be raising assert errors in `__contains__`. Anything that isn't our expected input should just return `False`.

sdk/core/azure-core/tests/testserver_tests/async_tests/test_rest_headers_async.py
```diff
@@ -7,9 +7,124 @@
...
+@pytest.mark.asyncio
+async def test_headers_response(get_response_headers):
+    h = await get_response_headers(HttpRequest("GET", "/headers/duplicate/numbers"))
+    assert "a" in h
+    assert "A" in h
+    assert "b" in h
+    assert "B" in h
+    assert "c" not in h
+    assert h["a"] == "123, 456"
+    assert h.get("a") == "123, 456"
```
annatisch, line 37: Can we test these two case-insensitive as well?
```python
assert h["A"] == "123, 456"
assert h.get("A") == "123, 456"
```

sdk/core/azure-core/tests/testserver_tests/async_tests/test_rest_headers_async.py
```diff
@@ -7,9 +7,124 @@
...
+@pytest.mark.asyncio
+async def test_headers_response_keys(get_response_headers):
+    h = await get_response_headers(HttpRequest("GET", "/headers/duplicate/numbers"))
+    # basically want to make sure this behaves like dict {"a": "123, 456", "b": "789"}
+    ref_dict = {"a": "123, 456", "b": "789"}
+    assert list(h.keys()) == list(ref_dict.keys())
+    assert repr(h.keys()) == repr(ref_dict.keys())
+    assert "a" in h.keys()
+    assert "b" in h.keys()
+    assert set(h.keys()) == set(ref_dict.keys())
```
annatisch, line 55: For testing keys/values/items, we can also test that mutability is preserved:
```python
before_mutation_keys = h.keys()
h['c'] = '000'
assert 'c' in before_mutation_keys
```

sdk/core/azure-core/azure/core/rest/_aiohttp.py
```diff
@@ -23,14 +23,62 @@
...
+    def items(self):
+        """Return a new view of the dictionary's keys."""
+        return _ItemsView(super().items())
+
+    def values(self):
+        """Return a new view of the dictionary's values."""
+        return {k: v for k, v in self.items()}.values()  # pylint: disable=unnecessary-comprehension
+
+    def __getitem__(self, key: str) -> str:
+        return ", ".join(self.getall(key))
```
annatisch, line 75: What does `getall` return if the key isn't found - `None` or empty list/tuple?

sdk/core/azure-core/azure/core/rest/_aiohttp.py
```diff
@@ -23,14 +23,62 @@
...
+
+    def get(self, key, default=None):
+        values = self.getall(key, default)
+        if values:
+            return ", ".join(values)
+        return values
```
annatisch, line 81: Same questions as above - is it possible we're returning an empty list/tuple here instead of None (where no default is used).

sdk/core/azure-core/tests/testserver_tests/async_tests/test_rest_headers_async.py
```diff
@@ -7,9 +7,124 @@
...
+async def test_headers_response(get_response_headers):
+    h = await get_response_headers(HttpRequest("GET", "/headers/duplicate/numbers"))
+    assert "a" in h
+    assert "A" in h
+    assert "b" in h
+    assert "B" in h
+    assert "c" not in h
+    assert h["a"] == "123, 456"
+    assert h.get("a") == "123, 456"
+    assert h.get("nope", default=None) is None
```
annatisch, line 38: Could we add a test for this that doesn't set `default`?

sdk/core/azure-core/azure/core/rest/_aiohttp.py
```diff
@@ -23,14 +23,62 @@
...
+        """Return a new view of the dictionary's keys."""
+        return {k: v for k, v in self.items()}.keys()  # pylint: disable=unnecessary-comprehension
+
+    def items(self):
+        """Return a new view of the dictionary's keys."""
+        return _ItemsView(super().items())
+
+    def values(self):
+        """Return a new view of the dictionary's values."""
+        return {k: v for k, v in self.items()}.values()  # pylint: disable=unnecessary-comprehension
```
annatisch, line 72: Does is need to be a comprehension? I think you can just do `dict(self.items()).values()`? (And can remove the pylint msg)

sdk/core/azure-core/azure/core/rest/_aiohttp.py
```diff
@@ -23,14 +23,62 @@
...
+
+    def __repr__(self):
+        return f"dict_items({list(self.__iter__())})"
+
+
+class _CIMultiDict(CIMultiDict):
+    """Dictionary with the support for duplicate case-insensitive keys."""
+
+    def __iter__(self):
+        return iter(self.keys())
```
annatisch, line 60: Not sure if I saw tests for this one? e.g. `assert list(headers) == [...]`

sdk/core/azure-core/azure/core/rest/_aiohttp.py
```diff
@@ -23,14 +23,62 @@
...
+
+    def keys(self):
+        """Return a new view of the dictionary's keys."""
+        return {k: v for k, v in self.items()}.keys()  # pylint: disable=unnecessary-comprehension
+
+    def items(self):
+        """Return a new view of the dictionary's keys."""
```
johanste, line 67:
```suggestion
        """Return a new view of the dictionary's items."""
```

sdk/core/azure-core/azure/core/rest/_aiohttp.py
johanste: ...and @annatisch's comment applies to `keys()` as well....

sdk/core/azure-core/azure/core/rest/_aiohttp.py
iscai-msft: good point, let me add that!

sdk/core/azure-core/azure/core/rest/_aiohttp.py
iscai-msft: thanks, will fix!

sdk/core/azure-core/azure/core/rest/_aiohttp.py
```diff
@@ -23,14 +23,62 @@
...
+
+    def __iter__(self):
+        for key, groups in groupby(self._ref.__iter__(), lambda x: x[0]):
+            yield tuple([key, ", ".join(group[1] for group in groups)])
+
+    def __contains__(self, item):
+        assert isinstance(item, (list, tuple))
+        assert len(item) == 2
+        for k, v in self.__iter__():
+            if item[0] == k and item[1] == v:
```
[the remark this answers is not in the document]
iscai-msft: yep that makes sense! I'l update it to check if the case insensitive key is in the items. For values we're not doing case-insensitive, so won't look for case insensitive values.

sdk/core/azure-core/azure/core/rest/_aiohttp.py
iscai-msft: yay like that a lot better, thanks!

sdk/core/azure-core/azure/core/rest/_aiohttp.py
iscai-msft: It raises an error if you don't pass in a default. I'll go ahead and fix this, thanks so much!

sdk/core/azure-core/azure/core/rest/_aiohttp.py
iscai-msft: yep, switched!

sdk/core/azure-core/tests/testserver_tests/async_tests/test_rest_headers_async.py
iscai-msft: yep!

sdk/core/azure-core/tests/testserver_tests/async_tests/test_rest_headers_async.py
iscai-msft: @annatisch I looked into this behavior, and httpx and aiohttp both don't natively have this behavior. Is this something we want to override from them?

sdk/core/azure-core/azure/core/rest/_aiohttp.py
```diff
@@ -23,14 +23,102 @@
...
+
+class _CIMultiDict(CIMultiDict):
+    """Dictionary with the support for duplicate case-insensitive keys."""
+
+    def __iter__(self):
+        return iter(self.keys())
+
+    def keys(self):
+        """Return a new view of the dictionary's keys."""
+        return _KeysView(self.items())
```
iscai-msft, line 101: couldn't just use dict comprehension here because that made the .keys() and .values() immutable

sdk/core/azure-core/tests/testserver_tests/async_tests/test_rest_headers_async.py
iscai-msft: fixed!

sdk/core/azure-core/tests/testserver_tests/test_rest_headers.py
```diff
@@ -4,154 +4,247 @@
...
+@pytest.mark.skipif(sys.version_info < (3, 0),
+                    reason="In 2.7, .values() are not mutable")
+def test_headers_response_values_mutability(get_response_headers):
+    h = get_response_headers(HttpRequest("GET", "/headers/duplicate/numbers"))
+    # test mutability
+    before_mutation_values = h.values()
+    h['c'] = '000'
+    assert '000' in before_mutation_values
+
+def test_headers_response_items(get_response_headers):
```
iscai-msft, line 111: in 2.7, `collections.KeysView`, `collections.ItemsView`, and `collections.ValuesView` are immutable, while in `collections.abc`, they are mutable:

2.7 code snippet 
Screen Shot 2021-08-23 at 12 09 02 PM

code snippet in 3.9:
Screen Shot 2021-08-23 at 12 09 50 PM

annatisch (approved):

PR #20250, opened 2021-08-12
Call connection wrapper changes

ravithanneeru:


sdk/communication/azure-communication-callingserver/azure/communication/callingserver/_generated/models/_azure_communication_calling_server_service_enums.py
```diff
@@ -91,3 +91,14 @@ class ToneValue(with_metaclass(_CaseInsensitiveEnumMeta, str, Enum)):
     C = "c"
     D = "d"
     FLASH = "flash"
+
```
zihzhan-msft, line 94: generated code should not be manually overrided.

PR #20290, opened 2021-08-16
[rest] change text from a property to a method

iscai-msft:


sdk/core/azure-core/azure/core/rest/_requests_basic.py
```diff
@@ -85,10 +85,11 @@ def encoding(self, value):
         self._encoding = value
         self._internal_response.encoding = value
 
-    @property
-    def text(self):
+    def text(self, encoding=None):
         # this will trigger errors if response is not read in
         self.content  # pylint: disable=pointless-statement
+        if encoding:
+            self._internal_response.encoding = encoding
```
xiangyan99, line 92: Should we update _internal_response.encoding in text method?

This will cause some side effect?

iscai-msft: @xiangyan99 talked to @lmazuel offline, and we agreed to not handle requests text and encoding as a special case, so all of this code is gone now!

sdk/core/azure-core/tests/testserver_tests/test_rest_http_response.py
```diff
@@ -294,5 +294,20 @@ def test_send_request_return_pipeline_response(client):
...
+
+def test_text_and_encoding(send_request):
+    response = send_request(
+        request=HttpRequest("GET", "/encoding/emoji"),
+    )
+    assert response.content == u"👩".encode("utf-8")
+    assert response.text() == u"👩"
+
+    # try setting encoding as a property
+    response.encoding = "utf-16"
```
iscai-msft, line 308: this decodes it with `utf-16`

sdk/core/azure-core/tests/testserver_tests/test_rest_http_response.py
```diff
@@ -294,5 +294,20 @@ def test_send_request_return_pipeline_response(client):
...
-    assert response.http_response.text == "Hello, world!"
+    assert response.http_response.text() == "Hello, world!"
     assert hasattr(response.http_request, "content")
+
+def test_text_and_encoding(send_request):
+    response = send_request(
+        request=HttpRequest("GET", "/encoding/emoji"),
+    )
+    assert response.content == u"👩".encode("utf-8")
+    assert response.text() == u"👩"
```
iscai-msft, line 305: we start off with the default `utf-8-sig`

sdk/core/azure-core/tests/testserver_tests/test_rest_http_response.py
```diff
@@ -294,5 +294,20 @@ def test_send_request_return_pipeline_response(client):
...
+
+    # try setting encoding as a property
+    response.encoding = "utf-16"
+    assert response.text() == u"鿰ꦑ" == response.content.decode(response.encoding)
+
+    # assert latin-1 changes text decoding without changing encoding property
+    assert response.text("latin-1") == 'ð\x9f\x91©' == response.content.decode("latin-1")
```
iscai-msft, line 312: this decodes it with `latin-1`

sdk/core/azure-core/azure/core/rest/_rest.py
```diff
@@ -235,17 +235,23 @@ def encoding(self, value):
...
-        if self._text is None:
+    def text(self, encoding=None):
+        # type: (Optional[str]) -> str
+        """Returns the response body as a string
+
+        :param optional[str] encoding: The encoding you want to decode the text with. Can
+         also be set independently through our encoding property
+        :return: The response's content decoded as a string.
+        """
+        if self._text is None or encoding:
```
xiangyan99, line 248: It looks to me if text() is called once. Changing encoding will not take affect any more.

e.g. after called text(encoding="utf-8"), if I call text(encoding="ascii"), can I get the "ascii" decoded string?

Or is this by design?

xiangyan99: Never mind. :)

xiangyan99 (approved):
LGTM

sdk/core/azure-core/azure/core/rest/_rest_py3.py
```diff
@@ -262,16 +262,22 @@ def encoding(self) -> Optional[str]:
...
+        :return: The response's content decoded as a string.
+        """
+        if self._text is None or encoding:
             content = self.content
             if not content:
                 self._text = ""
             else:
-                self._text = decode_to_text(self.encoding, self.content)
+                encoding_to_pass = encoding or self.encoding
+                self._text = decode_to_text(encoding_to_pass, self.content)
```
johanste, line 280: Why use `self.content` when you fished out the value on line 275. Alternatively, why are you fishing out content on line 275? Do we want to save the instructions needed to access the attribute from the instance?

johanste: Is it because accessing the content property has side-effects?

iscai-msft: Acccessign the `content` property will raise the `ResponseNotRead` errors if users mishandle. You ahve a good point, I don't have to store it in another variable, so I'll just stick to using `self.content`

johanste: I agree with pylint that the side effects of accessing `self.content` are too subtle. You are going to access self.content anyways, why do you need it *also* call it here?

iscai-msft: sorry i'm not quite understanding your comment. Does the current code there look better to you?

johanste: Why not just say:

```python
if not self.content:  # <- this will raise if data is not read. Line 275 is not needed. 
    self._text = ""
else:
    ...
```

iscai-msft: ah i think i was confused bc we're looking at diff code versions. What I currently have is
```python
if self._text is None or encoding:
    encoding_to_pass = encoding or self.encoding
    self._text = decode_to_text(encoding_to_pass, self.content)
return self._text
```
which i think solves the issue you brought up (i moved the empty content detection to `decode_to_text`

Issue #19925, opened 2021-07-23
HttpLoggingPolicy supports integrating logs into one record


freeloop914:
I'm using azure function app and using SDK to operate Azure resources like blob list/download.
I need these HTTP request/response logs for troubleshooting debug.
But currently they are separated to multiple records.
It's very difficult to query.

For example, this is what current log looks like:
```
[7/29/2020 6:55:16 PM] Request URL: 'https://xxx.documents.azure.com:443/'
[7/29/2020 6:55:16 PM] Request method: 'GET'
[7/29/2020 6:55:16 PM] Request headers:
[7/29/2020 6:55:16 PM]     'Cache-Control': 'no-cache'
[7/29/2020 6:55:16 PM]     'x-ms-version': 'REDACTED'
[7/29/2020 6:55:16 PM]     'x-ms-documentdb-query-iscontinuationexpected': 'REDACTED'
[7/29/2020 6:55:16 PM]     'x-ms-consistency-level': 'REDACTED'
[7/29/2020 6:55:16 PM]     'x-ms-session-token': 'REDACTED'
[7/29/2020 6:55:16 PM]     'x-ms-date': 'REDACTED'
[7/29/2020 6:55:16 PM]     'authorization': 'REDACTED'
[7/29/2020 6:55:16 PM]     'Accept': 'application/json'
[7/29/2020 6:55:16 PM]     'Content-Length': '0'
[7/29/2020 6:55:16 PM]     'User-Agent': 'azsdk-python-cosmos/4.0.0 Python/3.8.5 (Windows-10-10.0.19041-SP0)'

```

What I expect is :

```
[7/29/2020 6:55:16 PM] Request URL: 'https://xxx.documents.azure.com:443/'
 Request method: 'GET'
 Request headers:
     'Cache-Control': 'no-cache'
     'x-ms-version': 'REDACTED'
     'x-ms-documentdb-query-iscontinuationexpected': 'REDACTED'
     'x-ms-consistency-level': 'REDACTED'
     'x-ms-session-token': 'REDACTED'
     'x-ms-date': 'REDACTED'
     'authorization': 'REDACTED'
     'Accept': 'application/json'
     'Content-Length': '0'
     'User-Agent': 'azsdk-python-cosmos/4.0.0 Python/3.8.5 (Windows-10-10.0.19041-SP0)'

```

It looks more clean and tidy ( At least from Azure function app Monitoring Log interface. )
and when I query the log by 

[...]

PR #20300, opened 2021-08-16
integrate logs

xiangyan99:
#19925


sdk/core/azure-core/azure/core/pipeline/policies/_universal.py
```diff
@@ -420,26 +477,52 @@ def on_request(self, request):
...
+            if multi_record:
+                logger.info("Request URL: %r", redacted_url)
+                logger.info("Request method: %r", http_request.method)
+                logger.info("Request headers:")
+                for header, value in http_request.headers.items():
+                    value = self._redact_header(header, value)
+                    logger.info("    %r: %r", header, value)
+                if isinstance(http_request.body, types.GeneratorType):
+                    logger.info("File upload")
+                    return
```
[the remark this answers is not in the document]
johanste: Sorry, read the code wrong (missed the `return` instead of `else` on line 451)

sdk/core/azure-core/azure/core/pipeline/policies/_universal.py
```diff
@@ -280,27 +281,31 @@ def on_request(self, request):
                 return
 
             try:
-                _LOGGER.debug("Request URL: %r", http_request.url)
-                _LOGGER.debug("Request method: %r", http_request.method)
-                _LOGGER.debug("Request headers:")
+                log_string = "Request URL: '{}'".format(http_request.url)
```
tjprescott, line 284: Why are we not looking at the multi record environment variable here but we do elsewhere?

tjprescott (commented):
One comment but otherwise LGTM.

sdk/core/azure-core/azure/core/pipeline/policies/_universal.py
xiangyan99: Because it is logger.info in HttpLoggingPolicy but _LOGGER.debug here.

You know, we are changing the log format which is technically a breaking change.

For logger.info, we added multi_record to allow user to opt in the old behavior.

https://github.com/Azure/azure-sdk-for-python/pull/20300#discussion_r697737641 for more information.
But for debug log, there is not need for that.

tjprescott (approved):

sdk/core/azure-core/azure/core/pipeline/policies/_universal.py
```diff
@@ -280,27 +281,31 @@ def on_request(self, request):
...
+                log_string += "/nRequest method: '{}'".format(http_request.method)
```
jiasli, line 285: Shouldn't `\n` be used? `/n` makes the output unreadable:

image

sdk/core/azure-core/tests/test_http_logging_policy.py
```diff
@@ -280,12 +296,14 @@ def emit(self, record):
     policy.on_response(request, response)
 
     assert all(m.levelname == 'INFO' for m in mock_handler.messages)
```
indera, line 298: Usage of `INFO` here seems inconsistent with the other code... 

Example: sdk/keyvault/azure-keyvault-certificates/tests/test_certificates_client.py

PR #20043, opened 2021-08-02
OnBehalfOfCredential

chlowell:
Closes #19308. API here is still missing a property to tell authentication policies not to cache this credential's tokens.

PR #20313, opened 2021-08-17
Add AccessToken.refresh_on

chlowell:
This adds a refresh_on property to AccessToken, allowing credentials a means to indicate when a token should be refreshed i.e. replaced. A mechanism like this is required by OnBehalfOfCredential (#20043), whose tokens should never be cached by callers, and has other potential uses. The default refresh_on value aligns with BearerTokenCredentialPolicy's current refresh behavior.

AccessToken is a namedtuple, making it tricky to extend without breaking code. My solution here is to override `__new__` to return the expected tuple, and use `__init__` to initialize the new property. The resulting object can be constructed like the former tuple and behaves the same when used like a tuple but has the new property.


chlowell:
Closing in favor of #20451

PR #20314, opened 2021-08-17
[ACR]Bug fix for lacking of 'https://' prefix in endpoint in preparer

YalinLi0312:


sdk/containerregistry/azure-containerregistry/tests/recordings/test_anon_access.test_delete_manifest.yaml
```diff
@@ -46,7 +46,7 @@ interactions:
       code: 401
       message: Unauthorized
 - request:
-    body: service=seankaneanon.azurecr.io&scope=repository%3Alibrary%2Fhello-world%3Ametadata_read&grant_type=password
+    body: service=yalinlitestsanon.azurecr.io&scope=repository%3Alibrary%2Fhello-world%3Ametadata_read&grant_type=password
```
seankane-msft, line 49: Nit: The account name shows up in the service of the request body here. This should probably be scrubbed and replaced with `fake_url` by the preparer. It's not breaking because the anon are read only but if the non-readonly are in recordings those probably should be scrubbed out

seankane-msft (approved):

mccoyp (approved):

PR #20335, opened 2021-08-18
[purview account] first release for azure-purview-account

msyyc:
`autorest --version=3.4.5 --python --track2 --use=D:\dev3\autorest.python --use=@autorest\modelerfour@4.19.2 --python-sdks-folder=D:\dev3\azure-sdk-for-python\sdk --python-mode=update D:\dev3\azure-sdk-for-python\sdk\purview\azure-purview-account\swagger\README.md --version-tolerant`


sdk/purview/azure-purview-account/README.md
```diff
@@ -0,0 +1,164 @@
...
+
+credential = DefaultAzureCredential()
+client = PurviewAccountClient(endpoint="https://<my-account-name>.purview.azure.com", credential=credential)
+
+request = accounts.build_list_keys_request()
+
+response = client.send_request(request)
+try:
+    response.raise_for_status()
+    assert response.status_code == 200
```
iscai-msft, line 83: I think this sample is out of date because we've embedded the requests. Can you switch this to calling `client.list_keys()`? Additionally, can you iterate over the returned keys and print out information on them? thanks!

iscai-msft: p.s., we also generally don't include `assert`s in our sample code, we use `assert` for testing instead

sdk/purview/azure-purview-account/README.md
```diff
@@ -0,0 +1,164 @@
...
+
+# This client will log detailed information about its HTTP sessions, at DEBUG level
+client = PurviewAccountClient(endpoint=endpoint, credential=credential, logging_enable=True)
+```
+
+Similarly, `logging_enable` can enable detailed logging for a single `send_request` call,
+even when it isn't enabled for the client:
+
+```python
+result = client.send_request(request, logging_enable=True)
```
iscai-msft, line 130: I think it would be better to switch this call to an operation call, for example

```python
result = client.list_keys(..., logging_enable=True)
```

sdk/purview/azure-purview-account/tests/test_smoke.py
```diff
@@ -0,0 +1,26 @@
...
+        response = client.accounts.get_access_keys()
+        assert set(response.keys()) == set(['atlasKafkaPrimaryEndpoint', 'atlasKafkaSecondaryEndpoint'])
+
+    @PurviewAccountPowerShellPreparer()
+    def test_collections_list(self, purviewaccount_endpoint):
+        self.recording_processors.append(PurviewAccountCollectionsRecordingProcessor())
+        client = self.create_client(endpoint=purviewaccount_endpoint)
+        response = client.collections.list_collections()
+        result = [item for item in response]
+        assert len(result) > 0
```
iscai-msft, line 26: I'd also verify some of the attributes in the key response if possible, for example verify that their names are what you expect etc

iscai-msft: side note: i refer to "key names", not 100% sure that's actually whats in the response body, but what I mean is just identifying information about each key

iscai-msft (commented):
almost there, jsut some docstring / test nits. Thanks!

sdk/purview/azure-purview-account/tests/test_smoke.py
msyyc: I think detailed check is not necessary since the resource is created manually in advance. And the resource may be changed in the future. But we could check whether keys exist.

sdk/purview/azure-purview-account/tests/test_smoke_async.py
```diff
@@ -0,0 +1,28 @@
...
+from testcase_async import PurviewAccountTestAsync
+from _util import PurviewAccountRecordingProcessor, PurviewAccountCollectionsRecordingProcessor
+
+
+class PurviewAccountSmokeTestAsync(PurviewAccountTestAsync):
+
+    @PurviewAccountPowerShellPreparer()
+    async def test_basic_smoke_test(self, purviewaccount_endpoint):
+        self.recording_processors.append(PurviewAccountRecordingProcessor())
+        client = self.create_async_client(endpoint=purviewaccount_endpoint)
```
iscai-msft, line 17: like the tests!

iscai-msft (approved):
the documentation could be improved, but I don't want to block the release on this

PR #20361, opened 2021-08-19
[rest] use azure json encoder for json input bodies

iscai-msft:
shoutout to @mccoyp for all his hard work, it has essentially made this PR a one-liner


sdk/core/azure-core/azure/core/rest/_helpers.py
```diff
@@ -182,7 +183,7 @@ def set_content_body(content):
 
 def set_json_body(json):
     # type: (Any) -> Tuple[Dict[str, str], Any]
-    body = dumps(json)
+    body = dumps(json, cls=AzureJSONEncoder)
```
xiangyan99, line 186: Can we remove 
```py
from json import dumps
```
?

xiangyan99 (approved):

PR #20379, opened 2021-08-20
[Key Vault] Get default API version from sync client base

mccoyp:
Rather than redefining `DEFAULT_VERSION` in `async_client_base.py`, it's less error-prone to follow suit with `azure-keyvault-administration` and import the value from `client_base.py`.


chlowell:
Do you want to add 7.3-preview code to certs and secrets before doing this? Then you could have the same `_shared` in all 3 libraries.

mccoyp:
Sure, that works -- there's no rush.

YalinLi0312 (approved):
LGTM

PR #20380, opened 2021-08-20
Chain exceptions from LibsecretPersistence

chlowell:
Currently we expect only ImportError, but constructing LibsecretPersistence could raise ValueError instead. The type of the exception doesn't matter for our purposes, so this PR broadens the except clause to handle all exceptions the same way: if the application allows unencrypted storage, fall back to that; if not, raise a ValueError from the LibsecretPersistence error, to preserve the msal-extensions error message explaining what went wrong.

### DIFF
--- a/sdk/identity/azure-identity/azure/identity/_persistent_cache.py
+++ b/sdk/identity/azure-identity/azure/identity/_persistent_cache.py
@@ -92,11 +92,13 @@ def _get_persistence(allow_unencrypted, account_name, cache_name):
                 file_path, cache_name, {"MsalClientID": "Microsoft.Developer.IdentityService"}, label=account_name
             )
         except Exception as ex:  # pylint:disable=broad-except
+            _LOGGER.debug('msal-extensions is unable to encrypt a persistent cache: "%s"', ex, exc_info=True)
             if not allow_unencrypted:
-                _LOGGER.debug('msal-extensions is unable to encrypt a persistent cache: "%s"', ex, exc_info=True)
                 error = ValueError(
-                    "Persistent cache encryption isn't available in this environment. Please install encryption "
-                    + 'dependencies or specify "allow_unencrypted_storage=True" to store the cache without encryption.'
+                    "Cache encryption is impossible because libsecret dependencies are not installed or are unusable,"
+                    + " for example because no display is available (as in an SSH session). The chained exception has"
+                    + ' more information. Specify "allow_unencrypted_storage=True" to store the cache unencrypted'
+                    + " instead of raising this exception."
                 )
                 six.raise_from(error, ex)
             return msal_extensions.FilePersistence(file_path)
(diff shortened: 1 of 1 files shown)


sdk/identity/azure-identity/azure/identity/_persistent_cache.py
```diff
@@ -86,12 +88,13 @@ def _get_persistence(allow_unencrypted, account_name, cache_name):
             return msal_extensions.LibsecretPersistence(
                 file_path, cache_name, {"MsalClientID": "Microsoft.Developer.IdentityService"}, label=account_name
             )
-        except ImportError:
+        except Exception as ex:  # pylint:disable=broad-except
             if not allow_unencrypted:
-                raise ValueError(
-                    "PyGObject is required to encrypt the persistent cache. Please install that library or "
-                    + 'specify "allow_unencrypted_storage=True" to store the cache without encryption.'
+                error = ValueError(
```
xiangyan99, line 93: Why don't we explicitly tell users we need PyGObject to work?

chlowell: To avoid having to keep this message up to date with msal-extensions. It may stop requiring PyGObject, or require a different version of it (it's a runtime dependency). The details of the error aren't lost; msal-extensions includes them in the exception we're raising from. Having written this I'm thinking our message should be even less specific, because this one is misleading when we can't encrypt because we're in an SSH session.

xiangyan99 (approved):

sdk/identity/azure-identity/azure/identity/_persistent_cache.py
xiangyan99: Could you give me more information that other than ImportError, which error(s) we want to cover here?

chlowell: The next version of msal-extensions may raise ValueError or RuntimeError as well as ImportError (code is [here](https://github.com/AzureAD/microsoft-authentication-extensions-for-python/blob/dev/msal_extensions/libsecret.py)). It might also be possible for us to see an unexpected (by msal-extensions) PyGObject exception. Handling all exceptions here appeals to me because I don't want to rely too much on the current behavior of msal-extensions, and for us the particular exception isn't important.

xiangyan99: I am fine with that. The only question I think would be do we want to give different error messages for different types of error so it can be more actionable for users to fix them?

chlowell: msal-extensions does that with its errors, so by chaining them we give users the details without depending on how msal-extensions handles libsecret/PyGObject errors.

xiangyan99: LGTM

mccoyp (approved):

Issue #20301, opened 2021-08-16
Character '"' is not valid at position 20 in 'PartitionKey='"A'aaa"_bbbb3',RowKey='"A'aaa"_bbbb3''.


646826:
- **Package Name**: azure-data-tables
- **Package Version**: 12.1.0
- **Operating System**: Linux and Windows
- **Python Version**: 3.8

**Describe the bug**
```
TableTransactionError: 0:Syntax error: character '"' is not valid at position 20 in 'PartitionKey='"A'aaa"_bbbb3',RowKey='"A'aaa"_bbbb3''.
RequestId:8452bf63-8002-0099-18eb-9265f1000000
Time:2021-08-16T22:08:14.5833819Z
ErrorCode:InvalidInput
---------------------------------------------------------------------------
TableTransactionError                     Traceback (most recent call last)
/Untitled-1.py in 
     15 azure_table_client = TableServiceClient(endpoint=f"https://{account_name}.table.core.windows.net", credential=credential)
     16 azure_table_service = azure_table_client .create_table_if_not_exists(to_table_name)
---> 17 azure_table_service .submit_transaction(batch)

/usr/local/lib/python3.8/site-packages/azure/data/tables/_table_client.py in submit_transaction(self, operations, **kwargs)
    729             except AttributeError:
    730                 raise ValueError("Unrecognized operation: {}".format(operation[0]))
--> 731         return self._batch_send(*batched_requests.requests, **kwargs)  # type: ignore

/usr/local/lib/python3.8/site-packages/azure/data/tables/_base_client.py in _batch_send(self, *reqs, **kwargs)
    301                     error_message="The transaction request was too large",
    302                     error_type=RequestTooLargeError)
--> 303             raise _decode_error(
    304                 response=error_parts[0],
    305                 error_type=TableTransactionError

TableTransactionError: 0:Syntax error: character '"' is not valid at position 20 in 'PartitionKey='"A'aaa"_bbbb3',RowKey='"A'aaa"_bbbb3''.
RequestId:8452bf63-8002-0099-18eb-9265f1000000
Time:2021-08-16T22:08:14.5833819Z
ErrorCode:InvalidInput
```

**To Reproduce**
`pip install azure-data-tables`
```
from azure.core.credentials import AzureNamedKeyCredential
from azure.data.tables import TableServiceClient
from azure.data.tables import TableEntity

account_name=""
account_key=""
to_table_name=""

batch = []
[...]

PR #20386, opened 2021-08-23
[Tables] Fix single quote char handling in keys

annatisch:
Fixes bug reported in #20301

### DIFF
--- a/sdk/tables/azure-data-tables/tests/test_table_entity.py
+++ b/sdk/tables/azure-data-tables/tests/test_table_entity.py
@@ -1221,7 +1221,7 @@ class StorageTableEntityTest(AzureTestCase, TableTestCase):
 
             # Act
             sent_entity = self._create_updated_entity_dict(entity['PartitionKey'], entity['RowKey'])
-            resp = self.table.upsert_entity(mode=UpdateMode.MERGE, entity=sent_entity)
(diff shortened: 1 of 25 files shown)


sdk/tables/azure-data-tables/tests/test_table_entity.py
```diff
@@ -1221,7 +1221,7 @@ def test_operations_on_entity_with_partition_key_having_single_quote(self, table
 
             # Act
             sent_entity = self._create_updated_entity_dict(entity['PartitionKey'], entity['RowKey'])
-            resp = self.table.upsert_entity(mode=UpdateMode.MERGE, entity=sent_entity)
+            resp = self.table.upsert_entity(mode=UpdateMode.REPLACE, entity=sent_entity)
```
YalinLi0312, line 1224: Why change the UpdateMode here and some places below?

annatisch: Excellent question!
The original tests were always broken - and had been failing for some time, and we were skipping them because we _thought_ the reason they were failing was because this character wasn't supported by the service. In fact it was because we had a bug, and once that bug was fixed, the tests needed to be fixed to also run.

YalinLi0312 (approved):

PR #20387, opened 2021-08-23
Validate Application ID in User Agent

LuChen-Microsoft:
1. Application Id needs to be less or equal than 24 characters.
2. Spaces are not allowed in the application Id based on the docs.

### DIFF
--- a/sdk/core/azure-core/azure/core/pipeline/policies/_universal.py
+++ b/sdk/core/azure-core/azure/core/pipeline/policies/_universal.py
@@ -187,6 +187,8 @@ class UserAgentPolicy(SansIOHTTPPolicy):
     """
     _USERAGENT = "User-Agent"
     _ENV_ADDITIONAL_USER_AGENT = 'AZURE_HTTP_USER_AGENT'
+    _MAX_APPLICATION_ID_LENGTH = 24
+    _INVALID_APPLICATION_ID_LENGTH = "'applicationId' length cannot be greater than " + _MAX_APPLICATION_ID_LENGTH
 
     def __init__(self, base_user_agent=None, **kwargs):  # pylint: disable=super-init-not-called
         # type: (Optional[str], **Any) -> None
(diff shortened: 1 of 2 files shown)


sdk/core/azure-core/azure/core/pipeline/policies/_universal.py
```diff
@@ -205,6 +206,11 @@ def __init__(self, base_user_agent=None, **kwargs):  # pylint: disable=super-ini
             )
 
         if application_id:
+            if application_id.length > self._MAX_APPLICATION_ID_LENGTH:
+                raise ValueError("'applicationId' length cannot be greater than " + self._MAX_APPLICATION_ID_LENGTH)
```
xiangyan99, line 210: If the app_id is too long, I prefer truncating the string rather than failing the app. 

Your opinions? @johanste  @annatisch

sdk/core/azure-core/azure/core/pipeline/policies/_universal.py
```diff
@@ -205,6 +206,11 @@ def __init__(self, base_user_agent=None, **kwargs):  # pylint: disable=super-ini
...
+            if ' ' in application_id:
+                raise ValueError("'applicationId' cannot contain spaces.")
+
```
tjprescott, line 213: I would recommend combining these so that the user gets one error message. Otherwise, if the application ID is more than 24 characters AND has spaces, they will get runtime errors twice, once for each error. See:
https://github.com/Azure/azure-sdk-for-ios/pull/1000

sdk/core/azure-core/azure/core/pipeline/policies/_universal.py
LuChen-Microsoft: It throws error in C# and java. Here are the similar validation 
https://github.com/Azure/azure-sdk-for-java/blob/main/sdk/core/azure-core/src/main/java/com/azure/core/util/ClientOptions.java#L57
https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/src/DiagnosticsOptions.cs#L109

tjprescott:
This should have a changelog entry, and it must be pointed out that making this change would be a breaking change for AzureCore.

PR #20389, opened 2021-08-23
[EventHubs] Stress test refactor

yunhaoling:
~~currently blocked by:~~

~~> helm install -n python-eventhubs-stress-test-ns python-eventhubs-stress-test .
Error: unable to build kubernetes objects from release manifest: error validating "": error validating data: [ValidationError(Job): unknown field "Error" in io.k8s.api.batch.v1.Job, ValidationError(Job.spec.template.spec): missing required field "containers" in io.k8s.api.core.v1.PodSpec]~~

-- resolved by changing indent to 6


sdk/eventhub/azure-eventhub/stress/templates/testjob.yml
```diff
@@ -0,0 +1,16 @@
...
+{{- define "stress.python-eh-stress" -}}
+metadata:
+  labels:
+    testName: "deploy-python-eh-stress"
+    testInstance: "deploy-python-eh-stress-instance"
+    chaos: "true"
+spec:
+  containers:
+    - name: python-eh-stress
+      image: stresstestregistry.azurecr.io/yuling/py-eh-stress:v2
```
benbp, line 11: The labels/image should be updated to be guaranteed unique for our auto-release pipelines (no `testInstance` targeting overlaps with chaos, and a dynamic image name that matches what gets built by the scripts in the pipeline).

```
  labels:
    testName: "deploy-python-eh-stress"
    testInstance: "eventhub-{{ .Release.Name }}-{{ .Release.Revision }}"
    chaos: "true"
spec:
  containers:
    - name: python-eh-stress
      image: {{ default "stresstestregistry.azurecr.io/python" .Values.repository }}/stress:{{ default "v2" .Values.tag }}
```

sdk/eventhub/azure-eventhub/stress/templates/testjob.yml
```diff
@@ -0,0 +1,16 @@
...
+      {{- include "stress-test-addons.container-env" . | nindent 6 }}
+      command: ['bash', '-c', 'python3 azure_eventhub_producer_stress.py & python3 azure_eventhub_consumer_stress_sync.py']
```
benbp, line 13: I would swap these two lines, to keep template includes separate from user editable fields.

```suggestion
      command: ['bash', '-c', 'python3 azure_eventhub_producer_stress.py & python3 azure_eventhub_consumer_stress_sync.py']
      {{- include "stress-test-addons.container-env" . | nindent 6 }}
```

sdk/eventhub/azure-eventhub/stress/test-resources.json
```diff
@@ -0,0 +1,169 @@
+{
```
benbp, line 1: You can if you want, but technically this doesn't need to be checked in, as the scripts will build the bicep file on the fly. I had it checked in for service bus because it pre-dated this functionality.

sdk/eventhub/azure-eventhub/stress/templates/network_loss.yaml
```diff
@@ -0,0 +1,31 @@
...
+    cron: '@every 30s'
+  duration: '10s'
+  action: loss
+  direction: to
+  externalTargets:
+    - 'eh-stress-{{ .Release.Name }}-{{ .Release.Revision}}.servicebus.windows.net'
+  mode: one
+  selector:
+    labelSelectors:
+      testInstance: 'deploy-python-eh-stress-instance'
```
benbp, line 23:
```suggestion
    testInstance: "eventhub-{{ .Release.Name }}-{{ .Release.Revision }}"
```

sdk/eventhub/azure-eventhub/stress/templates/testjob.yml
benbp: The `testInstance field can be whatever you want, but it has to include both `.Release.Name` and `.Release.Revision`.

benbp (approved):

sdk/eventhub/azure-eventhub/stress/templates/network_loss.yaml
```diff
@@ -0,0 +1,31 @@
...
+    release: '{{ $.Release.Name }}'
+    revision: '{{ $.Release.Revision }}'
+spec:
+  scheduler:
+    cron: '@every 30s'
+  duration: '10s'
+  action: loss
+  direction: to
+  externalTargets:
+    - 'eh-stress-{{ .Release.Name }}-{{ .Release.Revision}}.servicebus.windows.net'
```
yunhaoling, line 19: question to Ben: where is this `eh-stress-{{ .Release.Name }}-{{ .Release.Revision}}` coming from?

it is automatically generated resource? how is the name being generated?

sdk/eventhub/azure-eventhub/stress/templates/testjob.yml
```diff
@@ -0,0 +1,16 @@
+{{- include "stress-test-addons.deploy-job-template.from-pod" (list . "stress.python-eh-stress") -}}
+{{- define "stress.python-eh-stress" -}}
+metadata:
+  labels:
+    testName: "deploy-python-eh-stress"
+    testInstance: "eventhub-{{ .Release.Name }}-{{ .Release.Revision }}"
```
yunhaoling, line 6: where are `.Release.Name` and `.Release.Revision` coming from? helm chart?

sdk/eventhub/azure-eventhub/stress/test-resources.json
yunhaoling: I just copied and pasted our EH test-resrouces.json, we don't have a bicep in our repo yet.

I'll still unclear on what's difference between bicep/json.

What I'm understanding is that they are almost the same thing defining the resources to be created, while bicep is more powerful (with nice syntax sugar?)

Can we say that ultimately we will just have bicep files, all json files will be gone?

benbp: Correct. Bicep compiles to and de-compiles from ARM templates. Beyond syntactic sugar it also has some useful features like module imports and good error messages.

We are moving towards a goal state of having all json files be gone. If you want to start by copying over test-resources.json that's fine, though you may have to remove some template parameters that won't be provided today by the script. @ckairen is working on migrating deployment to use the common New-TestResources.ps1 script though which would fix this.

sdk/eventhub/azure-eventhub/stress/templates/network_loss.yaml
benbp: This is the URI that the stress test itself will try to hit. It may be simpler to use `servicebus.windows.net` instead of trying to enumerate the actual URIs. I believe that for the JS tests this is generated from the resource group name, which itself is generated using these helm release parameters (https://github.com/Azure/azure-sdk-for-js/blob/main/sdk/servicebus/service-bus/test/stress/test-resources.json#L66).

sdk/eventhub/azure-eventhub/stress/templates/testjob.yml
benbp: Correct, the script runs a command like `helm install <release name> <chart path>`. The revision gets incremented if there is already a release with that name deployed into the cluster. This scheme ensures that there is always unique naming for resources and labels, to ensure things don't overlap and chaos isn't mis-targeted.

sdk/eventhub/azure-eventhub/stress/templates/network_loss.yaml
yunhaoling: yea.. I was confused because I don't see how `eh-stress-{{ .Release.Name }}-{{ .Release.Revision}}` is constructed from the file `test-resources.json` ..

It's indeed easier if we could just use the domain name `servicebus.windows.net` as the target

benbp: I think you can but we may need to test it (e.g. whether it needs to be `.servicebus.windows.net` or `*.servicebus.windows.net` or just `servicebus.windows.net`.

sdk/eventhub/azure-eventhub/stress/templates/testjob.yml
benbp: `version`, `appVersion` and `apiVersion` are helm specific fields that we don't really need but have to supply. `name` is also helm-specific, but in the `deploy-stress-tests.ps1` script we use that value to determine the name of the helm release.

So `Chart.yaml -> name` == `Release.Name` and if you run the script multiple times, then `Release.Revision` will increment.

Issue #20226, opened 2021-08-11
[translation] fix poller.details.id


kristapratico:
Our custom poller has two properties added to it:
![image](https://user-images.githubusercontent.com/31998003/129062685-fa6c4d81-6bd3-4d8c-a146-3f26893a2add.png)

We should ensure that along with poller.id, poller.details.id gets populated immediately with the ID from the initial response.

PR #20392, opened 2021-08-24
[translation] fix poller.details

kristapratico:
Resolves https://github.com/Azure/azure-sdk-for-python/issues/20226

Noticed this in our user study. `poller.details` represents the status metadata (TranslationStatus) returned from the GET. Depending on when a user accesses this property, that information might not be returned yet (did the POST, about to do the GET). We do, however, know the operation ID from the initial POST so we can populate that property immediately. 

Also fixed an incorrect type in the docstrings for async client.

### DIFF
--- a/sdk/translation/azure-ai-translation-document/azure/ai/translation/document/_polling.py
+++ b/sdk/translation/azure-ai-translation-document/azure/ai/translation/document/_polling.py
@@ -67,8 +67,7 @@ class DocumentTranslationLROPoller(LROPoller):
             return TranslationStatus._from_generated(  # pylint: disable=protected-access
                 self._polling_method._current_body  # pylint: disable=protected-access
             )
-        else:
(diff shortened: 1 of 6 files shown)


sdk/translation/azure-ai-translation-document/azure/ai/translation/document/_polling.py
```diff
@@ -63,9 +63,11 @@ def details(self):
 
         :rtype: ~azure.ai.translation.document.TranslationStatus
         """
-        return TranslationStatus._from_generated(  # pylint: disable=protected-access
-            self._polling_method._current_body  # pylint: disable=protected-access
-        )
+        if self._polling_method._current_body:  # pylint: disable=protected-access
```
iscai-msft, line 66: nit: include as bug fix in changelog

iscai-msft (approved):

PR #20416, opened 2021-08-25
[Key Vault] Add support for key rotation

### LINKED ISSUE #19840: [Key Vault] Add support for Key Rotation
To support KeyVault Key Rotation, implement the following changes (modeled from JS, but use whatever API is idiomatic for your language):

## ARM Template changes

When creating a keyvault, add `rotate` to the keys' accessPolicies so that your test application has permissions to rotate and set the rotation policy.

## Key Client API

```tsx
getKeyRotationPolicy(name: string, options?: GetKeyRotationPolicyOptions): Promise<KeyRotationPolicy | undefined>; // see (1) below
rotateKey(name: string, options?: RotateKeyOptions): Promise<KeyVaultKey>;
updateKeyRotationPolicy(name: string, policy: KeyRotationPolicyProperties, options?: UpdateKeyRotationPolicyOptions): Promise<KeyRotationPolicy>;
```

Note that policy behavior is against the latest version of a key.

## New models

```tsx
// @public
export interface KeyRotationLifetimeAction {
    action: KeyRotationPolicyAction;
    timeAfterCreate?: string; // see (2) below
    timeBeforeExpiry?: string;  // see (2) below
}

// @public
export interface KeyRotationPolicy extends KeyRotationPolicyProperties {
    readonly createdOn: Date;
    readonly id: string;
    readonly updatedOn?: Date;
}

// @public
export type KeyRotationPolicyAction = "Rotate" | "Notify";

// @public
export interface KeyRotationPolicyProperties { // see (3) below
    expiresIn?: string;  // see (2) below
    lifetimeActions?: KeyRotationLifetimeAction[];
}
```

## Swagger inconsistencies to be mindful of

The swagger defines the possible values for LifetimeActionsType as an enum consisting of `rotate` and `notify` but in reality it returns `Rotate` and `Notify`  - the service accepts both casing, is case-insensitive, but in order to model the correct values here we went with the following swagger transformation:

```yaml
directive:
  - from: swagger-document
    where: $.definitions.LifetimeActionsType.properties["type"]["x-ms-enum"]
    transform: >
      $.values[0].value = "Rotate";
      $.values[1].value = "Notify";
```

## Notes

1. Calling GET /keys/{name}/rotationpolicy will return an object with all properties set to null. We're discussing having that endpoint 404, but until then I am return undefined if the id doesn't exist
2. We're still discussing whether the swagger should model this as an ISO8601 duration. In JS, the architects prefer a string, but model this duration in whatever way is idiomatic (TimeSpan, etc). If the swagger changes, I'll update the feature crew
3. In order to properly model input / output values I separated out the KeyRotationPolicy and the KeyRotationPolicyProperties . The former extends the latter and adds the server-side attributes (timestamps and ID).

mccoyp:
Resolves #19840.

### DIFF
--- a/sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_client.py
+++ b/sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_client.py
@@ -8,7 +8,7 @@ from azure.core.tracing.decorator import distributed_trace
 from ._shared import KeyVaultClientBase
 from ._shared.exceptions import error_map as _error_map
 from ._shared._polling import DeleteRecoverPollingMethod, KeyVaultOperationPoller
-from ._models import DeletedKey, KeyVaultKey, KeyProperties, KeyRotationLifetimeAction, KeyRotationPolicy, RandomBytes, ReleaseKeyResult
+from ._models import DeletedKey, KeyVaultKey, KeyProperties, KeyRotationPolicy, RandomBytes, ReleaseKeyResult
 
 try:
     from typing import TYPE_CHECKING
@@ -729,7 +729,7 @@ class KeyClient(KeyVaultClientBase):
         return RandomBytes(value=result.value)
 
     @distributed_trace
-    async def get_key_rotation_policy(self, name, **kwargs):
+    def get_key_rotation_policy(self, name, **kwargs):
         # type: (str, **Any) -> Optional[KeyRotationPolicy]
         """Get the rotation policy of a Key Vault key.
 
@@ -781,7 +781,7 @@ class KeyClient(KeyVaultClientBase):
         if lifetime_actions:
             lifetime_actions = [
                 self._models.LifetimeActions(
-                    action=action.action,
+                    action=self._models.LifetimeActionsType(type=action.action),
                     trigger=self._models.LifetimeActionsTrigger(
                         time_after_create=action.time_after_create, time_before_expiry=action.time_before_expiry
                     ),
@@ -789,7 +789,7 @@ class KeyClient(KeyVaultClientBase):
                 for action in lifetime_actions
             ]
 
-        attributes = self._models.KeyRotationPolicyAttributes(expiry_time=kwargs.pop("exires_in", None))
+        attributes = self._models.KeyRotationPolicyAttributes(expiry_time=kwargs.pop("expires_in", None))
         policy = self._models.KeyRotationPolicy(lifetime_actions=lifetime_actions, attributes=attributes)
         result = self._client.update_key_rotation_policy(
             vault_base_url=self._vault_url, key_name=name, key_rotation_policy=policy
--- a/sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_client.py
+++ b/sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_client.py
@@ -17,7 +17,7 @@ except ImportError:
 
 if TYPE_CHECKING:
     # pylint:disable=unused-import
-    from typing import Any, Optional, Union
+    from typing import Any, Iterable, Optional, Union
     from azure.core.paging import ItemPaged
     from azure.core.polling import LROPoller
     from ._models import JsonWebKey
@@ -730,13 +730,14 @@ class KeyClient(KeyVaultClientBase):
 
     @distributed_trace
     def get_key_rotation_policy(self, name, **kwargs):
-        # type: (str, **Any) -> Optional[KeyRotationPolicy]
+        # type: (str, **Any) -> KeyRotationPolicy
         """Get the rotation policy of a Key Vault key.
 
         :param str name: The name of the key.
 
-        :return: The key rotation policy, or None if there is no policy.
-        :rtype: ~azure.keyvault.keys.KeyRotationPolicy or None
+        :return: The key rotation policy.
+        :rtype: ~azure.keyvault.keys.KeyRotationPolicy
+        :raises: :class: `~azure.core.exceptions.HttpResponseError`
         """
         policy = self._client.get_key_rotation_policy(vault_base_url=self._vault_url, key_name=name, **kwargs)
         return KeyRotationPolicy._from_generated(policy)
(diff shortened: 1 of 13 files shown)


sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_client.py
```diff
@@ -727,3 +727,71 @@ def get_random_bytes(self, count, **kwargs):
...
+        # type: (str, **Any) -> Optional[KeyRotationPolicy]
+        """Get the rotation policy of a Key Vault key.
+
+        :param str name: The name of the key.
+
+        :return: The key rotation policy, or None if there is no policy.
+        :rtype: ~azure.keyvault.keys.KeyRotationPolicy or None
+        """
+        policy = self._client.get_key_rotation_policy(vault_base_url=self._vault_url, key_name=name, **kwargs)
+        if policy.id:
```
maorleger, line 742: I realize now that I originally had this in both the issue and my JS PR, but after some discussions with the team we agreed that it would not be necessary. There is a concept of a "default" rotation policy which means that we can always expect to get data back. 

Now, it might not have an id (if the user has never set one) which means that we need to model this differently (i.e. id is optional, as are the timestamps)... this was impactful for TypeScript, but might not be necessary for Python?

At any rate, what we're doing now is always return a KeyRotationPolicy, never `undefined` (or an Optional[T]) - let me know if you want to discuss this.

The TL;DR is - assume returning `KeyRotationPolicy._from_generated(policy)` is always safe (as long as id, createdOn, and updatedOn are nullable)

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_enums.py
```diff
@@ -39,6 +39,13 @@ class KeyOperation(with_metaclass(CaseInsensitiveEnumMeta, str, Enum)):
     export = "export"
 
 
+class KeyRotationPolicyAction(with_metaclass(CaseInsensitiveEnumMeta, str, Enum)):
```
maorleger, line 42: What is CaseInsensitiveEnumMeta? I'm just curious if it does what I think it does 😄

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_client.py
mccoyp: Got it! I remembered you mentioning that a default policy had been discussed but I wasn't sure if this was implemented yet. I'll make those updates!

sdk/keyvault/azure-keyvault-keys/tests/test_key_client.py
```diff
@@ -22,10 +29,19 @@
...
 logging_enabled = get_decorator(logging_enable=True)
 logging_disabled = get_decorator(logging_enable=False)
 
 
+def _assert_rotation_policies_equal(p1, p2):
+    assert p1.id == p2.id
+    assert p1.expires_in == p2.expires_in
+    assert p1.created_on == p2.created_on
+    assert p1.updated_on == p2.updated_on
+    assert len(p1.lifetime_actions) == len(p2.lifetime_actions)
```
maorleger, line 42: Totally up to you - but I would suggest testing that the lifetime_actions are deeply equal (or whatever is equivalent in python). Is that possible? What do you think? 

My reasoning is just concern over deserialization / transformation missing some data - either on input or output - which testing the length of the actions won't discover

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_enums.py
mccoyp: [Here's](https://docs.microsoft.com/en-us/python/api/azure-core/azure.core.caseinsensitiveenummeta?view=azure-python) the class reference -- having our enums inherit from this is a recent addition to the Python guidelines

sdk/keyvault/azure-keyvault-keys/tests/test_key_client.py
mccoyp: That's true, I would like to test for that. My reason for not including a thorough comparison was because I wasn't sure if we can assume some ordering of these lists, and testing for equality if they're unordered gets a bit messy for a helper method. The better solution is probably to just test deep equality on a case-by-case basis so I know what to expect and don't have to write messy, generalized logic. I'll do that, thanks for the suggestion!

mccoyp: Improving these tests caught a bug with how I was handling generated lifetime action objects, so thank you for pointing this out! You saved me from future pain 🤠

maorleger: Awesome! 🎉

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_client.py
```diff
@@ -727,3 +727,69 @@ def get_random_bytes(self, count, **kwargs):
...
         return RandomBytes(value=result.value)
+
+    @distributed_trace
+    def get_key_rotation_policy(self, name, **kwargs):
+        # type: (str, **Any) -> Optional[KeyRotationPolicy]
+        """Get the rotation policy of a Key Vault key.
+
+        :param str name: The name of the key.
+
+        :return: The key rotation policy, or None if there is no policy.
```
maorleger, line 738: nit: this comment is inaccurate

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_enums.py
```diff
@@ -39,6 +39,13 @@ class KeyOperation(with_metaclass(CaseInsensitiveEnumMeta, str, Enum)):
...
+    """The action that will be executed in a key rotation policy"""
+
+    ROTATE = "Rotate"  #: Rotate the key based on the key policy.
+    NOTIFY = "Notify"  #: Trigger Event Grid events. Not configurable in preview; defaults to 30 days before expiry.
```
maorleger, line 46: Consider removing the "Not configurable in preview;" part of the comment if this is something customers will see

maorleger (approved):
🚀

PR #20419, opened 2021-08-26
Suppress CredScan warnings for test certs and fake password examples

mccoyp:
This should address https://github.com/Azure/azure-sdk-for-python-pr/issues/659, https://github.com/Azure/azure-sdk-for-python-pr/issues/660, https://github.com/Azure/azure-sdk-for-python-pr/issues/661, and [warnings](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=1064526&view=logs&j=0874ffeb-2919-5b4b-20de-16a97970b94c&t=439e917e-962a-53e0-8a4b-b8bb4f8ed372&l=57) from new Identity test certificates.

The ```":code:`<pfx-file-password>\`"``` value is an example password value that shows up in many generated docstrings (one such example [here](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/compute/azure-mgmt-compute/azure/mgmt/compute/v2015_06_15/models/_models.py#L3115)).

### DIFF
--- a/eng/CredScanSuppression.json
+++ b/eng/CredScanSuppression.json
@@ -11,6 +11,7 @@
                 "kt#_gahr!@aGERDXA",
                 "Aa1!zyx_",
                 "Aa!1()-xyz",
+                ":code:`<pfx-file-password>`",
                 "000000000000000000000000000000000000000000000000000",
                 "UsernameAndPassword",
                 "123456",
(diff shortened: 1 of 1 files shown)


eng/CredScanSuppression.json
```diff
@@ -11,6 +11,7 @@
                 "kt#_gahr!@aGERDXA",
                 "Aa1!zyx_",
                 "Aa!1()-xyz",
+                ":code:`<pfx-file-password>`",
```
weshaggard, line 14: Do you know if it needs this entire string or just `<pfx-file-password>`?

weshaggard:
Lets watch the results of the run I triggered https://dev.azure.com/azure-sdk/internal/_build/results?buildId=1074828&view=results to make sure these suppress the what we expect.

eng/CredScanSuppression.json
mccoyp: Based on the row and column of the warning, it looks like it should be the whole string -- but verifying with a test run is the smart thing to do, thanks for starting that!

weshaggard (approved):

PR #20431, opened 2021-08-26
Search: Skill Versions

### LINKED ISSUE #20432: Search: Add Skill Versions
Add a version field to disambiguate between the original (V1) and V3 versions of EntityRecogntion and Sentiment skills, similar to Java and .NET.

tjprescott:
Closes #20432.

Adds `SentimentSkillVersion` and `EntityRecognitionSkillVersion`, and a customized model for `SentimentSkill` and `EntityRecognitionSkill` that uses this version to create the correct model. 

- [x] Add tests


sdk/search/azure-search-documents/azure/search/documents/indexes/models/_models.py
```diff
@@ -590,3 +944,13 @@ def unpack_analyzer(analyzer):
...
+
+
+def validate(kwargs, version, unsupported):
+    unsupported = [unsupported] if isinstance(unsupported, str) else unsupported
+    errors = []
+    for param in unsupported:
+        if param in kwargs:
+            errors.append(param)
+    if errors:
+        raise ValueError("Unsupported parameters for skill version {}: {}".format(version, ', '.join(errors)))
```
xiangyan99, line 956: It looks like a breaking change.

In old behavior, non-supported parameters are ignored rather than raise exception in ctor.

tjprescott: My understanding was you wanted an error thrown:

> This is from Alan said: if user calls setter for non-V1 properties in V1 model, they raise exceptions.
> 
> And it looks to me this is like the scenario user wants to access non-existing properties.
> 
> We can double check with Johan/Anna.

xiangyan99: It is in property's setter (and maybe getter as well). But if we add it in ctor, it is a breaking change.

xiangyan99 (commented):
nit: I think this change should not impact any network request/response. So no need to re-record the tests. Of course, there is no harm to do so.

sdk/search/azure-search-documents/azure/search/documents/indexes/models/_models.py
```diff
@@ -3,24 +3,378 @@
...
     SearchIndexerDataSource as _SearchIndexerDataSource,
+    SearchIndexerSkill,
+    SearchIndexerSkillset as _SearchIndexerSkillset,
+    SentimentSkill as _SentimentSkillV1,
+    SentimentSkillV3 as _SentimentSkillV3,
     SynonymMap as _SynonymMap,
     DataSourceCredentials,
     AzureActiveDirectoryApplicationCredentials,
 )
+from .._generated.models._search_client_enums import _CaseInsensitiveEnumMeta
```
xiangyan99, line 28: Maybe we should use azure.core.CaseInsensitiveEnumMeta

tjprescott:
I had to re-record the tests because I needed to change the skillset tests... and then re-recorded all to ensure I didn't accidentally break anything.

sdk/search/azure-search-documents/azure/search/documents/indexes/models/_models.py
tjprescott: @annatisch can you provide guidance here? It seems odd to me to just silently accept bad params in the constructor but fail on a setter/getter. It seems we should either always validate or never validate.

sdk/search/azure-search-documents/azure/search/documents/indexes/models/_models.py
tjprescott: If we did, we would be the only SDK doing so. It seems all SDKs declare this class in their `_foo_client_enum.py` file. Presumably, this is part of the autorest generation.

xiangyan99: The problem is if we reply on auto-generated private code, it may be subject to change(break). Of course, we can fix that when it breaks.

tjprescott: As it turns out, using `CaseInsensitiveEnum` at all in custom code causes pylint to vomit in a way that you can't disable. Anna told me to make it a regular, case sensitive enum, which fixed the problem.

sdk/search/azure-search-documents/azure/search/documents/indexes/models/_models.py
```diff
@@ -13,14 +15,196 @@
...
+
 
 DELIMITER = "|"
 
 
+class EntityRecognitionSkillVersion(with_metaclass(_CaseInsensitiveEnumMeta, str, Enum)):
+    """Specifies the Entity Recognition skill version to use."""
+
+    #: Use Entity Recognition skill V1.
+    V1 = "#Microsoft.Skills.Text.EntityRecognitionSkill"
```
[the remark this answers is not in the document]
annatisch: The string value here is certainly unfriendly to type out. Especially give that it starts with a '#' character....
Are these strings some kind of 'known' value? e.g. if I were a customer, could I have copied this from the portal? Or conceptual reference docs/API docs anywhere and paste it in?

sdk/search/azure-search-documents/azure/search/documents/indexes/models/_models.py
```diff
@@ -13,14 +15,196 @@
...
+
+    def __init__(
+        self,
+        **kwargs
+    ):
+        super(SentimentSkill, self).__init__(**kwargs)
+        skill_version = kwargs.get('skill_version', SentimentSkillVersion.V1)
+        self.odata_type = skill_version  # type: str
+        self.default_language_code = kwargs.get('default_language_code', None)
+        self.include_opinion_mining = kwargs.get('include_opinion_mining', False)
```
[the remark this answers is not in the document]
annatisch: I would think that if the user explicitly sends a value not supported by their target version, that would raise. However we wouldn't send the value not supported by the target version by default

sdk/search/azure-search-documents/azure/search/documents/indexes/models/_models.py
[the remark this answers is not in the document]
xiangyan99: These are the odata_type from swagger.

I don't think there is an easy way for customers to find them.

sdk/search/azure-search-documents/azure/search/documents/indexes/models/_models.py
xiangyan99: I am afraid the answer is NO. 

Service team may add/remove properties so it is possible V3 has unsupported properties against V4 and/or V4 has unsupported properties against V3.

sdk/search/azure-search-documents/azure/search/documents/indexes/models/_models.py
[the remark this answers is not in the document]
tjprescott: But isn't that the point of the enum? SkillVersion dot all-the-values?

sdk/search/azure-search-documents/azure/search/documents/indexes/models/_models.py
annatisch: In that case, a customer is always going to be broken if they're using LATEST. It's just a matter of where the error is raised - either at the model or at the operation.
We also need to accommodate the fact that people may not use models, and pass in their inputs as dicts, in which case, again, the error will be at the operation level.

sdk/search/azure-search-documents/azure/search/documents/indexes/models/_models.py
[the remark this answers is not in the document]
tjprescott: Yes because we get the skillset and transform it to the custom models that contain this property.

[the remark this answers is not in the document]
annatisch: I would leave these as the over-the-wire values for now. We can reassess before GA.

sdk/search/azure-search-documents/azure/search/documents/indexes/models/_models.py
tjprescott: I'll work on moving the validation into the operation.

sdk/search/azure-search-documents/azure/search/documents/indexes/models/_models.py
tjprescott: I created an issue to track this known bug/limitation:
https://github.com/Azure/azure-sdk-for-python/issues/20520

sdk/search/azure-search-documents/azure/search/documents/indexes/models/_models.py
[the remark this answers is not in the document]
tjprescott: I have left these as the over-the-wire values for now, as requested.

sdk/search/azure-search-documents/azure/search/documents/indexes/models/_models.py
tjprescott: Per our Teams discussion, I've removed validation entirely for these release. We can align on the desired behavior, because the service does not throw when unsupported kwargs are passed.

sdk/search/azure-search-documents/azure/search/documents/indexes/models/_models.py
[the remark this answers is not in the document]
tjprescott: Per our discussion in Teams, I've removed all validation until we align across languages on what we want to happen.

tjprescott:
@annatisch @xiangyan99 I created an issue to track the client-side validation aspects: https://github.com/Azure/azure-sdk-for-python/issues/20526

xiangyan99 (approved):

PR #20456, opened 2021-08-30
add six to testserver setup.py

iscai-msft:


xiangyan99 (approved):

scbedd:
@iscai-msft we should merge this.

PR #20463, opened 2021-08-30
[Storage]fix live test

xiafu-msft:
Service made a quick query fix which breaks the live test. 
Previous service behavior: 
input format has_header=True, output doesn't have header no matter if has_header=True or not in output format

Current behavior:
input format has_header=True, customer can choose if they want to include header in the output format.
input format=False customer can only set has_header=False in output format.


sdk/storage/azure-storage-blob/tests/test_quick_query.py
```diff
@@ -228,7 +228,8 @@ def test_quick_query_iter_records_with_headers(self, resource_group, location, s
         blob_client.upload_blob(CSV_DATA, overwrite=True)
 
         input_format = DelimitedTextDialect(has_header=True)
-        reader = blob_client.query_blob("SELECT * from BlobStorage", blob_format=input_format)
+        output_format = DelimitedTextDialect(has_header=False)
```
kasobol-msft, line 231: do we have coverage for `output_format = DelimitedTextDialect(has_header=True)` ?

xiafu-msft: good question, the test failed because we default the output_format to input_format which means has_header=True such that the assertion failed. So I explicitly set has_header=True to make the original assertion pass.
I think we should keep the original setting as well, let me add a test.

kasobol-msft: yes,we should add coverage in this case.

tasherif-msft (approved):
LGTM

sdk/storage/azure-storage-blob/tests/test_quick_query.py
xiafu-msft: Just a note: I added the tests for has_header=True, just accidentally amended to the existing commit.

Issue #19610, opened 2021-06-30
[ACR] [Before GA] Update cloud configuration API


annelo-msft:
Once the service team has added data plane scopes, we will update our `authentication_scope` kwarg as follows:

- [x] Remove `authentication_scope` kwarg
- [x] Add `audience` (or similarly named) kwarg
- [x] `audience` is an extensible enum, containing the audience strings for known clouds; the user can override with a string if needed
- [x] If `audience` is set, client library will append `/.default` to `audience` value to specify the default scope
- [x] `audience` will default to data-plane scope root, which should be common across clouds [Note: this is not yet available from the service]
- [x] Throw an exception is `audience` is not set and update samples to reflect this

In Python, we will also
- [x] Initially not support `credential_scopes` kwarg
- If we do evenually support the `credential_scopes` kwarg, raise an error if this is different from the scope determined from `audience`

Please see https://github.com/azure/azure-sdk-for-net/issues/21603 for further details.

annelo-msft:
@johanste @lmazuel @schaabs FYI

YalinLi0312:
Fixed.

PR #20464, opened 2021-08-30
[ACR] Update cloud configuration API

YalinLi0312:
For issue: https://github.com/Azure/azure-sdk-for-python/issues/19610


sdk/containerregistry/azure-containerregistry/azure/containerregistry/_enums.py
```diff
@@ -0,0 +1,17 @@
...
+# pylint:skip-file (avoids crash due to six.with_metaclass https://github.com/PyCQA/astroid/issues/713)
+from enum import Enum
+from six import with_metaclass
+
+from azure.core import CaseInsensitiveEnumMeta
+
+class ContainerRegistryAudience(with_metaclass(CaseInsensitiveEnumMeta, str, Enum)):
+    """Supported container registry audience"""
+
+    ARM_CHINA_VALUE = "https://management.chinacloudapi.cn"
```
annelo-msft, line 14: In .NET we've called the enum values `AzureResourceManagerChina`, e.g., because we have ARM in another enum indicating processor architecture, and we don't have the `_VALUE` suffix.  See [the APIView here](https://apiview.dev/Assemblies/Review/7ebfcb36ff6149ba96170bd059094513#Azure.Containers.ContainerRegistry.ContainerRegistryAudience).  I don't have a strong opinion regarding Python values, but I'd suggest to follow the precedent set elsewhere in Python APIs.

sdk/containerregistry/azure-containerregistry/azure/containerregistry/aio/_async_container_registry_client.py
```diff
@@ -46,12 +46,13 @@ def __init__(self, endpoint: str, credential: Optional["AsyncTokenCredential"] =
...
                 :caption: Instantiate an instance of `ContainerRegistryClient`
-        """
+        """       
+        defaultScope = [audience + "/.default"]
         if not endpoint.startswith("https://") and not endpoint.startswith("http://"):
             endpoint = "https://" + endpoint
         self._endpoint = endpoint
         self._credential = credential
-        super(ContainerRegistryClient, self).__init__(endpoint=endpoint, credential=credential, **kwargs)
+        super(ContainerRegistryClient, self).__init__(endpoint=endpoint, credential=credential, credential_scopes=defaultScope, **kwargs)
```
annelo-msft, line 55: My understanding is that Python wasn't going to use the `credential_scopes` kwarg in this library at all.  It would be worth following up with Johan, Laurent, or Anna on this as an implementation detail.

sdk/containerregistry/azure-containerregistry/tests/testcase.py
```diff
@@ -220,19 +218,25 @@ def get_authority(endpoint):
...
+        return AzureAuthorityHosts.AZURE_GERMANY
     raise ValueError("Endpoint ({}) could not be understood".format(endpoint))
 
 
-def get_authorization_scope(authority):
+def get_audience(authority):
     if authority == AzureAuthorityHosts.AZURE_PUBLIC_CLOUD:
         logger.warning("Public auth scope")
-        return ["https://management.core.windows.net/.default"]
+        return ContainerRegistryAudience.ARM_PUBLIC_CLOUD_VALUE
```
annelo-msft, line 230: It looks like the Python precedent is not to use the `_VALUE` suffix (just looking at `AzureAuthorityHosts`)

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_enums.py
mccoyp: In [azure-identity](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/identity/azure-identity/azure/identity/_constants.py#L14) we have `AZURE_CHINA`, `AZURE_GERMANY`, etc. so it makes sense to drop `_VALUE`. I think using `ARM_` instead of `AZURE_RESOURCE_MANAGER_` would be nice for brevity, but that's a question for Anna and Johan

sdk/containerregistry/azure-containerregistry/azure/containerregistry/aio/_async_container_registry_client.py
YalinLi0312: Yes, `credential_scopes` has been removed from kwarg in client class in this PR. The `credential_scopes` here is a kwarg in base class, just for internal use. I've talked with @annatisch , it's okay to use it internally.

@annatisch Please correct me if I misunderstood.

YalinLi0312:
@annelo-msft Thanks for your review! 

I updated the name to be without the `_VALUE` suffix, and the `credential_scopes` kwarg is waiting Anna's confirmation. 

Thanks

annelo-msft (approved):
LGTM.  Please make sure someone from the Python teams signs off on this as well.

sdk/containerregistry/azure-containerregistry/azure/containerregistry/aio/_async_container_registry_client.py
annatisch: Is this a concern about the logic/mechanics around using 'credential_scope'? Or just the parameter name?
If the latter, then this is an internal implementation detail that could be refactored at any time and is not surfaced to the public API in any way.
If the former, please provide more details :)

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_enums.py
annatisch: @johanste - having looked into the cloud configuration topic - do you have any concerns with `ARM_CHINA/ARM_GERMANY/ARM_FOO` naming convention?

sdk/containerregistry/azure-containerregistry/azure/containerregistry/aio/_async_container_registry_client.py
annelo-msft: My concern was the latter.  There was a conversation between Johan and Laurent about whether to expose the `credential_scopes` kwarg, which I understand Python is trying to sunset, in favor of the `audience` kwarg.  If this implementation doesn't put that decision in danger, then I'm not worried at all.

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_enums.py
johanste: I would just document the possible values. No need for a public type.

johanste: (which also side-steps the naming convention question :))

annatisch (approved):

PR #20469, opened 2021-08-31
[Test proxy] Add migration guide

mccoyp:
This adds a migration guide to get folks started on migrating to the test proxy -- this was initially written as an informal gist, so it may be terrible (apologies)

### DIFF
--- a/doc/dev/test_proxy_migration_guide.md
+++ b/doc/dev/test_proxy_migration_guide.md
@@ -170,7 +170,7 @@ case.
 [docker_start_proxy]: https://github.com/Azure/azure-sdk-for-python/blob/main/eng/common/testproxy/docker-start-proxy.ps1
 [general_docs]: https://github.com/Azure/azure-sdk-tools/blob/main/tools/test-proxy/README.md
 [proxy_cert_docs]: https://github.com/Azure/azure-sdk-tools/blob/main/tools/test-proxy/documentation/trusting-cert-per-language.md
-[pytest_collection]: https://docs.pytest.org/en/latest/explanation/goodpractices.html#test-discovery
-[pytest_fixtures]: https://docs.pytest.org/en/6.2.x/fixture.html#scope-sharing-fixtures-across-classes-modules-packages-or-session
+[pytest_collection]: https://docs.pytest.org/latest/goodpractices.html#test-discovery
+[pytest_fixtures]: https://docs.pytest.org/latest/fixture.html#scope-sharing-fixtures-across-classes-modules-packages-or-session
 [sanitizers]: https://github.com/Azure/azure-sdk-tools/blob/main/tools/test-proxy/Azure.Sdk.Tools.TestProxy/README.md#session-and-test-level-transforms-sanitiziers-and-matchers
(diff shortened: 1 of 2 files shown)


scbedd (approved):
I like this writeup a lot.

doc/dev/test_proxy_migration_guide.md
```diff
@@ -0,0 +1,174 @@
...
+
+## Run the tests
+
+### Perform one-time setup
+
+The test proxy is made available for your tests via a Docker container. Run the following command to build this
+container, providing the full path to the root of your local `azure-sdk-for-python` repo in place of `{path}`:
+
+```cmd
+docker container create -v C:\\{path}\\azure-sdk-for-python:/etc/testproxy -p 5001:5001 -p 5000:5000 --name ambitious_azsdk_test_proxy azsdkengsys.azurecr.io/engsys/testproxy-lin
```
scbedd, line 63: Do we want to just point them to the `docker-start-proxy.ps1` under `eng/common/testproxy`?

It does pretty much exactly this (and figures out where the root of the repo is for you).

doc/dev/test_proxy_migration_guide.md
```diff
@@ -0,0 +1,174 @@
...
+Since the test proxy doesn't use [`vcrpy`][vcrpy], tests don't use a scrubber to sanitize values in recordings.
+Instead, sanitizers (as well as matchers and transforms) can be registered on the proxy as detailed in
+[this][sanitizers] section of the proxy documentation. At the time of writing, sanitizers can be registered via
+the `add_sanitizer` method in `devtools_testutils`. For example, at the start of a test file, you can set up a URI
+sanitizer for all tests by doing something like the following:
+
+```python
+from devtools_testutils import add_sanitizer, AzureRecordedTestCase, ProxyRecordingSanitizer
+
+add_sanitizer(ProxyRecordingSanitizer.URI, value="fakeendpoint")
```
scbedd, line 107: in our example we should probably provide best practices, which probably means putting this under a session fixture in `conf.py`.

scbedd: I could be missing something about that though. Feel free to tell me!

doc/dev/test_proxy_migration_guide.md
```diff
@@ -0,0 +1,174 @@
...
+
+class TestExample(AzureRecordedTestCase):
+    ...
+```
+
+`add_sanitizer` accepts a sanitizer type from the ProxyRecordingSanitizer enum as a required parameter. A regular
+expression to match and a value to use in recordings can be optionally provided as `regex` and `value` arguments,
```
scbedd, line 114: Right now we only provide `regex` and `value` right.

All the regex sanitizers also allow a 3rd argument `groupForReplace`. Maybe we swap `add_sanitizer` to variable number of arguments, throwing if we don't get the expected ones. We could generate some nice errors fairly simply.

scbedd (approved):
I left some feedback about `sanitizer`. As long as we cover `groupForReplace` I'm ok with leaving design as is. Leaving it up to you!

PR #20481, opened 2021-08-31
[rest] allow LRO to handle rest requests and responses

iscai-msft:


xiangyan99 (commented):
Lots of overlaps between this one and #20448?

iscai-msft:
@xiangyan99 yeah let me redo this PR

PR #20483, opened 2021-08-31
have LRO support rest requests and responses

iscai-msft:
i haven't started out with a changelog, since it's not quite public, how I'm able to do LRO polling


sdk/core/azure-core/azure/core/polling/base_polling.py
```diff
@@ -574,10 +575,19 @@ def request_status(self, status_link):
         """
         if self._path_format_arguments:
             status_link = self._client.format_url(status_link, **self._path_format_arguments)
-        request = self._client.get(status_link)
         # Re-inject 'x-ms-client-request-id' while polling
         if "request_id" not in self._operation_config:
             self._operation_config["request_id"] = self._get_request_id()
+        if hasattr(self._initial_response.http_response, "content"):
```
xiangyan99, line 581: Why not isinstance?

xiangyan99: And given you don't need _operation_config, you can move your code to right after

```py
if self._path_format_arguments:
            status_link = self._client.format_url(status_link, **self._path_format_arguments)
```

and no need to remove request = self._client.get(status_link)

iscai-msft: i try to go for duck-typing more, i think it's more pythonic.

@xiangyan99 that was a bug I had that @annatisch caught, jsut fixed it!

iscai-msft: @xiangyan99 also am not removing `request = self._client.get(status_link)`, it's just that that returns an old request, and if a user gives me a new response, i want to create a new request to continue the chain

annatisch (approved):

xiangyan99 (approved):
LGTM.

PR #20485, opened 2021-08-31
Search: Add normalizer property to AnalyzeText.

tjprescott:
Adds the `normalizer_name` property to `AnalyzeTextOptions`.

Not sure why the regenerated syntax diffs with the previously committed syntax.


xiangyan99 (approved):
Please update changelog.

PR #20489, opened 2021-08-31
[Exporter] Support redirect response in exporter

lzchen:
307 and 308 are redirect responses from Breeze.

azure.core supports automatic redirection via `RedirectPolicy`, but does not set the new endpoint appropriately, so we extract out the logic and perform redirects in the exporter.


sdk/monitor/azure-monitor-opentelemetry-exporter/tests/test_base_exporter.py
```diff
@@ -12,12 +12,14 @@
...
 from azure.core.exceptions import HttpResponseError, ServiceRequestError
+from azure.core.pipeline.transport import HttpResponse
 from azure.monitor.opentelemetry.exporter.export._base import (
     BaseExporter,
     ExportResult,
     get_trace_export_result,
 )
-from azure.monitor.opentelemetry.exporter._generated.models import TelemetryItem
+from azure.monitor.opentelemetry.exporter._generated import AzureMonitorClient
+from azure.monitor.opentelemetry.exporter._generated.models import TelemetryItem, TrackResponse
```
rakshith91, line 22: are these something we want to expose to the users?

rakshith91 (approved):

Issue #19158, opened 2021-06-08
Run mypy in Key Vault CI


chlowell:
Fix current errors and add packages to [mypy_hard_failure_packages.py](https://github.com/Azure/azure-sdk-for-python/blob/master/eng/tox/mypy_hard_failure_packages.py):
- [x] azure-keyvault-administration (#19246)
- [x] azure-keyvault-certificates(#20525)
- [x] azure-keyvault-keys(#20545)
- [x] azure-keyvault-secrets(#20507)

mccoyp:
For reference: to run mypy for a package (e.g. `azure-keyvault-certificates`), run
`mypy --exclude _generated --ignore-missing azure`
from the top of the package -- in this case, from `azure-sdk-for-python/sdk/keyvault/azure-keyvault-certificates`

mccoyp:
Steps to resolve:

For each package:

1. Run mypy check
2. Resolve warnings
3. Once there are no warnings, add package to hard failures file

YalinLi0312:
Fixed.

PR #20507, opened 2021-09-01
Run mypy in azure-keyvault-secrets CI

YalinLi0312:
Part of issue https://github.com/Azure/azure-sdk-for-python/issues/19158.

This PR fixed all errors in `azure-keyvault-secrets`, and all shared folder errors in `azure-keyvault-keys` and `azure-keyvault-certifiates`.


sdk/keyvault/azure-keyvault-secrets/azure/keyvault/secrets/_client.py
```diff
@@ -294,7 +294,7 @@ def restore_secret_backup(self, backup, **kwargs):
 
     @distributed_trace
     def begin_delete_secret(self, name, **kwargs):
-        # type: (str, **Any) -> DeletedSecret
+        # type: (str, **Any) -> KeyVaultOperationPoller
```
mccoyp, line 297: KeyVaultOperationPoller is a private type since it's in `_shared`, so we should list the return type as an LROPoller:
```suggestion
        # type: (str, **Any) -> LROPoller
```

sdk/keyvault/azure-keyvault-certificates/azure/keyvault/certificates/_shared/_polling.py
```diff
@@ -33,11 +33,11 @@ class KeyVaultOperationPoller(LROPoller):
     # pylint: disable=arguments-differ
     def __init__(self, polling_method):
         # type: (PollingMethod) -> None
-        super(KeyVaultOperationPoller, self).__init__(None, None, None, NoPolling())
+        super(KeyVaultOperationPoller, self).__init__(None, None, self.result, NoPolling())
```
mccoyp, line 36: Was there a mypy complaint about passing `None` here? I don't think `self.result` would be a fitting deserialization callback here -- the `deserialization_callback` provided to LROPoller should accept a Response as input and return a deserialized object ([link](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/core/azure-core/azure/core/polling/_poller.py#L151)).

sdk/keyvault/azure-keyvault-secrets/azure/keyvault/secrets/_models.py
```diff
@@ -11,19 +11,19 @@
...
     from datetime import datetime
     from ._generated.v7_1 import models as _models
 
 
 class SecretProperties(object):
     """A secret's id and attributes."""
 
-    def __init__(self, attributes, vault_id, **kwargs):
-        # type: (_models.SecretAttributes, str, **Any) -> None
+    def __init__(self, attributes=None, vault_id=None, **kwargs):
```
mccoyp, line 22: Is there a reason to make these parameters optional?

sdk/keyvault/azure-keyvault-certificates/azure/keyvault/certificates/_shared/_polling.py
YalinLi0312: This was the error:
`azure\keyvault\secrets\_shared\_polling.py:36: error: Argument 3 to "__init__" of "LROPoller" has incompatible type "None"; expected "Callable[..., Any]"`

mccoyp: Ah, makes sense. I would either add a `type: ignore` comment here or in that case provide a callable that returns `None`; for example, `lambda *_: None`:
```python
super(KeyVaultOperationPoller, self).__init__(None, None, lambda *_: None, NoPolling())
```

sdk/keyvault/azure-keyvault-secrets/azure/keyvault/secrets/_models.py
mccoyp: Hm... yeah, that's true, and it doesn't look like we verify anywhere that these properties should be non-`None`. That being said, I'm not sure if we can/should make these property return types optional since these are properties that we should really ensure have a value. @chlowell, is there any context I'm missing for why we assume these properties are populated (aside from things having worked well thus far)?

chlowell: I doubt there's a case in which these have no value--that would be a secret with no ID or management attributes. The autorest client makes these properties optional either by general design or because they aren't required by the spec. I suppose this is one reason to favor instance attributes over properties on models, though that pushes the problem of None checking onto applications.

mccoyp: Would it be reasonable to make the property -> instance attribute switch here? If not, I think I'd feel more comfortable ignoring the mypy check here and evaluating this again in a more focused PR/issue

chlowell: I'm uncertain that's the best solution. It doesn't eliminate the need to anticipate None values. And what if there's no case for this library in which SecretBundle.attributes or .id is null? That's a broader question for the service team. For now I think it's reasonable to reflect the reality of the code we have, i.e. to acknowledge that in principle at least, some values could be None, and satisfy mypy by accounting for that. Which is orthogonal to giving these parameters default values--let's not do that. The fact that a parameter could have value None, and so is "optional" in a type checking sense, doesn't imply it should have a default value None, and so be "optional" in a "caller isn't required to pass a value" sense.

sdk/keyvault/azure-keyvault-secrets/azure/keyvault/secrets/_models.py
```diff
@@ -87,55 +87,55 @@ def key_id(self):
 
     @property
     def enabled(self):
-        # type: () -> bool
+        # type: () -> Union[bool, None]
```
chlowell, line 90: Union[T, None] == Optional[T]

YalinLi0312: I'll update it, and in all other places.

sdk/keyvault/azure-keyvault-secrets/azure/keyvault/secrets/_models.py
YalinLi0312: I'll ignore the mypy warning here and evaluate later.

sdk/keyvault/azure-keyvault-secrets/azure/keyvault/secrets/_models.py
```diff
@@ -69,10 +69,10 @@ def content_type(self):
 
     @property
     def id(self):
-        # type: () -> str
+        # type: () -> Optional[str]
         """The secret's id
 
-        :rtype: str
+        :rtype: optional[str]
```
mccoyp, line 75: nit: I think this is a bit clearer in generated documentation (this applies to all optional return types)
```suggestion
        :rtype: str or None
```

sdk/keyvault/azure-keyvault-secrets/azure/keyvault/secrets/_models.py
```diff
@@ -51,8 +51,8 @@ def _from_secret_item(cls, secret_item):
         # type: (_models.SecretItem) -> SecretProperties
         """Construct a SecretProperties from an autorest-generated SecretItem"""
         return cls(
-            secret_item.attributes,
-            secret_item.id,
+            secret_item.attributes,  # type: ignore
+            secret_item.id,  # type: ignore
             content_type=secret_item.content_type,
```
mccoyp, line 56: It looks like `content_type`, `tags`, and possibly other properties should technically be optional as well

YalinLi0312: Here is ignoring the errors that expect `attributes` and `id` to be **optional** while we don't want to set them with `None`.

sdk/keyvault/azure-keyvault-secrets/azure/keyvault/secrets/_models.py
YalinLi0312: Make sense, I'll update them to be this style.

sdk/keyvault/azure-keyvault-secrets/azure/keyvault/secrets/_models.py
mccoyp: True -- I just mean that I'm surprised mypy doesn't complain despite the `content_type` property on line 62, for example, not being listed as possibly `None`

sdk/keyvault/azure-keyvault-secrets/azure/keyvault/secrets/_models.py
```diff
@@ -191,7 +191,7 @@ class KeyVaultSecret(object):
     """All of a secret's properties, and its value."""
 
     def __init__(self, properties, value):
-        # type: (SecretProperties, str) -> None
+        # type: (SecretProperties, Optional[str]) -> None
```
mccoyp, line 194: This is a case where the type checking gets a little confusing. A SecretBundle can have a `None` `value`, which is why this `Optional[str]` type makes sense -- but also, [SecretSetParameters](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/keyvault/azure-keyvault-secrets/azure/keyvault/secrets/_generated/v7_2/models/_models_py3.py#L560) requires a non-`None` value. We have to send over SecretSetParameters to create a secret in the first place, meaning that -- in theory -- any secret we get back and turn into a KeyVaultSecret _should_ have a value for `value`. I don't know what to do about this yet, but I'm thinking aloud a bit about why this is awkward.

chlowell: When does SecretBundle have a None value?

mccoyp: It doesn't in any real scenario I've seen (and I doubt it ever would), but its [type signature](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/keyvault/azure-keyvault-secrets/azure/keyvault/secrets/_generated/v7_2/models/_models_py3.py#L128) leaves the possibility open.

sdk/keyvault/azure-keyvault-secrets/azure/keyvault/secrets/_models.py
```diff
@@ -69,10 +69,10 @@ def content_type(self):
 
     @property
     def id(self):
-        # type: () -> str
+        # type: () -> Optional[str]
```
mccoyp, line 72: Since we're enforcing that the `attributes` and `vault_id` given to `__init__` are non-`None`, this property can assume it's non-`None` and we can remove any `if self._attributes` and `if self._vault_id` checks in properties

sdk/keyvault/azure-keyvault-secrets/azure/keyvault/secrets/_models.py
```diff
@@ -23,7 +23,7 @@ def __init__(self, attributes, vault_id, **kwargs):
         # type: (_models.SecretAttributes, str, **Any) -> None
         self._attributes = attributes
         self._id = vault_id
-        self._vault_id = KeyVaultSecretIdentifier(vault_id)
+        self._vault_id = KeyVaultSecretIdentifier(vault_id) if vault_id else None
```
mccoyp, line 26: We require a non-`None` `vault_id` so we can remove this check

sdk/keyvault/azure-keyvault-secrets/azure/keyvault/secrets/_models.py
YalinLi0312: Then this introduces one question that happened several times in mypy ci: 
when swagger defines a property to be optional while we think there's no chance to be None, should we ignore mypy error in this case?

mccoyp (approved):

Issue #16333, opened 2021-01-25
[Key Vault] Fix flaky certificate request cancellation test


mccoyp:
[`test_async_request_cancellation_and_deletion`](https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/keyvault/azure-keyvault-certificates/tests/test_certificates_client.py#L385) often fails locally and in nightly builds because the certificate creation request succeeds before it can be cancelled. Keeping the operation in a pending state until cancellation, possibly by specifying a different issuer in the certificate's policy, would resolve this. The async version of the test is currently skipped -- we should also get this test running.

PR #20510, opened 2021-09-02
[Key Vault] Skip flaky certificate cancellation test in live mode

mccoyp:
Not a real solution to #16333 yet, but this is a temporary fix to recently increased flakiness that's been making live pipeline tests fail consistently.


sdk/keyvault/azure-keyvault-certificates/tests/test_certificates_client.py
```diff
@@ -381,6 +381,9 @@ def test_recover_and_purge(self, client, **kwargs):
     @all_api_versions()
     @client_setup
     def test_async_request_cancellation_and_deletion(self, client, **kwargs):
+        if self.is_live:
+            pytest.skip("Skipping by default because of operation flakiness in pipeline testing")
```
chlowell, line 385: Consider linking the issue here for the benefit of the parallel universe in which the issue gets fixed but the fixer forgets to unskip this test 😸

chlowell (approved):

PR #20525, opened 2021-09-02
Run mypy in azure-keyvault-certificates CI

YalinLi0312:
This PR is for part of issue #19158, it fixed all errors in azure-keyvault-certificates package.

### DIFF
--- a/sdk/keyvault/azure-keyvault-certificates/azure/keyvault/certificates/_models.py
+++ b/sdk/keyvault/azure-keyvault-certificates/azure/keyvault/certificates/_models.py
@@ -795,7 +795,10 @@ class CertificatePolicy(object):
     def _from_certificate_policy_bundle(cls, certificate_policy_bundle):
         # type: (Optional[models.CertificatePolicy]) -> CertificatePolicy
         """Construct a CertificatePolicy from an autorest-generated CertificatePolicy"""
-        if certificate_policy_bundle and certificate_policy_bundle.lifetime_actions:
+        if certificate_policy_bundle is None:
+            return cls()
+
+        if certificate_policy_bundle.lifetime_actions:
             lifetime_actions = [
                 LifetimeAction(
                     action=CertificatePolicyAction(item.action.action_type) if item.action else None,
@@ -806,21 +809,19 @@ class CertificatePolicy(object):
             ]  # type: Optional[List[LifetimeAction]]
         else:
             lifetime_actions = None
-        x509_certificate_properties = (
-            certificate_policy_bundle.x509_certificate_properties if certificate_policy_bundle else None
-        )
+        x509_certificate_properties = certificate_policy_bundle.x509_certificate_properties
         if x509_certificate_properties and x509_certificate_properties.key_usage:
             key_usage = [
                 KeyUsageType(k) for k in x509_certificate_properties.key_usage
             ]  # type: Optional[List[KeyUsageType]]
         else:
             key_usage = None
-        key_properties = certificate_policy_bundle.key_properties if certificate_policy_bundle else None
+        key_properties = certificate_policy_bundle.key_properties
         curve_name = getattr(key_properties, "curve", None)  # missing from 2016-10-01 KeyProperties
         if curve_name:
             curve_name = KeyCurveName(curve_name)
 
-        issuer_parameters = certificate_policy_bundle.issuer_parameters if certificate_policy_bundle else None
+        issuer_parameters = certificate_policy_bundle.issuer_parameters
         return cls(
             issuer_name=issuer_parameters.name if issuer_parameters else None,
             subject=(x509_certificate_properties.subject if x509_certificate_properties else None),
@@ -837,12 +838,11 @@ class CertificatePolicy(object):
             key_usage=key_usage,
             content_type=(
                 CertificateContentType(certificate_policy_bundle.secret_properties.content_type)
-                if certificate_policy_bundle and
-                certificate_policy_bundle.secret_properties and
+                if certificate_policy_bundle.secret_properties and
                 certificate_policy_bundle.secret_properties.content_type
                 else None
             ),
-            attributes=certificate_policy_bundle.attributes  if certificate_policy_bundle else None,
+            attributes=certificate_policy_bundle.attributes,
             san_emails=(
                 x509_certificate_properties.subject_alternative_names.emails
                 if x509_certificate_properties and x509_certificate_properties.subject_alternative_names
--- a/sdk/keyvault/azure-keyvault-certificates/azure/keyvault/certificates/_models.py
+++ b/sdk/keyvault/azure-keyvault-certificates/azure/keyvault/certificates/_models.py
@@ -61,25 +61,25 @@ class AdministratorContact(object):
     @property
     def email(self):
         # type: () -> Optional[str]
-        """:rtype: optional[str]"""
+        """:rtype: str or None"""
         return self._email
 
     @property
     def first_name(self):
         # type: () -> Optional[str]
-        """:rtype: optional[str]"""
+        """:rtype: str or None"""
         return self._first_name
 
     @property
     def last_name(self):
         # type: () -> Optional[str]
-        """:rtype: optional[str]"""
+        """:rtype: str or None"""
         return self._last_name
 
     @property
     def phone(self):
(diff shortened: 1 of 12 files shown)


sdk/keyvault/azure-keyvault-certificates/azure/keyvault/certificates/_polling.py
```diff
@@ -6,6 +6,16 @@
 import time
 
 from azure.core.polling import PollingMethod
+from azure.keyvault.certificates import KeyVaultCertificate, CertificateOperation
```
mccoyp, line 9: We can move these inside the `if TYPE_CHECKING:` block since they're only referenced in typehints

sdk/keyvault/azure-keyvault-certificates/azure/keyvault/certificates/_models.py
```diff
@@ -513,104 +513,104 @@ def _from_certificate_operation_bundle(cls, certificate_operation_bundle):
...
+        return self._vault_id.vault_url if self._vault_id else None
 
     @property
     def issuer_name(self):
-        # type: () -> str
+        # type: () -> Union[str, WellKnownIssuerNames, None]
         """The name of the issuer of the certificate.
 
-        :rtype: str
+        :rtype: union[str, WellKnownIssuerNames, None]
```
mccoyp, line 540: Similarly to using `Type or None` for `Optional[Type]`, I would probably change this to
```suggestion
        :rtype: str or WellKnownIssuerNames or None
```

sdk/keyvault/azure-keyvault-certificates/azure/keyvault/certificates/_models.py
```diff
@@ -369,25 +369,25 @@ def key_id(self):
 
     @property
     def secret_id(self):
-        # type: () -> str
-        """:rtype: str"""
+        # type: () -> Optional[Any]
```
mccoyp, line 372: It looks like the [secret ID of a CertificateBundle](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/keyvault/azure-keyvault-certificates/azure/keyvault/certificates/_generated/v7_2/models/_models_py3.py#L240) should always be a string, so should this be `Optional[str]`?

sdk/keyvault/azure-keyvault-certificates/azure/keyvault/certificates/_models.py
```diff
@@ -513,104 +513,104 @@ def _from_certificate_operation_bundle(cls, certificate_operation_bundle):
...
+        return self._vault_id.vault_url if self._vault_id else None
 
     @property
     def issuer_name(self):
-        # type: () -> str
+        # type: () -> Union[str, WellKnownIssuerNames, None]
         """The name of the issuer of the certificate.
 
-        :rtype: str
+        :rtype: str or WellKnownIssuerNames or None
```
mccoyp, line 540: I would use the fully qualified name of WellKnownIssuerNames here: `~azure.keyvault.certificates.WellKnownIssuerNames` ([example](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_models.py#L384) from `azure-keyvault-keys`). For the type hint on line 537, we can probably also do `Union[WellKnownIssuerNames, None]` because WellKnownIssuerNames is a string enum, and that lets the user know that the value will only be from that enum's set of values.

sdk/keyvault/azure-keyvault-certificates/azure/keyvault/certificates/_models.py
```diff
@@ -768,12 +770,13 @@ def _to_certificate_policy_bundle(self):
...
                 curve=self.key_curve_name,
-            )
+            )  # type: Optional[models.KeyProperties]
         else:
             key_properties = None
 
         if self.content_type:
-            secret_properties = models.SecretProperties(content_type=self.content_type)
+            secret_properties = models.SecretProperties(
+                content_type=self.content_type)  # type: Optional[models.SecretProperties]
```
mccoyp, line 779: nit: I think the parenthesis after `self.content_type` should be on the next line -- did you try running `black` on this file?

sdk/keyvault/azure-keyvault-certificates/azure/keyvault/certificates/_models.py
```diff
@@ -789,34 +792,38 @@ def _to_certificate_policy_bundle(self):
 
     @classmethod
     def _from_certificate_policy_bundle(cls, certificate_policy_bundle):
-        # type: (models.CertificatePolicy) -> CertificatePolicy
+        # type: (Optional[models.CertificatePolicy]) -> CertificatePolicy
         """Construct a CertificatePolicy from an autorest-generated CertificatePolicy"""
-        if certificate_policy_bundle.lifetime_actions:
+        if certificate_policy_bundle and certificate_policy_bundle.lifetime_actions:
```
mccoyp, line 797: Instead of checking if the `certificate_policy_bundle` is none throughout the method, it would probably be easier to, at the start of the method, return a `cls()` if `certificate_policy_bundle is None` since all fields will be None by default. Then the rest of the method logic doesn't need to be changed, apart from maybe a mypy comment if there's still a warning.

sdk/keyvault/azure-keyvault-certificates/azure/keyvault/certificates/_models.py
```diff
@@ -1283,8 +1291,8 @@ class LifetimeAction(object):
         should be between 1 and 972 (36 * 27).
     """
 
-    def __init__(self, action, lifetime_percentage=None, days_before_expiry=None):
-        # type: (CertificatePolicyAction, Optional[int], Optional[int]) -> None
+    def __init__(self, action=None, lifetime_percentage=None, days_before_expiry=None):
```
mccoyp, line 1294: I would keep the type hint you wrote for this method on the next line, but we probably shouldn't give `action` a default value of `None` even if `None` can be provided. We want to make sure an `action` (`None` or otherwise) still needs to be provided when creating a LifetimeAction.

sdk/keyvault/azure-keyvault-certificates/azure/keyvault/certificates/_models.py
YalinLi0312: Nope. But I see some other places are putting end parenthesis in a new line, so will follow with others.

sdk/keyvault/azure-keyvault-certificates/azure/keyvault/certificates/_models.py
YalinLi0312: I'm not sure how the `action` works at this stage, I'll leave this question to @chlowell .

sdk/keyvault/azure-keyvault-certificates/azure/keyvault/certificates/_models.py
YalinLi0312: Good point! I didn't check CertificateBundle, in this case I think `Optional[str]` would make more sense.

sdk/keyvault/azure-keyvault-certificates/azure/keyvault/certificates/_models.py
YalinLi0312: I think we should keep `str` in Union, because on line496 the right side of "=" is type of Optional[str].

sdk/keyvault/azure-keyvault-certificates/azure/keyvault/certificates/_models.py
chlowell: I don't recall the history of `action` but in general we believed required positional parameters really are required, either by Key Vault or for the client to construct a model. Regardless, we shouldn't change the API to solve a type checking problem. Whether a parameter is required (by Python) is unrelated to the type of its value.

sdk/keyvault/azure-keyvault-certificates/azure/keyvault/certificates/_models.py
```diff
@@ -1283,8 +1292,8 @@ class LifetimeAction(object):
...
```
mccoyp, line 1295: The PR looks good to me if `action` is made a required parameter here 👍

mccoyp (approved):

PR #20537, opened 2021-09-03
[Batch] Redact test account keys in recordings

mccoyp:
This resolves [CredScan warnings](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=1078334&view=logs&j=0874ffeb-2919-5b4b-20de-16a97970b94c&t=439e917e-962a-53e0-8a4b-b8bb4f8ed372&l=56) raised by test account keys that are present in recordings. To redact these values while preserving test functionality (ensuring that the keys are unique), these keys are redacted with a hash of their original value.


maririos (approved):
general logic LGTM. Not sure Python implementation :)

msyyc (approved):

PR #20543, opened 2021-09-04
[Identity] Resolve CredScan warnings

mccoyp:
Resolves the warnings described in https://github.com/Azure/azure-sdk-for-python-pr/issues/657.

This moves test PowerShell outputs into a newly-suppressed helper file, and redacts test tokens from existing CAE test recordings (using the [RecordingRedactor logic](https://github.com/Azure/azure-sdk-for-python/blob/a6cfc776292ef9f8841e5ea3dc68c8fff4ae7ee8/sdk/identity/azure-identity/tests/recording_processors.py#L58) that was added to CaeTestCase after their introduction to the repo).


xiangyan99 (approved):

eng/CredScanSuppression.json
```diff
@@ -40,9 +40,10 @@
                 "eng/common/testproxy/dotnet-devcert.pfx",
                 "sdk/confidentialledger/azure-confidentialledger/tests/_shared/constants.py",
                 "sdk/keyvault/azure-keyvault-certificates/tests/ca.key",
-                "sdk/identity/azure-identity/tests/ec-certificate.pem",
                 "sdk/identity/azure-identity/tests/certificate.pfx",
                 "sdk/identity/azure-identity/tests/certificate-with-password.pfx",
+                "sdk/identity/azure-identity/tests/ec-certificate.pem",
+                "sdk/identity/azure-identity/tests/helpers.py",
```
weshaggard, line 46: Can we put these into a file that is only intended for these fake secrets and is named as such. Right now by blindly adding this helpers.py file to the suppression list we risk another potential real leak in that file because it is used for other things beyond fake secrets.

mccoyp: Sure, I can split this into a `fake_credentials` (or similar) file, that seems reasonable.

weshaggard (approved):

sdk/identity/azure-identity/tests/fake_credentials.py
```diff
@@ -0,0 +1,10 @@
+# ------------------------------------
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+# ------------------------------------
+
+FAKE_CLIENT_ID = "fake-client-id"
```
chlowell, line 6: Was this triggering warnings (surprising because client IDs are not secret)?

mccoyp: This wasn't, but I felt like a test client ID belonged more in this file than in `helpers.py` (since it's a component of constructing a credential) so I brought it over. I'm not strongly opposed at all to moving this back to the helpers file though!

chlowell: It's only pedantry making me write that it doesn't fit here. While I'm being pedantic though, nothing in this file is a credential, so this is more like `credscan_ignore.py`, and this constant still doesn't fit because credscan ignores it already. But I leave it up to you, I'm only driving by 💨

mccoyp: Yeah, the naming is really awkward because you're right, nothing here is a "credential". The contents of the PowerShell strings just raise CredScan flags because of the contained token references. Your suggestion seems reasonable considering this file is more about satisfying CredScan than storing fake credentials centrally.

PR #20545, opened 2021-09-04
Run mypy in azure-keyvault-keys CI

YalinLi0312:
This PR is for part of issue #19158, it fixed all errors in azure-keyvault-keys package.


sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/crypto/_client.py
```diff
@@ -112,34 +112,34 @@ def __init__(self, key, credential, **kwargs):
...
             self._key_id = parse_key_vault_id(key)
-            self._keys_get_forbidden = None  # type: Optional[bool]
+            self._keys_get_forbidden = False
         elif self._jwk:
             self._key = key
         else:
             raise ValueError("'key' must be a KeyVaultKey instance or a key ID string including a version")
 
-        if not (self._jwk or self._key_id.version):
+        if not (self._jwk or self._key_id.version if self._key_id else None):
```
mccoyp, line 129: To disambiguate the statement and fix pipeline errors:
```suggestion
        if not (self._jwk or (self._key_id.version if self._key_id else None)):
```

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/crypto/_client.py
```diff
@@ -112,34 +112,34 @@ def __init__(self, key, credential, **kwargs):
...
+                self._local_provider = get_local_cryptography_provider(cast(JsonWebKey, self._key))
                 self._initialized = True
             except Exception as ex:  # pylint:disable=broad-except
                 six.raise_from(ValueError("The provided jwk is not valid for local cryptography"), ex)
         else:
             self._local_provider = NoLocalCryptography()
             self._initialized = False
 
-        self._vault_url = None if self._jwk else self._key_id.vault_url
+        self._vault_url = None if self._jwk else self._key_id.vault_url if self._key_id else None  # type: ignore
```
mccoyp, line 142: nit: since we make `self._vault_url` in multiple cases here, I think reordering the evaluation a bit makes things clearer:
```suggestion
        self._vault_url = None if (self._jwk or self._key_id is None) else self._key_id.vault_url  # type: ignore
```

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/crypto/_providers/local_provider.py
```diff
@@ -64,34 +64,37 @@ def encrypt(self, algorithm, plaintext, iv=None):
         # type: (EncryptionAlgorithm, bytes, Optional[bytes]) -> EncryptResult
         self._raise_if_unsupported(KeyOperation.encrypt, algorithm)
         ciphertext = self._internal_key.encrypt(plaintext, algorithm=algorithm.value, iv=iv)
-        return EncryptResult(key_id=self._key.kid, algorithm=algorithm, ciphertext=ciphertext, iv=iv)
+        return EncryptResult(
+            key_id=self._key.kid, algorithm=algorithm, ciphertext=ciphertext, iv=iv)  # type: ignore[attr-defined]
```
mccoyp, line 68: nit: for each of these, we should put the trailing parenthesis on its own line
```suggestion
        return EncryptResult(
            key_id=self._key.kid, algorithm=algorithm, ciphertext=ciphertext, iv=iv
        )  # type: ignore[attr-defined]
```

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/crypto/aio/_client.py
```diff
@@ -61,34 +61,35 @@ def __init__(self, key: "Union[KeyVaultKey, str]", credential: "AsyncTokenCreden
...
+                self._local_provider = get_local_cryptography_provider(cast(JsonWebKey, self._key))
                 self._initialized = True
             except Exception as ex:  # pylint:disable=broad-except
                 raise ValueError("The provided jwk is not valid for local cryptography") from ex
         else:
             self._local_provider = NoLocalCryptography()
             self._initialized = False
 
-        self._vault_url = None if self._jwk else self._key_id.vault_url
+        self._vault_url = None if self._jwk else self._key_id.vault_url  # type: ignore
```
mccoyp, line 92: This line should match the sync client's, and other comments I made apply here (just emphasizing since I've forgotten to carry changes over to async clients before 😅)

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/crypto/_models.py
```diff
@@ -18,11 +18,11 @@ class DecryptResult:
     :param bytes plaintext: The decrypted bytes
     """
 
-    def __init__(self, key_id, algorithm, plaintext):
-        # type: (str, EncryptionAlgorithm, bytes) -> None
-        self.key_id = key_id
+    def __init__(self, algorithm, plaintext, key_id=None):
+        # type: (EncryptionAlgorithm, bytes, Optional[str]) -> None
```
mccoyp, line 22: For each of these types we should avoid changing the signature and just update type hints (as much as we both dislike accepting `None` for a required parameter):
```suggestion
    def __init__(self, key_id, algorithm, plaintext):
        # type: (Optional[str], EncryptionAlgorithm, bytes) -> None
```

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/crypto/aio/_client.py
YalinLi0312: Good reminder!

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/crypto/aio/_client.py
```diff
@@ -61,34 +61,35 @@ def __init__(self, key: "Union[KeyVaultKey, str]", credential: "AsyncTokenCreden
...
         elif self._jwk:
             self._key = key
         else:
             raise ValueError("'key' must be a KeyVaultKey instance or a key ID string including a version")
 
-        if not (self._jwk or self._key_id.version):
+        if not (self._jwk or (self._key_id.version if self._key_id else None)):
             raise ValueError("'key' must include a version")
 
+        self._initialized = None  # type: Optional[bool]
```
mccoyp, line 81: We can remove this line now (it's already gone in the synchronous client)
```suggestion
```

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/crypto/_client.py
```diff
@@ -311,16 +314,16 @@ def decrypt(self, algorithm, ciphertext, **kwargs):
...
+            key_name=self._key_id.name if self._key_id else None,
+            key_version=self._key_id.version if self._key_id else None,
             parameters=self._models.KeyOperationsParameters(
                 algorithm=algorithm, value=ciphertext, iv=iv, tag=tag, aad=aad
             ),
             **kwargs
         )
 
-        return DecryptResult(key_id=self.key_id, algorithm=algorithm, plaintext=operation_result.result)
+        return DecryptResult(algorithm=algorithm, plaintext=operation_result.result, key_id=self.key_id)
```
mccoyp, line 326: These parameters can be returned to their original order in methods returning `*Result`s here and in the async client

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/crypto/_providers/local_provider.py
```diff
@@ -64,34 +64,40 @@ def encrypt(self, algorithm, plaintext, iv=None):
         # type: (EncryptionAlgorithm, bytes, Optional[bytes]) -> EncryptResult
         self._raise_if_unsupported(KeyOperation.encrypt, algorithm)
         ciphertext = self._internal_key.encrypt(plaintext, algorithm=algorithm.value, iv=iv)
-        return EncryptResult(key_id=self._key.kid, algorithm=algorithm, ciphertext=ciphertext, iv=iv)
+        return EncryptResult(
+            key_id=self._key.kid, algorithm=algorithm, ciphertext=ciphertext, iv=iv
+        )  # type: ignore[attr-defined]
```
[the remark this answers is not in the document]
YalinLi0312: Right, I forgot to rerun mypy before updates.

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/crypto/_client.py
YalinLi0312: Also, have reverted the order in other *Results in this file and async one.

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/crypto/_client.py
```diff
@@ -440,14 +443,14 @@ def sign(self, algorithm, digest, **kwargs):
...
-            key_version=self._key_id.version,
+            vault_base_url=self._key_id.vault_url if self._key_id else None,
+            key_name=self._key_id.name if self._key_id else None,
+            key_version=self._key_id.version if self._key_id else None,
             parameters=self._models.KeySignParameters(algorithm=algorithm, value=digest),
             **kwargs
         )
 
-        return SignResult(key_id=self.key_id, algorithm=algorithm, signature=operation_result.result)
+        return SignResult(algorithm=algorithm, signature=operation_result.result, key_id=self.key_id)
```
mccoyp, line 453:
```suggestion
        return SignResult(key_id=self.key_id, algorithm=algorithm, signature=operation_result.result)
```

YalinLi0312: Good catch!

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/crypto/aio/_client.py
```diff
@@ -338,9 +341,9 @@ async def unwrap_key(self, algorithm: "KeyWrapAlgorithm", encrypted_key: bytes,
...
         operation_result = await self._client.unwrap_key(
-            vault_base_url=self._key_id.vault_url,
-            key_name=self._key_id.name,
-            key_version=self._key_id.version,
+            vault_base_url=self._key_id.vault_url if self._key_id else None,
+            key_name=self._key_id.name if self._key_id else None,
+            key_version=self._key_id.version if self._key_id else None,
             parameters=self._models.KeyOperationsParameters(algorithm=algorithm, value=encrypted_key),
             **kwargs
         )
```
mccoyp, line 349: The change you had made on line 351 (`key_id=self._key_id` --> `key_id=self.key_id`) was actually a really good catch that I had missed before -- if we reference `self._key_id` directly instead of the `self.key_id` property, the `key_id` we pass in will be `None` in some cases [when we might not want it to be](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/crypto/aio/_client.py#L104). If you re-introduce that change we should be good to get this merged 🥳

mccoyp (approved):

PR #20547, opened 2021-09-04
Smoke test package verification

ckairen:
Making sure we are checking the correct package for smoke test.

Pipeline Ref: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=1081799&view=logs&j=811de798-cac6-5a82-8dac-ad597ae5b1be&t=0737808f-afb0-5c24-e7c8-8a6b739b1c03&l=162

### DIFF
--- a/eng/pipelines/templates/jobs/smoke.tests.yml
+++ b/eng/pipelines/templates/jobs/smoke.tests.yml
@@ -167,7 +167,7 @@ jobs:
                 return @{ "name" = $matches[1]; "version" = $matches[2] }
               }
             }
-            if (!($artifacts.name -contains "${{parameters.Artifact.name}}")) {
+            if ($artifacts.name -notcontains "${{parameters.Artifact.name}}") {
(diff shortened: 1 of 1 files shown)


eng/pipelines/templates/jobs/smoke.tests.yml
```diff
@@ -167,6 +167,10 @@ jobs:
                 return @{ "name" = $matches[1]; "version" = $matches[2] }
               }
             }
+            if (!($artifacts.name -contains "${{parameters.Artifact.name}}")) {
```
weshaggard, line 170:
```suggestion
            if ($artifacts.name -notcontains "${{parameters.Artifact.name}}") {
```

weshaggard (approved):
Assuming in all cases we only have one artifact this should work.

ckairen:
Yup there's always going to be one artifact, that was done when smoke test got isolated per package

PR #20559, opened 2021-09-07
[FileShare]Bug Fix

xiafu-msft:


xiafu-msft:
> should we add test for this?

done.

PR #20571, opened 2021-09-07
Allow overriding client_id for token exchange

chlowell:
ManagedIdentityCredential should use AZURE_CLIENT_ID as a default but allow overriding via `client_id`.


sdk/identity/azure-identity/azure/identity/aio/_credentials/managed_identity.py
```diff
@@ -68,7 +68,7 @@ def __init__(self, **kwargs: "Any") -> None:
 
             self._credential = TokenExchangeCredential(
                 tenant_id=os.environ[EnvironmentVariables.AZURE_TENANT_ID],
-                client_id=os.environ[EnvironmentVariables.AZURE_CLIENT_ID],
+                client_id=kwargs.pop("client_id", os.environ[EnvironmentVariables.AZURE_CLIENT_ID]),
```
xiangyan99, line 71: How is identity_config consumed?

chlowell: Currently it's ignored. The platform only accepts "client_id", so really there's no need for `identity_config`. Of course someone may use it anyway. There are some design considerations to think through before deciding how to handle it. We don't need a solution for this release, so I created #20575.

xiangyan99: Then should we remove 

```py
:keyword identity_config: a mapping ``{parameter_name: value}`` specifying a user-assigned identity by its object
        or resource ID, for example ``{"object_id": "..."}``. Check the documentation for your hosting environment to
        learn what values it expects.
:paramtype identity_config: Mapping[str, str]
```

from the docstring?

chlowell: No, all other implementations observe it.

xiangyan99 (approved):

sdk/identity/azure-identity/azure/identity/_constants.py
```diff
@@ -48,4 +48,4 @@ class EnvironmentVariables:
     AZURE_REGIONAL_AUTHORITY_NAME = "AZURE_REGIONAL_AUTHORITY_NAME"
 
     AZURE_FEDERATED_TOKEN_FILE = "AZURE_FEDERATED_TOKEN_FILE"
-    TOKEN_EXCHANGE_VARS = (AZURE_CLIENT_ID, AZURE_TENANT_ID, AZURE_FEDERATED_TOKEN_FILE)
+    TOKEN_EXCHANGE_VARS = (AZURE_AUTHORITY_HOST, AZURE_TENANT_ID, AZURE_FEDERATED_TOKEN_FILE)
```
xiangyan99, line 51: Is this by design?

I see we have code:

```py
        elif all(os.environ.get(var) for var in EnvironmentVariables.TOKEN_EXCHANGE_VARS):
            _LOGGER.info("%s will use token exchange", self.__class__.__name__)
            from .token_exchange import TokenExchangeCredential

            self._credential = TokenExchangeCredential(
                tenant_id=os.environ[EnvironmentVariables.AZURE_TENANT_ID],
                client_id=os.environ[EnvironmentVariables.AZURE_CLIENT_ID],
                token_file_path=os.environ[EnvironmentVariables.TOKEN_FILE_PATH],
                **kwargs
            )
```

chlowell: Yes. I've updated that code as well, did you copy it from an old revision?

The gist is, AZURE_AUTHORITY_HOST, AZURE_TENANT_ID, AZURE_FEDERATED_TOKEN_FILE are certain to have values in this environment. AZURE_CLIENT_ID may not have a value, in which case we require a value for client_id.

xiangyan99 (approved):

PR #20577, opened 2021-09-07
[rest] correctly pickle rest aiohttp responses

iscai-msft:
Currently, pickling for `RestAioHttpTransportResponse`s are failing, because the internal response is not being correctly removed. In `RestAioHttpTransportResponse`, the internal response is a hidden property, `._internal_response`, so we were not properly removing it in getstate

### DIFF
--- a/sdk/core/azure-core/tests/async_tests/test_rest_polling_async.py
+++ b/sdk/core/azure-core/tests/async_tests/test_rest_polling_async.py
@@ -118,3 +118,16 @@ async def test_delete_operation_location(lro_poller):
 async def test_request_id(lro_poller):
     result = await (await lro_poller(HttpRequest("POST", "/polling/request-id"), request_id="123456789")).result()
     assert result['status'] == "Succeeded"
+
+@pytest.mark.asyncio
+async def test_continuation_token(client, lro_poller, deserialization_callback):
+    poller = await lro_poller(HttpRequest("POST", "/polling/post/location-and-operation-location"))
+    token = poller.continuation_token()
+    new_poller = AsyncLROPoller.from_continuation_token(
+        continuation_token=token,
+        polling_method=AsyncLROBasePolling(0),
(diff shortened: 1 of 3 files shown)


sdk/core/azure-core/tests/async_tests/test_rest_polling_async.py
```diff
@@ -118,3 +118,16 @@ async def test_delete_operation_location(lro_poller):
 async def test_request_id(lro_poller):
     result = await (await lro_poller(HttpRequest("POST", "/polling/request-id"), request_id="123456789")).result()
     assert result['status'] == "Succeeded"
+
+@pytest.mark.asyncio
+async def test_continuation_token(client, lro_poller, deserialization_callback):
```
xiangyan99, line 123: Could you tell me which scenario we want to test in these tests?

iscai-msft: we want to test that we can correctly [get a continuation token from a poller](https://docs.microsoft.com/en-us/python/api/azure-core/azure.core.polling.lropoller?view=azure-python#continuation-token--), and correctly [restart polling with that continuation token](https://docs.microsoft.com/en-us/python/api/azure-core/azure.core.polling.lropoller?view=azure-python#from-continuation-token-polling-method--continuation-token----kwargs). 

To create the continuation token, we have to pickle the response. As you can tell from the comment, and the [corresponding code](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/core/azure-core/azure/core/pipeline/transport/_aiohttp.py#L354) in `azure.core.pipeline.transport.AioHttpTransportResponse`, we have to get rid of parts of the response to pickle it. My code is copied from `azure.core.pipeline.transport.AioHttpTransportResponse`, but because my internal response is a private property, I have to update the code to reflect that

annatisch (approved):

Issue #20487, opened 2021-08-31
Add support to configure Interactive Browser Client Id on Default Azure Credential 


g2vinay:
Allow the user to configure interactive browser client id in Default Azure Credential similar to other client ids.

PR #20591, opened 2021-09-08
add interactive_browser_client_id

xiangyan99:


sdk/identity/azure-identity/azure/identity/_credentials/default.py
```diff
@@ -102,6 +104,7 @@ def __init__(self, **kwargs):
         managed_identity_client_id = kwargs.pop(
             "managed_identity_client_id", os.environ.get(EnvironmentVariables.AZURE_CLIENT_ID)
         )
+        interactive_browser_client_id = kwargs.pop("interactive_browser_client_id")
```
chlowell, line 107: `pop` with no default raises when the key doesn't exist
```suggestion
        interactive_browser_client_id = kwargs.pop("interactive_browser_client_id", None)
```

sdk/identity/azure-identity/azure/identity/_credentials/default.py
```diff
@@ -70,6 +70,8 @@ class DefaultAzureCredential(ChainedTokenCredential):
         AZURE_TENANT_ID, if any. If unspecified, users will authenticate in their home tenants.
     :keyword str managed_identity_client_id: The client ID of a user-assigned managed identity. Defaults to the value
         of the environment variable AZURE_CLIENT_ID, if any. If not specified, a system-assigned identity will be used.
+    :keyword str interactive_browser_client_id: The client ID to be used in interactive browser credential. If not
+        specified, a system-assigned identity will be used.
```
chlowell, line 74: System and user assigned identities apply only to managed identity
```suggestion
        specified, users will authenticate to an Azure development application.
```

sdk/identity/azure-identity/azure/identity/_credentials/default.py
```diff
@@ -137,7 +140,13 @@ def __init__(self, **kwargs):
         if not exclude_powershell_credential:
             credentials.append(AzurePowerShellCredential(**kwargs))
         if not exclude_interactive_browser_credential:
-            credentials.append(InteractiveBrowserCredential(tenant_id=interactive_browser_tenant_id, **kwargs))
+            if interactive_browser_client_id:
+                credentials.append(InteractiveBrowserCredential(
+                    tenant_id=interactive_browser_tenant_id,
+                    client_id=interactive_browser_client_id,
```
chlowell, line 146: InteractiveBrowserCredential expects this argument to be set only with a valid client ID (i.e. a string). So, we should pass a value here only when interactive_browser_client_id has a string value.

sdk/identity/azure-identity/tests/test_default.py
```diff
@@ -387,6 +387,21 @@ def validate_tenant_id(credential):
...
 
+def test_interactive_browser_client_id():
+    """the credential should allow configuring a client ID for InteractiveBrowserCredential by kwarg"""
+
+    client_id = "client-id"
+
+    def validate_client_id(credential):
+        assert len(credential.call_args_list) == 1, "InteractiveBrowserCredential should be instantiated once"
+        _, kwargs = credential.call_args
+        assert kwargs == {"client_id": client_id, "tenant_id": "72f988bf-86f1-41af-91ab-2d7cd011db47"}
```
chlowell, line 398: Suggest validating only that you got the expected client_id, because tenant_id is irrelevant to this feature and varies by environment configuration

sdk/identity/azure-identity/azure/identity/_credentials/default.py
xiangyan99: I don't see we do similar validation for other client ids. Should we keep consistent with others?

chlowell: `InteractiveBrowserCredential(client_id=None, ...)` creates a credential that doesn't work. `ManagedIdentityCredential(client_id=None)` creates a credential that could work, depending on the environment. Maybe both credentials should be consistent in how they handle `client_id=None` but they aren't today. (For what it's worth, I think InteractiveBrowserCredential is more correct here.)

sdk/identity/azure-identity/azure/identity/_credentials/default.py
```diff
@@ -102,6 +105,9 @@ def __init__(self, **kwargs):
         managed_identity_client_id = kwargs.pop(
             "managed_identity_client_id", os.environ.get(EnvironmentVariables.AZURE_CLIENT_ID)
         )
+        interactive_browser_client_id = kwargs.pop("interactive_browser_client_id", None)
+        if not isinstance(interactive_browser_client_id, six.string_types):
+            raise ValueError('"interactive_browser_client_id" must be a string')
```
chlowell, line 110: But the variable defaults to `None`, causing `DefaultAzureCredential()` to raise this error.

xiangyan99: Hah, Good catch!.

chlowell: How about:
```py

browser_args = dict(kwargs, tenant_id=interactive_browser_tenant_id)
interactive_browser_client_id = kwargs.pop("interactive_browser_client_id", None)
if interactive_browser_client_id:
    browser_args["client_id"] = interactive_browser_client_id
...
if not exclude_interactive_browser_credential:
    credentials.append(InteractiveBrowserCredential(**browser_args))
```

xiangyan99: We want to pass in the client_id after validating client_id is a string, right?

Are you proposing we doing it in ctor of InteractiveBrowserCredential?

chlowell: We want to pass client_id if and only if the caller gave us a value for interactive_browser_client_id. My code snippet is one way to achieve that. Only a string will work for client_id but in general we don't validate types and there's no special need to do so here.

xiangyan99: I think we are on same page and I guess it is already addressed?

(please refresh the page, I moved the validation into 

```py
if not exclude_interactive_browser_credential:
    if interactive_browser_client_id:
```
)

chlowell: Yes, that's another way to do it. No need for the `isinstance` though.

xiangyan99:
#20487

sdk/identity/azure-identity/azure/identity/_credentials/default.py
```diff
@@ -4,6 +4,7 @@
 # ------------------------------------
 import logging
 import os
+import six
```
chlowell, line 7: This is unused now

chlowell (approved):

PR #20592, opened 2021-09-08
Update docstrings for custom models.

tjprescott:
Updates our custom models to use `:keyword` annotations instead of `:param`. This causes them to be displayed correctly in APIView.

### DIFF
--- a/sdk/search/azure-search-documents/azure/search/documents/indexes/models/_index.py
+++ b/sdk/search/azure-search-documents/azure/search/documents/indexes/models/_index.py
@@ -31,26 +31,26 @@ class SearchField(msrest.serialization.Model):
 
     All required parameters must be populated in order to send to Azure.
 
-    :param name: Required. The name of the field, which must be unique within the fields collection
+    :keyword name: Required. The name of the field, which must be unique within the fields collection
      of the index or parent field.
-    :type name: str
-    :param type: Required. The data type of the field. Possible values include: "Edm.String",
+    :paramtype name: str
+    :keyword type: Required. The data type of the field. Possible values include: "Edm.String",
      "Edm.Int32", "Edm.Int64", "Edm.Double", "Edm.Boolean", "Edm.DateTimeOffset",
      "Edm.GeographyPoint", "Edm.ComplexType".
-    :type type: str or ~azure.search.documents.indexes.models.SearchFieldDataType
-    :param key: A value indicating whether the field uniquely identifies documents in the index.
+    :paramtype type: str or ~azure.search.documents.indexes.models.SearchFieldDataType
+    :keyword key: A value indicating whether the field uniquely identifies documents in the index.
      Exactly one top-level field in each index must be chosen as the key field and it must be of
      type Edm.String. Key fields can be used to look up documents directly and update or delete
      specific documents. Default is false for simple fields and null for complex fields.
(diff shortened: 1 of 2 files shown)


sdk/search/azure-search-documents/azure/search/documents/indexes/models/_index.py
```diff
@@ -31,26 +31,26 @@ class SearchField(msrest.serialization.Model):
 
     All required parameters must be populated in order to send to Azure.
 
-    :param name: Required. The name of the field, which must be unique within the fields collection
+    :keyword name: Required. The name of the field, which must be unique within the fields collection
```
xiangyan99, line 34: I think we should use "param" rather than "keyword" for required parameters?

xiangyan99: From guideline:

```py
def get(*args, **kwargs):
    """Calls `request` with the method "GET" and forwards all other arguments.

    :param str method-param: The method-param parameter
    :keyword int method-kwarg: The optional method-kwarg parameter

    For additional request configuration options, please see https://aka.ms/azsdk/python/options.
    """
    return request("GET", *args, **kwargs)
```


https://azure.github.io/azure-sdk/python_documentation.html

tjprescott: https://github.com/Azure/azure-sdk-for-python/blob/main/doc/dev/docstring_typehint.md#model-and-client-docstrings

I don't think the distinction between `:param` and `:keyword` is whether the argument is required or not, but rather whether it is positional or keyword-only, and in this case (all cases in this PR) they are keyword-only.

@johanste @annatisch can you comment here?

If we leave these as `:param`, APIView will _not_ display them.

tjprescott:
Here is the APIView that shows this change produces the correct APIView:
https://apiview.dev/Assemblies/Review/8ba6867a14c142b296bc2399e0ee36b4?diffRevisionId=470b5dcf6ec84b97bafa8445f14e6382&doc=False&diffOnly=False

Also, I have verified that this change also does not affect the display of generated docs.

tjprescott:
This is also implementing the solution being pursued in Autorest: https://github.com/Azure/autorest.python/pull/1026

tjprescott:
@xiangyan99 this is also the same logic that was applied by the latest autorest in https://github.com/Azure/azure-sdk-for-python/pull/20597

xiangyan99 (approved):

xiangyan99:
> @xiangyan99 this is also the same logic that was applied by the latest autorest in #20597

Could you give me a link where in #20597 that use keyword for required parameters?

tjprescott:
@xiangyan99 here's the autorest PR that made the change: https://github.com/Azure/autorest.python/pull/1026/files

If you look in `model_base_serializer.py` you can see it was basically a simply replacement. There's no logic based on whether something is required or not.

sdk/search/azure-search-documents/azure/search/documents/indexes/models/_index.py
johanste: Correct -  there is nothing that prevents keyword-only required parameters (although we generally try to avoid them in convenience layers if possible)

Issue #20524, opened 2021-09-02
[Search] Changes to SearchClient.search


Mohit-Chakraborty:
![image](https://user-images.githubusercontent.com/8271806/131925115-c22e4d9b-3a35-42af-a26a-5c3029b14523.png)

tjprescott:
Summary of changes needed for Python:
- [ ] Change `speller` to `query_speller`
- [ ] Construct `answers` property from two custom properties `query_answer` and `query_answer_count`
- [ ] Construct `captions` property from two custom properties `query_caption` and `query_caption_highlight`.

PR #20602, opened 2021-09-08
[Search] Update SearchClient.search API

tjprescott:
Closes https://github.com/Azure/azure-sdk-for-python/issues/20524

### DIFF
--- a/sdk/search/azure-search-documents/azure/search/documents/_search_client.py
+++ b/sdk/search/azure-search-documents/azure/search/documents/_search_client.py
@@ -277,11 +277,11 @@ class SearchClient(HeadersMixin):
 
         query_answer = kwargs.pop("query_answer", None)
         query_answer_count = kwargs.pop("query_answer_count", None)
-        answers = '{}|count-{}'.format(query_answer, query_answer_count)
+        answers = query_answer if not query_answer_count else '{}|count-{}'.format(query_answer, query_answer_count)
 
         query_caption = kwargs.pop("query_caption", None)
         query_caption_highlight = kwargs.pop("query_caption_highlight", None)
-        captions = '{}|highlight-{}'.format(query_caption, query_caption_highlight)
+        captions = query_caption if not query_caption_highlight else '{}|highlight-{}'.format(query_caption, query_caption_highlight)
 
         semantic_fields = kwargs.pop("semantic_fields", None)
 
--- a/sdk/search/azure-search-documents/azure/search/documents/_search_client.py
+++ b/sdk/search/azure-search-documents/azure/search/documents/_search_client.py
@@ -200,12 +200,23 @@ class SearchClient(HeadersMixin):
         :keyword query_language: A value that specifies the language of the search query. Possible values
          include: "none", "en-us".
         :paramtype query_language: str or ~azure.search.documents.models.QueryLanguage
-        :keyword speller: A value that specified the type of the speller to use to spell-correct
+        :keyword query_speller: A value that specified the type of the speller to use to spell-correct
          individual search query terms. Possible values include: "none", "lexicon".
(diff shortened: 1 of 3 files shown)


sdk/search/azure-search-documents/azure/search/documents/_search_client.py
```diff
@@ -259,11 +270,21 @@ def search(self, search_text, **kwargs):
...
-        speller = kwargs.pop("speller", None)
-        answers = kwargs.pop("answers", None)
+        query_speller = kwargs.pop("query_speller", None)
         select = kwargs.pop("select", None)
         skip = kwargs.pop("skip", None)
         top = kwargs.pop("top", None)
+
+        query_answer = kwargs.pop("query_answer", None)
+        query_answer_count = kwargs.pop("query_answer_count", None)
+        answers = '{}|count-{}'.format(query_answer, query_answer_count)
```
xiangyan99, line 280: We only have this when query_answer_count  is not None, right?

Or we should not contain "|count-" in answers.

sdk/search/azure-search-documents/azure/search/documents/_search_client.py
```diff
@@ -259,11 +270,21 @@ def search(self, search_text, **kwargs):
...
+
+        query_caption = kwargs.pop("query_caption", None)
+        query_caption_highlight = kwargs.pop("query_caption_highlight", None)
+        captions = '{}|highlight-{}'.format(query_caption, query_caption_highlight)
```
xiangyan99, line 284: Same.

sdk/search/azure-search-documents/azure/search/documents/_search_client.py
tjprescott: I pushed a change and verified it works.

sdk/search/azure-search-documents/azure/search/documents/_search_client.py
```diff
@@ -259,11 +270,21 @@ def search(self, search_text, **kwargs):
...
-        speller = kwargs.pop("speller", None)
-        answers = kwargs.pop("answers", None)
+        query_speller = kwargs.pop("query_speller", None)
         select = kwargs.pop("select", None)
         skip = kwargs.pop("skip", None)
         top = kwargs.pop("top", None)
+
+        query_answer = kwargs.pop("query_answer", None)
+        query_answer_count = kwargs.pop("query_answer_count", None)
+        answers = query_answer if not query_answer_count else '{}|count-{}'.format(query_answer, query_answer_count)
```
xiangyan99, line 280: Is it valid with query_answer = "None" and query_answer_count = 3?

tjprescott: No, you will get a service error InvalidRequestParameters.

xiangyan99: Then should we add into the docstring it only works when query_answer is "extractive"?

xiangyan99: (I see you have it for query_caption_highlight)

tjprescott: Fixed.

tjprescott:
Here's the APIView diff (yay it shows properties!)
https://apiview.dev/Assemblies/Review/8ba6867a14c142b296bc2399e0ee36b4?diffRevisionId=470b5dcf6ec84b97bafa8445f14e6382&doc=False&diffOnly=True

xiangyan99 (approved):
Do we need tests?

tjprescott:
There were no tests for semantic search beforehand, and since the account has to be whitelisted, there's no way to write a test without hard-coding resources, which is not good.

PR #20603, opened 2021-09-08
[Tables] Add support bug fix opt-out

annatisch:
The bug fix for character escaping in partition and row keys has the potential to be breaking if anyone was taking a dependency on the broken behaviour. So we have exposed the escaping mechanism to allow it to be patched - letting people revert back to the previous behaviour.


sdk/tables/azure-data-tables/azure/data/tables/_table_client.py
```diff
@@ -84,6 +84,7 @@ def __init__(  # pylint: disable=missing-client-constructor-parameter-credential
             raise ValueError("Please specify a table name.")
         _validate_table_name(table_name)
         self.table_name = table_name
+        self.prepare_key = _prepare_key
```
YalinLi0312, line 87: Shouldn't the value be `None or _prepare_key`? Otherwise, why you are not using _prepare_key directly.

annatisch: By assigning the function to a public attribute on the client - a user can easily patch the behaviour to revert it, as is demonstrated in the tests:
```python
# This will replace the escaping function with a no-op, essentially reverting the fix.
client.prepare_key = lambda key: key
```

PR #20623, opened 2021-09-09
Add 3.10 to Nightly Edge Build

scbedd:
Missed the PR on this.

@mccoyp Take a look? I'm only concerned about the pytest upgrade, but I think it will be fine. Just gotta run storage and core most likely.


mccoyp:
Currently running pipelines with the pytest bump on https://github.com/Azure/azure-sdk-for-python/pull/20662

praveenkuttappan (approved):

PR #20658, opened 2021-09-12
Handle Errors and Tables in Query

rakshith91:
- [x] Introduce QueryPartialErrorException
- [x] Throw errors as discussed here https://gist.github.com/rakshith91/d08caf649feac0081703d4c1e5850326
- [x] Add iterator to the LogsQueryResults to iterate over tables.

### DIFF
--- a/sdk/monitor/azure-monitor-query/CHANGELOG.md
+++ b/sdk/monitor/azure-monitor-query/CHANGELOG.md
@@ -4,7 +4,13 @@
 
 ### Features Added
 
+- Added `QueryPartialErrorException` and `LogsQueryError` to handle errors.
+- Added `partial_error` and `is_error` attributes to `LogsQueryResult`.
+- Added an option `allow_partial_errors` that defaults to False, which can be set to not throw if there are any partial errors.
+
 ### Breaking Changes
+
+- `LogsQueryResult` now iterates over the tables directly as a convinience.
 
 ### Bugs Fixed
 
--- a/sdk/monitor/azure-monitor-query/azure/monitor/query/_exceptions.py
+++ b/sdk/monitor/azure-monitor-query/azure/monitor/query/_exceptions.py
@@ -1,0 +1,83 @@
+#
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+# --------------------------------------------------------------------------
+from azure.core.exceptions import HttpResponseError
+
+class LogsQueryError(object):
+    """The code and message for an error.
+
+    All required parameters must be populated in order to send to Azure.
+
+    :ivar code: A machine readable error code.
+    :vartype code: str
+    :ivar message: A human readable error message.
+    :vartype message: str
+    :ivar details: error details.
+    :vartype details: list[~monitor_query_client.models.ErrorDetail]
+    :ivar innererror: Inner error details if they exist.
+    :vartype innererror: ~azure.monitor.query.LogsQueryError
+    :ivar additional_properties: Additional properties that can be provided on the error info
+     object.
+    :vartype additional_properties: object
+    :ivar bool is_error: Boolean check for error item when iterating over list of
+        results. Always True for an instance of a LogsQueryError.
+    """
+    def __init__(
+        self,
+        **kwargs
+    ):
+        self.code = kwargs.get('code', None)
+        self.message = kwargs.get('message', None)
+        self.details = kwargs.get('details', None)
+        self.innererror = kwargs.get('innererror', None)
+        self.additional_properties = kwargs.get('additional_properties', None)
+        self.is_error = True
+
+    @classmethod
+    def _from_generated(cls, generated):
+        if not generated:
+            return None
+        details = None
+        if generated.details is not None:
+            details=[d.serialize() for d in generated.details]
+        return cls(
+            code=generated.code,
+            message=generated.message,
+            innererror=cls._from_generated(generated.innererror) if generated.innererror else None,
+            additional_properties=generated.additional_properties,
+            details=details,
+            is_error=True
+        )
+
+class QueryPartialErrorException(HttpResponseError):
+    """There is a partial failure in query operation. This is thrown for a single query operation
+     when allow_partial_errors is set to False.
+
+    :ivar code: A machine readable error code.
+    :vartype code: str
+    :ivar message: A human readable error message.
+    :vartype message: str
+    :ivar details: error details.
+    :vartype details: list[~monitor_query_client.models.ErrorDetail]
+    :ivar innererror: Inner error details if they exist.
+    :vartype innererror: ~azure.monitor.query.LogsQueryError
+    :ivar additional_properties: Additional properties that can be provided on the error info
+     object.
+    :vartype additional_properties: object
+    :ivar bool is_error: Boolean check for error item when iterating over list of
+        results. Always True for an instance of a LogsQueryError.
+    """
+
+    def __init__(self, **kwargs):
+        error = kwargs.pop('error', None)
+        if error:
+            self.code = error.code
+            self.message = error.message
+            self.details = [d.serialize() for d in error.details] if error.details else None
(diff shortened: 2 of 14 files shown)


sdk/monitor/azure-monitor-query/azure/monitor/query/_generated/models/_models_py3.py
```diff
@@ -10,7 +10,7 @@
 from typing import Dict, List, Optional, Union
 
 from azure.core.exceptions import HttpResponseError
-import msrest.serialization
+from
```
rakshith91, line 13:
```suggestion
import msrest.serialization
```

sdk/monitor/azure-monitor-query/CHANGELOG.md
```diff
@@ -4,8 +4,14 @@
 
 ### Features Added
 
+- Added `QueryPartialErrorException` and `LogsQueryError` to handle errors.
+- Added `partial_error` and `is_error` attributes to `LogsQueryResult`.
+- Added an option `allow_partial_errors` that defaults to False, which can be set to not throw if there are any partial errors.
+
 ### Breaking Changes
 
+- `LogsQueryResult` now iterates over the tables directly as a convinience.
```
xiangyan99, line 13:
```suggestion
- `LogsQueryResult` now iterates over the tables directly as a convenience.
```

sdk/monitor/azure-monitor-query/azure/monitor/query/_exceptions.py
```diff
@@ -0,0 +1,83 @@
+#
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+# --------------------------------------------------------------------------
+from azure.core.exceptions import HttpResponseError
+
+class LogsQueryError(object):
```
xiangyan99, line 9: Is it an AzureError?

sdk/monitor/azure-monitor-query/azure/monitor/query/_exceptions.py
```diff
@@ -0,0 +1,83 @@
...
+        return cls(
+            code=generated.code,
+            message=generated.message,
+            innererror=cls._from_generated(generated.innererror) if generated.innererror else None,
+            additional_properties=generated.additional_properties,
+            details=details,
+            is_error=True
+        )
+
+class QueryPartialErrorException(HttpResponseError):
```
xiangyan99, line 55: QueryPartialErrorException or QueryPartialError?

sdk/monitor/azure-monitor-query/azure/monitor/query/_helpers.py
```diff
@@ -41,13 +41,21 @@ def get_metrics_authentication_policy(
 
     raise TypeError("Unsupported credential")
 
-def process_error(exception):
-    raise_error = HttpResponseError
-    raise raise_error(message=exception.message, response=exception.response)
-
-def order_results(request_order, mapping, obj):
+def order_results(request_order, mapping, obj, err, allow_partial_errors=False):
```
xiangyan99, line 44: Do we want to put allow_partial_errors into kwargs?

sdk/monitor/azure-monitor-query/azure/monitor/query/_helpers.py
```diff
@@ -90,3 +98,22 @@ def native_col_type(col_type, value):
 
 def process_row(col_types, row):
     return [native_col_type(col_types[ind], val) for ind, val in enumerate(row)]
+
+def process_error(error):
```
xiangyan99, line 102: Here, error is an HttpResponseError, right?

sdk/monitor/azure-monitor-query/azure/monitor/query/_exceptions.py
rakshith91: This is always raised - hence the name. inspired from `PartialBatchErrorException` in storage

sdk/monitor/azure-monitor-query/azure/monitor/query/_helpers.py
rakshith91: normally i would say yes - but here, i don't see the kwargs expanding anytime soon - plus this is an internal method, so we can change it if we think it's necessary

sdk/monitor/azure-monitor-query/azure/monitor/query/_helpers.py
rakshith91: yes that is correct

sdk/monitor/azure-monitor-query/azure/monitor/query/_exceptions.py
rakshith91: this need not be an azure error.

sdk/monitor/azure-monitor-query/azure/monitor/query/_helpers.py
xiangyan99: Could you help me understand what is this method for?

sdk/monitor/azure-monitor-query/azure/monitor/query/_exceptions.py
```diff
@@ -0,0 +1,83 @@
...
+        details = None
+        if generated.details is not None:
+            details = [d.serialize() for d in generated.details]
+        return cls(
+            code=generated.code,
+            message=generated.message,
+            innererror=cls._from_generated(generated.innererror) if generated.innererror else None,
+            additional_properties=generated.additional_properties,
+            details=details,
+            is_error=True
```
xiangyan99, line 52: Do we need it? (is_error=True)

rakshith91: yes - we return `Union[LogsQueryResult, LogsQueryError]` for the batch api - the workflow would be 
```
if result.is_error:
   # handle logsquery erroe
elif not result.is_error:
  # hanfle log query result
```

xiangyan99: If I read correctly, we don't honor "is_error" in ctor.

```py
    def __init__(
        self,
        **kwargs
    ):
        self.code = kwargs.get('code', None)
        self.message = kwargs.get('message', None)
        self.details = kwargs.get('details', None)
        self.innererror = kwargs.get('innererror', None)
        self.additional_properties = kwargs.get('additional_properties', None)
        self.is_error = True
```

sdk/monitor/azure-monitor-query/azure/monitor/query/_helpers.py
rakshith91: actually, made some small changes to this - missed a commit earlier
this is to convert the model into the logsqueryerror intead of generated internal model.

sdk/monitor/azure-monitor-query/azure/monitor/query/_exceptions.py
rakshith91: that is correct - this is always `True` for a `LogsQueryError` and always `False` for a `LogsQueryResult` - it's very similar to what we have in text analytics against DocumentError

sdk/monitor/azure-monitor-query/azure/monitor/query/_helpers.py
```diff
@@ -41,13 +41,21 @@ def get_metrics_authentication_policy(
...
+    for item in ordered:
+        if not item.body.error:
+            results.append(obj._from_generated(item.body)) # pylint: disable=protected-access
+        else:
+            error = item.body.error
+            if allow_partial_errors and error.code == 'PartialError':
+                res = obj._from_generated(item.body) # pylint: disable=protected-access
+                res.partial_error = err._from_generated(error) # pylint: disable=protected-access
+                results.append(res)
+            else:
```
xiangyan99, line 56: Shouldn't we raise if allow_partial_errors = False and there is error?

rakshith91: no - in case of batch, we return the error object instead of the result object

of course, we raise when something fatal happens (like auth error for instance)

xiangyan99 (approved):

PR #20662, opened 2021-09-13
Bump pytest version in pipeline tests

mccoyp:


mccoyp:
Seems like there are some problems with storage blob tests, but I haven't yet figured out how/if the pytest change could be related

PR #20669, opened 2021-09-13
[Search] Add skillset validation

### LINKED ISSUE #20526: Search: resolve validation strategy for multi-version skills
The service does not fail if unsupported arguments are sent to it when creating skillsets. We should decide what degree of client-side validation is appropriate, if any, ideally aligning strategies across languages.

tjprescott:
Adds skillset versioning validation. This validation is checked just before a request would be sent, as this is the most reliable way to ensure validation occurs whether the customer supplies strongly-typed objects or a dict. 

Closes #20526

### DIFF
--- a/sdk/search/azure-search-documents/CHANGELOG.md
+++ b/sdk/search/azure-search-documents/CHANGELOG.md
@@ -11,6 +11,7 @@
 
 - Renamed `SearchClient.speller` to `SearchClient.query_speller`.
 - Removed keyword arguments from `SearchClient`: `answers` and `captions`.
+- `SentimentSkill`, `EntityRecognitionSkill`: added client-side validation to prevent sending unsupported parameters.
 
 ### Bugs Fixed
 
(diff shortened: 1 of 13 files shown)


sdk/search/azure-search-documents/CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 
 - Renamed `SearchClient.speller` to `SearchClient.query_speller`.
 - Removed keyword arguments from `SearchClient`: `answers` and `captions`.
+- `SentimentSkill`, `EntityRecognitionSkill`: added client-side validation to prevent sending unsupported parameters.
```
xiangyan99, line 14: Let's double check with @johanste because in our guidelines, we have

```py
DO NOT validate service parameters. This includes null checks, empty strings, and other common validating conditions. Let the service validate any request parameters.
```

https://azure.github.io/azure-sdk/general_implementation.html

Does this conflict against our guideline?

tjprescott: I agree @johanste should weigh in. The discussion we had with Anna was that we should validate because the service does not. Her issue was with validating in the constructor, believing we should instead validate in a way that works with dict input as well as strongly-typed models (which is what this does).

johanste: Client side validation is acceptable when it prevents silent failures. Which seem to be the case here. If the service had done proper validation (and given a reasonable error message), client side validation would not have been needed.

xiangyan99 (approved):

PR #20701, opened 2021-09-15
[Test Proxy] Make add_sanitizer a module-level method

mccoyp:
This moves `add_sanitizer` out of AzureRecordedTestCase's body, since it never needed to be there. This makes setting up sanitizers on a test file-basis much easier.

### DIFF
--- a/tools/azure-sdk-tools/devtools_testutils/azure_recorded_testcase.py
+++ b/tools/azure-sdk-tools/devtools_testutils/azure_recorded_testcase.py
@@ -37,6 +37,19 @@ if TYPE_CHECKING:
 load_dotenv(find_dotenv())
 
 
+def add_sanitizer(sanitizer, regex=None, value=None):
+    # type: (ProxyRecordingSanitizer, Optional[str], Optional[str]) -> None
+    if sanitizer == ProxyRecordingSanitizer.URI:
+        requests.post(
+            "{}/Admin/AddSanitizer".format(PROXY_URL),
+            headers={"x-abstraction-identifier": ProxyRecordingSanitizer.URI.value},
+            json={
+                "regex": regex or "[a-z]+(?=(?:-secondary)\\.(?:table|blob|queue)\\.core\\.windows\\.net)",
+                "value": value or "fakevalue"
+            },
+        )
+
+
 def is_live():
     """A module version of is_live, that could be used in pytest marker."""
     if not hasattr(is_live, "_cache"):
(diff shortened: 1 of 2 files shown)


tools/azure-sdk-tools/devtools_testutils/azure_recorded_testcase.py
```diff
@@ -37,6 +37,19 @@
 load_dotenv(find_dotenv())
 
 
+def add_sanitizer(sanitizer, regex=None, value=None):
```
scbedd, line 40: The `/Admin/AddSanitizer` response is going to be passing back an `id` in the near future. That `id` will be later used by `/Admin/RemoveSanitizer` to remove one at a time if we don't want to have to do a full reset.

Given that, it may be valuable to add the newly POSTED sanitizer to a dictionary that we can import from later.

scbedd (approved):
Left some thoughts about storing sanitizers locally, but approve nonetheless.

tools/azure-sdk-tools/devtools_testutils/azure_recorded_testcase.py
scbedd: The more I think about it, the more I think we should make the arguments `kwargs` instead of setting them exactly.

That way in each sanitizer, we can access the kwarg that we expect to be present, and throw if it's not.

I say this because right now you're only dealing with configuration that needs a `value` and a `regex`...but what happens if you want to provide a `regexGroup`?

scbedd (changes requested):
After reading your document and leaving a comment on it, I decided to change my review on this one.

tools/azure-sdk-tools/devtools_testutils/azure_recorded_testcase.py
mccoyp: That makes total sense -- I had made these optional for simplicity when it was still in `integrate-proxy`, but shifting to kwargs is definitely better. Will do!

mccoyp:
@scbedd I refactored the method to accept `value`, `regex`, and `group_for_replace` as kwargs for now, without any throwing behavior for now. Were you thinking about having the specified sanitizer throw an error if required kwargs aren't provided, rather than having defaults (e.g. throwing a ValueError if a URI sanitizer is added without a `value` or `regex`)?

I think it would be great to abstract away the regexes, at least, as much as possible -- ideally, making them completely optional. Or, at least, having an enum of common regexes that would be service or endpoint specific, maybe. That might not be necessary if there's a catch-all regex that would apply to all common endpoints in recordings

scbedd:
> Were you thinking about having the specified sanitizer throw an error if required kwargs aren't provided, rather than having defaults (e.g. throwing a ValueError if a URI sanitizer is added without a `value` or `regex`)?

I'm totally fine with any amount of `layering` of sanitizers that we want to add by default for `python`. In _general_, we definitely want to come up with a `base` set of sanitizers that could be shared _across teams_. For instance, the `subscriptionId` sanitizer. That one is a great candidate for _always_ setting URI subscriptionIds to 0s. We shouldn't have to configure that per language.

scbedd (approved):

PR #20725, opened 2021-09-16
Protobuf 3.18.0 does not support python 2.7

scbedd:
Pinning it to prevent test failures on py27.


scbedd:
@xiangyan99 `windows 2.7` has gone green 👍

danieljurek (approved):

PR #20752, opened 2021-09-18
[Key Vault] Run parameterized administration tests in live pipeline

### LINKED ISSUE #19414: [Key Vault] Run azure-keyvault-administration live pipeline tests

### LINKED ISSUE #20363: [Key Vault] Parameterize tests for administration
In a spirit similar to other KV packages, run administration tests against all API versions. This will require some refactoring of how we fetch endpoint and credential configuration in these tests.

mccoyp:
Resolves #19414 and resolves #20363.

This refactors `azure-keyvault-administration` tests to use a similar test class structure as other KV packages, and in doing so allows tests to run for each available API version in playback (with only the latest API version being used in live tests). Now that tests are able to determine at runtime if their environment is configured correctly or not, the package has been added to `tests.yml` to run administration tests in the live `python - keyvault - tests` pipeline. At the moment, all administration tests should be skipped in this pipeline except for those run in the single Managed HSM-enabled platform.


sdk/keyvault/azure-keyvault-administration/tests/test_access_control_async.py
```diff
@@ -94,7 +94,7 @@ async def test_role_definitions(self, client):
 
         async for d in client.list_role_definitions(scope):
             assert (d.id != definition.id), "the role definition should have been deleted"
-        await asyncio.sleep(30)  # additional waiting to avoid conflicts with resources in other tests
+        await asyncio.sleep(60)  # additional waiting to avoid conflicts with resources in other tests
```
benbp, line 97: Are these tests filtered to only run in live mode so we can avoid sleeps in CI?

mccoyp: That's a great point -- I should definitely do that. Thank you for pointing that out!

PR #20769, opened 2021-09-21
[Test Proxy] Update migration guide and default add_sanitizer regex

mccoyp:
This mentions that `autouse=True` should be added to session fixture declarations if users want sanitizers to be added automatically when running tests, and updates the default regex for `add_sanitizer` to correctly replace the storage account name in URLs while preserving `-secondary` suffixes

### DIFF
--- a/tools/azure-sdk-tools/devtools_testutils/azure_recorded_testcase.py
+++ b/tools/azure-sdk-tools/devtools_testutils/azure_recorded_testcase.py
@@ -52,7 +52,9 @@ def add_sanitizer(sanitizer, **kwargs):
     """
     request_args = {}
     request_args["value"] = kwargs.get("value") or "fakevalue"
-    request_args["regex"] = kwargs.get("regex") or "[a-z]+(?=(?:-secondary)\\.(?:table|blob|queue)\\.core\\.windows\\.net)"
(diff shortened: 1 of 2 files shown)


tools/azure-sdk-tools/devtools_testutils/azure_recorded_testcase.py
```diff
@@ -62,7 +64,7 @@ def add_sanitizer(sanitizer, **kwargs):
             json={
                 "regex": request_args["regex"],
                 "value": request_args["value"],
-                "groupForReplace": request_args["group_for_replace"]
+                "groupForReplace": request_args["group_for_replace"],
```
YalinLi0312, line 67: Need to remove the extra `,`

YalinLi0312 (changes requested):

tools/azure-sdk-tools/devtools_testutils/azure_recorded_testcase.py
mccoyp: This was actually added by `black` -- I think it's a style preference and shouldn't affect functionality

YalinLi0312 (approved):

PR #20770, opened 2021-09-21
[Monitor exporter] Update OT Exporter to latest spec changes

lzchen:
https://github.com/microsoft/common-schema/pull/280
https://github.com/microsoft/common-schema/pull/279
https://github.com/microsoft/common-schema/pull/282
https://github.com/microsoft/common-schema/pull/285
https://github.com/microsoft/common-schema/pull/286

Also replaces hardcoded strings with semantic convention constants from semconv package.

### DIFF
--- a/sdk/monitor/azure-monitor-opentelemetry-exporter/azure/monitor/opentelemetry/exporter/export/trace/_exporter.py
+++ b/sdk/monitor/azure-monitor-opentelemetry-exporter/azure/monitor/opentelemetry/exporter/export/trace/_exporter.py
@@ -184,7 +184,7 @@ def _convert_span_to_envelope(span: Span) -> TelemetryItem:
                         path,
                     )
                 except Exception:  # pylint: disable=broad-except
-                    logger.warning("Error while parsing url.")
+                    pass
             else:
                 envelope.tags["ai.operation.name"] = span.name
             if SpanAttributes.HTTP_STATUS_CODE in span.attributes:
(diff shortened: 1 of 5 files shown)


sdk/monitor/azure-monitor-opentelemetry-exporter/azure/monitor/opentelemetry/exporter/export/trace/_exporter.py
```diff
@@ -116,94 +118,113 @@ def _convert_span_to_envelope(span: Span) -> TelemetryItem:
...
+                    parse_url = urlparse(url)
+                    path = parse_url.path
+                    if not path:
+                        path = "/"
+                    envelope.tags["ai.operation.name"] = "{} {}".format(
+                        span.attributes[SpanAttributes.HTTP_METHOD],
+                        path,
+                    )
+                except Exception:  # pylint: disable=broad-except
+                    logger.warning("Error while parsing url.")
```
hectorhdzg, line 187: Maybe just log it?, SpanAttributes could have wrong data not sure we want to report it as warning, I guess it depends what is the difference between log and warning in this exporter

hectorhdzg (approved):
LGTM

sdk/monitor/azure-monitor-opentelemetry-exporter/azure/monitor/opentelemetry/exporter/export/trace/_exporter.py
lzchen: Good point. The alternative is just to `pass` on exception (which does nothing). I'll change it to that so the user does not get bombarded with logs.

PR #20777, opened 2021-09-21
Test on the dev version setting.

sima-zhu:


sima-zhu:
az run python - template

sima-zhu:
az pipelines build queue python - template

sima-zhu:
az pipelines runs list

PR #20812, opened 2021-09-23
[ServiceBus] Add large message support to SBAdministrationClient

yunhaoling:
addressing issue: https://github.com/Azure/azure-sdk-for-python/issues/20668
swagger update PR: https://github.com/Azure/azure-rest-api-specs/pull/16125


sdk/servicebus/azure-servicebus/tests/async_tests/mgmt_tests/test_mgmt_queues_async.py
```diff
@@ -260,6 +260,13 @@ async def test_async_mgmt_queue_create_with_queue_description(self, servicebus_n
             max_size_in_megabytes=3072,
             requires_session=True
         )
+
+        with pytest.raises(HttpResponseError):
```
swathipil, line 264: is this also testing that `max_message_size_in_kilobytes` is > 1024? if so, doesn't the test `test_async_mgmt_queue_premium_create_with_queue_description` below cover this scenario?

sdk/servicebus/azure-servicebus/tests/async_tests/mgmt_tests/test_mgmt_queues_async.py
```diff
@@ -299,6 +306,99 @@ async def test_async_mgmt_queue_create_with_queue_description(self, servicebus_n
...
+            await mgmt_service.create_queue(
+                queue_name_3,
+                max_message_size_in_kilobytes=1023  # min allowed is 1024
+            )
+
+        with pytest.raises(HttpResponseError):
+            await mgmt_service.create_queue(
+                queue_name_3,
+                max_message_size_in_kilobytes=102401  # max allowed is 102400
```
swathipil, line 358: where are the max and min specified? is there a default value? I don't believe it's specified in the docstring

sdk/servicebus/azure-servicebus/tests/async_tests/mgmt_tests/test_mgmt_topics_async.py
```diff
@@ -88,6 +89,78 @@ async def test_async_mgmt_topic_create_with_topic_description(self, servicebus_n
             assert topic_2.enable_express
             assert topic_2.enable_partitioning
             assert topic_2.max_size_in_megabytes % 3072 == 0
+
+            with pytest.raises(HttpResponseError):
+                await mgmt_service.create_topic(
+                    topic_name_3,
+                    max_message_size_in_kilobytes=10
```
swathipil, line 96: same question: is this needed here when min size is being tested below?

sdk/servicebus/azure-servicebus/azure/servicebus/aio/management/_management_client_async.py
```diff
@@ -614,6 +619,7 @@ async def create_topic(self, topic_name: str, **kwargs) -> TopicProperties:
             availability_status=None,
             enable_express=kwargs.pop("enable_express", None),
             user_metadata=kwargs.pop("user_metadata", None),
+            max_message_size_in_kilobytes=kwargs.pop("max_message_size_in_kilobytes", None)
```
swathipil, line 622: I can't seem to spot where this is being used other than as a property like `self.max_message_size_in_kilobytes`. Is it meant to be used as a constraint somewhere? so like (` if size(message)< self.max_message_size_in_kilobytes:`)

swathipil: ah, just checked the issue - it says it's all service side validation. In that case, this is something the user sets so that the service can validate with this value?

sdk/servicebus/azure-servicebus/tests/async_tests/mgmt_tests/test_mgmt_queues_async.py
yunhaoling: this test case is to test that `max_message_size_in_kilobytes` isn't supported on standard tier

sdk/servicebus/azure-servicebus/tests/async_tests/mgmt_tests/test_mgmt_queues_async.py
yunhaoling: default value is the same as min which is 1024 kb, max is 102400 kb (100 mb). I will update the doc string

sdk/servicebus/azure-servicebus/tests/async_tests/mgmt_tests/test_mgmt_topics_async.py
yunhaoling: same, this is testing that standard tier doesn't support `max_message_size_in_kilobytes`, I should update the test case not to use 10 which is misleading as if I were testing the min threshold.

sdk/servicebus/azure-servicebus/azure/servicebus/aio/management/_management_client_async.py
yunhaoling: yes, and to be more precise on `user sets` -- user can create/update queue/topic of the property max_message_size_in_kilobytes being specified.

sdk/servicebus/azure-servicebus/tests/async_tests/mgmt_tests/test_mgmt_queues_async.py
yunhaoling: test updated to 1024

sdk/servicebus/azure-servicebus/tests/async_tests/mgmt_tests/test_mgmt_queues_async.py
yunhaoling: doc string updated

sdk/servicebus/azure-servicebus/tests/async_tests/mgmt_tests/test_mgmt_topics_async.py
yunhaoling: test updated to 1024

sdk/servicebus/azure-servicebus/azure/servicebus/aio/management/_management_client_async.py
swathipil: makes sense!

swathipil (approved):
lgtm!

PR #20819, opened 2021-09-24
[Test Proxy] Add registration methods for all sanitizers

mccoyp:
This replaces `add_sanitizer` with `add_*_sanitizer` methods for every sanitizer currently available to the proxy. This makes it easier to document how each sanitizer can be configured, and makes sanitizer registration more explicit. All sanitizer parameters are left as keyword-only arguments for now, but making some required in the future would probably be ideal. If we do go with this style of sanitizer methods, I'll also update the test proxy migration doc.


tools/azure-sdk-tools/devtools_testutils/sanitizers.py
```diff
@@ -0,0 +1,199 @@
...
+# --------------------------------------------------------------------------
+import requests
+from typing import TYPE_CHECKING
+
+from .config import PROXY_URL
+
+if TYPE_CHECKING:
+    from typing import Any, Dict
+
+
```
YalinLi0312, line 14: I think we can have a general add_sanitizer()  which including all add_XXX_sanitizer(), it would be easier to call in test proxy migration.

tools/azure-sdk-tools/devtools_testutils/sanitizers.py
```diff
@@ -0,0 +1,199 @@
...
+    :keyword str regex: A regex. Can be defined as a simple regex, or if a ``group_for_replace`` is provided, a
+        substitution operation.
+    :keyword str group_for_replace: The capture group that needs to be operated upon. Do not provide if you're invoking
+        a simple replacement operation.
+    """
+
+    request_args = {}
+    request_args["value"] = kwargs.get("value")
+    request_args["regex"] = kwargs.get("regex")
+    request_args["groupForReplace"] = kwargs.get("group_for_replace")
```
YalinLi0312, line 56: There are some similar lines in constructing request_args, is it possible to do some refactor based on this point? For example, have one method for preparing request arguments.

tools/azure-sdk-tools/devtools_testutils/sanitizers.py
mccoyp: I would like a single method, but it gets complicated to detail how the various kwargs should be used, depending on what kind of sanitizer you want. Splitting the method into one for each sanitizer makes it a lot easier to document any parameters you can provide and how they should be used, which is why I went for this.

I guess we could have a general method call the other methods based on the sanitizer type provided, yeah... I just wonder if that extra call layer would hide the parameter documentation and make things confusing. What do you think?

tools/azure-sdk-tools/devtools_testutils/sanitizers.py
```diff
@@ -174,14 +148,24 @@ def add_uri_regex_sanitizer(**kwargs):
...
+
+
+def _get_request_args(**kwargs):
+    # type: (**Any) -> Dict
+    """Returns a dictionary of sanitizer constructor headers"""
+
     request_args = {}
-    request_args["value"] = kwargs.get("value") or "fakevalue"
-    request_args["regex"] = (
-        kwargs.get("regex") or "(?<=\\/\\/)[a-z]+(?=(?:|-secondary)\\.(?:table|blob|queue)\\.core\\.windows\\.net)"
```
YalinLi0312: why getting rid of the second regex in your updates?

mccoyp: Having a storage-centric URI regex as the default regex doesn't really seem ideal going forward, so I thought I'd remove it while doing the refactoring to get request args from a helper method -- I'll try tracking down a more general URI regex that we could use here as a default, but until then I think it makes sense to remove since we have a regex that will work with Tables during the test refactoring

YalinLi0312 (approved):

PR #20867, opened 2021-09-25
[formrecognizer] regenerate on latest

kristapratico:
New swagger corrects API version from v3.0-preview.1 to v2021-09-30-preview


sdk/formrecognizer/azure-ai-formrecognizer/azure/ai/formrecognizer/_document_analysis_client.py
```diff
@@ -83,9 +83,9 @@ def begin_analyze_document(self, model, document, **kwargs):
             "prebuilt-document", "prebuilt-layout".
         :param document: JPEG, PNG, PDF, TIFF, or BMP type file stream or bytes.
         :type document: bytes or IO[bytes]
-        :keyword list[str] pages: Custom page numbers for multi-page documents(PDF/TIFF). Input the page numbers
+        :keyword str pages: Custom page numbers for multi-page documents(PDF/TIFF). Input the page numbers
```
kristapratico, line 86: latest swagger makes the change from list[str] to str. Since we're mostly just exposing the generated code here, we can do that change and discuss design later in beta2

sdk/formrecognizer/azure-ai-formrecognizer/azure/ai/formrecognizer/_models.py
```diff
@@ -8,7 +8,7 @@
 
 from enum import Enum
 from collections import namedtuple
-from ._generated.models import ModelInfo, Error
+from ._generated.v2021_09_30_preview.models import ModelInfo, Error
```
kristapratico, line 11: Being more explicit when importing from generated models now that we have some identical names, but different shapes across api versions.

sdk/formrecognizer/azure-ai-formrecognizer/azure/ai/formrecognizer/_generated/models.py
```diff
@@ -4,6 +4,4 @@
 # Licensed under the MIT License. See License.txt in the project root for
 # license information.
 # --------------------------------------------------------------------------
-from .v2_0.models import *
-from .v2_1.models import *
```
catalinaperalta, line 8: I know this is the generated code, but is it right for these to be removed?

catalinaperalta (approved):

PR #20887, opened 2021-09-27
Resolving Build Warnings

scbedd:
- [x] Meets intended purpose (addressing yellow builds)
- [x] Checked with `Run.Regression=true` and `SetDevVersion=true`
   - [x]  core
   - [x] appconfiguration


eng/tox/tox.ini
```diff
@@ -43,6 +43,7 @@ platform = linux: linux
 passenv = *
 setenv =
   SPHINX_APIDOC_OPTIONS=members,undoc-members,inherited-members
+  PIP_EXTRA_INDEX_URL=https://pypi.python.org/simple
```
weshaggard, line 46: Is this just working around the issue? Do we still end up with 2 feeds given this one is the extra feed?

scbedd: **yes**.

I'm working on another commit that will actually check the package before installing from the feed. If it's a non-azure-sdk publisher, we will hard error out.

This will happen in `create_package_and_install.py`. The mechanism that we install packages.

weshaggard (approved):
This is a good first step we should still follow-up to try and split the resolves in our dev package case.

PR #20899, opened 2021-09-27
[Perf Framework] Use urljoin to concatenate URLs

mikeharder:
- Existing code fails if `test_proxy` option ends with `/`

### DIFF
--- a/tools/azure-devtools/src/azure_devtools/perfstress_tests/perf_stress_test.py
+++ b/tools/azure-devtools/src/azure_devtools/perfstress_tests/perf_stress_test.py
@@ -6,11 +6,7 @@
 import os
 import aiohttp
 
-try:
-    from urllib.parse import urljoin
-except ImportError:
-    from urlparse import urljoin
-
+from urllib.parse import urljoin
 from ._policies import PerfTestProxyPolicy
 
 


tools/azure-devtools/src/azure_devtools/perfstress_tests/perf_stress_test.py
```diff
@@ -6,6 +6,11 @@
 import os
 import aiohttp
 
+try:
+    from urllib.parse import urljoin
+except ImportError:
+    from urlparse import urljoin
```
annatisch, line 12: This try/except around the import is probably unnecessary - as the perfstress framework can only be run under Python 3 anyway, as it's natively async.

tools/azure-devtools/src/azure_devtools/perfstress_tests/perf_stress_test.py
```diff
@@ -67,7 +72,7 @@ async def stop_playback(self):
             "x-recording-id": self._recording_id,
             "x-purge-inmemory-recording": "true"
         }
-        url = self.args.test_proxy + "/playback/stop"
+        url = urljoin(self.args.test_proxy, "/playback/stop")
```
annatisch, line 75: I think this will only work if `self.args.test_proxy` is only ever a hostname. From memory, ff it also has a path, I think that path gets replaced by the supplied value, rather than concatenated. Which I think is sufficient to our needs - but just making sure.

mikeharder: The current contract is that `test_proxy` should just specify the protocol (`http` vs `http`), hostname, and port.  For example, when redirecting a request from a live service to the test proxy, we use the protocol, hostname, and port from `test_proxy`, but use original `path` without trying to combine with an exsting path in `test_proxy`.

In the future we could decide to add support for a base path in `test_proxy`, and this code would need to remove the leading slash (from `urljoin(test_proxy, "/playback/stop")` to `urljoin(test_proxy, "playback/stop")`).  But more changes would be required to merge the base path when redirecting requests.

annatisch (approved):

PR #20904, opened 2021-09-28
[formrecognizer] Print fields in custom documents samples

### LINKED ISSUE #20845: [formrecognizer] sample_analyze_custom_documents should show fields
@kristapratico commented on [Fri Sep 24 2021](https://github.com/Azure/azure-sdk-for-python-pr/issues/791)

we should show printing out the fields in these samples

catalinaperalta:
Fixes #20845

### DIFF
--- a/sdk/formrecognizer/azure-ai-formrecognizer/samples/v3.2-beta/async_samples/sample_analyze_custom_documents_async.py
+++ b/sdk/formrecognizer/azure-ai-formrecognizer/samples/v3.2-beta/async_samples/sample_analyze_custom_documents_async.py
@@ -34,6 +34,17 @@ import os
 import asyncio
 
 
+def format_bounding_region(bounding_regions):
+    if not bounding_regions:
+        return "N/A"
+    return ", ".join("Page #{}: {}".format(region.page_number, format_bounding_box(region.bounding_box)) for region in bounding_regions)
+
+def format_bounding_box(bounding_box):
+    if not bounding_box:
+        return "N/A"
+    return ", ".join(["[{}, {}]".format(p.x, p.y) for p in bounding_box])
+
+
 async def analyze_custom_documents_async(custom_model_id):
     path_to_sample_documents = os.path.abspath(
         os.path.join(
@@ -69,6 +80,12 @@ async def analyze_custom_documents_async(custom_model_id):
         print("Document has type {}".format(document.doc_type))
         print("Document has document type confidence {}".format(document.confidence))
         print("Document was analyzed with model with ID {}".format(result.model_id))
+        print("...within {} bounding regions".format(format_bounding_region(document.bounding_regions)))
+        for key, value in document.fields.items():
+            field_value = value.content
+            if value.value:
+                field_value = value.value
+            print("......found field of type '{}' with value '{}' within '{}' bounding regions and with confidence {}".format(value.value_type, field_value, format_bounding_region(value.bounding_regions), value.confidence))
 
     # iterate over tables, lines, and selection marks on each page
     for page in result.pages:
--- a/sdk/formrecognizer/azure-ai-formrecognizer/samples/v3.2-beta/async_samples/sample_analyze_custom_documents_async.py
+++ b/sdk/formrecognizer/azure-ai-formrecognizer/samples/v3.2-beta/async_samples/sample_analyze_custom_documents_async.py
@@ -34,17 +34,6 @@ import os
 import asyncio
 
 
-def format_bounding_region(bounding_regions):
-    if not bounding_regions:
-        return "N/A"
-    return ", ".join("Page #{}: {}".format(region.page_number, format_bounding_box(region.bounding_box)) for region in bounding_regions)
-
-def format_bounding_box(bounding_box):
-    if not bounding_box:
-        return "N/A"
-    return ", ".join(["[{}, {}]".format(p.x, p.y) for p in bounding_box])
-
-
 async def analyze_custom_documents_async(custom_model_id):
     path_to_sample_documents = os.path.abspath(
         os.path.join(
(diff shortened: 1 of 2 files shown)


sdk/formrecognizer/azure-ai-formrecognizer/samples/v3.2-beta/async_samples/sample_analyze_custom_documents_async.py
```diff
@@ -34,6 +34,17 @@
...
 
+def format_bounding_region(bounding_regions):
+    if not bounding_regions:
+        return "N/A"
+    return ", ".join("Page #{}: {}".format(region.page_number, format_bounding_box(region.bounding_box)) for region in bounding_regions)
+
+def format_bounding_box(bounding_box):
+    if not bounding_box:
+        return "N/A"
+    return ", ".join(["[{}, {}]".format(p.x, p.y) for p in bounding_box])
```
maririos, line 45: Are both of these something that might be worth adding to the model object itself so a user can also consume it?
As an example, in V3 of .NET we added a convenience `ToString` for `BoundingBox` https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/formrecognizer/Azure.AI.FormRecognizer/src/FieldBoundingBox.cs#L64

sdk/formrecognizer/azure-ai-formrecognizer/samples/v3.2-beta/async_samples/sample_analyze_custom_documents_async.py
```diff
@@ -34,6 +34,17 @@
 import asyncio
 
 
+def format_bounding_region(bounding_regions):
+    if not bounding_regions:
+        return "N/A"
```
maririos, line 39: the user will see  `... within N/A bounding regions` right?
what about not printing that line at all if the bounding region is not present? 

Same comment will apply for bounding box

maririos (approved):

sdk/formrecognizer/azure-ai-formrecognizer/samples/v3.2-beta/async_samples/sample_analyze_custom_documents_async.py
catalinaperalta: That's a nice convenience to have! Although, I'd rather wait and see if we get requests for it, rather than add it without validating that there's a large use case for it. Maybe we could see if there are users who mention this in UX studies.

maririos: We did add it in Bounding Box because of customer feedback. At least in .NET :)

sdk/formrecognizer/azure-ai-formrecognizer/samples/v3.2-beta/async_samples/sample_analyze_custom_documents_async.py
```diff
@@ -69,6 +80,12 @@ async def analyze_custom_documents_async(custom_model_id):
         print("Document has type {}".format(document.doc_type))
         print("Document has document type confidence {}".format(document.confidence))
         print("Document was analyzed with model with ID {}".format(result.model_id))
+        print("...within {} bounding regions".format(format_bounding_region(document.bounding_regions)))
+        for key, value in document.fields.items():
```
kristapratico, line 84: I think we should stick with `name, field` as to not confuse with key-value pairs.

sdk/formrecognizer/azure-ai-formrecognizer/samples/v3.2-beta/async_samples/sample_analyze_custom_documents_async.py
```diff
@@ -69,6 +80,12 @@ async def analyze_custom_documents_async(custom_model_id):
...
+            field_value = value.content
+            if value.value:
+                field_value = value.value
```
kristapratico, line 87: Could do...

`field_value = field.value if field.value else field.content`

sdk/formrecognizer/azure-ai-formrecognizer/samples/v3.2-beta/async_samples/sample_analyze_custom_documents_async.py
kristapratico: Just a general comment on this - in order to keep the samples simple and focused in v2 we created a separate sample `get_bounding_boxes` for people who were more interested in the location of text on the page... I'm kind of inclined to do the same here so we can leave this sample to just be mostly fields/values/confidences

catalinaperalta: That sounds good...I'll remove this from the sample

catalinaperalta: Done

sdk/formrecognizer/azure-ai-formrecognizer/samples/v3.2-beta/async_samples/sample_analyze_custom_documents_async.py
catalinaperalta: Done

sdk/formrecognizer/azure-ai-formrecognizer/samples/v3.2-beta/async_samples/sample_analyze_custom_documents_async.py
catalinaperalta: Done

kristapratico (approved):

PR #20919, opened 2021-09-28
[formrecognizer] README feedback

catalinaperalta:
Addressing some of Paul's feedback on the README.


catalinaperalta:
Addressed in separate PR

PR #20894, opened 2021-09-27
[formrecognizer] Add prebuilt-document samples and tests

catalinaperalta:
Fixes #20847, #20842

This PR adds samples for prebuilt-document, along with tests. Also includes some fixes to `_from_generated()` methods, testcase checks and a misspelling in the layout samples.

PR #20959, opened 2021-09-30
[formrecognizer] Small sample fixes

catalinaperalta:
Adding fixes for feedback received in #20894

### DIFF
--- a/sdk/formrecognizer/azure-ai-formrecognizer/samples/v3.2-beta/async_samples/sample_analyze_general_document_async.py
+++ b/sdk/formrecognizer/azure-ai-formrecognizer/samples/v3.2-beta/async_samples/sample_analyze_general_document_async.py
@@ -7,18 +7,14 @@
 # --------------------------------------------------------------------------
 
 """
-FILE: sample_analyze_document.py
+FILE: sample_analyze_document_async.py
 
 DESCRIPTION:
     This sample demonstrates how to extract general document information from a document
     given through a file.
 
-    Note that selection marks returned from begin_analyze_document() do not return the text associated with
-    the checkbox. For the API to return this information, build a custom model to analyze the checkbox and its text.
-    See sample_build_model.py for more information.
-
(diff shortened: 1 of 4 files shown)


sdk/formrecognizer/azure-ai-formrecognizer/samples/v3.2-beta/async_samples/sample_analyze_general_document_async.py
```diff
@@ -7,18 +7,14 @@
 # --------------------------------------------------------------------------
 
 """
-FILE: sample_analyze_document_async.py
+FILE: sample_analyze_general_document_async.py
```
catalinaperalta, line 10: Just going to move the naming conversation over here. So you suggest prebuilt instead of general? I'll update it.

kristapratico: Unless we can find a better name, I think "prebuilt" tells you more about what it is than "general" does. Not sure what other languages are doing though

sdk/formrecognizer/azure-ai-formrecognizer/samples/v3.2-beta/async_samples/sample_analyze_general_document_async.py
```diff
@@ -155,7 +151,7 @@ async def analyze_document():
                 )
         if kv_pair.value:
             print(
-                    "Value '{}' found within '{}' bounding regions".format(
+                    "\nValue '{}' found within '{}' bounding regions".format(
```
kristapratico, line 154: sorry I wasn't clear, I meant \n at the end of the string so hopefully the output would be like:

Key: thing
Value: thing

Key: thing2
Value: thing2

But I'm open to whatever looks good printed on the console.

kristapratico (approved):

Issue #20839, opened 2021-09-25
[formrecognizer] fix FTC TODOs in code/tests/samples


kristapratico:
@kristapratico commented on [Thu Aug 26 2021](https://github.com/Azure/azure-sdk-for-python-pr/issues/694)

Pending on service updates not yet available.

PR #20978, opened 2021-09-30
[formrecognizer] re-enable tests

kristapratico:
Resolves https://github.com/Azure/azure-sdk-for-python/issues/20839

### DIFF
--- a/sdk/formrecognizer/azure-ai-formrecognizer/tests/test_receipt_async.py
+++ b/sdk/formrecognizer/azure-ai-formrecognizer/tests/test_receipt_async.py
@@ -400,14 +400,13 @@ class TestReceiptFromStreamAsync(AsyncFormRecognizerTest):
 
     @FormRecognizerPreparer()
     @GlobalClientPreparer()
-    @pytest.mark.skip("the service is returning a different error code")
     async def test_receipt_locale_error(self, client):
         with open(self.receipt_jpg, "rb") as fd:
             receipt = fd.read()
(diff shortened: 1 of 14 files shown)


sdk/formrecognizer/azure-ai-formrecognizer/tests/test_receipt_async.py
```diff
@@ -400,14 +400,13 @@ async def test_receipt_locale_specified(self, client):
...
     @GlobalClientPreparer()
-    @pytest.mark.skip("the service is returning a different error code")
     async def test_receipt_locale_error(self, client):
         with open(self.receipt_jpg, "rb") as fd:
             receipt = fd.read()
         with pytest.raises(HttpResponseError) as e:
             async with client:
                 await client.begin_analyze_document("prebuilt-receipt", receipt, locale="not a locale")
-        assert "UnsupportedLocale" == e.value.error.code
+        assert "InvalidArgument" == e.value.error.code
```
catalinaperalta, line 409: So is this the error code to expect from now on here?

catalinaperalta (approved):
🚢

sdk/formrecognizer/azure-ai-formrecognizer/tests/test_receipt_async.py
kristapratico: Yes, for v2021-09-30-preview this is the error code. Also if the service changes up the error response due to REST feedback during the betas we'll know with this test.

Issue #20973, opened 2021-09-30
[formrecognizer] update readme link with migration guide + example for prebuilt doc


kristapratico:
- [x] README needs migration guide link
- [x] README needs prebuilt-document example
- [x] sample README needs migration guide link

PR #20996, opened 2021-09-30
[formrecognizer] add prebuilt document to readme examples

kristapratico:
last of Paul's feedback on readme to separate prebuilt document into its own example

Part of #20973

### DIFF
--- a/sdk/formrecognizer/azure-ai-formrecognizer/README.md
+++ b/sdk/formrecognizer/azure-ai-formrecognizer/README.md
@@ -191,6 +191,7 @@ The following section provides several code snippets covering some of the most c
 
 * [Extract layout](#extract-layout "Extract Layout")
 * [Using Prebuilt Models](#using-prebuilt-models "Using Prebuilt Models")
+* [Using Prebuilt Document](#using-prebuilt-document "Using Prebuilt Document")
 * [Build a Model](#build-a-model "Build a model")
 * [Analyze Documents Using a Custom Model](#analyze-documents-using-a-custom-model "Analyze Documents Using a Custom Model")
 * [Manage Your Models](#manage-your-models "Manage Your Models")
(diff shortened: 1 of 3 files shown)


sdk/formrecognizer/azure-ai-formrecognizer/samples/v3.2-beta/async_samples/sample_analyze_prebuilt_document_async.py
```diff
@@ -63,12 +63,10 @@ async def analyze_document():
...
 
-    for idx, style in enumerate(result.styles):
-        print(
-            "Document contains {} content".format(
-                "handwritten" if style.is_handwritten else "no handwritten"
-            )
-        )
+    for style in result.styles:
+        if style.is_handwritten:
+            print("Document contains handwritten content: ")
```
catalinaperalta, line 68: Do we not want to say that there was no handwritten content on the document? I dont really mind, just wondering because that means if people copy this they wont see the locations of the none handwritten content.

sdk/formrecognizer/azure-ai-formrecognizer/README.md
```diff
@@ -310,6 +311,83 @@ You are not limited to receipts! There are a few prebuilt models to choose from,
...
+print("----Key-value pairs found in document----")
+for kv_pair in result.key_value_pairs:
+    if kv_pair.key:
+        print(
+                "Key '{}' found within '{}' bounding regions".format(
+                    kv_pair.key.content,
+                    kv_pair.key.bounding_regions,
+                )
+            )
+    if kv_pair.value:
```
catalinaperalta, line 349: Is the actual sample in this order? If not then I think the sample should be changed to this, I cant remember if I updated it I know I thought that this was the right order to showcase things.

catalinaperalta (approved):
LGTM

sdk/formrecognizer/azure-ai-formrecognizer/samples/v3.2-beta/async_samples/sample_analyze_prebuilt_document_async.py
kristapratico: Before it would print "Document contains handwritten content" for every instance of handwritten content. I think this is a good opportunity to show people how to use the spans to find the specific handwritten content (sans helper methods this beta). We could set a var to hold whether any handwritten content was found while iterating over the list and then print "No handwritten content" but since we aren't doing that for any of the other features like selection marks, entities, etc I left as-is. What do you mean by "locations of the none handwritten content"?

sdk/formrecognizer/azure-ai-formrecognizer/README.md
kristapratico: I just re-ordered because I wanted the output to match the sentence order preceding the sample.

kristapratico: I don't think order is important? but I can change the sample if you think it should match

sdk/formrecognizer/azure-ai-formrecognizer/samples/v3.2-beta/async_samples/sample_analyze_prebuilt_document_async.py
catalinaperalta: I like your idea. I just meant also printing the spans for places where there's no handwritten content. But I think it's enough to show it with one, shouldnt be too hard for users to switch it up for whichever one they want. This LGTM as is then :)

sdk/formrecognizer/azure-ai-formrecognizer/README.md
catalinaperalta: Nah it's ok, it's fine as is for now. Was just curious

maririos (approved):

Issue #20621, opened 2021-09-09
Add getCryptographyClient to KeyClient


maorleger:
Add a method `getCryptographyClient` to `KeyClient` that accepts a key name and optional key version and returns a `CryptographyClient` that reuses the same pipeline (like in dotnet) or generated client (like in JS). 

Motivation for this change is both discoverability as well as perf (by sharing a single pipeline)

Issue #21005, opened 2021-10-01
[Key Vault] Ensure crypto operations work as expected for all API versions


mccoyp:
Context: the generated [KeyOperationResult](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_generated/v7_2/models/_models_py3.py#L635) class only has `iv`, `authentication_tag`, and `additional_authenticated_data` attributes in API versions 7.2+. This isn't usually an issue, since most or all algorithms using these values require symmetric keys -- these can only be used with Managed HSM for now, which only supports API versions 7.2+.

However, some crypto operations that use some of these attributes (AES-CBCPAD, for example) can be performed locally. It's possible that someone would -- for whatever reason -- create a CryptographyClient with an earlier API version than 7.2 and attempt to perform crypto operations that use these attributes. That may not work since the generated models loaded by the CryptographyClient may not have support for the attributes.

To fix this, if it is a problem (which I have reason to think it is), we'll probably want to add a check before performing certain crypto operations -- either when we check that an algorithm is supported by our local crypto provider, or inside the `encrypt`/`decrypt`/etc. method.

PR #21006, opened 2021-10-01
[Key Vault] Add KeyClient.get_cryptography_client method

mccoyp:
Resolves https://github.com/Azure/azure-sdk-for-python/issues/20621. This also fixes three bugs:

- When providing a generated client to a KV client, we don't currently set any generated models for the created type. As a result, any reference to `self._models` from the client will raise an AttributeError
  - This updates the base client (for `administration`, `certificates`, `keys`, and `secrets`) to accept a set of generated models, and will otherwise set them based on the provided or default API version
- The base client for `azure-keyvault-administration` expects a NotImplementedError in the case of an incorrect API version, but should instead expect a ValueError (this change in the generated client happened, I think, in API version 7.2)
  - Like the other packages, `administration`'s base client will still raise a NotImplementedError in this case but will know to do so when it catches a ValueError
- We currently assume that an EncryptResult will always have an IV, authentication tag, and additional authenticated data. This is only true for recent API versions, so we get an AttributeError for older versions in some cases
  - This updates CryptographyClient to first check for these attributes on the EncryptResult we get back from the operation

It's still possible that there are API version-dependent attribute issues in some cases I didn't have time to consider; looking into that is tracked by https://github.com/Azure/azure-sdk-for-python/issues/21005.

### DIFF
--- a/sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_client.py
+++ b/sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_client.py
@@ -68,6 +68,7 @@ class KeyClient(KeyVaultClientBase):
             HTTP client as this :class:`~azure.keyvault.keys.KeyClient`.
         :rtype: ~azure.keyvault.keys.crypto.CryptographyClient
         """
+        # We provide a fake credential because the generated client already has the KeyClient's real credential
         return CryptographyClient(
             key, object(), generated_client=self._client, generated_models=self._models  # type: ignore
         )
--- a/sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_client.py
+++ b/sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_client.py
@@ -19,11 +19,15 @@ except ImportError:
 if TYPE_CHECKING:
     # pylint:disable=unused-import
     from typing import Any, Iterable, Optional, Union
-    from azure.core.credentials import TokenCredential
     from azure.core.paging import ItemPaged
     from azure.core.polling import LROPoller
     from ._models import JsonWebKey
     from ._enums import KeyType
+
+
+def _get_key_id(vault_url, key_name, version=None):
+    without_version = "{}/keys/{}".format(vault_url, key_name)
+    return without_version + "/" + version if version else without_version
 
 
 class KeyClient(KeyVaultClientBase):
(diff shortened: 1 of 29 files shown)


mccoyp:
Commit [31e3ac4](https://github.com/Azure/azure-sdk-for-python/pull/21006/commits/31e3ac40a5b669889e2b782a37fc248af8339e0e) touches on the main points of interest for review

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/aio/_client.py
```diff
@@ -61,20 +62,23 @@ def _get_attributes(self, enabled, not_before, expires_on, exportable=None):
...
         :returns: A :class:`~azure.keyvault.keys.crypto.aio.CryptographyClient` using the same options, credentials, and
             HTTP client as this :class:`~azure.keyvault.keys.aio.KeyClient`.
         :rtype: ~azure.keyvault.keys.crypto.aio.CryptographyClient
         """
+        key_id = _get_key_id(self._vault_url, key_name, kwargs.get("version"))
+
         # We provide a fake credential because the generated client already has the KeyClient's real credential
         return CryptographyClient(
-            key, object(), generated_client=self._client, generated_models=self._models  # type: ignore
+            key_id, object(), generated_client=self._client, generated_models=self._models  # type: ignore
```
maorleger, line 81: Maybe I missed it, but just to confirm - this `generated_client` will be the one that is used to create a cryptoClient? And do we skip instantiating a throwaway `generated_client` in that case?

maorleger (commented):
Nice! Two questions - one is just for my own understanding. The fake credential is a bit weird but I completely understand why it's necessary (and the simplest approach here)

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_client.py
```diff
@@ -19,13 +19,17 @@
...
     from typing import Any, Iterable, Optional, Union
-    from azure.core.credentials import TokenCredential
     from azure.core.paging import ItemPaged
     from azure.core.polling import LROPoller
     from ._models import JsonWebKey
     from ._enums import KeyType
 
 
+def _get_key_id(vault_url, key_name, version=None):
+    without_version = "{}/keys/{}".format(vault_url, key_name)
```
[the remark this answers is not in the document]
mccoyp: Good question -- luckily we know that a KeyClient's vault URL shouldn't have a trailing / because it's [stripped by the base client](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_shared/client_base.py#L45)

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/aio/_client.py
mccoyp: Yep! We just directly set the generated client of the resulting CryptographyClient to be the one we pass in here. The tests for `get_cryptography_client` check that the `._client` of both clients reference the same object (using an `==` equality, rather than just checking for properties)

maorleger (approved):

PR #21025, opened 2021-10-02
Resolve Installation Ordering during dev build

scbedd:
@weshaggard  this is the followup that we discussed in #20887 

This addresses any standard installation of a package or wheel. It does this by understanding and downloading the requirements of a package _before_ it is installed. 

Once we've confirmed download works, we install facing the greater pypi.


eng/tox/create_package_and_install.py
```diff
@@ -165,21 +173,39 @@ def build_and_discover_package(setuppy_path, dist_dir, target_setup, package_typ
...
 
+    commands_options = []
+    built_pkg_path = ""
     setup_py_path = os.path.join(args.target_setup, "setup.py")
+    additional_downloaded_reqs = []
+    tmp_dl_folder = os.path.join(args.distribution_directory, "dl")
+    os.mkdir(tmp_dl_folder)
+
+    # If extra index URL is passed then set it as argument to pip command
+    if args.extra_index_url:
```
weshaggard, line 193: Perhaps I'm missing something but shouldn't we no longer need to pass the extra-index-url after this update?

eng/tox/create_package_and_install.py
```diff
@@ -165,21 +173,39 @@ def build_and_discover_package(setuppy_path, dist_dir, target_setup, package_typ
...
+        commands_options.extend(["--extra-index-url", args.extra_index_url])
+
+    # preview version is enabled when installing dev build so pip will install dev build version from devpos feed
+    if args.install_preview or os.getenv("SetDevVersion"):
```
weshaggard, line 197: Can we consistently use either SetDevVersion or --pre-download-disable to trigger this fork?

scbedd: Will definitely give it a shot.

eng/tox/create_package_and_install.py
```diff
@@ -188,16 +202,39 @@ def build_and_discover_package(setuppy_path, dist_dir, target_setup, package_typ
...
+                        sys.executable,
+                        "-m",
+                        "pip",
+                        "download",
+                        "-d",
+                        tmp_dl_folder,
+                        "--no-deps",
+                    ]
+
+                    download_command.extend(azure_requirements)
```
praveenkuttappan, line 230: Are we always downloading latest of required packages? Ideally that should be fine since dependency is against what's in repo unless some package decides to pin dependency. right?

scbedd: Correct! We actually explicitly _disable_ this piece in `latestdependency` and `mindependency`. My offline text chat describes the issue with those however.

eng/tox/create_package_and_install.py
```diff
@@ -165,21 +165,35 @@ def build_and_discover_package(setuppy_path, dist_dir, target_setup, package_typ
         default="wheel",
     )
 
+    parser.add_argument(
+        "--pre-download-disable",
+        dest="pre_download",
```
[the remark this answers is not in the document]
scbedd: I think you're right. Good call. Thanks!

praveenkuttappan (approved):
Looks good

scbedd:
The single core failure above is already resolved in another PR. Merging this.

PR #21028, opened 2021-10-03
Refactor AzureJSONEncoder

Codejune:
Relation of #17102 
### Descriptions
- Added `_timestr_as_isostr()` method for reducing indent
- Remove duplicate method calls try `super().default()`


sdk/core/azure-core/azure/core/serialization.py
```diff
@@ -95,29 +107,16 @@ class AzureJSONEncoder(JSONEncoder):
     """A JSON encoder that's capable of serializing datetime objects and bytes."""
 
     def default(self, o):  # pylint: disable=too-many-return-statements
+        if isinstance(o, (bytes, bytearray)):
```
mccoyp, line 110: I think that `try`ing to return `super(AzureJSONEncoder, self).default(o)` first, and then trying the rest in the `except` block, is probably the most Pythonic approach.

You raise a good point though -- rather than calling `super(...).default(o)` both at the start and at the end, we can probably just raise the exception we got from the first attempt at the end of the method, since we know what will happen. So something like
```python
try:
    return super(AzureJSONEncoder, self).default(o)
except TypeError as e:
    (try different serializations)
    raise e
```

sdk/core/azure-core/azure/core/serialization.py
```diff
@@ -95,29 +107,16 @@ class AzureJSONEncoder(JSONEncoder):
...
+            return base64.b64encode(o).decode()
+        try:
```
mccoyp, line 112: nit: to make it more clear what order we're trying things in inside the method body:
```suggestion
        try:
            # First, try datetime.datetime and datetime.date/datetime.time
```

sdk/core/azure-core/azure/core/serialization.py
```diff
@@ -82,7 +82,19 @@ def _timedelta_as_isostr(value):
 
     return "P" + date + time
 
-
+def _timestr_as_isostr(o):
```
mccoyp, line 85: nit: it's hard to come up with a good name for this helper, since we're checking for a `datetime`, `date`, or `time`... but maybe `_time_as_isostr` would be a more fitting name, since the object we're passing in is a non-string ("timestr" implies to me that we're being given a time as a string). What do you think?

mccoyp: Also, since this has been moved out of the `AzureJSONEncoder.default` method, it might be nice to rename the parameter `o` to something more descriptive 🙂 Maybe `value` to align with `_timedelta_as_isostr`?

mccoyp (commented):
Thank you so much for opening a PR for this! I just have a few suggestions, but this is a great refactoring

Codejune:
I confirmed that in Azure Pipelines/python-autorest-pr it still throws an error. Am I missing something about this step?

PR #21033, opened 2021-10-04
[formrecognizer] Run black on code

catalinaperalta:
NOTE: this PR runs black on all the code except `_models.py` due to some formatting issues that conflict with pylint.


kristapratico (approved):
🎉

PR #21063, opened 2021-10-05
Enabling Python 3.10 in build matrix

scbedd:
Also adding skip logic so that an incompatible package can be omitted from the tests on py 3.10.


scbedd:
Dependency installation stumbled on `cmarkgfm`

May need to build a macos wheel manually for it.

scbedd:
Checks out. 
image

Check-enforcer overriding and merging this PR.

PR #21093, opened 2021-10-06
[Key Vault] Adjust key rotation test for preview behavior

mccoyp:
Context: live rotation policy tests have started [failing](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=1129470&view=ms.vss-test-web.build-test-results-tab&runId=24486150&resultId=100322&paneView=debug) because the service has started honoring a restriction that was specified in generated enums: [rotation policies can't configure the notification time for a notifying policy in preview](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_generated/v7_3_preview/models/_key_vault_client_enums.py#L35). This adjusts tests to use the required time before expiry for these policies, and also fixes a previously-uncaught bug in setting the expiry time for key policies.


vcolin7 (approved):
LGTM :)

Issue #21057, opened 2021-10-05
Regenerate ACR package on stable/2021-07-01 rest api


YalinLi0312:
Regenerate on [Azure/azure-rest-api-specs@`c8d9a26a2857828e095903efa72512cf3a76c15d`](https://github.com/Azure/azure-rest-api-specs/commit/c8d9a26a2857828e095903efa72512cf3a76c15d)

YalinLi0312:
Resolved.

PR #21110, opened 2021-10-06
[ACR] React to stable/2021-07-01 swagger changes

YalinLi0312:
For issue: https://github.com/Azure/azure-sdk-for-python/issues/21057


YalinLi0312:
> Looks like it's all changes to generated code. LGTM, probably should also have someone from Python team sign off

Can you help with that? I'm actually don't know who is the best person to tag.

PR #21117, opened 2021-10-07
[Blob]ARM migration

xiafu-msft:


xiafu-msft:
azp run python - storage - tests

xiafu-msft:
I'm pretty sure the failed 1 test is just flaky

xiafu-msft:
> > I'm pretty sure the failed 1 test is just flaky, I clicked rerun and it took 33 min already...
> 
> do we know why is it flaky? i.e. is it problem with test code or do we have a bug (race) or there's eventual consistency happening on the service side?

It should be eventual consistency, as the blob_tier_inferred is None, if it happens frequently we can add sleep time to test (put it in doc). For now as the test have passed I will merge the pr.
```
            blob_ref = blob.get_blob_properties()
            self.assertIsNotNone(blob_ref.blob_tier)
>           self.assertTrue(blob_ref.blob_tier_inferred)
E           AssertionError: None is not true
```

kasobol-msft:
> > > I'm pretty sure the failed 1 test is just flaky, I clicked rerun and it took 33 min already...
> > 
> > 
> > do we know why is it flaky? i.e. is it problem with test code or do we have a bug (race) or there's eventual consistency happening on the service side?
> 
> It should be eventual consistency, as the blob_tier_inferred is None, if it happens frequently we can add sleep time to test (put it in doc). For now as the test have passed I will merge the pr.
> 
> ```
>             blob_ref = blob.get_blob_properties()
>             self.assertIsNotNone(blob_ref.blob_tier)
> >           self.assertTrue(blob_ref.blob_tier_inferred)
> E           AssertionError: None is not true
> ```

sleep or better polling with timeout.
good to merge, please create work item to improve that test.

PR #21123, opened 2021-10-07
[formrecognizer] Add perfstress test for new clients

catalinaperalta:
Part of #20983


catalinaperalta:
Closing since perf tests we added in #22163

PR #21152, opened 2021-10-08
[textanalytics] support multiple actions of the same type

kristapratico:
Resolves https://github.com/Azure/azure-sdk-for-python/issues/19612

Using `task_name` to correlate inputs with outputs, but not exposing this property publicly. This lets us guarantee correct ordering returned.

Had to rerecord all tests now that we always set `task_name` for every call to `begin_analyze_actions`.


kristapratico:
> curious on what is the reason to not exposing ActionName to the user and let them set the value

Python has a more flattened approach than other languages - so we don't have a type that corresponds to the action result itself (where an action_name might live). Since action_name would be used to correlate inputs/outputs, we can get away with not exposing since order is guaranteed already.

PR #21162, opened 2021-10-08
[Perf] Call Run() once before starting recording

mikeharder:
- Avoids capturing one-time setup like authorization requests
- Ported from https://github.com/Azure/azure-sdk-for-net/pull/24263


mikeharder:
@annatisch: All comments resolved, please re-review

PR #21195, opened 2021-10-11
Enable API change detection for Python PR

praveenkuttappan:
Add step to identify python API change detection in PR validation.


praveenkuttappan:
I have pushed a dummy change to verify diffing again.

weshaggard:
@praveenkuttappan why do the diff show empty lines? Is that expected from an apiview diff?

PR #21203, opened 2021-10-11
[Perf] Support multiple test proxies

mikeharder:
- Some scenarios require multiple test proxies to bottleneck client
- Proxies are assigned round-robin to parallel test instances
- Ported from https://github.com/Azure/azure-sdk-for-net/pull/24265


mikeharder:
@annatisch: ping

PR #21242, opened 2021-10-13
[formrecognizer] Fix get operations samples

catalinaperalta:
These samples fail when no operations are returned.


catalinaperalta:
> I'm a little curious why this sample passes in most CI environments, but not all. Since the samples run after all test cases there should be plenty of operations to return in the 24h window.
> 
> But I do think this change is a good addition anyway in case someone tries to run the sample having not created any operations yet.

Yeah I had the same thought, my reasoning was that those environments that fail are either too fast or slow so there are no operations left by the time it reaches those samples.

Issue #20698, opened 2021-09-15
Add support for multi-tenant authentication to Key Vault


heaths:
Support for multi-tenant authentication may have been added to your language's version of Azure.Core. Since support for multi-tenant authentication has been a frequent customer ask in some languages, let's add support in all languages where possible.

See https://github.com/Azure/azure-sdk-for-net/issues/18359 for more information and help for how to conceptually implement it.

heaths:
See https://github.com/Azure/azure-sdk-for-net/pull/24558 for .NET's PR as an example.

PR #21290, opened 2021-10-15
[Key Vault] Add support for multi-tenant authentication

mccoyp:
Resolves https://github.com/Azure/azure-sdk-for-python/issues/20698.

Context: now that `azure-identity` supports providing a tenant ID to token requests, we can allow Key Vault clients to make use of tenant discovery. This updates the challenge authentication policy to parse out the tenant ID provided in a challenge and authenticate requests with that tenant. Based off of [work Charles did](https://github.com/chlowell/azure-sdk-for-python/blob/keyvault-challenge-policy/sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_shared/challenge_auth_policy.py) in his fork of the repo, this also updates the challenge auth policy to inherit from `azure-core`'s BearerTokenCredentialPolicy instead of re-implementing the wheel. As a result, doing so also requires `azure-core` >= 1.15.0 (see [changelog](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/core/azure-core/CHANGELOG.md#1150-2021-06-04)).


sdk/keyvault/azure-keyvault-administration/azure/keyvault/administration/_internal/async_challenge_auth_policy.py
```diff
@@ -15,65 +15,61 @@
...
-    from azure.core.credentials_async import AsyncTokenCredential
-    from azure.core.pipeline import PipelineRequest
-    from azure.core.pipeline.transport import AsyncHttpResponse
-    from . import HttpChallenge
+    from typing import Any, Optional
+    from azure.core.pipeline import PipelineRequest, PipelineResponse
 
 
-class AsyncChallengeAuthPolicy(ChallengeAuthPolicyBase, AsyncHTTPPolicy):
+class AsyncChallengeAuthPolicy(AsyncBearerTokenCredentialPolicy):
```
xiangyan99, line 28: Could you give more context about this change?

mccoyp: Just updated the PR description with more context!

sdk/keyvault/azure-keyvault-administration/azure/keyvault/administration/_internal/async_challenge_auth_policy.py
```diff
@@ -15,65 +15,60 @@
...
-
-    async def send(self, request: "PipelineRequest") -> "AsyncHttpResponse":
+    def __init__(self, *args: "Any", **kwargs: "Any") -> None:
+        self._last_tenant_id = None  # type: Optional[str]
+        super().__init__(*args, **kwargs)
+
+    async def on_request(self, request: "PipelineRequest") -> None:
+        challenge = ChallengeCache.get_challenge_for_url(request.http_request.url)
+        if challenge:
+            if self._last_tenant_id == challenge.tenant_id:
```
chlowell, line 38: When is this False? `on_challenge` updates the cached challenge and sets `_last_tenant_id` to match, so it appears they will always be equal.

sdk/keyvault/azure-keyvault-administration/azure/keyvault/administration/_internal/http_challenge.py
```diff
@@ -40,6 +40,10 @@ def __init__(self, request_uri, challenge, response_headers=None):
         if "authorization" not in self._parameters and "authorization_uri" not in self._parameters:
             raise ValueError("Invalid challenge parameters")
 
+        authorization_uri = self.get_authorization_server()
+        # the authoritzation server URI should look something like https://login.windows.net/tenant-id
+        self.tenant_id = authorization_uri.split("/")[-1] or None
```
chlowell, line 45: Just the paranoia talking, but I would take the first segment of the path, not the last.

sdk/keyvault/azure-keyvault-administration/azure/keyvault/administration/_internal/async_challenge_auth_policy.py
mccoyp: Hm, yeah I think you're right. I was confusing this with evaluating the tenant ID of a new challenge, but this is being fetched from the cache so they should always be equal.

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_shared/async_challenge_auth_policy.py
```diff
@@ -15,65 +15,50 @@
...
-
-    async def send(self, request: "PipelineRequest") -> "AsyncHttpResponse":
+    async def on_request(self, request: "PipelineRequest") -> None:
         _enforce_tls(request)
+        challenge = ChallengeCache.get_challenge_for_url(request.http_request.url)
+        if challenge:
+            # Super can handle this. Its cached token, if any, probably isn't from a different tenant, and
+            # it knows the scope to request for a new token. Note that if the vault has moved to a new
+            # tenant since our last request for it, this request will fail.
+            await super().on_request(request)
```
chlowell, line 38: When the tenant baked into the credential doesn't match the vault's tenant, we need to pass the latter in every `get_token()`, and `super` doesn't know anything about it. My first idea is, don't bother with `super`--handle the credential and cached token here.

mccoyp: Just fixed, and extended tests to check for this

sdk/keyvault/azure-keyvault-administration/azure/keyvault/administration/_internal/async_challenge_auth_policy.py
```diff
@@ -67,8 +59,31 @@ async def on_challenge(self, request: "PipelineRequest", response: "PipelineResp
...
 
         return True
+
+    async def _handle_challenge(self, request: "PipelineRequest", challenge: "HttpChallenge") -> None:
+        """Authenticate according to challenge, add Authorization header to request"""
+
+        scope = challenge.get_scope() or challenge.get_resource() + "/.default"
+        await self.authorize_request(request, scope, tenant_id=challenge.tenant_id)
+
+    async def _handle_response(self, request: "PipelineRequest", response: "PipelineResponse") -> "PipelineResponse":
```
chlowell, line 72: What went wrong with `on_challenge`?

sdk/keyvault/azure-keyvault-administration/azure/keyvault/administration/_internal/async_challenge_auth_policy.py
```diff
@@ -21,35 +21,27 @@
...
-            if self._last_tenant_id == challenge.tenant_id:
-                # Super can handle this. Its cached token, if any, probably isn't from a different tenant, and
-                # it knows the scope to request for a new token. Note that if the vault has moved to a new
-                # tenant since our last request for it, this request will fail.
-                await super().on_request(request)
-            else:
-                # acquire a new token because this vault is in a different tenant
-                await self.authorize_request(request, *self._scopes, tenant_id=challenge.tenant_id)
+            # Note that if the vault has moved to a new tenant since our last request for it, this request will fail.
+            await self._handle_challenge(request, challenge)
```
chlowell, line 36: This effectively disables token caching on the policy because nothing uses the cached token anymore--`_handle_challenge` calls `authorize_request`, which replaces the cached token with a new one.

sdk/keyvault/azure-keyvault-administration/azure/keyvault/administration/_internal/async_challenge_auth_policy.py
mccoyp: Adding `_handle_response` was actually a mistake -- I can't really say why I thought this was necessary. It's also responsible for all the playback errors, as I'm sure you noticed if you looked at the pipeline. My bad!

sdk/keyvault/azure-keyvault-certificates/CHANGELOG.md
```diff
@@ -3,12 +3,16 @@
 ## 4.4.0b2 (Unreleased)
 
 ### Features Added
+- Added support for multi-tenant authentication against Managed HSM when using
```
chlowell, line 6:
```suggestion
- Added support for multi-tenant authentication when using
```

sdk/keyvault/azure-keyvault-certificates/azure/keyvault/certificates/_shared/async_challenge_auth_policy.py
```diff
@@ -15,65 +15,59 @@
...
-        self._credential = credential
-        super(AsyncChallengeAuthPolicy, self).__init__(**kwargs)
-
-    async def send(self, request: "PipelineRequest") -> "AsyncHttpResponse":
+    async def on_request(self, request: "PipelineRequest") -> None:
         _enforce_tls(request)
+        challenge = ChallengeCache.get_challenge_for_url(request.http_request.url)
+        if challenge:
+            # Note that if the vault has moved to a new tenant since our last request for it, this request will fail.
+            await self._handle_challenge(request, challenge)
```
chlowell, line 36: It's odd to have `_handle_challenge` and `on_challenge`, and more so when neither calls the other. Consider inlining `_handle_challenge` here.

sdk/keyvault/azure-keyvault-administration/azure/keyvault/administration/_internal/async_challenge_auth_policy.py
```diff
@@ -15,65 +15,59 @@
...
 
-        return response
+    async def on_challenge(self, request: "PipelineRequest", response: "PipelineResponse") -> bool:
+        try:
+            challenge = _update_challenge(request, response)
+            scope = challenge.get_scope() or challenge.get_resource() + "/.default"
+        except ValueError:
+            return False
+
+        self._scopes = (scope,)
```
chlowell, line 56: Do we need to set this now that we don't call `super` for anything?

sdk/keyvault/azure-keyvault-administration/azure/keyvault/administration/_internal/async_challenge_auth_policy.py
```diff
@@ -13,67 +13,67 @@
...
-    from . import HttpChallenge
+    from azure.core.pipeline import PipelineRequest, PipelineResponse
 
 
-class AsyncChallengeAuthPolicy(ChallengeAuthPolicyBase, AsyncHTTPPolicy):
+class AsyncChallengeAuthPolicy(AsyncBearerTokenCredentialPolicy):
     """policy for handling HTTP authentication challenges"""
 
-    def __init__(self, credential: "AsyncTokenCredential", **kwargs: "Any") -> None:
+    def __init__(self, credential: "AsyncTokenCredential", *scopes: str, **kwargs: "Any") -> None:
```
mccoyp, line 35: I added an `__init__` back in because `_need_new_token` made me realize that we shouldn't add any dependencies on private parent class fields. I had been referencing `self._credential` and `self._token` from the parent class before

sdk/keyvault/azure-keyvault-administration/azure/keyvault/administration/_internal/async_challenge_auth_policy.py
```diff
@@ -13,67 +13,67 @@
...
+
+        body = request.context.pop("key_vault_request_data", None)
+        request.http_request.set_text_body(body)  # no-op when text is None
+        await self.authorize_request(request, scope, tenant_id=challenge.tenant_id)
+
+        return True
+
+    @property
+    def _need_new_token(self) -> bool:
+        # pylint:disable=invalid-overridden-method
```
mccoyp, line 78: I debated internally a bit over adding this override as a property. I suppressed this warning because the situation in `azure-core` is likely unintentional, where `_need_new_token` is a property for BearerTokenCredentialPolicy but a method for AsyncBearerTokenCredentialPolicy. Assuming that will be changed in the future, we can remove this suppression, but I figured that makes more sense than re-implementing the property/method pattern here for the sake of consistency

sdk/keyvault/azure-keyvault-administration/azure/keyvault/administration/_internal/challenge_auth_policy.py
```diff
@@ -44,20 +42,10 @@ def _enforce_tls(request):
...
-    challenge_request.headers["Content-Length"] = "0"
-
-    options = copy.deepcopy(request.context.options)
-    context = PipelineContext(request.context.transport, **options)
-
-    return PipelineRequest(http_request=challenge_request, context=context)
+def _get_challenge_scope(challenge):
+    # type: (HttpChallenge) -> str
+    # azure-identity credentials require an AADv2 scope but the challenge may specify an AADv1 resource
+    return challenge.get_scope() or challenge.get_resource() + "/.default"
```
chlowell, line 48: To me, reusing this line isn't worth the function call

sdk/keyvault/azure-keyvault-keys/CHANGELOG.md
```diff
@@ -3,12 +3,16 @@
 ## 4.5.0b5 (Unreleased)
 
 ### Features Added
+- Added support for multi-tenant authentication against Key Vault and Managed HSM when using
```
chlowell, line 6:
```suggestion
- Added support for multi-tenant authentication when using
```

sdk/keyvault/azure-keyvault-secrets/CHANGELOG.md
```diff
@@ -3,12 +3,16 @@
 ## 4.4.0b2 (Unreleased)
 
 ### Features Added
+- Added support for multi-tenant authentication against Managed HSM when using
```
chlowell, line 6:
```suggestion
- Added support for multi-tenant authentication when using
```

sdk/keyvault/azure-keyvault-certificates/CHANGELOG.md
chlowell: still refers to MHSM

chlowell (approved):

PR #21404, opened 2021-10-25
[Cosmos] CosmosDB asynchronous client

simorenoh:
This PR contains the changes for the working language-native asynchronous client for cosmosdb. The async samples that were added showcase how the different functions can be used with async/await, and serve to show end to end functionality of each resource.

Any and all feedback is appreciated!


moderakh (changes requested):
1) as we are forking the implementation code and copying from the non async io code path, we need to have a plan on maintenance. now if there is any bug let's say in the query stack we need to fix this into different codebase.

multiple options to consider for future:

* make the sync api a shim layer on top of async API, rather than having two separate implementations
* deprecated the sync API

2. API `self.client_connection.last_response_headers` may cause indeterministic behaviour, as for example in the context of cross partition query or when there is any concurrency it is not possible to identify really which header response associates with which set of result

3. I don't see any test for the new async API added. Did we add test?

4. in general we should try to follow python culture but at the same time be close as possible to the existing SDKs, you can do a quick cross check between the new python api and the java sdk.
few deviations:
- in java v4 we don't explicitly need to enable cross partition query, (we shouldn't need explicit `enable_cross_partition_query`)
- disable ssl verification, on java we decided to not provide this functionality see here: https://github.com/Azure/azure-sdk-for-java/pull/22990

5. please review some of the options passed in the public contract, we should only have the options which are applicable on each API, for example, populate_query_metrics is only applicable to query_items but that is added to all public contracts (e.g., create_database)

6. we have code for python2 in the new implementation as the new async io api is only for python3 we should remove python2 related code.

sdk/cosmos/azure-cosmos/README.md
```diff
@@ -428,6 +428,84 @@ print(json.dumps(container_props['defaultTtl']))
...
+client = CosmosClient(url, credential=key)
+database_name = 'testDatabase'
+database = client.get_database_client(database_name)
+container_name = 'products'
+container = database.get_container_client(container_name)
+
+async def create_lists():
+    results = await container.query_items(
+            query='SELECT * FROM products p WHERE p.productModel = "Model 2"',
+            enable_cross_partition_query=True)
```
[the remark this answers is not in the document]
simorenoh: does this mean it should be enabled by default if no partition key value is passed in by the user?

sdk/cosmos/azure-cosmos/azure/cosmos/aio/cosmos_client.py
```diff
@@ -0,0 +1,407 @@
...
+            response_hook(self.client_connection.last_response_headers)
+        return result
+
+    @distributed_trace
+    def query_databases(
+        self,
+        query=None,  # type: Optional[str]
+        parameters=None,  # type: Optional[List[str]]
+        enable_cross_partition_query=None,  # type: Optional[bool]
+        max_item_count=None,  # type:  Optional[int]
```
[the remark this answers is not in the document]
simorenoh: This type is meant to be a List of Dicts, thanks for catching that.

This bit, however, seems to vary on the team's implementation for the python sdk's; storage blob and eventhub checkpointstore sdk just pass in the query and send everything else into **kwargs, table sdk sends in a filter and reads the rest from **kwargs, and others like data factory and synapse have their own models:
image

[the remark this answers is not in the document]
simorenoh: s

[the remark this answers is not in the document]
simorenoh: in this case, I think it's worth the time for us to split it out of kwargs since it's information still directly pertinent to the query that will be passed in by the user (query and query params)
 
for comparison, the other keywords that are used by the method from kwargs are: initial_headers, request_options, response_hook, session_token - which relate more to configs and settings, not the query per se

sdk/cosmos/azure-cosmos/azure/cosmos/aio/database.py
```diff
@@ -0,0 +1,768 @@
...
+                indexing_policy=indexing_policy,
+                default_ttl=default_ttl,
+                populate_query_metrics=populate_query_metrics,
+                offer_throughput=offer_throughput,
+                unique_key_policy=unique_key_policy,
+                conflict_resolution_policy=conflict_resolution_policy,
+                analytical_storage_ttl=analytical_storage_ttl
+            )
+
+    def get_container_client(self, container):
```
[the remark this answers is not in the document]
simorenoh: The method currently also takes in a ContainerProxy instance, but I will change that to be just a string since it doesn't really make sense to take in a ContainerProxy and then return it again

sdk/cosmos/azure-cosmos/azure/cosmos/aio/database.py
```diff
@@ -0,0 +1,768 @@
...
+        response_hook = kwargs.pop('response_hook', None)
+
+        result = await self.client_connection.DeleteUser(
+            user_link=self._get_user_link(user), options=request_options, **kwargs
+        )
+        if response_hook:
+            response_hook(self.client_connection.last_response_headers, result)
+
+    @distributed_trace_async
+    async def read_offer(self, **kwargs):
```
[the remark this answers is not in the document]
simorenoh: agreed, will rename `offer` to `throughput`

sdk/cosmos/azure-cosmos/azure/cosmos/aio/container.py
```diff
@@ -0,0 +1,742 @@
...
+        return "<ContainerProxy [{}]>".format(self.container_link)[:1024]
+
+    async def _get_properties(self):
+        # type: () -> Dict[str, Any]
+        if self._properties is None:
+            self._properties = await self.read()
+        return self._properties
+
+    @property
+    async def is_system_key(self):
```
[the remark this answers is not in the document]
simorenoh: Seems to be used to check if partitionKey is system generated or not - although with that said the SDK makes partition key a required argument in container creation and so does the portal, so this could be removed, would like to know more about "system generated" partition keys before removing though

sdk/cosmos/azure-cosmos/azure/cosmos/aio/_cosmos_client_connection_async.py
```diff
@@ -0,0 +1,2492 @@
...
+        # Parses the paths into a list of token each representing a property
+        partition_key_parts = base.ParsePaths(partitionKeyDefinition.get("paths"))
+        # Check if the partitionKey is system generated or not
+        is_system_key = partitionKeyDefinition["systemKey"] if "systemKey" in partitionKeyDefinition else False
+
+        # Navigates the document to retrieve the partitionKey specified in the paths
+        return self._retrieve_partition_key(partition_key_parts, document, is_system_key)
+
+    # Navigates the document to retrieve the partitionKey specified in the partition key parts
+    def _retrieve_partition_key(self, partition_key_parts, document, is_system_key):
```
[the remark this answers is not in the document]
simorenoh: agreed, will take time for refactoring all methods to be consistent afterwards

sdk/cosmos/azure-cosmos/azure/cosmos/aio/container.py
```diff
@@ -0,0 +1,742 @@
...
+            document_link=item_link, new_document=body, options=request_options, **kwargs
+        )
+        if response_hook:
+            response_hook(self.client_connection.last_response_headers, result)
+        return result
+
+    @distributed_trace_async
+    async def delete_item(
+        self,
+        item,  # type: Union[Dict[str, Any], str]
```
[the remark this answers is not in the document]
simorenoh: I'll change it to be `item_id` string only, but scenario 1 is effectively what is being done here

sdk/cosmos/azure-cosmos/azure/cosmos/aio/container.py
```diff
@@ -0,0 +1,742 @@
...
+        parameters=None,  # type: Optional[List[Dict[str, object]]]
+        partition_key=None,  # type: Optional[Any]
+        enable_cross_partition_query=None,  # type: Optional[bool]
+        max_item_count=None,  # type: Optional[int]
+        enable_scan_in_query=None,  # type: Optional[bool]
+        populate_query_metrics=None,  # type: Optional[bool]
+        **kwargs  # type: Any
+    ):
+        # type: (...) -> Iterable[Dict[str, Any]]
+        """Return all results matching the given `query`.
```
moderakh, line 302: I can't find any api for returning response headers other than `self.client_connection.last_response_headers` for any of the APIs. 

what's the workaround for getting response headers, assuming that's something python users would need?

simorenoh: that's all the sdk currently has as of now - once the response is returned after the function gets sent into _PipelineRunFunction, the last response headers get returned along with the response body

sdk/cosmos/azure-cosmos/azure/cosmos/aio/container.py
```diff
@@ -0,0 +1,742 @@
...
+            response_hook(self.client_connection.last_response_headers, items)
+        return items
+
+    @distributed_trace_async
+    async def query_items(
+        self,
+        query,  # type: str
+        parameters=None,  # type: Optional[List[Dict[str, object]]]
+        partition_key=None,  # type: Optional[Any]
+        enable_cross_partition_query=None,  # type: Optional[bool]
```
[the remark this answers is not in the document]
simorenoh: outcome of this is we branch on a pk value being passed in to set this option
- need to look into upgrading query pipeline

sdk/cosmos/azure-cosmos/azure/cosmos/aio/container.py
moderakh: the last header response is not well defined when iterating through the results of a query, also it is not possible to associate the items in the result to the response header (there is no clear boundary).

@simorenoh could you check which pattern the other azure python SDKs are using for this?

sdk/cosmos/azure-cosmos/azure/cosmos/aio/user.py
```diff
@@ -0,0 +1,295 @@
...
+        # type: (Any) -> Dict[str, Any]
+        """Read user propertes.
+
+        :keyword Callable response_hook: A callable invoked with the response metadata.
+        :returns: A dictionary of the retrieved user properties.
+        :raises ~azure.cosmos.exceptions.CosmosHttpResponseError: If the given user couldn't be retrieved.
+        :rtype: dict[str, Any]
+        """
+        request_options = build_options(kwargs)
+        response_hook = kwargs.pop('response_hook', None)
```
moderakh, line 82: we need to check what the use case for response_hook is.
If we are already returning the result why do we need resonse_hook

moderakh (approved):
Thanks @simorenoh we discussed comments in the review session.

Please after addressing the essential ones. track addressing the rest after the first preview.
I am good on the PR and the first preview, assuming that we agree that after the first preview we may need to do some breaking changes to further improve the public APIs, and we may have to do a few more previews before finalizing the public contract.

Thanks @simorenoh  for driving this.

sdk/cosmos/azure-cosmos/azure/cosmos/aio/container.py
```diff
@@ -0,0 +1,714 @@
...
+            database_or_container_link=self.container_link, document=body, options=request_options, **kwargs
+        )
+        if response_hook:
+            response_hook(self.client_connection.last_response_headers, result)
+        return result
+
+    @distributed_trace_async
+    async def read_item(
+        self,
+        item,  # type: Union[str, Dict[str, Any]]
```
moderakh, line 201: I see two options here:
1. we pass item as json and SDK infers both id and partitionKey
2. we pass itemId as string and partitionKey explicitly.

We should not mix the two options. same for the other APIs as well. We should look into this after the first preview.

sdk/cosmos/azure-cosmos/azure/cosmos/aio/container.py
```diff
@@ -0,0 +1,714 @@
...
+
+        items = self.client_connection.ReadItems(
+            collection_link=self.container_link, feed_options=feed_options, response_hook=response_hook, **kwargs
+        )
+        if response_hook:
+            response_hook(self.client_connection.last_response_headers, items)
+        return items
+
+    @distributed_trace_async
+    async def query_items(
```
moderakh, line 270: we should look into if our async io implementation truly will parallelize the cross partition query execution? or if the execution is serial.

please test and check.

sdk/cosmos/azure-cosmos/azure/cosmos/aio/container.py
```diff
@@ -0,0 +1,714 @@
...
+            response_hook(self.client_connection.last_response_headers, self._properties)
+
+        return cast('Dict[str, Any]', self._properties)
+
+    @distributed_trace_async
+    async def create_item(
+        self,
+        body,  # type: Dict[str, Any]
+        pre_trigger_include=None,  # type: Optional[str]
+        post_trigger_include=None,  # type: Optional[str]
```
moderakh, line 155: after the first preview, we should consider making commonly used parameters as explicit parameters and make non commonly used one as non-explicit in the kwargs

sdk/cosmos/azure-cosmos/README.md
```diff
@@ -428,6 +432,83 @@ print(json.dumps(container_props['defaultTtl']))
...
+        await container.upsert_item({
+                'id': 'item{0}'.format(i),
+                'productName': 'Widget',
+                'productModel': 'Model {0}'.format(i)
+            }
+        )
+    await client.close()
+```
+
+It is also worth pointing out that the asynchronous client has to be closed manually after its use, either by initializing it using async with or calling the close() method directly like shown above.
```
gahl-levy, line 462: @simorenoh 
> closed manually after its use
Does it need to be closed after every function it's used in (like above), or can the client be passed around safely when doing things async?

sdk/cosmos/azure-cosmos/README.md
```diff
@@ -428,6 +432,83 @@ print(json.dumps(container_props['defaultTtl']))
...
 
+### Using the asynchronous client
+
+The asynchronous cosmos client looks and works in a very similar fashion to the already existing client, with the exception of its package within the sdk and the need of using async/await keywords in order to interact with it.
+
+```Python
+from azure.cosmos.aio import CosmosClient
+import os
+
+url = os.environ['ACCOUNT_URI']
```
gahl-levy, line 443: Constants such as these should be in all caps: https://www.python.org/dev/peps/pep-0008/#constants

sdk/cosmos/azure-cosmos/README.md
```diff
@@ -428,6 +432,83 @@ print(json.dumps(container_props['defaultTtl']))
...
+It is also worth pointing out that the asynchronous client has to be closed manually after its use, either by initializing it using async with or calling the close() method directly like shown above.
+
+```Python
+from azure.cosmos.aio import CosmosClient
+import os
+
+url = os.environ['ACCOUNT_URI']
+key = os.environ['ACCOUNT_KEY']
+database_name = 'testDatabase'
+container_name = 'products'
```
gahl-levy, line 471: ditto. I was a bit confused when I saw these being referred in the function below without realizing they are constants.

sdk/cosmos/azure-cosmos/README.md
```diff
@@ -428,6 +432,83 @@ print(json.dumps(container_props['defaultTtl']))
...
+async with CosmosClient(url, credential=key) as client:
+    database = client.get_database_client(database_name)
+    container = database.get_container_client(container_name)
+    for i in range(1, 10):
+        await container.upsert_item({
+                'id': 'item{0}'.format(i),
+                'productName': 'Widget',
+                'productModel': 'Model {0}'.format(i)
+            }
+        )
```
gahl-levy, line 482: Do we need to close the client here too?

await client.close()

sdk/cosmos/azure-cosmos/README.md
```diff
@@ -428,6 +432,83 @@ print(json.dumps(container_props['defaultTtl']))
...
+```
+
+### Queries with the asynchronous client
+
+Queries work the same way for the most part, with one exception being the absence of the `enable_cross_partition` flag in the request; queries without a specified partition key value will now by default atempt to do a cross partition query. Results can be directly iterated on, but because queries made by the asynchronous client return AsyncIterable objects, results can't be cast into lists directly; instead, if you need to create lists from your results, use Python's list comprehension to populate a list:
```
gahl-levy, line 487:
```suggestion
Queries work the same way for the most part, with one exception being the absence of the `enable_cross_partition` flag in the request; queries without a specified partition key value will now by default attempt to do a cross partition query. Results can be directly iterated on, but because queries made by the asynchronous client return AsyncIterable objects, results can't be cast into lists directly; instead, if you need to create lists from your results, use Python's list comprehension to populate a list:
```

gahl-levy: 1. Is this the reason that we don't have an await when querying? Should we make that more explicit?
2. Why do we need the word "async" again in the function
3. IIRC point queries operate differently. Should we add an example and explain the differentiation?

sdk/cosmos/azure-cosmos/README.md
```diff
@@ -428,6 +432,83 @@ print(json.dumps(container_props['defaultTtl']))
...
+
+```Python
+from azure.cosmos.aio import CosmosClient
+import os
+
+url = os.environ['ACCOUNT_URI']
```
gahl-levy, line 493: ditto on the constants

sdk/cosmos/azure-cosmos/README.md
simorenoh: It can be passed around safely, but needs to be manually disposed of once you're done using it - the alternative to this being to use the client using a `with` statement:
`async with CosmosClient(url, key) as client:` and then starting your cosmos db logic within that context

sdk/cosmos/azure-cosmos/README.md
simorenoh: this is that second way to initialize I had mentioned above - the way it is done here, it gets disposed once the `with` statement gets exited

sdk/cosmos/azure-cosmos/README.md
simorenoh: 1. Yeah this is that point I had brought up with you before, I feel like I tried to explain it but I'm not sure it's completely understandable... let's talk on this to reach a conclusion.
2. We need the word async in the function definition not because of the query, but because we do an `async for` loop right after - any function that has either `await` or `async` within has to be defined as async by python
3. This is the case even for the normal sync client, and I believe should be documented in the actual Cosmos docs (not just SDK), found this for instance: https://devblogs.microsoft.com/cosmosdb/point-reads-versus-queries/

sdk/cosmos/azure-cosmos/README.md
```diff
@@ -428,6 +432,83 @@ print(json.dumps(container_props['defaultTtl']))
...
+key = os.environ['ACCOUNT_KEY']
+client = CosmosClient(url, credential=key)
+database_name = 'testDatabase'
+database = client.get_database_client(database_name)
+container_name = 'products'
+container = database.get_container_client(container_name)
+
+async def create_lists():
```
gahl-levy, line 501: Do we need to close the client in this example too? I'm not seeing "with" being used either.

sdk/cosmos/azure-cosmos/README.md
```diff
@@ -428,6 +432,83 @@ print(json.dumps(container_props['defaultTtl']))
...
+    results = container.query_items(
+            query='SELECT * FROM products p WHERE p.productModel = "Model 2"')
+
+    # Iterating directly on results
+    async for item in results:
+        print(item)
+
+    # Making a list from the results
+    item_list = [item async for item in results]
+```
```
gahl-levy, line 511:
```suggestion
await client.close()
```
```

File: sdk/cosmos/azure-cosmos/README.md [hunk 22]
```diff
@@ -428,6 +432,83 @@ print(json.dumps(container_props['defaultTtl']))
 
 For more information on TTL, see [Time to Live for Azure Cosmos DB data][cosmos_ttl].
 
+### Using the asynchronous client
+
+The asynchronous cosmos client looks and works in a very similar fashion to the already existing client, with the exception of its package within the sdk and the need of using async/await keywords in order to interact with it.
```
gahl-levy, line 437:
```suggestion
The asynchronous cosmos client is a separate client that looks and works in a similar fashion to the existing synchronous client. However, the async client needs to be imported separately and its methods need to be used with the async/await keywords.
```

sdk/cosmos/azure-cosmos/README.md
```diff
@@ -428,6 +432,83 @@ print(json.dumps(container_props['defaultTtl']))
...
+URL = os.environ['ACCOUNT_URI']
+KEY = os.environ['ACCOUNT_KEY']
+client = CosmosClient(URL, credential=KEY)
+DATABASE_NAME = 'testDatabase'
+database = client.get_database_client(DATABASE_NAME)
+CONTAINER_NAME = 'products'
+container = database.get_container_client(CONTAINER_NAME)
+
+async def create_items():
+    for i in range(1, 10):
```
gahl-levy, line 452:
```suggestion
    for i in range(10):
```

sdk/cosmos/azure-cosmos/README.md
```diff
@@ -428,6 +432,83 @@ print(json.dumps(container_props['defaultTtl']))
...
+        await container.upsert_item({
+                'id': 'item{0}'.format(i),
+                'productName': 'Widget',
+                'productModel': 'Model {0}'.format(i)
+            }
+        )
+    await client.close()
```
gahl-levy, line 459:
```suggestion
    await client.close() # the async client must be closed manually if it's not initialized in a with statement
```

sdk/cosmos/azure-cosmos/README.md
gahl-levy: @simorenoh  What happens if these async clients are not closed properly? Memory leak or something else?

sdk/cosmos/azure-cosmos/README.md
```diff
@@ -428,6 +432,83 @@ print(json.dumps(container_props['defaultTtl']))
...
+```Python
+from azure.cosmos.aio import CosmosClient
+import os
+
+URL = os.environ['ACCOUNT_URI']
+KEY = os.environ['ACCOUNT_KEY']
+DATABASE_NAME = 'testDatabase'
+CONTAINER_NAME = 'products'
+
+async with CosmosClient(URL, credential=KEY) as client:
```
gahl-levy, line 473:
```suggestion
async with CosmosClient(URL, credential=KEY) as client: # the with statement will automatically close the async client
```

sdk/cosmos/azure-cosmos/README.md
```diff
@@ -428,6 +432,83 @@ print(json.dumps(container_props['defaultTtl']))
...
+    database = client.get_database_client(DATABASE_NAME)
+    container = database.get_container_client(CONTAINER_NAME)
+    for i in range(1, 10):
```
gahl-levy, line 476:
```suggestion
    for i in range(10):
```

sdk/cosmos/azure-cosmos/README.md
```diff
@@ -428,6 +432,83 @@ print(json.dumps(container_props['defaultTtl']))
...
+                'id': 'item{0}'.format(i),
+                'productName': 'Widget',
+                'productModel': 'Model {0}'.format(i)
+            }
+        )
+```
+
+### Queries with the asynchronous client
+
+Queries work the same way for the most part, with one exception being the absence of the `enable_cross_partition` flag in the request; queries without a specified partition key value will now by default attempt to do a cross partition query. Results can be directly iterated on, but because queries made by the asynchronous client return AsyncIterable objects, results can't be cast into lists directly; instead, if you need to create lists from your results, use Python's list comprehension to populate a list:
```
gahl-levy, line 487:
```suggestion
Unlike the synchronous client, the async client does not have an `enable_cross_partition` flag in the request. Queries without a specified partition key value will attempt to do a cross partition query by default. 

Query results can be iterated, but query results return an asynchronous iterator. This means that each object from the iterator is already being "awaited" and does not contain that actual query result, only an awaited object to asynchronously retrieve it. 

These results can't be cast into lists directly; instead, if you need to create lists from your results, use Python's list comprehension to populate a list:
```

gahl-levy: @simorenoh  Not sure if "awaited object" is the correct way to describe it. We should probably also add an example of what the results iterator returns.

sdk/cosmos/azure-cosmos/README.md
```diff
@@ -428,6 +432,83 @@ print(json.dumps(container_props['defaultTtl']))
...
+DATABASE_NAME = 'testDatabase'
+database = client.get_database_client(DATABASE_NAME)
+CONTAINER_NAME = 'products'
+container = database.get_container_client(CONTAINER_NAME)
+
+async def create_lists():
+    results = container.query_items(
+            query='SELECT * FROM products p WHERE p.productModel = "Model 2"')
+
+    # Iterating directly on results
```
gahl-levy, line 505:
```suggestion
    # Iterating directly on results
    item_list = []
```

sdk/cosmos/azure-cosmos/README.md
```diff
@@ -428,6 +432,83 @@ print(json.dumps(container_props['defaultTtl']))
...
+    async for item in results:
+        print(item)
```
gahl-levy, line 507:
```suggestion
        item_list.append(item)
```

sdk/cosmos/azure-cosmos/README.md
```diff
@@ -428,6 +432,83 @@ print(json.dumps(container_props['defaultTtl']))
...
+
+    # Making a list from the results
```
gahl-levy, line 509:
```suggestion
    # Asynchronously creates a complete list of the actual query results
```

sdk/cosmos/azure-cosmos/README.md
simorenoh: I would have to test, but I assume that is the case - you also get an error in your code stating the client was not properly disposed of

sdk/cosmos/azure-cosmos/README.md
```diff
@@ -428,7 +432,90 @@ print(json.dumps(container_props['defaultTtl']))
...
+                'productModel': 'Model {0}'.format(i)
+            }
+        )
+```
+
+### Queries with the asynchronous client
+
+Unlike the synchronous client, the async client does not have an `enable_cross_partition` flag in the request. Queries without a specified partition key value will attempt to do a cross partition query by default. 
+
+Query results can be iterated, but query results return an asynchronous iterator. This means that each object from the iterator is already being "awaited" and does not contain that actual query result, only an awaitable object to asynchronously retrieve it.
```
simorenoh, line 489:
```suggestion
Query results can be iterated, but the query's raw output returns an asynchronous iterator. This means that each object from the iterator is an awaitable object, and does not yet contain the true query result. In order to obtain the query results you can use an async for loop, which awaits each result as you iterate on the object, or manually await each query result as you iterate over the asynchronous iterator.
```

sdk/cosmos/azure-cosmos/README.md
```diff
@@ -428,7 +432,90 @@ print(json.dumps(container_props['defaultTtl']))
...
+        )
+```
+
+### Queries with the asynchronous client
+
+Unlike the synchronous client, the async client does not have an `enable_cross_partition` flag in the request. Queries without a specified partition key value will attempt to do a cross partition query by default. 
+
+Query results can be iterated, but query results return an asynchronous iterator. This means that each object from the iterator is already being "awaited" and does not contain that actual query result, only an awaitable object to asynchronously retrieve it. 
 
+These results can't be cast into lists directly; instead, if you need to create lists from your results, use an async for loop or Python's list comprehension to populate a list:
```
simorenoh, line 491:
```suggestion
Since the query results are an asynchronous iterator, they can't be cast into lists directly; instead, if you need to create lists from your results, use an async for loop or Python's list comprehension to populate a list:
```

tjprescott (changes requested):
One small change.

sdk/cosmos/azure-cosmos/CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
-## 4.2.1 (Unreleased)
+## 4.3.0 (Unreleased)
+**New features**
+- Added language native async i/o client
 
+## 4.2.1 (Unreleased)
```
simorenoh, line 5:
```suggestion
```

tjprescott (approved):

Issue #18622, opened 2021-05-10
[ACR] Add Perf Tests


annelo-msft:
Add perf tests for ACR types using language-appropriate approach

.NET perf test scenarios: https://github.com/Azure/azure-sdk-for-net/tree/main/sdk/containerregistry/Azure.Containers.ContainerRegistry/perf/Azure.Containers.ContainerRegistry.Perf/Scenarios

YalinLi0312:
Fixed.

PR #21421, opened 2021-10-26
[ACR] Add performance tests

YalinLi0312:
Resolves issue https://github.com/Azure/azure-sdk-for-python/issues/18622


sdk/containerregistry/azure-containerregistry/tests/perfstress_tests/list_artifacts.py
```diff
@@ -0,0 +1,35 @@
...
+
+        # Create clients
+        account_url = self.get_from_env("CONTAINERREGISTRY_ANONREGISTRY_ENDPOINT")
+        audience = "https://management.azure.com"
+        self.anon_client = ContainerRegistryClient(endpoint=account_url, credential=None, audience=audience)
+        self.async_anon_client = AsyncContainerRegistryClient(endpoint=account_url, credential=None, audience=audience)       
+        self.repository = "node"
+
+    async def global_setup(self):
+        await super().global_setup()
```
annatisch, line 23: Looks like we don't need this?

sdk/containerregistry/azure-containerregistry/tests/perfstress_tests/list_artifacts.py
```diff
@@ -0,0 +1,35 @@
...
+
+    async def close(self):
+        await self.async_anon_client.close()
+        await super().close()
+
+    def run_sync(self):
+        for manifest in self.anon_client.list_manifest_properties(self.repository):
+            self.anon_client.get_manifest_properties(self.repository, manifest.digest)
```
annatisch, line 31: This test seems to be combining multiple operations in a single perf-run.
We don't normally do that, as it makes it harder to see and debug where perf loss is happening.
Is there a reason you've done it here? Is this the scenario that's being tested across languages?

sdk/containerregistry/azure-containerregistry/tests/perfstress_tests/list_artifacts.py
```diff
@@ -0,0 +1,35 @@
...
+
+    async def global_setup(self):
+        await super().global_setup()
+
+    async def close(self):
+        await self.async_anon_client.close()
+        await super().close()
+
+    def run_sync(self):
+        for manifest in self.anon_client.list_manifest_properties(self.repository):
```
annatisch, line 30: How many artifacts is this listing?
The test needs to be repeatable, and comparable across languages - so the actual resources that we run against should have fixed parameters (e.g. "lists X manifest properties in Y seconds")

annatisch: If these resources need to be preconfigured, and not run dynamically - please update the the readme to specify exactly what resources are needed in order for this test to be run in a repeatable way.

YalinLi0312: It needs to pull one image to repo before running this test, it's in README.md section _Import docker images into the newly created registry_, but didn't add a description. (Will do that).

YalinLi0312: > How many artifacts is this listing? The test needs to be repeatable, and comparable across languages - so the actual resources that we run against should have fixed parameters (e.g. "lists X manifest properties in Y seconds")

1, we're using the same artifact across languages.

sdk/containerregistry/azure-containerregistry/tests/perfstress_tests/list_artifacts.py
YalinLi0312: Yes, I followed what they did in other languages, and both tests are being tested across languages. i.e. [JS](https://github.com/Azure/azure-sdk-for-js/blob/main/sdk/containerregistry/perf-tests/container-registry/test/listArtifacts.spec.ts#L15), [.NET](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/containerregistry/Azure.Containers.ContainerRegistry/perf/Azure.Containers.ContainerRegistry.Perf/Scenarios/ListArtifacts.cs#L45)

sdk/containerregistry/azure-containerregistry/tests/perfstress_tests/list_artifacts.py
YalinLi0312: Right, can get rid of the global_setup here since it's not set up anything.

sdk/containerregistry/azure-containerregistry/tests/perfstress_tests/list_artifacts.py
jeremymeng: JS is only calling service in the listing method, the for loop is only doing local operation `getArtifact()` to create client side helper.  Since Python is using a flatter hierarchy, `get_manifest_properties()` is likely different from other languages' `getArtifact()`

annatisch: Thanks @jeremymeng!
In that case @YalinLi0312 can you made sure that `get_manifest_properties` is either not making any service requests - or replace it with something that doesn't?
You can check the `ListBlobsTest` for an example - the just do an empty loop using `pass`

YalinLi0312: By checking the src code, `get_manifest_properties` does making service requests([link](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/containerregistry/azure-containerregistry/azure/containerregistry/_generated/operations/_container_registry_operations.py#L1491)). And same in `get_repository_properties`. In this case, I think python should have 4 separate tests for each operation.

annatisch (approved):

PR #21426, opened 2021-10-26
Handling Proper Defaulting of `BuildTargetingString` within nested templates

scbedd:
We want to ensure that setting a `BuildTargetingString` at queue time is available. This enables us to exclude problematic packages from a run when attempting to release another package in the same service.

Of course, the ideal is that we never have any red pipelines, but the fact of the matter is sometimes random stuff crops up.

[The original "solution"](https://github.com/Azure/azure-sdk-for-python/blob/a62fe6624baef5eaf62cec5b0c94df2e883a8b3e/eng/pipelines/templates/stages/archetype-sdk-client.yml#L77) that I came up with is not working properly. I _thought_ I had tested it, but obviously I missed something during verification.

The core problem is that the way that `parameters` resolve. Parameters are resolved at _compile time_, so even if you set up something at the very top level of your build yaml like I linked above:

image

While _within this template scope_ we can use `$(BuildTargetingStringValue)`. If we pass that to _another_ template, it immediately reverts to the `default` parameter argument. For reasons of how the templates are compiled.

This PR just recognizes and deals with that fact. So essentially:

1. Run the `resolve-targeting-string.yml` if you need to _use_ resolved `BuildTargetingString` in your current template. After running template `resolve-targeting-string.yml`, just reference to the coalesced value by `$(TargetingString)`
2. Otherwise, just pass `${{ paramters.BuildTargetingString }}` to other template calls. Given that resolution will default anyway, might as well make it explicit.

Once `resolve-targeting-string.yml` runs _once_ in the build, it won't run again.

### Test Runs

* UnTargeted, Standard is the standard build for this PR.
* [Targeted, Standard](https://dev.azure.com/azure-sdk/public/_build/results?buildId=1167268&view=results)
* [Untargeted, Daily Dev](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=1167266&view=results)
* [Targeted, Daily Dev](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=1167267&view=results)


weshaggard (approved):

doc/eng_sys_checks.md
```diff
@@ -31,6 +31,20 @@ In both `public` and `internal` projects, all builds allow a filter to be introd
    1. For example, setting filter string `azure-mgmt-*` will filter a build to only management packages. A value of `azure-keyvault-secrets` will result in only building THAT specific package.
 3. Once it's set, run the build!
 
+### Within Template YML
```
weshaggard, line 34: Might also be worth adding a note about how to override this at queue time as needed so you have a place to point folks to for that.

scbedd: That's actually covered directly above this new section that calls out the "how". I'll amend the wording here to make it more explicit that this is describing _where_ in the template yaml this happens.

scbedd: I actually decided that having this detail here when we explicitly call out these same locations in code comments will just confuse folks. I've cut the addition entirely.

Issue #21462, opened 2021-10-28
[Schema Registry] bug bash - single method doesn't take logging_enable keyword argument


yunhaoling:
when calling register_schema with logging_enable:
```
properties = schema_registry_client.register_schema(GROUP_NAME, "testnew", SCHEMA_STRING, "avro", logging_enable=True)
```

it throws:
```
Traceback (most recent call last):
  File "D:\Projects\azure-sdk-for-python\sdk\schemaregistry\azure-schemaregistry\bugbash\bugbash.py", line 74, in <module>
    properties = schema_registry_client.register_schema(GROUP_NAME, "testnew", SCHEMA_STRING, "avro", logging_enable=True)
  File "C:\Users\yuling\Anaconda3\envs\srbugbash\lib\site-packages\azure\schemaregistry\_schema_registry_client.py", line 117, in register_schema
    request = schema_rest.build_register_request(
  File "C:\Users\yuling\Anaconda3\envs\srbugbash\lib\site-packages\azure\schemaregistry\_generated\rest\schema\_request_builders_py3.py", line 185, in build_register_request
    return HttpRequest(
  File "C:\Users\yuling\Anaconda3\envs\srbugbash\lib\site-packages\azure\core\rest\_rest_py3.py", line 122, in __init__
    raise TypeError(
TypeError: You have passed in kwargs 'logging_enable' that are not valid kwargs.
```

because keyword argument is not passed properly

Issue #21177, opened 2021-10-08
[Schema Registry] full pass over docstring


swathipil:
for sync/async SR + Avro
- [ ] Are all params documented?
- [ ] Is priority order reasonable
- [ ] Return vals + descriptions
- [ ] Exceptions/error modes documented?
- [ ] #types + in-docstring + inline types?

- [ ] ask for api approval: https://apiview.dev/Assemblies/Review/21d5a1c19a6845cf9549faf0e30db2e1

PR #21488, opened 2021-10-29
[SchemaRegistry] fix logging_enable bug

swathipil:
fixes: #21462
from bash here: #21177


sdk/schemaregistry/azure-schemaregistry/azure/schemaregistry/_schema_registry_client.py
```diff
@@ -123,7 +124,7 @@ def register_schema(
             **kwargs
```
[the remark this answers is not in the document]
yunhaoling: I think the question here is that some kwargs may be used `send_request` while some may be used in `build_query_id_by_content_request`.

from what I observe is that `build_query_id_by_content_request` checks whether input keywoard arguments is valid for building a request -- keyword arguments for `send_request` are **invalid** for building a request.

@iscai-msft , do you have any suggestion on this situation? we have a mixed keyword argument bag in which some of them are for sending request while some are for building request, do we need to have a kinda black/white list (popping out keyword argument first)?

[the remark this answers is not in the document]
iscai-msft: ah i see. Sorry we're stricter here, but we wanted to make sure no unwanted kwargs went to HttpRequest. For now, you can use `HttpRequest`'s kwargs to determine which ones are good to pass to build_x_request https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/core/azure-core/azure/core/rest/_rest_py3.py#L91

sdk/schemaregistry/azure-schemaregistry/azure/schemaregistry/_schema_registry_client.py
```diff
@@ -114,16 +114,18 @@ def register_schema(
         except AttributeError:
             pass
 
+        http_request_keywords = ["params", "headers", "json", "data", "files"]
+        http_request_kwargs = {key: kwargs.pop(key, None) for key in http_request_keywords if key in kwargs}
```
swathipil, line 118: This is a little repetitive so I can move it out to a helper method in utils.

yunhaoling: yes, let's move it to helper.
also I think we could put `"content_type" into the `http_request_keywords` as well, so that we don't need the extra `content_type=kwargs.pop("content_type", "application/json")` when building request

swathipil: quick clarification: do you mean we should do this `content_type=kwargs.pop("content_type", "application/json")` in the util method? or that we should do the following? so that content_type is set to None by default unless included? like so:
`
http_request_keywords = ["content_type", "params", "headers", "json", "data", "files"]
http_request_kwargs = {key: kwargs.pop(key, None) for key in http_request_keywords if key in kwargs}
`

swathipil: I'm not sure, in either case, that this would work b/c `get_schema` does not take `content_type` as a kwarg to the build_request method.

yunhaoling: oh, sorry for not making it clear, depending on the current implementation, I think it could be done in the _schema_registry_client.py (not util.py) like the following:
```python
        http_request_kwargs = get_http_request_kwargs(content_type=kwargs.pop("content_type", "application/json"), kwargs)
        request = schema_rest.build_register_request(
            group_name=group_name,
            schema_name=name,
            content=schema_definition,
            serialization_type=format,
            **http_request_kwargs
        )
```

sdk/schemaregistry/azure-schemaregistry/samples/sync_samples/schema_registry.py
```diff
@@ -76,7 +76,7 @@ def register_schema(client, group_name, name, schema_string, format):
 
 def get_schema_by_id(client, id):
     print("Getting schema by id...")
-    schema = client.get_schema(id)
+    schema = client.get_schema(id, logging_enable=True)
```
yunhaoling, line 79: I think for samples we could drop logging_enable?

sdk/schemaregistry/azure-schemaregistry/azure/schemaregistry/_schema_registry_client.py
swathipil: so according the the HttpRequest class that Izzy mentioned, `content_type` is not a keyword: https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/core/azure-core/azure/core/rest/_rest_py3.py#L91

so I don't think the above would work?

yunhaoling: the above would technically work as build_get_by_id_request would pop the `content_type`.
you're right from design perspective, `HttpRequest` doesn't take `content_type` (content_type is part of the header).

but I feel a bit strange about the usage of the keyword arguments.. maybe it's because `content_type` is not documented in the docstring..

let's just keep the implementation for now

yunhaoling (approved):

PR #21505, opened 2021-10-30
Generate and publish API stub file for Python

praveenkuttappan:
Generate api stub file and publish it as artifact in package folder


scripts/devops_tasks/tox_harness.py
```diff
@@ -435,6 +435,11 @@ def prep_and_run_tox(targeted_packages, parsed_args, options_array=[]):
         if parsed_args.tenvparallel:
             tox_execution_array.extend(["-p", "all"])
 
+        if parsed_args.tox_env == "apistub":
```
praveenkuttappan, line 438: I prefer not to do this if tox has an option to choose a specific arg from a list of args in {posargs} and pass to it to command. Something like {posargs:out_path} but this didn't work.

eng/pipelines/templates/steps/analyze.yml
```diff
@@ -108,12 +108,6 @@ steps:
       artifactName: 'packages'
       targetPath: $(Build.ArtifactStagingDirectory)
 
-  - template: ../steps/run_apistub.yml
-    parameters:
-      ServiceDirectory: ${{ parameters.ServiceDirectory }}
-      TestMarkArgument: ${{ parameters.TestMarkArgument }}
```
weshaggard, line 114: Do we need any of these other args in the new usages?

praveenkuttappan: We have removed this step in analyze and only service directory is required in build job. I will remove other params from template itself.

praveenkuttappan: Done.

praveenkuttappan:
@scbedd Can you please have a look at this PR when you get a chance?

scripts/devops_tasks/tox_harness.py
scbedd: I think this is fine. I see your difficulty.

scbedd (approved):

eng/tox/run_apistubgen.py
```diff
@@ -56,4 +63,9 @@ def get_package_wheel_path(pkg_root):
     if not pkg_path:
         pkg_path = args.target_package
 
-    check_call(["apistubgen", "--pkg-path", pkg_path,], cwd=args.work_dir)
+    cmds = ["apistubgen", "--pkg-path", pkg_path]
```
weshaggard, line 66: For my understanding where do we install the apistubgen code from azure-sdk-tools repo?

PR #21520, opened 2021-11-01
support charset_normalizer

xiangyan99:


sdk/core/azure-core/CHANGELOG.md
```diff
@@ -1,6 +1,6 @@
 # Release History
 
-## 1.19.1 (Unreleased)
+## 1.20.0 (2021-11-01)
```
iscai-msft, line 3: Since all that's being released are bug fixes, I think this should be 1.19.1

iscai-msft (approved):

sdk/core/azure-core/CHANGELOG.md
xiangyan99: I thought about it. The problem is we did add a new feature - if `charset_normalizer` is installed but `chardet` is not, we still work. Which means some code that works in 1.20.0 does not work in 1.19.0.

PR #21525, opened 2021-11-01
regenerate artifacts

xiangyan99:


sdk/synapse/azure-synapse-managedprivateendpoints/setup.py
```diff
@@ -71,6 +71,7 @@
         'Programming Language :: Python :: 3.7',
         'Programming Language :: Python :: 3.8',
         'Programming Language :: Python :: 3.9',
+        'Programming Language :: Python :: 3.10',
```
iscai-msft, line 74: you should bump azure-core to a verssion that contains azure.core.rest, I think we might as well bump itto `1.20.0` to get all of the fixes

iscai-msft (approved):
besides min azure-core dep looks good to me

PR #21527, opened 2021-11-01
Added List System Containers

tasherif-msft:


sdk/storage/azure-storage-blob/tests/settings/settings_fake.py
```diff
@@ -22,7 +22,7 @@
 CLIENT_SECRET = "00000000-0000-0000-0000-000000000000"
 
 ACCOUNT_URL_SUFFIX = 'core.windows.net'
-RUN_IN_LIVE = "False"
-SKIP_LIVE_RECORDING = "True"
+RUN_IN_LIVE = "True"
```
seanmcc-msft, line 25: Do you need to revert these 2 lines?

seanmcc-msft (commented):
What about tests for Data Lake?

sdk/storage/azure-storage-blob/tests/settings/settings_fake.py
tasherif-msft: yep yep

tasherif-msft:
TODO: add datalake tests

seanmcc-msft (approved):

PR #21529, opened 2021-11-01
[rest] add kwargs to iter_bytes and iter_raw

iscai-msft:
In the future, we might think of adding additional arguments to the `iter_` metods (i.e., `chunk_size`). Thus, we want to add kwargs to the methods right now so our hands aren't tied later on


annatisch (approved):
LGTM :)

xiangyan99 (approved):

PR #21530, opened 2021-11-01
[formrecognizer] Add parent references for to_dict and from_dict methods

### LINKED ISSUE #21470: [formrecognizer] Add parent element references in dict methods on models
Will need to store parent references in to and from dict methods to ensure proper functionality when users call the get_children methods.

catalinaperalta:
Fixes #21470 

Adding parent info in the to and from dict methods allows continued functionality for `get_words` and `get_lines` calls on models that have been converted.

### DIFF
--- a/sdk/formrecognizer/azure-ai-formrecognizer/azure/ai/formrecognizer/_models.py
+++ b/sdk/formrecognizer/azure-ai-formrecognizer/azure/ai/formrecognizer/_models.py
@@ -2258,6 +2258,7 @@ class AnalyzedDocument(object):
             if self.fields
             else {},
             "confidence": self.confidence,
+            "_parent": self._parent,
         }
 
     @classmethod
@@ -2281,6 +2282,7 @@ class AnalyzedDocument(object):
             if data.get("fields")
             else {},
             confidence=data.get("confidence", None),
+            _parent=data.get("_parent", None),
         )
 
     def get_words(self, **kwargs):  # pylint: disable=unused-argument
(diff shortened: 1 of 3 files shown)


sdk/formrecognizer/azure-ai-formrecognizer/azure/ai/formrecognizer/_models.py
```diff
@@ -2258,6 +2258,7 @@ def to_dict(self):
             if self.fields
             else {},
             "confidence": self.confidence,
+            "_parent": self._parent,
```
catalinaperalta, line 2261: Thoughts on special processing for this? At the moment this works with the get_words/get_lines calls.

kristapratico: This is setting the generated model to the `_parent` key, though, so not converting the whole response to_dict as promised.

Check with Anna, but I don't think we want to include the _parent in the to_dict/from_dict, similar to how Will was talking about how the information wouldn't be round-tripped through JSON. This adds a lot to the dict, and it's no longer just a ref once we fully convert the parent to a dict :(

catalinaperalta: Yeah agree, will talk to Anna and see what she things :/

catalinaperalta:
Closing since this feature is pending

PR #21538, opened 2021-11-02
[Test Proxy] Add fixture to automatically start/stop Docker container

mccoyp:
This adds a session-level fixture, `test_proxy`, that can be used to start the proxy Docker container before recorded tests run and shut it down after all tests in the pytest session are run. It also exposes two helper methods used by the fixture, `start_test_proxy` and `stop_test_proxy`, that can be called directly to manually start or stop the proxy container.

The container needs some time to warm up before it'll accept incoming requests, so the `start_test_proxy` method pings a proxy endpoint with GET requests until receiving a 200 (or until a 10-minute timeout is hit).

**Usage example (that will be added to the migration guide soon)**
In a package's `conftest.py`, make the fixture name a parameter to any proxy-dependent, session-level fixtures:
```python
import os
from devtools_testutils import add_general_regex_sanitizer, test_proxy

@pytest.fixture(scope="session", autouse=True)
def sanitize_account_name(test_proxy):
    add_general_regex_sanitizer(value="fakeendpoint", regex=os.envrion["TABLES_STORAGE_ACCOUNT_NAME"])
```
This will make sure the the container startup script is run before any sanitizers are registered (otherwise, we'd get errors). If no sanitizers are being registered with session-level fixtures, a dummy fixture can be added to the package's `conftest.py` instead, that just invokes the `test_proxy` fixture:
```python
from devtools_testutils import test_proxy

@pytest.fixture(scope="session", autouse=True)
def start_proxy(test_proxy):
    return
```

### DIFF
--- a/tools/azure-sdk-tools/devtools_testutils/proxy_docker_startup.py
+++ b/tools/azure-sdk-tools/devtools_testutils/proxy_docker_startup.py
@@ -27,6 +27,7 @@ CONTAINER_NAME = "ambitious_azsdk_test_proxy"
 LINUX_IMAGE_SOURCE_PREFIX = "azsdkengsys.azurecr.io/engsys/testproxy-lin"
 WINDOWS_IMAGE_SOURCE_PREFIX = "azsdkengsys.azurecr.io/engsys/testproxy-win"
 CONTAINER_STARTUP_TIMEOUT = 6000
+PROXY_MANUALLY_STARTED = os.getenv('PROXY_MANUAL_START', False)
 
 REPO_ROOT = os.path.abspath(os.path.join(os.path.abspath(__file__), "..", "..", "..", ".."))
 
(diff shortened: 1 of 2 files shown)


tools/azure-sdk-tools/devtools_testutils/proxy_docker_startup.py
```diff
@@ -0,0 +1,136 @@
...
+    else:
+        _LOGGER.debug("No instance of the test proxy container found. Attempting creation...")
+        create_container()
+
+    _LOGGER.debug("Attempting to start the test proxy container...")
+
+    proc = subprocess.Popen(shlex.split("docker container start " + CONTAINER_NAME))
+    proc.communicate()
+    # wait for the proxy server to become available
+    time.sleep(10)
```
scbedd, line 112: We should 100% spin on a loop that GET `http://localhost:5001/Info/Online`

scbedd: 10 second timeout will lead to possible wonky intermittent failures.

tools/azure-sdk-tools/devtools_testutils/proxy_docker_startup.py
```diff
@@ -0,0 +1,136 @@
...
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+    )
+
+    output, stderr = proc.communicate()
+    try:
+        return json.loads(output)
+    # We'll get a JSONDecodeError on Py3 (ValueError on Py2) if output is empty (i.e. there's no proxy container)
+    except ValueError:
+        return None
```
YalinLi0312, line 61: If could log/print out the error details would be helpful while debugging I think.

tools/azure-sdk-tools/devtools_testutils/proxy_docker_startup.py
```diff
@@ -0,0 +1,136 @@
...
+
+if TYPE_CHECKING:
+    from typing import Optional
+
+
+_LOGGER = logging.getLogger()
+
+CONTAINER_NAME = "ambitious_azsdk_test_proxy"
+LINUX_IMAGE_SOURCE_PREFIX = "azsdkengsys.azurecr.io/engsys/testproxy-lin"
+WINDOWS_IMAGE_SOURCE_PREFIX = "azsdkengsys.azurecr.io/engsys/testproxy-win"
```
YalinLi0312, line 25: Well, I didn't know there's an image for windows, was using LINUX_IMAGE on my windows machine. And it looks it works well.

mccoyp: Yep, that's expected! [This comment](https://github.com/Azure/azure-sdk-for-python/pull/21538/files#diff-d59281f1d615c6b201013a6521648c6a1491359f395213003fa9f848f89b6142R67) explains a bit more. I also use Linux containers for any local Docker-ing

tools/azure-sdk-tools/devtools_testutils/proxy_docker_startup.py
mccoyp: The details of this are logged in `start_test_proxy` since that's where we interpret the output of `get_container_info`, but I added some comments to make the behavior more clear

scbedd (commented):
We need to handle `PROXY_MANUAL_START` here.

scbedd (approved):

Issue #21495, opened 2021-10-29
[formrecognizer] Optimize get children functionality


catalinaperalta:
Improve internal algorithm for get children to reduce processing time.

catalinaperalta:
This feature was not added.

PR #21554, opened 2021-11-03
[formrecognizer] Improve get children implementation

catalinaperalta:
Adding a check to stop searching spans once we've reached elements that end beyond the current span. 

Also changed search elements from being a list to a string. 

Part of #21495

### DIFF
--- a/sdk/formrecognizer/azure-ai-formrecognizer/azure/ai/formrecognizer/_models.py
+++ b/sdk/formrecognizer/azure-ai-formrecognizer/azure/ai/formrecognizer/_models.py
@@ -2290,7 +2290,7 @@ class AnalyzedDocument(object):
         :return: iterable[DocumentWord]
         :rtype: iterable[DocumentWord]
         """
-        return _get_children(self, search_elements=["word"], cross_page=True)
+        return _get_children(self, search_element="word", cross_page=True)
 
     def get_lines(self, **kwargs):  # pylint: disable=unused-argument
         # type: (Any) -> Iterable[DocumentLine]
@@ -2299,7 +2299,7 @@ class AnalyzedDocument(object):
         :return: iterable[DocumentLine]
         :rtype: iterable[DocumentLine]
         """
-        return _get_children(self, search_elements=["line"], cross_page=True)
+        return _get_children(self, search_element="line", cross_page=True)
 
 
 class DocumentEntity(object):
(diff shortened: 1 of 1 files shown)


sdk/formrecognizer/azure-ai-formrecognizer/azure/ai/formrecognizer/_models.py
```diff
@@ -4178,23 +4178,29 @@ def from_dict(cls, data):
...
         return result
 
     # look for elements on a specific page
-    for elem in search_elements:
-        for element in _get_element_list(source_element._parent, elem):
-            if _in_span(element, source_element.spans):
-                result.append(element)
+    for element in _get_element_list(source_element._parent, search_element):
+        if _in_span(element, source_element.spans):
+            result.append(element)
```
catalinaperalta, line 4203: Need to add the _beyond_span check here.

sdk/formrecognizer/azure-ai-formrecognizer/azure/ai/formrecognizer/_models.py
```diff
@@ -4178,23 +4178,29 @@ def from_dict(cls, data):
...
+            for element in _get_element_list(
+                _get_page(
+                    source_element._parent.pages,
+                    region.page_number,
+                    ),
+                search_element,
+            ):
+                if _in_span(element, source_element.spans):
+                    result.append(element)
+                if _beyond_span(element, source_element.spans):
```
kristapratico, line 4196: do we need to sort the spans before doing a check like this? I thought that was Paul's latest feedback.

kristapratico, line 4196: could use `elif`?

catalinaperalta: I dont remember seeing that feedback O.o. Where was it? Or should I ping Paul??

kristapratico: It was in the FR SDK sync chat:

> The spans should be ordered by reading order, which ideally has incrementing indices. Unfortunately, this is not true in general since we don't yet include selection marks in the overall reading order computation (they are listed at the end). So though the assumption about the index ordering should be true in the future, I would recommend that we sort the spans in the implementation to avoid relying on such behavior.

catalinaperalta: Ah yes, I had only taken that as relevant for selection marks...but hmmm you're right. I can do it since we dont have any assurances for the order although so far I've seen ordered spans. This will add some more computation :/ ok will think about this and add it.

catalinaperalta:
Closing this until we add the get children methods back in again.

PR #21568, opened 2021-11-03
Use working region/subscription for search deployments

benbp:
We are running into deployment issues with the search RP and deletion metadata. This switches our search deployments to a working region/subscription until their fix rolls out

### DIFF
--- a/sdk/search/tests.yml
+++ b/sdk/search/tests.yml
@@ -4,6 +4,9 @@ stages:
   - template: /eng/pipelines/templates/stages/archetype-sdk-tests.yml
     parameters:
       BuildTargetingString: azure-search-documents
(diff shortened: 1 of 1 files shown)


sdk/search/tests.yml
```diff
@@ -4,6 +4,8 @@ stages:
   - template: /eng/pipelines/templates/stages/archetype-sdk-tests.yml
     parameters:
       BuildTargetingString: azure-search-documents
+      # TODO: change/remove location back to default westus2 after search RP fixes deletion metadata issue
```
weshaggard, line 7: Please include a link to the issue for this todo.

weshaggard (approved):

benbp:
Overriding because the actual test failures cannot be fixed until the deployment issue is fixed. The latter is verified by this change and the pipeline runs.

PR #21591, opened 2021-11-04
STG79 Preview

tasherif-msft:


sdk/storage/azure-storage-blob/azure/storage/blob/_generated/models/_azure_blob_storage_enums.py
```diff
@@ -175,6 +175,7 @@ class ListContainersIncludeType(with_metaclass(_CaseInsensitiveEnumMeta, str, En
 
     METADATA = "metadata"
     DELETED = "deleted"
+    SYSTEM = "system"
```
seanmcc-msft, line 178: Is this supposed to be here?  This is an STG 80 feature.  If this class isn't public it doesn't matter.

seanmcc-msft (approved):

sdk/storage/azure-storage-blob/azure/storage/blob/_generated/models/_azure_blob_storage_enums.py
tasherif-msft: It is not public

seanmcc-msft (approved):

Issue #21560, opened 2021-11-03
container.read_item() should return dict[str, Any]


marcelvb:
- **Package Name**: azure-cosmos
- **Package Version**: 4.2.0
- **Operating System**: Any
- **Python Version**: 3.8

**Describe the bug**
The type hint for the return type of `read_item()` in `container.py` is `dict[str, str]`. The documentation says it's dict[str, Any] which I think the code should also be.

Code: https://github.com/Azure/azure-sdk-for-python/blob/c8291ac6cb0dbd865da03a88dd2bcb9279e2c4a6/sdk/cosmos/azure-cosmos/azure/cosmos/container.py#L163

Docs: https://github.com/Azure/azure-sdk-for-python/blob/c8291ac6cb0dbd865da03a88dd2bcb9279e2c4a6/sdk/cosmos/azure-cosmos/azure/cosmos/container.py#L175

Other methods also use `dict[str, Any]` for container items. With `dict[str, str]` as a return type you will get type errors when using nesting in json.

tjprescott:
Thank you @marcelvb for reporting this! @Rodrigossz could you take a look?

simorenoh:
Hi @marcelvb, thank you for catching this. You are right - this is a miss on our end. Here is the PR to address this:
https://github.com/Azure/azure-sdk-for-python/pull/21593

Thank you for helping us improve our product!

PR #21593, opened 2021-11-04
[Cosmos] update type hints to reflect actual document body

simorenoh:
Based on https://github.com/Azure/azure-sdk-for-python/issues/21560

Fixing type hints to reflect actual document bodies, which can contain any mix of type pairings

### DIFF
--- a/sdk/cosmos/azure-cosmos/azure/cosmos/container.py
+++ b/sdk/cosmos/azure-cosmos/azure/cosmos/container.py
@@ -160,7 +160,7 @@ class ContainerProxy(object):
         post_trigger_include=None,  # type: Optional[str]
         **kwargs  # type: Any
     ):
-        # type: (...) -> Dict[str, str]
+        # type: (...) -> Dict[str, Any]
         """Get the item identified by `item`.
 
         :param item: The ID (name) or dict representing item to retrieve.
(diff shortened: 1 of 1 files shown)


sdk/cosmos/azure-cosmos/azure/cosmos/container.py
```diff
@@ -376,7 +376,7 @@ def replace_item(
         post_trigger_include=None,  # type: Optional[str]
         **kwargs  # type: Any
     ):
-        # type: (...) -> Dict[str, str]
+        # type: (...) -> Dict[str, Any]
```
xinlian12, line 379: Do we have tests to validate after the change, no following errors?
`With dict[str, str] as a return type you will get type errors when using nesting in json.`

xinlian12 (commented):
LGTM. Thanks

xinlian12 (approved):

sdk/cosmos/azure-cosmos/azure/cosmos/container.py
simorenoh: We don't have any particular type checks for tests but If dict[str, str] wasn't breaking anything, this won't either (str falls under Any as you would expect). Creating a test for this would be weird as well, since it might be a first thought to parse through the dict and check for the values having different types, but a user could also choose to have it all be strings mapped to only strings.

I also ran the existing tests and samples with this and had same results as before, so we should be good!

tjprescott (approved):

PR #21618, opened 2021-11-05
Resolve `azure-search-documents` errors related to `typing-extensions`

scbedd:
Based on the fix though, I checked a few packages that _may_ have the same issue with `aiohttp` and came up with a list

  * azure-communication-chat
  * azure-communication-identity
  * azure-communication-networktraversal
  * azure-communication-phonenumbers
  * azure-communication-sms

EDIT: they are fine.

So after this PR I'm going to check them for the same error.

@tjprescott here ya go!

@xiangyan99 aiohttp doesn't have a dep on typing-extensions anymore, so the version that _was_ getting installed no longer is. Now `mindependency` resolves `typing-extensions` to a much earlier version (due to fact that aiohttp doesn't require a higher one).

This is the earliest version I found that passed the search tests. Do you protest this update? If so, then we need to instead pin the minimum version of `typing-extensions` we test in the mindependency resolver. However, globally we will only test using a minimum version of `typing-extensions>=3.7.4.3` as a result of that.


tjprescott (approved):

tjprescott:
Thank you @scbedd! You are my hero!

xiangyan99 (approved):

PR #21619, opened 2021-11-05
[SchemaRegistry] avro docs cleanup

swathipil:


sdk/schemaregistry/azure-schemaregistry-avroserializer/README.md
```diff
@@ -27,7 +27,22 @@ To use this package, you must have:
 * Python 2.7, 3.6 or later - [Install Python][python]
 
 ### Authenticate the client
```
yunhaoling, line 29: I think maybe we could drop this section?
since for Avro Serializer, we just pass in the SRClient.

What about other languages' readme for the serializer?

sdk/schemaregistry/azure-schemaregistry-avroserializer/README.md
```diff
@@ -27,7 +27,22 @@ To use this package, you must have:
...
-Interaction with Schema Registry Avro Serializer starts with an instance of AvroSerializer class. You need the fully qualified namespace, AAD credential and schema group name to instantiate the client object.
+Interaction with Schema Registry Avro Serializer starts with an instance of AvroSerializer class. The client constructor takes the fully qualified namespace, schema group name, and and Azure Active Directory credential:
```
yunhaoling, line 30: the serializer takes SRClient and group name.. so actually it's not the SR Serializer being authenticated but the SRClient passsed in which makes me wonder if we need this section or not

sdk/schemaregistry/azure-schemaregistry-avroserializer/README.md
```diff
@@ -27,7 +27,22 @@ To use this package, you must have:
...
+pip install azure-identity
+```
+
+* Additionally, to use the async API supported on Python 3.6+, you must first install an async transport, such as [aiohttp](https://pypi.org/project/aiohttp/):
+
+```Bash
+pip install aiohttp
+```
 
 **Create client using the azure-identity library:**
```
yunhaoling, line 47: same comment as the above one, I prefer stripping it out or we could replace with other titles e.g. `Create Avro Serializer using azure-schemaregistry library`.

sdk/schemaregistry/azure-schemaregistry-avroserializer/README.md
```diff
@@ -99,7 +115,7 @@ fully_qualified_namespace = os.environ['SCHEMAREGISTRY_FULLY_QUALIFIED_NAMESPACE
 group_name = "<your-group-name>"
 
 schema_registry_client = SchemaRegistryClient(fully_qualified_namespace, token_credential)
-serializer = AvroSerializer(client=schema_registry_client, group_name=group_name)
+serializer = AvroSerializer(client=schema_registry_client, group_name=group_name, auto_register_schemas=True)
```
yunhaoling, line 118: Is there a  reason why we want it to be True?

because it makes feel like we want users to always turn on this feature, and if it's the case, shouldn't we be considering making this option default to true within out SDK?

sdk/schemaregistry/azure-schemaregistry-avroserializer/README.md
swathipil: yeah that's a good point

sdk/schemaregistry/azure-schemaregistry-avroserializer/README.md
swathipil: yeah that makes more sense

sdk/schemaregistry/azure-schemaregistry-avroserializer/README.md
swathipil: that's fair, maybe I can add a snippet where we register it first with the client? and don't pass in the auto_register_schemas=True?

sdk/schemaregistry/azure-schemaregistry-avroserializer/README.md
swathipil: .NET does have this section, but they specify that the constructor to pass things into is the client constructor: https://github.com/Azure/azure-sdk-for-net/tree/main/sdk/schemaregistry/Microsoft.Azure.Data.SchemaRegistry.ApacheAvro#authenticate-the-client
Java does too: https://github.com/Azure/azure-sdk-for-java/tree/main/sdk/schemaregistry/azure-data-schemaregistry-apacheavro#create-schemaregistryasyncclient-with-azure-active-directory-credential
JS does not have this section.

Personally, I would prefer keeping and re-wording this. I personally as a user would want all necessary info here, and having to click links to authenticate, etc. would not be super convenient.

sdk/schemaregistry/azure-schemaregistry-avroserializer/README.md
swathipil: I'm not sure why the decision was made to not auto_register_schemas by default, but I think maybe it was because registering schemas is makes a call to the service. If we were to set this to True by default, that feels like silent behavior. The user may not want to make so many calls to the service, especially if they only have the one schema to register but lots of data to serialize for it.

yunhaoling: I get the point, it helps the onboarding experience if users haven't registered the schema into SR already.
let's keep `auto_register_schemas=True` for now.

sdk/schemaregistry/azure-schemaregistry-avroserializer/README.md
```diff
@@ -112,6 +127,10 @@ schema_string = """
  ]
 }"""
 
+schema_registry_client = SchemaRegistryClient(fully_qualified_namespace, token_credential)
+schema_register_client.register(group_name, schema_name, schema_string, format)
```
swathipil, line 131: did one manual register, and one example with `auto_register_schemas=True`

sdk/schemaregistry/azure-schemaregistry-avroserializer/README.md
```diff
@@ -27,16 +27,32 @@ To use this package, you must have:
...
+
+**Create AvroSerializer using the azure-schemaregistry library:**
 
 ```python
 from azure.schemaregistry import SchemaRegistryClient
 from azure.schemaregistry.serializer.avroserializer import AvroSerializer
 from azure.identity import DefaultAzureCredential
 
 credential = DefaultAzureCredential()
+# Namespace should be similar to: '<your-eventhub-namespace>.servicebus.windows.net/'
```
yunhaoling, line 55: kitten nit, no slash at the end?
```suggestion
# Namespace should be similar to: '<your-eventhub-namespace>.servicebus.windows.net'
```

yunhaoling (approved):

Issue #350, opened 2015-04-29
wait_for_operation needs to flush as it prints


zooba:
The wait function prints dots while it is working, but it doesn't explicitly flush the output stream and so the dots don't appear until the next newline.

PR #21620, opened 2021-11-05
[rest] store decompressed body in response content for aiohttp with decompression headers

iscai-msft:


sdk/core/azure-core/azure/core/utils/_pipeline_transport_rest_shared.py
```diff
@@ -353,10 +353,11 @@ def _aiohttp_body_helper(response):
     enc = enc.lower()
     if enc in ("gzip", "deflate"):
         if response._decompressed_content:
```
xiangyan99, line 355: Should we move this to line #350?

iscai-msft: so before w check for content encoding, if we see that we've already decompressed the response, we just go ahead and return it?

I think that could be valid, that way we don't have to check the response headers each time

sdk/core/azure-core/azure/core/utils/_pipeline_transport_rest_shared.py
```diff
@@ -353,10 +353,11 @@ def _aiohttp_body_helper(response):
...
-            return response._decompressed_content
+            return response._content
         import zlib
         zlib_mode = 16 + zlib.MAX_WBITS if enc == "gzip" else zlib.MAX_WBITS
         decompressor = zlib.decompressobj(wbits=zlib_mode)
-        response._decompressed_content = decompressor.decompress(response._content)
-        return response._decompressed_content
+        response._content = decompressor.decompress(response._content)
```
xiangyan99, line 360: Is it valid that user load_body then get content?

In this case, content is still raw.

xiangyan99: IOW, is it possible user gets content before calling _aiohttp_body_helper?

sdk/core/azure-core/tests/async_tests/test_rest_stream_responses_async.py
```diff
@@ -217,3 +216,32 @@ async def test_pass_kwarg_to_iter_raw(client):
     response = await client.send_request(request, stream=True)
     async for part in response.iter_raw(chunk_size=5):
         assert part
+
+@pytest.mark.asyncio
+async def test_decompress_compressed_header(client):
```
annatisch, line 221: As discussed in chat - lets make sure we duplicate these tests across the transports to ensure consistent behaviour.

annatisch (approved):

xiangyan99 (approved):

Issue #21174, opened 2021-10-08
[KeyVault] - Add live tests for SKR in KV


maorleger:
SKR is not fully supported in Key Vault, only MHSM.

Once it is available globally in KV implement the following changes:

- [x]  Remove enableHsm conditions on mock attestation service elements in test-resources.json to install unconditionally (assuming #20404 is complete, disregard otherwise)
- [x] Ensure the same suite of tests are run against MHSM and KV (depending on your infrastructure, you might remove some annotation or you might need to duplicate the tests)
  - **Note (11/5/21):** Creating a key with a release policy and releasing it currently works on Key Vault and Managed HSM. However, importing a key with a release policy, or updating a key's existing release policy, work on Managed HSM but not Key Vault.
    - Importing a key with a release policy specified results in a key that has `None` for a release policy.
    - Updating a key's release policy fails with the following error, even when the `exportable` attribute isn't specified in `update_key_properties`: `azure.core.exceptions.HttpResponseError: (BadParameter) AKV.SKR.1014: Cannot modify the exportable state of a key.`

PR #21621, opened 2021-11-05
[Key Vault] Enable tests for SKR on Key Vault

mccoyp:
Part of https://github.com/Azure/azure-sdk-for-python/issues/21174.

This enables key release on Key Vault by adding the "release" permission to `test-resources.json` and runs some key release tests on KV (they were previously only run on MHSM). Running the remaining release-related tests on KV will be done after investigating behavior differences between KV and MHSM.


YalinLi0312 (approved):
LGTM

PR #21671, opened 2021-11-09
Fix Mixed Reality nspkg

craigktreasure:
The Mixed Reality nspkg appears to be incomplete. This attempts to fix it.


craigktreasure:
Turns out we were doing it "correct". So no changes necessary.

PR #21684, opened 2021-11-10
[webpubsub] regen with hub as a client parameter

iscai-msft:


msyyc:
the content is in another PR :https://github.com/Azure/azure-sdk-for-python/pull/21688

PR #21685, opened 2021-11-10
[Cosmos] Use repo's general tox.ini

mccoyp:
This removes the `tox.ini` file in `azure-cosmos` to instead use the repo-general `tox.ini` under `/eng/tox`.

@scbedd, you're more familiar with `tox` than I am. Live tests are passing -- does this change look like it won't cause any issues?


simorenoh (approved):
Approving since using the main tox.ini file didn't affect running tests

scbedd:
@mccoyp I have no issue with this. If the cosmos folks are ok with it?

PR #21691, opened 2021-11-10
[Communication] Introducing Media Composition SDK for Python

minnieliu:


sdk/communication/azure-communication-mediacomposition/CHANGELOG.md
```diff
@@ -0,0 +1,6 @@
+# Release History
+
+## 1.0.0b1 (Unreleased)
```
GoWang, line 3: I am not familiar with the python version format, but other languages set the version as `1.0.0-beta.1`

minnieliu: For Python, the versioning format is a bit different. For example, Identity changelog looks like this: 
image

PR #21693, opened 2021-11-10
Use arm template parameters as sole input for live test environment variables

benbp:
For sovereign cloud testing, many cloud-specific arm template parameters and environment variables get passed in from an external keyvault source. This makes them hard to manage and track. I plan to move these values into source control, and as a pre-requisite I am consolidating as many values into the arm template parameters as possible (as opposed to a mismash of parameters and env var declarations).

### DIFF
--- a/sdk/attestation/test-resources.json
+++ b/sdk/attestation/test-resources.json
@@ -16,6 +16,13 @@
         "description": "The location of the resource. By default, this is the same as the resource group."
       }
     },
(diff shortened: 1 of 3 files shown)


mccoyp (approved):
Looks good for Key Vault

weshaggard (approved):

benbp:
Some managed HSM test timeouts (likely due to region overlaps), but the live tests are looking good.

PR #21721, opened 2021-11-11
[SchemaRegistry] avro docs update

swathipil:


sdk/schemaregistry/azure-schemaregistry-avroserializer/README.md
```diff
@@ -102,7 +102,7 @@ The following sections provide several code snippets covering some of the most c
 ### Serialization
 
 Use `AvroSerializer.serialize` method to serialize dict data with the given avro schema.
-The method would automatically register the schema to the Schema Registry Service and keep the schema cached for future serialization usage.
+The method would use a schema previously registered to the Schema Registry service and keep the schema cached for future serialization usage. It is also possible to avoid pre-registering the schema to the service and automatically register with the `serialize` method by passing in the keyword argument `auto_register_schemas=True`.
```
yunhaoling, line 105: thanks for making it clear! probably I would go a bit further with something like, otherwise it makes me feel it's the serialize method taking `auto_register_schemas`:
```suggestion
The method would use a schema previously registered to the Schema Registry service and keep the schema cached for future serialization usage. It is also possible to avoid pre-registering the schema to the service and automatically register with the `serialize` method by instantiating the `AvroSerializer` with keyword argument `auto_register_schemas=True`.
```

yunhaoling:
This is a puppy nit, so I think there's an extra slash in the end of the namespace XD
image

yunhaoling (approved):

sdk/schemaregistry/azure-schemaregistry-avroserializer/README.md
swathipil: ah good catch!

PR #21746, opened 2021-11-13
[Mgmt Resource] Migrate (most) tests to test proxy

mccoyp:
This makes use of AzureMgmtRecordedTestCase class (added separately in https://github.com/Azure/azure-sdk-for-python/pull/21767), which is to be used as a parent test class for management tests using the test proxy.

This migrates all tests that aren't skipped by default to use the test proxy, but tests from `test_mgmt_resource_locks.py` haven't been re-recorded. This is because my test service principal doesn't have permissions necessary to get the tests passing, and I don't have enough knowledge of the package to know how to fix this.

Tests failing in CI is expected for now, by the way -- that's being sorted as we speak and should be resolved soon.


mccoyp:
Tagging @scbedd to prompt GitHub notifications

PR #21755, opened 2021-11-15
Update async legacy module to newer async module in authenticate_with_azure_named_key_credential_async.py

sammiee5311:
I've modified one of the samples that is written as `asyncio.get_event_loop()` to `asyncio.run()` since get_running_loop() is deprecated (since version 3.10).

https://docs.python.org/3/library/asyncio-eventloop.html#event-loop

> Deprecated since version 3.10: Deprecation warning is emitted if there is no running event loop. In future Python releases, this function will be an alias of get_running_loop().


swathipil:
@sammiee5311 Thanks for your contribution! This looks good to me!

swathipil (approved):

PR #21774, opened 2021-11-16
[formrecognizer] fix failing tests

kristapratico:


catalinaperalta (approved):
So strange that this changed O.o

PR #17985, opened 2021-04-12
[Communication] - SDK - Synced SMS and Phone Numbers shared folders

jbeauregardb:
SMS and Phone Numbers packages had an outdated version of their shared folders. Identity and chat were updated so I replicated the changes on these two other SDKs.

PR #21777, opened 2021-11-16
Sync the _shared folder for Communication

petrsvihlik:
I found out that the _shared folder (`sdk\communication\azure-communication-<package>\azure\communication\<package>\_shared`) is not in sync across the Communication packages (SMS, PhoneNumbers, Chat, Identity, and NetworkTraversal).

The differences made our _shared folders split into the following branches:
- Branch A:
  - PhoneNumbers
  - SMS
  - Chat (1 difference compared to SMS and PhoneNumbers)
- Branch B:
  - Identity
  - NetworkTraversal (1 little change related to linting compared to Identity)

@jbeauregardb attempted to unify the folders in the past (#17985) but Identity was not updated (and I think NetworkTraversal was introduced later and that's how Branch B was created).

The second important milestone was #18026 that removed `replace(tzinfo=TZ_UTC)` from creation of the current timestamp.
- There was a [comment](https://github.com/Azure/azure-sdk-for-python/pull/18026#issuecomment-820656013) saying that we "should have some unit tests including both naïve and TZ-aware inputs." which wasn't processed. I'll probably add some tests for that.
- Also, it'd probably make sense to be able to handle both the naïve and TZ-aware inputs in the code to be compatible with the older versions of Python. E.g. similarly to what they do in [Azure.Core](https://github.com/Azure/azure-sdk-for-python/blob/5557136179aa3df49ab9c4abfd3b6d1a9a38975c/sdk/core/azure-core/azure/core/serialization.py#L91-L94). I'll take a closer look at it and suggest a change.

This PR is an attempt to cherry pick the right changes and unify the folders. 
I'm including all past contributors and owners of concerned packages as reviewers.

### DIFF
--- a/sdk/communication/azure-communication-chat/azure/communication/chat/_shared/utils.py
+++ b/sdk/communication/azure-communication-chat/azure/communication/chat/_shared/utils.py
@@ -15,7 +15,7 @@ import calendar
 from msrest.serialization import TZ_UTC
 from azure.core.credentials import AccessToken
 
-def _convert_datetime_to_utc_int(expires_on):
+def _convert_datetime_to_utc_int(input_datetime):
     """
     Converts DateTime in local time to the Epoch in UTC in second.
 
@@ -24,7 +24,7 @@ def _convert_datetime_to_utc_int(expires_on):
     :return: Integer
     :rtype: int
     """
-    return int(calendar.timegm(expires_on.utctimetuple()))
+    return int(calendar.timegm(input_datetime.utctimetuple()))
 
 def parse_connection_str(conn_str):
     # type: (str) -> Tuple[str, str, str, str]
--- a/sdk/communication/azure-communication-chat/azure/communication/chat/_shared/utils.py
+++ b/sdk/communication/azure-communication-chat/azure/communication/chat/_shared/utils.py
@@ -87,7 +87,7 @@ def create_access_token(token):
         padded_base64_payload = base64.b64decode(parts[1] + "==").decode('ascii')
         payload = json.loads(padded_base64_payload)
         return AccessToken(token,
-                           _convert_datetime_to_utc_int(datetime.fromtimestamp(payload['exp']).replace(tzinfo=TZ_UTC)))
+                           _convert_datetime_to_utc_int(datetime.fromtimestamp(payload['exp'], TZ_UTC)))
     except ValueError:
         raise ValueError(token_parse_err_msg)
 
--- a/sdk/communication/azure-communication-identity/azure/communication/identity/_shared/utils.py
+++ b/sdk/communication/azure-communication-identity/azure/communication/identity/_shared/utils.py
@@ -6,7 +6,6 @@
 
 import base64
 import json
-import time
 from typing import (  # pylint: disable=unused-import
     cast,
     Tuple,
@@ -15,6 +14,7 @@ from datetime import datetime
 import calendar
 from msrest.serialization import TZ_UTC
 from azure.core.credentials import AccessToken
+
 
 def _convert_datetime_to_utc_int(expires_on):
     return int(calendar.timegm(expires_on.utctimetuple()))
@@ -50,6 +50,7 @@ def get_current_utc_time():
     # type: () -> str
     return str(datetime.utcnow().strftime("%a, %d %b %Y %H:%M:%S ")) + "GMT"
 
+
 def get_current_utc_as_int():
     # type: () -> int
     current_utc_datetime = datetime.utcnow().replace(tzinfo=TZ_UTC)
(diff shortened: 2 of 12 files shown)


petrsvihlik:
Since we don't expose the value of `datetime.utcnow()` anywhere outside the `get_current_utc_as_int()`, I think we can keep it without the `.replace(tzinfo=TZ_UTC)`. I added a test for UTC without a timezone specified to make sure the `_convert_datetime_to_utc_int` works in all cases.
I ran the tests on both Python 2.x and 3.x and all seems to be passing.

sdk/communication/azure-communication-identity/azure/communication/identity/_shared/utils.py
```diff
@@ -17,6 +16,14 @@
 from azure.core.credentials import AccessToken
 
 def _convert_datetime_to_utc_int(expires_on):
+    """
+    Converts DateTime in local time to the Epoch in UTC in second.
+
+    :param input_datetime: Input datetime
```
AikoBB, line 22: you have to change the param name to `input_datetime` , same for other duplications in other packages

juancamilor:
@LuChen-Microsoft FYI

sdk/communication/azure-communication-identity/azure/communication/identity/_shared/utils.py
petrsvihlik: well spotted! :)
that was a copy-paste from chat. it seems that the linter doesn't check whether comments are matching param names.
fixed in 191dc62

petrsvihlik:
Added a new line to CODEOWNERS to enforce the reviews for the _shared folders https://github.com/Azure/azure-sdk-for-python/pull/21777/commits/bbaff7683367ad959041a9034c32c4e3a510f29c

petrsvihlik:
> Also, it'd probably make sense to be able to handle both the naïve and TZ-aware inputs in the code to be compatible with the older versions of Python. E.g. similarly to what they do in Azure.Core. I'll take a closer look at it and suggest a change.

It seems that TZ_UTC covers this well.

I also did some extensive testing and discovered that deserialization of the `expires_on` doesn't work correctly - there was a 1hr difference between the string payload and the deserialized int. So in 5371ea77a103c37a12df9c6e712761fc102e42c6, I replaced this:

```python
return AccessToken(token, _convert_datetime_to_utc_int(datetime.fromtimestamp(payload['exp']).replace(tzinfo=TZ_UTC)))
```
with:
```python
return AccessToken(token, _convert_datetime_to_utc_int(datetime.fromtimestamp(payload['exp'], TZ_UTC)))
```

Now it should be all good to review. 

@AriZavala2 @LuChen-Microsoft @RoyHerrod @jbeauregardb can you please verify the changes?

petrsvihlik:
Confirmed with @annatisch that [this change](https://github.com/Azure/azure-sdk-for-python/pull/17451/files) is legit. Explanation:
> token types follow a specific protocol, so the method signatures needed to be updated to be consistent.

image

This PR unifies the code to follow the change made in the Chat lib.

annatisch (approved):

sdk/communication/azure-communication-chat/azure/communication/chat/_shared/utils.py
```diff
@@ -15,7 +15,7 @@
 from msrest.serialization import TZ_UTC
 from azure.core.credentials import AccessToken
 
-def _convert_datetime_to_utc_int(expires_on):
+def _convert_datetime_to_utc_int(input_datetime):
```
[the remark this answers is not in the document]
petrsvihlik: generally, yes - this applies to all code in the _shared folders. in the case of this specific method, it's being used by two other methods from utils.py - `get_current_utc_as_int` and `create_access_token`.

annatisch (approved):

sdk/communication/azure-communication-networktraversal/azure/communication/networktraversal/_shared/user_credential.py
```diff
@@ -79,12 +79,8 @@ def _wait_till_inprogress_thread_finish_refreshing(self):
         self._lock.acquire()
 
     def _token_expiring(self):
-        return self._token.expires_on - self._get_utc_now() <\
-            timedelta(minutes=self._ON_DEMAND_REFRESHING_INTERVAL_MINUTES)
+        return self._token.expires_on - get_current_utc_as_int() <\
+            timedelta(minutes=self._ON_DEMAND_REFRESHING_INTERVAL_MINUTES).total_seconds()
```
AriZavala2, line 83: Why is it now the .total_seconds() added?

AriZavala2 (approved):

LuChen-Microsoft (approved):

sdk/communication/azure-communication-networktraversal/azure/communication/networktraversal/_shared/user_credential.py
petrsvihlik: Because the `self._token.expires_on` is an int (as you can see in utils.py), the  `get_current_utc_as_int()` is an int, and the difference is therefore also an int. If you'd compare it with a timedelta you'd get:
> TypeError: '<' not supported between instances of 'int' and 'datetime.timedelta'

I don't think this has ever worked. The tests were disabled. We enabled them and added new ones in the follow-up PR: https://github.com/Azure/azure-sdk-for-python/pull/21834/files

Issue #21750, opened 2021-11-15
azure-cosmos not able to disable "endpoint_discovery"


gavinHuang:
- **Package Name**: Azure-cosmos
- **Package Version**: 4.2.0 (and all other version)
- **Operating System**:Windows
- **Python Version**: 3.7

**Describe the bug**
"enable_endpoint_discovery" can't set to False at Cosmos_Client.py (line 82):
`policy.EnableEndpointDiscovery = kwargs.pop('enable_endpoint_discovery', None) or policy.EnableEndpointDiscovery
`
Even specifically set the value to False, this code will alway use the default value which is True.

**To Reproduce**
1.
`
cosmos_client = CosmosClient(
    url="https://accountname.documents.azure.com:443",
    credential="account_key",
    enable_endpoint_discovery=False
)
`

**Expected behavior**
Disable endpoint discovery 

**Screenshots**

**Additional context**
I don't need account discovery, the discovery change the url to something like:
`accountname-[region_name].documents.azure.com`
which in my environment is not a valid domain name, so I get error:
`
azure.core.exceptions.ServiceRequestError: <urllib3.connection.HTTPSConnection object at 0x0000029C0E3B0588>: Failed to establish a new connection: [Errno 11001] getaddrinfo failed
`

swathipil:
@gavinHuang Thanks for the feedback - directing this to the right people!

simorenoh:
Hi @gavinHuang, thank you for reporting this. You are completely right, this is a current gap in the SDK; the line you outlined will always evaluate to True since the second condition (policy.EnableEndpointDiscovery) is set to True (and naturally any or statement with one True condition ends up being True).

I have gone ahead and fixed this here: https://github.com/Azure/azure-sdk-for-python/pull/21778

Thanks again for reporting this and helping us improve the SDK! :)

gavinHuang:
Thanks for the quick turn around!

PR #21778, opened 2021-11-16
[Cosmos] fix endpoint_discovery logic to be user-configurable

simorenoh:
The current implementation for setting the enable_endpoint_discovery option utilizes an or statement where one of the two clauses will always be True, and as such fails to recognize any other input.

With these changes users can actually set it to False if they wish to.

Based on this issue: https://github.com/Azure/azure-sdk-for-python/issues/21750


simorenoh:
@annatisch tagging you for sanity check :)

kushagraThapar (approved):

kushagraThapar (changes requested):
Overall looks good to me, requesting changes on a test case. 
Please add a test case to verify this new behavior.

kushagraThapar (approved):

Issue #21748, opened 2021-11-14
Return code and HTTP errors in Event Grid client


edoboker:
Hi,

I was wondering why both sync and async methods of 'send' do not return standard HTTP errors and responses. It would immensely help when trying to handle 429 or 413 (rate too high and message too large). Currently, my only way of knowing if send actions have failed is to check the logs of the event grid of shift to other SDKs (which have their own issues).

I believe it has to do with [this ](https://github.com/Azure/azure-sdk-for-python/blob/fe9b0f8c254a26464e3acabf72d69b4724dd070e/sdk/eventgrid/azure-eventgrid/azure/eventgrid/_publisher_client.py#L204) line, which does not return the HTTP response of the underlying azure.core._pipeline_client (which in its turn, [does](https://github.com/Azure/azure-sdk-for-python/blob/878a12ebb8911a40017228327172b3dcfc85b6e7/sdk/core/azure-core/azure/core/_pipeline_client.py#L200) return an HTTP response. 

Am I mistaken or missing anything about error handling in this SDK? I could not find any exception handling in the code as well (which is the way they've implemented it in the Golang SDK, I believe), which made me wonder if any error handling mechanism exists at all...

swathipil:
@edoboker Thanks for the question! Routing this to people who can answer!

rakshith91:
Hi @edoboker 
Thanks for the issue. The reason EG does not return anything here is because the service itself does not return a body.

You can always use a `raw_response_hook` to get the status code.

Define a callback method like:
```
def callback(response):
    print(response.http_response.status_code)
```

and pass it as `raw_response_hook` in the send method
```
client.send([
   events,
    raw_response_hook=callback
)
```

edoboker:
Thanks for the reply! I think it would be great if this could be added to the documentation as well, being such an important feature.
Just out of curiosity - why not return an HTTP response without a body then? The status code is still meaningful... In addition - I can see in other SDKs and when accessing the service directly, I can see HTTP errors body received from the service itself, such as:
```
{{
    "error": {
        "code": "RequestEntityTooLarge",
        "message": "The maximum size for the JSON content (1049600) has been exceeded. Report '5f11aac2-8b20-41c2-bce4-5fe6be317e33:1:11/16/2021 12:15:17 PM (UTC)' to our forums for assistance or raise a support ticket.",
        "details": [{
            "code": "PayloadTooLarge",
[...]

PR #21790, opened 2021-11-17
Raise in Eventgrid

rakshith91:
Fixes https://github.com/Azure/azure-sdk-for-python/issues/21748


sdk/eventgrid/azure-eventgrid/azure/eventgrid/_publisher_client.py
```diff
@@ -201,7 +201,7 @@ def send(self, events, **kwargs):
         elif isinstance(events[0], EventGridEvent) or _is_eventgrid_event(events[0]):
             for event in events:
                 _eventgrid_data_typecheck(event)
-        self._client.send_request(  # pylint: disable=protected-access
+        return self._client.send_request(  # pylint: disable=protected-access
```
rakshith91, line 204: This technically returns 'None' instead of not returning anything

sdk/eventgrid/azure-eventgrid/samples/async_samples/sample_use_raw_response_hook_async.py
```diff
@@ -0,0 +1,52 @@
+# --------------------------------------------------------------------------
```
annatisch, line 1: We can delete this sample.

annatisch: And the sync version

sdk/eventgrid/azure-eventgrid/tests/test_exceptions.py
```diff
@@ -0,0 +1,83 @@
...
+    def test_raise_on_auth_error(self, resource_group, eventgrid_topic, eventgrid_topic_primary_key, eventgrid_topic_endpoint):
+        akc_credential = AzureKeyCredential("bad credential")
+        client = EventGridPublisherClient(eventgrid_topic_endpoint, akc_credential)
+        eg_event = EventGridEvent(
+                subject="sample", 
+                data={"sample": "eventgridevent"}, 
+                event_type="Sample.EventGrid.Event",
+                data_version="2.0"
+                )
+        with pytest.raises(HttpResponseError, match="The request authorization key is not authorized for*"):
```
annatisch, line 50: This should raise an ClientAuthenticationError?

sdk/eventgrid/azure-eventgrid/tests/test_exceptions.py
```diff
@@ -0,0 +1,83 @@
...
+        path  = os.path.abspath(os.path.join(os.path.abspath(__file__), "..", "./large_data.json"))
+        with open(path) as json_file:
+            data = json.load(json_file)
+        eg_event = EventGridEvent(
+                subject="sample", 
+                data=data,
+                event_type="Sample.EventGrid.Event",
+                data_version="2.0"
+                )
+        with pytest.raises(HttpResponseError):
```
annatisch, line 82: Could we also make sure that the error data is correctly exposed/accessible? It should be, as this code is adapted from the previous generated code - but just a couple of asserts that the message/status/etc can be read would be good :)

sdk/eventgrid/azure-eventgrid/azure/eventgrid/_publisher_client.py
```diff
@@ -21,6 +21,13 @@
     HttpLoggingPolicy,
     UserAgentPolicy,
 )
+from azure.core.exceptions import (
+    ClientAuthenticationError,
+    HttpResponseError,
+    ResourceNotFoundError,
+    ResourceExistsError,
+    map_error
```
johanste, line 29: `map_error` was intended as a means of inversion-of-control (e.g. a way to inject behavior into generated code). What does using it buy us here?

rakshith91: Getting specific errors like ClientAuthenticationError etc instead of generic HttpResponseError

annatisch (approved):

Issue #20932, opened 2021-09-29
[Test Proxy] Streamline doc for setting up proxy dev cert


mccoyp:
Disorganized notes of work items:

- [x] Clarify which, if any, of the general steps need to be followed if using Docker and Python
- [x] ~Link to dev cert is broken in the current doc (should be [this one](https://github.com/Azure/azure-sdk-for-python/blob/main/eng/common/testproxy/dotnet-devcert.crt))~
- [x] Fix link to [trust_proxy_cert.py](https://github.com/Azure/azure-sdk-for-python/blob/main/scripts/devops_tasks/trust_proxy_cert.py)
- [x] Can make it more clear that the `trust_proxy_cert.py` does many of the setup steps for you
- [x] Should emphasize that environment variables need to be persistently set (e.g. using SETX instead of SET)

PR #21797, opened 2021-11-17
[Test Proxy] Update test_proxy_migration_guide.md

mccoyp:
@scbedd This adds documentation for auto-starting the test proxy, optionally starting/stopping the proxy manually, and handling test recordings.

### DIFF
--- a/doc/dev/test_proxy_migration_guide.md
+++ b/doc/dev/test_proxy_migration_guide.md
@@ -62,22 +62,28 @@ on how to do so can be found [here][proxy_cert_docs].
 
 ### Start the proxy server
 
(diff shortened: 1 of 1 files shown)


scbedd (approved):
This all looks great. We should add an additional item about `scripts/devops_tasks/trust_proxy_cert.py`. 

Currently, it's assumed that those variables are set in our default, and I don't think it automatically falls back on `http` if the cert isn't trusted and updated in REQUESTS_CA_CERT.

mccoyp:
For record keeping: tracking the documentation of `trust_proxy_cert.py`'s use is part of https://github.com/Azure/azure-sdk-for-python/issues/20932

PR #21812, opened 2021-11-18
[Perf] Add PerfTestBase and BatchPerfTest

### LINKED ISSUE #20816: [Perf] Add "--insecure" option to disable SSL validation
The perf framework should validate SSL certificates by default, and disable validation if the "--insecure" command line option is specified.  This behavior should apply regardless of whether the test-proxy is being used or not.

Currently, the Python perf framework always disable SSL validation when using test-proxy, and I don't believe it supports disabling SSL validation when **not** using test-proxy.

https://github.com/Azure/azure-sdk-for-python/blob/0ff2bb65589142ef1bc881d1705a60bdf1d9ae47/tools/azure-devtools/src/azure_devtools/perfstress_tests/_policies.py#L20

### LINKED ISSUE #21159: [Perf] Create dev_requirements.txt and README.md for system_perfstress tests
Currently, I don't believe there are documented instructions for running the `system_perfstress` tests.  I also believe dependencies need to be installed manually, since there is no `dev_requirements.txt` specific to `system_perfstress`, and `/tools/azure-devtools/dev_requirements.txt` does not include the required packages.

I recommend making the `system_perfstress` tests work as similar as possible to SDK tests like https://github.com/Azure/azure-sdk-for-python/tree/main/sdk/storage/azure-storage-blob/tests/perfstress_tests.

### LINKED ISSUE #21384: [Perf] Add PerfTestBase and BatchPerfTest
Port https://github.com/Azure/azure-sdk-for-net/pull/24653 to Python.

annatisch:
Fixes #20816
Fixes #21384 
Fixes #21159


mikeharder (approved):
I didn't review the code line-by-line, but the APIs look correct.

I recommend adding a sample test that extends BatchPerfTest, something like https://github.com/Azure/azure-sdk-for-net/pull/24653/files#diff-5ebbae98f50d4148984513195f10702a48ae2b56dfbfaa9220b591c88b710978.

I also recommend running a few real-world tests like blop upload/download/list-blobs to verify no change in results.

annatisch:
Thanks @mikeharder - I added that sample to the documentation, and ran the blob storage tests (with and without proxy)

mikeharder:
> Thanks @mikeharder - I added that sample to the documentation, and ran the blob storage tests (with and without proxy)

How do you feel about adding the sample `MessageReceiveTest` to the `system_perfstress` folder, so it can be easily run without needing to copy/paste from the docs?

annatisch:
Sure thing. Added.

PR #21822, opened 2021-11-18
[formrecognizer] Migrate to test proxy

### LINKED ISSUE #21144: [formrecognizer] Migrate tests to test proxy

catalinaperalta:
Fixes #21144


sdk/formrecognizer/azure-ai-formrecognizer/tests/conftest.py
```diff
@@ -6,9 +6,40 @@
...
+    add_body_key_sanitizer(
+        json_path="source",
+        value="blob_sas_url",
+        regex="(?<=\\/\\/)[a-z-]+(?=\\.blob\\.core\\.windows\\.net)(.*)$",
+    )
+    add_body_key_sanitizer(
+        json_path="accessToken",
+        value="redacted",
+        group_for_replace="(\"accessToken\": \"([0-9a-z-]*)\")",
+    )
```
catalinaperalta, line 45: Please note that I pushed this but this is untested and probably will change.

sdk/formrecognizer/azure-ai-formrecognizer/azure/ai/formrecognizer/_form_base_client.py
```diff
@@ -64,6 +64,11 @@ def __init__(self, endpoint, credential, **kwargs):
             }
         )
 
+        try:
+            endpoint = endpoint.rstrip("/")
+        except AttributeError:
+            raise ValueError("Parameter 'endpoint' must be a string.")
+
```
kristapratico, line 71: we'll want for async too (realize you might not have started converting async yet, but wanted to call out)

sdk/formrecognizer/azure-ai-formrecognizer/tests/conftest.py
kristapratico: I found this super useful for testing: https://regexr.com/

sdk/formrecognizer/azure-ai-formrecognizer/tests/preparers.py
```diff
@@ -18,7 +18,7 @@
 FormRecognizerPreparer = functools.partial(
     PowerShellPreparer,
     'formrecognizer',
-    formrecognizer_test_endpoint="https://region.api.cognitive.microsoft.com/",
+    formrecognizer_test_endpoint="https://fakeendpoint.cognitiveservices.azure.com/",
```
kristapratico, line 21: also remove the trailing slash here so it matches the conftest sanitizer value. In fact, I think PowershellPreparer sanitizing is now enabled so you might not even need the conftest sanitizer anymore. Haven't tested that yet myself.

sdk/formrecognizer/azure-ai-formrecognizer/tests/recordings/test_business_card_async.pyTestBusinessCardAsynctest_business_card_v2.json
```diff
@@ -0,0 +1,4 @@
+{
```
kristapratico, line 1: if the test doesn't create a recording you shouldn't need the recorded_by_proxy decorator

sdk/formrecognizer/azure-ai-formrecognizer/tests/test_custom_forms.py
```diff
@@ -61,7 +64,8 @@ def test_custom_form_multipage_unlabeled(self, client, formrecognizer_multipage_
 
     @FormRecognizerPreparer()
     @FormTrainingClientPreparer()
-    def test_custom_form_labeled(self, client, formrecognizer_storage_container_sas_url_v2):
+    @recorded_by_proxy
+    def test_custom_form_labeled(self, client, formrecognizer_storage_container_sas_url_v2, **kwargs):
```
kristapratico: are you getting an error if you don't add kwargs?

catalinaperalta: Yes I am, I noticed that the tests that take in several params need it or else will raise an error.

sdk/formrecognizer/azure-ai-formrecognizer/azure/ai/formrecognizer/_form_base_client.py
kristapratico, line 71: it didn't really matter before since we weren't manipulating the endpoint, but just realized that now that we're rstriping it we should make sure `self._endpoint` on L26 also gets this value. It's used when any `get_*client` call is made.

sdk/formrecognizer/azure-ai-formrecognizer/tests/conftest.py
```diff
@@ -6,9 +6,51 @@
...
 
 # Ignore async tests for Python < 3.6
 collect_ignore_glob = []
 if sys.version_info < (3, 6):
     collect_ignore_glob.append("*_async.py")
+
+# autouse=True will trigger this fixture on each pytest run, even if it's not explicitly used by a test method
+# @pytest.fixture(scope="session", autouse=True)
+# def start_proxy(test_proxy):
+#     return
```
kristapratico, line 21: nit: seems like we can remove this

sdk/formrecognizer/azure-ai-formrecognizer/tests/preparers.py
```diff
@@ -53,7 +53,7 @@ def create_resource(self, name, **kwargs):
             form_recognizer_account_key = os.environ["FORMRECOGNIZER_TEST_API_KEY"]
             polling_interval = 5
         else:
-            form_recognizer_account = "https://region.api.cognitive.microsoft.com/"
+            form_recognizer_account = "https://fakeendpoint.cognitiveservices.azure.com"
             form_recognizer_account_key = "fakeZmFrZV9hY29jdW50X2tleQ=="
             polling_interval = 0
```
kristapratico, line 59: I can't comment on it, but the first recording file in this PR is quite big. Wondering if we should mark the stream tests live for now

sdk/formrecognizer/azure-ai-formrecognizer/tests/test_custom_forms.py
kristapratico: what is the error? wondering if this is a bug we need to open.

kristapratico (commented):
consider running live/without test proxy in CI as well just to verify all changes are good

sdk/formrecognizer/azure-ai-formrecognizer/tests/test_custom_forms.py
catalinaperalta: I thought this was expected behavior because of what's said here: https://github.com/Azure/azure-sdk-for-python/blob/main/doc/dev/test_proxy_migration_guide.md#record-test-variables

The error basically says that a `variables` kwarg is being passed in to these tests and it's not specified. That's why I added the **kwargs to these tests. Let me know if you think this is a bug.

sdk/formrecognizer/azure-ai-formrecognizer/tests/preparers.py
catalinaperalta: Deleted most of the larger file tests and marked them as live only. Done!

sdk/formrecognizer/azure-ai-formrecognizer/tests/test_custom_forms.py
kristapratico: I'm curious about this because when I converted over TA, my tests also took params but I didn't need to include kwargs. I read that section of the migration guide as you need only accept `variables` or `kwargs` if you want to use recorded test variables (which we don't do in FR yet). I'll look into this

kristapratico: Okay, seems to work in live/playback with or without kwargs. I don't think it hurts to keep it, though.

catalinaperalta: Yeah I've run all the tests on my local machine both live and in playback and (after adding the kwargs) all the tests ran fine

kristapratico (approved):

PR #21823, opened 2021-11-19
[Test Proxy] Add bodiless matcher, sanitize test SAS tokens

mccoyp:
@scbedd

### DIFF
--- a/tools/azure-sdk-tools/devtools_testutils/proxy_testcase.py
+++ b/tools/azure-sdk-tools/devtools_testutils/proxy_testcase.py
@@ -9,6 +9,7 @@ import requests
 import six
 import sys
 from typing import TYPE_CHECKING
+import pdb
 
 try:
     # py3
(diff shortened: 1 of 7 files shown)


tools/azure-sdk-tools/devtools_testutils/proxy_testcase.py
```diff
@@ -40,9 +40,13 @@
 PLAYBACK_START_URL = "{}/playback/start".format(PROXY_URL)
 PLAYBACK_STOP_URL = "{}/playback/stop".format(PROXY_URL)
 
-# TODO, create a pytest scope="session" implementation that can be added to a fixture such that unit tests can
-# startup/shutdown the local test proxy
-# this should also fire the admin mapping updates, and start/end the session for commiting recording updates
+# we store recording ID in a module-level variable so that sanitizers can access it
+this = sys.modules[__name__]
+this.recording_id = None
```
seankane-msft, line 45: Just a suggestion from my Go experience. I ended up storing this in a map of test name -> recording ID just to prevent any parallelization that might be done in the future. I don't know if pytest runs in parallel anywhere but this could be a problem if it does.

seankane-msft (approved):

tools/azure-sdk-tools/devtools_testutils/proxy_testcase.py
mccoyp: That's a really good suggestion! I think I'll give that a try and follow suit

PR #21824, opened 2021-11-19
Issue #9324 by DaisyCisneros

DaisyCisneros:
Added an offline version of the test_compress_plain_no_header test in test_streaming.py by utilizing a new function I added in streams.py. This new test named test_compress_plain_no_header_offline, passes!


kristapratico:
Related issue: https://github.com/Azure/azure-sdk-for-python/issues/9324

sdk/core/azure-core/tests/test_streaming.py
```diff
@@ -43,6 +44,17 @@ def test_decompress_plain_no_header(http_request):
...
 
+@pytest.mark.parametrize("http_request", HTTP_REQUESTS)
+def test_compress_plain_no_header_offline(port, http_request):
+    # thanks to Daisy Cisneros for this test!
+    # expect plain text
+    request = http_request(method="GET", url="http://localhost:{}/streams/string".format(port))
+    with RequestsTransport() as sender:
+        response = sender.send(request)
+        response.raise_for_status()
+        assert response.text() == "test"
```
iscai-msft, line 55: thanks @DaisyCisneros this looks really great! Can you also add to this test by trying to stream download the response? You can follow lines 66 -> 71 in this file

sdk/core/azure-core/tests/test_streaming.py
```diff
@@ -43,6 +44,17 @@ def test_decompress_plain_no_header(http_request):
     decoded = content.decode('utf-8')
     assert decoded == "test"
 
+@pytest.mark.parametrize("http_request", HTTP_REQUESTS)
```
iscai-msft, line 47: love that you're parameterizing this test with our two different kinds of http requests!

for more background information, we have an older `HttpRequest` class and a newer one. Eventually, we want to move to the newer one, but we want to make sure our tests run on both of them, so thank you for making this test run with both request classes!

sdk/core/azure-core/tests/test_streaming.py
```diff
@@ -43,6 +44,17 @@ def test_decompress_plain_no_header(http_request):
...
+def test_compress_plain_no_header_offline(port, http_request):
+    # thanks to Daisy Cisneros for this test!
+    # expect plain text
+    request = http_request(method="GET", url="http://localhost:{}/streams/string".format(port))
+    with RequestsTransport() as sender:
+        response = sender.send(request)
+        response.raise_for_status()
+        assert response.text() == "test"
+
+@pytest.mark.live_test_only
```
iscai-msft, line 57: perfect decorator!

sdk/core/azure-core/tests/testserver_tests/coretestserver/coretestserver/test_routes/streams.py
```diff
@@ -36,3 +36,9 @@ def iterable():
 @streams_api.route('/error', methods=['GET'])
 def error():
     return Response(stream_json_error(), status=400)
+
+@streams_api.route('/string', methods=['GET'])
+def string():
+    return Response(
```
iscai-msft, line 42: hey @DaisyCisneros can you make this yield the value "test" and return a generator, instead of returning it immediately? This will more closely mock a service that is streaming its response to users (i.e., not giving the whole response at the beginning, but giving them in smaler chunks when they ask for it.

[Here's](https://stackoverflow.com/questions/231767/what-does-the-yield-keyword-do) a good stack overflow post that helps explain yield / generators / iterables, and in terms of code, you can look at how either `streaming_body` or `streaming_json_error` are created and returned in the other endpoints in this file

iscai-msft (commented):
overall really great work @DaisyCisneros! Thank you so much for helping us out with this!

sdk/core/azure-core/tests/test_streaming.py
iscai-msft: also just to be clear, you don't have to switch to `pipeline.run()`, the transport.send() method takes in kwarg `stream` as well, so just make sure you pass in `stream=True`

sdk/core/azure-core/tests/test_streaming.py
DaisyCisneros: Thank you! 😊

sdk/core/azure-core/tests/test_streaming.py
DaisyCisneros: Thank you!! I appreciate learning the background on why both kinds of http requests are needed here. 😊 You are welcome!

sdk/core/azure-core/tests/testserver_tests/coretestserver/coretestserver/test_routes/streams.py
DaisyCisneros: Got it! Thank you for your feedback! The stack overflow post was very informative and the examples in the file were also very helpful! This is was implemented in the most recent commit. 😊

sdk/core/azure-core/tests/test_streaming.py
DaisyCisneros: Thank you! Lines 66-71 were a very helpful guide to making the response stream download. I made sure to pass in stream=True as well 😊. These changes are reflected in the most recent commit!

kristapratico (approved):
@DaisyCisneros this looks great to me! There are some test errors unrelated to your changes that we'll work on getting green! 😎

xiangyan99 (approved):

xiangyan99:
Looks great!

Thanks for the contribution.

iscai-msft:
@DaisyCisneros can you pull from main? I think once that happens this PR is good to go!

PR #21834, opened 2021-11-19
Implement token autorefresh

petrsvihlik:
## Motivation
- Bring the (azure-communication-)common functionality on par with .NET, JS, and Java client libraries

## Implementation
This PR:
- contains an implementation of proactive token refreshing that is already present in .NET, JS, and Java client libraries
- addresses the problem that if the token returned by `token_refresher` callback (implemented by Contoso) is soon to expire or expired, the `AutoRefreshTokenCredential` would keep looping constantly trying to fetch a new token. To prevent this, we implemented a back-off mechanism for the case the token is soon-to-expire and throw an exception if the user fails to provide a non-expired token via the `token_refresher`. We're fixing this particular problem in other SDKs too (see e.g. [JS/TS](https://github.com/Azure/azure-sdk-for-js/pull/19804)). This approach was discussed and recommended by the [Azure Board](https://github.com/Azure/azure-sdk/issues/3712) during an SDK API review.

## Merging notes

- [x] Do not merge before https://github.com/Azure/azure-sdk-for-python/pull/21777 gets merged. If there are any timezone-related conflicts, resolve them  in favor of #21777.

## TODOs
- [x] remove options bag according to the [Python guidelines](https://azure.github.io/azure-sdk/python_design.html#client-configuration) - Communication is the only SDK that has an `*options.py` file - we should remove it
- [x] enabled and fix existing tests
- [x] implement the autorefresh logic for the **sync** version
  - [x] add tests
  - [x] get the naming right (shorten the param names)
  - [x] reflect the datetime vs utc fix in utils.py in #21777
- [x] implement the autorefresh logic for the **async** version
  - [x] add tests
- [x] make sure tests are passing for python 2.*
- [x] make the autrefresh interval configurable
- [x] replicate the logic across the `_shared` folders in all Communication packages


petrsvihlik:
hey @AikoBB @rosebyte @iwonapytel can I pls get some initial feedback on this?

sdk/communication/azure-communication-identity/azure/communication/identity/_shared/user_credential.py
```diff
@@ -4,36 +4,56 @@
...
 
-from .user_token_refresh_options import CommunicationTokenRefreshOptions
 
 class CommunicationTokenCredential(object):
     """Credential type used for authenticating to an Azure Communication service.
     :param str token: The token used to authenticate to an Azure Communication service
-    :keyword token_refresher: The token refresher to provide capacity to fetch fresh token
+    :keyword token_refresher: The async token refresher to provide capacity to fetch fresh token
+    :keyword refresh_proactively: Whether to refresh the token proactively or not
+    :keyword refresh_time_before_expiry: The time before the token expires to refresh the token
```
AikoBB, line 27: we have agreed on the property name `refreshTimeBeforeTokenExpiry` in other languages, so should it be the same in python (` refresh_time_before_token_expiry`)?

sdk/communication/azure-communication-identity/azure/communication/identity/_shared/user_credential.py
```diff
@@ -4,36 +4,56 @@
...
+from .utils import create_access_token
 
-from .user_token_refresh_options import CommunicationTokenRefreshOptions
 
 class CommunicationTokenCredential(object):
     """Credential type used for authenticating to an Azure Communication service.
     :param str token: The token used to authenticate to an Azure Communication service
-    :keyword token_refresher: The token refresher to provide capacity to fetch fresh token
+    :keyword token_refresher: The async token refresher to provide capacity to fetch fresh token
+    :keyword refresh_proactively: Whether to refresh the token proactively or not
```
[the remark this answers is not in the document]
petrsvihlik: right...is "callable" the right thing [here](https://github.com/Azure/azure-sdk-for-python/pull/21834/commits/eb43cb3b35123c109bb4cd0bbf31f685eef2a1e7#)?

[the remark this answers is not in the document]
AikoBB: yes

sdk/communication/azure-communication-identity/azure/communication/identity/_shared/user_credential.py
petrsvihlik: During the last review, it turned out that Python developers prefer rather short names. The thing is that when the parameter is part of `kwargs`, devs need to explicitly type its name. I tried to shorten it here by removing the word token but it doesn't really make much of a difference. Do you have an idea for a better name in this case?

AikoBB: Oh I see, then it is better to leave as it is

sdk/communication/azure-communication-chat/azure/communication/chat/_shared/user_credential.py
```diff
@@ -3,49 +3,70 @@
...
+from .utils import create_access_token
 
 
 class CommunicationTokenCredential(object):
     """Credential type used for authenticating to an Azure Communication service.
     :param str token: The token used to authenticate to an Azure Communication service
-    :keyword token_refresher: The token refresher to provide capacity to fetch fresh token
+    :keyword callable token_refresher: The async token refresher to provide capacity to fetch fresh token
+    :keyword bool refresh_proactively: Whether to refresh the token proactively or not
+    :keyword timedelta refresh_time_before_expiry: The time before the token expires to refresh the token
```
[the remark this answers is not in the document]
petrsvihlik: we're dropping this

sdk/communication/azure-communication-chat/azure/communication/chat/_shared/user_credential.py
```diff
@@ -3,49 +3,70 @@
...
     :raises: TypeError
     """
 
     _ON_DEMAND_REFRESHING_INTERVAL_MINUTES = 2
+    _DEFAULT_AUTOREFRESH_INTERVAL_MINUTES = 10
```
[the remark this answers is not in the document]
petrsvihlik: we're keeping 10 minutes

AikoBB (approved):

sdk/communication/azure-communication-identity/azure/communication/identity/_shared/utils.py
```diff
@@ -84,18 +87,20 @@ def create_access_token(token):
         raise ValueError(token_parse_err_msg)
 
     try:
-        padded_base64_payload = base64.b64decode(parts[1] + "==").decode('ascii')
+        padded_base64_payload = base64.b64decode(
+            parts[1] + "==").decode('ascii')
```
martinbarnas-ms, line 91: Is this formatting idiomatic Python style or is it a mistake?

petrsvihlik: this is what happens if you hit "format document"...so I suppose yes.

martinbarnas-ms (approved):

sdk/communication/azure-communication-phonenumbers/CHANGELOG.md
```diff
@@ -15,6 +15,8 @@ Python 2.7 is no longer supported. Please use Python version 3.6 or later.
 
 - Users can now purchase United Kingdom (GB) toll free and geographic phone numbers for PSTN Calling
 - Users can now purchase Denmark (DK) toll free and geographic phone numbers for PSTN Calling
+- Added support for proactive refreshing of tokens
```
lucasrsant, line 18: Since this version is already released, I believe this should be part of a new version `1.1.0b2`.

sdk/communication/azure-communication-phonenumbers/azure/communication/phonenumbers/_shared/user_credential.py
```diff
@@ -3,84 +3,120 @@
...
-    def _wait_till_inprogress_thread_finish_refreshing(self):
+    def _schedule_refresh(self):
+        if self._timer is not None:
+            self._timer.cancel()
+
+        token_ttl = self._token.expires_on - get_current_utc_as_int()
+
+        if self._is_token_expiring_soon(self._token):
+            # Schedule the next refresh for when it reaches a certain percentage of the remaining lifetime.
+            timespan = token_ttl / 2
```
lucasrsant, line 98: I'm not sure how the `Timer` constructor behaves if you pass a decimal number, I'd suggest using `timespan = token_ttl // 2` to round it.

sdk/communication/azure-communication-phonenumbers/azure/communication/phonenumbers/_shared/user_credential.py
```diff
@@ -3,84 +3,120 @@
...
-    def _is_currenttoken_valid(self):
-        return get_current_utc_as_int() < self._token.expires_on
+    def _is_token_expiring_soon(self, token):
+        if self._refresh_proactively:
+            interval = timedelta(
+                minutes=self._DEFAULT_AUTOREFRESH_INTERVAL_MINUTES)
+        else:
+            interval = timedelta(
+                minutes=self._ON_DEMAND_REFRESHING_INTERVAL_MINUTES)
+        return token.expires_on - get_current_utc_as_int() <\
```
lucasrsant, line 117: nit:
```
return ((token_expires_on - get_current_utc_as_int())
            < interval.total_seconds())
```

sdk/communication/azure-communication-phonenumbers/azure/communication/phonenumbers/_shared/user_credential_async.py
```diff
@@ -3,87 +3,119 @@
...
-    async def _wait_till_inprogress_thread_finish_refreshing(self):
+    def _schedule_refresh(self):
+        if self._timer is not None:
+            self._timer.cancel()
+
+        token_ttl = self._token.expires_on - get_current_utc_as_int()
+
+        if self._is_token_expiring_soon(self._token):
+            # Schedule the next refresh for when it reaches a certain percentage of the remaining lifetime.
+            timespan = token_ttl / 2
```
lucasrsant, line 92: I'm not sure how the `AsyncTimer` constructor behaves if you pass a decimal number, I'd suggest using `timespan = token_ttl // 2` to round it.

sdk/communication/azure-communication-phonenumbers/azure/communication/phonenumbers/_shared/user_credential_async.py
```diff
@@ -3,87 +3,119 @@
...
-        return self._token.expires_on - get_current_utc_as_int() <\
-            timedelta(minutes=self._ON_DEMAND_REFRESHING_INTERVAL_MINUTES).total_seconds()
+    def _is_token_expiring_soon(self, token):
+        if self._refresh_proactively:
+            interval = timedelta(
+                minutes=self._DEFAULT_AUTOREFRESH_INTERVAL_MINUTES)
+        else:
+            interval = timedelta(
+                minutes=self._ON_DEMAND_REFRESHING_INTERVAL_MINUTES)
+        return token.expires_on - get_current_utc_as_int() <\
```
lucasrsant, line 113: nit:
```
return ((token_expires_on - get_current_utc_as_int())
            < interval.total_seconds())
```

sdk/communication/azure-communication-phonenumbers/azure/communication/phonenumbers/_shared/utils.py
```diff
@@ -84,18 +87,20 @@ def create_access_token(token):
         raise ValueError(token_parse_err_msg)
 
     try:
-        padded_base64_payload = base64.b64decode(parts[1] + "==").decode('ascii')
+        padded_base64_payload = base64.b64decode(
+            parts[1] + "==").decode('ascii')
```
lucasrsant, line 91: We should keep quotes consistency.

lucasrsant (changes requested):
Thanks you for the PR. I left some comments in the `azure-communication-phonenumbers` part of the change.

petrsvihlik:
@AikoBB can you please remove the following files?

```
\communication\azure-communication-sms\azure\communication\sms\_shared\user_token_refresh_options.py
\communication\azure-communication-phonenumbers\azure\communication\phonenumbers\_shared\user_token_refresh_options.py
\communication\azure-communication-networktraversal\azure\communication\networktraversal\_shared\user_token_refresh_options.py
\communication\azure-communication-chat\azure\communication\chat\_shared\user_token_refresh_options.py
```

petrsvihlik:
plus, we need to fix this:
image

it's probably caused by too complex type or the multiline comment

sdk/communication/azure-communication-chat/azure/communication/chat/_shared/user_credential.py
```diff
@@ -3,84 +3,121 @@
...
-from .user_token_refresh_options import CommunicationTokenRefreshOptions
+from .utils import create_access_token
 
 
 class CommunicationTokenCredential(object):
     """Credential type used for authenticating to an Azure Communication service.
-    :param str token: The token used to authenticate to an Azure Communication service
-    :keyword token_refresher: The token refresher to provide capacity to fetch fresh token
+    :param str token: The token used to authenticate to an Azure Communication service.
+    :keyword callable token_refresher: The async token
```
annatisch, line 18: This token refresher isn't async is it?

sdk/communication/azure-communication-chat/azure/communication/chat/_shared/user_credential.py
```diff
@@ -3,84 +3,121 @@
...
-    def __init__(self,
-                 token,  # type: str
-                 **kwargs
-                 ):
-        token_refresher = kwargs.pop('token_refresher', None)
-        communication_token_refresh_options = CommunicationTokenRefreshOptions(token=token,
-                                                                               token_refresher=token_refresher)
-        self._token = communication_token_refresh_options.get_token()
-        self._token_refresher = communication_token_refresh_options.get_token_refresher()
+    _DEFAULT_AUTOREFRESH_INTERVAL_MINUTES = 10
```
annatisch, line 26: I thought per the previous archboard review, that the proactive refresh would be done according to a percentage of the remaining validity rather than an absolute time interval?

sdk/communication/azure-communication-chat/azure/communication/chat/_shared/user_credential.py
```diff
@@ -3,84 +3,121 @@
...
+        if not isinstance(token, six.string_types):
+            raise TypeError("Token must be a string.")
+        self._token = create_access_token(token)
+        self._token_refresher = kwargs.pop('token_refresher', None)
+        self._refresh_proactively = kwargs.pop('refresh_proactively', False)
+        self._timer = None
         self._lock = Condition(Lock())
         self._some_thread_refreshing = False
+        if self._refresh_proactively:
+            self._schedule_refresh()
```
annatisch, line 38: I would prefer we didn't start the refresh thread until the first API call is made (i.e. the first time the token is used).
Proactive refresh is about continuously maintaining an existing auth session - however until the first API call is made, so such continuous session exists, so we do need to be actively refreshing it yet.
Potentially we could do this in `__enter__` for people who are using this in a context manager - but for people not using a context manager, the best time to start up the time would be from the first API call.

sdk/communication/azure-communication-chat/azure/communication/chat/_shared/user_credential.py
```diff
@@ -3,84 +3,121 @@
...
-        self._token = communication_token_refresh_options.get_token()
-        self._token_refresher = communication_token_refresh_options.get_token_refresher()
+    _DEFAULT_AUTOREFRESH_INTERVAL_MINUTES = 10
+
+    def __init__(self, token: str, **kwargs: Any):
+        if not isinstance(token, six.string_types):
+            raise TypeError("Token must be a string.")
+        self._token = create_access_token(token)
+        self._token_refresher = kwargs.pop('token_refresher', None)
+        self._refresh_proactively = kwargs.pop('refresh_proactively', False)
```
annatisch, line 33: As mentioned in the meeting - let's make sure it doesn't silently fail if I enable proactive refresh without providing the refresher function.

sdk/communication/azure-communication-chat/azure/communication/chat/_shared/user_credential.py
```diff
@@ -3,84 +3,121 @@
...
                     break
 
         if should_this_thread_refresh:
             try:
-                newtoken = self._token_refresher()  # pylint:disable=not-callable
-
+                new_token = self._token_refresher()
+                if not self._is_token_valid(new_token):
+                    raise ValueError(
+                        "The token returned from the token_refresher is expired.")
```
annatisch, line 77: Should we attempt to validate this in the client?
It seems we could just send the token at let the service reject it with some kind of auth error?

sdk/communication/azure-communication-chat/azure/communication/chat/_shared/user_credential.py
annatisch: Ah - it looks like we are, and this is more like a default window

sdk/communication/azure-communication-chat/azure/communication/chat/_shared/user_credential_async.py
```diff
@@ -3,87 +3,119 @@
...
+            < interval.total_seconds())
 
-    def _is_currenttoken_valid(self):
-        return get_current_utc_as_int() < self._token.expires_on
+    @classmethod
+    def _is_token_valid(cls, token):
+        return get_current_utc_as_int() < token.expires_on
 
     async def close(self) -> None:
         pass
```
annatisch, line 121: As discussed in the meeting - this should implement the logic currently in `__aexit__` - which should then call this function. (We also need this on the sync credential).

sdk/communication/azure-communication-chat/azure/communication/chat/_shared/user_credential_async.py
```diff
@@ -92,4 +124,6 @@ async def __aenter__(self):
         return self
 
     async def __aexit__(self, *args):
+        if self._timer is not None:
+            self._timer.cancel()
```
annatisch, line 128: self._timer should be reset to None so this method is repeatable.

sdk/communication/azure-communication-chat/azure/communication/chat/_shared/utils_async.py
```diff
@@ -0,0 +1,30 @@
...
+    def start(self):
+        self._task = asyncio.ensure_future(self._job())
+
+    async def _job(self):
+        await asyncio.sleep(self._interval)
+        await self._callback()
+
+    def cancel(self):
+        if self._task is not None:
+            self._task.cancel()
```
annatisch, line 30: `self._task = None`?

sdk/communication/azure-communication-chat/tests/_shared/helper.py
```diff
@@ -0,0 +1,49 @@
...
+from datetime import datetime, timedelta
+from functools import wraps
+try:
+    from urllib.parse import urlparse
+except ImportError:
+    from urlparse import urlparse
+import sys
+
+if sys.version_info[0] < 3 or sys.version_info[1] < 4:
+    # python version < 3.3
```
annatisch, line 18: We shouldn't need these checks any longer, currently supported Python versions are only 3.6 - 3.10

sdk/communication/azure-communication-phonenumbers/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## 1.2.0 (Unreleased)
 
+- Added support for proactive refreshing of tokens
+  - `CommunicationTokenCredential` exposes a new boolean keyword argument `refresh_proactively` that defaults to `False`. If set to `True`, the refreshing of the token will be scheduled in the background ensuring continuous authentication state.
```
annatisch, line 6: I'm not sure this needs to be in every package changelog. Most packages weren't exposing these types in the public API - I think it was only Chat and Identity.

petrsvihlik: actually, only chat

sdk/communication/azure-communication-chat/azure/communication/chat/_shared/user_credential.py
AikoBB: it should be sync, fixed

sdk/communication/azure-communication-chat/azure/communication/chat/_shared/user_credential.py
AikoBB: yeah, it is the default expiring soon window. Here is the design of how the autorefresh mechanism works https://github.com/Azure/azure-sdk/issues/3712#issuecomment-1011186174

sdk/communication/azure-communication-chat/azure/communication/chat/_shared/user_credential.py
petrsvihlik: We discussed it with Dom some time ago and agreed we should set clear expectations about how the callback method is supposed to by implemented by Contoso. It doesn't make any sense to allow expired tokens, and throwing an exception is, I think, the right thing to do in this case. Also, this way the code follows the "throw early, catch late" principle. Is it ok from your PoV?

annatisch: We don't think clock skew could be a problem where we raise on a token that might actually have been accepted?
Likewise in reverse - this wont prevent expired token being sent to the service, it will just filter out some of the _really_ expired tokens - so users will have to handle _both_ ValueErrors and Authentication errors in the scenario of expired tokens.

sdk/communication/azure-communication-chat/azure/communication/chat/_shared/user_credential.py
```diff
@@ -3,35 +3,38 @@
...
 class CommunicationTokenCredential(object):
     """Credential type used for authenticating to an Azure Communication service.
-    :param str token: The token used to authenticate to an Azure Communication service
-    :keyword token_refresher: The token refresher to provide capacity to fetch fresh token
-    :raises: TypeError
+    :param str token: The token used to authenticate to an Azure Communication service.
+    :keyword token_refresher: The sync token refresher to provide capacity to fetch a fresh token.
+     The returned token must be valid (expiration date must be in the future).
+    :paramtype token_refresher: Callable[[], AccessToken]
+    :keyword bool proactive_refresh: Whether to refresh the token proactively or not.
```
annatisch, line 21: Maybe it would be helpful to include some information in the docstring about how this would happen? e.g. something like:
"If True, the credential will use a background thread to attempt to refresh the token within 10 minutes, or N% or the tokens remaining validity"

sdk/communication/azure-communication-chat/azure/communication/chat/_shared/user_credential.py
```diff
@@ -3,35 +3,38 @@
...
+    _DEFAULT_AUTOREFRESH_INTERVAL_MINUTES = 10
+
+    def __init__(self, token: str, **kwargs: Any):
+        if not isinstance(token, six.string_types):
+            raise TypeError("Token must be a string.")
+        self._token = create_access_token(token)
+        self._token_refresher = kwargs.pop('token_refresher', None)
+        self._proactive_refresh = kwargs.pop('proactive_refresh', False)
+        if(self._proactive_refresh and self._token_refresher is None):
+            raise ValueError("'token_refresher' must not be None.")
```
annatisch, line 36: I would make this a tiny bit more explicit: "When 'proactive_refresh' is True, 'token_refresher' must not be None"

sdk/communication/azure-communication-chat/azure/communication/chat/_shared/user_credential.py
```diff
@@ -41,46 +44,88 @@ def get_token(self, *scopes, **kwargs):  # pylint: disable=unused-argument
...
+            interval = timedelta(
+                minutes=self._ON_DEMAND_REFRESHING_INTERVAL_MINUTES)
+        return ((token.expires_on - get_current_utc_as_int())
+            < interval.total_seconds())
+
+    @classmethod
+    def _is_token_valid(cls, token):
+        return get_current_utc_as_int() < token.expires_on
+
+    def __enter__(self):
```
annatisch, line 120: If a customer _doesn't_ use a context manager - when's the first time the `schedule_refresh` will be called?
If I'm following the changes correctly - it looks like it will happen at the end of `get_token`, during the `_update_token_and_reschedule` call?

sdk/communication/azure-communication-chat/azure/communication/chat/_shared/user_credential_async.py
```diff
@@ -3,93 +3,134 @@
...
+        self._token = create_access_token(token)
+        self._token_refresher = kwargs.pop('token_refresher', None)
+        self._proactive_refresh = kwargs.pop('proactive_refresh', False)
+        if(self._proactive_refresh and self._token_refresher is None):
+            raise ValueError("'token_refresher' must not be None.")
+        self._timer = None
+        self._async_mutex = Lock()
+        if sys.version_info[:3] == (3, 10, 0):
+            # Workaround for Python 3.10 bug(https://bugs.python.org/issue45416):
+            getattr(self._async_mutex, '_get_loop', lambda: None)()
```
annatisch, line 43: Interesting bug!
I'm a _tiny_ bit concerned that the issue report mentions the fix is compatible with 3.7+, given that we're still tied to 3.6, we should make sure to test it :)

sdk/communication/azure-communication-chat/azure/communication/chat/_shared/utils_async.py
```diff
@@ -0,0 +1,29 @@
...
+
+    def start(self):
+        self._task = asyncio.ensure_future(self._job())
+
+    async def _job(self):
+        await asyncio.sleep(self._interval)
+        await self._callback()
+
+    def cancel(self):
+        self._task = None
```
annatisch, line 29: Wondering if this should call `self._task.cancel()`?

sdk/communication/azure-communication-chat/azure/communication/chat/_shared/user_credential.py
AikoBB: Yeah, as you have mentioned in your previous [comment](https://github.com/Azure/azure-sdk-for-python/pull/21834#discussion_r819115398) for people not using a context manager the scheduling happens from the first API call in the _update_token_and_reschedule.

sdk/communication/azure-communication-chat/azure/communication/chat/_shared/user_credential_async.py
AikoBB: I ran user credential samples and tests on 3.6.12 locally and works fine

sdk/communication/azure-communication-chat/azure/communication/chat/_shared/user_credential.py
petrsvihlik: This is a valid point and yes, it can happen. 
I wonder - Would you feel different about an argument null check? To me, they feel very similar - they're just a set of pre-conditions that help eliminate possible invalid scenarios, but never 100% guarantee that something won't fail later on.
The necessity of handling this type of errors still feels like a small price to pay for receiving a clear and concise message about the source of the error early on. 
I'm really curious about your PoV.

sdk/communication/azure-communication-chat/samples/user_credential_sample.py
```diff
@@ -0,0 +1,72 @@
...
+
+import os
+from azure.communication.chat import CommunicationTokenCredential
+from azure.communication.identity import CommunicationIdentityClient
+
+class CommunicationTokenCredentialSamples(object):
+    
+    connection_string = os.environ.get("COMMUNICATION_SAMPLES_CONNECTION_STRING", None)
+    if not connection_string:
+        raise ValueError("Set COMMUNICATION_SAMPLES_CONNECTION_STRING env before run this sample.")
```
petrsvihlik, line 30:
```suggestion
        raise ValueError("Set COMMUNICATION_SAMPLES_CONNECTION_STRING env before running this sample.")
```

sdk/communication/azure-communication-chat/samples/user_credential_sample.py
```diff
@@ -0,0 +1,72 @@
...
+            print("Token issued with value: " + tokenresponse.token)
+
+    def create_credential_with_refreshing_callback(self):
+        # Alternatively, for long-lived clients, you can create a `CommunicationTokenCredential` with a callback to renew tokens if expired.
+        # Here we assume that we have a function `fetch_token_from_server` that makes a network request to retrieve a token string for a user.
+        # It's necessary that the `fetch_token_from_server` function returns a valid token (with an expiration date set in the future) at all times.
+        fetch_token_from_server = lambda: None 
+        with CommunicationTokenCredential(
+            self.token, token_refresher=fetch_token_from_server) as credential:
+            tokenresponse = credential.get_token()
```
petrsvihlik, line 51:
```suggestion
            token_response = credential.get_token()
```

sdk/communication/azure-communication-chat/samples/user_credential_sample.py
```diff
@@ -0,0 +1,72 @@
...
+            print("Token issued with value: " + tokenresponse.token)
```
petrsvihlik, line 52:
```suggestion
            print("Token issued with value: " + token_response.token)
```

sdk/communication/azure-communication-chat/samples/user_credential_sample.py
```diff
@@ -0,0 +1,72 @@
...
+
+    def create_credential_with_proactive_refreshing_callback(self):
+        # Optionally, you can enable proactive token refreshing where a fresh token will be acquired as soon as the
+        # previous token approaches expiry. Using this method, your requests are less likely to be blocked to acquire a fresh token
+        fetch_token_from_server = lambda: None 
+        with CommunicationTokenCredential(
+            self.token, token_refresher=fetch_token_from_server, proactive_refresh=True) as credential:
+            tokenresponse = credential.get_token()
```
petrsvihlik, line 60:
```suggestion
            token_response = credential.get_token()
```

sdk/communication/azure-communication-chat/samples/user_credential_sample.py
```diff
@@ -0,0 +1,72 @@
...
+            print("Token issued with value: " + tokenresponse.token)
```
petrsvihlik, line 61:
```suggestion
            print("Token issued with value: " + token_response.token)
```

petrsvihlik: the same in aio

sdk/communication/azure-communication-chat/samples/user_credential_sample.py
petrsvihlik: the same in aio

sdk/communication/azure-communication-chat/azure/communication/chat/_shared/user_credential.py
annatisch: There are similarities, whereas an argument check feels like it would be more of an error that would occur during development, causing someone to correct their code.
My feeling is that this could more likely occur due (depending on how the users callback acquires a token) to awkward timing, maybe a flaky network for example, that could result in the same code sometimes giving a ValueError and sometimes given a service AuthenticationError.
I guess another factor to consider here would be retry - if a ValueError is raised because the callback ran slowly for some reason - should that be retryable? Is that up to the user to retry? If instead it's rejected by the service - will that/should that be retried in the pipeline or by customers? (I don't think the pipeline automatically retries auth errors).

annatisch: hmm good question - if I send an expired token to the service, what status code would I receive?
It's possible this wont be retried, and simply raised directly to the customer.
We can double check the code path/pipeline retry policy to confirm :)

petrsvihlik: @AikoBB informed me that you've discussed this one before she left for a vacation. -> resolving
MicrosoftTeams-image

sdk/communication/azure-communication-chat/azure/communication/chat/_shared/user_credential.py
```diff
@@ -3,84 +3,139 @@
...
-    :keyword token_refresher: The token refresher to provide capacity to fetch fresh token
-    :raises: TypeError
+    :param str token: The token used to authenticate to an Azure Communication service.
+    :keyword token_refresher: The sync token refresher to provide capacity to fetch a fresh token.
+     The returned token must be valid (expiration date must be in the future).
+    :paramtype token_refresher: Callable[[], AccessToken]
+    :keyword bool proactive_refresh: Whether to refresh the token proactively or not.
+     If the proactive refreshing is enabled ('proactive_refresh' is true), the credential will use
+     a background thread to attempt to refresh the token within 10 minutes before the cached token expires,
+     the proactive refresh will request a new token by calling the 'token_refresher' callback.
```
annatisch, line 24: Could we add a note to the documentation that if `proactive_refresh` is enabled, the Credential object must be either run within a context manager, or `close` called once the object has been finished with.

sdk/communication/azure-communication-chat/azure/communication/chat/_shared/user_credential_async.py
```diff
@@ -3,93 +3,137 @@
...
-
-    async def close(self) -> None:
-        pass
+    @classmethod
+    def _is_token_valid(cls, token):
+        return get_current_utc_as_int() < token.expires_on
 
     async def __aenter__(self):
+        if self._proactive_refresh:
+            self._schedule_refresh()
```
annatisch, line 130: It looks like `aenter` and `aexit` were already defined on the async credential object - in this case there might already be customers who are reusing this object. In that case we should only raise the RuntimeError _if_ they have set `proactive_refresh=True`

sdk/communication/azure-communication-chat/azure/communication/chat/_shared/user_credential.py
```diff
@@ -3,84 +3,139 @@
...
+        return ((token.expires_on - get_current_utc_as_int())
+            < interval.total_seconds())
+
+    @classmethod
+    def _is_token_valid(cls, token):
+        return get_current_utc_as_int() < token.expires_on
+
+    def __enter__(self):
+        if self._is_closed.is_set():
+            raise RuntimeError("An instance of CommunicationTokenCredential cannot be reused once it has been closed.")
```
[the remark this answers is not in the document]
AikoBB: Yes, I have already added a test for the context manager [here](https://github.com/Azure/azure-sdk-for-python/blob/dcf0e826cc676c02b2e5ad09ca8cfa3e68f68347/sdk/communication/azure-communication-identity/tests/test_user_credential.py#L217), will add another one for the case using credential without context manager and will update async credential as well

sdk/communication/azure-communication-chat/azure/communication/chat/_shared/user_credential.py
```diff
@@ -3,84 +3,139 @@
...
-
                 raise
+        if self._proactive_refresh:
+            self._schedule_refresh()
         return self._token
 
-    def _wait_till_inprogress_thread_finish_refreshing(self):
+    def _schedule_refresh(self):
+        if self._is_closed.is_set():
+            return
```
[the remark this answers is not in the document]
AikoBB: This check was added for pypy issue when the canceled thread tries to reschedule another timer, and for that reason, I think we do not have to throw an exception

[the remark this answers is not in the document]
annatisch: Yeah I think that makes sense.
I'm just wondering if someone calls `credential.close()` then attempts to reuse the credential object without a context manager - this would silently ignore the proactive refresh.
But maybe there's a better place to check for that.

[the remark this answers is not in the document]
AikoBB: Yeah, I understood your point, I think the good place to check is the get_token method, what do you think?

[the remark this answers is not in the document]
annatisch: That could work - though it would probably need to be within the `if self._proactive_refresh` clause.

sdk/communication/azure-communication-chat/azure/communication/chat/_shared/user_credential.py
AikoBB: Added [async](https://github.com/Azure/azure-sdk-for-python/pull/21834/commits/be1aa9f9e6e232bf443706daac6ce17d6c974674#diff-dbfc5aa9546a717c64a9b0df2b572ecdda52759bde3aa7e51af95c44a700be80) and [sync](https://github.com/Azure/azure-sdk-for-python/pull/21834/commits/be1aa9f9e6e232bf443706daac6ce17d6c974674#diff-616d1098edeb23af7081e6a0b40703cb5e2ca2b0b97b680fef43e99ec813eac1) tests for the using credential without a context manager scenarios

sdk/communication/azure-communication-chat/azure/communication/chat/_shared/user_credential.py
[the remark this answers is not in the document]
AikoBB: Added [async](https://github.com/Azure/azure-sdk-for-python/pull/21834/commits/be1aa9f9e6e232bf443706daac6ce17d6c974674#diff-dbfc5aa9546a717c64a9b0df2b572ecdda52759bde3aa7e51af95c44a700be80) and [sync](https://github.com/Azure/azure-sdk-for-python/pull/21834/commits/be1aa9f9e6e232bf443706daac6ce17d6c974674#diff-616d1098edeb23af7081e6a0b40703cb5e2ca2b0b97b680fef43e99ec813eac1) tests for the using credential without a context manager scenarios

sdk/communication/azure-communication-chat/azure/communication/chat/_shared/user_credential_async.py
AikoBB: Yeah, I put the logic inside the [proactive refresh check](https://github.com/Azure/azure-sdk-for-python/blob/be1aa9f9e6e232bf443706daac6ce17d6c974674/sdk/communication/azure-communication-identity/azure/communication/identity/_shared/user_credential_async.py#L138) and did the same for [the sync version](https://github.com/Azure/azure-sdk-for-python/blob/be1aa9f9e6e232bf443706daac6ce17d6c974674/sdk/communication/azure-communication-identity/azure/communication/identity/_shared/user_credential.py#L132) to make them consistent

annatisch (approved):
Thanks so much @AikoBB for all your work on this!!!

LuChen-Microsoft (approved):

LuChen-Microsoft (approved):
Looks good to me

Issue #21101, opened 2021-10-06
[EventHubs] Updates async sample code to drop loop


yunhaoling:
in current async samples, we use the following pattern to run code
```
loop = asyncio.get_event_loop()
loop.run_until_complete(...)
```

However, starting from python 3.10, loop is deprecated.
we should update to use e.g. `asyncio.run(...)`.

NOTE:  asyncio.run is introduced in python 3.7
our samples will be run on python3.6, so need to check if it would break our CI.

PR #21857, opened 2021-11-23
[EventHubs] Update async samples

sammiee5311:
Related to #21101

I've updated EventHubs samples to drop `loop.run_until_complete(...)`(updated to `asyncio.run(...)`)

I did not modified those files.

- [`client_creation_async.py`](https://github.com/Azure/azure-sdk-for-python/blob/fe37155dcb6bf66491cf59a4d3cd65763b2aa01c/sdk/eventhub/azure-eventhub/samples/async_samples/client_creation_async.py#L100)
- [`connection_to_custom_endpoint_address_async.py`](https://github.com/Azure/azure-sdk-for-python/blob/fe37155dcb6bf66491cf59a4d3cd65763b2aa01c/sdk/eventhub/azure-eventhub/samples/async_samples/connection_to_custom_endpoint_address_async.py#L73)
- [`sample_code_eventhub_async.py`](https://github.com/Azure/azure-sdk-for-python/blob/fe37155dcb6bf66491cf59a4d3cd65763b2aa01c/sdk/eventhub/azure-eventhub/samples/async_samples/sample_code_eventhub_async.py#L210)

Because I'm not sure if the behavior will be changed or not once I modified the code.

For examples, 

**from**
``` python
loop.run_until_complete(create_producer_client())
loop.run_until_complete(create_consumer_client())
```
**to**
``` python
async def main():
    await create_producer_client()
    await create_consumer_client()

asyncio.run(main())
```

### DIFF
--- a/sdk/eventhub/azure-eventhub/samples/async_samples/authenticate_with_sas_token_async.py
+++ b/sdk/eventhub/azure-eventhub/samples/async_samples/authenticate_with_sas_token_async.py
@@ -83,7 +83,6 @@ async def create_with_sas_token():
         await producer_client.send_batch(event_data_batch)
 
 
-loop = asyncio.get_event_loop()
 start_time = time.time()
-loop.run_until_complete(create_with_sas_token())
+asyncio.run(create_with_sas_token())
 print("Send messages in {} seconds.".format(time.time() - start_time))
--- a/sdk/eventhub/azure-eventhub/samples/async_samples/client_identity_authentication_async.py
+++ b/sdk/eventhub/azure-eventhub/samples/async_samples/client_identity_authentication_async.py
@@ -61,6 +61,6 @@ async def run():
                     break
             await producer.send_batch(event_data_batch)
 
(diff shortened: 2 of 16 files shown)


swathipil:
@sammiee5311 great, thank you! These look good to me! I'll run the tests and merge this PR once the tests pass.

Regarding your example, those two snippets are doing different things, as you mentioned. The `await` is non-blocking execution, so it will allow other tasks to execute while waiting, whereas the `run_until_complete` is blocking, so it will ensure that `create_producer_client` is completed before calling `create_consumer_client`.

It shouldn't matter much in this scenario, since it's just an example of initializing clients. However, we should keep the original behavior, so I can update those examples as needed later.

Thanks again for your contribution!

sammiee5311:
@swathipil thank you so much for detailed explaination ! It helps me understand very well so now I understand how `await` works !

No worries. Even if It is not a big contribution, thanks for your kindness !

swathipil (approved):

Issue #19939, opened 2021-07-26
[textanalytics] establish consistency on setting default params


kristapratico:
If the service specifies that a param defaults to something, e.g. `disable_service_logs` defaults to `True`, then it would be better if we don't set that default in the client library directly and just let the service handle it on their side. We have a mix of both strategies in TA for parameters with default values and should work on taking a consistent approach while ensuring any changes do not cause unintended behavior.

PR #21866, opened 2021-11-23
[textanalytics] establish consistency on setting default params

kristapratico:
Resolves https://github.com/Azure/azure-sdk-for-python/issues/19939

If the service specifies that a param defaults to something, e.g. `disable_service_logs` defaults to `True`, then it would be better if we don't set that default in the client library directly and just let the service handle it on their side (similar to how the other languages do it). We have a mix of both strategies in TA for parameters with default values and should work on taking a consistent approach while ensuring any changes do not cause unintended behavior.

Needed to re-record all tests since we used to pass `show_stats=False` for all endpoints unless otherwise specified in the tests.

~~PR pending on re-recording tests - we're seeing a few InternalServerErrors with the analyze endpoint (reported to service team).~~


iscai-msft (approved):
thank you queen

PR #21869, opened 2021-11-23
[ServiceBus] Fix TTL on messages

jyggen:
Using `seconds` on `timedelta` to set TTL leads to unexpected behaviour with, for example, `timedelta(days=1)`: 

```
azure.servicebus.exceptions.ServiceBusError: Argument TimeToLIve must be a positive timeout value. The provided value was 00:00:00.
Parameter name: TimeToLIve
Actual value was 00:00:00
```

Instead of `seconds`, `total_seconds()` should be used:
```
>>> timedelta(days=1).seconds
0
>>> timedelta(days=1).total_seconds()
86400.0
```


jyggen:
Realized that this isn't compatible with Python 2.7, so will have to wait until 2022 I guess :)

yunhaoling:
nice catch! thanks @jyggen for the PR!
it looks good to me, will merge it after passing the tests.

yunhaoling (approved):
test passed, appreciate your contribution, your fix will be included in our next release🎉

PR #21885, opened 2021-11-24
add error message when no recording file exists

kristapratico:
just a suggestion, feel free to reject the PR.

A `KeyError` is raised when the `x-recording-id` cannot be found. My understanding is that this happens when trying to run playback without a recording file present for the specific test (I've accidentally done this a few times myself). Suggestion to raise an error message that helps explains the problem.


mccoyp (approved):
I somehow just saw this, so sorry! Thanks for this change!

Issue #21849, opened 2021-11-22
Async `EventHubProducerClient` + `AzureFunctions`


Stael:
Hello !
I am posting this ticket here, but please let me know if you think I should close it and post it on [Azure/azure-functions-python-worker](https://github.com/Azure/azure-functions-python-worker/issues/755) instead.

We are currently using several AzureFunctions in production :
- Premium V2 pricing tier
- Operating system: Linux
- Plan type: App service plan
- Python version 3.6

Use case:
- Receive data from HTTP requests and send those data to EventHub using `azure-eventhub` package.

The AzureFunction has a timeout of X seconds, meaning that after X seconds, the function is "aborted" (we have been experimenting with 15/30/60/120sec).

---

We have 2 versions of the code : 

"Threaded Version": the one currently in production which doesn't use async/await but uses threading (See [here](https://github.com/Azure/azure-functions-python-worker/issues/755)). This version was created to mitigate issue 2 (see below).
```python
def threaded_execution(f):
    def wrapper(*args, **kwargs):
        thread = Thread(target=f, args=args, kwargs=kwargs)
        thread.start()

    return wrapper

class EventProducer:
    def __init__(self, password: str):
        self._producer: EventHubProducerClient = EventHubProducerClient.from_connection_string(password)

    @threaded_execution
    def produce(self, data: dict):
        bson_encoded_data = bson.BSON.encode(data)
        b64_encoded_data = base64.b64encode(bson_encoded_data)
        try:
            self._producer.send_batch([EventData(b64_encoded_data)])
        except Exception as e:
            logging.error(...)
```

The function code looks like this:
```python
event_producer: EventProducer = EventProducer(...)

def main(req: func.HttpRequest) -> func.HttpResponse:
    data: dict = build_event_data(req)
    event_producer.produce(data)
    return func.HttpResponse(status_code=200)
```

"Async/Await Version": Another one which we created and released in "staging" because we were unable to debug issue 1 while using the "threading version" of the code.

Our producer code looks like this: 

```python
class EventProducer:
    def __init__(self, password: str):
        self._password: str = password

    async def produce(self, data: dict):
        bson_encoded_data = bson.BSON.encode(data)
        b64_encoded_data = base64.b64encode(bson_encoded_data)
        client = EventHubProducerClient.from_connection_string(self._password)
        async with client:
            await client.send_batch([EventData(b64_encoded_data)])
```

The function code looks like this:
```python
[...]

PR #21927, opened 2021-11-30
[EventHubs] Fix bug in reusing EventHubProducerClient

yunhaoling:
Addressing issue: https://github.com/Azure/azure-sdk-for-python/issues/21849

fix EventHubProducerClient reusability


sdk/eventhub/azure-eventhub/CHANGELOG.md
```diff
@@ -8,6 +8,8 @@
 
 ### Bugs Fixed
 
+- Fixed a bug that `EventHubProduerClient` is not reusable after getting closed (issue #21849).
```
swathipil, line 11: can you explain what you mean by not reusable currently? (just for my own sake, sorry - I read through the issue and the code, but I don't fully understand what the bug is).

yunhaoling: it is the case when the same client which is opened, closed, then opened again.

```python
producer = EventHubProducerClient(...)
with producer:
   #...
   # the context manager will close the underlying amqp connection

# error would be raised if I want to reuse this producer
with producer:  # the same producer can't work because of the bug when retrieving internal partition producers
```

swathipil (approved):
small nit, but o/w LGTM!

PR #21940, opened 2021-11-30
[Key Vault] Use shared test-resources.json

mccoyp:
This aligns Python's KV `test-resources.json` with [.NET's](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/keyvault/test-resources.json), which is or will be used by other languages as well.


maorleger (approved):
🚀

PR #21945, opened 2021-11-30
add license files

xiangyan99:


weshaggard:
I cannot successfully review this on the web so just adding comments here. Can you please use the exact text from https://docs.opensource.microsoft.com/releasing/maintain/license/? Also from a quick look it seems like some of the places have both a LICENSE and LICENSE.txt file. Please remove the LICENSE.txt file and any references to them.

PR #21950, opened 2021-12-01
Pin Crypto Temporarily

scbedd:
The true path forward will be one of two options.

1. Stop using `UsePythonVersion` in favor of our own hand-rolled version. It's not too bad, couple hours.
2. Finish [the PR to the task itself](https://github.com/microsoft/azure-pipelines-tasks/pull/15553). I'm not optimistic about engagement there.

For those of you that are lost. The timeline of events:

1. crypto 36 released
2. it started crashing on pypy (even though they ship pypy wheels)

The reason behind this is that the `UsePythonVersion` task is hardcoded to go after `pypy w/2.7` or `pypy w/3.6` ONLY. It legit doesn't even parse the versionspec.

### DIFF
--- a/eng/test_tools.txt
+++ b/eng/test_tools.txt
@@ -2,6 +2,7 @@ pip==20.3
 
 # python 3.10 pinned packages
 cffi==1.15.0rc2; python_version >= '3.10'
+cryptography==35.0.0
 
 # requirements leveraged by ci for testing
 pytest==4.6.9; python_version == '2.7'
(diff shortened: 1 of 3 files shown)


RyanZzzz (approved):

eng/test_tools.txt
```diff
@@ -2,6 +2,8 @@ pip==20.3
 
 # python 3.10 pinned packages
 cffi==1.15.0rc2; python_version >= '3.10'
+cryptography==35.0.0; python_version > '2.7'
+cryptography==3.3; python_version == '2.7'
```
scbedd, line 6: 3.3 == last version that supported py 2.7

zihzhan-msft (approved):

scbedd:
It's a transient failure, overriding and merging.

PR #21960, opened 2021-12-01
work around for crypto issue

xiangyan99:


mccoyp:
This should work for Key Vault, so I think it looks good. Is the PEM formatting issue on 35.0.0 the reason not to pin to that version?

scbedd (approved):

PR #21990, opened 2021-12-02
Issue #9324 adding offline test for test_streaming.py and streams.py

emilychang9:
This PR will:
- mark test_decompress_plain_header() and async test_decompress_plain_header() as live tests only
- add the offline equivalent test for synchronous test_decompress_plain_header() called test_decompress_plain_header_offline()
- add a function to the offline test server to simulate the test behavior

### DIFF
--- a/sdk/core/azure-core/tests/test_streaming.py
+++ b/sdk/core/azure-core/tests/test_streaming.py
@@ -115,12 +115,15 @@ def test_decompress_plain_header(http_request):
 
 @pytest.mark.parametrize("http_request", HTTP_REQUESTS)
 def test_decompress_plain_header_offline(port, http_request):
-    request = http_request(method="GET", url="http://localhost:{}/streams/compressed".format(port),
-                           headers={"Content-Type": "gzip"})
+    request = http_request(method="GET", url="http://localhost:{}/streams/compressed".format(port))
     with RequestsTransport() as sender:
-        response = sender.send(request)
+        response = sender.send(request, stream=True)
         response.raise_for_status()
-        assert response.text() == '{"error": headers={"Content-Type": "gzip"}}'
+        try:
+            result = response.read()
+            assert result.text() == '{"error": headers={"Content-Type": "gzip"}}'
+        except AttributeError:
+            pass
 
 @pytest.mark.parametrize("http_request", HTTP_REQUESTS)
 def test_compress_plain_header(http_request):
--- a/sdk/core/azure-core/tests/test_streaming.py
+++ b/sdk/core/azure-core/tests/test_streaming.py
@@ -26,6 +26,7 @@
 import pytest
 from azure.core import PipelineClient
 from azure.core.exceptions import DecodeError
+from azure.core.pipeline.transport import RequestsTransport
 from utils import HTTP_REQUESTS
 
 @pytest.mark.parametrize("http_request", HTTP_REQUESTS)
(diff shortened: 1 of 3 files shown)


sdk/core/azure-core/tests/test_streaming.py
```diff
@@ -111,6 +113,19 @@ def test_decompress_plain_header(http_request):
...
         pass
 
+@pytest.mark.parametrize("http_request", HTTP_REQUESTS)
+def test_decompress_plain_header_offline(port, http_request):
+    request = http_request(method="GET", url="http://localhost:{}/streams/compressed".format(port))
+    with RequestsTransport() as sender:
+        response = sender.send(request, stream=True)
+        response.raise_for_status()
+        try:
+            result = response.read()
```
kristapratico, line 123: Like in the above online test, I think we want to stream download the response similar to L109 with the `decompress=True` keyword argument passed. This should signal that we want to decompress the body based on the Content-Encoding header and we should get the expected behavior of failure

sdk/core/azure-core/tests/testserver_tests/coretestserver/coretestserver/test_routes/streams.py
```diff
@@ -36,3 +39,8 @@ def iterable():
 @streams_api.route('/error', methods=['GET'])
 def error():
     return Response(stream_json_error(), status=400)
+
+@streams_api.route('/compressed', methods=['GET'])
+def compressed():
+    return Response(stream_compressed_header_error(), status=300, headers={"Content-Type": "gzip"})
```
kristapratico, line 45: I believe the header passed should be "Content-Encoding" here instead of "Content-Type". @iscai-msft - can you give me a sanity check? :)

kristapratico (commented):
Awesome job!! Looking really good 😎

sdk/core/azure-core/tests/test_streaming.py
kristapratico: with pytest.raises(DecodeError):
   list(data)

sdk/core/azure-core/tests/testserver_tests/coretestserver/coretestserver/test_routes/streams.py
emilychang9: Done!

sdk/core/azure-core/tests/test_streaming.py
emilychang9: Included these changes in the most recent commit!

sdk/core/azure-core/tests/test_streaming.py
```diff
@@ -105,11 +107,18 @@ def test_decompress_plain_header(http_request):
     pipeline_response = client._pipeline.run(request, stream=True)
     response = pipeline_response.http_response
     data = response.stream_download(client._pipeline, decompress=True)
-    try:
-        content = b"".join(list(data))
-        assert False
-    except DecodeError:
-        pass
+    with pytest.raises(DecodeError):
```
iscai-msft, line 110: yay thank you so much for improving the old test!

iscai-msft (approved):
@emilychang9 looks like it's passing the core CI. I think there are additional failures related to identity and storage CI, and I don't think they have anything to do with you. Let me try to help get this PR green. Thank you so so much again, this is great work!

kristapratico (approved):
Thanks for addressing my comments. Great job! 🚀

PR #21992, opened 2021-12-03
[translation] remove sample tests dependency on static storage acc

kristapratico:
The sample tests for Document Translation depend on a persistent storage account which needs a container with two small files to test the scenarios in the live CI. We originally were using a static storage account for this in the test-resources.json, but I want to fix this because...

1) This isn't the right way to use static resources in our live tests
2) This complicates adding additional cloud configs, e.g. Dogfood. However, I don't think it makes sense to have a static storage account across clouds for just two small files.

The rest of the resources and containers needed for tests/samples can be be dynamically created and torn down as expected.

Live test run pass: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=1226274&view=results

### DIFF
--- a/sdk/translation/azure-ai-translation-document/samples/assets/translate.txt
+++ b/sdk/translation/azure-ai-translation-document/samples/assets/translate.txt
@@ -1,0 +1,1 @@
+hello world
--- a/sdk/translation/test-resources-post.ps1
+++ b/sdk/translation/test-resources-post.ps1
@@ -1,20 +1,3 @@
-[CmdletBinding(SupportsShouldProcess = $true, ConfirmImpact = 'Medium')]
-param (
-    [Parameter(Mandatory = $true)]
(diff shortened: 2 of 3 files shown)


kristapratico:
Hey @benbp wanted to get your opinion on this PR. 

I want to add the Dogfood config to translation live tests, but realized that we currently have a static storage account defined in the test-resources.json which is used to run samples in the CI. All we really need to run samples successfully are two small files which exist locally in the repo. I'm wondering if this is appropriate usage of the test-resources-post.ps1 step to just upload these files to the container created in the template so we can run samples and then tear everything down at the end.

Let me know if there is another way you would recommend over this approach.

benbp:
This is great. I'm totally behind removing static resources in favor of a reproducible/cloud independent approach.

benbp (approved):

PR #22000, opened 2021-12-03
[translation] update yaml to include dogfood

kristapratico:
Resolves https://github.com/Azure/azure-sdk-for-python/issues/22001

~Pending on some questions/answers from the service team.~

### DIFF
--- a/sdk/translation/azure-ai-translation-document/tests/test_translation.py
+++ b/sdk/translation/azure-ai-translation-document/tests/test_translation.py
@@ -32,7 +32,7 @@ class TestTranslation(DocumentTranslationTest):
     def test_active_directory_auth(self):
         token = self.generate_oauth_token()
         endpoint = self.get_oauth_endpoint()
-        client = DocumentTranslationClient(endpoint, token, credential_scopes=["https://ppe.cognitiveservices.azure.com/.default"])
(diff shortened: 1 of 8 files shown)


kristapratico:
@benbp would you mind reviewing this? This enables dogfood testing and conditionally deploys a storage acc for prod. I set up a static storage acc in the static-test-resources rg for dogfood.

sdk/translation/azure-ai-translation-document/tests/test_translation.py
```diff
@@ -32,7 +32,10 @@ class TestTranslation(DocumentTranslationTest):
     def test_active_directory_auth(self):
         token = self.generate_oauth_token()
         endpoint = self.get_oauth_endpoint()
-        client = DocumentTranslationClient(endpoint, token)
+        kwargs = {}
+        if os.getenv("AZURE_AUTHORITY_HOST", "").find("ppe") != -1:
+            kwargs["credential_scopes"] = ["https://ppe.cognitiveservices.azure.com/.default"]
```
benbp, line 37: An alternative would be to set the credential scope as an env var in the subscription config or as an output in the arm template, that way you don't have to hardcode endpoints into the test.

benbp (approved):

sdk/translation/azure-ai-translation-document/tests/test_translation.py
kristapratico: Great idea, going to update to do that.

PR #22013, opened 2021-12-06
[Search] Update Similarity property metadata

mrwbarg:
Rename property values for _subtype_map under Similarity.


tjprescott:
Hi @mrwbarg is there a particular issue you are fixing with this change?

mrwbarg:
Hello,

Yes! There are no keys on the "objects" dictionary corresponding to the
current names. Both have the "Algorithm" suffix.

tjprescott:
@mrwbarg the change you made was in the _generated folder, which cannot be manually modified. If you are encountering an issue with serialization, please open a bug report and I can take a look to address it.

PR #22014, opened 2021-12-06
[Search] Update Similiarity metadata (in Py3 models).

mrwbarg:
Rename _subtype_map property values under Similarity.


tjprescott:
Hi @mrwbarg is there a specific issue you are encountering that this PR fixes?

tjprescott:
@mrwbarg the change you made was in the _generated folder, which cannot be manually modified. If you are encountering an issue with serialization, please open a bug report and I can take a look to address it.

PR #22024, opened 2021-12-06
Resolve Installation Ordering Issues Preventing PR Checks of dependent packages

scbedd:
@xiangyan99 @iscai-msft 

The issue was introduced when I implemented the addition of some additional restrictions when pulling from PyPI.

Nightly builds would still be pulling the latest version of azure-core, but PR builds wouldn't be getting the checks properly as @xiangyan99 and @iscai-msft identified to me this morning.

I believe this should be the only fix, but I'm running some additional verification jobs to be certain of that.

The cause of this bug is that when we are installing a target package, we evaluate it's dependencies for `azure` specific packages. We go and grab those from a our single dev feed. The problem is that we _always_ do this, meaning that we **replace** the relative dependency version already installed with `dev_requirements.txt`.

The bugfix checks the already installed packages and takes no action if an already installed version matches a version spec from `install_requires`.

### DIFF
--- a/eng/tox/create_package_and_install.py
+++ b/eng/tox/create_package_and_install.py
@@ -17,6 +17,7 @@ import sys
 import glob
 import shutil
 from pkg_resources import parse_version
+import pdb
 
 
 from tox_helper_tasks import find_whl, find_sdist, get_package_details, get_pip_list_output, parse_req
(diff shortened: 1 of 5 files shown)


iscai-msft (approved):

scbedd:
This looks g2g afaict. I'll take look at all the daily builds tomorrow.

PR #22030, opened 2021-12-07
Issue #9324 create test_compress_compressed_no_header_offline

marcusc2:
- mark test_compress_compressed_no_header() as live test only
- create test_compress_compressed_no_header_offline()
- add stream_compress() and compress() in streams.py to replicate online test behavior


sdk/core/azure-core/tests/testserver_tests/coretestserver/coretestserver/test_routes/streams.py
```diff
@@ -25,6 +25,14 @@ def stream_json_error():
     yield '{"error": {"code": "BadRequest", '
     yield' "message": "You made a bad request"}}'
 
+def stream_compress():
+    with open("testserver_tests/coretestserver/coretestserver/files/test.tar.gz", "rb") as fd:
```
kristapratico, line 29: You may need to use an absolute path to the file here. Something like:

```
import os
file_path = os.path.abspath(os.path.join(os.path.abspath(__file__), "..", "..", "./files/test.tar.gz"))
```

kristapratico (commented):
Woohoo looking good! 😃

iscai-msft (approved):
Looks great! Once CI is fixed I think we're good to go! Thank you so much @marcusc2

sdk/core/azure-core/tests/test_streaming.py
```diff
@@ -76,6 +77,19 @@ def test_decompress_compressed_no_header(http_request):
...
+@pytest.mark.parametrize("http_request", HTTP_REQUESTS)
+def test_compress_compressed_no_header_offline(port, http_request):
+    # expect compressed text
+    request = http_request(method="GET", url="http://localhost:{}/streams/compress".format(port))
+    with RequestsTransport() as sender:
+        response = sender.send(request, stream=True)
+        response.raise_for_status()
+        data = response.stream_download(sender, decompress=False)
+        content = b"".join(list(data))
+        with pytest.raises(UnicodeDecodeError):
```
iscai-msft, line 89: Love the with pytest.raises!

catalinaperalta (approved):
This looks great! :) I think if you just fix the file path like Krista and Izzy suggested CI should pass, since it's pointing out that the file could not be found.

scbedd:
We have triggered a build storm with this PR. It may be just a bad git merge, or something else, but the files changed is reflecting pretty much every package in the repo that I can tell.

@marcusc2 no more commits to this PR please.

PR #22102, opened 2021-12-09
Update the file based on the fabric config data

AlexGhiondea:
/cc @jsquire @lilyjma @lmazuel


AlexGhiondea:
Where it makes sense we should see if we can collapse the PR label and the issue triage (if they have the SAME list of people).

@lmazuel can you help with that?

jsquire (approved):
Changes line up with bot rules, in spot checked cases.   LGTM

lmazuel (approved):

PR #22104, opened 2021-12-09
update shared req for mgmt libraries

kristapratico:
When this [PR ](https://github.com/Azure/azure-sdk-for-python/commit/50471e7f07b430ab0bbc1984cd2fc102d908de41)went in, the msrest version was bumped in many of the mgmt libraries due to the auto-packaging updates.


kristapratico:
Updating all of these to  msrest>=0.6.21, but wondering if the frozen requirement for msrest should just be updated instead? 

https://github.com/Azure/azure-sdk-for-python/blob/4db8247d519b48051eb9f88d1c2c44949b6a45ba/shared_requirements.txt#L117

scbedd (approved):

scbedd:
@kristapratico if none remain with the old one, it's definitely better to just update the frozen one.

kristapratico:
> @kristapratico if none remain with the old one, it's definitely better to just update the frozen one.

I'm actually not sure. Going to merge this to unblock PRs and then I'll try in separate PR.

PR #22112, opened 2021-12-09
Added offline version of test_decompress_compressed_header

zhzhh97:
Added test_decompress_compressed_header_offline()

Added stream_decompress() and decompress() in streams.py to mimic the online test behavior

Marked test_decompress_compressed_header() as live_test_only


sdk/core/azure-core/tests/test_streaming.py
```diff
@@ -150,6 +151,18 @@ def test_decompress_compressed_header(http_request):
...
     assert decoded == "test"
 
+
+@pytest.mark.parametrize("http_request", HTTP_REQUESTS)
+def test_decompress_compressed_header_offline(port, http_request):
+    request = http_request(method="GET", url="http://localhost:{}/streams/decompress".format(port))
+    with RequestsTransport() as sender:
+        response = sender.send(request, stream=True)
+        response.raise_for_status()
+        data = response.stream_download(sender, decompress=False)
```
kristapratico, line 161: For this test scenario, we have a compressed file and we want to pass `decompress=True` when stream downloading the response. This will signal to decompress the body based on the `Content-Encoding` header and then we'll successfully download the file contents (i.e. "test")

sdk/core/azure-core/tests/testserver_tests/coretestserver/coretestserver/files/test_with_header.tar.gz
```diff
@@ -0,0 +1 @@
+test
```
kristapratico, line 1: We can use the same file that's added in this PR: https://github.com/Azure/azure-sdk-for-python/pull/22030/files

sdk/core/azure-core/tests/testserver_tests/coretestserver/coretestserver/test_routes/streams.py
```diff
@@ -40,7 +47,16 @@ def iterable():
...
+
 @streams_api.route('/compressed', methods=['GET'])
 def compressed():
     return Response(stream_compressed_header_error(), status=300, headers={"Content-Encoding": "gzip"})
-    
\ No newline at end of file
+
+@streams_api.route('/decompress', methods=['GET'])
+def decompress():
+    return Response(stream_decompress(), status=200)
```
kristapratico, line 62: looks good! We just need to pass the `Content-Encoding` header like on L58!

sdk/core/azure-core/tests/testserver_tests/coretestserver/coretestserver/test_routes/streams.py
```diff
@@ -28,6 +30,16 @@ def stream_json_error():
...
+def stream_decompress_header():
+    try:
+        file_path = os.path.abspath(os.path.join(os.path.abspath(__file__), "..", "./files/test.tar.gz"))
+        with open(file_path, "rb") as fd:
+            yield fd.read()
+    except:
+        file_path = os.path.abspath(os.path.join(os.path.abspath(__file__), "..", "..", "./files/test.tar.gz"))
+        with open(file_path, "rb") as fd:
+            yield fd.read()
+
```
kristapratico, line 42: @zhzhh97 I think you're hitting the same error in your PR as we see in Marcus's PR (CI isn't running on this PR because it looks like there is a merge conflict). See this comment for a possible resolution: https://github.com/Azure/azure-sdk-for-python/pull/22624#discussion_r812512287

sdk/core/azure-core/tests/testserver_tests/coretestserver/coretestserver/test_routes/streams.py
```diff
@@ -40,7 +55,17 @@ def iterable():
 def error():
     return Response(stream_json_error(), status=400)
 
+
+@streams_api.route('/string', methods=['GET'])
+def string():
+    return Response(
+        streaming_test(), status=200, mimetype="text/plain"
+    )
```
kristapratico, line 63: It doesn't look like this code gets used in this PR - seems we can safely remove it.

kristapratico (commented):
Looking good! You can remove the test.tar.gz file from the PR now that you create it dynamically.

I suspect once the merge conflicts are fixed that this PR should pass all tests 🚀

sdk/core/azure-core/tests/async_tests/test_streaming_async.py
```diff
@@ -154,6 +154,7 @@ async def test_compress_plain_header(http_request):
         decoded = content.decode('utf-8')
         assert decoded == "test"
 
+@pytest.mark.live_test_only
```
kristapratico, line 157: would you mind adding the same test in the async file since we skip the online one here?

sdk/core/azure-core/tests/test_streaming.py
```diff
@@ -165,6 +166,22 @@ def test_decompress_compressed_header(http_request):
...
+    request = http_request(method="GET", url="http://localhost:{}/streams/decompress_header".format(port))
+    with RequestsTransport() as sender:
+        try:
+            response = client._pipeline.run(request, stream=True).http_response
+            response.raise_for_status()
+            data = response.stream_download(sender, decompress=True)
+            content = b"".join(list(data))
+            decoded = content.decode('utf-8')
+            assert decoded == "test"
+        except HttpResponseError as e:
```
kristapratico, line 182: we can remove the try/except debugging code now that we fixed the issue with the gzip file.

sdk/core/azure-core/tests/testserver_tests/coretestserver/coretestserver/test_routes/streams.py
```diff
@@ -43,13 +55,10 @@ def iterable():
 def error():
     return Response(stream_json_error(), status=400)
 
-@streams_api.route('/string', methods=['GET'])
-def string():
-    return Response(
-        streaming_test(), status=200, mimetype="text/plain"
-    )
-
```
kristapratico, line 51: I think L46-L51 might have been accidentally removed - the tests are failing because a few of them depend on this code.

sdk/core/azure-core/tests/async_tests/test_streaming_async.py
kristapratico: we'll do the async test later

sdk/core/azure-core/tests/test_streaming.py
kristapratico: @zhzhh97 sorry if this was unclear -- I meant the try/except from the test you are adding here. It looks like the try/except from a different test might have accidentally been removed at the end of this file.

sdk/core/azure-core/tests/test_utils.py
```diff
@@ -1,42 +0,0 @@
-# ------------------------------------
```
kristapratico, line 1: @zhzhh97 I think this file might have accidentally gotten deleted when you rebased? Can we add it back?

kristapratico, line 1: @zhzhh97 I think this file might have accidentally gotten deleted when you rebased? Can we add it back?

kristapratico:
hey @zhzhh97 I pushed a few commits to your PR to see if we can get this merged. :)

iscai-msft (approved):

sdk/core/azure-core/tests/testserver_tests/coretestserver/coretestserver/test_routes/streams.py
```diff
@@ -67,4 +66,16 @@ def compressed_no_header():
 @streams_api.route('/compressed', methods=['GET'])
 def compressed():
     return Response(stream_compressed_header_error(), status=300, headers={"Content-Encoding": "gzip"})
+
+def stream_decompress_header():
+    with gzip.open('test.tar.gz', 'wb') as f:
```
johanste, line 71: Suggestion: use the [tempfile](https://docs.python.org/3/library/tempfile.html) module to help with cleanup etc.

sdk/core/azure-core/tests/testserver_tests/coretestserver/coretestserver/test_routes/streams.py
```diff
@@ -67,4 +66,16 @@ def compressed_no_header():
...
+        f.write(b"test")
+
+    with open(os.path.join(os.path.abspath('test.tar.gz')), "rb") as fd:
+        yield fd.read()
```
johanste, line 75: I would suggest `yield`ing the contents at least twice (and possibly more). The idea being that we make the receiver have to call `read` multiple times. Which will verify that decompression works as expected even when presented with boundaries that may not align with blocks.

sdk/core/azure-core/tests/testserver_tests/coretestserver/coretestserver/test_routes/streams.py
xiangyan99: In fact, I am thinking of checking test.tar.gz into our tests so we don't need to generate it every time.

How do you think?

johanste: Not writing files is goodness. Binary files checked in is less so. If the file is this small, then inlining the data would probably be simpler. We can also use `zlib` compress to compress data inline if we wanted to produce larger amounts of data.

sdk/core/azure-core/tests/testserver_tests/coretestserver/coretestserver/test_routes/streams.py
```diff
@@ -67,4 +67,15 @@ def compressed_no_header():
 @streams_api.route('/compressed', methods=['GET'])
 def compressed():
     return Response(stream_compressed_header_error(), status=300, headers={"Content-Encoding": "gzip"})
+
+def stream_decompress_header():
```
johanste, line 71: Nitpick; I'd consider renaming this because what it really does/represents is a gzip content-encoded stream (iterable) of bytes.

sdk/core/azure-core/tests/test_streaming.py
```diff
@@ -178,6 +179,18 @@ def test_decompress_compressed_header(http_request):
     decoded = content.decode('utf-8')
     assert decoded == "test"
 
+@pytest.mark.parametrize("http_request", HTTP_REQUESTS)
+def test_decompress_compressed_header_offline(port, http_request):
```
johanste, line 183: Do we want to test aiohttp as well?

johanste (approved):

sdk/core/azure-core/tests/test_streaming.py
xiangyan99: Yes. We will add async test in another PR.

Issue #22110, opened 2021-12-09
[textanalytics] add a sample that shows how to set model_version


kristapratico:
Request from the docs team. See https://github.com/Azure/azure-sdk-for-js/pull/19087 for inspiration.

PR #22132, opened 2021-12-10
[textanalytics] add sample showing how to set model_version

kristapratico:
Resolves https://github.com/Azure/azure-sdk-for-python/issues/22110


deyaaeldeen (approved):
Looks good. Do the result objects carry the model version used by any chance? If so, it might be nice to print that too.

kristapratico:
> Looks good. Do the result objects carry the model version used by any chance? If so, it might be nice to print that too.

Thought about showing that, but because Python's design is a bit different, the model_version from the result obj isn't accessed as nicely (through a callback): https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/textanalytics/azure-ai-textanalytics/samples/sample_get_detailed_diagnostics_information.py#L57

mssfang (approved):

PR #22133, opened 2021-12-10
[Blob][Datalake] Dynamically set `read_timeout`

tasherif-msft:


sdk/storage/azure-storage-blob/azure/storage/blob/_download.py
```diff
@@ -16,12 +16,27 @@
...
 from azure.core.tracing.common import with_current_context
+from ._shared.base_client import TransportWrapper
+from ._shared.constants import READ_TIMEOUT
 from ._shared.encryption import decrypt_blob
 from ._shared.request_handlers import validate_and_format_range_headers
 from ._shared.response_handlers import process_storage_error, parse_length_from_content_range
 from ._deserialize import get_page_ranges_result
 
 
+def get_transfer_timeout(client, request_data_size):
```
zezha-msft, line 27: Minor: naming convention -> since the method is not returning anything it should perhaps be called set_transfer_timeout

sdk/storage/azure-storage-blob/azure/storage/blob/aio/_download_async.py
```diff
@@ -14,12 +14,28 @@
...
+from .._shared.base_client_async import AsyncTransportWrapper
+from .._shared.constants import READ_TIMEOUT
 from .._shared.encryption import decrypt_blob
 from .._shared.request_handlers import validate_and_format_range_headers
 from .._shared.response_handlers import process_storage_error, parse_length_from_content_range
 from .._deserialize import get_page_ranges_result
 from .._download import process_range_and_offset, _ChunkDownloader
 
+
+def get_transfer_timeout(client, request_data_size):
```
zezha-msft, line 26: can we avoid code duplication?

tasherif-msft: the issue with avoiding duplication is the imports. the AsyncTransportWrapper causes import issues on python 2 which is still part of our pipelines/testing framework. I couldn't think of a clean way to check if the transporter object was a transport wrapper object (sync or async). So I was left duplicating in the both the async/sync code

selimb:
As I mentioned in https://github.com/Azure/azure-sdk-for-python/pull/24031/files#r852410779, I believe the role of `read_timeout` is misunderstood.
See https://github.com/psf/requests/issues/3099 for a more elaborate discussion on the topic, in which large uploads/downloads come up.

PR #22146, opened 2021-12-13
[Search] Use test-proxy and test-resources.

### LINKED ISSUE #20692: [Search] Test reliability issues with azure-search-documents
Two thorny issues that arose during the most recent Search work:

1. The `SearchServicePreparer` seemingly creates deterministic names instead of random ones. This causes tests to fail when the previous iteration of a test has not fully deleted. Like `StorageAccountPreparer`, the `SearchServicePreparer` should use a completely random name when running live that maps to a deterministic name to facilitate playback.
2. Sporadically, tests fail with a "Connection was force closed". The problem here is that this failure takes several minutes to manifest. The service should ideally _not_ fail, but if it does, it should fail fast.

### LINKED ISSUE #21493: [Search] Update Search to use Test Proxy server
https://github.com/Azure/azure-sdk-for-python/blob/main/doc/dev/test_proxy_migration_guide.md

tjprescott:
Fixes #21493. Fixes #20692.


tjprescott:
Closing in favor of #22187

PR #22147, opened 2021-12-13
[formrecognizer] Update async samples to use asyncio.run()

### LINKED ISSUE #21212: [formrecognizer] Update async samples to use asyncio.run instead of loop

catalinaperalta:
Fixes #21212


kristapratico (approved):
🚢

PR #22157, opened 2021-12-13
Parse SAS connection string correctly for DFS endpoint

tasherif-msft:


seanmcc-msft (commented):
I'm not sure if this is going to work - https://docs.microsoft.com/en-us/rest/api/storageservices/naming-and-referencing-containers--blobs--and-metadata#blob-names

"A blob name can contain any combination of characters."

I think this means someone could name their blob "my.blob.3"

tasherif-msft:
> I'm not sure if this is going to work - https://docs.microsoft.com/en-us/rest/api/storageservices/naming-and-referencing-containers--blobs--and-metadata#blob-names
> 
> "A blob name can contain any combination of characters."
> 
> I think this means someone could name their blob "my.blob.3"

these aren't blob names - its the endpoint

seanmcc-msft (approved):

PR #21717, opened 2021-11-11
Update keyvault to enable live testing in sovereign clouds for multiple services

v-xuto:
These changes enable keyvault to run live tests against Public, UsGov and China.
1. According to JS issue https://github.com/Azure/azure-sdk-for-js/issues/18267, there is the same error here . 
So skip the following tests:
- `test_key_client.py::KeyClientTests::test_key_crud_operations`, 
- `test_keys_async.py::KeyVaultKeyTest::test_key_crud_operations`,
- `test_samples_keys.py::TestExamplesKeyVault::test_example_key_crud_operations`,
- `test_samples_keys_async.py::TestExamplesKeyVault::test_example_key_crud_operations`
2. According to JS issue https://github.com/Azure/azure-sdk-for-js/issues/18142, there is the same error here. 
So skip the following tests:

- `test_key_client.py::KeyClientTests::test_key_rotation`,
- `test_key_client.py::KeyClientTests::test_key_rotation_policy`,
- `test_keys_async.py::KeyVaultKeyTest::test_key_rotation`,
- `test_keys_async.py::KeyVaultKeyTest::test_key_rotation_policy`

3. According to JS issue https://github.com/Azure/azure-sdk-for-js/issues/18414, there is the same error here. 
So add code `MatrixFilters:ArmTemplateParameters=^(?!.*enableHsm.*true)` in UsGov and China cloudconfig in tests.yml.

Pipeline results:
https://dev.azure.com/azure-sdk/internal/_build/results?buildId=1194909&view=results

@benbp , @jameszliao-msft , @lmazuel , @AlexGhiondea , @chlowell for notification.

PR #22179, opened 2021-12-15
[Key Vault] Update HSM deployment scripts

mccoyp:
HSM deployments have been stalling during live test pipelines for the past week. Discussion of the issue can be found in https://github.com/Azure/azure-sdk-for-python/pull/21717.

### DIFF
--- a/sdk/keyvault/test-resources.json
+++ b/sdk/keyvault/test-resources.json
@@ -37,7 +37,7 @@
         },
         "hsmLocation": {
             "type": "string",
-            "defaultValue": "westus",
+            "defaultValue": "canadacentral",
             "allowedValues": [
                 "australiacentral",
                 "canadacentral",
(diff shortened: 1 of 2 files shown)


mccoyp:
@benbp There are some `azure-keyvault-administration` errors that seem to have to do with SAS tokens, but this will unblock the issue with not being able to deploy HSMs

sdk/keyvault/test-resources.json
```diff
@@ -37,7 +37,7 @@
         },
         "hsmLocation": {
             "type": "string",
-            "defaultValue": "westus",
+            "defaultValue": "canadacentral",
```
benbp, line 40: We will still pass in `westus2` as the default in the deployment scripts unless an override is specified in `tests.yml`

benbp (approved):

PR #22237, opened 2021-12-20
[Cosmos] split proof queries sync client

simorenoh:
This is just a draft PR to address the current gap in the cosmos sdk not being split proof

The gone_retry_policy is used exclusively to refresh the partition key range cache and raise the exception immediately in order to repair the document producers and not waste time retrying several times.

One note on these changes is that I still need to remove certain print statements and comments but want to get input first so I can keep logging the right things in case some further iteration is needed. 

I also left the testing file that I was using (`test_xq.py`) in order to walk through the process for triggering a partition split that I was using, but I will remove this file once the changes are approved.


sdk/cosmos/azure-cosmos/azure/cosmos/_execution_context/base_execution_context.py
```diff
@@ -120,7 +119,14 @@ def _fetch_items_helper_no_retries(self, fetch_function):
                 self._has_started = True
             new_options = copy.deepcopy(self._options)
             new_options["continuation"] = self._continuation
-            (fetched_items, response_headers) = fetch_function(new_options)
+            try:
+                (fetched_items, response_headers) = fetch_function(new_options)
+            except exceptions.CosmosHttpResponseError as e:
```
xinlian12, line 124: should we also check sub status code? only raise for split related exception?

xinlian12, line 124: question, if we do not capture and raise the exception specifically, then the exception should automatically bubble up to up layer?

sdk/cosmos/azure-cosmos/azure/cosmos/_execution_context/document_producer.py
```diff
@@ -58,7 +58,13 @@ def __init__(self, partition_key_target_range, client, collection_link, query, d
         collection_id = _base.GetResourceIdOrFullNameFromLink(collection_link)
 
         def fetch_fn(options):
-            return self._client.QueryFeed(path, collection_id, query, options, partition_key_target_range["id"])
+            try:
+                return self._client.QueryFeed(path, collection_id, query, options, partition_key_target_range["id"])
+            except exceptions.CosmosHttpResponseError as e:
+                if e.status_code == http_constants.StatusCodes.GONE:
```
xinlian12, line 64: same question as above

sdk/cosmos/azure-cosmos/azure/cosmos/_execution_context/multi_execution_aggregator.py
```diff
@@ -129,6 +136,37 @@ def fetch_next_block(self):
...
+            try:
+                # TODO: we can also use more_itertools.peekable to be more python friendly
+                targetQueryExContext.peek()
+                # if there are matching results in the target ex range add it to the priority queue
+
+                self._orderByPQ.push(targetQueryExContext)
+
+            except exceptions.CosmosHttpResponseError as e:
+                if document_producer.partition_range_is_gone(e):
+                    # error shouldn't bubble up all the way up here, but we raise to make sure we retry if it does
```
xinlian12, line 164: in case the exception reached here, where we are going to handle the exception?

sdk/cosmos/azure-cosmos/azure/cosmos/_gone_retry_policy.py
```diff
@@ -0,0 +1,56 @@
...
+
+    def __init__(self, client, *args):
+        self._max_retry_attempt_count = 3
+        self.current_retry_attempt_count = 0
+        self.retry_after_in_milliseconds = 1000
+        self.args = args
+        self.client = client
+
+
+    def ShouldRetry(self, exception):
```
xinlian12, line 40: do we need to validate the status code here?

sdk/cosmos/azure-cosmos/azure/cosmos/_gone_retry_policy.py
```diff
@@ -0,0 +1,56 @@
...
+
+# pylint: disable=protected-access
+
+
+class GoneRetryPolicy(object):
+
+    def __init__(self, client, *args):
+        self._max_retry_attempt_count = 3
+        self.current_retry_attempt_count = 0
+        self.retry_after_in_milliseconds = 1000
```
xinlian12, line 35: retry after is not used?

sdk/cosmos/azure-cosmos/azure/cosmos/_execution_context/multi_execution_aggregator.py
simorenoh: it didn't at any point honestly, I can probably take out this logic now that I think about it

sdk/cosmos/azure-cosmos/azure/cosmos/_gone_retry_policy.py
simorenoh: nope because we send the error into here only when we already verified a 410

sdk/cosmos/azure-cosmos/azure/cosmos/_execution_context/base_execution_context.py
simorenoh: seems like you're right, there's no need to capture the exception at this point :)

sdk/cosmos/azure-cosmos/azure/cosmos/_execution_context/document_producer.py
simorenoh: seems like this one is necessary, otherwise the error does not get handled properly and instead gets stuck within the code execution (not thrown but not used)

sdk/cosmos/azure-cosmos/azure/cosmos/_gone_retry_policy.py
simorenoh: we verify the substatus code as a way to future proof - if we need to use the gone policy for something else, we can branch on substatus code

sdk/cosmos/azure-cosmos/azure/cosmos/_gone_retry_policy.py
simorenoh: while we don't currently use it, it is called within the retry utility to call a cosmos client timeout - and will be used if the sub status code is not PARTITION_KEY_RANGE_GONE (1002)

sdk/cosmos/azure-cosmos/azure/cosmos/_gone_retry_policy.py
```diff
@@ -0,0 +1,55 @@
...
+        :param (exceptions.CosmosHttpResponseError instance) exception:
+        :rtype: boolean
+
+        """
+        if exception.sub_status == http_constants.SubStatusCodes.PARTITION_KEY_RANGE_GONE:
+            # refresh routing_map_provider to refresh partition key range cache
+            # return False to raise error to multi_execution_aggregator and repair document producer
+            self.client.refresh_routing_map_provider()
+            return False
+        if self.current_retry_attempt_count < self._max_retry_attempt_count:
```
kushagraThapar, line 52: why are we only checking the current retry attempt count in this if condition, and not when updating the partition key range cache. 
Let's say if we keep getting gone exception, this if condition will never hit, and we will never get out of this retry policy.

sdk/cosmos/azure-cosmos/azure/cosmos/_execution_context/multi_execution_aggregator.py
```diff
@@ -129,6 +136,32 @@ def fetch_next_block(self):
 
         raise NotImplementedError("You should use pipeline's fetch_next_block.")
 
+    def _repair_document_producer(self):
```
kushagraThapar, line 139: Would be good to add some documentation to this function, like what it does and how it repairs the document producers.

sdk/cosmos/azure-cosmos/samples/test_xq.py
```diff
@@ -0,0 +1,82 @@
+import time
```
kushagraThapar, line 1: why have we added this test in samples ?
Also, I see license header is missing for this one and some other samples, we should add it.

kushagraThapar (changes requested):

sdk/cosmos/azure-cosmos/azure/cosmos/_gone_retry_policy.py
simorenoh: I was returning a False immediately for sub status 1002 in order to further handle the exception on a first-try basis, since we need the exception to bubble up in order to repair the execution context.

I think this is the way to go, but I see what you mean. I'll add a flag that only returns False pre-emptively if it's the first time we run into a 1002 with that instance of the gone retry policy.

sdk/cosmos/azure-cosmos/azure/cosmos/_execution_context/document_producer.py
```diff
@@ -99,7 +105,12 @@ def peek(self):
 
         """
         if self._cur_item is None:
-            self._cur_item = next(self._ex_context)
+            try:
+                self._cur_item = next(self._ex_context)
+            except exceptions.CosmosHttpResponseError as e:
```
xinlian12, line 110: what would happen for other exceptions if we only raise gone exception?

sdk/cosmos/azure-cosmos/azure/cosmos/_execution_context/multi_execution_aggregator.py
xinlian12: will this also handle multiple split (split after split? ) scenario?

sdk/cosmos/azure-cosmos/azure/cosmos/_execution_context/document_producer.py
simorenoh: i was wrong ignore me

sdk/cosmos/azure-cosmos/azure/cosmos/_gone_retry_policy.py
```diff
@@ -0,0 +1,59 @@
...
+            self.current_retry_attempt_count += 1
+            if (self.refresh_partition_key_range_cache
+                    and exception.sub_status == http_constants.SubStatusCodes.PARTITION_KEY_RANGE_GONE):
+                # refresh routing_map_provider to refresh partition key range cache
+                # make refresh_partition_key_range_cache False to skip this check on subsequent Gone exceptions
+                # return False to raise error to multi_execution_aggregator and repair document producer context
+                self.client.refresh_routing_map_provider()
+                self.refresh_partition_key_range_cache = False
+                return False
+            return True
```
simplynaveen20, line 58: Why we want to retry after one attempt without refreshing the cache . PartitionKeyRangeGoneRetryPolicy in V4 Java SDK only retrying one time https://github.com/Azure/azure-sdk-for-java/blob/c4c068cc0d2061df9a4fcc49419acc10be223d4a/sdk/cosmos/azure-cosmos/src/main/java/com/azure/cosmos/implementation/PartitionKeyRangeGoneRetryPolicy.java#L60

sdk/cosmos/azure-cosmos/samples/test_xq.py
```diff
@@ -0,0 +1,82 @@
...
+        container.create_item(body=body)
+        if i == 5000 or i == 10000:
+            print("created 5k items")
+    print("15k items took {} seconds".format(time.time() - start))
+    success, errors = 0, 0
+    print("--------------------------------")
+    print("created items, waiting 4m before queries")
+    print("change offer to 11k manually")
+    print("--------------------------------")
+    time.sleep(240)
```
simplynaveen20, line 46: Current test will not guarantee that your split completed , please refer https://github.com/Azure/azure-sdk-for-java/blob/c4c068cc0d2061df9a4fcc49419acc10be223d4a/sdk/cosmos/azure-cosmos/src/test/java/com/azure/cosmos/rx/QueryValidationTests.java#L397

simplynaveen20 (approved):
Please have a look to the comments I added , apart from that LGTM

sdk/cosmos/azure-cosmos/azure/cosmos/_execution_context/multi_execution_aggregator.py
xinlian12: Discussed offline, Ideally, the code should handle multiple splits scenario. But generally split will take some time, the one split right after another one is rare to happen, so will not put too much effort it at this moment.

sdk/cosmos/azure-cosmos/azure/cosmos/_execution_context/document_producer.py
xinlian12: Discussed offline.

sdk/cosmos/azure-cosmos/azure/cosmos/_retry_utility.py
```diff
@@ -32,8 +32,10 @@
 from . import _resource_throttle_retry_policy
 from . import _default_retry_policy
 from . import _session_retry_policy
+from . import _gone_retry_policy
 from .http_constants import HttpHeaders, StatusCodes, SubStatusCodes
 
+
```
xinlian12, line 38: nit: remove extra line

sdk/cosmos/azure-cosmos/azure/cosmos/_retry_utility.py
```diff
@@ -32,8 +32,10 @@
 from . import _resource_throttle_retry_policy
 from . import _default_retry_policy
 from . import _session_retry_policy
+from . import _gone_retry_policy
```
xinlian12, line 35: question: usually the import statements should be sorted in java & SDK, do we need to sort in Python as well?

sdk/cosmos/azure-cosmos/test/test_config.py
```diff
@@ -82,6 +82,16 @@ def create_database_if_not_exist(cls, client):
         cls.IS_MULTIMASTER_ENABLED = client.get_database_account()._EnableMultipleWritableLocations
         return cls.TEST_DATABASE
 
+    @classmethod
+    def create_database_if_not_exist_with_throughput(cls, client, throughput):
+        # type: (CosmosClient) -> Database
+        if cls.TEST_DATABASE is not None:
+            return cls.TEST_DATABASE
+        cls.try_delete_database(client)
```
xinlian12, line 90: why need to delete the database if it does not exist?

xinlian12 (approved):
LGTM, thanks for the fix 👍

sdk/cosmos/azure-cosmos/azure/cosmos/_retry_utility.py
simorenoh: yes but this is something that I need to do probably in every file, will take care of all of it in one PR later

kushagraThapar (approved):

PR #22262, opened 2021-12-27
[Cosmos] deprecate offer-named methods in favor of new throughput-named methods

simorenoh:
With these changes, we can officially mark read_offer() for deprecation and instantiate the more relevantly named read_throughput() method for database and container levels.

I went ahead and created a new ThroughputProperties object to replace the previously existing Offer object as well to match with the .NET and Java SDKs, although this bit is completely extra and could be reverted if we feel it's not neccessary.

I also know that the reasoning for this was the current state of documentation (which mostly references throughput as opposed to offers) but I found this too: https://docs.microsoft.com/en-us/rest/api/cosmos-db/offers - not sure if this is all there is, but wanted to double check on this point.


sdk/cosmos/azure-cosmos/azure/cosmos/aio/database.py
```diff
@@ -144,15 +145,15 @@ async def read(self, **kwargs):
...
-        **kwargs  # type: Any
+            self,
+            id,  # type: str  # pylint: disable=redefined-builtin
+            partition_key,  # type: Any
+            indexing_policy=None,  # type: Optional[Dict[str, Any]]
+            default_ttl=None,  # type: Optional[int]
+            offer_throughput=None,  # type: Optional[int]
+            unique_key_policy=None,  # type: Optional[Dict[str, Any]]
+            conflict_resolution_policy=None,  # type: Optional[Dict[str, Any]]
+            **kwargs  # type: Any
```
tjprescott, line 156: The original indentation here looks correct. What's the reason for this change?

tjprescott (changes requested):
Seems a tooling mismatch is causing non-standard indents of many methods. Also, please add a changelog entry documenting these changes. This looks reasonable to me, but I'll defer to @annatisch on the deprecation specifics.

sdk/cosmos/azure-cosmos/azure/cosmos/aio/database.py
simorenoh: Pycharm did this on its own to "format the file correctly"

tjprescott:
It seems the non-standard indenting is still showing in the diff.

tjprescott (changes requested):
So we still have a breaking change on account of wholesale renaming `Offer` to `ThroughputProperties`. @annatisch how do you recommend resolving that? Alias Offer to ThroughputProperties?

sdk/cosmos/azure-cosmos/azure/cosmos/throughput_properties.py
```diff
@@ -1,5 +1,5 @@
 # The MIT License (MIT)
```
annatisch, line 1: We should just leave this file in place.

annatisch (commented):
Yeah as @tjprescott mentioned - anyone who's written the following code should remain unbroken:
```python
from azure.cosmos.offer import Offer
o = client.read_offer()
if isinstance(o, Offer):
    print(o.properties)
```
I'm pretty sure for the model itself this could just be aliased. However you'll probably need to keep the `offer.py` file around.

We're fortunate that this is a response type only (I can't see anywhere a customer would have constructed this themselves), so hopefully the number of people using this directly should be low.

sdk/cosmos/azure-cosmos/azure/cosmos/aio/container.py
```diff
@@ -32,7 +33,7 @@
 from .._base import build_options as _build_options, validate_cache_staleness_value
 from ..exceptions import CosmosResourceNotFoundError
 from ..http_constants import StatusCodes
-from ..offer import Offer
+from ..offer import ThroughputProperties
```
tjprescott, line 36: I think we still need to import `Offer`, we just don't need to add it to `__all__`

sdk/cosmos/azure-cosmos/azure/cosmos/aio/container.py
```diff
@@ -547,16 +548,35 @@ async def delete_item(
 
     @distributed_trace_async
     async def read_offer(self, **kwargs):
```
annatisch, line 550: Can we just remove these entirely from async?
Seems odd to deprecate an API that hasn't even GA'ed...
Let's just use the deprecation for sync, and async only has the new APIs.

sdk/cosmos/azure-cosmos/azure/cosmos/offer.py
```diff
@@ -19,18 +19,21 @@
...
+    To read and update throughput properties use the associated methods on the :class:`Container`.
     """
 
     def __init__(self, offer_throughput, properties=None):  # pylint: disable=super-init-not-called
         # type: (int, Dict[str, Any]) -> None
         self.offer_throughput = offer_throughput
         self.properties = properties
+
+
+Offer = ThroughputProperties
```
annatisch, line 39: We didn't add any tests for this?

sdk/cosmos/azure-cosmos/azure/cosmos/aio/container.py
simorenoh: completely agreed, removed!

sdk/cosmos/azure-cosmos/azure/cosmos/aio/container.py
simorenoh: why do we need to import Offer again? It's not being used anywhere anymore with the replacement of ThroughputProperties

sdk/cosmos/azure-cosmos/azure/cosmos/offer.py
simorenoh: just added them

annatisch (approved):
Looks good - probably just needs a changelog entry?

PR #22263, opened 2021-12-27
[Cosmos] create new async user agent for async client

simorenoh:
Draft PR to address one of the points in the list of things to take care of while async is in preview

With these changes we should be able to see the difference between customers using the sync/async clients in our logs, making it easier to pinpoint if an issue is coming from one or the other.


kushagraThapar (approved):
LGTM, thanks @simorenoh  - would be good to add some unit tests for these 2 user agents.

kushagraThapar (approved):

sdk/cosmos/azure-cosmos/test/test_client_user_agent.py
```diff
@@ -0,0 +1,53 @@
...
+
+    async def test_client_user_agent(self):
+        client_sync = sync_client.CosmosClient(url=_test_config.host, credential=_test_config.masterKey)
+        client_async = async_client.CosmosClient(url=_test_config.host, credential=_test_config.masterKey)
+
+        self.assertTrue(client_sync.client_connection._user_agent.startswith("azsdk-python-cosmos/"))
+        self.assertTrue(client_async.client_connection._user_agent.startswith("azsdk-python-cosmos-async/"))
+        self.assertTrue(client_async.client_connection._user_agent != client_sync.client_connection._user_agent)
+
+        await client_async.close()
```
kushagraThapar, line 48: Do we also need to close `client_sync` ?

simorenoh: nope, since the other client is synchronous it closes gracefully upon exiting the execution context is it being used in

kushagraThapar: Sounds good, thanks!

PR #22264, opened 2021-12-27
[DRAFT] [Cosmos] mark populate_query_metrics flag to be deleted when used and remove docs

simorenoh:
Draft PR to address one of the points in the list of things to take care of while async is in preview

Removed the docs around the usage of the `populate_query_metrics` flag in methods that don't need it, and added a user warning in case it is being passed in.

### DIFF
--- a/sdk/cosmos/azure-cosmos/azure/cosmos/container.py
+++ b/sdk/cosmos/azure-cosmos/azure/cosmos/container.py
@@ -25,6 +25,7 @@
 from typing import Any, Dict, List, Optional, Union, Iterable, cast  # pylint: disable=unused-import
 
 import six
+import warnings
 from azure.core.tracing.decorator import distributed_trace  # type: ignore
 
 from ._cosmos_client_connection import CosmosClientConnection
@@ -120,7 +121,6 @@ class ContainerProxy(object):
         # type: (...) -> Dict[str, Any]
         """Read the container properties.
 
-        :param populate_query_metrics: Enable returning query metrics in response headers.
         :param populate_partition_key_range_statistics: Enable returning partition key
             range statistics in response headers.
         :param populate_quota_info: Enable returning collection storage quota information in response headers.
@@ -135,6 +135,10 @@ class ContainerProxy(object):
         request_options = build_options(kwargs)
         response_hook = kwargs.pop('response_hook', None)
         if populate_query_metrics is not None:
+            warnings.warn(
+                "the populate_query_metrics flag does not apply to this method and will be removed in the future",
+                UserWarning,
+            )
             request_options["populateQueryMetrics"] = populate_query_metrics
         if populate_partition_key_range_statistics is not None:
             request_options["populatePartitionKeyRangeStatistics"] = populate_partition_key_range_statistics
@@ -165,7 +169,6 @@ class ContainerProxy(object):
 
         :param item: The ID (name) or dict representing item to retrieve.
         :param partition_key: Partition key for the item to retrieve.
-        :param populate_query_metrics: Enable returning query metrics in response headers.
         :param post_trigger_include: trigger id to be used as post operation trigger.
         :keyword str session_token: Token for use with Session consistency.
         :keyword dict[str,str] initial_headers: Initial headers to be sent as part of the request.
@@ -191,6 +194,10 @@ class ContainerProxy(object):
         if partition_key is not None:
             request_options["partitionKey"] = self._set_partition_key(partition_key)
         if populate_query_metrics is not None:
+            warnings.warn(
+                "the populate_query_metrics flag does not apply to this method and will be removed in the future",
+                UserWarning,
+            )
             request_options["populateQueryMetrics"] = populate_query_metrics
         if post_trigger_include is not None:
             request_options["postTriggerInclude"] = post_trigger_include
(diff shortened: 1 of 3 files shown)


sdk/cosmos/azure-cosmos/azure/cosmos/container.py
```diff
@@ -119,7 +121,6 @@ def read(
         # type: (...) -> Dict[str, Any]
         """Read the container properties.
 
-        :param populate_query_metrics: Enable returning query metrics in response headers.
```
annatisch, line 122: If you're keen.... we could go further than this and unbreakingly remove it from the method signature.....
Something like:
```python
    @overload
    def read(
        self,
        *,
        populate_partition_key_range_statistics: Optional[bool] = None,
        populate_quota_info: Optional[bool] = None,
        **kwargs
    )

    def read(
        self,
        *args,
        **kwargs
    ):
        populate_query_metrics = args[0] if args else None or kwargs.pop('populate_query_metrics', None)
        if populate_query_metrics:
            ...  # warning
```

simorenoh: I don't think we need to go that far since regardless the option can still be passed in and never did anything to begin with. This is a good point of reference for future refactoring though, thank you! :)

annatisch: That's true - the reason I raised it is that currently the API reference documentation will display a positional keyword parameter that has no accompanying documentation. Which may confuse.
What we _could_ do - is leave the param documentation in, but just describe it as a deprecated no-op.

But yes - adding the fancier overloads are a non-breaking refactor we could do later before the GA of the async APIs to clean up the old sync APIs to match any 'improvements' we want to add in the async version.

simorenoh: I see what you mean... how would we accomplish the work for this item then though? From our initial conversation, we had concluded on deleting the documentation for this option and marking it as deprecated if users passed it in. Does this mean we now want to still provide this documentation (even though the option is useless and marked to be removed)?

sdk/cosmos/azure-cosmos/azure/cosmos/container.py
```diff
@@ -164,7 +181,6 @@ def read_item(
 
         :param item: The ID (name) or dict representing item to retrieve.
         :param partition_key: Partition key for the item to retrieve.
-        :param populate_query_metrics: Enable returning query metrics in response headers.
```
annatisch, line 167: Do we want to open an issue to track the full `overload` treatment for the rest of these methods?

sdk/cosmos/azure-cosmos/azure/cosmos/container.py
```diff
@@ -133,8 +143,15 @@ def read(
         """
         request_options = build_options(kwargs)
         response_hook = kwargs.pop('response_hook', None)
-        if populate_query_metrics is not None:
+        populate_query_metrics = args[0] if args else None or kwargs.pop('populate_query_metrics', None)
```
annatisch, line 146: I would also recommend adding some tests here to validate the non-breaking behaviour for both positional and keyword parameters.

sdk/cosmos/azure-cosmos/azure/cosmos/container.py
```diff
@@ -133,8 +143,15 @@ def read(
...
+        if populate_query_metrics:
+            warnings.warn(
+                "the populate_query_metrics flag does not apply to this method and will be removed in the future",
+                UserWarning,
+            )
             request_options["populateQueryMetrics"] = populate_query_metrics
+        populate_partition_key_range_statistics = kwargs.pop("populate_partition_key_range_statistics", None)
+        populate_quota_info = kwargs.pop("populate_quota_info", None)
```
annatisch, line 154: These two parameters also need to be checked for in the positional args list:
```python
populate_partition_key_range_statistics = args[1] if args else None or kwargs.pop("populate_partition_key_range_statistics", None)
populate_quota_info = args[2] if args else None or kwargs.pop("populate_quota_info", None)
```

sdk/cosmos/azure-cosmos/azure/cosmos/container.py
simorenoh: what do you mean by this?

sdk/cosmos/azure-cosmos/azure/cosmos/container.py
```diff
@@ -133,8 +143,15 @@ def read(
...
-            request_options["populateQueryMetrics"] = populate_query_metrics
+        populate_query_metrics = args[0] if args else kwargs.pop('populate_query_metrics', None)
+        if populate_query_metrics:
+            warnings.warn(
+                "the populate_query_metrics flag does not apply to this method and will be removed in the future",
+                UserWarning,
+            )
+        populate_partition_key_range_statistics = args[1] if args else kwargs.pop(
+            "populate_partition_key_range_statistics", None)
+        populate_quota_info = args[2] if args else kwargs.pop("populate_quota_info", None)
```
annatisch, line 154: This could potentially raise an IndexError - to be safe I think you'd need:
`... = args[2] if len(args) > 2 else ...`
(and similar for line 152)

sdk/cosmos/azure-cosmos/azure/cosmos/container.py
annatisch: Just that it would be nice to add the overloads everywhere - but it's too much work to do right now, and non-breaking to add later. So we should open an issue to track it for post-GA.

annatisch (approved):

sdk/cosmos/azure-cosmos/azure/cosmos/container.py
simorenoh: Follow-up issue to this: https://github.com/Azure/azure-sdk-for-python/issues/24423

PR #22284, opened 2021-12-30
[Cosmos] Remove GetAuthorizationHeader method from client

simorenoh:
PR to address one of the points in the list of things to take care of while async is in preview

As part of this effort, I have made the auth module private (since it's only meant to be used internally regardless) and have removed samples/ tests that used this method along with the REST API rather than with the SDK.


sdk/cosmos/azure-cosmos/azure/cosmos/auth.py
```diff
@@ -27,12 +27,23 @@
 import hmac
 
 import six
+import warnings
 
 from . import http_constants
 
 
 def GetAuthorizationHeader(
```
simorenoh, line 35: I didn't directly remove the method due to the usages in tests and samples since regardless this would give the user insights into it - waiting for confirmation on plan of action.

tjprescott (commented):
Will defer to @annatisch on if this correctly implements the Python SDK deprecation requirements, but there should definitely be a changelog entry for this.

sdk/cosmos/azure-cosmos/azure/cosmos/auth.py
```diff
@@ -27,12 +27,23 @@
...
-    cosmos_client_connection, verb, path, resource_id_or_fullname, is_name_based, resource_type, headers
+        cosmos_client_connection, verb, path, resource_id_or_fullname, is_name_based, resource_type, headers
+):
+    warnings.warn("This method is being removed from the SDK, please use recommended authorization patterns by passing "
+                  "in credentials directly within the creation of the cosmos client instance", UserWarning)
```
annatisch, line 39: I think we should use the `DeprecationWarning` flag here rather than `UserWarning`.
I would also phrase the message accordingly - something like:
"This method has been deprecated and will be removed in a future release."

sdk/cosmos/azure-cosmos/azure/cosmos/auth.py
```diff
@@ -43,7 +54,7 @@ def GetAuthorizationHeader(
     :param str resource_type:
     :param dict headers:
     :return: The authorization headers.
-    :rtype: dict
+    :rtype: str
```
annatisch, line 57: Did we actually change the return type? Or is this just mis-documented?

sdk/cosmos/azure-cosmos/samples/nonpartitioned_container_operations.py
```diff
@@ -109,6 +109,7 @@ def create_nonpartitioned_container(db):
 
 
 def get_authorization(client, verb, resource_id_or_fullname, resource_type, headers):
+    # Here the method is being called
```
annatisch, line 112: This entire file looks like it should be... reconsidered...
There looks like too much manual url manipulation, direct request/response handling...
There must be a way to achieve this using standard API's...?

simorenoh: Completely agreed, so much of this isn't even using the SDK directly... @kushagraThapar would love your input when you get a moment

sdk/cosmos/azure-cosmos/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 #### Bugs Fixed
 
 #### Other Changes
+- Made `auth.py` a private file, making the GetAuthorizationMethod no longer public
```
tjprescott, line 12: This immediately strikes me as a breaking change--removing public surface area.

simorenoh: I was torn between that and this since we didn't export the auth file or methods anywhere - but if it's a breaking change regardless can I just remove the authorization method rather than deprecating it?

annatisch: We can't break it... we can really only throw deprecation warnings around...
There might be ways to mess with dunder get_attr in the init file to support backwards compatibility... but we haven't experimented with this to know what it's limitations/drawbacks would be.

simorenoh: ill make the file public again and just stick to the current deprecation message in that case

sdk/cosmos/azure-cosmos/samples/nonpartitioned_container_operations.py
simorenoh: spoke with Kushagra and ended up removing a lot of the code - should be good now

PR #22302, opened 2022-01-03
[core] add error body to HttpResponseError str

### LINKED ISSUE #21760: Exception messages in Python not showing nicely

iscai-msft:
fixes #21760


sdk/core/azure-core/azure/core/exceptions.py
```diff
@@ -333,6 +333,16 @@ def _parse_odata_body(error_format, response):
             pass
         return None
 
+    def __str__(self):
+        retval = super(HttpResponseError, self).__str__()
+        try:
+            body = self.response.text()
+            if body and not self.error:
```
annatisch, line 340: If `self.error` evaluates to true, does that mean it's already included in `retval`?

sdk/identity/azure-identity/tests/test_aad_client.py
```diff
@@ -44,7 +44,7 @@ def test_error_reporting():
         assert transport.send.call_count == 1
         transport.send.reset_mock()
 
-
+@pytest.mark.skip(reason="Adding body to HttpResponseError str. Not an issue bc we don't automatically log errors")
```
annatisch, line 47: Is this anything in these tests that should be preserved outside of thigs new str behaviour?

annatisch (approved):

sdk/identity/azure-identity/tests/test_aad_client.py
xiangyan99: Second. Seems like we changed our expected behavior. Maybe we should remove the tests?

iscai-msft: I checked with @mccoyp and @johanste and they're ok with skipping the tests (at least for now). The test basically mocks a response with secrets, and wants to make sure we don't expose the mock secrets in the str of error response.

@xiangyan99 maybe we can delete, but I think for now we can just skip, just in case

sdk/core/azure-core/azure/core/exceptions.py
iscai-msft: Yep, if `self.error` exists, it means the error body was correctly parsed with odatav4 (code [here](https://github.com/Azure/azure-sdk-for-python/blob/5c006b5b53a6bfade999b8f0a90ab4c7a9fff8e1/sdk/core/azure-core/azure/core/exceptions.py#L310)). In that case, the relevant error body details are already parsed out and included in the str error response

xiangyan99 (approved):
LGTM

Issue #10979, opened 2020-04-21
[ServiceBus] Add fixed retry support


yunhaoling:
Add support for RetryMode.Fixed

https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/core/azure-core/azure/core/pipeline/policies/_retry.py#L83

PR #21884, opened 2021-11-24
[EventHubs] add fixed backoff retry mode

swathipil:
fixes: #8670

PR #22317, opened 2022-01-04
[ServiceBus] fixed retry backoff

swathipil:
fixes: #10979

based on changes to #21884


yunhaoling (approved):
LGTM! also we're missing changelog updates and version bump

Issue #11685, opened 2020-05-28
[ServiceBus] Expand **kwargs in both sync and async APIs to improve intellisense


yunhaoling:
In current public async API, kwargs are not expanded which means only required parameters shows up when IDE/Editor try to do intellisense stuff -- auto-code-complete and hint.

To make intellisense work better, expand `(param, **kwargs)` to `(param1, *, kwarg1=default, kwarg2=default, ...)` in async APIs.

not applicable for sync due to the compatibility with py2.7

richardpark-msft:
Short of being able to remove Python 2.7 perhaps we can introduce some newlines or something into the text between the individual keyword args? 

That would at least help with readability - it's a huge block of fairly undelimited text otherwise.

KieranBrantnerMagee:
Bumping milestone as we're addressing this incrementally (redoing kwargs/typing as we touch components of the code) to continue tracking this work.

PR #22353, opened 2022-01-06
[ServiceBus] expand kwargs in public API

swathipil:
fixes: #11685


sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_client.py
```diff
@@ -225,8 +278,20 @@ def get_queue_sender(self, queue_name, **kwargs):
         self._handlers.add(handler)
         return handler
 
-    def get_queue_receiver(self, queue_name, **kwargs):
-        # type: (str, Any) -> ServiceBusReceiver
+    def get_queue_receiver(
+        self,
+        queue_name: str,
+        *,
+        session_id: Union[str, NEXT_AVAILABLE_SESSION] = None,
```
swathipil, line 285: why is NEXT_AVAILABLE_SESSION available in the public API, rather than `ServiceBusSessionFilter`. I know it's the only option for now, but it seems strange.

If we only want that constant to be available, rather than any other future values added to `ServiceBusSessionFilter`, should the type be `Union[str, int]`? And, since None is an option, shouldn't the type here be `Optional[Union[str, int]]`?

sdk/servicebus/azure-servicebus/azure/servicebus/amqp/_amqp_message.py
```diff
@@ -123,7 +136,7 @@ def __init__(self, **kwargs):
             return
 
         # manually constructed AMQPAnnotatedMessage
-        input_count_validation = len([key for key in ("data_body", "sequence_body", "value_body") if key in kwargs])
+        input_count_validation = len([key for key in (data_body, sequence_body, value_body) if key])
```
swathipil, line 139: These cannot be included as explicit kwargs, b/c in the case that the user passes `value_body=None`, this will detect count as 0. @yunhaoling - do you know if there's a way to check that the user passed in `value_body=None`? or if no bodies are passed in, can we set `value_body` to `None` by default? and then validate that `input_count_validation <=1`?

sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_client.py
yunhaoling: you're right... it's a bit strange here as `NEXT_AVAILABLE_SESSION` is a constant value not a class. Also, ServiceBusSessionFilter is not an extensible Enum so it's not of type int, Union[str, int] wouldn't work unless we make it an extensible Enum -- `class ServiceBusSessionFilter(int, Enum)`.

I feel that the right approach is to expose `ServiceBusSessionFilter`, however, I feel it's overkilling to expose an Enum with just one member.

 how about let's keep it for now unless someone explains?

swathipil: I'm def with you on that. Unfortunately, this Enum is not a valid type, so I get a mypy error. Maybe I can just try to ignore the type? Not sure if that would work, but is that preferable?

sdk/servicebus/azure-servicebus/azure/servicebus/amqp/_amqp_message.py
yunhaoling: interesting.. I see what you mean that we cannot detect whether it's a None passed by user or default None.

if we set value_body to None by default, from a user perspective, previously they have to explicitly pass something, while now they can create an empty message with value_body being None.

bascially:

`msg = AmqpAnnotatedMessage(value_body=None)`
vs
`msg = AmqpAnnotatedMessage()`

and it shouldn't be a breaking changing to users, right? I feel I'm onboard with this idea but could you double check that this won't be a breaking change

swathipil: Here are the cases in which there would be a difference:
1) 
```python
AmqpAnnotatedMessage(...) # no bodies passed in
```
- Before: results `ValueError` b/c less than 1 body passed in.
- Now: created Message with value_body=None, without ValueError

2)
```python
AmqpAnnotatedMessage(data_body=..., value_body=...) # two bodies passed in, one is value_body
AmqpAnnotatedMessage(sequence_body=..., value_body=...) # two bodies passed in, one is value_body
```
- Before: results in ValueError since >1 bodies passed in
- Now: No ValueError. Message gets created with sequence_body

So the only difference is that what previously resulted in ValueError no longer does in these cases. It still does if multiple bodies are passed in and none of them are set to `None`.

sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_client.py
```diff
@@ -74,15 +86,40 @@ class ServiceBusClient(object):
...
+        self,
+        fully_qualified_namespace: str,
+        credential: Union[
+            "TokenCredential", "AzureSasCredential", "AzureNamedKeyCredential"
+        ],
+        *,
+        http_proxy: Optional[Dict[str, Any]] = None,
+        user_agent: Optional[str] = None,
+        logging_enable: Optional[bool] = False,
+        transport_type: Optional[TransportType] = TransportType.Amqp,
```
yunhaoling, line 99: This should be `azure.servicebus.TransportType.Amqp`, not a uamqp one to avoid uamqp type leaking.

sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_client.py
```diff
@@ -2,11 +2,12 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License. See License.txt in the project root for license information.
 # --------------------------------------------------------------------------------------------
-from typing import Any, Union, TYPE_CHECKING
+from typing import Any, Dict, Union, Optional, TYPE_CHECKING
 import logging
 from weakref import WeakSet
 
 import uamqp
+from uamqp.constants import TransportType
```
yunhaoling, line 10: I think this line is correlated with "transport_type: Optional[TransportType] = TransportType.Amqp"

however, we should not leak uamqp types, we should use `~azure.servicebus.TransportType`

sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_client.py
```diff
@@ -225,15 +282,27 @@ def get_queue_sender(self, queue_name, **kwargs):
         self._handlers.add(handler)
         return handler
 
-    def get_queue_receiver(self, queue_name, **kwargs):
-        # type: (str, Any) -> ServiceBusReceiver
+    def get_queue_receiver(
+        self,
+        queue_name: str,
+        *,
+        session_id: Optional[Union[str, ServiceBusSessionFilter]] = None,
```
yunhaoling, line 289: `ServiceBusSessionFilter` is not part of the public API surface

what mypy error are you getting with this: session_id: Union[str, NEXT_AVAILABLE_SESSION] = None? is there a link to the pipeline log?

sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_session.py
```diff
@@ -157,8 +153,7 @@ def set_state(self, state, **kwargs):
             timeout=timeout,
         )
 
-    def renew_lock(self, **kwargs):
-        # type: (Any) -> datetime.datetime
+    def renew_lock(self, timeout: Optional[float] = None) -> datetime.datetime:
```
yunhaoling, line 156:
```suggestion
    def renew_lock(self, *, timeout: Optional[float] = None) -> datetime.datetime:
```

sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_session.py
```diff
@@ -121,8 +119,7 @@ def get_state(self, **kwargs):
         session_state = response.get(MGMT_RESPONSE_SESSION_STATE)  # type: ignore
         return session_state
 
-    def set_state(self, state, **kwargs):
-        # type: (Union[str, bytes, bytearray], Any) -> None
+    def set_state(self, state: Union[str, bytes, bytearray], timeout: Optional[float] = None) -> None:
```
yunhaoling, line 122:
```suggestion
    def set_state(self, state: Union[str, bytes, bytearray], *, timeout: Optional[float] = None) -> None:
```

sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_session.py
```diff
@@ -88,8 +88,7 @@ class ServiceBusSession(BaseSession):
             :caption: Get session from a receiver
     """
 
-    def get_state(self, **kwargs):
-        # type: (Any) -> bytes
+    def get_state(self, timeout: Optional[float] = None) -> bytes:
```
yunhaoling, line 91:
```suggestion
    def get_state(self, *, timeout: Optional[float] = None) -> bytes:
```

sdk/servicebus/azure-servicebus/azure/servicebus/_common/_connection_string_parser.py
```diff
@@ -12,13 +13,21 @@ class ServiceBusConnectionStringProperties(DictMixin):
...
-        self._shared_access_key_name = kwargs.pop("shared_access_key_name", None)
-        self._shared_access_key = kwargs.pop("shared_access_key", None)
+    def __init__(
+        self,
+        fully_qualified_namespace: Optional[str] = None,
+        endpoint: Optional[str] = None,
+        entity_path: Optional[str] = None,
+        shared_access_signature: Optional[str] = None,
+        shared_access_key_name: Optional[str] = None,
+        shared_access_key: Optional[str] = None,
```
yunhaoling, line 23:
```suggestion
    def __init__(
        self,
        *,
        fully_qualified_namespace: Optional[str] = None,
        endpoint: Optional[str] = None,
        entity_path: Optional[str] = None,
        shared_access_signature: Optional[str] = None,
        shared_access_key_name: Optional[str] = None,
        shared_access_key: Optional[str] = None,
```

sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_client.py
swathipil: https://dev.azure.com/azure-sdk/public/_build/results?buildId=1277503&view=logs&j=b70e5e73-bbb6-5567-0939-8415943fadb9&t=e3f8f308-6e5f-58aa-3a2a-fe425958de32&l=70

sdk/servicebus/azure-servicebus/azure/servicebus/aio/_servicebus_client_async.py
```diff
@@ -2,11 +2,12 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License. See License.txt in the project root for license information.
 # --------------------------------------------------------------------------------------------
-from typing import Any, Union, TYPE_CHECKING
+from typing import Any, Dict, Union, Optional, TYPE_CHECKING
 import logging
 from weakref import WeakSet
 
 import uamqp
+from uamqp.constants import TransportType
```
yunhaoling, line 10: same transport concern as the sync one

sdk/servicebus/azure-servicebus/azure/servicebus/aio/_servicebus_client_async.py
```diff
@@ -222,7 +271,20 @@ def get_queue_sender(self, queue_name: str, **kwargs: Any) -> ServiceBusSender:
         self._handlers.add(handler)
         return handler
 
-    def get_queue_receiver(self, queue_name: str, **kwargs: Any) -> ServiceBusReceiver:
+    def get_queue_receiver(
+        self,
+        queue_name: str,
+        *,
+        session_id: Optional[Union[str, ServiceBusSessionFilter]] = None,
```
yunhaoling, line 278: same as the sync one

sdk/servicebus/azure-servicebus/azure/servicebus/aio/_servicebus_receiver_async.py
```diff
@@ -120,7 +121,21 @@ class ServiceBusReceiver(collections.abc.AsyncIterator, BaseHandler, ReceiverMix
...
-        credential: Union["AsyncTokenCredential", AzureSasCredential, AzureNamedKeyCredential],
+        credential: Union["AsyncTokenCredential", "AzureSasCredential", "AzureNamedKeyCredential"],
+        *,
+        queue_name: Optional[str] = None,
+        topic_name: Optional[str] = None,
+        subscription_name: Optional[str] = None,
+        http_proxy: Optional[Dict[str, Any]] = None,
+        user_agent: Optional[str] = None,
+        logging_enable: Optional[bool] = False,
+        transport_type: Optional[TransportType] = TransportType.Amqp,
```
yunhaoling, line 132: same transport

sdk/servicebus/azure-servicebus/azure/servicebus/aio/_servicebus_session_async.py
```diff
@@ -40,7 +40,7 @@ class ServiceBusSession(BaseSession):
             :caption: Get session from a receiver
     """
 
-    async def get_state(self, **kwargs: Any) -> bytes:
+    async def get_state(self, timeout: Optional[float] = None) -> bytes:
```
yunhaoling, line 43:
```suggestion
    async def get_state(self, *, timeout: Optional[float] = None) -> bytes:
```

sdk/servicebus/azure-servicebus/azure/servicebus/aio/_servicebus_session_async.py
```diff
@@ -72,7 +71,7 @@ async def get_state(self, **kwargs: Any) -> bytes:
         return session_state
 
     async def set_state(
-        self, state: Union[str, bytes, bytearray], **kwargs: Any
+        self, state: Union[str, bytes, bytearray], timeout: Optional[float] = None
```
yunhaoling, line 74:
```suggestion
        self, state: Union[str, bytes, bytearray], *, timeout: Optional[float] = None
```

sdk/servicebus/azure-servicebus/azure/servicebus/aio/_servicebus_session_async.py
```diff
@@ -108,7 +106,7 @@ async def set_state(
             timeout=timeout,
         )
 
-    async def renew_lock(self, **kwargs: Any) -> datetime.datetime:
+    async def renew_lock(self, timeout: Optional[float] = None) -> datetime.datetime:
```
yunhaoling, line 109:
```suggestion
    async def renew_lock(self, *, timeout: Optional[float] = None) -> datetime.datetime:
```

sdk/servicebus/azure-servicebus/azure/servicebus/aio/management/_management_client_async.py
```diff
@@ -295,7 +298,29 @@ async def get_queue_runtime_properties(
         )
         return runtime_properties
 
-    async def create_queue(self, queue_name: str, **kwargs) -> QueueProperties:
+    async def create_queue(     # pylint: disable=too-many-locals
```
yunhaoling, line 301: baby nit:
```suggestion
    async def create_queue(  # pylint: disable=too-many-locals
```

sdk/servicebus/azure-servicebus/azure/servicebus/aio/management/_management_client_async.py
```diff
@@ -1062,7 +1104,14 @@ async def get_rule(
         return rule_description
 
     async def create_rule(
-        self, topic_name: str, subscription_name: str, rule_name: str, **kwargs
+        self,
+        topic_name: str,
+        subscription_name: str,
+        rule_name: str,
+        *,
+        filter: Union[CorrelationRuleFilter, SqlRuleFilter] = TrueRuleFilter(), # pylint: disable=redefined-builtin
```
yunhaoling, line 1112: baby nit
```suggestion
        filter: Union[CorrelationRuleFilter, SqlRuleFilter] = TrueRuleFilter(),  # pylint: disable=redefined-builtin
```

sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_client.py
annatisch: Are we sure we want to promote all these to keyword-only args?
We could never do this for pipeline configuration options in an HTTP-based client....

sdk/servicebus/azure-servicebus/azure/servicebus/_common/message.py
```diff
@@ -92,8 +92,24 @@ class ServiceBusMessage(
 
     """
 
-    def __init__(self, body, **kwargs):
-        # type: (Optional[Union[str, bytes]], Any) -> None
+    def __init__(
+        self,
+        body: Optional[Union[str, bytes]],
+        *,
+        application_properties: Optional[dict] = None,
```
annatisch, line 99: This should be `typing.Dict[?, ?]` (Not sure if there's a particular structure to application properties.... maybe the keys can be bytes? str? So maybe something like Dict[str, Any]?)

sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_receiver.py
```diff
@@ -120,19 +124,45 @@ class ServiceBusReceiver(
...
+        http_proxy: Optional[Dict[str, Any]] = None,
+        user_agent: Optional[str] = None,
+        logging_enable: Optional[bool] = False,
+        transport_type: Optional[TransportType] = TransportType.Amqp,
+        receive_mode: Union[
+            ServiceBusReceiveMode, str
+        ] = ServiceBusReceiveMode.PEEK_LOCK,
+        max_wait_time: Optional[float] = None,
+        auto_lock_renewer: Optional["AutoLockRenewer"] = None,
+        prefetch_count: int = 0,
```
annatisch, line 144: I would not expose `http_proxy`, `logging_enable`, `transport_type` - I'd just leave these in kwargs + docs.

sdk/servicebus/azure-servicebus/azure/servicebus/amqp/_amqp_message.py
yunhaoling: the second situation is a bit tricky, previously we have a stronger constraint by checking if users have passed a key; however now users can do 

`AmqpAnnotatedMessage(value_body=None, sequence_body=[xxxxx])` which is valid, and we can't tell whether users want to send a message with the value_body None or the sequence_body..

let's do not expand kwargs for those bodies. I'd rather not to introduce behavior changes just for the sake of making type hint look better.

sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_receiver.py
```diff
@@ -120,19 +124,45 @@ class ServiceBusReceiver(
...
+        fully_qualified_namespace: str,
+        credential: Union["TokenCredential", "AzureSasCredential", "AzureNamedKeyCredential"],
+        *,
+        queue_name: Optional[str] = None,
+        topic_name: Optional[str] = None,
+        subscription_name: Optional[str] = None,
+        http_proxy: Optional[Dict[str, Any]] = None,
+        user_agent: Optional[str] = None,
+        logging_enable: Optional[bool] = False,
+        transport_type: Optional[TransportType] = TransportType.Amqp,
```
swathipil, line 138: remove

sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_client.py
```diff
@@ -133,8 +170,20 @@ def close(self):
...
-    def from_connection_string(cls, conn_str, **kwargs):
-        # type: (str, Any) -> ServiceBusClient
+    def from_connection_string(
+        cls,
+        conn_str: str,
+        *,
+        http_proxy: Optional[Dict[str, Any]] = None,
+        user_agent: Optional[str] = None,
+        logging_enable: bool = False,
+        transport_type: TransportType = TransportType.Amqp,
```
swathipil, line 180: remove

sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_client.py
swathipil, line 99: remove

sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_sender.py
```diff
@@ -149,17 +153,26 @@ class ServiceBusSender(BaseHandler, SenderMixin):
...
+        self,
+        fully_qualified_namespace: str,
+        credential: Union["TokenCredential", "AzureSasCredential", "AzureNamedKeyCredential"],
+        *,
+        queue_name: Optional[str] = None,
+        topic_name: Optional[str] = None,
+        http_proxy: Optional[Dict[str, Any]] = None,
+        user_agent: Optional[str] = None,
+        logging_enable: bool = False,
+        transport_type: TransportType = TransportType.Amqp,
```
swathipil, line 166: remove

sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_sender.py
```diff
@@ -149,17 +153,26 @@ class ServiceBusSender(BaseHandler, SenderMixin):
...
+        **kwargs
```
swathipil, line 167:
```suggestion
        **kwargs: Any
```

sdk/servicebus/azure-servicebus/azure/servicebus/aio/_servicebus_client_async.py
```diff
@@ -75,15 +82,36 @@ class ServiceBusClient(object):
...
         fully_qualified_namespace: str,
-        credential: Union["AsyncTokenCredential", AzureSasCredential, AzureNamedKeyCredential],
+        credential: Union[
+            "AsyncTokenCredential", AzureSasCredential, AzureNamedKeyCredential
+        ],
+        *,
+        http_proxy: Optional[Dict[str, Any]] = None,
+        user_agent: Optional[str] = None,
+        logging_enable: bool = False,
+        transport_type: TransportType = TransportType.Amqp,
```
swathipil, line 92: remove

sdk/servicebus/azure-servicebus/azure/servicebus/aio/_servicebus_client_async.py
```diff
@@ -111,7 +139,20 @@ async def _create_uamqp_connection(self):
...
     @classmethod
-    def from_connection_string(cls, conn_str: str, **kwargs: Any) -> "ServiceBusClient":
+    def from_connection_string(
+        cls,
+        conn_str: str,
+        *,
+        http_proxy: Optional[Dict[str, Any]] = None,
+        user_agent: Optional[str] = None,
+        logging_enable: bool = False,
+        transport_type: TransportType = TransportType.Amqp,
```
swathipil, line 149: remove

sdk/servicebus/azure-servicebus/azure/servicebus/aio/_servicebus_receiver_async.py
swathipil, line 132: remove

sdk/servicebus/azure-servicebus/azure/servicebus/aio/_servicebus_sender_async.py
```diff
@@ -91,6 +92,13 @@ def __init__(
         self,
         fully_qualified_namespace: str,
         credential: Union["AsyncTokenCredential", AzureSasCredential, AzureNamedKeyCredential],
+        *,
+        queue_name: Optional[str] = None,
+        topic_name: Optional[str] = None,
+        http_proxy: Optional[Dict[str, Any]] = None,
+        user_agent: Optional[str] = None,
+        logging_enable: bool = False,
+        transport_type: TransportType = TransportType.Amqp,
```
swathipil, line 101: remove

sdk/servicebus/azure-servicebus/azure/servicebus/aio/management/_management_client_async.py
```diff
@@ -560,7 +577,26 @@ async def get_topic_runtime_properties(
         )
         return topic_description
 
-    async def create_topic(self, topic_name: str, **kwargs) -> TopicProperties:
+    async def create_topic(
+        self,
+        topic_name: str,
+        *,
+        default_message_time_to_live: Optional[Union[datetime.timedelta, str]] = None,
+        max_size_in_megabytes: Optional[float] = None, # TODO: said long? should it be float
```
swathipil, line 585: @yunhaoling should this be float? it says long in the docstring

sdk/servicebus/azure-servicebus/azure/servicebus/amqp/_amqp_message.py
```diff
@@ -112,8 +113,17 @@ class AmqpAnnotatedMessage(object):
     :paramtype delivery_annotations: Optional[dict]
     """
 
-    def __init__(self, **kwargs):
-        # type: (Any) -> None
+    def __init__(
+        self,
+        *,
+        header: Optional["AmqpMessageHeader"] = None,
+        footer: Optional[dict] = None,
```
swathipil, line 120: dict should be typing Dict

sdk/servicebus/azure-servicebus/azure/servicebus/amqp/_amqp_message.py
```diff
@@ -112,8 +113,17 @@ class AmqpAnnotatedMessage(object):
...
+        properties: Optional["AmqpMessageProperties"] = None,
+        application_properties: Optional[dict] = None,
+        annotations: Optional[dict] = None,
+        delivery_annotations: Optional[dict] = None,
```
swathipil, line 124: typing.Dict

sdk/servicebus/azure-servicebus/azure/servicebus/management/_management_client.py
```diff
@@ -552,8 +583,26 @@ def get_topic_runtime_properties(self, topic_name, **kwargs):
...
         return topic_description
 
-    def create_topic(self, topic_name, **kwargs):
-        # type: (str, Any) -> TopicProperties
+    def create_topic(
+        self,
+        topic_name: str,
+        *,
+        default_message_time_to_live: Optional[Union[datetime.timedelta, str]] = None,
+        max_size_in_megabytes: Optional[float] = None, # TODO: said long? should it be float
```
swathipil, line 591: @yunhaoling same question: float vs. long with typing?

sdk/servicebus/azure-servicebus/azure/servicebus/amqp/_amqp_message.py
swathipil: yep that makes more sense to me too

sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_client.py
swathipil: removing for now. BUT NEED TO REVISIT!!

sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_client.py
```diff
@@ -225,15 +282,27 @@ def get_queue_sender(self, queue_name, **kwargs):
...
+        **kwargs: Any
+    ) -> ServiceBusReceiver:
         """Get ServiceBusReceiver for the specific queue.
 
         :param str queue_name: The path of specific Service Bus Queue the client connects to.
         :keyword session_id: A specific session from which to receive. This must be specified for a
          sessionful queue, otherwise it must be None. In order to receive messages from the next available
          session, set this to ~azure.servicebus.NEXT_AVAILABLE_SESSION.
-        :paramtype session_id: Union[str, ~azure.servicebus.NEXT_AVAILABLE_SESSION]
+        :paramtype session_id: Optional[Union[str, ~azure.servicebus.NEXT_AVAILABLE_SESSION]]
```
swathipil, line 305: remove Optional here, sphinx doesn't like Optional here

swathipil: https://github.com/Azure/azure-sdk-for-python/issues/22051

sdk/servicebus/azure-servicebus/azure/servicebus/aio/management/_management_client_async.py
swathipil: should be int, fix docstring too

sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_client.py
swathipil: remove explicit kwarg for now, if resolution not reached

sdk/servicebus/azure-servicebus/azure/servicebus/management/_management_client.py
swathipil: int for Python 3

sdk/servicebus/azure-servicebus/azure/servicebus/_common/_connection_string_parser.py
```diff
@@ -12,13 +13,22 @@ class ServiceBusConnectionStringProperties(DictMixin):
...
+    ) -> None:
+        self._fully_qualified_namespace = fully_qualified_namespace
+        self._endpoint = endpoint
+        self._entity_path = entity_path
+        self._shared_access_signature = shared_access_signature
+        self._shared_access_key_name = shared_access_key_name
+        self._shared_access_key = shared_access_key
```
yunhaoling, line 31: I don't think we need to expand kwargs here as users are not expected to create a `ServiceBusConnectionStringProperties` by themselves.. The kwargs are privately used by us.

sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_receiver.py
```diff
@@ -685,8 +729,13 @@ def receive_deferred_messages(self, sequence_numbers, **kwargs):
                     self._auto_lock_renewer.register(self, message)
             return messages
 
-    def peek_messages(self, max_message_count=1, **kwargs):
-        # type: (int, Any) -> List[ServiceBusReceivedMessage]
+    def peek_messages(
```
yunhaoling, line 732: could you also help add type hint for the abandon/complete/deadletter/defer_message method? :P

though there're not kwargs there, but I feel since we're adding type hint everywhere, we shall keep consistency

sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_sender.py
```diff
@@ -172,7 +180,9 @@ def __init__(self, fully_qualified_namespace, credential, **kwargs):
             super(ServiceBusSender, self).__init__(
                 fully_qualified_namespace=fully_qualified_namespace,
                 credential=credential,
-                entity_name=entity_name,
+                entity_name=str(entity_name),
```
yunhaoling, line 183: why here we changed to cast it to a str?

sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_session.py
```diff
@@ -121,8 +119,7 @@ def get_state(self, **kwargs):
         session_state = response.get(MGMT_RESPONSE_SESSION_STATE)  # type: ignore
         return session_state
 
-    def set_state(self, state, **kwargs):
-        # type: (Union[str, bytes, bytearray], Any) -> None
+    def set_state(self, *, state: Union[str, bytes, bytearray], timeout: Optional[float] = None) -> None:
```
yunhaoling, line 122:
```suggestion
    def set_state(self, state: Union[str, bytes, bytearray], *, timeout: Optional[float] = None) -> None:
```

sdk/servicebus/azure-servicebus/azure/servicebus/aio/_servicebus_receiver_async.py
```diff
@@ -688,7 +718,7 @@ async def receive_deferred_messages(
             return messages
 
     async def peek_messages(
-        self, max_message_count: int = 1, **kwargs: Any
+        self, max_message_count: int = 1, *, sequence_number: int = 0, timeout: Optional[float] = None
```
yunhaoling, line 721: same for async, could you help add type hint for complete/abandon/defer/dead_letter message method?

sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_sender.py
yunhaoling: oh, I see that it's what we're doing in async, yea, I think cast to str does no harm, we can do it

yunhaoling (approved):

PR #22372, opened 2022-01-06
[Storage] Drop support for Python 2.7 and 3.6

jalauzon-msft:
Dropping support for Python 2.7 and 3.6 with the following steps for each package:

- Modifying `setup.py`.
- Modifying `dev_requirements.txt`.
- Modifying `mypy.ini` to use Python 3.7.
- Modifying package `README.md`.
- Modifying `samples/README.md`.
- Removing `setup.cfg`.

### DIFF
--- a/sdk/storage/azure-storage-blob-changefeed/README.md
+++ b/sdk/storage/azure-storage-blob-changefeed/README.md
@@ -1,3 +1,6 @@
+## _Disclaimer_
+_Azure SDK Python packages support for Python 2.7 has ended 01 January 2022. For more information and questions, please refer to https://github.com/Azure/azure-sdk-for-python/issues/20691_
+
 # Azure Storage Blob ChangeFeed client library for Python
 
 This preview package for Python enables users to get blob change feed events. These events can be lazily generated, iterated by page, retrieved for a specific time interval, or iterated from a specific continuation token.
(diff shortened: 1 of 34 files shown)


sdk/storage/azure-storage-queue/setup.py
```diff
@@ -62,11 +47,7 @@
     classifiers=[
         "Development Status :: 4 - Beta",
         'Programming Language :: Python',
-        'Programming Language :: Python :: 2',
-        'Programming Language :: Python :: 2.7',
         'Programming Language :: Python :: 3',
```
annatisch, line 50: "Programming Language :: Python :: 3 :: Only"

sdk/storage/azure-storage-blob-changefeed/README.md
```diff
@@ -1,3 +1,6 @@
+## _Disclaimer_
```
annatisch, line 1: Could we add a note to the changelog for the next release for each of these packages?

jalauzon-msft: Good idea. Done.

sdk/storage/azure-storage-queue/setup.py
jalauzon-msft: Oops, missed one. Good catch.

tasherif-msft:
Looks good, we should also remove test checks and import checks for python 2 for example:
https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/storage/azure-storage-blob/tests/test_container.py#L939
and 
https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/storage/azure-storage-blob/azure/storage/blob/_blob_service_client.py#L15-L18

tasherif-msft (approved):

jalauzon-msft:
Opened a follow-up issue to remove Python 2 specific checks from the code. #22427

PR #22376, opened 2022-01-07
Update text analytics live test config to support sovereign cloud testing

benbp:
This updates the live test pipeline configuration to reference environment variables dynamically on a per-cloud basis.
Currently there are environment variables defined which reference static resources generated for tests. These resources
exist in the public cloud and depend on features that are currently only available in public cloud. In order to skip
these tests in sovereign clouds, we need to keep the environment variables undefined.


kristapratico (approved):
This looks good to me, thanks for doing this @benbp!

PR #22386, opened 2022-01-07
[translation] py3 changes

kristapratico:
Ran pyupgrade on doc translation and manually made some changes to readme, kwargs, type hints, etc.

Live tests: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=1337497&view=results


sdk/translation/azure-ai-translation-document/azure/ai/translation/document/_models.py
```diff
@@ -299,8 +317,7 @@ class TranslationStatus(
     :ivar int total_characters_charged: Total characters charged across all documents within the translation operation.
     """
 
-    def __init__(self, **kwargs):
-        # type: (**Any) -> None
+    def __init__(self, **kwargs: Any) -> None:
```
kristapratico, line 320: this model and everything below is an output model only so don't feel compelled to add the kwonly args to the init

iscai-msft (approved):
:shipit:

PR #22388, opened 2022-01-07
Remove py27 and py36 from all test matrices. Leave py27 testing enabled for core.

scbedd:
@xiangyan99 this PR will:

1. Disable py27 test runs for everyone but core.
2. We will leave this while people submit PRs to remove their py27 support.
3. I merge the follow-up PR #22225  to enforce py3only regardless of test status.


xiangyan99 (approved):

scbedd:
This does not make any existing failures worse, core is clean and reflecting the continued presence of `py27` job.

PR #22394, opened 2022-01-08
[formrecognizer] Add more perf tests 

catalinaperalta:
Adding perf tests to cover some response handling

### DIFF
--- a/sdk/formrecognizer/azure-ai-formrecognizer/tests/perfstress_tests/perf_dmac_list_models.py
+++ b/sdk/formrecognizer/azure-ai-formrecognizer/tests/perfstress_tests/perf_dmac_list_models.py
@@ -1,0 +1,40 @@
+# coding=utf-8
+# ------------------------------------
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+# ------------------------------------
+
+import os
+from azure_devtools.perfstress_tests import PerfStressTest
+from azure.core.credentials import AzureKeyCredential
+from azure.ai.formrecognizer import DocumentModelAdministrationClient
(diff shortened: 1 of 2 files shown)


sdk/formrecognizer/azure-ai-formrecognizer/tests/perfstress_tests/perf_dmac_list_models.py
```diff
@@ -0,0 +1,40 @@
...
+
+    async def close(self):
+        """This is run after cleanup."""
+        await self.async_admin_client.close()
+        self.admin_client.close()
+        await super().close()
+
+    def run_sync(self):
+        """The synchronous perf test."""
+        result = self.admin_client.list_models()
```
kristapratico, line 34: A call to list_models won't make any calls to the service until you begin to iterate over the result.

```
for _ in list_models():
    pass
```
Aside from that, I'm wondering if list_* methods are good candidates for perf since I think the time to run will depend on how many models the resource has to list.

catalinaperalta: Oh yes! Thanks, I meant to call this out. I was wondering whether maybe the test should make one call to the service to get the first page and then be done. Not sure if that's better though. Let me know what you think.

kristapratico: My only thought on that is if it could depend on the service-side paging size. E.g. If it's like 100 and I get one full page with my resource that has >100 models, will we see a non-negligible difference compared with a resource that only has 2 models? Might be worth testing, not sure if that small of a difference will matter.

PR #22397, opened 2022-01-08
[EventHubs] Pure Python AMQP Sycn Implementation Integration

yunhaoling:
pure python amqp changes is here: https://github.com/Azure/azure-uamqp-python/pull/306


sdk/eventhub/azure-eventhub/azure/eventhub/_consumer.py
```diff
@@ -216,21 +209,20 @@ def receive(self, batch=False, max_batch_size=300, max_wait_time=None):
         self._receive_start_time = self._receive_start_time or time.time()
         deadline = self._receive_start_time + (max_wait_time or 0)  # max_wait_time can be None
         if len(self._message_buffer) < max_batch_size:
+            # TODO: the retry here is a bit tricky as we are using low-level api from the amqp client.
+            #  besides, the retry will create a new AMQPClient + new link properties (offset)
+            #  so we need to create new amqp client or update the link properties of existing one
+            #  I'd like to add more optional steps into the retry policy of amqp lib
+            #  to allow updating existing properties for recreation of a client if link/connection error happens
```
annatisch, line 216: _Should_ we be creating a new client?
Maybe the same client should recreate its connection?
A new client certainly shouldn't be needed just for a new link

sdk/eventhub/azure-eventhub/azure/eventhub/_consumer.py
```diff
@@ -216,21 +209,20 @@ def receive(self, batch=False, max_batch_size=300, max_wait_time=None):
...
                 try:
                     if self._open():
-                        messages =self._handler.receive_message_batch(
-                            max_batch_size=max_batch_size
-                        )
-                        for message in messages:
-                            uamqp_message = Message(**message)
-                            self._message_buffer.appendleft(uamqp_message)
-                        #self._handler.do_work()  # type: ignore
+                        self._handler.do_work(batch=self._prefetch)  # type: ignore
```
annatisch, line 220: It looks like we're changing the EH Consumer to work based on callbacks rather than the receive_batch function. What was the motivation for this?

sdk/eventhub/azure-eventhub/azure/eventhub/_producer.py
```diff
@@ -112,72 +119,41 @@ def __init__(self, client, target, **kwargs):
...
-            self._handler.send_message(self._unsent_events[0])
+            timeout = timeout_time - time.time() if timeout_time else 0
+            self._handler.send_message(self._unsent_events[0], timeout=timeout)
             self._unsent_events = None
-            # self._unsent_events = self._handler.pending_messages  # type: ignore
-            # if self._outcome != constants.MessageSendResult.Ok:
-            #     if self._outcome == constants.MessageSendResult.Timeout:
-            #         self._condition = OperationTimeoutError("Send operation timed out")
-            #     if self._condition:
-            #         raise self._condition
```
annatisch, line 162: We no longer need to monitor unsent messages?

sdk/eventhub/azure-eventhub/azure/eventhub/pyamqp/__init__.py
```diff
@@ -10,4 +10,4 @@
 from ._connection import Connection
 from ._transport import SSLTransport
 
-from .client import ReceiveClient, SendClient
+from .client import AMQPClient, ReceiveClient, SendClient
```
annatisch, line 13: Why is `pyamqp` public? This module should be private?

sdk/eventhub/azure-eventhub/tests/livetest/synctests/test_reconnect.py
```diff
@@ -64,30 +66,38 @@ def test_send_with_long_interval_sync(live_eventhub, sleep):
...
     with sender:
-            sender._open_with_retry()
-            time.sleep(11)
-            sender._unsent_events = [ed.message]
-            ed.message.on_send_complete = sender._on_outcome
-            with pytest.raises((uamqp.errors.ConnectionClose,
-                                uamqp.errors.MessageHandlerError, OperationTimeoutError)):
-                # Mac may raise OperationTimeoutError or MessageHandlerError
-                sender._send_event_data()
-            sender._send_event_data_with_retry()
```
annatisch, line 80: We're no longer getting these exceptions raised because now we've built the retry logic into pyamqp rather than EH?

sdk/eventhub/azure-eventhub/azure/eventhub/_consumer.py
yunhaoling: I agree, reusing the same client is the preferred approach.

however, I have not yet come up with a design on how to deal with callback mode retry as we're only calling the `client.do_work` here which has limitation.

maybe I would add a new method into amqp client to include sophisticated logic for retry handling in callback mode..

sdk/eventhub/azure-eventhub/azure/eventhub/_consumer.py
yunhaoling: it's not changed, current EH consumer is also based on callbacks

sdk/eventhub/azure-eventhub/tests/livetest/synctests/test_reconnect.py
yunhaoling: well, yes and no.

as we build our retry logic into pyamqp, the `_send_event_data` would raise error when there's no retry and succeed if there's retry.

I split the test into two, one with 0 retry config client which would raise, and another with normal retry config client which would succeed

sdk/eventhub/azure-eventhub/azure/eventhub/pyamqp/__init__.py
yunhaoling: agreed, I will make it private

sdk/eventhub/azure-eventhub/azure/eventhub/_producer.py
yunhaoling: We need to monitor because amqp send client `queue` and `wait` were used to send messages -- We didn't use uamqp send_message because there was an issue with the implementation (probably dealing with timeout?).

but now we have reimplemented the lib, I think it's time to use the `send_message` method which is simpler and the right way to do it

sdk/eventhub/azure-eventhub/azure/eventhub/_consumer.py
annatisch: Is this needed for this preview?
If not - let's have a chat about what's needed here and how we can cleanly incorporate it.

yunhaoling: I'm afraid it is needed -- otherwise, the behavior is somewhat breaking because per retry would receive duplicate messages not from the last received event's offset when error happened.

annatisch: Okay - well we can do whatever is needed to make the preview happy - but ultimately I don't think this should exist as a separate method - we should make the default path successful.

PR #22414, opened 2022-01-10
Generate Manifests alongside build artifacts

scbedd:
This adds two steps to the `build-artifacts` stage of our yaml pipelines.

Related to Azure/azure-sdk-tools#2273

### DIFF
--- a/eng/pipelines/templates/steps/build-artifacts.yml
+++ b/eng/pipelines/templates/steps/build-artifacts.yml
@@ -108,7 +108,7 @@ steps:
 
   - task: AzureArtifacts.manifest-generator-task.manifest-generator-task.ManifestGeneratorTask@0
     displayName: 'Upload Documentation SBOM'
-    condition: and(succeeded(), ${{ parameters.CustomCondition }})
+    condition: and(succeeded(), ${{ parameters.BuildDocs }})
     inputs:
       BuildDropPath: $(Build.SourcesDirectory)/_docs
 
--- a/eng/pipelines/templates/steps/build-artifacts.yml
+++ b/eng/pipelines/templates/steps/build-artifacts.yml
@@ -101,6 +101,17 @@ steps:
 
   - ${{ parameters.BeforePublishSteps }}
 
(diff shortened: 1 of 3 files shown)


eng/pipelines/templates/steps/build-artifacts.yml
```diff
@@ -101,6 +101,17 @@ steps:
 
   - ${{ parameters.BeforePublishSteps }}
 
+  - task: AzureArtifacts.manifest-generator-task.manifest-generator-task.ManifestGeneratorTask@0
+    displayName: 'Upload Package SBOM'
+    inputs:
+      BuildDropPath: $(Build.ArtifactStagingDirectory)
+
+  - task: AzureArtifacts.manifest-generator-task.manifest-generator-task.ManifestGeneratorTask@0
```
weshaggard, line 109: Do we need a sbom for the docs?

eng/pipelines/templates/variables/globals.yml
```diff
@@ -1,4 +1,5 @@
 variables:
   PythonVersion: '3.6'
   skipComponentGovernanceDetection: true
-  AzureSDKCondaChannel: https://azuresdkconda.blob.core.windows.net/channel1/
\ No newline at end of file
+  AzureSDKCondaChannel: https://azuresdkconda.blob.core.windows.net/channel1/
+  Package.EnableSBOMSigning: true
```
weshaggard, line 5: Do we need to set this yet? Will it even work for public builds?

weshaggard (approved):

eng/pipelines/templates/variables/globals.yml
scbedd: Either way, it won't do any harm! (At least not according to their documentation)

I think that at the outset they won't enforce on this, but they _will_. Leaving it on by default is good imo.

eng/pipelines/templates/steps/build-artifacts.yml
scbedd: I think so, yes. I figure it's a shipped artifact so 🤷‍♂️

PR #22418, opened 2022-01-10
[ServiceBus&EventHubs] api fixes

swathipil:
updated the following, after API review:
EH:
1) update `EventData.body` to type primitives (not Any): None/bool/int/float/bytes/str/dict/list/uuid.UUID

SB:
2) update type of kwarg `session_id` in sync/async `ServiceBusClient.get_queue_receiver/get_subscription_receiver` to `Union[str, Literal[NEXT_AVAILABLE_SESSION]]`
3) update return type of async/sync `ServiceBusSession.get_state()` to `bytes`
4) move docstring type of kwarg `sub_queue`  to `:paramtype` line in async/sync `ServiceBusClient.get_queue_receiver`

### DIFF
--- a/sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
+++ b/sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
@@ -64,6 +64,7 @@ PrimitiveTypes = Optional[Union[
     int,
     float,
     bytes,
+    bool,
     str,
     Dict,
     List,
@@ -339,7 +340,7 @@ class EventData(object):
         For :class:`azure.eventhub.amqp.AmqpMessageBodyType.VALUE<azure.eventhub.amqp.AmqpMessageBodyType.VALUE>`,
         the body could be any type.
 
-        :rtype: int or float or bytes or str or Dict or List or uuid.UUID
+        :rtype: int or bool or float or bytes or str or Dict or List or uuid.UUID
         """
         try:
             return self._raw_amqp_message.body
--- a/sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
+++ b/sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
@@ -340,7 +340,7 @@ class EventData(object):
         For :class:`azure.eventhub.amqp.AmqpMessageBodyType.VALUE<azure.eventhub.amqp.AmqpMessageBodyType.VALUE>`,
         the body could be any type.
 
-        :rtype: int or bool or float or bytes or str or Dict or List or uuid.UUID
+        :rtype: int or bool or float or bytes or str or dict or list or uuid.UUID
         """
         try:
             return self._raw_amqp_message.body
--- a/sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
+++ b/sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
@@ -6,6 +6,7 @@ from __future__ import unicode_literals
 
 import json
 import logging
+import uuid
 from typing import (
     Union,
     Dict,
@@ -58,6 +59,17 @@ from .amqp import (
 
 if TYPE_CHECKING:
     import datetime
+
+PrimitiveTypes = Union[
+    None,
+    int,
+    float,
+    bytes,
+    str,
+    Dict,
+    List,
+    uuid.UUID,
+]
 
 _LOGGER = logging.getLogger(__name__)
 
@@ -319,7 +331,7 @@ class EventData(object):
 
     @property
     def body(self):
-        # type: () -> Any
+        # type: () -> PrimitiveTypes
         """The body of the Message. The format may vary depending on the body type:
         For :class:`azure.eventhub.amqp.AmqpMessageBodyType.DATA<azure.eventhub.amqp.AmqpMessageBodyType.DATA>`,
         the body could be bytes or Iterable[bytes].
@@ -328,7 +340,7 @@ class EventData(object):
         For :class:`azure.eventhub.amqp.AmqpMessageBodyType.VALUE<azure.eventhub.amqp.AmqpMessageBodyType.VALUE>`,
         the body could be any type.
 
-        :rtype: Any
+        :rtype: Union[None, int, float, bytes, str, Dict, List, uuid.UUID]
         """
         try:
             return self._raw_amqp_message.body
--- a/sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
+++ b/sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
@@ -60,8 +60,7 @@ from .amqp import (
 if TYPE_CHECKING:
     import datetime
 
-PrimitiveTypes = Union[
-    None,
+PrimitiveTypes = Optional[Union[
     int,
     float,
     bytes,
@@ -69,7 +68,7 @@ PrimitiveTypes = Union[
     Dict,
     List,
     uuid.UUID,
-]
+]]
 
 _LOGGER = logging.getLogger(__name__)
 
--- a/sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
+++ b/sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
@@ -339,7 +339,7 @@ class EventData(object):
         For :class:`azure.eventhub.amqp.AmqpMessageBodyType.VALUE<azure.eventhub.amqp.AmqpMessageBodyType.VALUE>`,
         the body could be any type.
 
-        :rtype: Union[None, int, float, bytes, str, Dict, List, uuid.UUID]
+        :rtype: int or float or bytes or str or Dict or List or uuid.UUID
         """
         try:
             return self._raw_amqp_message.body
--- a/sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_client.py
+++ b/sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_client.py
@@ -5,7 +5,6 @@
 from typing import Any, Union, Optional, TYPE_CHECKING
 import logging
 from weakref import WeakSet
-from typing_extensions import Literal
 
 import uamqp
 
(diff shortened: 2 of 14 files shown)


sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
```diff
@@ -59,6 +60,17 @@
 if TYPE_CHECKING:
     import datetime
 
+PrimitiveTypes = Union[
+    None,
```
annatisch, line 64: Should this be NoneType? Or maybe instead the whole union should be wrapped in Optional?

sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_client.py
```diff
@@ -285,13 +288,14 @@ def get_queue_receiver(
         :keyword session_id: A specific session from which to receive. This must be specified for a
          sessionful queue, otherwise it must be None. In order to receive messages from the next available
          session, set this to ~azure.servicebus.NEXT_AVAILABLE_SESSION.
-        :paramtype session_id: Union[str, ~azure.servicebus.NEXT_AVAILABLE_SESSION]
-        :keyword Optional[Union[ServiceBusSubQueue, str]] sub_queue: If specified, the subqueue this receiver will
+        :paramtype session_id: Union[str, Literal[~azure.servicebus.NEXT_AVAILABLE_SESSION]]
```
annatisch, line 291: I would not document it like this in the docstring. Docstring formatting != type hints.
It should be something like:
`:paramtype session_id: str or ~azure.servicebus.NEXT_AVAILABLE_SESSION`

sdk/servicebus/azure-servicebus/azure/servicebus/aio/_servicebus_client_async.py
```diff
@@ -3,6 +3,7 @@
 # Licensed under the MIT License. See License.txt in the project root for license information.
 # --------------------------------------------------------------------------------------------
 from typing import Any, Union, Optional, TYPE_CHECKING
+from typing_extensions import Literal
```
swathipil, line 6: I planned to do this:
```
try:
  from typing import Literal
except ImportError:
  from typing_extensions import Literal
```
but `typing.Literal` does not exist in Python 3.7, so it results in a mypy error. I'm not sure if there's a better way to do this? @annatisch

sdk/servicebus/azure-servicebus/azure/servicebus/aio/_servicebus_client_async.py
```diff
@@ -21,13 +22,16 @@
...
-    ServiceBusReceiveMode
+    ServiceBusReceiveMode,
+    NEXT_AVAILABLE_SESSION,
 )
 from ._async_utils import create_authentication
 
 if TYPE_CHECKING:
     from azure.core.credentials_async import AsyncTokenCredential
 
+NextAvailableSessionType = Literal[NEXT_AVAILABLE_SESSION]  # type: ignore
```
swathipil, line 33: This line (without the type: ignore) also resulted in a mypy error for all python versions: `error: Variable "azure.servicebus.aio._servicebus_client_async.NextAvailableSessionType" is not valid as a type`, so I ignored type.
I'm not sure if there's a better alternative? @annatisch

annatisch: Hah - if we have to do `type: ignore` then the whole thing is moot and we should just remove it ;)
Did you check the `Literal` docs? My suggestion was not researched, so may have been invalid...

swathipil: yep that makes sense! I tried this method instead here: 

I had to update this line to the specific enum (`ServiceBusSessionFilter.NEXT_AVAILABLE`), to get rid of the mypy error. I think it only takes the exact enum as a parameter.

This would expose the `ServiceBusSessionFilter` to users, which I know we didn't want to do. Since we also don't want type: ignore, I'm not sure that this would work.

Thoughts @annatisch ?

swathipil: resolution: expose ServiceBusSessionFilter

sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
```diff
@@ -59,6 +60,16 @@
 if TYPE_CHECKING:
     import datetime
 
+PrimitiveTypes = Optional[Union[
+    int,
```
yunhaoling, line 64: `bool` is also one of the types we support

sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
```diff
@@ -328,7 +339,7 @@ def body(self):
         For :class:`azure.eventhub.amqp.AmqpMessageBodyType.VALUE<azure.eventhub.amqp.AmqpMessageBodyType.VALUE>`,
         the body could be any type.
 
-        :rtype: Any
+        :rtype: int or float or bytes or str or Dict or List or uuid.UUID
```
yunhaoling, line 342: bool is missing, also connecting by str looks a bit weird.. should it be like the following for sphinx:
`int, float, bytes, str, Dict, List or uuid.UUID`?

sdk/servicebus/azure-servicebus/azure/servicebus/aio/_servicebus_sender_async.py
```diff
@@ -257,7 +261,7 @@ async def schedule_messages(
             )
 
     async def cancel_scheduled_messages(
-        self, sequence_numbers: Union[int, List[int]], *, timeout: Optional[float] = None
+        self, sequence_numbers: Union[int, List[int]], *, timeout: Optional[float] = None, **kwargs:Any
```
yunhaoling, line 264: nit & pylint failure
```suggestion
        self, sequence_numbers: Union[int, List[int]], *, timeout: Optional[float] = None, **kwargs: Any
```

sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
swathipil: good catch!

I think they should each be or b/c that's how I've seen it in other places. For ex: https://github.com/Azure/azure-sdk-for-python/blob/60a7f629d56a8380806df322d1d5a3fa1559b718/sdk/servicebus/azure-servicebus/azure/servicebus/aio/_servicebus_client_async.py#L48

sdk/servicebus/azure-servicebus/setup.py
```diff
@@ -69,5 +69,6 @@
         'azure-core<2.0.0,>=1.14.0',
         "isodate>=0.6.0",
         "six>=1.11.0",
+        "typing-extensions>=3.7.4.3",
```
swathipil, line 72: qq @annatisch: is this the right place to put this? or should this be elsewhere?

sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
yunhaoling: looks like you're right:

image


https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html

yunhaoling: but one more thing, does sphinx support `Dict` or `List` which I feel are more of a "typing" thing as Laurent pointed out in the issue:https://github.com/Azure/azure-sdk-for-python/issues/22051

probably should just be `dict` or `list` (lower case)

swathipil: ah yep!

sdk/servicebus/azure-servicebus/setup.py
annatisch: Should only be need for <3.8:
https://github.com/Azure/azure-sdk-for-python/blob/feature/communication-CallingServer-preview/sdk/communication/azure-communication-chat/setup.py#L59-L67

sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_client.py
```diff
@@ -5,6 +5,7 @@
 from typing import Any, Union, Optional, TYPE_CHECKING
 import logging
 from weakref import WeakSet
+from typing_extensions import Literal
```
annatisch, line 8: As you mentioned in a comment elsewhere - I think this will need to be:
```python
try:
    from typing import Literal
except ImportError:
    from typing_extensions import Literal
```

sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_receiver.py
```diff
@@ -683,6 +685,8 @@ def receive_deferred_messages(
                 :caption: Receive deferred messages from ServiceBus.
 
         """
+        if kwargs:
+            warnings.warn(f"Unsupported keyword args: {kwargs}")
```
annatisch, line 689: Is the `f {}` formatting support in 3.6+?

swathipil: yes it is

sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_client.py
swathipil: this causes a mypy error, same as this one: https://github.com/python/typeshed/issues/3500

They mention checking python version as the solution. I'm not sure if that's what we want to do?

annatisch: If we add the setup.py amendment above (which we probably should - we don't want unnecessary dependencies) then typing_extensions wont be available in <3.8.
As linked to within the issue you linked to - Mypy should support/understand Python version checks.

swathipil: sounds good, thank you!

swathipil: It looks like there's still an issue with mypy, but I think this is because this mypy check in the pipeline is using Python 3.7.

I can ask Scott about this.

annatisch (approved):

jaystyles (commented):
F

Issue #21951, opened 2021-12-01
Manually install pypy37 in `UsePythonVersion` custom task. Remove pinned `cryptography`


scbedd:
[This PR](https://github.com/microsoft/azure-pipelines-tasks/pull/15553/files) highlights the issue at play here. We can't use the default `UsePythonVersion` task because it's hardcoded for pypy3.6 only.

Need to use our installer script to pull down the pypy version, and bypass the task / manually prepend the path for the targeted  pypy3 versions.

After we address issues with pypy not being truly compatible with the UsePythonVersion task, we can remove this.

PR #22419, opened 2022-01-10
[core] fix ci

iscai-msft:


iscai-msft:
this is caused by #21951, @scbedd is working on a fix for this, so closing this PR

PR #22425, opened 2022-01-10
[tox] fix too many outputs bug

iscai-msft:
all creds to @kristapratico


scbedd (approved):
This definitely seems like the better way to do it!

FYI to readers without context pip list isn't guaranteed to have exactly two outputs like is visible in the code here. In some cases, it will return 3. By changing to `--format freeze`, we ensure that we can delimit the split on a known value, and eliminate the edge case.

PR #22449, opened 2022-01-11
Fix typos in README.md

seankane-msft:
Fixes markdown formatting for multiplication values which were appearing as italicized.


jalauzon-msft (approved):
Thanks for doing this!

Issue #22234, opened 2021-12-20
[mgmt] drop python2 


msyyc:
RE: Python 3 packaging updates starting Jan 1st

related issue : https://github.com/Azure/azure-sdk-for-python/issues/20691
referenced PR :https://github.com/Azure/azure-sdk-for-python/pull/22192

msyyc:
RE: Python 3 packaging updates starting Jan 1st

PR #22459, opened 2022-01-12
add python36 back

msyyc:
https://github.com/Azure/azure-sdk-for-python/issues/22234

RE: Problem with last Python SDK


BigCat20196 (approved):

ofek:
When will support for 3.6 end? It's EOL now

PR #22475, opened 2022-01-12
[Cosmos] Remove support for Python27, set Python36 as minimum, and update cosmos emulator pipeline

simorenoh:
This PR takes care of updating Cosmos to remove Python27 support and upgrade the minimum version to 3.6 as per current guidelines: https://github.com/Azure/azure-sdk-for-python/blob/main/doc/dev/packaging.md. 

It also takes care of updating the cosmos emulator pipeline step to go against 3.6.


sdk/cosmos/azure-cosmos/dev_requirements.txt
```diff
@@ -1,3 +1,4 @@
 -e ../../../tools/azure-sdk-tools
 ../../core/azure-core
--e ../../../tools/azure-devtools
\ No newline at end of file
+-e ../../../tools/azure-devtools
+python_version >= '3.6'
```
annatisch, line 4: Don't think we need this line

sdk/cosmos/azure-cosmos/test/test_location_cache.py
```diff
@@ -12,19 +12,14 @@
...
 
 class RefreshThread(threading.Thread):
     def __init__(self, group=None, target=None, name=None,
                  args=(), kwargs=None, verbose=None):
-        if six.PY2:
-            super(RefreshThread, self).__init__(group=group, target=target,
-                                      name=name, verbose=verbose)
-        else:
-            super().__init__()
+        super().__init__()
```
annatisch, line 22: hmm I wonder why the previous py2 code was passing in keyword params and the py3 code doesn't?

sdk/cosmos/azure-cosmos/setup.py
```diff
@@ -25,14 +25,26 @@
...
+    raise RuntimeError("Cannot find version information")
+
 with open("README.md", encoding="utf-8") as f:
     readme = f.read()
 with open("CHANGELOG.md", encoding="utf-8") as f:
     changelog = f.read()
 
+exclude_packages = [
+    "tests",
+    "tests.*",
```
annatisch, line 38: What did we need this one for?

sdk/cosmos/azure-cosmos/test/test_location_cache.py
simorenoh: not too sure to be honest, doesn't seem like they were even being set anywhere within the test

sdk/cosmos/azure-cosmos/setup.py
simorenoh: good catch, it's also the wrong name for cosmos - we have tests as `test` for whatever reason, ill update

annatisch (approved):

gahl-levy:
LGTM, thanks for updating this @simorenoh!

PR #22533, opened 2022-01-18
[ServiceBus&EventHubs] cleanup pylint/docs

### LINKED ISSUE #15432: [ServiceBus] Clean up pylint ignores where applicable
We aspire to be pylint green, but aren't always perfect.  There are places that pylint ignores are emplaced, that could potentially be improved.  On the easier end would be things like line length, but one should feel empowered to address any that they think are safe and self-contained.

Address the code to prevent the need for the pylint ignores, as applicable.

swathipil:
fixes #15432
- cleaning up pylint disable, where possible
- update dict --> typing.Dict in type hints

### DIFF
--- a/sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
+++ b/sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
@@ -386,7 +386,7 @@ class EventData(object):
 
         :param encoding: The encoding to use for decoding event data.
          Default is 'UTF-8'
-        :rtype: dict
+        :rtype: Dict[str, Any]
         """
         data_str = self.body_as_str(encoding=encoding)
         try:
--- a/sdk/eventhub/azure-eventhub/azure/eventhub/amqp/_amqp_message.py
+++ b/sdk/eventhub/azure-eventhub/azure/eventhub/amqp/_amqp_message.py
@@ -4,7 +4,7 @@
 # license information.
 # -------------------------------------------------------------------------
 
-from typing import Optional, Any, cast, Mapping
+from typing import Optional, Any, cast, Mapping, Dict
 
 import uamqp
 
@@ -34,15 +34,15 @@ class AmqpAnnotatedMessage(object):
     :keyword header: The amqp message header.
     :paramtype header: Optional[~azure.eventhub.amqp.AmqpMessageHeader]
     :keyword footer: The amqp message footer.
-    :paramtype footer: Optional[dict]
+    :paramtype footer: Optional[Dict]
     :keyword properties: Properties to add to the amqp message.
     :paramtype properties: Optional[~azure.eventhub.amqp.AmqpMessageProperties]
     :keyword application_properties: Service specific application properties.
-    :paramtype application_properties: Optional[dict]
+    :paramtype application_properties: Optional[Dict]
     :keyword annotations: Service specific message annotations.
-    :paramtype annotations: Optional[dict]
+    :paramtype annotations: Optional[Dict]
     :keyword delivery_annotations: Service specific delivery annotations.
-    :paramtype delivery_annotations: Optional[dict]
+    :paramtype delivery_annotations: Optional[Dict]
     """
 
     def __init__(self, **kwargs):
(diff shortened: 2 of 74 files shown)


sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
```diff
@@ -386,7 +386,7 @@ def body_as_json(self, encoding="UTF-8"):
 
         :param encoding: The encoding to use for decoding event data.
          Default is 'UTF-8'
-        :rtype: dict
+        :rtype: Dict[str, Any]
```
yunhaoling, line 389: quick question: have you run sphinx to see how it's rendered in the sphinx doc?

I'm wondering how it would look as `rtype` is a sphinx thing while `Dict` is from typing -- Laurent mentioned in the [issue](https://github.com/Azure/azure-sdk-for-python/issues/22051): "Sphinx do not use typing, but a more classical system"

sdk/eventhub/azure-eventhub/azure/eventhub/amqp/_amqp_message.py
```diff
@@ -34,15 +34,15 @@ class AmqpAnnotatedMessage(object):
     :keyword header: The amqp message header.
     :paramtype header: Optional[~azure.eventhub.amqp.AmqpMessageHeader]
     :keyword footer: The amqp message footer.
-    :paramtype footer: Optional[dict]
+    :paramtype footer: Optional[Dict]
```
yunhaoling, line 37: should be `dict or None`? as Laurent mentioned that "can add or None if the optionality matters, but we should clarify it in the text."

yunhaoling:
I'm a bit uncertain about this "update dict --> typing.Dict in type hints" -- as what we're doing here is updating the docstring which is a sphinx thing, while sphinx doesn't "like" `typing` module.

let's check how it's rendered

sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
swathipil: I have not, but there are actually places in the docs where we already have typing.Dict and it looks fine. For example, here's Dict looks in the APIView for type `properties` in `EventData`: https://apiview.dev/Assemblies/Review/3e49ce1ca870470289360023dcb51edf#

sdk/eventhub/azure-eventhub/azure/eventhub/amqp/_amqp_message.py
swathipil: Since this is a problem in a lot of places, not just this line, is it fine to leave it here and address it in the issue Laurent made?

sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
yunhaoling: okay.. the doc of "<xref:Iterable>[<xref:Dict>[<xref:str,Any>]]" doesn't seem correct -- which means changing from `rtype: dict` to `:rtype: Dict[str, Any]` would lead to Api reference doc "broken" which is not a good documentation experience.

I think we should keep the `rtype: dict` or changed to `rtype: dict[str, any]` instead of using typing one `:rtype: Dict[str, Any]`.

sdk/eventhub/azure-eventhub/azure/eventhub/amqp/_amqp_message.py
yunhaoling: I agree that we should tackle one issue at a time.

As long as we ultimately update the sphinx docstring to be using `dict or None` instead of `Optional[Dict]`, I'm good.

sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
swathipil: sounds good - does this mean we should update all Dict to dict then?

swathipil: as per OOB discussion: leaving this as Dict for now. Plan on updating all types that don't follow sphinx rules in the future.

yunhaoling (approved):
approved, as per OOB discussion, would need to have a separate PR for sphinx docstring correctness

PR #22566, opened 2022-01-19
Remove Remaining azure-core py2.7

scbedd:
@lmazuel @xiangyan99 as requested in yesterdays scrum.


xiangyan99:
I have done it in #22531?

scbedd:
Not the same thing though. This removes the _run_ entirely.

scbedd:
Yep. You're right. Closing as dup.

PR #22570, opened 2022-01-20
[formrecognizer] Expose new properties on models

catalinaperalta:
This PR adds the new `api_version` and `tags` properties to ModelOperationInfo, ModelOperation, DocumentModelInfo, and DocumentModel. Added `build_mode` of `DocTypeInfo`.

NOTE: other build model tests will be enabled in a separate PR. This PR enables tests that can verify the new properties in service responses. There will be a few other properties added in a separate PR after regeneration with the latest swagger.

Related issues and pending work: #22719, #22718


sdk/formrecognizer/azure-ai-formrecognizer/tests/test_dmac_mgmt.py
```diff
@@ -145,10 +148,12 @@ def test_get_list_operations(self, client):
         # check successful op
         if successful_op:
             op = client.get_operation(successful_op.operation_id)
+            # TODO not seeing this returned at the operation level
+            # assert op.api_version
```
catalinaperalta, line 152: Tracking issue:  #22719

sdk/formrecognizer/azure-ai-formrecognizer/azure/ai/formrecognizer/_models.py
```diff
@@ -3258,6 +3258,10 @@ class ModelOperationInfo(object):
     :vartype kind: str
     :ivar resource_location: URL of the resource targeted by this operation.
     :vartype resource_location: str
+    :ivar api_version: API version used to create this operation.
+    :vartype api_version: str
+    :ivar tags: List of key-value tag attributes associated with the model.
```
kristapratico, line 3263: should we say "user-defined" kv attributes here?

sdk/formrecognizer/azure-ai-formrecognizer/azure/ai/formrecognizer/_models.py
```diff
@@ -3268,18 +3272,22 @@ def __init__(self, **kwargs):
...
+            "kind={}, resource_location={}, api_version={}, tags={})".format(
                 self.operation_id,
                 self.status,
                 self.percent_completed,
                 self.created_on,
                 self.last_updated_on,
                 self.kind,
                 self.resource_location,
+                self.api_version,
+                repr(self.tags),
```
kristapratico, line 3290: I don't think it matters a ton, but I don't think we need the repr here around a standard dict

sdk/formrecognizer/azure-ai-formrecognizer/azure/ai/formrecognizer/_models.py
```diff
@@ -3809,6 +3861,9 @@ class DocTypeInfo(object):
     fields and types, and the confidence for those fields.
 
     :ivar str description: A description for the model.
+    :ivar build_mode: The build mode used when building the custom model.
+     Possible values include: "template", "neural".
+    :vartype build_mode: str or :class:`~azure.ai.formrecognizer.DocumentBuildMode`
```
kristapratico, line 3866: this is an output model only, right? I think vartype should be str only. We can say in the ivar doc that its value is represented by DocumentBuildMode but the user won't ever get an enum returned here.

sdk/formrecognizer/azure-ai-formrecognizer/tests/test_dmac_mgmt.py
```diff
@@ -125,14 +127,15 @@ def test_mgmt_model(self, client, formrecognizer_storage_container_sas_url, **kw
...
-    def test_get_list_operations(self, client):
+    def test_get_list_operations(self, client, **kwargs):
         operations = client.list_operations()
         successful_op = None
         failed_op = None
         for op in operations:
             assert op.operation_id
             assert op.status
-            assert op.percent_completed is not None
+            # FIXME check why some operations aren't returned with a percent_completed field
```
kristapratico, line 137: I thought Paul mentioned that they haven't fully finished this feature :/

kristapratico (approved):

sdk/formrecognizer/azure-ai-formrecognizer/azure/ai/formrecognizer/_models.py
catalinaperalta: That seems pretty safe, I did pause for a moment in case this set of tags could ever include non-user-defined tags, but that is probably very unlikely. I'll add this and just confirm with the service team if this is suitable wording for this property.

PR #22624, opened 2022-01-21
Issue #9324 create test_compress_compressed_no_header_offline

marcusc2:
This PR will:
- mark test_compress_compressed_no_header() as live test only
- create test_compress_compressed_no_header_offline()
- add stream_compress() and compress() in streams.py to replicate online test behavior

Please disregard the previous PR


sdk/core/azure-core/tests/testserver_tests/coretestserver/coretestserver/test_routes/streams.py
```diff
@@ -48,6 +54,10 @@ def string():
     return Response(
         streaming_test(), status=200, mimetype="text/plain"
     )
+
+@streams_api.route('/compress', methods=['GET'])
```
xiangyan99, line 58: Is this compressed stream w/o compress header?

Could you add this information into the API?

sdk/core/azure-core/tests/testserver_tests/coretestserver/coretestserver/test_routes/streams.py
```diff
@@ -31,6 +32,11 @@ def streaming_test():
 def stream_compressed_header_error():
     yield b'test'
 
+def stream_compress():
+    file_path = os.path.abspath(os.path.join(os.path.abspath(__file__), "..", "..", "./files/test.tar.gz"))
```
kristapratico, line 36: Okay I see the error in CI now: https://dev.azure.com/azure-sdk/public/_build/results?buildId=1318031&view=logs&j=c7500c44-708b-56ec-f52b-02af33ea1457&t=df38eb02-057c-5be2-a959-7d0c1b0db6de&l=3054

It looks like it's not finding the file in CI. I'm not totally sure why the path created locally is different than what we see in CI, but we can try adding a try/except here removing one of the ".." so that it works locally and in CI.

sdk/core/azure-core/tests/testserver_tests/coretestserver/coretestserver/test_routes/streams.py
```diff
@@ -32,8 +32,11 @@ def streaming_test():
...
 
-def stream_compress():
-    file_path = os.path.abspath(os.path.join(os.path.abspath(__file__), "..", "..", "./files/test.tar.gz"))
+def stream_compressed_no_header():
+    try:
+        file_path = os.path.abspath(os.path.join(os.path.abspath(__file__), "..", "..", "./files/test.tar.gz"))
+    except:
+        file_path = os.path.abspath(os.path.join(os.path.abspath(__file__), "..", "./files/test.tar.gz"))
     with open(file_path, "rb") as fd:
         yield fd.read()
```
kristapratico, line 41: The FileNotFoundError arises from the "with open" code so I think you need to include that in your try/except

sdk/core/azure-core/tests/test_streaming.py
```diff
@@ -92,6 +93,21 @@ def test_decompress_compressed_no_header(http_request):
     except UnicodeDecodeError:
         pass
 
+@pytest.mark.parametrize("http_request", HTTP_REQUESTS)
+def test_compress_compressed_no_header_offline(port, http_request):
+    # expect compressed text
+    request = http_request(method="GET", url="http://localhost:{}/streams/compressed_no_header".format(port))
+    with RequestsTransport() as sender:
+        try:
+            response = sender.send(request, stream=True)
```
iscai-msft, line 102: @marcusc2 can you use `pipeline.run` instead of `send`? For our newer azure core rest, everything needs to be properly read, so easier to use `pipeline.run` instead.

```python
client = PipelineClient("")
request = http_request(method="GET", url="http://localhost:{}/streams/compressed_no_header".format(port))
pipeline_response = client._pipeline.run(request, stream=True)
```

sdk/core/azure-core/tests/testserver_tests/coretestserver/coretestserver/test_routes/streams.py
```diff
@@ -31,6 +32,16 @@ def streaming_test():
...
 
+def stream_compressed_no_header():
+    try:
+        file_path = os.path.abspath(os.path.join(os.path.abspath(__file__), "..", "./files/test.tar.gz"))
+        with open(file_path, "rb") as fd:
+            yield fd.read()
+    except:
+        file_path = os.path.abspath(os.path.join(os.path.abspath(__file__), "..", "..", "./files/test.tar.gz"))
+        with open(file_path, "rb") as fd:
+            yield fd.read()
```
kristapratico, line 43: @marcusc2 Okay I took a closer look at the CI failures and it seems that the issue is still that the test.tar.gz file cannot be found. I think this is arising from the coretestserver being installed on the machine and then CI searching through the installed package path instead of what's here in azure-core (at least that's what it looks like from this path [here](https://dev.azure.com/azure-sdk/public/_build/results?buildId=1386309&view=logs&j=69247760-d00b-59ba-802c-4e283c193199&t=6654d4af-1717-52b7-cca2-b6c5ec5727ae&l=2605)).

I think an easy way to workaround this (and not have to commit the file to the repo) is to just create this file on the fly. The python standard library includes `gzip` which we can use to create the file, read from it, and then delete it after the test completes. Suggest trying something like this:

```
import gzip
with gzip.open('test.tar.gz', 'wb') as f:
    f.write(b"test")

with open(os.path.join(os.path.abspath('test.tar.gz')), "rb") as fd:
    yield fd.read()

os.remove("test.tar.gz")
```

sdk/core/azure-core/tests/testserver_tests/coretestserver/coretestserver/test_routes/streams.py
```diff
@@ -31,6 +33,24 @@ def streaming_test():
...
+        os.remove("test.tar.gz")
+    except:
+        with gzip.open('test.tar.gz', 'wb') as f:
+            f.write(b"test")
+
+        with open(os.path.join(os.path.abspath('test.tar.gz')), "rb") as fd:
+            yield fd.read()
+            
+        os.remove("test.tar.gz")
+
```
kristapratico, line 53: yay! it looks like this fixed the failing tests (other test failures are unrelated). One small thing - we don't need the try/except code anymore since we only added that before when we weren't sure what the expected path to the file was. We can remove the duplicate code here and just include lines L38-L44

sdk/core/azure-core/tests/testserver_tests/coretestserver/coretestserver/test_routes/streams.py
```diff
@@ -48,7 +59,11 @@ def string():
     return Response(
         streaming_test(), status=200, mimetype="text/plain"
     )
-    
+
+@streams_api.route('/compressed_no_header', methods=['GET'])
+def compressed_no_header():
+    return Response(stream_compressed_no_header(), status=300, headers={"Content-Encoding": "gzip"})
```
kristapratico, line 65: @marcusc2 On re-review of this, I noticed that this test is for compressed file with _no_ header so we shouldn't send the `Content-Encoding` header here. We do pass the arg `decompress=False` in the actual test so it's not actually looking at this header, but we should remove it so that it matches the intended test case scenario.

sdk/core/azure-core/tests/test_streaming.py
```diff
@@ -92,6 +92,22 @@ def test_decompress_compressed_no_header(http_request):
...
+    try:
+        request = http_request(method="GET", url="http://localhost:{}/streams/compressed_no_header".format(port))
+        pipeline_response = client._pipeline.run(request, stream=True)
+        response = pipeline_response.http_response
+        data = response.stream_download(client._pipeline, decompress=False)
+        content = b"".join(list(data))
+        with pytest.raises(UnicodeDecodeError):
+            decoded = content.decode('utf-8')
+    except HttpResponseError as e:
+        print(e.response.text())
```
iscai-msft, line 108: we shouldn't be catching the HttpResponseError because we're not expecting the service to throw an error, so here we're swallowing if we do get a network error.

Is the current code getting a network error?

iscai-msft (commented):
overall looks good! i worry that by catching the HttpResponseError we're swallowing potential errors that we shouldn't be having. Thanks!

xiangyan99 (commented):
Thanks for the contribution.

Do we need an async test too?

Or we plan to add it in a separate PR?

Thanks.

sdk/core/azure-core/tests/test_streaming.py
```diff
@@ -92,6 +92,22 @@ def test_decompress_compressed_no_header(http_request):
...
+
+@pytest.mark.live_test_only
```
iscai-msft, line 110: the async test is https://github.com/Azure/azure-sdk-for-python/blob/33c845aa7674c26bf851313daeae23e2f3f5459c/sdk/core/azure-core/tests/async_tests/test_streaming_async.py#L94

iscai-msft (approved):

PR #22723, opened 2022-01-21
[formrecognizer] use kwargs where we don't record tests

kristapratico:
same as https://github.com/Azure/azure-sdk-for-python/pull/22614

@iscai-msft

### DIFF
--- a/sdk/formrecognizer/azure-ai-formrecognizer/tests/test_ftc_compose_model.py
+++ b/sdk/formrecognizer/azure-ai-formrecognizer/tests/test_ftc_compose_model.py
@@ -59,7 +59,8 @@ class TestTraining(FormRecognizerTest):
 
     @FormRecognizerPreparer()
     @FormTrainingClientPreparer(client_kwargs={"api_version": "2.0"})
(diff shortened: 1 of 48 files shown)


sdk/formrecognizer/azure-ai-formrecognizer/tests/test_ftc_compose_model.py
```diff
@@ -59,7 +59,8 @@ def test_compose_model_invalid_unlabeled_models_v21(self, client, formrecognizer
 
     @FormRecognizerPreparer()
     @FormTrainingClientPreparer(client_kwargs={"api_version": "2.0"})
-    def test_compose_model_bad_api_version(self, client, formrecognizer_storage_container_sas_url, **kwargs):
```
catalinaperalta, line 62: Just double checking the storage container sas url is not used in this test right?

catalinaperalta: Nvm, just checked, thanks for removing it :)

catalinaperalta (approved):
LGTM. Thanks for doing this!

Issue #22489, opened 2022-01-13
[EngSys] CI Windows Python3.6 failure: Could not install packages due to an EnvironmentError: [WinError 145] The directory is not empty 


yunhaoling:
CI would sometimes fail because "Could not install packages due to an EnvironmentError: [WinError 145] The directory is not empty" on Windows Python 3.6. Retry could help, but it happens quite frequently.

[link to sample error from azure-servicebus CI](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=1293557&view=logs&j=29cbd47f-5b04-5aae-27ea-c5db931f8609&t=6f4236c0-e0ce-5979-2129-59b4ad74eed4&l=3877)

```
...
Successfully built idna-ssl
ERROR: astroid 2.5 has requirement wrapt<1.13,>=1.11, but you'll have wrapt 1.13.3 which is incompatible.
Installing collected packages: urllib3, idna, certifi, charset-normalizer, requests, six, azure-core, pycparser, cffi, cryptography, pywin32, portalocker, PyJWT, msal, msal-extensions, azure-identity, ConfigArgParse, pyyaml, wrapt, multidict, typing-extensions, yarl, vcrpy, azure-devtools, azure-common, azure-mgmt-core, oauthlib, requests-oauthlib, isodate, msrest, azure-mgmt-resource, azure-mgmt-keyvault, webencodings, pyparsing, packaging, bleach, docutils, Pygments, cmarkgfm, readme-renderer, wcwidth, py, attrs, atomicwrites, colorama, zipp, importlib-metadata, more-itertools, pluggy, pytest, pytest-asyncio, json-delta, python-dateutil, azure-storage-common, wheel, coverage, pytest-cov, python-dotenv, azure-mgmt-storage, pyOpenSSL, MarkupSafe, Jinja2, pytoml, azure-sdk-tools, adal, msrestazure, azure-mgmt-servicebus, async-timeout, asynctest, frozenlist, aiosignal, idna-ssl, aiohttp, pip, pytest-custom-exit-code, pytest-forked, execnet, pytest-xdist, smmap, gitdb, GitPython, pbr, stevedore, bandit, chardet, toml, isort, mccabe, typed-ast, lazy-object-proxy, astroid, pylint
  Attempting uninstall: wheel
    Found existing installation: wheel 0.34.2
    Uninstalling wheel-0.34.2:
[...]

PR #22746, opened 2022-01-25
Address astroid / wrapt installation error output

scbedd:
@mccoyp @kristapratico while this is not failing any builds at this point in time, we _are_ seeing some error output that isn't failing the builds.

[Example Build](https://dev.azure.com/azure-sdk/public/_build/results?buildId=1284955&view=logs&j=92b97d34-cf4d-5fc2-e574-c46d36e621e2&t=22fc1b08-c6b5-52f6-00fb-ed3e5e692f60&l=1323)

We've been seeing this error on and off in our builds, sometimes associated with another issue #22489 , sometimes on its own.

The origin of this error is two packages.

1. `vcrpy~=3.0.0` from azure-devtools setup.py -> `wrapt`
2. `pylint==2.5.2` from test_tools.txt. `pylint==2.5.2` -> `astroid==2.5` -> `wrapt >=1.12,<1.13`

Checking the [setup](https://github.com/kevin1024/vcrpy/blob/8245bd4f84146b2d144857b9446e467de4516796/setup.py#L27) for vcrpy, we can see that `wrapt` is actually totally unpinned. This means that it's an order of installation problem.

If `vcrpy` is installed first, `wrapt` will install at the latest, then during installation of `pylint`, `1.12.1` version actually is installed. This means that there is no conflict, and everything is fine.

If, however, `pylint` is installed first, `wrapt==1.11.1` is installed, then _supplanted_ by the `vcrpy` install. THIS is where the `ERROR` text pops up.  We lucked into a situation where we're not actually using any functionalities that break, which is why we don't see this associated with any failed builds. 

I'm not absolutely certain why we're triggering this logic, as its not consistent. IMO, it should recognize that there _is_ a `wrapt` installation already completed. I think we're hitting edge cases in the old pip==20.3 resolver.

By pinning `wrapt` to the version required by `astroid==2.5`, I believe we should be able to eliminate this warning.


eng/tox/tox.ini
```diff
@@ -41,6 +41,8 @@ pkgs =
 
 
 [testenv]
+commands_pre=
+    {envbindir}/python -m pip install pip==20.3
```
kristapratico, line 45: was this change intended in this PR?

scbedd: I don't know why it's popping in. I just rebased from main and submitted, and it's still showing up :|

scbedd: Fixed now.

kristapratico (approved):

PR #22790, opened 2022-01-26
[formrecognizer] Add a default for percent_completed

catalinaperalta:
Default percent_completed to 0. After talking to the service team, this should be defaulted to 0 when not returned. Generally our tests fail with operations that have a `notStarted` status since this is currently not returning a percent_completed value.

### DIFF
--- a/sdk/formrecognizer/azure-ai-formrecognizer/azure/ai/formrecognizer/_models.py
+++ b/sdk/formrecognizer/azure-ai-formrecognizer/azure/ai/formrecognizer/_models.py
@@ -3257,7 +3257,7 @@ class ModelOperationInfo(object):
     def __init__(self, **kwargs):
         self.operation_id = kwargs.get("operation_id", None)
         self.status = kwargs.get("status", None)
-        self.percent_completed = kwargs.get("percent_completed", None)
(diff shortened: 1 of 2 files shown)


sdk/formrecognizer/azure-ai-formrecognizer/azure/ai/formrecognizer/_models.py
```diff
@@ -3257,7 +3257,7 @@ class ModelOperationInfo(object):
     def __init__(self, **kwargs):
         self.operation_id = kwargs.get("operation_id", None)
         self.status = kwargs.get("status", None)
-        self.percent_completed = kwargs.get("percent_completed", None)
+        self.percent_completed = kwargs.get("percent_completed", 0)
```
kristapratico, line 3260: it might be worth adding this as a "bug fix" on the changelog

kristapratico (approved):

sdk/formrecognizer/azure-ai-formrecognizer/azure/ai/formrecognizer/_models.py
catalinaperalta: Good catch. Done!

PR #22792, opened 2022-01-27
[Cosmos] Change default consistency level to the database account consistency level

simorenoh:
These changes make it so the default consistency level that gets set when initializing clients becomes that of the database account, as opposed to the previous default (which was just "Session").

With these changes, the user can be sure that if they don't pass in a given consistency level their clients will run using the one present in their cosmos account. This also makes it easier for users seeking to try the dedicated gateway/ integrated cache since they won't have to specify for the clients to have a specific consistency when their account is set up to use these new features.


sdk/cosmos/azure-cosmos/README.md
```diff
@@ -463,7 +465,7 @@ async def create_items():
     await client.close() # the async client must be closed manually if it's not initialized in a with statement
 ```
 
-It is also worth pointing out that the asynchronous client has to be closed manually after its use, either by initializing it using async with or calling the close() method directly like shown above.
+However, instead of taking care of it manually with those extra lines, you can use the `async with` keywords. This creates a context manager that will warm up, initialize and later clean up and close the client once you're out of the statement. The example below shows how to start the client this way.
```
gahl-levy, line 468:
```suggestion
Instead of manually opening and closing the client, you can use the `async with` keywords. This creates a context manager that will initialize and later close the client once you're out of the statement. The example below shows how to do so.
```

sdk/cosmos/azure-cosmos/README.md
```diff
@@ -474,16 +476,17 @@ KEY = os.environ['ACCOUNT_KEY']
...
-    container = database.get_container_client(CONTAINER_NAME)
-    for i in range(10):
-        await container.upsert_item({
-                'id': 'item{0}'.format(i),
-                'productName': 'Widget',
-                'productModel': 'Model {0}'.format(i)
-            }
-        )
+async def create_products():
+    async with CosmosClient(URL, credential=KEY) as client: # the with statement will automatically warm up, initialize,# and close the async client
```
gahl-levy, line 480:
```suggestion
    async with CosmosClient(URL, credential=KEY) as client: # the with statement will automatically initialize and close the async client
```

sdk/cosmos/azure-cosmos/README.md
```diff
@@ -439,20 +439,22 @@ For more information on TTL, see [Time to Live for Azure Cosmos DB data][cosmos_
 ### Using the asynchronous client (Preview)
 
 The asynchronous cosmos client is a separate client that looks and works in a similar fashion to the existing synchronous client. However, the async client needs to be imported separately and its methods need to be used with the async/await keywords.
+Due to its asynchronous nature, it also needs to be warmed up and then closed down when used. The example below shows how by using the client's __aenter__() and close() methods.
```
gahl-levy, line 442:
```suggestion
The Async client needs to be initialized and closed after usage. The example below shows how to do so by using the client's __aenter__() and close() methods.
```

sdk/cosmos/azure-cosmos/azure/cosmos/_cosmos_client_connection.py
```diff
@@ -207,9 +198,44 @@ def __init__(
...
+        :type consistency_level: Optional[str]
+        :rtype: None
+        """
+        if consistency_level is None:
+            # Set to default level present in account
+            user_consistency_policy = database_account.ConsistencyPolicy
+            consistency_level = user_consistency_policy.get(constants._Constants.DefaultConsistencyLevel)
+
+        # Set consistency level header to be used for the client
+        self.default_headers[http_constants.HttpHeaders.ConsistencyLevel] = consistency_level
```
j82w, line 229: Should this throw if the user attempts to set the SDK level to greater than the account consistency?

sdk/cosmos/azure-cosmos/azure/cosmos/aio/_cosmos_client_connection_async.py
```diff
@@ -232,10 +224,54 @@ def _ReadEndpoint(self):
...
+        """Checks if consistency level param was passed in by user and sets it to that value or to the account default.
+        :param database_account: The database account to be used to check consistency levels
+        :type database_account: ~azure.cosmos.documents.DatabaseAccount
+        :rtype: None
+        """
+        # Set to default level present in account
+        user_consistency_policy = database_account.ConsistencyPolicy
+        consistency_level = user_consistency_policy.get(constants._Constants.DefaultConsistencyLevel)
+
+        self.default_headers[http_constants.HttpHeaders.ConsistencyLevel] = consistency_level
```
j82w, line 265: Should it be setting the consistency level header if it using the default account consistency level?

sdk/cosmos/azure-cosmos/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 ## Release History
 
+### 4.3.0b3 (Unreleased)
+
+### Bugs fixed
```
j82w, line 5: Should this be a 4.2.1 hotfix instead? Then if possible unlist/remove 4.2.0.

sdk/cosmos/azure-cosmos/azure/cosmos/_cosmos_client_connection.py
```diff
@@ -207,9 +198,44 @@ def __init__(
         # Routing map provider
         self._routing_map_provider = routing_map_provider.SmartRoutingMapProvider(self)
 
+        # Set header setting to lowest latency consistency level for the first request to _GetDatabaseAccount
+        self.default_headers[http_constants.HttpHeaders.ConsistencyLevel] = documents.ConsistencyLevel.Eventual
```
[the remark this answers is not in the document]
simorenoh: Good catch thanks, I thought setting a consistency level for the first request was necessary

simorenoh:
look into changing python code docs here too: https://docs.microsoft.com/en-us/azure/cosmos-db/sql/how-to-manage-consistency?tabs=portal%2Cdotnetv2%2Capi-async#override-the-default-consistency-level

sdk/cosmos/azure-cosmos/CHANGELOG.md
simorenoh: we'll be releasing this in the next beta so no worries on the versioning

sdk/cosmos/azure-cosmos/README.md
```diff
@@ -463,7 +465,7 @@ async def create_items():
     await client.close() # the async client must be closed manually if it's not initialized in a with statement
 ```
 
-It is also worth pointing out that the asynchronous client has to be closed manually after its use, either by initializing it using async with or calling the close() method directly like shown above.
+Instead of manually opening and closing the client, you can and should use the `async with` keywords. This creates a context manager that will initialize and later close the client once you're out of the statement. The example below shows how to do so.
```
kushagraThapar, line 468:
```suggestion
Instead of manually opening and closing the client, it is highly recommended to use the `async with` keywords. This creates a context manager that will initialize and later close the client once you're out of the statement. The example below shows how to do so.
```

sdk/cosmos/azure-cosmos/azure/cosmos/_cosmos_client_connection.py
```diff
@@ -210,6 +201,38 @@ def __init__(
...
+    ) -> None:
+        """Checks if consistency level param was passed in by user and sets it to that value or to the account default.
+
+        :param database_account: The database account to be used to check consistency levels
+        :type database_account: ~azure.cosmos.documents.DatabaseAccount
+        :param consistency_level: The consistency level passed in by the user
+        :type consistency_level: Optional[str]
+        :rtype: None
+        """
+        if consistency_level is None:
```
kushagraThapar, line 220: Currently there is no verification for Consistency level, shouldn't we add some validation to this. Otherwise this may result in unexpected behavior. 
If python has enums, it is best to use them here, and then use enums in internal implementation. Please see java code for more details.

sdk/cosmos/azure-cosmos/azure/cosmos/_cosmos_client_connection.py
```diff
@@ -210,6 +201,38 @@ def __init__(
...
+            # Set to default level present in account
+            user_consistency_policy = database_account.ConsistencyPolicy
+            consistency_level = user_consistency_policy.get(constants._Constants.DefaultConsistencyLevel)
+        else:
+            # Set consistency level header to be used for the client
+            self.default_headers[http_constants.HttpHeaders.ConsistencyLevel] = consistency_level
+
+        if consistency_level == documents.ConsistencyLevel.Session:
```
kushagraThapar, line 228: I see we have enums, best to validate it when taking the consistency level from user.

sdk/cosmos/azure-cosmos/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
...
+- Default consistency level for the sync and async clients is no longer "Session" and will instead be set to the 
+  consistency level of the user's cosmos account setting on initialization
```
kushagraThapar, line 7:
```suggestion
  consistency level of the user's cosmos account setting on initialization if not passed during client initialization. 
  This change will impact client application in terms of RUs and latency. Users relying on default `Session` consistency will need to pass it explicitly if their account consistency is different than `Session`. Please see [Consistency Levels in Azure Cosmos DB](https://docs.microsoft.com/en-us/azure/cosmos-db/consistency-levels) for more details.  
```

kushagraThapar (changes requested):

sdk/cosmos/azure-cosmos/azure/cosmos/_cosmos_client_connection.py
simorenoh: The service already throws an error if the users pass in strings that don't match any consistency level

sdk/cosmos/azure-cosmos/azure/cosmos/_cosmos_client_connection.py
simorenoh: what do you mean by verification? the backend will throw an error if the user attempts to pass in an invalid consistency

j82w (approved):

kushagraThapar (approved):

sdk/cosmos/azure-cosmos/azure/cosmos/_cosmos_client_connection.py
```diff
@@ -210,6 +201,39 @@ def __init__(
...
+            # create a session - this is maintained only if the default consistency level
+            # on the client is set to session, or if the user explicitly sets it as a property
+            # via setter
+            self.default_headers[http_constants.HttpHeaders.ConsistencyLevel] = consistency_level
```
xinlian12, line 232: why for session consistency, we always need to set the header?

sdk/cosmos/azure-cosmos/azure/cosmos/aio/_cosmos_client_connection_async.py
```diff
@@ -143,14 +144,6 @@ def __init__(
         # Keeps the latest response headers from server.
         self.last_response_headers = None
 
-        if consistency_level == documents.ConsistencyLevel.Session:
```
xinlian12, line 146: this comment is for line 141:
    if consistency_level is not None:
            self.default_headers[http_constants.HttpHeaders.ConsistencyLevel] = consistency_level

the above statements have been removed from sync client implementation, why not using similar logic here as well?

sdk/cosmos/azure-cosmos/azure/cosmos/aio/_cosmos_client_connection_async.py
```diff
@@ -232,10 +225,53 @@ def _ReadEndpoint(self):
         return self._global_endpoint_manager.get_read_endpoint()
 
     async def _setup(self):
```
xinlian12, line 227: is this a new function added? when it will called?

sdk/cosmos/azure-cosmos/azure/cosmos/aio/_cosmos_client_connection_async.py
```diff
@@ -232,10 +225,53 @@ def _ReadEndpoint(self):
...
+        # Save the choice that was made (either None or some value) and branch to set or get the consistency
+        if self.default_headers.get(http_constants.HttpHeaders.ConsistencyLevel):
+            user_defined_consistency = self.default_headers[http_constants.HttpHeaders.ConsistencyLevel]
+        else:
+            user_defined_consistency = None
+
+        if user_defined_consistency == documents.ConsistencyLevel.Session:
+            # create a Session if the user wants Session consistency
+            self.session = _session.Session(self.url_connection)
```
xinlian12, line 236: if the user defined consistency level is None, but account consistency is session, I thought we would still want to main a Session? 

Should we have the similar logic as in sync client?

sdk/cosmos/azure-cosmos/samples/examples_async.py
```diff
@@ -6,143 +6,157 @@
...
 async def examples_async():
     # All interaction with Cosmos DB starts with an instance of the CosmosClient
     # In order to use the asynchronous client, we need to use async/await keywords,
     # which can only be used within async methods like examples_async() here
 
+    # Since this is an asynchronous client, in order to properly use it you also have to warm it up and close it down.
+    # One way to do it would be like below (all of these statements would be necessary if you want to do it this way).
+
+    async_client = CosmosClient(url, key)
+    await async_client.__aenter__()
```
xinlian12, line 19: should we also add the "async with" here? (It mentioned in the readme)

sdk/cosmos/azure-cosmos/test/test_user_configs.py
```diff
@@ -46,5 +58,54 @@ def test_enable_endpoint_discovery(self):
         self.assertTrue(client_default.client_connection.connection_policy.EnableEndpointDiscovery)
         self.assertTrue(client_true.client_connection.connection_policy.EnableEndpointDiscovery)
 
+    def test_default_account_consistency(self):
```
xinlian12, line 61: Do we have similar test in async?

sdk/cosmos/azure-cosmos/azure/cosmos/aio/_cosmos_client_connection_async.py
```diff
@@ -232,10 +225,53 @@ def _ReadEndpoint(self):
...
+            database_account = await self._global_endpoint_manager._GetDatabaseAccount(
                 **self._setup_kwargs)
+            self._setup_kwargs['database_account'] = database_account
             await self._global_endpoint_manager.force_refresh(self._setup_kwargs['database_account'])
+        else:
+            database_account = self._setup_kwargs.get('database_account')
+
+        # Use database_account if no consistency passed in to verify consistency level to be used
+        if user_defined_consistency is None:
+            self._check_if_session_consistency(database_account)
```
xinlian12, line 251: if we first get the database account info, then the logic of handling session consistency can be combined? (similar to what you have done in sync client)

sdk/cosmos/azure-cosmos/azure/cosmos/_cosmos_client_connection.py
simorenoh: we need to set the header for session consistency because the way we check whether or not the session needs to be updated is through the header - the only place the Python SDK currently stores the given consistency level a user is using the header

sdk/cosmos/azure-cosmos/azure/cosmos/aio/_cosmos_client_connection_async.py
simorenoh: yeah, this function is an addition to the async client that is necessary in order to properly warm up the client - it gets called when a client gets initialized using `async with` or when a user calls `client.__aenter__()` to manually warm it up

sdk/cosmos/azure-cosmos/test/test_user_configs.py
simorenoh: async tests for the python sdk will come altogether in a later PR, we dont currently have any

sdk/cosmos/azure-cosmos/azure/cosmos/aio/_cosmos_client_connection_async.py
simorenoh: because we need to store the level of consistency the user is passing somewhere in order to do the checks later to see if starting a session is necessary

sdk/cosmos/azure-cosmos/azure/cosmos/aio/_cosmos_client_connection_async.py
simorenoh: yep you're right, I'll make those changes now

sdk/cosmos/azure-cosmos/azure/cosmos/aio/_cosmos_client_connection_async.py
xinlian12: gotcha, thanks for the explanation

xinlian12 (approved):
LGTM, thanks

Issue #22787, opened 2022-01-26
[TestProxy] tests should not have to return variables to record


tjprescott:
Discovered in #22187, when a tests uses `@recorded_by_proxy` but does not return anything (because it does not use the variables API), no recording is produced.

Tests should NOT have to return anything in order to be recorded. Per #22275, even using the variables API should not require a return statement.

PR #22799, opened 2022-01-27
Always return dictionary from tests

mccoyp:
# Description

Resolves https://github.com/Azure/azure-sdk-for-python/issues/22787. This is a patch-up for a bug discovered by @tjprescott. After the recent change to put `"x-recording-file"` in the body of recording start requests, tests have stopped producing recordings whenever they don't return a dictionary. Returning anything in a test should only be necessary for tests using the `variables` API.

While I work on a second, more user-friendly iteration of the `variables` API (and track down the code that produced this bug in the test proxy tool), this change will make it unnecessary for every test to return a dictionary in order to record.

Tagging those affected by the bug and @scbedd for notifs.

- _N/A_ Pull request includes test coverage for the included changes.
  - Tested manually with already-migrated tests

### DIFF
--- a/tools/azure-sdk-tools/devtools_testutils/aio/proxy_testcase_async.py
+++ b/tools/azure-sdk-tools/devtools_testutils/aio/proxy_testcase_async.py
@@ -69,14 +69,15 @@ def recorded_by_proxy_async(test_func):
         test_output = None
         try:
             try:
-                test_output = await test_func(*args, variables=variables, **trimmed_kwargs)
+                # tests don't record successfully unless test_output is a dictionary
+                test_output = await test_func(*args, variables=variables, **trimmed_kwargs) or {}
             except TypeError:
                 logger = logging.getLogger()
                 logger.info(
                     "This test can't accept variables as input. The test method should accept `**kwargs` and/or a "
                     "`variables` parameter to make use of recorded test variables."
                 )
-                test_output = await test_func(*args, **trimmed_kwargs)
+                test_output = await test_func(*args, **trimmed_kwargs) or {}
         except ResourceNotFoundError as error:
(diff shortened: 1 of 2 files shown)


catalinaperalta (approved):
LGTM. Thanks for adding this meanwhile you find the fix!

tools/azure-sdk-tools/devtools_testutils/aio/proxy_testcase_async.py
```diff
@@ -69,14 +69,15 @@ async def combined_call(*args, **kwargs):
         test_output = None
         try:
             try:
-                test_output = await test_func(*args, variables=variables, **trimmed_kwargs)
+                # tests don't record successfully unless test_output is a dictionary
+                test_output = await test_func(*args, variables=variables, **trimmed_kwargs) or {}
```
tjprescott, line 73: This is fine with me if it works, but I would be more curious to see where is the logic that is skipped when `test_output` is `None`? It seems to me that would be the more appropriate place for a fix.

mccoyp: That's a good point. We actually send `test_output` to the proxy in the [stop_record_or_playback](https://github.com/Azure/azure-sdk-for-python/blob/main/tools/azure-sdk-tools/devtools_testutils/proxy_testcase.py#L130) method, which would be a more appropriate place to fix this. I'll make that change in a moment

tjprescott: In that case, this is a good defense against wonkiness in the test-proxy

tjprescott (approved):

Issue #22516, opened 2022-01-14
[Schema Registry] implement AvroEncoder


swathipil:
implementation after finalizing design: #21606

Issue #21549, opened 2021-11-02
[SchemaRegistry] bump avro dependency version after dropping Python 2.7


swathipil:
bump avro version to 1.11.0

yunhaoling:
avro >= 1.11.0

in general we should not pin "=="

PR #22816, opened 2022-01-28
[SchemaRegistry] AvroEncoder

swathipil:
fixes: #22516
fixes: #21549 

Implementation for AvroEncoder as described in gist here: https://gist.github.com/swathipil/9eed26e7f85c6b57d9e788d2691c76de


sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
```diff
@@ -110,10 +112,33 @@ class EventData(object):
...
+    @overload
+    def __init__(self, body: Optional[Union[str, bytes, List[AnyStr]]] = None) -> None:
+        ...
+
+    def __init__(
+        self,
+        body: Optional[Union[str, bytes, List[AnyStr]]] = None,
+        *,
+        data: bytes = None,
+        content_type: str = None,
```
swathipil, line 128: @yunhaoling - Should we document that we take 'data' and 'content_type' as kwargs now? Do we want to keep this "private" for use in the encoder, since users don't really need to know about this?

swathipil: OOB: add to docstring + changelog

sdk/schemaregistry/azure-schemaregistry-avroencoder/azure/schemaregistry/encoder/avroencoder/_schema_registry_avro_encoder.py
```diff
@@ -0,0 +1,276 @@
...
+        return data, content_type
+
+    def decode(
+        self,
+        *,
+        message: Optional[MessageType] = None,
+        data: Optional[bytes] = None,
+        content_type: Optional[str] = None,
+        **kwargs,
+    ) -> Dict[str, Any]:
```
yunhaoling, line 224: it's just returning `Dict` or `MessageType` as well?

sdk/schemaregistry/azure-schemaregistry-avroencoder/azure/schemaregistry/encoder/avroencoder/_schema_registry_avro_encoder.py
```diff
@@ -0,0 +1,276 @@
...
+        """
+        Decode bytes data using schema ID in the content type field. One of the following is required:
+            1) `message`
+            2) `data` and `content_type`
+        Data must follow format of associated Avro RecordSchema:
+        https://avro.apache.org/docs/1.10.0/gettingstartedpython.html#Defining+a+schema
+
+        :param bytes value: The bytes data needs to be decoded.
```
yunhaoling, line 232: Is `value` needed anymore?

sdk/schemaregistry/azure-schemaregistry-avroencoder/azure/schemaregistry/encoder/avroencoder/aio/_schema_registry_avro_encoder_async.py
```diff
@@ -0,0 +1,269 @@
...
+        **kwargs,
+    ) -> Dict[str, Any]:
+        """
+        Decode bytes data using schema ID in the content type field. One of the following is required:
+            1) `message`
+            2) `data` and `content_type`
+        Data must follow format of associated Avro RecordSchema:
+        https://avro.apache.org/docs/1.10.0/gettingstartedpython.html#Defining+a+schema
+
+        :param bytes value: The bytes data needs to be decoded.
```
yunhaoling, line 226: is value needed?

sdk/schemaregistry/azure-schemaregistry-avroencoder/azure/schemaregistry/encoder/avroencoder/aio/_schema_registry_avro_encoder_async.py
```diff
@@ -0,0 +1,269 @@
...
+                        If using an Azure SDK model class, please check the README.md for the full list
+                        of supported Azure SDK models and their corresponding versions."""
+                ).raise_with_traceback()
+        if not data:
+            raise ValueError("'data' value cannot be None.")
+
+        # include in first preview for back compatibility
+        data, content_type = self._convert_preamble_format(data, content_type)
+        if not content_type:
+            raise ValueError("'content' type cannot be None.")
```
yunhaoling, line 258: nit: should be this?
```suggestion
            raise ValueError("'content_type' cannot be None.")
```

sdk/schemaregistry/azure-schemaregistry-avroencoder/samples/README.md
```diff
@@ -0,0 +1,59 @@
...
+[eventhub_send_integration_sample]:  https://github.com/Azure/azure-sdk-for-python/tree/main/sdk/schemaregistry/azure-schemaregistry-avroencoder/samples/sync_samples/eventhub_send_integration.py
+[eventhub_receive_integration_sample]:  https://github.com/Azure/azure-sdk-for-python/tree/main/sdk/schemaregistry/azure-schemaregistry-avroencoder/samples/sync_samples/eventhub_receive_integration.py
+[encode_and_decode_event_data_message_async_sample]: https://github.com/Azure/azure-sdk-for-python/tree/main/sdk/schemaregistry/azure-schemaregistry-avroencoder/samples/async_samples/encode_and_decode_event_data_message_async.py
+[eventhub_send_integration_async_sample]:  https://github.com/Azure/azure-sdk-for-python/tree/main/sdk/schemaregistry/azure-schemaregistry-avroencoder/samples/async_samples/eventhub_send_integration_async.py
+[eventhub_receive_integration_async_sample]:  https://github.com/Azure/azure-sdk-for-python/tree/main/sdk/schemaregistry/azure-schemaregistry-avroencoder/samples/async_samples/eventhub_receive_integration_async.py
+[api_reference]: https://docs.microsoft.com/python/api/
```
yunhaoling, line 59: it points to a general page.. is there a more schema registry specific one?

sdk/schemaregistry/azure-schemaregistry-avroencoder/samples/README.md
```diff
@@ -0,0 +1,59 @@
...
+    * Receive `EventData` from Event Hubs and decode the received bytes.
+
+## Prerequisites
+- Python 3.6 or later.
+- **Microsoft Azure Subscription:**  To use Azure services, including Azure Schema Registry, you'll need a subscription.
+If you do not have an existing Azure account, you may sign up for a free trial or use your MSDN subscriber benefits when you [create an account](https://account.windowsazure.com/Home/Index).
+
+## Setup
+
+1. Install the Azure Schema Registry Avro Encoder client library and Azure Identity client library for Python with [pip](https://pypi.org/project/pip/):
```
yunhaoling, line 32: I think mentioning that eventhub also needed to be installed to run sample code which uses `EventData` brings clarity

sdk/schemaregistry/azure-schemaregistry-avroencoder/azure/schemaregistry/encoder/avroencoder/aio/_schema_registry_avro_encoder_async.py
```diff
@@ -0,0 +1,269 @@
...
+        if data[0:RECORD_FORMAT_IDENTIFIER_LENGTH] == record_format_identifier:
+            schema_id = data[
+                SCHEMA_ID_START_INDEX : (SCHEMA_ID_START_INDEX + SCHEMA_ID_LENGTH)
+            ].decode("utf-8")
+            content_type = f"{AVRO_MIME_TYPE}+{schema_id}"
+            data = data[DATA_START_INDEX:]
+
+        return data, content_type
+
+    async def decode(
```
JoshLove-msft, line 211: We will probably need an update before GA to allow the user to pass in a different "reader" schema.

swathipil: makes sense! I'm planning to do that in a separate PR after this one is merged for this upcoming release.

sdk/schemaregistry/azure-schemaregistry-avroencoder/samples/README.md
swathipil: just doing this temporarily for this release, since the actual link for avroencoder is not valid yet. Will update after this release!

sdk/eventhub/azure-eventhub/CHANGELOG.md
```diff
@@ -1,9 +1,11 @@
 # Release History
 
-## 5.7.1 (Unreleased)
+## 5.9.0b1 (Unreleased)
 
 ### Features Added
 
+- `EventData` constructor now takes `data` and `content_type` as keyword args for interoperability with the Schema Registry Avro Encoder library, where `body` will be set to `data` if passed in.
```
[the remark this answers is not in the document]
swathipil: yeah that makes sense to me.

sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
```diff
@@ -110,10 +116,33 @@ class EventData(object):
...
+        **kwargs
+    ) -> None:
         self._last_enqueued_event_properties = {}  # type: Dict[str, Any]
         self._sys_properties = None  # type: Optional[Dict[bytes, Any]]
+
+        if kwargs:
+            raise TypeError(f"__init__() got an unexpected keyword argument '{list(kwargs.keys())[0]}'")
```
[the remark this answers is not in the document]
swathipil: is this valid? I thought the actual constructor needed to have the kwargs if the overloaded method has kwargs?

sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
```diff
@@ -180,6 +209,15 @@ def __str__(self):
         event_str += " }"
         return event_str
 
+    def __data__(self) -> bytes:
```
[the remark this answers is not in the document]
swathipil: was wondering if anyone else would use this except us internally with other SDKs, but raising an error makes sense to me.
(esp if we consider __data__ to be a part of the public API as defined by the protocol)

sdk/schemaregistry/azure-schemaregistry-avroencoder/CHANGELOG.md
```diff
@@ -0,0 +1,21 @@
+# Release History
+
+## 1.0.0b1 (Unreleased)
+
+This version and all future versions will require Python 3.6+. Python 2.7 is no longer supported.
+
+### Features Added
+
+- This package follows from the azure-schemaregistry-avroserializer.
```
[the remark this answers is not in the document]
swathipil: I believe that's what we're doing. I asked Laurent the same question, and he mentioned that avroserializer would be deprecated anyway when/if we update the `azure-schemaregistry` version in `setup.py`. So we wouldn't want to necessarily have code that would be broken here.

sdk/schemaregistry/azure-schemaregistry-avroencoder/azure/schemaregistry/encoder/avroencoder/aio/_schema_registry_avro_encoder_async.py
swathipil: so here `value` is the name of the parameter, following the same param name as in the serializer. We may want to discuss naming this data instead for the next release.

sdk/schemaregistry/azure-schemaregistry-avroencoder/azure/schemaregistry/encoder/avroencoder/_schema_registry_avro_encoder.py
swathipil: so here `value` is the name of the parameter, following the same param name as in the serializer. We may want to discuss naming this data instead for the next release.

sdk/schemaregistry/azure-schemaregistry-avroencoder/CHANGELOG.md
```diff
@@ -0,0 +1,21 @@
...
+- APIs have been updated to allow for encoding directly to and decoding from message type objects, where the data value is the Avro encoded payload.
+- The schema ID and format indicator are no longer prepended to the payload. Instead, the content type of the message will hold this information.
```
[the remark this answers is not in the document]
swathipil: makes sense to me!

sdk/schemaregistry/azure-schemaregistry-avroencoder/azure/schemaregistry/encoder/avroencoder/_schema_registry_avro_encoder.py
```diff
@@ -0,0 +1,276 @@
...
+        stream.write(data_bytes)
+        stream.flush()
+
+        payload = stream.getvalue()
+        stream.close()
+
+        if message_type:
+            try:
+                return message_type(data=payload, content_type=content_type, **kwargs)
+            except TypeError as e:
```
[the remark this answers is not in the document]
swathipil: I wasn't able to find any other errors that this would throw. These were the only scenarios that I accounted for, which both raise TypeErrors:
1) `message_type` is a class/object that isn't a callback, so this won't work.
2) `message_type` function call doesn't take the right params, so it throws a TypeError.

sdk/schemaregistry/azure-schemaregistry-avroencoder/azure/schemaregistry/encoder/avroencoder/_utils.py
```diff
@@ -0,0 +1,12 @@
...
+# Licensed under the MIT License. See License.txt in the project root for license information.
+# --------------------------------------------------------------------------------------------
+
+
+def get_http_request_kwargs(kwargs):
+    http_request_keywords = ["params", "headers", "json", "data", "files"]
+    http_request_kwargs = {
+        key: kwargs.pop(key, None) for key in http_request_keywords if key in kwargs
+    }
+    return http_request_kwargs
```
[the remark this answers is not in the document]
swathipil: yes, these are the only ones that we filter out. I took this from the `azure-schemaregistry` code, because we had this problem there as well. And there, I traced it back to what the `core` method would take.

sdk/schemaregistry/azure-schemaregistry-avroencoder/azure/schemaregistry/encoder/avroencoder/_schema_registry_avro_encoder.py
swathipil: `decode` just returns Dict, which is the decoded data. We don't return any of the message types here.

sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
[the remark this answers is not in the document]
swathipil: update: if we do this, it results in the following mypy error:
`error: Overloaded function implementation does not accept all possible arguments of signature 1`

I can add a `# type: ignore` to the constructor, but it seems like not a great solution. Thoughts?

sdk/schemaregistry/azure-schemaregistry-avroencoder/CHANGELOG.md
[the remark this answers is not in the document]
swathipil: I think that's the same reason why we want to do b1, so I'm good with changing this.

sdk/schemaregistry/azure-schemaregistry-avroencoder/azure/schemaregistry/encoder/avroencoder/_schema_registry_avro_encoder.py
```diff
@@ -0,0 +1,276 @@
...
+                SCHEMA_ID_START_INDEX : (SCHEMA_ID_START_INDEX + SCHEMA_ID_LENGTH)
+            ].decode("utf-8")
+            content_type = f"{AVRO_MIME_TYPE}+{schema_id}"
+            data = data[DATA_START_INDEX:]
+
+        return data, content_type
+
+    def decode(
+        self,
+        *,
```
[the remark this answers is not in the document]
swathipil: I removed it from the docstring and `value` is replaced with `data`. My thoughts were to keep `value` in `encode` b/c we decided on it before. However, for decode, I was thinking we wanted to have `data` instead of `value` to be consistent with the output of encode.
For ex.:
```python
# returns {'data': '...', 'content_type': ...}
metadata_dict = AvroEncoder.encode(value={'x': "val"}...)

# then data/content_type would correspond to the kwargs that `decode` accepts
decoded_data = AvroEncoder.decode(**metadata_dict)
```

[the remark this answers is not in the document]
swathipil: I can change `value` to `data` in encode too?

sdk/schemaregistry/azure-schemaregistry-avroencoder/samples/async_samples/encode_with_callback_async.py
```diff
@@ -0,0 +1,88 @@
...
+async def encode_with_callback(encoder):
+
+    # Callback MUST have these parameters in the following order: (data, content_type, **kwargs)
+    def sample_create_event_data(data, content_type, **kwargs):
+        print("Creating sample EventData with callback.")
+        return EventData(data=data, content_type=content_type, **kwargs)
+
+    dict_data_ben = {"name": "Ben", "favorite_number": 7, "favorite_color": "red"}
+    event_data_ben = await encoder.encode(
+        dict_data_ben, schema=SCHEMA_STRING, message_type=sample_create_event_data
```
deyaaeldeen, line 67: I am not sure if it is just me but `message_type=<callback>` feels weird

sdk/schemaregistry/azure-schemaregistry-avroencoder/samples/async_samples/encode_with_callback_async.py
```diff
@@ -0,0 +1,88 @@
...
+
+async def encode_with_callback(encoder):
+
+    # Callback MUST have these parameters in the following order: (data, content_type, **kwargs)
+    def sample_create_event_data(data, content_type, **kwargs):
+        print("Creating sample EventData with callback.")
+        return EventData(data=data, content_type=content_type, **kwargs)
+
+    dict_data_ben = {"name": "Ben", "favorite_number": 7, "favorite_color": "red"}
+    event_data_ben = await encoder.encode(
```
deyaaeldeen, line 66: you're not planning on using the name `encodeMessageData`?

sdk/schemaregistry/azure-schemaregistry-avroencoder/azure/schemaregistry/encoder/avroencoder/_schema_registry_avro_encoder.py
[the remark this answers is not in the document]
yunhaoling: I'm thinking of the case that user has their check and raises their own error in the callback/constructor like the following:

```python
def message_type(data, content_type, **kwargs):
    if kwargs.get("xxx") != "yyy":
        raise CustomizedErrorDefinedByUser("something wrong")
    return SomeData(...)
```

sdk/schemaregistry/azure-schemaregistry-avroencoder/azure/schemaregistry/encoder/avroencoder/_schema_registry_avro_encoder.py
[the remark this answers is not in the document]
yunhaoling: I don't have a strong preference here, `data` works for me.
I just feel flipping between different terms (data in decode and value in encode) makes the api a bit confusing to me..

I tend to use the same term, probably worthwhile to check how json/avro lib name the parameter.
also what about other languages, do we align the terminology "data"?

sdk/schemaregistry/azure-schemaregistry-avroencoder/azure/schemaregistry/encoder/avroencoder/_schema_registry_avro_encoder.py
yunhaoling: then we can remove this docstring for `value`, right? since it's not being used

sdk/schemaregistry/azure-schemaregistry-avroencoder/samples/async_samples/encode_with_callback_async.py
swathipil: yes, so Anna mentioned that we should prob go with `encode`. In this case, I think the shorter version is fine because the method itself is pretty straightforward about what is being encoded. (I.E. `encode` is clear so no need to `message_data`)

sdk/schemaregistry/azure-schemaregistry-avroencoder/samples/async_samples/encode_with_callback_async.py
swathipil: yes, I definitely agree with this. I originally just had only `MessageType` as the type, so this felt reasonable. I think we'll change this. I proposed `message_callback_type` in the APIView since that feels inclusive of `message_type`.

Will see if anyone has thoughts during Arch Board review + can also change before GA.

sdk/schemaregistry/azure-schemaregistry-avroencoder/azure/schemaregistry/encoder/avroencoder/_utils.py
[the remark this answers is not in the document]
yunhaoling: I looked into the code of `azure-schemaregirtry`:

- get_http_request_kwargs is to extract the arguments used by building a `request`
- other kwargs are kept for `send_request` method

```python
    def get_schema(self, schema_id, **kwargs):
        # type: (str, Any) -> Schema
        http_request_kwargs = get_http_request_kwargs(kwargs)
        request = schema_rest.build_get_by_id_request(
            id=schema_id, **http_request_kwargs
        )
        response = self._generated_client.send_request(request, **kwargs)
        response.raise_for_status()
        return _parse_response_schema(response)
```

I would you not need to duplicate this filter process in the avro encoder since it's taken care by `azure-schemaregistry` already, besides it's dropping the kwargs for `send_request`.

or did you encounter a real error not doing such?

sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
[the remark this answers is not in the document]
yunhaoling: if mypy/pylint doesn't complain, then I'm good with removing it.

taking **kwargs could be incremental

sdk/schemaregistry/azure-schemaregistry-avroencoder/azure/schemaregistry/encoder/avroencoder/_schema_registry_avro_encoder.py
swathipil: also changing this param to `data` for consistency + that's what .NET is calling it in their `encode`

sdk/schemaregistry/azure-schemaregistry-avroencoder/azure/schemaregistry/encoder/avroencoder/_schema_registry_avro_encoder.py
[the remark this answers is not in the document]
swathipil: in this case, wouldn't we just want this error to be raised as normal? would we catch this?

sdk/schemaregistry/azure-schemaregistry-avroencoder/azure/schemaregistry/encoder/avroencoder/_schema_registry_avro_encoder.py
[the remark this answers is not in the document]
swathipil: same term sounds good to me as well. .NET uses `data` in their encode method.

sdk/schemaregistry/azure-schemaregistry-avroencoder/azure/schemaregistry/encoder/avroencoder/_utils.py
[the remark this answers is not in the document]
swathipil: ah good catch, this is a bug here: https://github.com/Azure/azure-sdk-for-python/blob/f363c224e41fc2878ff595901451c0497941b4c7/sdk/schemaregistry/azure-schemaregistry/azure/schemaregistry/_schema_registry_client.py#L141

and send_request takes the following kwargs:
`stream`: https://github.com/Azure/azure-sdk-for-python/blob/f363c224e41fc2878ff595901451c0497941b4c7/sdk/core/azure-core/azure/core/_pipeline_client.py#L189
`_return_pipeline_response`: https://github.com/Azure/azure-sdk-for-python/blob/f363c224e41fc2878ff595901451c0497941b4c7/sdk/core/azure-core/azure/core/_pipeline_client.py#L194
`transport`/`policies`/`per_call_policies`/`per_retry_policies`: https://github.com/Azure/azure-sdk-for-python/blob/f363c224e41fc2878ff595901451c0497941b4c7/sdk/core/azure-core/azure/core/_pipeline_client.py#L114
and I think a bunch more from inside `_build_pipeline` method.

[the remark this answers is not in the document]
swathipil: `request_id`/`auto_request_id`/`response_encoding`/`network_span_namer`/`tracing_attributes`/`session`/`session_owner`/`use_env_settings` are all things here. I'm not sure if we document these as reserved words anywhere.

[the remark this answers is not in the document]
swathipil: OOB: not passing kwargs when calling schema_registry_client methods for now

sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
```diff
@@ -110,10 +116,33 @@ class EventData(object):
...
+        **kwargs
+    ) -> None:
         self._last_enqueued_event_properties = {}  # type: Dict[str, Any]
         self._sys_properties = None  # type: Optional[Dict[bytes, Any]]
+
+        if kwargs:
+            raise TypeError(f"__init__() got an unexpected keyword argument '{list(kwargs.keys())[0]}'")
+
+        if data:
+            if body:
```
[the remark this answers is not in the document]
swathipil: I think if we combine, like this:
```
if data and body:
  warnings.warn('overwrite')
  body = data
```
then it will not work if `body=None` and `data='x'`. body ends up as None. If we do this:
```
if data and body:
  warnings.warn(overwrite)
body = data
```
it will not work for `data=None` and `body='x'`.  body ends up as None.

Sorry, maybe I'm missing something obvious though? 😕

sdk/schemaregistry/azure-schemaregistry-avroencoder/azure/schemaregistry/encoder/avroencoder/_schema_registry_avro_encoder.py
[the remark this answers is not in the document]
yunhaoling: you're right, we're only catching TypeError here -- I'm dumb, nvm

sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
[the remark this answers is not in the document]
yunhaoling: could be the following?

```python
if data and body:
  warnings.warn(overwrite)
body = data or body
```

sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
```diff
@@ -17,7 +17,9 @@
     List,
     TYPE_CHECKING,
     cast,
+    overload
 )
+import warnings
```
yunhaoling, line 22: I don't recall that we use `warnings` in our SDKs -- we usually use logger.warning if I remember correctly.

sdk/schemaregistry/azure-schemaregistry-avroencoder/azure/schemaregistry/encoder/avroencoder/aio/_schema_registry_avro_encoder_async.py
yunhaoling: log an issue for Josh's comment?

swathipil: I actually just decided to add the reader schema to this PR. Not a big change + quicker to add here since CC is tomorrow.

sdk/schemaregistry/azure-schemaregistry-avroencoder/samples/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+```bash
+pip install azure-schemaregistry-avroencoder azure-identity
+```
+
+Additionally, if using with `azure.eventhub.EventData`, install `azure-eventhub>=5.9.0b1`:
+
+```bash
+pip install azure-eventhub>=5.9.0b1
```
yunhaoling, line 41: nit: I would pin to version ==5.9.0b1 since EH is a preview, maybe there is new stable feature coming in 5.10 before the GAing protocol.
```suggestion
pip install azure-eventhub==5.9.0b1
```

sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
swathipil: we use warnings.warn in a few places, for ex: https://github.com/Azure/azure-sdk-for-python/blob/f90387f693220527a9f8eb2408749bac94bfc458/sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_sender.py#L307

I can change it, but I think this works too.

swathipil: aligning with rest of EH by using _LOGGER.warning

sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
```diff
@@ -98,6 +99,10 @@ class EventData(object):
 
     :param body: The data to send in a single message. body can be type of str or bytes.
     :type body: str or bytes
+    :keyword data: The data to send in a single message. If set, will overwrite body argument.
+    :paramtype data: bytes or None
+    :keyword content_type: The content type descriptor.
+    :paramtype content_type: str or None
```
swathipil, line 105: remove

sdk/schemaregistry/azure-schemaregistry-avroencoder/README.md
```diff
@@ -129,45 +129,59 @@ definition = """
...
+
+    # OR 
 
-with serializer:
-    encoded_bytes = b'<data_encoded_by_azure_schema_registry_avro_serializer>'
-    decoded_data = serializer.deserialize(encoded_bytes)
+    encoded_bytes = b'<data_encoded_by_azure_schema_registry_avro_encoder>'
+    content_type = 'avro/binary+<schema_id_of_corresponding_schema>'
+    data_dict = {"data": encoded_bytes, "content_type": content_type}
+    decoded_data = encoder.decode(encoded_bytes, message=data_dict)
```
swathipil, line 173: `encoder.decode(data_dict)

sdk/schemaregistry/azure-schemaregistry-avroencoder/azure/schemaregistry/encoder/avroencoder/aio/_schema_registry_avro_encoder_async.py
```diff
@@ -0,0 +1,277 @@
...
+        self,
+        message: Union[MessageType, MessageMetadataDict],
+        *,
+        readers_schema: Optional[str] = None,
```
yunhaoling, line 221: plural or singular? -- reader_schema or readers_schema?

sdk/schemaregistry/azure-schemaregistry-avroencoder/samples/README.md
```diff
@@ -0,0 +1,65 @@
...
+1. Install the Azure Schema Registry Avro Encoder client library and Azure Identity client library for Python with [pip](https://pypi.org/project/pip/):
+
+```bash
+pip install azure-schemaregistry-avroencoder azure-identity
+```
+
+Additionally, if using with `azure.eventhub.EventData`, install `azure-eventhub==5.9.0b1`:
+
+```bash
+pip install azure-eventhub>=5.9.0b1
```
yunhaoling, line 41: haha, I saw you added "install `azure-eventhub==5.9.0b1`",
should this bash command also be updated?
```suggestion
pip install azure-eventhub==5.9.0b1
```

sdk/schemaregistry/azure-schemaregistry-avroencoder/azure/schemaregistry/encoder/avroencoder/_schema_registry_avro_encoder.py
```diff
@@ -0,0 +1,285 @@
...
+            content_type = f"{AVRO_MIME_TYPE}+{schema_id}"
+            data = data[DATA_START_INDEX:]
+
+        return data, content_type
+
+    def decode(
+        self,
+        message: Union[MessageType, MessageMetadataDict],
+        *,
+        readers_schema: Optional[str] = None,
```
yunhaoling, line 227: should we mention this in changelog?

yunhaoling, line 227: reader or reader**s** (plural)

sdk/schemaregistry/azure-schemaregistry-avroencoder/azure/schemaregistry/encoder/avroencoder/aio/_schema_registry_avro_encoder_async.py
swathipil: plural as per Apache Avro: https://github.com/apache/avro/blob/fa443cf8c12b3f4cddf85c4a28f6b6aab68a148f/lang/py/avro/io.py#L614

sdk/schemaregistry/azure-schemaregistry-avroencoder/samples/README.md
swathipil: 🤦

sdk/schemaregistry/azure-schemaregistry-avroencoder/azure/schemaregistry/encoder/avroencoder/_schema_registry_avro_encoder.py
swathipil: I thought about this and decided to not include it in the changelog, especially considering that all the other params changed as well and we didn't mention those.

I can put it in if you feel strongly.

swathipil: plural as per Apache Avro implementation

swathipil: oh, also just checked the other language changelogs and they don't mention this.

deyaaeldeen: I do mention it in JS: https://github.com/Azure/azure-sdk-for-js/blob/main/sdk/schemaregistry/schema-registry-avro/CHANGELOG.md:
> [](https://github.com/Azure/azure-sdk-for-js/blob/main/sdk/schemaregistry/schema-registry-avro/CHANGELOG.md#breaking-changes)decodeMessageData now supports decoding using a different but compatible schema

but I am happy to change it if we get consensus. /cc @JoshLove-msft

swathipil: ah gotcha, thanks for the link - I was looking for the specific param name.

sdk/schemaregistry/azure-schemaregistry-avroencoder/README.md
```diff
@@ -44,71 +44,71 @@ pip install azure-identity
...
 
-### Message format
-
-The same format is used by schema registry serializers across Azure SDK languages.
-
-Messages are encoded as follows:
+### Supported message models
 
-- 4 bytes: Format Indicator
+Support has been added to certain Azure Messaging SDK model classes for interoperability with the `AvroEncoder`. These models are subtypes of the `MessageType` protocol defined under the `azure.schemaregistry.encoder.avroencoder` namespace. Currently, the supported model classes are:
```
yunhaoling, line 71: nit: I feel we could copy paste the protocol definition code into this section?

sdk/schemaregistry/azure-schemaregistry-avroencoder/azure/schemaregistry/encoder/avroencoder/_schema_registry_avro_encoder.py
swathipil: If we decide to do this, I can just update it in another docs PR later. Going to merge this one for now.

yunhaoling (approved):
LGTM!

sdk/schemaregistry/azure-schemaregistry-avroencoder/README.md
swathipil: will fix in separate docs PR

PR #22848, opened 2022-01-31
Update tests.md

tjprescott:
Updates to the `docs/dev/tests.md` file to account for changes in how tests are written and run.

Anyone is encouraged to commit additional changes.

### DIFF
--- a/doc/dev/tests.md
+++ b/doc/dev/tests.md
@@ -1,19 +1,23 @@
-# Setup Python Development Environment
-In this document we will provide the introduction to the testing framework by:
+# Python SDK testing guide
 
-- [Setting up your development environment](#setup-the-development-environment)
-- [Integrating with pytest](#integrate-with-the-pytest-test-framework)
-- [Using Tox](#tox)
+This guide walks through the setup necessary to run tests in the Azure SDK for Python, gives an overview of the shared
+testing infrastructure, and demonstrates how to write and run tests for a service.
+
+### Table of contents
+
+- [Set up your development environment](#set-up-your-development-environment)
+- [Integrate with pytest](#integrate-with-the-pytest-test-framework)
+- [Use Tox](#tox)
 - [The `devtools_testutils` package](#devtools_testutils-package)
-- [Writing New Tests](#writing-new-tests)
-- [Define our credentials and settings](#define-credentials)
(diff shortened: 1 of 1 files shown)


doc/dev/tests.md
```diff
@@ -128,8 +126,11 @@ A quick description of the five commands above:
 ## `devtools_testutils` Package
 The Azure SDK team has created some in house tools to help with easier testing. These additional tools are located in the `devtools_testutils` package that was installed with your `dev_requirements.txt`. In this package is the [`AzureTestCase`](https://github.com/Azure/azure-sdk-for-python/blob/main/tools/azure-sdk-tools/devtools_testutils/azure_testcase.py#L99-L350) object which every test case object should inherit from. This management object takes care of creating and scrubbing recordings to make sure secrets are not added to the recordings files (and subsequently to the git history) and authenticating clients for test methods.
```
kristapratico, line 127: I'm wondering if we should move the sentences about test case objects inheriting from `AzureTestCase` to where we talk about "older SDK tests". Otherwise this guidance doesn't mix with test proxy. I can make the change if you agree

doc/dev/tests.md
```diff
@@ -15,11 +15,10 @@ In this document we will provide the introduction to the testing framework by:
 
 ## Setup your development environment
 
-The Azure SDK Python team creates libraries that are compatible with Python 2.7 and 3.5 and up. We will set up working Python virtual environments for Python 2.7, 3.5, and 3.9. It is recommended to do your development work in Python3, however it is helpful to have virtual environments for other versions to make debugging PRs easier locally.
+The Azure SDK Python team creates libraries that are compatible with Python 3.6 and up. We will set up working Python virtual environments for Python 3.6, and 3.9. It is recommended to do your development work in Python3, however it is helpful to have virtual environments for other versions to make debugging PRs easier locally.
```
johanste, line 18: Is there a specific reason we don't recommend Python 3.10?

doc/dev/tests.md
tjprescott: My thought is that a lot of McCoy's test proxy stuff should go here.  @mccoyp do you agree?

doc/dev/tests.md
tjprescott: Not specifically no.

PR #22871, opened 2022-02-02
[Storage] Updates from arch review for STG81

jalauzon-msft:


sdk/storage/azure-storage-blob/azure/storage/blob/_models.py
```diff
@@ -925,8 +925,8 @@ class ContainerSasPermissions(object):
         To enable operations related to set/delete immutability policy.
         To get immutability policy, you just need read permission.
     """
-    def __init__(self, read=False, write=False, delete=False, delete_previous_version=False,
-                 list=False, tag=False, **kwargs):  # pylint: disable=redefined-builtin
+    def __init__(self, read=False, write=False, delete=False,
+                 list=False, delete_previous_version=False, tag=False, **kwargs):  # pylint: disable=redefined-builtin
```
jalauzon-msft, line 929: I'm moving this back to its original position.

tasherif-msft (approved):

PR #22877, opened 2022-02-02
Disable apistub gen artifact gen for python pipeline

praveenkuttappan:
Python APIStub generator has a new change that seems to have broken apistubgen for azure-core-tracing-opencensus. This PR is to temporarily remove this apistub gen step since we haven't rolled out change to utilize this artifact yet.

### DIFF
--- a/eng/pipelines/templates/steps/build-artifacts.yml
+++ b/eng/pipelines/templates/steps/build-artifacts.yml
@@ -95,9 +95,9 @@ steps:
         --service="${{ parameters.ServiceDirectory }}" 
         --toxenv=sphinx
 
-  - template: /eng/pipelines/templates/steps/run_apistub.yml
-    parameters:
-      ServiceDirectory: ${{ parameters.ServiceDirectory }}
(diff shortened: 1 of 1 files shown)


weshaggard:
@praveenkuttappan instead of disabling is it easy enough to rollback the version of the apistub tool we use? Also is there any way to try to test/check for these breaks when new changes are made to apistub?

praveenkuttappan:
We will add a pipeline to test this tool against all python packages.

scbedd (approved):

sima-zhu (approved):
The code is not super complicated. Better to remove or roll back.

praveenkuttappan:
> @praveenkuttappan instead of disabling is it easy enough to rollback the version of the apistub tool we use? Also is there any way to try to test/check for these breaks when new changes are made to apistub?

@tjprescott has submitted a PR to fix actual issue in apistubgen. This PR is to unblock any pipeline failures until that's merged.

praveenkuttappan:
This is a temporary change to unblock python team.

PR #22882, opened 2022-02-02
[translation] add keywords to assist in searching for the package in pypi

kristapratico:


catalinaperalta (approved):
Love this! Also, you'll know better for this package, but maybe also add `translation` or `translate`?

kristapratico:
> Love this! Also, you'll know better for this package, but maybe also add `translation` or `translate`?

Initially I was thinking to stick with product names, but I don't think it could hurt to try some words which describe what the package does. Will add.

PR #22883, opened 2022-02-02
Revert changes to disable apistubgen

praveenkuttappan:
Revert temporary change to disable apistubgen and create stub file for packages.


praveenkuttappan:
@scbedd Travis has merged fix in api stub generator to resolve issue we were getting earlier for azure-core-tracing-opencensus package. This PS is to revert my temporary change to disable api stub generator.

tjprescott (approved):

tjprescott:
Does the fact that this passed mean that when it's merged things won't be broken?

tjprescott:
@praveenkuttappan sorry I broke this. Here's the fix: https://github.com/Azure/azure-sdk-tools/pull/2763

Issue #22550, opened 2022-01-19
[AMQP Python] Event Hubs feature parity with custom_endpoint_address and connection_verify


yunhaoling:
The two feature `custom_endpoint_address, connection_verify` are temporarily pulled out, we should get them back.

PR #22884, opened 2022-02-02
[EH Pyproto] Async send and receive

yunhaoling:


sdk/eventhub/azure-eventhub/tests/livetest/asynctests/test_send_async.py
```diff
@@ -214,8 +233,8 @@ async def test_send_non_ascii_async(connstr_receivers):
     # receive_message_batch() returns immediately once it receives any messages before the max_batch_size
     # and timeout reach. Could be 1, 2, or any number between 1 and max_batch_size.
     # So call it twice to ensure the two events are received.
-    partition_0 = [EventData._from_message(x) for x in receivers[0].receive_message_batch(timeout=5000)] + \
-                  [EventData._from_message(x) for x in receivers[0].receive_message_batch(timeout=5000)]
+    partition_0 = [EventData._from_message(x) for x in receivers[0].receive_message_batch(timeout=5)] + \
```
swathipil, line 236: wondering why you reduced these + other timeouts?

yunhaoling: the time unit in python amqp is seconds instead of miliseconds now

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_encode.py
```diff
@@ -593,7 +593,7 @@ def encode_filter_set(value):
 
 
 def encode_unknown(output, value, **kwargs):
-    # type: (bytearray, Optional[Any], Any) -> None
+    # type: (bytearray, Optional[Any]) -> None
```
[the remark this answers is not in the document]
yunhaoling: thanks for pointing it out, however, I probably won't fix mypy/pylint issues in this PR, there are too many them -- I would love to have a separate PR dedicated for python3 mypy style type hint.

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_cbs_async.py
```diff
@@ -0,0 +1,218 @@
...
+        self._refresh_window = None
+
+        self._token_status_code = None
+        self._token_status_description = None
+
+        self.state = CbsState.CLOSED
+        self.auth_state = CbsAuthState.Idle
+
+    async def _put_token(self, token, token_type, audience, expires_on=None):
+        # type: (str, str, str, datetime) -> None
```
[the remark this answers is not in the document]
swathipil: will be fixed in a separate PR as per comment above.

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_client_base_async.py
```diff
@@ -194,22 +189,12 @@ async def _create_auth_async(self) -> authentication.JWTTokenAsync:
...
                 self._auth_uri,
                 functools.partial(self._credential.get_token, self._auth_uri),
-                token_type=token_type,
-                timeout=self._config.auth_timeout,
-                http_proxy=self._config.http_proxy,
-                transport_type=self._config.transport_type,
-                custom_endpoint_hostname=self._config.custom_endpoint_hostname,
-                port=self._config.connection_port,
-                verify=self._config.connection_verify,
-                refresh_window=300,
```
annatisch, line 208: We don't need all these settings?

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_client_base_async.py
```diff
@@ -287,19 +271,25 @@ async def _management_request_async(self, mgmt_msg: Message, op_type: bytes) ->
...
+                        description="Management authentication failed. Status code: {}, Description: {!r}".format(
+                            status_code,
+                            description
                         )
                     )
                 if status_code in [404]:
-                    raise ConnectError(
-                        "Management connection failed. Status code: {}, Description: {!r}".format(
-                            status_code, description
+                    raise errors.AMQPConnectionError(
```
annatisch, line 281: Is this raised to customers? Would that be breaking? Or do we handle this internally?

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_client_base_async.py
```diff
@@ -480,10 +466,11 @@ async def _close_connection_async(self) -> None:
         await self._client._conn_manager_async.reset_connection_if_broken()  # pylint:disable=protected-access
 
     async def _handle_exception(self, exception: Exception) -> Exception:
-        if not self.running and isinstance(exception, compat.TimeoutException):
-            exception = errors.AuthenticationException("Authorization timeout.")
-            return await _handle_exception(exception, self)
-
+        if not self.running and isinstance(exception, TimeoutError):
+            exception = errors.AuthenticationException(
```
annatisch, line 470: Why would we be raising pyamqp errors from EH code? Or is this not exposed to customers?

sdk/eventhub/azure-eventhub/tests/livetest/asynctests/test_auth_async.py
```diff
@@ -33,7 +33,7 @@ async def test_client_secret_credential_async(live_eventhub):
         batch.add(EventData(body='A single message'))
         await producer_client.send_batch(batch)
 
-    def on_event(partition_context, event):
+    async def on_event(partition_context, event):
```
annatisch, line 36: Was this test always broken?

sdk/eventhub/azure-eventhub/tests/livetest/asynctests/test_auth_async.py
```diff
@@ -45,6 +45,7 @@ def on_event(partition_context, event):
     assert on_event.called is True
     assert on_event.partition_id == "0"
     assert list(on_event.event.body)[0] == 'A single message'.encode('utf-8')
+    await credential.close()
```
annatisch, line 48: I didn't have to do this before?

sdk/eventhub/azure-eventhub/tests/livetest/asynctests/test_auth_async.py
```diff
@@ -74,7 +75,7 @@ async def test_client_sas_credential_async(live_eventhub):
         await producer_client.send_batch(batch)
 
     # Finally let's do it with SAS token + conn str
-    token_conn_str = "Endpoint=sb://{}/;SharedAccessSignature={};".format(hostname, token.decode())
```
annatisch, line 77: Would a customer ever interact with this `token`?

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_client_base_async.py
yunhaoling: there're advanced feature which are not supported in the python amqp yet.
I have created an issue for it: https://github.com/Azure/azure-sdk-for-python/issues/22550

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_client_base_async.py
yunhaoling: it's not exposed to customers and is handled within retry, we raise amqp errors because we would like to reuse current retry logic which takes action (close connection) based on amqp error type.

sdk/eventhub/azure-eventhub/tests/livetest/asynctests/test_auth_async.py
yunhaoling: yea.. the funny thing is that the test passed because the sync callback is executed.
in the while, await a sync call logs a error/warning if on_error is not callback which is the expected behavior..

I will make a PR to update the test case

sdk/eventhub/azure-eventhub/tests/livetest/asynctests/test_auth_async.py
yunhaoling: it's still not needed, I add it to avoid warning.

when I was testing this case, I was warning logs like "transport pipeline session not closed" which indicates the aio credential is not closed. so I add it

sdk/eventhub/azure-eventhub/tests/livetest/asynctests/test_auth_async.py
yunhaoling: when you say interact, do you string manipulation or just moving the whole token string around?

a customer may pass the token string to downstream for distribution, but as far as I know they should not change the token string itself which would make the token invalid.

annatisch: I ask because it looks like `token` has changed from bytes to str - so wondering if that's breaking that I no longer needed to call `token.decode()`

sdk/eventhub/azure-eventhub/tests/livetest/asynctests/test_auth_async.py
annatisch: Gotcha. So a customer would not normally need to call `close` on their credentials - this would be in the client?

yunhaoling: the event hub client won't call `credential.close` internally.
the life cycle of the `credential` is managed by users themselves.

sdk/eventhub/azure-eventhub/tests/livetest/asynctests/test_auth_async.py
yunhaoling: yea.. from "bytes to str" perspective, it's a breaking change..

I know our credential `AccessToken` protocol says we should return string, so we were wrong at the beginning.
I could update the code to return bytes instead of string in EH to make it non-breaking.

annatisch: Yeah if this is customer facing we'll probably have to do that.

PR #22893, opened 2022-02-03
[Cosmos] move system_key logic to query_iterable

simorenoh:
This is a problem that came up while we were having the big conversations around the async client in December, with the network call that is made through the `set_partition_key()` method that was present in most container methods. It seems as though the changes with moving the await to the QueryIterableAsync class' __aiter__ method did not end up working, and it has been reported by a customer that queries with partition keys didn't work, and I also verified myself (not sure how I tested this before that it didn't break, but here we are).

With these changes we move the overhead of this call to the container initialization instead, since regardless any method a user would've first used with the container would have spent time on this network call regardless. This way, we also simplify the query logic in the back, and make sure that containers all have that check done when initialized.

**_Edit:_** This change is no longer what it previously described - although it was a good solution, it does not affect users who initialize their container proxies by calling `get_container_client()`. Instead, the logic is now in the query_iterable class, and takes place during query execution.


sdk/cosmos/azure-cosmos/azure/cosmos/aio/database.py
```diff
@@ -230,7 +230,9 @@ async def create_container(
         if response_hook:
             response_hook(self.client_connection.last_response_headers, data)
 
-        return ContainerProxy(self.client_connection, self.database_link, data["id"], properties=data)
+        container = ContainerProxy(self.client_connection, self.database_link, data["id"], properties=data)
+        await container._check_is_system_key()
```
simorenoh, line 234: this is the big change, where we preemptively do this network call before sending back the container instance to the user

tjprescott: At @simorenoh do you have any concerns with this approach? This call is already awaiting other async calls, so this doesn't strike me as crazy. @annatisch?

simorenoh: Depends on what the user is doing, but so long as they're passing in a partition key value to any of the functions affected (read_item(), query_items(), delete_item(), query_conflicts(), read_conflict(), delete_conflict()) they will have to use this result.

Since the list above includes several item methods, I'd say there's a good chance they'd need it.

simorenoh: Nevermind, there is more things to think about... a proper concern with this in that users may not necessarily create containers (and instead use `get_container_client()` to obtain their existing instance) which would still not work, will think about this more.

sdk/cosmos/azure-cosmos/azure/cosmos/aio/_query_iterable_async.py
```diff
@@ -77,11 +78,6 @@ def __init__(
         )
         super(QueryIterable, self).__init__(self._fetch_next, self._unpack, continuation_token=continuation_token)
 
-    async def __aiter__(self):
-        if 'partition_key' in self._options:
-            self._options['partition_key'] = await self._options['partition_key']
-        return self
```
annatisch, line 83: Were there any of the other async iterators where we had logic in `aiter` that needs to be moved?
We should also probably remove any other empty `aiter`s if we added them.

simorenoh: the other three `aiter` only do `return self` and are also present as `iter` in their sync versions, so I think we should be good with just removing this

annatisch (approved):

PR #22895, opened 2022-02-03
Go from PyPy3.6 -> PyPy3.7

scbedd:
This method absolutely works for pypy3.8, but we do NOT have a wheel for `cryptography<4` at pypy3.8.

If we go to `pypy3.8` then we'll need to remove `crypto<4`, but last I heard crypto 36 has a breaking change that we haven't dealt with yet.

@xiangyan99 for pypy37 vs 38 and crypto discussion. @mccoyp for engsys changes.


eng/tox/tox.ini
```diff
@@ -51,21 +51,23 @@ platform = linux: linux
            macos: darwin
            windows: win32
 passenv = *
+download=true
 requires=
   {[packaging]pkgs}
 setenv =
   SPHINX_APIDOC_OPTIONS=members,undoc-members,inherited-members
   PIP_EXTRA_INDEX_URL=https://pypi.python.org/simple
+  VIRTUALENV_WHEEL=0.37.0
```
scbedd, line 60: I'm hoping this will have a positive impact on our `Windows 145` errors.

eng/tox/tox.ini
```diff
@@ -51,21 +51,23 @@ platform = linux: linux
...
+  VIRTUALENV_PIP=20.3.3
+  VIRTUALENV_SETUPTOOLS=60.7.0
 deps = {[base]deps}
 changedir = {toxinidir}
-install_command = python -m pip install {opts} {packages} --cache-dir {toxinidir}/../.tox_pip_cache_{envname}
-commands_pre=
-    {envbindir}/python -m pip install pip==20.3
+install_command = python -m pip install {opts} {packages} --cache-dir {toxinidir}/../.tox_pip_cache_{envname} --ignore-installed
```
scbedd, line 65: `--ignore-installed` is because on pypy3.7/pypy3.8, the _virtualenv_ pip is finding site-packages from the _global_ install. [This is what that looked like](https://dev.azure.com/azure-sdk/public/_build/results?buildId=1342443&view=logs&j=563de66a-0637-56b5-a77c-8b9a24c478c0&t=b0c3bd39-e5d0-5f0d-f1d7-d5569ea832ef&l=2670).

image

Aside from this install addition, I have not been able to get that behavior to STOP. For clarity, on `pypy3.6.12` works JUST FINE regardless if I prepend the path using MY python script vs the default `pypy3` selected by UsePythonVersion task.

That means that the issue causing this weird behavior is native to pypy, and not specific to devops.

mccoyp (approved):
Looks good as far as I can tell -- is [this](https://github.com/Azure/azure-sdk-for-python/pull/22895/files#diff-ff7f21df73c8ed2f0f699ac9fd4ef745c6908760ee95f15633a2720122658a47R95) how the output of the "PyPy3 Specific Path Prepend" task is produced? Just curious since the script doesn't seem to actually `return` anything and I'm unfamiliar with how these tasks work

scbedd:
Correct! These are devops [logging commands](https://docs.microsoft.com/en-us/azure/devops/pipelines/scripts/logging-commands?view=azure-devops&tabs=bash)

Issue #21653, opened 2021-11-08
AzureCliCredential fails to get token because strptime() argument 1 must be str, not None


sdebruyn:
- **Package Name**: azure-identity
- **Package Version**: 1.7.0
- **Operating System**: Linux-5.4.0-1059-azure-x86_64-with-glibc2.28 (Debian slim Buster)
- **Python Version**:  3.9.7

**Bug description**
AzureCliCredential fails to get token. Scope was `"https://database.windows.net//.default"`

```
[2021-11-09 00:08:54,452] {subprocess.py:78} INFO - 2021-11-08 23:08:54.451818 (MainThread): AzureCliCredential.get_token failed: strptime() argument 1 must be str, not None
[2021-11-09 00:08:54,452] {subprocess.py:78} INFO - Traceback (most recent call last):
[2021-11-09 00:08:54,452] {subprocess.py:78} INFO -   File "/home/airflow/.local/lib/python3.9/site-packages/azure/identity/_internal/decorators.py", line 30, in wrapper
[2021-11-09 00:08:54,452] {subprocess.py:78} INFO -     token = fn(*args, **kwargs)
[2021-11-09 00:08:54,452] {subprocess.py:78} INFO -   File "/home/airflow/.local/lib/python3.9/site-packages/azure/identity/_credentials/azure_cli.py", line 75, in get_token
[2021-11-09 00:08:54,453] {subprocess.py:78} INFO -     token = parse_token(output)
[2021-11-09 00:08:54,453] {subprocess.py:78} INFO -   File "/home/airflow/.local/lib/python3.9/site-packages/azure/identity/_credentials/azure_cli.py", line 94, in parse_token
[2021-11-09 00:08:54,453] {subprocess.py:78} INFO -     dt = datetime.strptime(token["expiresOn"], "%Y-%m-%d %H:%M:%S.%f")
[2021-11-09 00:08:54,453] {subprocess.py:78} INFO - TypeError: strptime() argument 1 must be str, not None
```

**Context**
Azure CLI is logged in using MSI

Command: `az login -i`
Output:
```
[2021-11-09 00:08:47,407] {subprocess.py:74} INFO - Output:
[2021-11-09 00:08:48,852] {subprocess.py:78} INFO - [
[2021-11-09 00:08:48,853] {subprocess.py:78} INFO -   {
[2021-11-09 00:08:48,853] {subprocess.py:78} INFO -     "environmentName": "AzureCloud",
[...]

PR #22899, opened 2022-02-03
handle null expireson

xiangyan99:
#21653

### DIFF
--- a/sdk/identity/azure-identity/azure/identity/_credentials/azure_cli.py
+++ b/sdk/identity/azure-identity/azure/identity/_credentials/azure_cli.py
@@ -2,7 +2,7 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
 # ------------------------------------
-from datetime import datetime
+from datetime import datetime, timedelta
 import json
 import os
 import platform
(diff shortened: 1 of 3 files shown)


sdk/identity/azure-identity/azure/identity/_credentials/azure_cli.py
```diff
@@ -91,7 +91,10 @@ def parse_token(output):
     """
     try:
         token = json.loads(output)
-        dt = datetime.strptime(token["expiresOn"], "%Y-%m-%d %H:%M:%S.%f")
+        expires_on_in_token = token["expiresOn"]
+        if not expires_on_in_token:
+            expires_on_in_token = (datetime.now()+timedelta(days=1)).strftime("%Y-%m-%d %H:%M:%S.%f")
```
chlowell, line 96: The token's lifetime depends on the tenant's configuration. We don't know anything about that, so we can't safely guess when the token expires. Underestimating degrades performance; overestimating causes the auth policy to send an expired token. Really this is a CLI bug that should be fixed in the CLI. I think we have two options for handling it here:
1. use the token only once (we can assume it's valid right now)
2. raise AuthenticationFailedError

Option 1 makes the application terribly slow; option 2 maybe crashes it 😄

chlowell (changes requested):

jiasli:
This issue has long been fixed (https://github.com/Azure/azure-cli/pull/20219). Azure CLI guarantees the existence of `expiresOn`.

PR #22902, opened 2022-02-03
Set AZURE_TEST_RUN_LIVE in pipeline

mccoyp:
# Description

Without setting this variable, test proxy tooling isn't able to detect that we're running tests live. Because of this, if `AZURE_TEST_RUN_LIVE` isn't set in a library's `tests.yml` file, we'll get errors because the test proxy isn't available to agents.

@scbedd for notifs

# All SDK Contribution checklist:
- [x] **The pull request does not introduce [breaking changes]**
- [x] **CHANGELOG is updated for new features, bug fixes or other significant changes.**
- [x] **I have read the [contribution guidelines](https://github.com/Azure/azure-sdk-for-python/blob/main/CONTRIBUTING.md).**

## General Guidelines and Best Practices
- [x] Title of the pull request is clear and informative.
- [x] There are a small number of commits, each of which have an informative message. This means that previously merged commits do not appear in the history of the PR. For more information on cleaning up the commits in your PR, [see this page](https://github.com/Azure/azure-powershell/blob/master/documentation/development-docs/cleaning-up-commits.md).

### [Testing Guidelines](https://github.com/Azure/azure-sdk-for-python/blob/main/CONTRIBUTING.md##building-and-testing)
- _N/A_ Pull request includes test coverage for the included changes.

PR #22908, opened 2022-02-03
Revert "Set AZURE_TEST_RUN_LIVE in pipeline"

benbp:
Reverts Azure/azure-sdk-for-python#22902

This breaks many existing pipelines that pre-define AZURE_TEST_RUN_LIVE because it ends up getting defined+injected twice:

https://github.com/Azure/azure-sdk-for-python/blob/main/eng/pipelines/templates/jobs/live.tests.yml#L114-L115

```
            AZURE_TEST_RUN_LIVE: 'true'
            ${{ insert }}: ${{ parameters.EnvVars }}
```

For example:

{"$id":"1","customProperties":**{"ValidationResults":[{"result":"error","message":"/eng/pipelines/templates/jobs/live.tests.yml (Line: 115, Col: 28): 'AZURE_TEST_RUN_LIVE' is already defined"}]}**,"innerException":null,"message":"Could not queue the build because there were validation errors or warnings.","typeName":"Microsoft.TeamFoundation.Build.WebApi.BuildRequestValidationFailedException, Microsoft.TeamFoundation.Build2.WebApi","typeKey":"BuildRequestValidationFailedException","errorCode":0,"eventId":3000}

FYI @mccoyp @scbedd

### DIFF
--- a/eng/pipelines/templates/jobs/live.tests.yml
+++ b/eng/pipelines/templates/jobs/live.tests.yml
@@ -111,7 +111,6 @@ jobs:
           CoverageArg: $(CoverageArg)
           EnvVars:
             AZURE_RUN_MODE: 'Live' #Record, Playback
-            AZURE_TEST_RUN_LIVE: 'true'
             ${{ insert }}: ${{ parameters.EnvVars }}
           PreSteps: ${{ parameters.PreSteps }}
           PostSteps: ${{ parameters.PostSteps }}


mccoyp (approved):
Yep, just became aware of this a few minutes ago. Thank you for beating me to opening a PR!

benbp:
There are a couple ways we can fix this, either with the templates or removing the duplicate definitions from the `tests.yml` files, but I wanted to put up this revert to unblock any existing pipeline failures.

PR #22912, opened 2022-02-04
[formrecognizer] Adding 2022-01-30-preview work

catalinaperalta:
This PR is moving the 2022-01-30-preview work checked into the private repo into main.


kristapratico (approved):
all work was already reviewed on the private repo 🐰

catalinaperalta:
For the live test pipeline results, the following samples are failing because they are missing the build_mode parameter:
'sample_create_composed_model.py', 'sample_manage_models.py', 'sample_create_composed_model_async.py', 'sample_manage_models_async.py'
This needs to fix the path to the sample file: 'sample_analyze_read_async.py'

Will merge and fix in the next PR.

PR #22919, opened 2022-02-04
Fix live search tests.

tjprescott:
Update to reenable live tests for search.


sdk/search/tests.yml
```diff
@@ -13,5 +13,5 @@ stages:
         AZURE_CLIENT_SECRET: $(SEARCH_CLIENT_SECRET)
         AZURE_TENANT_ID: $(SEARCH_TENANT_ID)
         AZURE_CLIENT_OID: $(SEARCH_CLIENT_OID)
-        TEST_MODE: 'RunLiveNoRecord'
```
mccoyp, line 16: I see that most `tests.yml` files still use this variable. It won't affect anything in the Python-side proxy tooling, but I don't know if it's used for pipelines in some other way. @scbedd or @benbp, do you know if this variable is necessary?

benbp: It does not appear to be used.

mccoyp (approved):

PR #22942, opened 2022-02-07
[core] add fake secrets in tests to credscan

iscai-msft:


lmazuel (approved):

eng/CredScanSuppression.json
```diff
@@ -32,6 +32,7 @@
...
+                "SharedKey account:G4jjBXA7LI/RnWKIOQ8i9xH4p76pAQ+4Fs4R1VxasaE="
```
weshaggard, line 35: Is it possible to reuse an existing placeholder instead of adding a new one?

PR #22946, opened 2022-02-07
[Cosmos] Add max_integrated_cache_staleness param option to item methods

simorenoh:
These changes make it so users can start trying out the CosmosDB integrated cache in the Python SDK with either the sync or async clients. These changes are basically just passing in additional headers.

Still missing marking these headers as a preview feature. We currently have the SDK in beta version while we work in the async client, but we can't have so many beta versions in a row for the SDK - users shy away from using beta versions, even if not all of it is in beta. We should find a way to make sure we communicate this is a preview header, but the client itself should be good to use.


sdk/cosmos/azure-cosmos/azure/cosmos/aio/container.py
```diff
@@ -198,16 +199,20 @@ async def create_item(
...
+            partition_key,  # type: Any
+            max_integrated_cache_staleness_in_ms=None,  # type: Optional[int],
+            **kwargs  # type: Any
     ):
         # type: (...) -> Dict[str, Any]
         """Get the item identified by `item`.
 
         :param item: The ID (name) or dict representing item to retrieve.
         :param partition_key: Partition key for the item to retrieve.
+        **Provisional** parameter max_integrated_cache_staleness_in_ms
```
kushagraThapar, line 213: @annatisch  - with this `Provisional` API mention, would we be fine GAing our SDK even if this stays `Provisional`?

sdk/cosmos/azure-cosmos/azure/cosmos/aio/container.py
```diff
@@ -198,16 +199,20 @@ async def create_item(
...
+        :param max_integrated_cache_staleness_in_ms: The max cache staleness for the integrated cache in milliseconds.
```
kushagraThapar, line 214: @simorenoh  - please update the documentation similar to below lines-

```
/**
     * Gets the staleness value associated with the request in the Azure CosmosDB service. For requests where the {@link
     * com.azure.cosmos.ConsistencyLevel} is {@link com.azure.cosmos.ConsistencyLevel#EVENTUAL} or {@link com.azure.cosmos.ConsistencyLevel#SESSION}, responses from the
     * integrated cache are guaranteed to be no staler than value indicated by this maxIntegratedCacheStaleness.
     *
     * <p>Default value is null</p>
     *
     * <p>Cache Staleness is supported in milliseconds granularity. Anything smaller than milliseconds will be ignored.</p>
     *
     * @return Duration of maxIntegratedCacheStaleness
     */
```

sdk/cosmos/azure-cosmos/azure/cosmos/aio/container.py
```diff
@@ -198,16 +199,20 @@ async def create_item(
...
     @distributed_trace_async
     async def read_item(
-        self,
-        item,  # type: Union[str, Dict[str, Any]]
-        partition_key,  # type: Any
-        **kwargs  # type: Any
+            self,
+            item,  # type: Union[str, Dict[str, Any]]
+            partition_key,  # type: Any
+            max_integrated_cache_staleness_in_ms=None,  # type: Optional[int],
```
kushagraThapar, line 205: @annatisch - is this python style of adding new parameters?
What if in the future, there are 10 more of these? Would we keep adding them here?
I am assuming, this is not a breaking change in python, and won't be in future if we keep doing this. Am I right?

sdk/cosmos/azure-cosmos/CHANGELOG.md
```diff
@@ -3,13 +3,17 @@
 ### 4.3.0b3 (Unreleased)
 
 #### Features Added
+- Added new **provisional** `max_integrated_cache_staleness_in_ms` parameter to item read and query methods in order
```
kushagraThapar, line 6:
```suggestion
- Added new **provisional** `max_integrated_cache_staleness_in_ms` parameter to read item and query items APIs in order
```

sdk/cosmos/azure-cosmos/azure/cosmos/_base.py
```diff
@@ -657,3 +661,9 @@ def ParsePaths(paths):
             tokens.append(token)
 
     return tokens
+
+
+def validate_cache_staleness_value(max_integrated_cache_staleness):
+    int(max_integrated_cache_staleness)  # Will throw error if data type cant be converted to int
+    if max_integrated_cache_staleness < 0:
```
kushagraThapar, line 668: I don't think we allow 0 milliseconds in Java SDK, worth checking with Compute Gateway team on this. 

FYI, this is the Java SDK logic -> 

```
if (maxIntegratedCacheStaleness.toNanos() > 0 && maxIntegratedCacheStaleness.toMillis() <= 0) {
            throw new IllegalArgumentException("MaxIntegratedCacheStaleness granularity is milliseconds");
        }
        if (maxIntegratedCacheStaleness.toMillis() < 0) {
            throw new IllegalArgumentException("MaxIntegratedCacheStaleness duration cannot be negative");
        }
        return maxIntegratedCacheStaleness.toMillis();
```

sdk/cosmos/azure-cosmos/azure/cosmos/aio/container.py
```diff
@@ -296,6 +312,9 @@ def query_items(
         :param enable_scan_in_query: Allow scan on the queries which couldn't be served as
             indexing was opted out on the requested paths.
         :param populate_query_metrics: Enable returning query metrics in response headers.
+        **Provisional** parameter max_integrated_cache_staleness_in_ms
+        :param max_integrated_cache_staleness_in_ms: The max cache staleness for the integrated cache in milliseconds.
```
kushagraThapar, line 316: Same here and everywhere else, please update the documentation.

kushagraThapar (changes requested):

sdk/cosmos/azure-cosmos/azure/cosmos/aio/container.py
simorenoh: I did see that Java had this whole thing, but I'll have to talk to Anna on this amount of comments for a parameter since in Java the object has it's own methods as opposed to Python, and it might end up cluttering the method description.

sdk/cosmos/azure-cosmos/azure/cosmos/_base.py
```diff
@@ -657,3 +661,9 @@ def ParsePaths(paths):
             tokens.append(token)
 
     return tokens
+
+
+def validate_cache_staleness_value(max_integrated_cache_staleness):
+    int(max_integrated_cache_staleness)  # Will throw error if data type cant be converted to int
+    if max_integrated_cache_staleness <= 0:
+        raise ValueError("Parameter 'max_integrated_cache_staleness_in_ms' can only be a positive integer.")
```
annatisch, line 669: We usually let the service do these kind of type/value validations - what would the service do if I set this value to `-1`? Or a string?

sdk/cosmos/azure-cosmos/azure/cosmos/aio/container.py
annatisch: Yes and no.
We would keep adding keyword arguments. And would keep doing so in future.
However unlike here - they would need to be keyword only (otherwise it _can_ lead to future breaking changes). 
@simorenoh in this case it will either need to be rolled into kwargs (probably advisable given it's status), or use the `*` syntax to mark it as keyword-only.

sdk/cosmos/azure-cosmos/azure/cosmos/aio/container.py
annatisch: There are a couple of questions still open around that.
For example - the SDK can not go GA pointing to a preview API version.
Given that this parameter effectively amounts to just a header - we can actually unblock customers who want to use this without adding it as an API feature. That way we have time to chose how to surface it in the SDK as a GA feature when the service goes GA, and customers who were already using it remain unbroken.
In this case it would be something like:
```python
client = CosmosClient(..., api_version="preview-version")
client.read_item(..., headers={"DedicatedGatewayCacheStaleness":"10"}
```

sdk/cosmos/azure-cosmos/azure/cosmos/_base.py
simorenoh: I just double checked and it seems like the backend does no validation on this value, and instead sets the default value of 5 mins so long as the dedicated gateway is configured for the account - one doesn't even have to pass the option in explicitly for the integrated cache to be used, since it seems it gets utilized regardless.

kushagraThapar (approved):

kushagraThapar:
@simorenoh  - was this PR only blocked because of me, if so, apologies for late approval. Please merge it if it is ready. If not, let me know what we need to do to get it in. Thanks!

sdk/cosmos/azure-cosmos/azure/cosmos/aio/container.py
simorenoh: This is a preview feature but does not require an API version change - it instead requires users to set up the dedicated gateway in their database accounts manually in order to point their service at a different endpoint, so these concerns are taken care of.

sdk/cosmos/azure-cosmos/samples/test_axq.py
```diff
@@ -0,0 +1,82 @@
...
+    db = client.create_database_if_not_exists(DATABASE_ID)
+    cont = db.create_container_if_not_exists(CONTAINER_ID, partition_key=PartitionKey(path="/id"))
+    body = get_test_item()
+    cont.create_item(body=body)
+
+    @patch.object(cont.client_connection, '_CosmosClientConnection__Get')
+    def get_headers():
+        cont.read_item(item=body["id"], partition_key=body["id"], max_integrated_cache_staleness_in_ms=(1000 * 60))
+
+    get_headers()
```
kushagraThapar, line 39: where are we validating the headers value?

kushagraThapar (changes requested):

sdk/cosmos/azure-cosmos/samples/test_axq.py
simorenoh: we're not, I just uploaded this in case Annie wanted to try something since she messaged me, it's not done

kushagraThapar (approved):

PR #22949, opened 2022-02-07
Change future value behavior of message state

rakshith91:


sdk/servicebus/azure-servicebus/azure/servicebus/_common/message.py
```diff
@@ -971,17 +971,19 @@ def dead_letter_source(self):
...
 
-        :rtype: ~azure.servicebus.ServiceBusMessageState
+        :rtype: int
         """
         try:
             message_state = self._raw_amqp_message.annotations.get(MESSAGE_STATE_NAME)
             if not message_state:
                 return ServiceBusMessageState.ACTIVE
+            elif message_state not in (0, 1, 2):
+                return message_state
```
annatisch, line 986: I think you could simply use:
```python
try:
    return ServiceBusMessageState(message_state)
except ValueError:
    return  message_state
```

yunhaoling (approved):
do we need to remove this `ServiceBusMessageState` class from our api?

rakshith91:
> do we need to remove this ServiceBusMessageState class from our api?

we would still need it :)

sdk/servicebus/azure-servicebus/azure/servicebus/_common/message.py
rakshith91: it'll be more like 
```
            try:
                return ServiceBusMessageState(message_state)
            except ValueError:
                if not message_state:
                    return ServiceBusMessageState.ACTIVE
                return message_state
```

annatisch: Yep, though I was just referring to the `elif`.... so:
```python
if not message_state:
    return ServiceBusMessageState.ACTIVE
try:
    return ServiceBusMessageState(message_state)
except ValueError:
    return message_state
```
But either/or. both should work.

Issue #21765, opened 2021-11-15
[ServiceBus] Stress test onboarding


yunhaoling:
after finishing up the issue: https://github.com/Azure/azure-sdk-for-python/issues/21565

get other python messaging crew onboard with the stress test project

yunhaoling:
- stress test scenarios
  - send and receive large messages

yunhaoling:
https://dev.azure.com/azure-sdk/internal/_wiki/wikis/internal.wiki/485/Stress-test-checklist
https://github.com/Azure/azure-sdk-tools/blob/main/tools/stress-cluster/chaos/README.md
https://github.com/Azure/azure-sdk-for-js/tree/main/sdk/servicebus/service-bus/test/stress
https://github.com/Azure/azure-sdk-for-python/tree/main/sdk/eventhub/azure-eventhub/stress

swathipil:
Following Richard's PR in Go, we should also test against ServicePrincipal/AAD and not just with connection strings: https://github.com/Azure/azure-sdk-for-go/blob/1a459bfba83ccae3e4db2982e6c8d5dc14ee9c54/sdk/messaging/azservicebus/internal/stress/stress-test-resources.bicep

PR #22982, opened 2022-02-08
[ServiceBus] initial stress test onboarding

swathipil:
fixes: #21765


swathipil:
@yunhaoling - Should we be ignoring these stress tests in the pipelines?

benbp:
Stress config pieces LGTM

sdk/servicebus/azure-servicebus/stress/templates/testjob.yaml
```diff
@@ -0,0 +1,17 @@
...
+    chaos: "true"
+spec:
+  containers:
+    - name: python-sb-stress
+      image: {{ .Values.image }}
+      imagePullPolicy: Always
+      command: ['bash', '-c', 'python3 test_stress_queues.py']
+      {{- include "stress-test-addons.container-env" . | nindent 6 }}
+      # async test command
+      # command: ['bash', '-c', 'python3 azure_eventhub_producer_stress.py -m stress_send_list_async & python3 azure_eventhub_consumer_stress_async.py']
```
yunhaoling, line 16: nit:
```suggestion
```

sdk/servicebus/azure-servicebus/stress/scripts/test_stress_queues.py
```diff
@@ -0,0 +1,332 @@
...
+                                    azure_monitor_metric=AzureMonitorMetric("test_stress_queue_check_for_dropped_messages")
+                                    )
+
+    result = stress_test.run()
+    assert(result.total_sent > 0)
+    assert(result.total_received > 0)
+
+if __name__ == '__main__':
+    #parser = ArgumentParser()
+    pytest.main()
```
yunhaoling, line 332: emmm, curious on where this "main" is defined?
what would happen with `python test_stress_queues.py`, would each test method be called?

yunhaoling (approved):

yunhaoling:
@swathipil , yes, I do think those stress tests should be ignored in our livetest pipeline.

Ideally, we should have separate pipeline for running stress tests which takes place weekly (less frequently) depending on how the stress tests are written.

sdk/servicebus/azure-servicebus/stress/scripts/test_stress_queues.py
swathipil: yeah this is one of the ways to run the tests rather than running the `pytest` command. All of the tests are run as normal when `python test_stress_queues.py` is called. We can also pass in all the same flags.

I did it this way b/c I thought we might want to add the ArgmentParser in the future when we have infinite send/receive to take args for stress tests.

sdk/servicebus/azure-servicebus/stress/Dockerfile
```diff
@@ -0,0 +1,10 @@
+# syntax=docker/dockerfile:1
+
+FROM python:3.8-slim-buster
+
+WORKDIR /app
+
+COPY ./scripts /app/stress/scripts
+
+WORKDIR /app/stress/scripts
+RUN pip3 install -r dev_requirements.txt
```
richardpark-msft, line 10: It must be nice to not have a build step :)

(I have to do a two-stage dockerfile for Go)

sdk/servicebus/azure-servicebus/stress/scripts/conftest.py
```diff
@@ -0,0 +1,46 @@
...
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+#--------------------------------------------------------------------------
+import pytest
+import os
+import sys
+from azure.servicebus import ServiceBusClient
+from dotenv import load_dotenv
+
+LOGGING_ENABLE = False
```
richardpark-msft, line 12: How verbose are your logs, btw? Wouldn't you want this on all the time?

sdk/servicebus/azure-servicebus/stress/scripts/conftest.py
```diff
@@ -0,0 +1,46 @@
...
+ENV_FILE = os.environ.get('ENV_FILE')
+load_dotenv(dotenv_path=ENV_FILE, override=True)
+
+# Allow us to pass stress_test_duration from the command line. TODO: may not actually be using this.
+def pytest_addoption(parser):
+    parser.addoption('--stress_test_duration_seconds', action="store", default=None)
```
richardpark-msft, line 18: Given the typical duration of stress tests (hours or even days!) you might want to consider a _slightly_ larger default unit (maybe minutes?)

sdk/servicebus/azure-servicebus/stress/scripts/conftest.py
richardpark-msft: @benbp  - now that we're auto-logging all output into appinsights perhaps we need to think about some reasonable measure here? Any recommendations?

benbp: @richardpark-msft I think we can keep an eye on it. @ckairen do you think it would be possible to graph and/or alert on high log counts per test or namespace? I think we only need to care if a test is uploading multiple GB/s per day, otherwise it's pretty cheap.

sdk/servicebus/azure-servicebus/stress/scripts/conftest.py
swathipil: noted for future changes

sdk/servicebus/azure-servicebus/stress/scripts/conftest.py
benbp: https://github.com/Azure/azure-sdk-tools/issues/2776

richardpark-msft: (storage costs aside, I'm mostly worried about someone uploading so much that they block ingestion)

sdk/servicebus/azure-servicebus/stress/scripts/conftest.py
richardpark-msft: A polite way of saying 'we're not doing this'. I can accept this.

sdk/servicebus/azure-servicebus/stress/scripts/conftest.py
richardpark-msft: (Resolving this since the issue been filed)

sdk/servicebus/azure-servicebus/stress/scripts/conftest.py
swathipil: nooo, I just meant noted so I can make the change this upcoming sprint :)

sdk/servicebus/azure-servicebus/stress/scripts/conftest.py
benbp: > (storage costs aside, I'm mostly worried about someone uploading so much that they block ingestion)

This would be extremely difficult :)

PR #22984, opened 2022-02-09
[Storage] Support OAuth for import/export managed disks

jalauzon-msft:
Two changes were needed for this feature:
- Ensure blob clients can support both OAuth and SAS together
- Follow the bearer challenge to the correct scope from an OAuth request failure using the Storage scope


sdk/storage/azure-storage-blob/azure/storage/blob/_shared/policies_async.py
```diff
@@ -218,3 +221,24 @@ def get_backoff_time(self, settings):
...
+    async def on_challenge(self, request, response):
+        # type: (PipelineRequest, PipelineResponse) -> bool
+        try:
+            auth_header = response.http_response.headers.get("WWW-Authenticate")
+            challenge = StorageHttpChallenge(auth_header)
+        except ValueError:
+            return False
+
+        scope = challenge.resource_id + DEFAULT_OAUTH_SCOPE
+        await self.authorize_request(request, scope, tenant_id=challenge.tenant_id)
```
evelyn-ys, line 242: Since for disk scenario, two requests will be sent:
- first one for storage scope
- second one for disk scope.

There would exist bugs when fetching `data_stream_total`: https://github.com/Azure/azure-sdk-for-python/blob/f424ae46c43dfd92265c55bc81f2df343965d80e/sdk/storage/azure-storage-blob/azure/storage/blob/_shared/policies.py#L306-L313

From the first request, we will get a response without `content-range` as it returns auth failure. The correct `content-range` is returned in second response. But at that time, `data_stream_total` won't change anymore.

This causes `data_stream_current` > `data_stream_total`, and it breaks CLI progress bar.

jalauzon-msft: Thanks for finding the issue! I was not aware of this form of progress tracking, and it was indeed broken for Bearer challenges. I have hopefully addressed this in the latest update.

sdk/storage/azure-storage-blob/azure/storage/blob/_shared/authentication.py
```diff
@@ -140,3 +141,41 @@ def on_request(self, request):
...
+
+class StorageHttpChallenge(object):
+    def __init__(self, challenge):
+        """ Parses an HTTP WWW-Authentication Bearer challenge from the Storage service. """
+        if not challenge:
+            raise ValueError("Challenge cannot be empty")
+
+        self._parameters = {}
+
+        split_challenge = challenge.strip().split(" ", 1)
```
tasherif-msft, line 154: You can just do:
```suggestion
        self.scheme, trimmed_challenge  = challenge.strip().split(" ", 1)
 ```

File: sdk/storage/azure-storage-blob/azure/storage/blob/_shared/policies.py [hunk 3]
```diff
@@ -620,3 +634,24 @@ def get_backoff_time(self, settings):
             if self.backoff > self.random_jitter_range else 0
         random_range_end = self.backoff + self.random_jitter_range
         return random_generator.uniform(random_range_start, random_range_end)
+
+
+class StorageBearerTokenCredentialPolicy(BearerTokenCredentialPolicy):
```
tasherif-msft, line 639: for different clouds (mooncake, gov, etc) how are we accounting for this for the scopes?

sdk/storage/azure-storage-blob/azure/storage/blob/_shared/authentication.py
jalauzon-msft: Good point. Will do

sdk/storage/azure-storage-blob/azure/storage/blob/_shared/policies.py
jalauzon-msft: I'm not exactly sure how OAuth works in those clouds, but everything should just work. We are using the same default scope as we were previously `STORAGE_OAUTH_SCOPE` (it just moved into this file) and then, in the case of a challenge, we will follow whatever scope is provided in the challenge (from the service). I'm not sure if customers override `STORAGE_OAUTH_SCOPE` in certain clouds or what happens but it should continue to work, just now with added support for challenges.

tasherif-msft: We should create an item to investigate and make sure that the Python SDK supports other clouds at some point.

tasherif-msft (approved):

sdk/storage/azure-storage-blob/azure/storage/blob/_shared/policies.py
jalauzon-msft: Sure. I've opened #23198 for that.

PR #22891, opened 2022-02-03
use ParamSpec to enable intellisense with distributed_trace decorators

kristapratico:
This PR is related to this [discussion](https://teams.microsoft.com/l/message/19:b97d98e6d22c41e0970a1150b484d935@thread.skype/1643319956805?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=3e17dcb0-4257-4a30-b843-77f47f1d4121&parentMessageId=1643319956805&teamName=Azure%20SDK&channelName=Language%20-%20Python&createdTime=1643319956805) on Teams where we noticed that our `distributed_trace` decorators were blocking the intellisense type help. Fortunately there is a solution to this called typing.ParamSpec which forwards the parameter types of one callable to another callable and restores our type hinting!

ParamSpec: https://docs.python.org/3/library/typing.html#typing.ParamSpec

~This only works for py3.10+, but does enable intellisense as expected:~ This works for all supported Python versions.

VSCode/pylance:

image

Pycharm:

image

PR #23017, opened 2022-02-10
[ServiceBus&EventHubs] fix mypy 0.931 errors

swathipil:
follow up to #22891

### DIFF
--- a/sdk/eventhub/azure-eventhub/azure/eventhub/_client_base.py
+++ b/sdk/eventhub/azure-eventhub/azure/eventhub/_client_base.py
@@ -46,7 +46,7 @@ if TYPE_CHECKING:
     from azure.core.credentials import TokenCredential
 
 _LOGGER = logging.getLogger(__name__)
-_Address = collections.namedtuple("Address", "hostname path")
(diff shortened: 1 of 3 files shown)


sdk/eventhub/azure-eventhub/azure/eventhub/_client_base.py
```diff
@@ -46,7 +46,7 @@
     from azure.core.credentials import TokenCredential
 
 _LOGGER = logging.getLogger(__name__)
-_Address = collections.namedtuple("Address", "hostname path")
+_Address = collections.namedtuple("Address", "hostname path")   # type: ignore
```
swathipil, line 49: @yunhaoling - Can we change the inner named tuple to `_Address`?

yunhaoling: yes, I think we could rename the tuple as long as this namedtuple is kept for internal usage.

swathipil: yep only used internally, not exposed to user.

PR #23021, opened 2022-02-10
[textanalytics] analyze actions fix for targeting v3.1 in beta

kristapratico:
This PR fixes a bug where incorrect action types were being sent in the request if targeting the older API version `v3.1` in the beta version of the client library. Luckily this is not an issue in the stable release itself. good call on the fix @iscai-msft!


iscai-msft (approved):
thanks @kristapratico

Issue #23006, opened 2022-02-10
Cosmos replace-functions (e.g. store procedures) do not work with "serverScript" key in the SPROC dict


mxkus:
- **Package Name**: azure-cosmos
- **Package Version**: At least 4.2.0, probably all other versions
- **Operating System**: Windows, but not OS specific
- **Python Version**: 3.9.5 but not version specific

**Describe the bug**
When replacing a stored procedure with the dictionary `{"id": SPROC_NAME, "serverScript": SPROC_CONTENT}`, there is a CosmosHttpResponseError. When using ` {"id": SPROC_NAME, "body": SPROC_CONTENT}`, it works fine.

**To Reproduce**
```
client = cosmos.CosmosClient("URI", credential="KEY")
database = client.get_database_client("DB_NAME")
container = database.get_container_client("CONTAINER_NAME")

sproc = {"id": SPROC_NAME, "serverScript": SPROC_CONTENT}
# this works
container.scripts.create_stored_procedure(body=sproc)

# this fails due to CosmosHttpResponseError
container.scripts.replace_stored_procedure(sproc=SPROC_NAME, body=sproc)
```

```CosmosHttpResponseError: (BadRequest) Message: {"Errors":["The input content is invalid - 'serverScript' is not a valid property in the current payload.","The request payload is invalid. Ensure to provide a valid request payload."]}```

The problem is that replacing and creating stuff (UDF, Trigger, SPROC) have different behaviours with regard to the "serverScript" key in `azure\cosmos\_cosmos_client_connection.py`:
* When creating, the key "serverScipt" is popped and mapped: 
```        sproc = sproc.copy()
        if sproc.get("serverScript"):
            sproc["body"] = str(sproc.pop("serverScript", ""))
        elif sproc.get("body"):
            sproc["body"] = str(sproc["body"])
```
* When replacing, the key "serverScipt" is mapped but stays in the dictionary:
```        sproc = sproc.copy()
        if sproc.get("serverScript"):
            sproc["body"] = str(sproc["serverScript"])
        elif sproc.get("body"):
            sproc["body"] = str(sproc["body"])
```

This leads to the described error.

**Expected behavior**
The stored procedures should also work with the `serverScript` key in the dict. It should also use the `pop` statement.

**Screenshots**
The problem is described and a solution provided.

**Additional context**

simorenoh:
Hi @mxkus, thank you for bringing this up!
[...]

PR #23076, opened 2022-02-15
[Cosmos] hotfix: replace sproc/trigger/udf request body invalid

simorenoh:
Based on this user-reported issue: https://github.com/Azure/azure-sdk-for-python/issues/23006

Basically, we were still passing in the "serverScript" body parameter to the backend as part of the request body when attempting to replace any of the listed resources. Due to this, the backend would return errors stating that the body was invalid.

This PR makes the changes to ensure the backend receives a proper body the same way it does when doing create operations for any of the above resources.

### DIFF
--- a/sdk/cosmos/azure-cosmos/azure/cosmos/aio/_cosmos_client_connection_async.py
+++ b/sdk/cosmos/azure-cosmos/azure/cosmos/aio/_cosmos_client_connection_async.py
@@ -1076,7 +1076,7 @@ class CosmosClientConnection(object):  # pylint: disable=too-many-public-methods
         CosmosClientConnection.__ValidateResource(udf)
         udf = udf.copy()
         if udf.get("serverScript"):
-            udf["body"] = str(udf["serverScript"])
+            udf["body"] = str(udf.pop("serverScript", ""))
         elif udf.get("body"):
             udf["body"] = str(udf["body"])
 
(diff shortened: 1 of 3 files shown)


mxkus (approved):
I have created the corresponding bug and to me, the solution seems correct.

sdk/cosmos/azure-cosmos/azure/cosmos/aio/_cosmos_client_connection_async.py
```diff
@@ -1105,7 +1105,7 @@ async def ReplaceTrigger(self, trigger_link, trigger, options=None, **kwargs):
         CosmosClientConnection.__ValidateResource(trigger)
         trigger = trigger.copy()
         if trigger.get("serverScript"):
-            trigger["body"] = str(trigger["serverScript"])
+            trigger["body"] = str(trigger.pop("serverScript", ""))
```
xinlian12, line 1108: Do we have any test for this change?

simorenoh: the test would probably end up being an end to end test on every method for stored procedures, triggers, and udfs, which does not exist as of now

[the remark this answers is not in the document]
simorenoh: yep, along with any other fields the body has, it just seems like the backend does not expect the body to have a serverScript field present and instead it's the sdk that parses/ makes sense of it

xinlian12: yea, it would be good to add some tests to make sure in future, no regression happens. Will not block on this PR, but good to have a working item to track. Thanks

xinlian12 (approved):
LGTM, thanks

PR #23089, opened 2022-02-16
[recoveryservicesbackup] fix multiple pages paging

iscai-msft:
Patching the generated code to fix multiple pages paging by adding a custom policy to remove duplicate query params

### DIFF
--- a/sdk/recoveryservices/azure-mgmt-recoveryservicesbackup/azure/mgmt/recoveryservicesbackup/activestamp/_patch.py
+++ b/sdk/recoveryservices/azure-mgmt-recoveryservicesbackup/azure/mgmt/recoveryservicesbackup/activestamp/_patch.py
@@ -51,6 +51,7 @@ class RemoveDuplicateParamsPolicy(SansIOHTTPPolicy):
 DUPLICATE_PARAMS_POLICY = RemoveDuplicateParamsPolicy(duplicate_param_names=["$filter", "$skiptoken", "api-version"])
 
 class RecoveryServicesBackupClient(RecoveryServicesBackupClientGenerated):
+    __doc__ = RecoveryServicesBackupClientGenerated.__doc__
     def __init__(
         self,
         credential: "TokenCredential",
--- a/sdk/recoveryservices/azure-mgmt-recoveryservicesbackup/azure/mgmt/recoveryservicesbackup/activestamp/_patch.py
+++ b/sdk/recoveryservices/azure-mgmt-recoveryservicesbackup/azure/mgmt/recoveryservicesbackup/activestamp/_patch.py
@@ -24,8 +24,42 @@
 # IN THE SOFTWARE.
 #
 # --------------------------------------------------------------------------
+import copy
+from typing import List
+import importlib
+import urllib.parse
+from ._recovery_services_backup_client import RecoveryServicesBackupClient as RecoveryServicesBackupClientGenerated
+from azure.core.pipeline.policies import SansIOHTTPPolicy
+
+class RemoveDuplicateParamsPolicy(SansIOHTTPPolicy):
+    def __init__(self, duplicate_param_names):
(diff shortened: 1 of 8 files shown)


sdk/recoveryservices/azure-mgmt-recoveryservicesbackup/azure/mgmt/recoveryservicesbackup/activestamp/_patch.py
```diff
@@ -24,8 +24,42 @@
...
+        # type: (List[str]) -> None
+        self.duplicate_param_names = duplicate_param_names
+
+    def on_request(self, request):
+        parsed_url = urllib.parse.urlparse(request.http_request.url)
+        query_params = urllib.parse.parse_qs(parsed_url.query)
+        updated_query_params = copy.copy(query_params)
+        for query in query_params:
+            if query in self.duplicate_param_names and (isinstance(query_params[query], list) and len(query_params[query]) > 1):
+                updated_query_params[query] = [query_params[query][-1]]
```
johanste, line 45: Do we want the first or the last value? This grabs the last value. And when would you not get a `list` as the value?

Python nitpick:
You may consider using dict comprehension for this:

```python
{ key: val[-1:] if key in self.duplicate_names else val for key, val in query_params.items()}
``` 

...although it is arguably not easier to understand compared to what you have now....

iscai-msft: the last value. urllib parse returns the queries in the order they are in the url. The way we end up formatting, we format at the beginning of the query, so the last value is going to be the one the service returned.

Yeah let me try with python dict comprehension, i thought i'd be ok with doing copy to keep it cleaner but it's bothering me more and more. thanks!

iscai-msft: updated!

johanste (approved):

msyyc:
please decompress the zip file to get private .whl file: 
[azure_mgmt_recoveryservicesbackup-4.1.0-py3-none-any.zip](https://github.com/Azure/azure-sdk-for-python/files/8085902/azure_mgmt_recoveryservicesbackup-4.1.0-py3-none-any.zip)

sdk/recoveryservices/azure-mgmt-recoveryservicesbackup/azure/mgmt/recoveryservicesbackup/activestamp/_patch.py
```diff
@@ -24,8 +24,40 @@
...
+            k: v[-1:] if k in self.duplicate_param_names else v
+            for k, v in query_params.items()
+        }
+        request.http_request.url = request.http_request.url.replace(parsed_url.query, "") + urllib.parse.urlencode(filtered_query_params, doseq=True)
+        return super().on_request(request)
+
+DUPLICATE_PARAMS_POLICY = RemoveDuplicateParamsPolicy(duplicate_param_names=["$filter", "$skiptoken", "api-version"])
+
+class RecoveryServicesBackupClient(RecoveryServicesBackupClientGenerated):
+    def __init__(self, credential, *args, **kwargs):
```
msyyc, line 51: We have to make `credential` clear because [azure-cli](https://github.com/Azure/azure-cli/blob/67568014e3df1f183b92ee1ad7699b3b86f35dc0/src/azure-cli-core/azure/cli/core/util.py#L631-L637).

iscai-msft: Let me update the other client too

msyyc (approved):

PR #23114, opened 2022-02-17
[core] fix mypy errors arising from latest mypy version

kristapratico:
Output using mypy==0.931 and Python 3.10

azure\core\settings.py:164: error: Incompatible return value type (got "Union[str, Type[AbstractSpan]]", expected "Optional[Type[AbstractSpan]]")
azure\core_pipeline_client.py:31: error: Module "collections" has no attribute "Iterable"
azure\core_pipeline_client.py:31: error: Name "Iterable" already defined (possibly by an import)


xiangyan99 (approved):
LGTM

Issue #21142, opened 2021-10-07
[AMQP Python] async performance tuning


yunhaoling:
- sending
- receiving
https://github.com/Azure/azure-sdk-for-python/issues/21143

todo: async recv perf improvement

yunhaoling:
after improvement

sending:
![image](https://user-images.githubusercontent.com/47871814/156045862-9ea8ea75-3ba5-4843-ac56-261e895648f3.png)

receiving:
![image](https://user-images.githubusercontent.com/47871814/156045893-60e1fb22-53fc-4c81-850c-f4dbabc364a8.png)
![image](https://user-images.githubusercontent.com/47871814/156045904-3c7d6b23-8575-4548-87ed-e9b55f513eb9.png)

PR #23122, opened 2022-02-18
[EH Pyproto] Async recv perf improvement

yunhaoling:
addressing issue: https://github.com/Azure/azure-sdk-for-python/issues/21142
change async receive implementation to embrace async coroutine tasks:
  - one coroutine does polling the underlying amqp connection
  - one coroutine does calling callbacks

to review amqp pr, check PR: https://github.com/Azure/azure-uamqp-python/pull/311


sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_connection_async.py
```diff
@@ -451,11 +452,10 @@ async def listen(self, wait=False, batch=1, **kwargs):
...
                 return
             try:
-                tasks = [asyncio.ensure_future(self._listen_one_frame(**kwargs)) for _ in range(batch)]
-                await asyncio.gather(*tasks)
+                for _ in range(batch):
+                    await asyncio.ensure_future(self._listen_one_frame(**kwargs))
             except ValueError:
-                for task in tasks:
-                    task.cancel()
+                pass
```
annatisch, line 458: Silent failure?

sdk/eventhub/azure-eventhub/azure/eventhub/_version.py
```diff
@@ -3,4 +3,4 @@
 # Licensed under the MIT License.
 # ------------------------------------
 
-VERSION = "5.8.0b3"
+VERSION = "5.8.0a3"
```
annatisch, line 6: The version is going backwards?

annatisch: Or is this a result of the bot-auto-version-bump PR after the last release?

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_consumer_async.py
```diff
@@ -162,7 +165,8 @@ async def _open_with_retry(self) -> None:
         await self._do_retryable_operation(self._open, operation_need_param=False)
 
     async def _message_received(self, message: Message) -> None:
-        self._message_buffer.append(message)
+        async with self._message_buffer_lock:
+            self._message_buffer.append(message)
```
annatisch, line 169: What does the lock around this achieve?
We're no longer processing the incoming message batch concurrently with `gather` - so shouldn't that remove the possibility of a race condition?

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_consumer_async.py
```diff
@@ -171,54 +175,67 @@ def _next_message_in_buffer(self):
...
+                        self._name,
+                        last_exception,
+                    )
+                    raise last_exception
+
+    async def receive(self, batch=False, max_batch_size=300, max_wait_time=None) -> None:
+        self._callback_task_run = True
+        self._last_callback_called_time = time.time()
+        callback_task = asyncio.ensure_future(self._callback_coroutine(batch, max_batch_size, max_wait_time))
+        receive_task = asyncio.ensure_future(self._receive_coroutine())
```
annatisch, line 235: Nitpick: Not sure I'm a fan of using `_coroutine` as a method name suffix....

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_consumer_async.py
```diff
@@ -171,54 +175,67 @@ def _next_message_in_buffer(self):
...
+                    self._last_callback_called_time = now_time
+                else:
+                    if max_wait_time and (now_time - self._last_callback_called_time) > max_wait_time:
+                        # no events received, and need to callback
+                        await self._on_event_received([] if batch else None)
+                        self._last_callback_called_time = now_time
+                    # backoff a bit to avoid throttling CPU when no events are coming
+                    await asyncio.sleep(0.05)
+            except asyncio.CancelledError:
+                raise
```
annatisch, line 198: Immediately raising kinda feels like this is redundant?

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_consumer_async.py
```diff
@@ -171,54 +175,67 @@ def _next_message_in_buffer(self):
...
-                await self._on_event_received([])
-            else:
-                await self._on_event_received(None)
+                ):
+                    raise await self._handle_exception(exception)
+                if not self.running:  # exit by close
+                    return
+                if self._last_received_event:
+                    self._offset = self._last_received_event.offset
+                last_exception = await self._handle_exception(exception)
```
annatisch, line 221: Do we think it could be a problem that these are now handled from a separate coroutine to the on-event callbacks? Could it lead someone to have to start locking things?

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_consumer_async.py
```diff
@@ -171,54 +175,67 @@ def _next_message_in_buffer(self):
...
-                try:
-                    await self._open()
-                    await cast(ReceiveClientAsync, self._handler).do_work_async(batch=self._prefetch)
-                    break
-                except asyncio.CancelledError:  # pylint: disable=try-except-raise
-                    raise
-                except Exception as exception:  # pylint: disable=broad-except
-                    if (
+        retried_times = 0
+        while retried_times <= max_retries:
```
annatisch, line 205: This was previously wrapped within a larger `while` loop - I'm not sure we want to just exit the receive loop after max_retries?

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_consumer_async.py
yunhaoling: we have two coroutines interacting with self._message_buffer simultaneously -- receive task keeps adding messages while callback task keeps pumping messages, race condition between the two tasks is what I'm trying to avoid here.

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_consumer_async.py
yunhaoling: I have changed the receiving implementation from previous polling to current coroutine task.

the receive means differently in two implementations:

previous: the receive acts as a single call, event processor keeps calling receive,
now: receive itself a coroutine task, which do while loop (retry)

in both implementations, if retry exhausted and error raised, error would be raised. so actually the code is doing loop or raise error, it doesn't exit the loop

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_consumer_async.py
yunhaoling: I could just rename it as `_task` :P

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_consumer_async.py
yunhaoling: I don't think so..

I think for user it's always there's one coroutine calls on_event callbacks.

The internal buffered queue is for decoupling receiving message + calling callbacks.

how we receive messages and dumps into a local queue is an implementation detail.
It happened that previously we do everything in a single coroutine, but now we split out the receiving part. however, the callback part for a user doesn't change.

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_consumer_async.py
yunhaoling: oh yes, I feel this line could be removed.
thanks for pointing out.

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_connection_async.py
yunhaoling: I have updated the code to align with the sync implementation (break instead of catching error by the returning value), perf is not impacted.

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_consumer_async.py
yunhaoling: updated to `_task`

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_consumer_async.py
annatisch: We could investigate this further in term of optimization - it seem expensive to await the lock for every individual message.

annatisch (approved):

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_consumer_async.py
yunhaoling: yeah.. it is interesting that actually in my v0, I indeed tried to batch things, however, I guess I didn't do thing correctly that the perf is not good.

I can put a todo for this

PR #23148, opened 2022-02-19
[Test Proxy] Add custom matcher setter and ignore unnecessary headers

mccoyp:
# Description

**For reviewers (@scbedd):** The only relevant files for test proxy work are under `devtools_testutils`.

**For package owners:** The only changes to affected libraries are to invocations of `set_bodiless_matcher`.

This adds a `set_custom_default_matcher` method for setting custom matchers, which was motivated by the case where "Authorization" headers get dropped in playback due to challenge authentication cache misses (relevant to KV and ACR).

While looking into this issue, though, it became apparent that we don't actually want to record some headers in the first place -- they were explicitly filtered out in our old `vcrpy` setup because they can contain sensitive information, interfere with matching, and/or take up recording space without a purpose. So, this PR adds a call to the `RemoveHeaderSanitizer` as part of test proxy startup to remove some headers -- `Authorization`, `x-ms-client-request-id`, and `x-ms-request-id`, in particular -- from recordings.

This also sets up a custom default matcher that excludes these headers, since tests will otherwise fail if the header removal sanitizer is set up but the headers exist in recordings (which they will, for all current recordings). I looked into only setting up this sanitizer in live mode, but we need it in playback since playback requests send these headers and the proxy will get upset over this by default. Removing the headers from each recorded test is infeasible because of how many there are, and this solution works when existing references to `set_bodiless_matcher` are replaced with a call to `set_custom_default_matcher` that turns off body matching _and_ excludes the sensitive headers.

Whenever tests are re-recorded, they'll now omit these headers. So, using a plain `set_bodiless_matcher` at that point will continue to work like it did before this PR. Eventually, once these headers are removed from all or most recordings, we'll be able to remove the call to `set_custom_default_matcher` in `proxy_startup.py`. Or, at the very least, we can update recordings library-by-library after this PR is merged to avoid a monstrously large PR today.

- _N/A_ Pull request includes test coverage for the included changes.
  - Changes are verified by running existing, migrated test suites

### DIFF
--- a/sdk/eventgrid/azure-eventgrid/tests/conftest.py
+++ b/sdk/eventgrid/azure-eventgrid/tests/conftest.py
@@ -27,7 +27,7 @@ import platform
 import sys
 import pytest
 from devtools_testutils import test_proxy
-from devtools_testutils.sanitizers import add_remove_header_sanitizer, add_general_regex_sanitizer, set_bodiless_matcher
+from devtools_testutils.sanitizers import add_remove_header_sanitizer, add_general_regex_sanitizer, set_custom_default_matcher
 
 # Ignore async tests for Python < 3.5
 collect_ignore_glob = []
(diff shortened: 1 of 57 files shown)


scbedd (approved):
@mccoyp I really appreciate the move of get_test_id and the associated recording_id tracking to `helper` away from `proxy_testcase`.  I was hitting circular import errors when PRing a different change, and this little move addresses my problems perfectly. 👍 👍 👍 

In general, I see where you're coming from RE: backwards compat. I recommend filing an issue to track the removal of the set-default after the recordings have been updated to pull the content out.

mccoyp:
Issue tracking the removal of `set_custom_default_matcher` inside `proxy_startup.py`: https://github.com/Azure/azure-sdk-for-python/issues/23592

mccoyp:
Issue tracking the removal of `set_custom_default_matcher` inside `proxy_startup.py`: https://github.com/Azure/azure-sdk-for-python/issues/23592

PR #23171, opened 2022-02-22
update pinned mypy version

kristapratico:
Updates our pinned mypy version to 0.931 and includes the third-party package stubs needed (these stopped shipping with mypy since >0.9XX). Includes some mypy fixes in libraries which started raising new errors with this version of mypy/python.


swathipil (approved):
EH/SB LGTM!

PR #23179, opened 2022-02-23
[Storage] Add missing SAS permissions to Storage packages

### LINKED ISSUE #8799: [Storage]Add all Missing SAS Permissions
We have some missing sas permissions for container
check sas for blob, queue and file(fileshare and file datalake) to make sure all permissions are provided in sdk.
https://docs.microsoft.com/en-us/rest/api/storageservices/create-service-sas#permissions-for-a-directory-container-or-blob

jalauzon-msft:
Resolves #8799.
Updating blob, file-datalake, and file-share with all missing SAS permissions and/or updating the SAS documentation.


sdk/storage/azure-storage-blob/azure/storage/blob/_shared_access_signature.py
```diff
@@ -66,13 +66,14 @@ def generate_blob(self, container_name, blob_name, snapshot=None, version_id=Non
         :param str snapshot:
             The snapshot parameter is an opaque DateTime value that,
             when present, specifies the blob snapshot to grant permission.
-        :param BlobSasPermissions permission:
```
tasherif-msft, line 69: why are we removing these?

jalauzon-msft: Just moving the type to the below `:type permission:` tag since I added an `or` to the type. It seems like the style we follow is, if the type has an `or` we have it declared with a separate `type` line.

NOTE: `str` has always been supported, just updated the docs to reflect it.

tasherif-msft (approved):

PR #23181, opened 2022-02-24
[ACR] Fix cspell errors

YalinLi0312:
For issue: https://github.com/Azure/azure-sdk-for-python/issues/22643


sdk/containerregistry/azure-containerregistry/azure/containerregistry/_models.py
```diff
@@ -206,7 +206,7 @@ def created_on(self):
         return self._created_on
 
     @property
-    def last_udpated_on(self):
+    def last_updated_on(self):
```
kristapratico, line 209: nit: You might want to include this as a bug fix on the changelog since it looks like part of the public surface.

kristapratico (approved):

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_models.py
YalinLi0312: Thanks for your review @kristapratico ! Have updated the changelog. :)

PR #23188, opened 2022-02-24
[KeyVault Keys] Fix cspell errors

YalinLi0312:
For issue: https://github.com/Azure/azure-sdk-for-python/issues/22662


danieljurek (changes requested):
In the case of the collections of random characters like `xcau` these look like they're all concentrated in a [single file](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/keyvault/azure-keyvault-keys/tests/keys.py) . 

Please exclude that entire file instead of adding exceptions for those collections of characters.

While it's possible that spelling errors may appear in that one file, the impact is lower because this is a test file.

YalinLi0312:
Have created a new one with clear commit history.

PR #22564, opened 2022-01-19
[Tables] Add the support for Python 3.6 back

YalinLi0312:

PR #23190, opened 2022-02-24
[Tables] Resolve dependency conflicts

YalinLi0312:
The conflict is from `azure-mgmt-core` where requires `azure-core` version to be at least **1.15.0**.
This is where the error was found: https://dev.azure.com/azure-sdk/public/_build/results?buildId=1386229&view=logs&j=4fa73ede-de59-50b0-851d-db96582ea108&t=813eb0cd-e8db-59f1-2e13-b4deadd36a36&l=2762
Thanks @kristapratico for letting me know!

### DIFF
--- a/sdk/tables/azure-data-tables/setup.py
+++ b/sdk/tables/azure-data-tables/setup.py
@@ -63,7 +63,7 @@ setup(
     ]),
     python_requires=">=3.7",
     install_requires=[
-        "azure-core<2.0.0,>=1.14.0",
+        "azure-core<2.0.0,>=1.15.0",
         "msrest>=0.6.21"
     ],
 )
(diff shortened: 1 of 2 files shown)


sdk/tables/azure-data-tables/setup.py
```diff
@@ -63,7 +63,7 @@
     ]),
     python_requires=">=3.7",
```
kristapratico, line 64: unrelated to this PR, but the setup should be updated to support 3.6 and 3.10

kristapratico (approved):
Thanks! The frozen requirement in the shared_requirements.txt just needs to be updated to pass the analyze check.

sdk/tables/azure-data-tables/setup.py
annatisch: Got spot - we should fix that and the classifiers while we're here.

annatisch: The 3.6 support is in a separate PR - so I'll merge that after this one.
https://github.com/Azure/azure-sdk-for-python/pull/22564

PR #23202, opened 2022-02-24
Don't pass loop if not defined

lmazuel:
Fix https://github.com/Azure/azure-sdk-for-python/issues/23196


tjprescott (approved):
This makes sense to me.

PR #23270, opened 2022-03-01
Update fabricbot.json to have new contacts for services

lilyjma:
Changed GitHub contacts for ARM, ARM subservices, and API Management per request.

### DIFF
--- a/.github/fabricbot.json
+++ b/.github/fabricbot.json
@@ -1356,7 +1356,8 @@
             "ARM"
           ],
           "mentionees": [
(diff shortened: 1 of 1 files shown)


jsquire (commented):
Changes here look good, but we'll also want to reflect these changes in the [CODEOWNERS](https://github.com/Azure/azure-sdk-for-python/blob/main/.github/CODEOWNERS#L171) so that if the rule gets regenerated, the changes are not lost.

Issue #23020, opened 2022-02-10
Release messages from internal buffer when there is no active receive call


ramya-rao-a:
Take the below scenario:
-  `ServiceBusReceiver.receive_messages()` is used to ask for 100 messages, but only 30 came over the wire before timeout. We give these 30 messages to the caller. 
- 40 messages come over the wire before the next call to `receive_messages`

If there is too much of a gap between the two calls to `receive_messages` then these 40 messages will be expired. Due to this the caller will get the lock lost error when settling these messages and will get the same messages again.

This is an expected loss if the user has chosen to use prefetch. But in the absence of prefetch, the client should not be holding on to messages if there are no receive calls from the user.

This feature request is to [release such messages](http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-messaging-v1.0-os.html#type-released) which will not affect the delivery count either.

annatisch:
I've done some initial testing here.
Our Python SDK design is slightly different to the .NET/Java SDK designs where prefetch is achieved via a background thread.
In the Python single-threaded design, extra messages can be received outside the context of the `receive_messages` call in the following circumstances:
- The link credit on the link is set according to the `max_message_count` value.
- Not all the messages allowed by the credit value are received at once (meaning there's extra credit pending on the connection)
- The connection is then used outside of the `receive_messages` context, e.g. for different operations like sending, renewing tokens, sending empty keep-alive frames etc.

This can allow that remaining link credit to be used up receiving extra messages that will then sit in the buffer until the next time `receive_messages` is called, and they will potentially be expired.

One possible solution for this would be to, if prefetch=0, set the `handler._on_message_received_callback` to function capable of releasing incoming messages when not in a receiving context.

yunhaoling:
my investigation based on @annatisch 's approach:

1. message releasing only works when the message lock hasn't been expired yet. release message after lock expiration still increases the delivery count.
2. messages could arrive at socket when the client is not active.
3. those messages will stay in the socket unable to be handled until there're method calls (e.g. receive) triggering the transport layer to consume bytes from socket, it is until that moment that we are able to deal with messages.

as a fix, we could have a background task which keeps listening on the socket and consume messages, we release messages when messages not arriving in a receive context.

in my draft pr, I overload the keep-alive thread to listen on socket periodically, however, it may not be a best practice as
- we violate single responsibility principle
- we increase cpu workload with shorter checking interval

PR #23285, opened 2022-03-02
[ServiceBus] release message in background

yunhaoling:
addressing issue: https://github.com/Azure/azure-sdk-for-python/issues/23020

based on @annatisch 's approach: https://github.com/Azure/azure-sdk-for-python/pull/23212

how it works:

- overload the keep-alive background task for polling socket when prefetch is 1
- using threading.event for indication of whether a message arrives in a receiving context or non-receiving context

cons:
overloading keep_alive result in bad-perf, increase cpu workload.

### DIFF
--- a/sdk/servicebus/azure-servicebus/azure/servicebus/_common/receiver_mixins.py
+++ b/sdk/servicebus/azure-servicebus/azure/servicebus/_common/receiver_mixins.py
@@ -85,7 +85,6 @@ class ReceiverMixin(object):  # pylint: disable=too-many-instance-attributes
                 "Messages received in RECEIVE_AND_DELETE receive mode cannot have their locks removed "
                 "as they have been deleted, providing an AutoLockRenewer in this mode is invalid."
             )
-        self._receive_context = threading.Event()
 
     def _build_message(self, received, message_type=ServiceBusReceivedMessage):
         message = message_type(
--- a/sdk/servicebus/azure-servicebus/azure/servicebus/_common/receiver_mixins.py
+++ b/sdk/servicebus/azure-servicebus/azure/servicebus/_common/receiver_mixins.py
@@ -5,7 +5,6 @@
 # -------------------------------------------------------------------------
 import uuid
 import functools
-import threading
 from typing import Optional, Callable
 
 from uamqp import Source
--- a/sdk/servicebus/azure-servicebus/azure/servicebus/_common/receiver_mixins.py
+++ b/sdk/servicebus/azure-servicebus/azure/servicebus/_common/receiver_mixins.py
@@ -178,3 +178,11 @@ class ReceiverMixin(object):  # pylint: disable=too-many-instance-attributes
     def _populate_message_properties(self, message):
         if self._session:
             message[MGMT_REQUEST_SESSION_ID] = self._session_id
+
+    def _enhanced_message_received(self, message):
+        # pylint: disable=protected-access
+        self._handler._was_message_received = True
+        if self._receive_context.is_set():
+            self._handler._received_messages.put(message)
+        else:
+            message.release()
--- a/sdk/servicebus/azure-servicebus/azure/servicebus/_common/receiver_mixins.py
+++ b/sdk/servicebus/azure-servicebus/azure/servicebus/_common/receiver_mixins.py
@@ -5,6 +5,7 @@
 # -------------------------------------------------------------------------
 import uuid
 import functools
+import threading
 from typing import Optional, Callable
 
 from uamqp import Source
@@ -84,6 +85,7 @@ class ReceiverMixin(object):  # pylint: disable=too-many-instance-attributes
                 "Messages received in RECEIVE_AND_DELETE receive mode cannot have their locks removed "
                 "as they have been deleted, providing an AutoLockRenewer in this mode is invalid."
             )
+        self._receive_context = threading.Event()
 
     def _build_message(self, received, message_type=ServiceBusReceivedMessage):
         message = message_type(
--- a/sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_receiver.py
+++ b/sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_receiver.py
@@ -199,7 +199,6 @@ class ServiceBusReceiver(
         self._session = (
             None if self._session_id is None else ServiceBusSession(self._session_id, self)
         )
-        self._keep_receiving_thread = None
         self._receive_context = threading.Event()
 
     def __iter__(self):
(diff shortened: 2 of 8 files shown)


sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_receiver.py
```diff
@@ -363,68 +371,77 @@ def _open(self):
...
-                    first_message_received = True
-                    abs_timeout_ms = (
-                        amqp_receive_client._counter.get_current_ms()
-                        + self._further_pull_receive_timeout_ms
-                    )
-            while (
-                not received_messages_queue.empty() and len(batch) < max_message_count
-            ):
+        try:
+            self._receive_context.set()
```
annatisch, line 384: Do we care that we're setting/clearing the threading Event regardless of whether we're actually using it (i.e. the callback is only replaced if prefetch is disabled)?
Probably not... doubtful that it would have much of a perf impact?

sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_receiver.py
```diff
@@ -246,18 +247,22 @@ def __next__(self):
...
+            message = self._build_message(uamqp_message)
+            if (
+                self._auto_lock_renewer
+                and not self._session
+                and self._receive_mode != ServiceBusReceiveMode.RECEIVE_AND_DELETE
+            ):
+                self._auto_lock_renewer.register(self, message)
+            return message
+        finally:
+            self._receive_context.clear()
```
annatisch, line 265: If I recall correctly, we had a system in place to allow for iterator receiving and batch message receiving from the same receiver.... is that correct?
And if so - how well is that handled when they're both sharing the same threading.Event?

sdk/servicebus/azure-servicebus/azure/servicebus/_common/receiver_mixins.py
```diff
@@ -84,6 +85,7 @@ def _populate_attributes(self, **kwargs):
                 "Messages received in RECEIVE_AND_DELETE receive mode cannot have their locks removed "
                 "as they have been deleted, providing an AutoLockRenewer in this mode is invalid."
             )
+        self._receive_context = threading.Event()
```
annatisch, line 88: Do we need an equivalent fix for async?

sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_receiver.py
yunhaoling: tested multiple scenarios (both sync and async, iterator with prefetch1 vs 300, receive_messages with max_message_count 1 vs 300), compare the perf results with currently GAed lib and the perf impact is neglectable (almost no per diff from what I observed)

sdk/servicebus/azure-servicebus/azure/servicebus/_common/receiver_mixins.py
```diff
@@ -176,3 +178,11 @@ def _on_attach(self, source, target, properties, error):
     def _populate_message_properties(self, message):
         if self._session:
             message[MGMT_REQUEST_SESSION_ID] = self._session_id
+
+    def _enhanced_message_received(self, message):
```
annatisch, line 182: We should take notes for how we can make this kind of fix easier in the new pyamqp....
For example have the `on_message_received` callback be more easily replaceable without having to set all these private internal attributes......

sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_receiver.py
```diff
@@ -546,6 +568,16 @@ def _close_handler(self):
         self._message_iter = None
         super(ServiceBusReceiver, self)._close_handler()
 
+    def _keep_receiving(self):
+        while not self._shutdown.is_set() and self._handler:
+            try:
+                self._handler._connection.work()  # pylint: disable=protected-access
+                time.sleep(5)
+                _LOGGER.debug("Keep receiving alive")
```
annatisch, line 576: Maybe this should be something like "Keep connection alive"

sdk/servicebus/azure-servicebus/azure/servicebus/aio/_servicebus_receiver_async.py
```diff
@@ -236,10 +239,14 @@ async def _inner_anext(self):
                 raise
 
     async def __anext__(self):
-        message = await self._inner_anext()
-        links = get_receive_links(message)
-        with receive_trace_context_manager(self, links=links):
-            return message
+        try:
+            self._receive_context.set()
```
annatisch, line 243: Is this not using an `asyncio.Event`?

yunhaoling: nope, I'm reusing the threading.Event because the only feature we used is `set` and `clear` which threading.Event seem to satisfy our goal here.

but since you called it out, I think we could change it to asyncio.Event for keep async consistency.

sdk/servicebus/azure-servicebus/azure/servicebus/_common/receiver_mixins.py
yunhaoling: async fixed added!

sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_receiver.py
yunhaoling: both would use the same threading.Event.

execution in sequential order is fine as we're doing set and clear in sequential order..

however, execution in nested situation/parallel execution would be a problem.. okay, I think I need to change it to counter variable instead of this threading.Event..

thanks for raising your concern!

sdk/servicebus/azure-servicebus/azure/servicebus/_common/receiver_mixins.py
yunhaoling: totally agreed, will add notes!

sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_receiver.py
yunhaoling: yep, that sounds good to me

sdk/servicebus/azure-servicebus/azure/servicebus/aio/_servicebus_receiver_async.py
yunhaoling: code updated to asyncio.Event

sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_receiver.py
annatisch: Do we have any tests that use both iterator an batch at the same time?

sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_receiver.py
```diff
@@ -546,6 +569,16 @@ def _close_handler(self):
         self._message_iter = None
         super(ServiceBusReceiver, self)._close_handler()
 
+    def _keep_receiving(self):
```
annatisch, line 572: Isn't this essentially just the same as setting the existing `keep_alive_interval` setting to 5?
Why did we need to replace the keep-alive function?

sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_receiver.py
yunhaoling: test added for iterator + iterator and iterator + batch

annatisch (approved):

PR #23286, opened 2022-03-02
[Key Vault] Arch board feedback and language alignment

### LINKED ISSUE #22981: [Key Vault] Address remaining architect feedback for keys
Per the most recent round of [feedback](https://apiview.dev/Assemblies/Review/2fe1d980066340af8e70b56592afe76a?diffRevisionId=a617921fa5c74b1c99d169f84f9ab047&doc=False&diffOnly=False) for `azure-keyvault-keys`, we need to take a look at a few more pieces of feedback before the upcoming stable release:

- [ ] Move the optional `version` parameter in `KeyClient.release_key` into keyword-only arguments
- [ ] Try to get `update_key_rotation_policy` and `get_key_rotation_policy` to round-trip
  - "If I `get` a rotation policy - I should be able to pass that back in again, with optional `kwargs` to overwrite specific attributes."
- [ ] Compare key encryption algorithm enums with `cryptography` names and align where possible

mccoyp:
# Description

Resolves #22981.


sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_enums.py
```diff
@@ -21,9 +21,9 @@ class KeyCurveName(with_metaclass(CaseInsensitiveEnumMeta, str, Enum)):
 class KeyExportEncryptionAlgorithm(with_metaclass(CaseInsensitiveEnumMeta, str, Enum)):
     """Supported algorithms for protecting exported key material"""
 
-    CKM_RSA_AES_KEY_WRAP = "CKM_RSA_AES_KEY_WRAP"
-    RSA_AES_KEY_WRAP_256 = "RSA_AES_KEY_WRAP_256"
-    RSA_AES_KEY_WRAP_384 = "RSA_AES_KEY_WRAP_384"
+    ckm_rsa_aes_key_wrap = "CKM_RSA_AES_KEY_WRAP"
```
seankane-msft, line 24: Why the change from upper case to lower case? I thought Python enums were supposed to be [capital letters](https://azure.github.io/azure-sdk/python_design.html#enumerations)

mccoyp: They are! Unfortunately we had already GA-ed with all lower-case in all packages except for `azure-keyvault-administration`. Anna suggested that we keep enum values consistent within a package (so lower-case here), but that continuing with proper upper-casing in `administration` is preferred

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_client.py
```diff
@@ -709,10 +708,11 @@ def release_key(self, name, target_attestation_token, version=None, **kwargs):
         :rtype: ~azure.keyvault.keys.ReleaseKeyResult
         :raises: :class:`~azure.core.exceptions.HttpResponseError`
         """
+        version = kwargs.pop("version", "")
         result = self._client.release(
             vault_base_url=self._vault_url,
             key_name=name,
-            key_version=version or "",
+            key_version=version,
```
YalinLi0312, line 715: Does the release function choose the latest version when the version is empty as I don't find it to take any action here?

mccoyp: Yes -- I think it's a swagger quirk that `key_version` is a required parameter for a couple of functions. Key Vault will use the default version when the version isn't specified, but because the parameter is required, we need to send an empty string instead of `None`.

maorleger: its because there's no great way to model optional path params (which key_version ends up being). So they are "required" in swagger but optional in reality. At least that's my understanding of the reasoning why...

maorleger (approved):

sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_models.py
```diff
@@ -320,18 +320,18 @@ class KeyRotationPolicy(object):
...
     :ivar updated_on: When the policy was last updated, in UTC
     :type updated_on: ~datetime.datetime
     """
 
-    def __init__(self, policy_id, **kwargs):
-        # type: (str, **Any) -> None
-        self.id = policy_id
+    def __init__(self, **kwargs):
+        # type: (**Any) -> None
+        self.id = kwargs.get("policy_id", None)
```
heaths, line 334: Shouldn't this just be `id`?

mccoyp: I would like it to be, but `id` is a built-in function in Python so linters get upset when `id` is used as a parameter name (even though it's acceptable as an attribute). We have to do similarly elsewhere, like accepting `key_id` in [KeyProperties](https://github.com/Azure/azure-sdk-for-python/blob/94dff336d6233be9fa067745cb4b9606c6c42479/sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_models.py#L67).

PR #23294, opened 2022-03-03
[Storage] Address comments from API Review for March release

jalauzon-msft:


sdk/storage/azure-storage-blob/azure/storage/blob/_models.py
```diff
@@ -988,13 +988,13 @@ def from_string(cls, permission):
...
         p_tag = 't' in permission
-        p_find = 'f' in permission
+        p_filter_by_tags = 'f' in permission
         p_move = 'm' in permission
         p_execute = 'e' in permission
         p_set_immutability_policy = 'i' in permission
         parsed = cls(read=p_read, write=p_write, delete=p_delete, list=p_list,
                      delete_previous_version=p_delete_previous_version, tag=p_tag, add=p_add,
-                     create=p_create, permanent_delete=p_permanent_delete, find=p_find, move=p_move,
+                     create=p_create, permanent_delete=p_permanent_delete, find=p_filter_by_tags, move=p_move,
```
annatisch, line 997: Shouldn't this be `filter_by_tags=p_filter_by_tags`?

sdk/storage/azure-storage-blob/azure/storage/blob/_models.py
```diff
@@ -923,7 +923,7 @@ class ContainerSasPermissions(object):
         Write a new blob, snapshot a blob, or copy a blob to a new blob.
     :keyword bool permanent_delete:
         To enable permanent delete on the blob is permitted.
-    :keyword bool find:
+    :keyword bool filter_by_tags:
```
annatisch, line 926: Is this going to be consistent with Storage reference docs?

sdk/storage/azure-storage-blob/azure/storage/blob/_models.py
jalauzon-msft: Good catch! Will update pending response to other comment :)

sdk/storage/azure-storage-blob/azure/storage/blob/_models.py
jalauzon-msft: So the storage reference docs have this as 'find' and that's why I originally went with that BUT this 'f' permission already exists on an account SAS and we called it `filter_by_tags`. It is the same permission just here it is specific to a container.

What do you think the best thing to do here is? Match the existing code or match the [docs](https://docs.microsoft.com/en-us/rest/api/storageservices/create-service-sas)

annatisch: In that case let's go with `filter_by_tags`. It will be consistent with the existing API, and also is more descriptive in terms of how it will be used/what permissions it's actually giving you :)

sdk/storage/azure-storage-file-share/azure/storage/fileshare/_directory_client.py
```diff
@@ -367,7 +367,7 @@ def rename_directory(
         :keyword file_last_write_time:
             Last write time for the file.
         :paramtype file_last_write_time:~datetime.datetime or str
-        :keyword dict(str,str) metadata:
+        :keyword Dict(str,str) metadata:
```
annatisch, line 370: Square brackets :) `Dict[str, str]`

annatisch (approved):

PR #23315, opened 2022-03-03
[textanalytics] release prep + fixing docstrings and links

kristapratico:


iscai-msft (approved):

kristapratico:
@iscai-msft sorry I think I'm going to have to revert that change to create a type alias for the inputs - it's breaking intellisense in VSCode:

image

I think I'd rather have a better experience for our users than have a type for our own convenience. 

Oddly it does work with Pycharm:

image

PR #23320, opened 2022-03-03
[Storage] Remove batch delete files feature for GA release

jalauzon-msft:
Temporarily removing the `delete_files()` feature for the GA release as it needs to be refactored before being released.


tasherif-msft (approved):

tasherif-msft:
@jalauzon-msft for future reference you can prefix with an `_` to say that something is "private" but commenting it out is also ok.

PR #23332, opened 2022-03-04
[pipeline] update existing README.md to drop py2

msyyc:
RE: Do the management plane libraries still support Python 2?

dependency: https://github.com/Azure/azure-sdk-for-python/pull/23333

PR #23374, opened 2022-03-07
update min version of azure-mgmt-core in shared reqs

kristapratico:
https://dev.azure.com/azure-sdk/public/_build/results?buildId=1414764&view=results

This PR: https://github.com/Azure/azure-sdk-for-python/pull/23332
 bumped the azure-mgmt-core version for most, but not all mgmt packages. The shared reqs needs to be updated to reflect that.


shared_requirements.txt
```diff
@@ -303,3 +303,24 @@ opentelemetry-sdk<2.0.0,>=1.5.0,!=1.10a0
 #override azure-mgmt-sqlvirtualmachine azure-mgmt-core>=1.3.0,<2.0.0
 #override azure-mgmt-rdbms msrest>=0.6.21
 #override azure-mgmt-rdbms azure-mgmt-core>=1.3.0,<2.0.0
```
mccoyp, line 305: We could remove a lot of these `azure-mgmt-core` overrides with this change; up to you!

mccoyp (approved):
Thank you for getting to this before me 🙂

shared_requirements.txt
kristapratico: good point, let me do that!

kristapratico: Analyze dependencies step is passing CI, merging this

xiangyan99:
I created same PR. https://github.com/Azure/azure-sdk-for-python/pull/23376:)

kristapratico:
> I created same PR. #23376:)

Oh no! I missed that. We're all so on top of it 😄

PR #23438, opened 2022-03-09
[SchemaRegistry] migrate to test proxy

### LINKED ISSUE #22517: [SchemaRegistry] Migrate SR tests to use test proxy

swathipil:
fixes #22517

- need to ignore pylint `import-error` due to same issue as here: https://github.com/PyCQA/pylint/issues/2862


sdk/schemaregistry/azure-schemaregistry-avroencoder/azure/__init__.py
```diff
@@ -1,26 +0,0 @@
-# --------------------------------------------------------------------------
```
yunhaoling, line 1: out of curiosity, the approach used here (removing __ init __.py) is called [native namespace package](https://packaging.python.org/en/latest/guides/packaging-namespace-packages/#native-namespace-packages)?

yunhaoling: I see, since we're now python 3 only, we are free to use the native nspkg pattern. nice!

yunhaoling (approved):
LGTM! thanks for doing this migration!

PR #23477, opened 2022-03-10
API Key Not Being Passed On Paged Requests

### LINKED ISSUE #23467: Async iteration over AsyncSearchItemPaged raises Forbidden on fetching next page
- **Package Name**: azure-search-documents 
- **Package Version**: 11.2.1
- **Operating System**: Ubuntu
- **Python Version**: 3.9.7

**Describe the bug**
When using the Async version of the SearchClient from azure-search-documents python SDK, I get an error when iterating over a Search result with more than 50 products. It looks like the error occurs when thz AsyncSearchItemPaged object tries to post for the next page. I then get a "HttpResponseError: Operation returned an invalid status 'Forbidden'"

**To Reproduce**
1. Create an index with at least 50 items
2. Instantiate a SearchClient from azure.search.documents.aio
3. Request the empty string to get all documents
4. iterate over the search results with: 
`async for x in results:
    print(x)`

**Expected behavior**
I would expect every result to be printed, or to get a way to iterate over the pages of results without getting an error.

kashifkhan:
Fixes #23467

The issue was caused by **kwargs not being passed in to `search_post` resulting in  an API call being without an `api key` in the header.

### DIFF
--- a/sdk/search/azure-search-documents/CHANGELOG.md
+++ b/sdk/search/azure-search-documents/CHANGELOG.md
@@ -7,7 +7,7 @@
 ### Breaking Changes
 
 ### Bugs Fixed
-- Bug fix for async paged requests to use api keys
+- Fixed issue where async `search` call would fail with a 403 error when retrieving large number of documents.
 
 ### Other Changes
 
--- a/sdk/search/azure-search-documents/CHANGELOG.md
+++ b/sdk/search/azure-search-documents/CHANGELOG.md
@@ -7,6 +7,7 @@
 ### Breaking Changes
 
 ### Bugs Fixed
+- Bug fix for async paged requests to use api keys
 
 ### Other Changes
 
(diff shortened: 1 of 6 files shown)


tjprescott (approved):
You should probably add a CHANGELOG entry for the bug fix.

sdk/search/azure-search-documents/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 ### Breaking Changes
 
 ### Bugs Fixed
+- Bug fix for async paged requests to use api keys
```
tjprescott, line 10: I'd recommend a more "customer facing" description. Something like: "Fixed issue where async list calls could fail with a 403 error when there was too much data to retrieve."

tjprescott (approved):

PR #23479, opened 2022-03-11
[Storage] Add remaining STG82 features for file-share

jalauzon-msft:
This change includes the following three features:
- Add `file_change_time` to APIs that support it
- Add `file_last_written_mode` to `upload_range()` and `upload_range_from_url()`
- Implement missing parameters for `create_directory()`. It seems the SMB properties were not previously added to `create_directory()` so this change adds those.


sdk/storage/azure-storage-file-share/azure/storage/fileshare/_directory_client.py
```diff
@@ -272,6 +272,33 @@ def create_directory(self, **kwargs):
         # type: (Any) -> Dict[str, Any]
         """Creates a new directory under the directory referenced by the client.
 
+        :keyword file_attributes:
```
tasherif-msft, line 275: if these attrs are applied to files/dirs - I think the kwarg `file-attributes` could be changed to something else? I'm not sure why in other kwargs we prefix it with `file` :/

sdk/storage/azure-storage-file-share/azure/storage/fileshare/_directory_client.py
```diff
@@ -272,6 +272,33 @@ def create_directory(self, **kwargs):
...
+            The file system attributes for files and directories.
+            If not set, the default value would be "none" and the attributes will be set to "Archive".
+            Here is an example for when the var type is str: 'Temporary|Archive'.
+            file_attributes value is not case sensitive.
+        :type file_attributes: str or :class:`~azure.storage.fileshare.NTFSAttributes`
+        :keyword file_creation_time:
```
tasherif-msft, line 281: same as above

sdk/storage/azure-storage-file-share/azure/storage/fileshare/_directory_client.py
jalauzon-msft: Yeah, I agree the prefix isn't ideal but I think its best to leave it to be consistent with the rest of the APIs here. `set_http_headers()`, `rename_file()`, etc. already prefix all the SMB properties with file.

This is also consistent with the service documentation / header names since they are the same for files and directories.
https://docs.microsoft.com/en-us/rest/api/storageservices/create-directory#request-headers

sdk/storage/azure-storage-file-share/azure/storage/fileshare/_directory_client.py
jalauzon-msft: Same reply as above :)

tasherif-msft (approved):

PR #23485, opened 2022-03-11
Fix remote rendering test pipeline

MichaelZp0:
# Description

The test pipeline for the remoterendering sdk is missing an environment variable, thus all tests run fail in the setup process.
This PR adds this environment variable and should fix the tests or at least will run them, so we can see if something is really broken and not only the environment.

### DIFF
--- a/sdk/remoterendering/azure-mixedreality-remoterendering/CHANGELOG.md
+++ b/sdk/remoterendering/azure-mixedreality-remoterendering/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Release History
 
+## 1.0.0b2 (2022-11-03)
+
 ## 1.0.0b1 (2021-11-15)
 
 - Initial release.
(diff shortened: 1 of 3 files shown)


ChristopherManthei (approved):

sdk/remoterendering/azure-mixedreality-remoterendering/CHANGELOG.md
```diff
@@ -1,14 +1,8 @@
...
+## 1.0.0b2 (2022-11-03)
 
-### Features Added
-
-### Breaking Changes
-
-### Bugs Fixed
-
-### Other Changes
+- Fixed bug in testing pipeline.
```
tjprescott, line 5: This isn't a customer facing change (or even an internal one) and thus nothing should be added to the changelog.

tjprescott (changes requested):

sdk/remoterendering/azure-mixedreality-remoterendering/CHANGELOG.md
MichaelZp0: Okay. I just removed the entry then.

tjprescott (approved):

PR #23504, opened 2022-03-14
[webpubsub] switch to customizations code

iscai-msft:
instead of rewriting the client and configuration, we're going to use the new customizations framework


catalinaperalta (approved):
Nice customizations!

PR #23509, opened 2022-03-15
[EventHub] add retry to send test for connection error

swathipil:
fix for #22961

### DIFF
--- a/sdk/eventhub/azure-eventhub/tests/livetest/asynctests/test_send_async.py
+++ b/sdk/eventhub/azure-eventhub/tests/livetest/asynctests/test_send_async.py
@@ -10,6 +10,7 @@ import asyncio
 import pytest
 import time
 import json
+import uamqp
 
 from azure.eventhub import EventData, TransportType, EventDataBatch
 from azure.eventhub.aio import EventHubProducerClient, EventHubConsumerClient
@@ -139,7 +140,7 @@ async def test_send_amqp_annotated_message(connstr_receivers):
 
 @pytest.mark.liveTest
 @pytest.mark.asyncio
-async def test_send_with_partition_key_async(connstr_receivers):
+async def test_send_with_partition_key_async(connstr_receivers, live_eventhub):
     connection_str, receivers = connstr_receivers
     client = EventHubProducerClient.from_connection_string(connection_str)
     async with client:
@@ -155,15 +156,38 @@ async def test_send_with_partition_key_async(connstr_receivers):
         await client.send_batch(await client.create_batch())
 
     found_partition_keys = {}
+    reconnect_receivers = []
     for index, partition in enumerate(receivers):
-        received = partition.receive_message_batch(timeout=5000)
-        for message in received:
+        retry_total = 0
+        while retry_total < 3:
+            timeout = 5000 + retry_total * 1000
             try:
-                event_data = EventData._from_message(message)
-                existing = found_partition_keys[event_data.partition_key]
-                assert existing == index
-            except KeyError:
-                found_partition_keys[event_data.partition_key] = index
+                received = partition.receive_message_batch(timeout=timeout)
+                for message in received:
+                    try:
+                        event_data = EventData._from_message(message)
+                        existing = found_partition_keys[event_data.partition_key]
+                        assert existing == index
+                    except KeyError:
+                        found_partition_keys[event_data.partition_key] = index
+                if received:
+                    break
+                retry_total += 1
+            except uamqp.errors.ConnectionClose:
+                uri = "sb://{}/{}".format(live_eventhub['hostname'], live_eventhub['event_hub'])
+                sas_auth = uamqp.authentication.SASTokenAuth.from_shared_access_key(
+                    uri, live_eventhub['key_name'], live_eventhub['access_key'])
+
+                source = "amqps://{}/{}/ConsumerGroups/{}/Partitions/{}".format(
+                    live_eventhub['hostname'],
+                    live_eventhub['event_hub'],
+                    live_eventhub['consumer_group'],
+                    index)
+                partition = uamqp.ReceiveClient(source, auth=sas_auth, debug=True, timeout=0, prefetch=500)
+                reconnect_receivers.append(partition)
+                retry_total += 1
+    for r in reconnect_receivers:
+        r.close()
 
 
 @pytest.mark.parametrize("payload", [b"", b"A single event"])
--- a/sdk/eventhub/azure-eventhub/tests/livetest/synctests/test_send.py
+++ b/sdk/eventhub/azure-eventhub/tests/livetest/synctests/test_send.py
@@ -11,6 +11,7 @@ import time
 import json
 import sys
 
+import uamqp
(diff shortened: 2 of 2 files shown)


sdk/eventhub/azure-eventhub/tests/livetest/asynctests/test_send_async.py
```diff
@@ -155,15 +156,38 @@ async def test_send_with_partition_key_async(connstr_receivers):
         await client.send_batch(await client.create_batch())
 
     found_partition_keys = {}
+    reconnect_receivers = []
     for index, partition in enumerate(receivers):
-        received = partition.receive_message_batch(timeout=5000)
-        for message in received:
+        retry_total = 0
+        while retry_total < 3:
```
swathipil, line 162: @yunhaoling - when using the consumer client, the retry is built in. However, since we're using the `receive_message_batch` method in `uamqp` (which is not actually used anywhere in the EH/SB code) and initializing our own `ReceiveClient`, we need to implement this retry logic. Similar logic exists in the `test_reconnect.py` file.

sdk/eventhub/azure-eventhub/tests/livetest/asynctests/test_send_async.py
```diff
@@ -155,15 +156,38 @@ async def test_send_with_partition_key_async(connstr_receivers):
...
+            timeout = 5000 + retry_total * 1000
```
swathipil, line 163: @yunhaoling - Roundtrips were around 3 minutes from East US to Australia (which should be longest distance/roundtrip from point to point) when this test was run locally. Might be longer in the pipeline, so setting to five minutes to start with, then bumping up.

yunhaoling: I'm not sure whether `5 minutes total round trip` is related to the code here: `timeout = 5000 + retry_total * 1000`,

but I just want to call it out that `receive_message_batch` takes timeout in the unit of milliseconds.

`setting 5 minutes` should be 60 * 5 * 1000 = 300000 then?

sdk/eventhub/azure-eventhub/tests/livetest/asynctests/test_send_async.py
```diff
@@ -155,15 +156,38 @@ async def test_send_with_partition_key_async(connstr_receivers):
...
+                    try:
+                        event_data = EventData._from_message(message)
+                        existing = found_partition_keys[event_data.partition_key]
+                        assert existing == index
+                    except KeyError:
+                        found_partition_keys[event_data.partition_key] = index
+                if received:
+                    break
+                retry_total += 1
+            except uamqp.errors.ConnectionClose:
```
yunhaoling, line 176: is partition receiver (ReceiveClient) defined in pytest fixture shared by multiple test cases? or for each test case it would be a new partition receiver obj

sdk/eventhub/azure-eventhub/tests/livetest/asynctests/test_send_async.py
swathipil: @mikeharder - as per @yunhaoling, the biggest reason is that the ConsumerClient uses push-based receiving, which means that it keeps receiving forever unless the user triggers a close. The ReceiveClient is pull-based receiving, so we can just receive for a length of time or a certain number of events. This makes it easy to test the ProducerClient. Switching over to using the ConsumerClient in the tests would take some work.

sdk/eventhub/azure-eventhub/tests/livetest/asynctests/test_send_async.py
swathipil: that makes more sense :D

sdk/eventhub/azure-eventhub/tests/livetest/asynctests/test_send_async.py
swathipil: It's shared by multiple test cases by default, unless we were to set the `fixture(scope="session")` or some other scope that isn't every test.

sdk/eventhub/azure-eventhub/tests/livetest/asynctests/test_send_async.py
swathipil: This was the timeout that worked locally (6/7/8 secs not 6/7/8 mins :P). So setting timeout to less than `5000 + retry_total * 1000` resulted in the UnknownError.

sdk/eventhub/azure-eventhub/tests/livetest/asynctests/test_send_async.py
```diff
@@ -155,15 +156,38 @@ async def test_send_with_partition_key_async(connstr_receivers):
...
+                    uri, live_eventhub['key_name'], live_eventhub['access_key'])
+
+                source = "amqps://{}/{}/ConsumerGroups/{}/Partitions/{}".format(
+                    live_eventhub['hostname'],
+                    live_eventhub['event_hub'],
+                    live_eventhub['consumer_group'],
+                    index)
+                partition = uamqp.ReceiveClient(source, auth=sas_auth, debug=True, timeout=0, prefetch=500)
+                reconnect_receivers.append(partition)
+                retry_total += 1
```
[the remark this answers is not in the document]
swathipil: added a raise error if retries exhausted.

sdk/eventhub/azure-eventhub/tests/livetest/asynctests/test_send_async.py
```diff
@@ -155,15 +156,38 @@ async def test_send_with_partition_key_async(connstr_receivers):
...
+                sas_auth = uamqp.authentication.SASTokenAuth.from_shared_access_key(
+                    uri, live_eventhub['key_name'], live_eventhub['access_key'])
+
+                source = "amqps://{}/{}/ConsumerGroups/{}/Partitions/{}".format(
+                    live_eventhub['hostname'],
+                    live_eventhub['event_hub'],
+                    live_eventhub['consumer_group'],
+                    index)
+                partition = uamqp.ReceiveClient(source, auth=sas_auth, debug=True, timeout=0, prefetch=500)
+                reconnect_receivers.append(partition)
```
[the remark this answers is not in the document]
swathipil: since the ReceiveClients created in the fixture will also be closed in the fixture, I will just close the ReceiveClients that were created in this test here. Will make sure no more than one ReceiveClient is created and open at a time.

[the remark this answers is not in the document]
yunhaoling: sounds good to me

sdk/eventhub/azure-eventhub/tests/livetest/asynctests/test_send_async.py
yunhaoling: yeah, an estimation of 3 minutes in total for connection to all partitions seems reasonable..
okay, now I know what this time unit it is and I'm good with the change.

yunhaoling (approved):

PR #23518, opened 2022-03-15
Add Azure Account extension version requirement to Identity README

scottaddie:
Addresses https://github.com/Azure/azure-sdk-for-net/issues/27263


xiangyan99 (approved):

joshfree (approved):
We should link to the tracking issue for the fix; so that customers who find this text can then easily find the open github issue which tracks the fix for the problem they are experiencing.

PR #23525, opened 2022-03-15
[textanalytics] adjustments to tests for new sentiment analysis model

kristapratico:
tracking possible regressions here: https://msazure.visualstudio.com/Cognitive%20Services/_workitems/edit/13848227


iscai-msft (approved):
rip

PR #23555, opened 2022-03-17
[Release helper] fix bug

BigCat20196:
fix bug when assignee is None


scripts/release_helper/common.py
```diff
@@ -46,7 +46,7 @@ def __init__(self, issue_package: IssuePackage, request_repo_dict: Dict[str, Rep
                  assignee_candidates: Set[str], language_owner: Set[str]):
         self.issue_package = issue_package
         self.request_repo_dict = request_repo_dict
-        self.assignee = issue_package.issue.assignee.login
+        self.assignee = assignee.login if (assignee := issue_package.issue.assignee) else ''
```
msyyc, line 49: We had better not use some grammar which is only for higher python version since it may bring problem when we debug in local env

msyyc (approved):

PR #23570, opened 2022-03-17
[Storage] Fix generated code for service operations

jalauzon-msft:
The previous update to the generated code from the new version of autorest in #23497 broke any Service Operations where the URL path was just `/` in the Swagger. This was due to a bug in a newly introduced directive. This was causing Live Test auth failures for any tests involving a Service Operation.

This change fixes the directive and re-generates code for all packages.


iscai-msft (approved):
thanks @jalauzon-msft !

vincenttran-msft (approved):

jalauzon-msft:
Failing live tests due to known issue with `azure-core` will override and merge

jalauzon-msft:
Failing live tests due to known issue with `azure-core` will override and merge

jalauzon-msft:
Live test failure was known flaky test. Will override.

PR #23578, opened 2022-03-18
[core] loosen check on stream input

iscai-msft:
Right now we're more stringent about `content` input, we expect it to be str / bytes or an iterable / asynciterable. This impacts storage live tests and is also incorrect because we should allow inputs with just a `read()` proeprty. I'm switching my code to be more like the pipeline transport code [here](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/core/azure-core/azure/core/pipeline/transport/_base.py#L243) and duck-typing instead.

Also at the same time, saw a gap to get rid of some `py2/py3` code, so did some code unification as well

Running against live storage: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=1443919&view=results. Will not see all tests passing, @jalauzon-msft is fixing the other failed tests in main right now, but should see the failing stream tests pass


lmazuel (approved):

xiangyan99 (approved):
Do we still need both rest and rest_py3?

iscai-msft:
@xiangyan99 no, I want to refactor that later in a separate PR

Issue #23351, opened 2022-03-04
type checking work across repo


kristapratico:
[Mypy](https://mypy.readthedocs.io/en/stable/introduction.html) and [pyright](https://github.com/microsoft/pyright) are static type checkers that can be run on code to evaluate type annotations and find bugs. The type annotations we write in our code ship as part of our client libraries and our developer customers rely on our typing as a further check of the code that they're writing. Therefore, it is important that we not only write type annotations into our client libraries, but also that these typing annotations are accurate and pass by the type checkers. We are aware of the customer impact of having incorrect type annotations shipped in our packages: https://github.com/Azure/azure-sdk-for-python/issues/20157 https://github.com/Azure/azure-sdk-for-python/issues/21223 https://github.com/Azure/azure-sdk-for-python/issues/20083 https://github.com/Azure/azure-sdk-for-python/issues/21847 https://github.com/Azure/azure-sdk-for-python/issues/23362

In our repo, we currently run mypy as part of CI on all libraries that opt-in on this [list](https://github.com/Azure/azure-sdk-for-python/blob/main/eng/tox/mypy_hard_failure_packages.py). This means that any new library must intentionally add themselves to the allow list to enable type checking on their library. We would like to change this workflow such that mypy and pyright is always run on every library _unless_ that library is placed on a blocklist. This would match the workflow we have with our other static analysis checker, pylint.

To make this a reality, there are a few steps we must take:

1) Decide the list of libraries we _don't_ want to run mypy/pyright on and move them to a blocklist.
2) Work towards getting our entire repo "type checker clean" (aka fix all the typing errors present in our code). Running mypy on all of the data-plane libraries in our repo results in 44 libraries that currently raise mypy errors:

      ```
      azure-agrifood-farming
      azure-ai-anomalydetector
      azure-ai-language-conversations
      azure-ai-language-questionanswering
      azure-applicationinsights
      azure-batch
      azure-common
      azure-communication-chat*
      azure-communication-identity*
      azure-communication-networktraversal*
      azure-communication-phonenumbers*
      azure-communication-sms*
      azure-confidentialledger
      azure-containerregistry*
      azure-core-tracing-opencensus
      azure-core-tracing-opentelemetry
      azure-cosmos
      azure-digitaltwins-core*
      azure-graphrbac
      azure-iot-deviceupdate
      azure-iot-modelsrepository
      azure-loganalytics
      azure-media-analytics-edge
      azure-media-videoanalyzer-edge
[...]

PR #23594, opened 2022-03-21
Mypy/typing dev doc

kristapratico:
Part of https://github.com/Azure/azure-sdk-for-python/issues/23351

### DIFF
--- a/doc/dev/static_type_checking.md
+++ b/doc/dev/static_type_checking.md
@@ -34,7 +34,7 @@ This guide walks through the setup necessary to run mypy, a static type checker,
 
 Python is a dynamically typed (or sometimes called "duck typed") language. This means that the interpreter only checks types as code runs and types are allowed to change.
 This is different from more statically typed languages, like Java, where you will have you declare types in code and check them at compile time.
-However, with PEP 484, type hints were introduced to Python which makes it possible to do the static type checking of Python code.
+However, with [PEP 484](https://peps.python.org/pep-0484/), type hints were introduced to Python which makes it possible to do the static type checking of Python code.
 Type hints can be written into Python code to indicate the type of a variable, argument, return type, etc. There are tools available to perform static type checking using type hints, like mypy. Note that type hints have no effect on the code at runtime and are not enforced by the interpreter.
 
 There are some key benefits to using and checking type hints in Python code:
@@ -99,8 +99,8 @@ So, in practice, what should you aim to type?
 1) Add type hints to publicly exposed APIs in the client library. Type hints get shipped with our client libraries and provide benefit to our users.
    To ensure that type hints do get shipped in your library and can be checked by a static type checker, follow the steps per [PEP 561](https://mypy.readthedocs.io/en/stable/installed_packages.html#creating-pep-561-compatible-packages) below:
 
-       a) add an empty `py.typed` file to your package directory. E.g. `.../sdk/azure-core/azure/core/py.typed`
-       b) include the `py.typed` file in the MANIFEST.in ([example](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/core/azure-core/MANIFEST.in))
+    - add an empty `py.typed` file to your package directory. E.g. `.../sdk/azure-core/azure/core/py.typed`
+    - include the `py.typed` file in the MANIFEST.in ([example](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/core/azure-core/MANIFEST.in))
    
 2) Add type hints anywhere in the source code where unit tests are worth writing. Consider typing/mypy as "free" tests for your library so focusing typing on
 high density/important areas of the code helps in detecting bugs.
@@ -113,11 +113,11 @@ Almost anything can be used as a type in annotations.
 2) Classes or abstract base classes - whether user-defined, from the standard library like [collections]() or [collections.abc](), or external packages
 3) Types from the [typing]() or [typing_extensions]() modules
 4) Built-in generic types, like `list` or `dict`*.
-   > *Note: Support for generic types was not added until Python 3.9 as well (or 3.7+ with `from __future__ import annotations` import). For example, this lets you pass in generic `list[str]` as a type hint rather than `typing.List[str]`. 
+   > *Note: Support for generic types was not added until Python 3.9 (or 3.7+ with `from __future__ import annotations` import). This lets you pass in generic `list[str]` as a type hint rather than `typing.List[str]`. 
 
 Here are a few considerations and notes on Python version support:
 
-- With PEP585 and PEP563, importing certain generic collection types from `typing` has been [deprecated](https://peps.python.org/pep-0585/#implementation) in favor for using the types in `collections.abc` or generic type hints (like `list`) from the standard library.
+- With [PEP585](https://peps.python.org/pep-0585/) and [PEP563](https://peps.python.org/pep-0563/), importing certain generic collection types from `typing` has been [deprecated](https://peps.python.org/pep-0585/#implementation) in favor for using the types in `collections.abc` or generic type hints (like `list`) from the standard library.
 At the time of writing, the Python SDK supports 3.6+ and so we must still import these types from `typing` or check the Python version
 at runtime before importing (Note that the `collections.abc` types did not have [] notation added until 3.9)
 
@@ -137,14 +137,14 @@ from typing_extensions import Protocol
 ```
 
 Our azure-core library takes a dependency on `typing-extensions` so for most client libraries the dependency is already implicit.
-See the [typing-extensions]() docs to check what has been backported.
+See the [typing-extensions](https://github.com/python/typing/tree/master/typing_extensions) docs to check what has been backported.
 
 ## Install and run mypy on your client library code
 
 Our Python SDK repo has the version of mypy that we run in CI pinned to a specific version. This is to avoid surprise typing errors raised when a new version of `mypy` ships.
 All client libraries in the Python SDK repo are automatically opted in to running mypy (TODO not yet).
 
-The easiest way to install and run mypy locally is with tox. This reproduces the exact mypy environment run in CI and brings in the third party stub packages necessary.
+The easiest way to install and run mypy locally is with [tox](https://github.com/Azure/azure-sdk-for-python/blob/main/doc/dev/tests.md#tox). This reproduces the exact mypy environment run in CI and brings in the third party stub packages necessary.
 
 `pip install tox tox-monorepo`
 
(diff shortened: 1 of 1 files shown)


doc/dev/static_type_checking.md
```diff
@@ -0,0 +1,894 @@
...
+not enforced by the interpreter.
+
+There are some key benefits to using and checking type hints in Python code:
+
+1) type hints checked by a static type checker can help developers find bugs in Python code (type hints can be thought
+   of as "free" unit tests, although they do not replace a thorough test suite).
+2) type hints are shipped in our packages so that customers can consume them to further type check their own
+   applications and get improved IDE and Intellisense experiences
+
+There are two styles of type hints: comment and annotation (preferred).
```
lmazuel, line 52: Don't do comment anymore. It's not prefered, it's comments are dead with Python 2

doc/dev/static_type_checking.md
```diff
@@ -0,0 +1,894 @@
...
+
+Type comments are Python 2 compatible and consist of comments in the code which indicate the types of arguments and
+return values. Note that these are truly just comments in the code which the type checker can use.
+
+```python
+from typing import Any, Optional
+
+
+def download_blob_from_url(
```
lmazuel, line 61: Skip everything about how to do comments. Given we have some in place, you may write something like "You may see comments because of legacy code, but really don't, and we will remove them all eventually"

doc/dev/static_type_checking.md
```diff
@@ -0,0 +1,894 @@
...
+Note that since Python 3.6, it is also possible to add type annotations to variables. The syntax follows the same as
+function arguments:
+
+```python
+length: int = 42
+width: int = 8
+area: int = length * width
+```
+
+When typing your library, note that almost anything can be used as a type - built-ins like `str`, `int`, `bool`, types
```
lmazuel, line 100: I remember `list` and `dist` were not usable at first, I don't remember on which Python version they were added, and if `from __future__ import annotations` make them work on 3.6

lmazuel: Ok, you replied later, it's from 3.7, so we can't recommend it _yet_

doc/dev/static_type_checking.md
```diff
@@ -0,0 +1,894 @@
...
+```python
+import sys
+
+if sys.version_info < (3, 9):
+    from typing import Sequence
+else:
+    from collections.abc import Sequence
+```
+
+- The `typing-extensions` library backports types that are introduced only in later versions of Python (e.g. `Protocol`
```
lmazuel, line 156: We should have a note somewhere that typing-extensions is doing a try/except, and therefore people should always import from type-extensions (and not try/except from typing/type-extensions)

doc/dev/static_type_checking.md
```diff
@@ -0,0 +1,894 @@
...
+```
+
+Our azure-core library takes a dependency on `typing-extensions` so for most client libraries the dependency is already
+implicit. See the [typing-extensions](https://github.com/python/typing/tree/master/typing_extensions) docs to check what
+has been backported.
+
+## Install and run mypy on your client library code
+
+Our Python SDK repo has the version of mypy that we run in CI pinned to a specific version. This is to avoid surprise
+typing errors raised when a new version of `mypy` ships. All client libraries in the Python SDK repo are automatically
```
lmazuel, line 170: What version should I isntall at the minimal? Should it be written?

doc/dev/static_type_checking.md
```diff
@@ -0,0 +1,894 @@
...
+
+```python
+from typing import Optional
+
+
+def tree(kind: Optional[str] = None):
+    ...
+```
+
+But not necessary here since a default argument of "oak" is provided:
```
lmazuel, line 303: Well, technically if `kind` can be `None`, then it's `Optional`. It never says that the default value _must_ be `None`

```python
def tree(kind: Optional[str]):  # I can be None, I have no default value, I'm required to be specified
    ...

def tree(kind: Optional[str] = "oak"):  # By default use "oak", but setting `None` is acceptable
    ...

```

Yes, this is terrible naming, and for those previous scenarios I think using `Union[X, None]` is more clear actually.

doc/dev/static_type_checking.md
```diff
@@ -0,0 +1,894 @@
...
+
+The `typing` module includes a pre-defined `TypeVar` named `AnyStr`. Its definition looks like this:
+
+```python
+from typing import TypeVar
+
+AnyStr = TypeVar('AnyStr', bytes, str)
+```
+
+`AnyStr` can be used in functions which accept or return both bytes and str.
```
lmazuel, line 530: There is another value to `AnyStr`, it keeps consistent typing with return type

image

doc/dev/static_type_checking.md
```diff
@@ -0,0 +1,894 @@
...
+
+def receive_deferred_messages(
+        self, sequence_numbers: Union[int, List[int]], *, timeout: Optional[float] = None, **kwargs: Any
+) -> List[ServiceBusReceivedMessage]:
+    if isinstance(sequence_numbers, int):
+        sequence_numbers = [sequence_numbers]
+    sequence_numbers = cast(List[int], sequence_numbers)  # mypy clarity
+```
+
+If you are using cast often, however, it might be an indication that the code should be otherwise written or refactored.
```
lmazuel, line 554: Let's add a note that it's cost-free, it doesn't slow down the implementation at runtime

doc/dev/static_type_checking.md
```diff
@@ -0,0 +1,894 @@
...
+
+def foo(b: "Baz") -> None:
+    b.hello()
+
+
+def bar() -> None:
+    print("bar")
+```
+
+Note that we also had to change Baz to a forward reference since it will not be defined at runtime under
```
lmazuel, line 660: Also use it to hide types that are costly to load at runtime and useful only for typing

Also, typings shoult not be under this, since types from typings are free to import

doc/dev/static_type_checking.md
```diff
@@ -0,0 +1,894 @@
...
+
+A key difference between Protocols and abstract classes (abc) is that protocols work at typing time while abcs work at
+runtime. A class does not need to _inherit_ from Protocol like an abc, the type checker will simply enforce its usage at
+typing time.
+
+A common Python SDK Protocol type we use
+is [azure.core.credentials.TokenCredential](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/core/azure-core/azure/core/credentials.py)
+. TokenCredential requires that a type implements the `get_token` method for it to be consistent with the Protocol.
+
+See more on Protocols in the [reference documentation](https://docs.python.org/3/library/typing.html#typing.Protocol).
```
lmazuel, line 766: Was runtime_checkable backported? If yes we should mention it:

https://docs.python.org/3/library/typing.html#typing.runtime_checkable

doc/dev/static_type_checking.md
kristapratico: Included it. I don't think we'll bump the version often enough that it will make it hard to keep this doc up-to-date.

kristapratico: Included it. I don't think we'll bump the version often enough that it will make it hard to keep this doc up-to-date.

doc/dev/static_type_checking.md
kristapratico: Good point, added this to the doc!

kristapratico: Good point, added this to the doc!

doc/dev/static_type_checking.md
```diff
@@ -0,0 +1,919 @@
...
+def analyze_text(text: str, analysis_kind: EntityRecognition) -> EntityRecognitionResult:
+    ...
+
+
+@overload
+def analyze_text(text: str, analysis_kind: SentimentAnalysis) -> SentimentResult:
+    ...
+
+
+def analyze_text(text, analysis_kind):  # the actual implementation doesn't need to be annotated
```
swathipil, line 831: Here you say "doesn't **need to** be". Wondering if this should be re-worded as **"shouldn't be"**? When I was testing this, I saw that actually annotating the implementation results in errors that would otherwise not be there.

For example, this does not result in mypy errors.
```python
@overload
def get_zero(a: Dict) -> str:
    ...
@overload
def get_zero(a: List) -> str:
    ...
def get_zero(a):
    try:
        zero = a[0]
    except AttributeError:
        zero = a.pop("0", "val")
    return zero
```

But annotating the implementation:
```python
def get_zero(a: Union[Dict, List]) -> str:
    try:
        zero = a[0]
    except AttributeError:
        zero = a.pop("0", "val")
    return zero
```
Results in:
```
encode_and_decode_event_data_message.py:102: error: Too many arguments for "pop" of "list"
encode_and_decode_event_data_message.py:102: error: Argument 1 to "pop" of "list" has incompatible type "str"; expected "SupportsIndex
```

doc/dev/static_type_checking.md
```diff
@@ -0,0 +1,984 @@
...
+    - [Use typing.Protocol for structural subtyping](TODO)
+- [How to ignore a mypy typing error](TODO)
+- [How to opt out of mypy type checking](TODO)
+
+## Intro to typing in Python
+
+Python is a dynamically typed (or sometimes called "duck typed") language. This means that the interpreter only checks
+types as code runs and types are allowed to change. This is different from more statically typed languages, like Java,
+where you will have to declare types in code and check them at compile time. However,
+with [PEP 484](https://peps.python.org/pep-0484/), type hints were introduced to Python which makes it possible to do
```
johanste, line 39: We can mention PEP 483 as well.

doc/dev/static_type_checking.md
```diff
@@ -0,0 +1,984 @@
...
+```python
+from typing import Union, List, Optional, Any, cast
+
+
+def receive_deferred_messages(
+        self, sequence_numbers: Union[int, List[int]], *, timeout: Optional[float] = None, **kwargs: Any
+) -> List[ServiceBusReceivedMessage]:
+    if isinstance(sequence_numbers, int):
+        sequence_numbers = [sequence_numbers]
+    sequence_numbers = cast(List[int], sequence_numbers)  # mypy clarity
```
swathipil, line 783: for clarity that this is being casted b/c list operations won't be valid on an "Int" type, can we include a list operation line after this? something like?
```suggestion
    sequence_numbers = cast(List[int], sequence_numbers)  # mypy clarity
    
    for num in sequence_numbers:    # if not casted as List[int], mypy error
```

doc/dev/static_type_checking.md
```diff
@@ -0,0 +1,959 @@
...
+Note that we also had to change Baz to a forward reference since it will not be defined at runtime under
+the `TYPE_CHECKING` conditional.
+
+Another reason to use `TYPE_CHECKING` is to hide importing types which are needed only for type annotations and are
+costly to load at runtime. Types imported from `typing` or `typing_extensions` do not need to be put under
+a `TYPE_CHECKING` conditional.
+
+### Use TypeVar for generic type hinting
+
+`TypeVar` is a generic type which can be bound to a specific type with each usage. Common usage might be to have the
```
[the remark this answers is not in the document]
kristapratico: Forgot to reply, meant to clarify on this - what did you mean by used for container elements? Did you mean commonly used?

doc/dev/static_type_checking.md
kristapratico: sure!

PR #23604, opened 2022-03-22
[Cosmos] AAD authentication sync client

simorenoh:
This PR has the changes for the sync client to utilize AAD authentication. I'll be making the PR for the async client separate, in order to reduce number of lines per PR.

The way the `@azure.identity` package uses AAD credentials to authenticate services is by adding those credentials into a policy that runs when requests are sent to the core pipelines. This policy makes sure to refresh the current token if needed and set the authentication header of requests going to the pipeline. The reason why Cosmos had to create their own policy in this instance is due to the prefix we utilize for our tokens, since the bearer token policy given by the identity module sends a different prefix altogether and as such does not work for us.

It was also recommended by the identity team to create our own policies entirely rather than attempting to override a couple methods, since this could break us on their end - specially for the `_update_headers()` method since it's private.

Sample is a simple run-through of what can and can't be done, if you think adding more examples would be helpful I can do so as well.


sdk/cosmos/azure-cosmos/azure/cosmos/_cosmos_client_connection.py
```diff
@@ -188,6 +191,10 @@ def __init__(
             HttpLoggingPolicy(**kwargs),
         ]
 
+        if self.aad_credentials:
+            scopes = base.create_scope_from_url(self.url_connection)
```
johanste, line 195: Are scopes constructed the same way in all clouds?

simorenoh: For the most part yes, they take whatever prefix comes before the port and assign `/.default`. Only difference is for Cosmos we utilize the user account endpoint. Here is Storage SDK scopes for comparison (file is constants.py under _shared directory for blob sdk): 
image

johanste: Are there scenario where customers use custom DNS names (either supported by the service - which I don't believe cosmos provides) or by custom proxies?

simorenoh: We generally do not recommend using DNS since it can create issues regardless, and I do believe you'd need the database account endpoint as a scope (which means you'd need this information regardless), but I'll have to ask the Cosmos RBAC team to see how this works.

simorenoh: Checked with the RBAC folks and it seems this isn't supported, so we should be good.

With that said, I do think the way I'm just replacing "443" with ".default" doesn't look so nice since we're searching the string for the 443... I thought of doing `return url.split(":")[0] + ":" + url.split(":")[1] + "/.default"` as well but not sure if that looks worse, any advice on this is appreciated!

sdk/cosmos/azure-cosmos/CHANGELOG.md
```diff
@@ -3,12 +3,7 @@
 ### 4.3.0b4 (Unreleased)
 
 #### Features Added
-
-#### Breaking Changes
-
-#### Bugs Fixed
-
-#### Other Changes
+- Added support for AAD authentication for the sync client
```
annatisch, line 6: Why only sync for now?

sdk/cosmos/azure-cosmos/README.md
```diff
@@ -76,6 +76,28 @@ KEY = os.environ['ACCOUNT_KEY']
...
+from azure.cosmos import CosmosClient
+from azure.identity import ClientSecretCredential
+
+import os
+URL = os.environ['ACCOUNT_URI']
+TENANT_ID = os.environ['TENANT_ID']
+CLIENT_ID = os.environ['CLIENT_ID']
+CLIENT_SECRET = os.environ['CLIENT_SECRET']
+
+AAD_CREDENTIALS = ClientSecretCredential(
```
annatisch, line 92: Our samples usually use the DefaultAzureCredential

sdk/cosmos/azure-cosmos/README.md
```diff
@@ -76,6 +76,28 @@ KEY = os.environ['ACCOUNT_KEY']
...
+    tenant_id=TENANT_ID,
+    client_id=CLIENT_ID,
+    client_secret=CLIENT_SECRET)
+
+client = CosmosClient(URL, AAD_CREDENTIALS)
```
annatisch, line 97: I'd just make these lowercase variables - looks like we're passing in constants.

sdk/cosmos/azure-cosmos/azure/cosmos/_base.py
```diff
@@ -663,6 +663,11 @@ def ParsePaths(paths):
     return tokens
 
 
+def create_scope_from_url(url):
+    # Need a better way to do this, regex?
+    return url.replace(":443", "") + ".default"
```
annatisch, line 668: If we actually do a urlparse we should be able to reconstruct the URL with the extra ".default" path and without the port?

sdk/cosmos/azure-cosmos/azure/cosmos/_cosmos_client_connection.py
annatisch: Just to go back Johan's original question - have we tested this with China and USGov clouds?
Does the AAD credentials work with Azure Stack or emulator - and if not - what kind of errors would I expect?

sdk/cosmos/azure-cosmos/azure/cosmos/_cosmos_client_connection.py
```diff
@@ -188,6 +191,10 @@ def __init__(
...
+            policies.append(CosmosBearerTokenCredentialPolicy(self.aad_credentials, scopes))
```
annatisch, line 196: I'm don't think we usually append the credentials policy to the end of the pipeline - could be worth checking where azure-core typically puts the authentication policy (e.g. before or after the retry policy)

sdk/cosmos/azure-cosmos/azure/cosmos/auth.py
```diff
@@ -138,9 +144,166 @@ def __GetAuthorizationTokenUsingResourceTokens(resource_tokens, path, resource_i
...
             if not segment in resource_types and sub_path in resource_tokens:
                 return resource_tokens[sub_path]
 
     return None
+
+
+# pylint:disable=too-few-public-methods
+class _CosmosBearerTokenCredentialPolicyBase(object):
+    """Base class for a Cosmos Bearer Token Credential Policy.
+    Based on ~azure.core.pipeline.policies.BearerTokenCredentialPolicy.
```
annatisch, line 158: How does this implementation differ from the implementation in Core?

sdk/cosmos/azure-cosmos/samples/config.py
```diff
@@ -5,4 +5,7 @@
     'master_key': os.environ.get('ACCOUNT_KEY', '[YOUR KEY]'),
     'database_id': os.environ.get('COSMOS_DATABASE', '[YOUR DATABASE]'),
     'container_id': os.environ.get('COSMOS_CONTAINER', '[YOUR CONTAINER]'),
+    'tenant_id': os.environ.get('TENANT_ID', '[YOUR CONTAINER]'),
+    'client_id': os.environ.get('COSMOS_CONTAINER', '[YOUR CONTAINER]'),
+    'client_secret': os.environ.get('COSMOS_CONTAINER', '[YOUR CONTAINER]'),
```
annatisch, line 10: I'm thinking the client_id and client_secret aren't the same thing as the "your container"?

sdk/cosmos/azure-cosmos/CHANGELOG.md
simorenoh: I'm splitting it into two PR's to not make it too many lines

sdk/cosmos/azure-cosmos/azure/cosmos/_cosmos_client_connection.py
simorenoh: pics from tables and storage sdks, I moved it to be before the DecodePolicy like in theirs:
image
image

sdk/cosmos/azure-cosmos/azure/cosmos/auth.py
simorenoh: Only in the authorization header, pushing changes now to reflect that.

simorenoh:
@annatisch I have spoken with Thomas Weiss from the security team and he has shared some of these answers with me; that data plane RBAC is available **is** sovereign clouds, but not in Azure Stack since CosmosDB isn't offered there. 

For the emulator, I can forward you an email with the full explanation, but we do not support AAD since the emulator does not have an identity service, and is only meant to be used alongside a primary key (there's no way for you to set up an identity that can authorize against the emulator).

sdk/cosmos/azure-cosmos/README.md
simorenoh: updated it to show user both - but if you think it's better to remove the ClientSecretCredential object as a whole I can do that too

sdk/cosmos/azure-cosmos/azure/cosmos/_cosmos_client_connection.py
simorenoh: @annatisch I asked Thomas Weiss from the security team and he has shared some of these answers with me; that data plane RBAC is available is sovereign clouds, but not in Azure Stack since CosmosDB isn't offered there.

For the emulator, I can forward you an email with the full explanation, but we do not support AAD since the emulator does not have an identity service, and is only meant to be used alongside a primary key (there's no way for you to set up an identity that can authorize against the emulator).

annatisch: Cool okay - so it looks like the only scenario we need to validate is the sovereign clouds :)

sdk/cosmos/azure-cosmos/azure/cosmos/_cosmos_client_connection.py
annatisch: I would start with using azure-core as your reference for things like this, which actually seems to have it _after_ the retry policy:
https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/core/azure-core/azure/core/_pipeline_client.py#L131
You could test it both to see what impact retry might have on resetting the auth header.

sdk/cosmos/azure-cosmos/samples/access_cosmos_with_aad.py
```diff
@@ -0,0 +1,102 @@
+import azure.cosmos.cosmos_client as cosmos_client
```
annatisch, line 1: Could we just import from `azure.cosmos`? The fact that the `cosmos_client` part of the namespace is even public is by accident... and I'd rather we didn't reinforce it in the samples :)

sdk/cosmos/azure-cosmos/azure/cosmos/_auth_policies.py
```diff
@@ -0,0 +1,161 @@
...
+    """Base class for a Bearer Token Credential Policy.
+
+    :param credential: The credential.
+    :type credential: ~azure.core.credentials.TokenCredential
+    :param str scopes: Lets you specify the type of access needed.
+    """
+
+    def __init__(self, credential, *scopes, **kwargs):  # pylint:disable=unused-argument
+        # type: (TokenCredential, *str, **Any) -> None
+        super(_CosmosBearerTokenCredentialPolicyBase, self).__init__()
```
annatisch, line 24: Why are we calling `super` when we're not inheriting from anything?
Though having said that - given that 90% of this is identical to the Core implementation - is there any reason you're not inheriting from the Core policy and simply overriding `_update_headers`?

annatisch: Yeah good point. And these are technically internal if we swap out the implementation later.

sdk/cosmos/azure-cosmos/azure/cosmos/_cosmos_client_connection.py
annatisch: Kay, let's leave it where it is for now - we can do a little homework before GA to figure out why it sits in different places in different SDKs to see if it will make a difference here.

sdk/cosmos/azure-cosmos/azure/cosmos/_cosmos_client_connection.py
simorenoh: I spoke to my team on AAD in the other SDKs and the use of sovereign clouds. Basically, the scope that we utilize for the token will always be this hardcoded convention using the user's database account, and does not depend on the cloud - it is a Cosmos specific scope.

sdk/cosmos/azure-cosmos/azure/cosmos/http_constants.py
```diff
@@ -379,6 +379,7 @@ class SubStatusCodes(object):
     REDUNDANT_COLLECTION_PUT = 1009
     SHARED_THROUGHPUT_DATABASE_QUOTA_EXCEEDED = 1010
     SHARED_THROUGHPUT_OFFER_GROW_NOT_NEEDED = 1011
+    AAD_REQUEST_NOT_AUTHORIZED = 5300
```
xinlian12, line 382: where this is used?

xinlian12 (approved):
LGTM, thanks

sdk/cosmos/azure-cosmos/azure/cosmos/http_constants.py
simorenoh: it's not used but I figured keeping track of it within the SDK would be useful in case we do need it, pretty sure we don't directly use a lot of the other ones either

sdk/cosmos/azure-cosmos/README.md
```diff
@@ -76,6 +76,33 @@ KEY = os.environ['ACCOUNT_KEY']
...
+    tenant_id=tenant_id,
+    client_id=client_id,
+    client_secret=client_secret)
+
+# Using DefaultAzureCredential (recommended)
+aad_credentials = DefaultAzureCredential()
+
+client = CosmosClient(url, aad_credentials)
+```
+More information on allowed operations for AAD authenticated clients: [RBAC Permission Model](https://aka.ms/cosmos-native-rbac)
```
milismsft, line 104: This is the wrong link; it should be https://docs.microsoft.com/en-us/azure/cosmos-db/how-to-setup-rbac
We should also mention explicitly that the AAD token must have permissions to:
Microsoft.DocumentDB/databaseAccounts/readMetadata

simorenoh: This is the link to the permission model showing the allowed operations based on permissions granted to the managed identity, I'll add the link you posted to help people with setup too though, that's good information to put there for sure.

sdk/cosmos/azure-cosmos/azure/cosmos/cosmos_client.py
```diff
@@ -60,10 +60,12 @@ def _build_auth(credential):
...
     elif hasattr(credential, '__iter__'):
         auth['permissionFeed'] = credential
+    elif hasattr(credential, 'get_token'):
+        auth['clientSecretCredential'] = credential
     else:
         raise TypeError(
-            "Unrecognized credential type. Please supply the master key as str, "
-            "or a dictionary or resource tokens, or a list of permissions.")
+            "Unrecognized credential type. Please supply the master key as a string "
+            "or a dictionary, or resource tokens, or a list of permissions, or a ClientSecretCredential.")
```
milismsft, line 68: ClientSecretCredential -> any instance of a class implementing TokenCredential (see azure.identity for specific implementations such as ClientSecretCredential)

milismsft (approved):

sdk/cosmos/azure-cosmos/README.md
milismsft: That previous link is different, the AAD token obtained though just RBAC setting will not work with this implementation.

PR #23626, opened 2022-03-23
add app service api 2017 backward compatibility support

xiangyan99:
If app service 2019 env vars are not found, we treat it as Azure ML credential.

Add this backward compatibility support in case it is running on early version of app service.


chlowell:
We should restore the old test cases as well, to have coverage of parsing the weird expires_on format.

mccoyp (approved):

PR #23629, opened 2022-03-23
[Storage] Added `creation_time` and `expiry_time` to `get_paths()`

jalauzon-msft:
`get_paths()` now returns the additional properties `creation_time` and `expiry_time` for each path. These properties are returned from the service as MS Filetimes which is not natively supported by Python's `datetime` so a custom parser was created to convert MS Filetime to `datetime` objects.


vincenttran-msft (approved):
LGTM! Great work, always fun trying to get everything to play nicely with different time formats 😁

Issue #17301, opened 2021-03-12
[Azure Storage Queue] receive_messages infinite loop


logachev:
- **Package Name**: 
- **Package Version**: 
- **Operating System**:
- **Python Version**: 

**Describe the bug**
It is not a bug, more like really unexpected behavior.

https://github.com/Azure/azure-sdk-for-python/blob/1ca37b47d156ed76922835e7097ee55396bf2fc5/sdk/storage/azure-storage-queue/azure/storage/queue/_queue_client.py#L546

Azure Storage Queue `receive_messages` doesn't have a way to define maximum number of messages to receive. If it is ever used against the queue that constantly gets new messages, it will just stuck in an infinite loop because it will try to query extra messages.

If it runs in a serverless environment, where execution is limited by time, it will lead to potential data loss when function is killed.

There is a workaround for this, but it is not obvious from the method description that it will continuously look for messages instead of dequeuing all available or X.. 

```
messages = queue_service.receive_messages()

try:
    return [m for m in messages.by_page().next()]
except StopIteration:
    return []
```

**To Reproduce**
1. 

**Expected behavior**

**Screenshots**

**Additional context**

yunhaoling:
thanks @logachev for reaching out! we'll investigate it asap.
adding @xiafu-msft who could further comment on the usage of the storage queue APIs.

tasherif-msft:
Hi @logachev, this is a good catch our rest API actually supports passing a max number of messages to retrieve. I will put out a PR to fix up our kwargs so that users can pass a set number of messages.

Thank you.

tasherif-msft:
@logachev I will be putting out a PR to resolve this soon!

logachev:
@tasherif-msft Thank you!

As a "feature request", ideally if it allowed to specify not only # of messages, but also # of pages to have behavior, similar to the original SDK when we do  a single query to retrieve no more than X messages vs multiple queries until queue is empty or we received X messages.

xiafu-msft:
just an update, I will sync up with @tasherif-msft about the pr

tasherif-msft:
I have just created a PR https://github.com/Azure/azure-sdk-for-python/pull/18783 it is still a work in progress.
Just to clarify, you're also asking for a feature where the user would limit the number of pages returned?

amishra-dev:
The pull request is still being worked on @xiafu-msft / @annatisch can we get this Pr merged?

amishra-dev:
[...]

PR #23634, opened 2022-03-24
[Storage] Add max_messages keyword to receive_messages()

vincenttran-msft:
As raised by a customer in #17301, their concern is that there is currently no support for defining a maximum number of messages to receive (and thus could be open to infinite loops under a set of certain circumstances). 

This PR adds `max_messages` keyword support for `recieve_messages()` and also adds sample code showcasing its use-case in `queue_samples_message.py`.


vincenttran-msft:
I completely messed up the git commit history- going to create a seperate branch and rein-corporate changes because I severely messed up using a rebase :)

Issue #23621, opened 2022-03-23
Example text is incorrect (Copy/paste error)


0xUnicorn:
Below `create_table_if_not_exists` it has an `Example` with the text `Deleting a table from the TableServiceClient object` which is also the text used below for `delete_table`.

The fix would be to change the text to: `Creating a table if it doesn't exist, from the TableServiceClient object`

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 4deb2d00-2362-4ee8-661e-fd93821ce4dc
* Version Independent ID: 473126fd-d011-e731-7f7e-f41cd3aed15b
* Content: [azure.data.tables.TableServiceClient class](https://docs.microsoft.com/en-us/python/api/azure-data-tables/azure.data.tables.tableserviceclient?view=azure-python#azure-data-tables-tableserviceclient-create-table-if-not-exists)
* Content Source: [docs-ref-autogen/azure-data-tables/azure.data.tables.TableServiceClient.yml](https://github.com/MicrosoftDocs/azure-docs-sdk-python/blob/master/docs-ref-autogen/azure-data-tables/azure.data.tables.TableServiceClient.yml)
* GitHub Login: @lmazuel
* Microsoft Alias: **lmazuel**

kashifkhan:
Hi @0xUnicorn , thanks for the feedback. We'll take a look at this and get back to you asap.

PR #23641, opened 2022-03-24
[Tables] Address example's caption

YalinLi0312:
For issue: https://github.com/Azure/azure-sdk-for-python/issues/23621


annatisch (approved):

sdk/tables/azure-data-tables/azure/data/tables/_table_service_client.py
```diff
@@ -213,7 +213,7 @@ def create_table_if_not_exists(self, table_name, **kwargs):
                 :end-before: [END create_table_if_not_exists]
                 :language: python
                 :dedent: 8
-                :caption: Deleting a table from the TableServiceClient object
+                :caption: Creating a table if it doesn't exist, from the TableServiceClient object
```
annatisch, line 216: Do we need to make any corrections to the async caption?

YalinLi0312: It's correct there: https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/tables/azure-data-tables/azure/data/tables/aio/_table_service_client_async.py#L220

PR #18783, opened 2021-05-18
[WIP] Added kwarg to limit number of messages received in Queues

tasherif-msft:
Resolves https://github.com/Azure/azure-sdk-for-python/issues/17301

PR #23654, opened 2022-03-24
[Storage] Added max_messages kwarg to receive_messages()

vincenttran-msft:
As raised by a customer in #17301, their concern is that there is currently no support for defining a maximum number of messages to receive (and thus could be open to infinite loops under a set of certain circumstances). 

This PR adds `max_messages` keyword support for `receive_messages()` and also adds sample code showcasing its use-case in `queue_samples_message.py` (and async-equivalents).

Thanks to @tasherif-msft for laying the groundwork in #18783 for this PR!


sdk/storage/azure-storage-queue/CHANGELOG.md
```diff
@@ -3,7 +3,11 @@
 ## 12.2.1 (Unreleased)
 
 ### Features Added
-- Added support for `create_if_not_exists()` for `QueueClient`
```
jalauzon-msft, line 6: Looks like something happened with your merge/diff as your PR is still trying to remove things already in `main`

sdk/storage/azure-storage-queue/azure/storage/queue/_models.py
```diff
@@ -264,8 +264,10 @@ class MessagesPaged(PageIterator):
     :param callable command: Function to retrieve the next page of items.
     :param int results_per_page: The maximum number of messages to retrieve per
         call.
+    :param int messages_to_retrieve: The maximum number of messages to retrieve from
+        the queue.
     """
-    def __init__(self, command, results_per_page=None, continuation_token=None):
+    def __init__(self, command, results_per_page=None, continuation_token=None, messages_to_retrieve=None):
```
jalauzon-msft, line 270: Nit: Can we rename this to `max_messages` to match the parameter.

sdk/storage/azure-storage-queue/azure/storage/queue/_models.py
```diff
@@ -275,9 +277,20 @@ def __init__(self, command, results_per_page=None, continuation_token=None):
         )
         self._command = command
         self.results_per_page = results_per_page
+        self.messages_to_retrieve = messages_to_retrieve
 
     def _get_next_cb(self, continuation_token):
         try:
+            if self.messages_to_retrieve is not None:
```
jalauzon-msft, line 284: What if it is `None`? Would we ever expect it to be `None` since we control it? Probably safe to remove this check especially since we don't do anything if it is. If it somehow miraculously was `None` I'd think we'd want to let the exception happen so we stop the iteration.

sdk/storage/azure-storage-queue/azure/storage/queue/_models.py
```diff
@@ -275,9 +277,20 @@ def __init__(self, command, results_per_page=None, continuation_token=None):
...
+                if self.results_per_page is None:
+                    self.results_per_page = 1
```
jalauzon-msft, line 286: I think maybe we should have this be the default value when popping the `kwarg` rather than checking for it here?

sdk/storage/azure-storage-queue/CHANGELOG.md
vincenttran-msft: Yeah... Just did another fetch/rebase/push/merge and the CHANGELOG.md just won't accept it.. will look into this more tomorrow

sdk/storage/azure-storage-queue/azure/storage/queue/_models.py
vincenttran-msft: Maybe I'm misunderstanding so we can follow-up tomorrow; but I believe the reason why this check is valid is because we want to have the same codepath if the customer does not define a `max_messages` value (which would propagate to `messages_to_retrieve`), then we want `receieve_messages()` to work exactly like it did before the introduction of the keyword `max_messages`.

sdk/storage/azure-storage-queue/azure/storage/queue/_models.py
vincenttran-msft: I am definitely open to suggestions to this- I think my thinking here was that I didn't want to change the default value from `None`, but in testing if we don't provide a value for `results_per_page` but we do provide a value for `mesages_to_retrieve` (thus falling into this conditional), we will face a `TypeError` for using an inequality on `None` (line 289)

sdk/storage/azure-storage-queue/azure/storage/queue/_models.py
vincenttran-msft: Resolved in offline discussion.

sdk/storage/azure-storage-queue/azure/storage/queue/_models.py
vincenttran-msft: Resolved in offline discussion.

sdk/storage/azure-storage-queue/azure/storage/queue/_models.py
```diff
@@ -275,9 +277,20 @@ def __init__(self, command, results_per_page=None, continuation_token=None):
         )
         self._command = command
         self.results_per_page = results_per_page
+        self.max_messages = max_messages
```
annatisch, line 280: Can we make this internal?

sdk/storage/azure-storage-queue/azure/storage/queue/_models.py
```diff
@@ -275,9 +277,20 @@ def __init__(self, command, results_per_page=None, continuation_token=None):
...
+            if self.max_messages is not None:
+                if self.results_per_page is None:
+                    self.results_per_page = 1
+                if self.max_messages == 0:
+                    raise StopIteration("End of paging")
+                if self.results_per_page > self.max_messages:
+                    self.results_per_page = self.max_messages
+                    self.max_messages = 0
+                else:
+                    self.max_messages = self.max_messages - self.results_per_page
```
annatisch, line 293: We should probably only set this value after we have successfully retrieved a page, and deduct the number of items that we actually retrieved rather than the number of items we hope to retrieve.

sdk/storage/azure-storage-queue/azure/storage/queue/_queue_client.py
```diff
@@ -242,28 +242,6 @@ def create_queue(self, **kwargs):
         except HttpResponseError as error:
             process_storage_error(error)
 
-    @distributed_trace
-    def create_if_not_exists(self, **kwargs):
```
annatisch, line 246: Looks like you're missing the latest from main?

sdk/storage/azure-storage-queue/azure/storage/queue/aio/_models.py
```diff
@@ -32,9 +34,20 @@ def __init__(self, command, results_per_page=None, continuation_token=None):
...
         self._command = command
         self.results_per_page = results_per_page
+        self.max_messages = max_messages
 
     async def _get_next_cb(self, continuation_token):
         try:
+            if self.max_messages is not None:
+                if self.results_per_page is None:
+                    self.results_per_page = 1
+                if self.max_messages == 0:
```
annatisch, line 44: <1 might be safer

sdk/storage/azure-storage-queue/azure/storage/queue/aio/_queue_client_async.py
```diff
@@ -486,11 +464,13 @@ def receive_messages(self, **kwargs):
         # type: (Optional[Any]) -> AsyncItemPaged[QueueMessage]
         """Removes one or more messages from the front of the queue.
 
-        When a message is retrieved from the queue, the response includes the message
+         When a message is retrieved from the queue, the response includes the message
```
annatisch, line 467: Extra whitespace?

sdk/storage/azure-storage-queue/tests/test_queue.py
```diff
@@ -405,6 +383,39 @@ def test_get_messages_with_options(self, storage_account_name, storage_account_k
             self.assertNotEqual('', message.expires_on)
             self.assertNotEqual('', message.next_visible_on)
 
+    @QueuePreparer()
+    def test_get_messages_with_max_messages(self, storage_account_name, storage_account_key):
```
annatisch, line 387: Could we please add another test of the reverse scenario?
I.e. load a queue up with 5 messages, and set max_messages=10

annatisch: We should also test other combinations of parameters - for example, if I set `messages_per_page=3`, then I set `max_messages=10`, we should confirm that our logic isn't actually gathering 12 messages and discarding the last 2 (the current implementation shouldn't have this problem, but by testing for it, we can protect against future regressions).

sdk/storage/azure-storage-queue/samples/queue_samples_message_async.py
```diff
@@ -279,6 +279,49 @@ async def update_message_async(self):
...
+        # Instantiate a queue client
+        from azure.storage.queue.aio import QueueClient
+        queue = QueueClient.from_connection_string(self.connection_string, "myqueue7")
+
+        # Create the queue
+        async with queue:
+            await queue.create_queue()
+
+            try:
+            # [START send_messages]
```
annatisch, line 293: I don't think we need the snippet tags here, unless you actually plan to add these into the reference documentation?

sdk/storage/azure-storage-queue/azure/storage/queue/_models.py
```diff
@@ -286,6 +295,8 @@ def _extract_data_cb(self, messages): # pylint: disable=no-self-use
         # There is no concept of continuation token, so raising on my own condition
         if not messages:
             raise StopIteration("End of paging")
+        if self._max_messages is not None:
+            self._max_messages = self._max_messages - self.results_per_page
```
annatisch, line 299: self._max_messages = self._max_messages - len(messages)?

vincenttran-msft: Good catch! I had this in my notes but completely missed it when moving this logic :) thanks Anna!

sdk/storage/azure-storage-queue/azure/storage/queue/_models.py
```diff
@@ -275,9 +277,16 @@ def __init__(self, command, results_per_page=None, continuation_token=None):
         )
         self._command = command
         self.results_per_page = results_per_page
+        self._max_messages = max_messages
```
jalauzon-msft, line 280: Nit: It feels weird to me to be modifying a member named `max_messages`. Perhaps the name of this member should be changed to something like `_message_count` or something that better reflects that this is actually a count of the remaining messages we want to grab, and it just starts at `max_messages`? (It's only max_messages in the beginning).

This is a nit though so feel free to WF :)

sdk/storage/azure-storage-queue/tests/test_queue_async.py
```diff
@@ -145,6 +148,156 @@ async def test_create_queue_with_options(self, storage_account_name, storage_acc
...
+        qsc = QueueServiceClient(self.account_url(storage_account_name, "queue"), storage_account_key)
+        queue_client = self._get_queue_reference(qsc)
+        await queue_client.create_queue()
+        await queue_client.send_message(u'message1')
+        await queue_client.send_message(u'message2')
+        await queue_client.send_message(u'message3')
+        await queue_client.send_message(u'message4')
+        await queue_client.send_message(u'message5')
+        result = []
+        async for m in queue_client.receive_messages(messages_per_page=5, max_messages=2):
```
jalauzon-msft, line 229: Maybe this case shouldn't be allowed and should raise a `ValueError`? I guess it "works" but also is a weird thing for the customer to do. I'll leave it up to you.

If you did check and raise an error, I would suggest doing that in the client method itself not the iterator.

jalauzon-msft (approved):
Overall looks good to me! Great work. I left two comments that you can take or leave :)

sdk/storage/azure-storage-queue/azure/storage/queue/_models.py
vincenttran-msft: For once I will actually take the opposite stance- I think `max_messages` is okay since it starts at the maximum number of messages we want to receive, and then is decreasing in value. Something like `message_count` in my eyes makes sense if we were starting at 0 and keeping track of messages we've received (i.e. increasing in value). So I will leave it for now, although always open to change if we have others chime in that feel strongly one way or the other 😃

sdk/storage/azure-storage-queue/tests/test_queue_async.py
vincenttran-msft: I agree that this functionality should not be expected. I think when I originally wrote this test and it passed, I made a judgement call and said "well, if the customer REALLY wants to do that, be my guest 😝" but yeah I think raising a `ValueError` here would be the right call!

Issue #20474, opened 2021-08-31
Add support to enable PII Logging in Identity SDK.


g2vinay:
Add API support to enable PII Logging in Identity SDK.
if enabled, log the Username that Shared Token Cache Credential uses.

PR to use as reference: https://github.com/Azure/azure-sdk-for-java/pull/23301

PR #23658, opened 2022-03-24
enable pii logging

xiangyan99:


kashifkhan:
There should probably be a CHANGELOG update for new behavior when logging is set to DEBUG ?

xiangyan99:
#20474

kashifkhan (approved):

mccoyp (commented):
Are there tests to either validate some aspects of the logging when we expect it, or more importantly, ensure we don't log this info when we don't expect to?

xiangyan99:
> Are there tests to either validate some aspects of the logging when we expect it, or more importantly, ensure we don't log this info when we don't expect to?

Is it by design that if user sets logging level to debug, PII is always logged. I will add tests to make sure it is not logged by default.

PR #23660, opened 2022-03-25
[Storage] Implement new paginated API `list_page_ranges()`, replacing `get_page_ranges()`

jalauzon-msft:
STG82 added optional pagination to the `get_page_ranges()` API. We took this opportunity to re-implement this API with two new methods, `list_page_ranges()` and `list_page_ranges_diff()`. These APIs are both paginated and returned auto-paging iterators. We also chose to split out the `diff` API into its own client method as it is in other languages. This protects against future changes that only apply to a single API.

The return type from the of these new methods differs from the existing `get_page_ranges()` as it combines the filled and cleared pages into a single list of `PageRage` objects which had a `cleared` property added to indicate whether the page was cleared or not.

Additionally, `get_page_ranges()` is marked as deprecated as part of this change.

NOTE: This change also includes some generated code for another STG82 feature (copy source tags on copy) which will be implemented in a future PR.


vincenttran-msft (approved):
LGTM! Great work :)

sdk/storage/azure-storage-blob/azure/storage/blob/_blob_client.py
```diff
@@ -2967,6 +2974,92 @@ def get_page_ranges( # type: ignore
...
 
+    @distributed_trace
+    def list_page_ranges(
+            self,
+            *,
+            offset: Optional[int] = None,
+            length: Optional[int] = None,
+            previous_snapshot: Optional[Union[str, Dict[str, Any]]] = None,
+            **kwargs: Any
+        ) -> ItemPaged[PageRange]:
```
annatisch, line 2985: Thank you for a nice Python3 style type hinted method signature :)

sdk/storage/azure-storage-blob/azure/storage/blob/_blob_client.py
```diff
@@ -2967,6 +2974,92 @@ def get_page_ranges( # type: ignore
             process_storage_error(error)
         return get_page_ranges_result(ranges)
 
+    @distributed_trace
+    def list_page_ranges(
+            self,
+            *,
+            offset: Optional[int] = None,
+            length: Optional[int] = None,
+            previous_snapshot: Optional[Union[str, Dict[str, Any]]] = None,
```
annatisch, line 2983: What's the `Dict[str, Any]` supported here?

sdk/storage/azure-storage-blob/azure/storage/blob/_models.py
```diff
@@ -836,11 +836,81 @@ class PageRange(DictMixin):
         Start of page range in bytes.
     :param int end:
         End of page range in bytes.
+    :ivar bool cleared:
+        Whether the range has been cleared.
     """
 
-    def __init__(self, start=None, end=None):
+    def __init__(self, start=None, end=None, *, cleared=False):
```
annatisch, line 843: Was this object previously used for any APIs apart from `get_page_ranges`?
And if I use this with the previous deprecated API, will I get `cleared=False` on my output when that may not be true?

sdk/storage/azure-storage-blob/azure/storage/blob/_models.py
```diff
@@ -836,11 +836,81 @@ class PageRange(DictMixin):
...
+    def _parse_page_list(page_list):
+        """Parse a generated PageList into a single list of PageRange sorted by start.
+        """
+        page_ranges = page_list.page_range
+        clear_ranges = page_list.clear_range
+
+        ranges = []
+        p_i, c_i = 0, 0
+
+        # Combine page ranges and clear ranges into single list, sorted by start
```
annatisch, line 858: Is this the logic that the .NET lib used to sort?
Does this hold true with the next page - i.e. the pages will be ordered according to start?

sdk/storage/azure-storage-blob/azure/storage/blob/_models.py
```diff
@@ -836,11 +836,81 @@ class PageRange(DictMixin):
...
     """
 
-    def __init__(self, start=None, end=None):
+    def __init__(self, start=None, end=None, *, cleared=False):
         self.start = start
         self.end = end
+        self.cleared = cleared
+
+    @staticmethod
+    def _parse_page_list(page_list):
```
annatisch, line 849: This should not be a staticmethod - let's move it into a utility function that can be shared between the sync and async iterators.

sdk/storage/azure-storage-blob/azure/storage/blob/aio/_models.py
```diff
@@ -141,3 +141,39 @@ def _build_item(item):
             blob = FilteredBlob(name=item.name, container_name=item.container_name, tags=tags)
             return blob
         return item
+
+
+class PageRangePaged(AsyncPageIterator):
```
annatisch, line 146: I think these days we might not have to actually create this entire PageIterator classes.... let me see if I can find an example of a more simple implementation.

sdk/storage/azure-storage-blob/azure/storage/blob/_blob_client.py
jalauzon-msft: Happy to start to be able to use this style :)

sdk/storage/azure-storage-blob/azure/storage/blob/_blob_client.py
jalauzon-msft: I believe this is to handle the output of `create_snapshot()` directly. That API outputs a `Dict[str, Any]` with some of the response headers with one of the being `"snapshot"`. We have this to allow that output to be passed directly and we parse this in `_get_page_ranges_options()`. A bit weird but I left it to be consistent with the existing API.

```python
if previous_snapshot_diff:
          try:
              options['prevsnapshot'] = previous_snapshot_diff.snapshot # type: ignore
          except AttributeError:
              try:
                  options['prevsnapshot'] = previous_snapshot_diff['snapshot'] # type: ignore
              except TypeError:
                  options['prevsnapshot'] = previous_snapshot_diff
```

sdk/storage/azure-storage-blob/azure/storage/blob/_models.py
jalauzon-msft: I couldn't actually find anywhere this existing class was used besides being exported. It's not used in the output of the existing `get_page_ranges()` as that just returns a raw Tuple. I also don't see it in any tests, samples, or as the input to anything. Searching the code, its only defined and exported.

So I decided it was safe to just modify and use it.

sdk/storage/azure-storage-blob/azure/storage/blob/_models.py
jalauzon-msft: Yes, I based this off of .NET. Looking at the service documentation, the list is guaranteed to come back ordered by range, so I assume the page and clear lists are already sorted.

https://docs.microsoft.com/en-us/rest/api/storageservices/get-page-ranges#response-body
"The response body includes a list of non-overlapping valid page ranges, sorted by increasing address page range."

sdk/storage/azure-storage-blob/azure/storage/blob/_models.py
jalauzon-msft: So, I actually initially tried to move this to another helper file (like `_deserialize` but the tricky part is this function depends on `PageRange` so moving it to any other file creates a circular dependency :)

Any suggestions to get around this? Could also just move this to a function at the top of the file.

sdk/storage/azure-storage-blob/azure/storage/blob/_models.py
annatisch: Sounds like lemonade to me.

sdk/storage/azure-storage-blob/azure/storage/blob/_models.py
annatisch: Just at the top of the file is fine :)

jalauzon-msft: Sounds good. I've moved it there.

PR #23661, opened 2022-03-25
[Storage] Attempt to fix flaky Live test

jalauzon-msft:
This test has been flaky in our Live Test pipeline for quite some time. It seems the service is flaky about sending back `x-ms-blob-tier-inferred` on a new blob. We need to follow-up with the service team on this but temporarily commenting out the offending line in hopes to stop this test from failing our Live Test pipeline.


vincenttran-msft (approved):
LGTM!

PR #23673, opened 2022-03-25
[formrecognizer] Fix dictionary methods on DocumentField model

catalinaperalta:
This PR adds support for converting DocumentFields of type list, dict, and CurrencyValue to and from a dictionary.

### DIFF
--- a/sdk/formrecognizer/azure-ai-formrecognizer/tests/test_dac_analyze_prebuilts.py
+++ b/sdk/formrecognizer/azure-ai-formrecognizer/tests/test_dac_analyze_prebuilts.py
@@ -5,6 +5,7 @@
 # ------------------------------------
 
 import pytest
+import json
 import functools
 from datetime import date, time
 from devtools_testutils import recorded_by_proxy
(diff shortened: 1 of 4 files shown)


sdk/formrecognizer/azure-ai-formrecognizer/tests/test_dac_analyze_prebuilts.py
```diff
@@ -182,6 +184,9 @@ def test_invoice_jpg(self, client, **kwargs):
         poller = client.begin_analyze_document("prebuilt-invoice", invoice)
 
         result = poller.result()
+        d = result.to_dict()
```
kristapratico, line 187: invoices have nested items right? just want to make sure we're converting those to_dict() as well.

kristapratico (approved):

sdk/formrecognizer/azure-ai-formrecognizer/tests/test_dac_analyze_prebuilts.py
catalinaperalta: I'm using invoices to get the CurrencyValue test, but receipt which is the other test does have the nested items.

PR #23689, opened 2022-03-26
[Storage] Add new `copy_source_tags` option to `start_copy_from_url()`

jalauzon-msft:
STG82 adds a new option for synchronous copy operations that is called `copy_source_tags`. This option allows callers to specify whether to copy or replace source blob tags during a copy operation.

NOTE: The generated code for this change includes code for other STG82 features as well, which will be implemented in other PRs.


sdk/storage/azure-storage-blob/azure/storage/blob/_generated/operations/_blob_operations.py
```diff
@@ -3628,6 +3632,9 @@ def copy_from_url(  # pylint: disable=inconsistent-return-statements
         :param copy_source_authorization: Only Bearer type is supported. Credentials should be a valid
          OAuth access token to copy source. Default value is None.
         :type copy_source_authorization: str
+        :param copy_source_tags: Optional, default 'replace'.  Indicates if source tags should be
```
vincenttran-msft, line 3635: nit: Remove one of the whitespaces between the period and the capital I in 'Indicates'

sdk/storage/azure-storage-blob/azure/storage/blob/_generated/operations/_block_blob_operations.py
```diff
@@ -889,6 +893,9 @@ def put_blob_from_url(  # pylint: disable=inconsistent-return-statements
         :param copy_source_authorization: Only Bearer type is supported. Credentials should be a valid
          OAuth access token to copy source. Default value is None.
         :type copy_source_authorization: str
+        :param copy_source_tags: Optional, default 'replace'.  Indicates if source tags should be
```
vincenttran-msft, line 896: Same nit: Extra whitespace

vincenttran-msft (changes requested):
LGTM! Just resolve the extra whitespace and should be good to go :)

sdk/storage/azure-storage-blob/azure/storage/blob/_generated/operations/_blob_operations.py
jalauzon-msft: This is generated code so unfortunately, I cannot change it :)

sdk/storage/azure-storage-blob/azure/storage/blob/_generated/operations/_block_blob_operations.py
jalauzon-msft: Same comment as above, this is generated.

PR #23715, opened 2022-03-28
Fix `BlobSasPermission` default value for `tag`

jalauzon-msft:
Bringing the hotfix that was just released as `12.11.0` into `main`. Changing the default value for `tag` on `BlobSasPermission` to `False`.


vincenttran-msft (approved):
LGTM!

PR #23717, opened 2022-03-28
[Cosmos] AAD authentication async client

simorenoh:
This PR has the changes for the async client to utilize AAD authentication.

The way the @azure.identity package uses AAD credentials to authenticate services is by adding those credentials into a policy that runs when requests are sent to the core pipelines. This policy makes sure to refresh the current token if needed and set the authentication header of requests going to the pipeline. The reason why Cosmos had to create their own policy in this instance is due to the prefix we utilize for our tokens, since the bearer token policy given by the identity module sends a different prefix altogether and as such does not work for us.

It was also recommended by the identity team to create our own policies entirely rather than attempting to override a couple methods, since this could break us on their end - specially for the `_update_headers()` method since it's private.

For the async client, the credentials seem to also require their context managers to be in place in order to not run into "Unclosed client session" errors once the context is over. Looks kindda weird with the double `async with`, so if there's any suggestions on this do let me know.

Sample is a simple run-through of what can and can't be done, if you think adding more examples would be helpful I can do so as well.


sdk/cosmos/azure-cosmos/samples/access_cosmos_with_aad_async.py
```diff
@@ -0,0 +1,111 @@
+from azure.cosmos.aio import CosmosClient
```
kushagraThapar, line 1: Missing license. 
We should also fix the existing samples and tests if the licenses are missing. 
Also, wondering why doesn't the azure engineering tooling catches this? 
cc @annatisch  - we do this on java SDKs, wondering if there is anything for this on python SDK ?

kushagraThapar (approved):
Thanks @simorenoh , looks good to me!

sdk/cosmos/azure-cosmos/samples/access_cosmos_with_aad_async.py
annatisch: That's actually a good question to which I don't have a good answer.
Let me raise it with the team and see if we can get a CI check added for that :)

xinlian12 (approved):
LGTM, thanks

sdk/cosmos/azure-cosmos/samples/access_cosmos_with_aad_async.py
simorenoh: Just realized none of our samples have licenses up top, will add them now

PR #23722, opened 2022-03-29
AMQP websocket implementation

rakshith91:


sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_transport.py
```diff
@@ -653,5 +654,82 @@ def Transport(host, connect_timeout=None, ssl=False, **kwargs):
...
+        if self._http_proxy:
+            http_proxy_host = self._http_proxy['proxy_hostname']
+            http_proxy_port = self._http_proxy['proxy_hostname']
+            username = self._http_proxy.get('username', None)
+            password = self._http_proxy.get('password', None)
+            if username or password:
+                http_proxy_auth = (username, password)
+        try:
+            from websocket import create_connection
+            # TODO: transform ssl to sslopt
```
rakshith91, line 686:
```suggestion
```

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/message.py
```diff
@@ -265,3 +265,5 @@ def __init__(self, message, state=MessageDeliveryState.WaitingToBeSent, expiry=N
         self.reason = None
         self.delivery = None
         self.error = None
+
+__all__ = ['Message']
```
rakshith91, line 269:
```suggestion
```

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_transport.py
```diff
@@ -392,7 +393,8 @@ def read(self, verify_frame_type=0, **kwargs):  # TODO: verify frame type?
         read_frame_buffer = BytesIO()
         try:
             frame_header = memoryview(bytearray(8))
-            read_frame_buffer.write(read(8, buffer=frame_header, initial=True))
+            data = read(8, buffer=frame_header, initial=True)
+            read_frame_buffer.write(data)
```
rakshith91, line 397:
```suggestion
            read_frame_buffer.write(read(8, buffer=frame_header, initial=True))
```

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_transport.py
```diff
@@ -47,11 +47,12 @@
 from threading import Lock
 
 import certifi
+import websockets
```
rakshith91, line 50:
```suggestion
```

rakshith91, line 50:
```suggestion
```

sdk/eventhub/azure-eventhub/azure/eventhub/_client_base.py
```diff
@@ -309,8 +309,6 @@ def _create_auth(self):
             functools.partial(self._credential.get_token, JWT_TOKEN_SCOPE),
             token_type=token_type,
             timeout=self._config.auth_timeout,
-            http_proxy=self._config.http_proxy,
-            transport_type=self._config.transport_type,
```
yunhaoling, line 313: I don't think those two lines should be removed?

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_connection.py
```diff
@@ -78,10 +83,12 @@ def __init__(self, endpoint, **kwargs):
         self.state = None  # type: Optional[ConnectionState]
 
         transport = kwargs.get('transport')
+        self._transport_type = kwargs.pop('transport_type')
         if transport:
             self._transport = transport
         elif 'sasl_credential' in kwargs:
-            self._transport = SASLTransport(
+            func = SASLWithWebSocket if self._transport_type is TransportType.AmqpOverWebsocket else SASLTransport
```
yunhaoling, line 90: if http proxy is passed, we should also use `SASLWithWebSocket`.

also I think we could change to a better name:
```suggestion
            sasl_transport = SASLWithWebSocket if (self._transport_type is TransportType.AmqpOverWebsocket or kwargs.get("http_proxy") else SASLTransport
```

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_connection.py
```diff
@@ -78,10 +83,12 @@ def __init__(self, endpoint, **kwargs):
...
+            self._transport = func(
```
yunhaoling, line 91: if http proxy is passed, we should also use `SASLWithWebSocket`.

also I think we could change to a better name:
```suggestion
            sasl_transport = SASLWithWebSocket if (self._transport_type is TransportType.AmqpOverWebsocket or kwargs.get("http_proxy") else SASLTransport
            self._transport = sasl_transport(
```

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_transport.py
```diff
@@ -653,5 +652,81 @@ def Transport(host, connect_timeout=None, ssl=False, **kwargs):
     Given a few parameters from the Connection constructor,
     select and create a subclass of _AbstractTransport.
     """
+    transport_type = kwargs.pop('transport_type')
+    if transport_type == TransportType.AmqpOverWebsocket:
+        transport = WebSocketTransport
     transport = SSLTransport if ssl else TCPTransport
```
yunhaoling, line 658: wouldn't the transport be overwritten by line 658 `transport = SSLTransport if ssl else TCPTransport` if `transport_type` is passed?

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_transport.py
```diff
@@ -653,5 +652,81 @@ def Transport(host, connect_timeout=None, ssl=False, **kwargs):
...
+    def _write(self, s):
+        """Completely write a string to the peer."""
+        try:
+            """
+            ABNF, OPCODE_BINARY = 0x2
+            See http://tools.ietf.org/html/rfc5234
+            http://tools.ietf.org/html/rfc6455#section-5.2
+            """
+            from websocket import ABNF
+            self.ws.send(s, opcode=ABNF.OPCODE_BINARY)
```
yunhaoling, line 730: I don't think we want to handle import error in the write method, it would cause slowness, as write will be called millions of times for sending.
we can just but the value 0x2.

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_connection_async.py
```diff
@@ -66,7 +72,8 @@ def __init__(self, endpoint, **kwargs):
         if transport:
             self.transport = transport
         elif 'sasl_credential' in kwargs:
-            self.transport = SASLTransport(
+            func = SASLWithWebSocket if self._transport_type is TransportType.AmqpOverWebsocket else SASLTransport
+            self._transport = func(
```
yunhaoling, line 76: same as the sync comment -- when http proxy is passed, we need SSLWithWebsocket

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_transport_async.py
```diff
@@ -360,3 +361,48 @@ async def negotiate(self):
...
+            ABNF, OPCODE_BINARY = 0x2
+            See http://tools.ietf.org/html/rfc5234
+            http://tools.ietf.org/html/rfc6455#section-5.2
+            """
+            from websocket import ABNF
+            await asyncio.get_event_loop().run_in_executor(
+                    None, self.ws.send, s, ABNF.OPCODE_BINARY
+                    )
+        except ImportError:
+            raise ValueError("Please install websocket-client library to use websocket transport.")
```
yunhaoling, line 408: same comment, we could just place code "0x2" instead of importing each time in "write"

yunhaoling (commented):
This looks great in general!
left some minor comments

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_transport.py
rakshith91: using the `send_binary` convinience

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_transport.py
```diff
@@ -653,5 +652,77 @@ def Transport(host, connect_timeout=None, ssl=False, **kwargs):
...
+            host, port, connect_timeout, **kwargs
+            )
+        self.ws = None
+        self._http_proxy = kwargs.get('http_proxy', None)
+
+    def connect(self):
+        http_proxy_host, http_proxy_port, http_proxy_auth = None, None, None
+        if self._http_proxy:
+            http_proxy_host = self._http_proxy['proxy_hostname']
+            http_proxy_port = self._http_proxy['proxy_hostname']
```
kashifkhan, line 678: will this be `self._http_proxy['proxy_port']` :)?

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/sasl.py
```diff
@@ -78,26 +103,23 @@ def __init__(self, host, credential, port=AMQPS_PORT, connect_timeout=None, ssl=
...
+
+class SASLWithWebSocket(WebSocketTransport, SASLTransportMixin):
+
+    def __init__(self, host, credential, port=WEBSOCKET_PORT, connect_timeout=None, ssl=None, **kwargs):
+        self.credential = credential
+        ssl = ssl or True
+        http_proxy = kwargs.pop('http_proxy', None)
+        self._transport = WebSocketTransport(
+            host,
+            port=WEBSOCKET_PORT,
```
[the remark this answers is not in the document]
rakshith91: fixed

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_connection.py
```diff
@@ -78,10 +83,14 @@ def __init__(self, endpoint, **kwargs):
         self.state = None  # type: Optional[ConnectionState]
 
         transport = kwargs.get('transport')
+        self._transport_type = kwargs.pop('transport_type')
         if transport:
             self._transport = transport
         elif 'sasl_credential' in kwargs:
-            self._transport = SASLTransport(
+            sasl_transport = SASLWithWebSocket if (
+                self._transport_type is TransportType.AmqpOverWebsocket or kwargs.get("http_proxy")
```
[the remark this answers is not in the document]
rakshith91: that is correct

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_sasl_async.py
```diff
@@ -100,3 +100,20 @@ async def negotiate(self):
             return
         else:
             raise ValueError("SASL negotiation failed.\nOutcome: {}\nDetails: {}".format(*fields))
+
+class SASLWithWebSocket(SASLTransport):
+
+    def __init__(
+        self, host, credential, port=WEBSOCKET_PORT, connect_timeout=None, ssl=None, **kwargs
+        ): # pylint: disable=super-init-not-called
```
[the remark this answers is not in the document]
rakshith91: They are all documented in the parent class - I can add the link though

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_connection_async.py
```diff
@@ -49,11 +50,16 @@ class Connection(object):
...
+    :keyword str transport_type: Determines if the transport type is Amqp or AmqpOverWebSocket.
+    :keyword Dict http_proxy: HTTP proxy settings. This must be a dictionary with the following
+     keys: `'proxy_hostname'` (str value) and `'proxy_port'` (int value).
+     Additionally the following keys may also be present: `'username', 'password'`.
     """
 
     def __init__(self, endpoint, **kwargs):
         parsed_url = urlparse(endpoint)
         self.hostname = parsed_url.hostname
+        self._transport_type = kwargs.pop('transport_type')
```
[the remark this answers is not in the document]
rakshith91: It's not exposed to the public. We are always passing a transport_type

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_transport.py
```diff
@@ -653,5 +652,77 @@ def Transport(host, connect_timeout=None, ssl=False, **kwargs):
     Given a few parameters from the Connection constructor,
     select and create a subclass of _AbstractTransport.
     """
-    transport = SSLTransport if ssl else TCPTransport
+    transport_type = kwargs.pop('transport_type')
```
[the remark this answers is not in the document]
rakshith91: same as async

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_transport.py
```diff
@@ -653,5 +652,77 @@ def Transport(host, connect_timeout=None, ssl=False, **kwargs):
     Given a few parameters from the Connection constructor,
```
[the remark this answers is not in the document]
rakshith91: https://github.com/Azure/azure-sdk-for-python/issues/24197


good catch - no idea, but it should be fixed. Created an issue. Let's track it in a different PR

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_transport.py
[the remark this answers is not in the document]
rakshith91: yep - transport_type is always passed. Again, it's not publicly exposed

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_sasl_async.py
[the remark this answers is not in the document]
swathipil: sounds good, thanks!

swathipil (commented):
version + changelog updates? I think you might not have the latest updates on your branch, b/c changelog doesn't include 5.8.0a1-3?

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_transport.py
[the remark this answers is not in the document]
annatisch: hmm I think we could be more intentional about that. It's currently treated as a required parameter in multiple places, but is never described as such. I don't like relying on just because it's currently "always passed in". If that is the case - let's just make it explicit.
I would say make it required in one place - probably the transport seems like a good fit - and actually make it a positional parameter. Then at the Connection layer or the Client layer we implement the correct "default" logic that gets passed in.

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_transport.py
```diff
@@ -653,5 +652,77 @@ def Transport(host, connect_timeout=None, ssl=False, **kwargs):
...
+            http_proxy_host = self._http_proxy['proxy_hostname']
+            http_proxy_port = self._http_proxy['proxy_port']
+            username = self._http_proxy.get('username', None)
+            password = self._http_proxy.get('password', None)
+            if username or password:
+                http_proxy_auth = (username, password)
+        try:
+            from websocket import create_connection
+            self.ws = create_connection(
+                url="wss://{}/$servicebus/websocket/".format(self._host),
```
annatisch, line 686: This should be removed out of pyamqp and into the SDK layer.

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_transport.py
```diff
@@ -653,5 +652,77 @@ def Transport(host, connect_timeout=None, ssl=False, **kwargs):
...
+                skip_utf8_validation=True,
+                sslopt=self.sslopts,
+                http_proxy_host=http_proxy_host,
+                http_proxy_port=http_proxy_port,
+                http_proxy_auth=http_proxy_auth
+            )
+        except ImportError:
+            raise ValueError("Please install websocket-client library to use websocket transport.")
+
+    def _read(self, n, buffer=None, **kwargs): # pylint: disable=unused-arguments
```
annatisch, line 698: Is this logic identical to the existing `read` method on the non-WS transport?

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_sasl_async.py
```diff
@@ -100,3 +100,20 @@ async def negotiate(self):
             return
         else:
             raise ValueError("SASL negotiation failed.\nOutcome: {}\nDetails: {}".format(*fields))
+
+class SASLWithWebSocket(SASLTransport):
```
annatisch, line 104: What's different about this class than the base SASLTransport class other than the construction of the WebSocketTransport?

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_transport_async.py
```diff
@@ -360,3 +361,43 @@ async def negotiate(self):
...
+        """Read exactly n bytes from the peer."""
+        
+        length = 0
+        view = buffer or memoryview(bytearray(n))
+        nbytes = self._read_buffer.readinto(view)
+        length += nbytes
+        n -= nbytes
+        while n:
+            data = await asyncio.get_event_loop().run_in_executor(
+                None, self.ws.recv
```
annatisch, line 377: So this websocket library doesn't support async? Were there any that do?

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/constants.py
```diff
@@ -14,6 +14,11 @@
 #: The port number is reserved for future transport mappings to these protocols.
 PORT = 5672
 
+# default port for AMQP over Websocket
+WEBSOCKET_PORT = 443
+
+# subprotocol for AMQP over Websocket
+AMQP_WS_SUBPROTOCOL = 'AMQPWSB10'
```
annatisch, line 21: Is this universal to AMQP? Or is this somehow Azure specific?

sdk/eventhub/azure-eventhub/tests/pyamqp_tests/synctests/test_websocket.py
```diff
@@ -0,0 +1,27 @@
...
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for license information.
+# --------------------------------------------------------------------------------------------
+
+import pytest
+
+from azure.eventhub._pyamqp import authentication, ReceiveClient
+from azure.eventhub._pyamqp.constants import TransportType
+
+def test_event_hubs_client_web_socket(live_eventhub):
```
annatisch, line 11: Async test?
And maybe some kind of test for the proxy settings?

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/constants.py
rakshith91: it's universal 
http://docs.oasis-open.org/amqp-bindmap/amqp-wsb/v1.0/amqp-wsb-v1.0.html

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_transport_async.py
rakshith91: no - unfortunately, this doesn't.

There is no library which supports both sync and async. We have other libraries which are async only.

So the two choices we have are:
1. Go with the wrapper for sync implementation like here
2. Recommend different libraries for both sync and async.

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_sasl_async.py
rakshith91: For the async version - nothing. Only the constructor is different

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_transport.py
rakshith91: nope - we'd have to use our own buffere here, since websocket does not have the capability to read n bytes at a time

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_transport.py
rakshith91: Can you clarify? Would you expect the users to provide a wss link by themselves?

annatisch: This doesn't look like it's generic - it looks specific to Azure services. So I would expect this value to be supplied by the EH or SB SDKs.
(So yes, customers of the pyamqp library would probably need to supply this - how was it being passed into uamqp?)

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_sasl_async.py
annatisch: In that case I don't think there's a reason to create a new class for this?

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_transport_async.py
annatisch: I'm leaning towards recommending different libraries...
We already do that for HTTP....
What's the quality/stability/maintenance of the async lib like?

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_connection.py
```diff
@@ -78,10 +83,14 @@ def __init__(self, endpoint, **kwargs):
         self.state = None  # type: Optional[ConnectionState]
 
         transport = kwargs.get('transport')
+        self._transport_type = kwargs.pop('transport_type')
```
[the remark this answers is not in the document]
rakshith91: A transport_type is always passed to a connection by the client. I can make it optional neverthless

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_sasl_async.py
rakshith91: the constructors are quite different and will involve too much unnecessary code - i think a better idea would be to put negotiate in a mixin and then use different constructors.

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_transport.py
rakshith91: hmm, That makes sense, but uamqp has been doing this instead of taking it from the client.

https://github.com/Azure/azure-uamqp-python/blob/master/src/wsio.pyx#L14

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_transport_async.py
rakshith91: realized async library i was looking at doesn't support http proxy - so not an option for now 
https://github.com/aaugustin/websockets/issues/364

sdk/eventhub/azure-eventhub/tests/pyamqp_tests/synctests/test_websocket.py
rakshith91: https://github.com/Azure/azure-uamqp-python/pull/323/files#diff-7ef7f7a671052d98771c46e57dc5b5fa98775fbe1e2e394d71df21cc25c792df

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_sasl_async.py
rakshith91: actually, updated this part a little bit - we don't need an async transport super class anymore

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_client_async.py
```diff
@@ -176,6 +176,23 @@ async def _do_retryable_operation_async(self, operation, *args, **kwargs):
...
+                if elapsed_time >= interval:
+                    _logger.info("Keeping %r connection alive. %r",
+                                 self.__class__.__name__,
+                                 self._connection._container_id)
+                    await self._connection._get_remote_timeout(current_time)
+                    start_time = current_time
+                await asyncio.sleep(1)
+        except Exception as e:  # pylint: disable=broad-except
+            _logger.info("Connection keep-alive for %r failed: %r.", self.__class__.__name__, e)
+
```
rakshith91, line 195: this method has not been vendored into this version, but existed in uamqp branch - adding this - this is not related to websocket , and create a different PR for this if required

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_connection.py
```diff
@@ -63,6 +64,10 @@ class Connection(object):
      Default value is `0.1`.
     :keyword bool network_trace: Whether to log the network traffic. Default value is `False`. If enabled, frames
      will be logged at the logging.INFO level.
+    :keyword str transport_type: Determines if the transport type is Amqp or AmqpOverWebSocket.
+    :keyword Dict http_proxy: HTTP proxy settings. This must be a dictionary with the following
+     keys: `'proxy_hostname'` (str value) and `'proxy_port'` (int value).
```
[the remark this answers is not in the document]
rakshith91: https://github.com/Azure/azure-sdk-for-python/issues/24306

I will track it here

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_connection.py
```diff
@@ -63,6 +64,10 @@ class Connection(object):
      Default value is `0.1`.
     :keyword bool network_trace: Whether to log the network traffic. Default value is `False`. If enabled, frames
      will be logged at the logging.INFO level.
+    :keyword str transport_type: Determines if the transport type is Amqp or AmqpOverWebSocket.
```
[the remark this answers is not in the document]
rakshith91: raised - also updated the documentation

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_transport_async.py
```diff
@@ -412,3 +418,76 @@ async def negotiate(self):
...
+
+
+class WebSocketTransportAsync(AsyncTransportMixin):
+    def __init__(self, host, port=WEBSOCKET_PORT, connect_timeout=None, ssl=None, **kwargs
+        ):
+        self.sslopts = ssl if isinstance(ssl, dict) else {}
+        self._connect_timeout = connect_timeout
+        self.host = host
+        super().__init__(
+            )
```
rakshith91, line 430:
```suggestion
        super().__init__()
```

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_transport_async.py
```diff
@@ -412,3 +418,76 @@ async def negotiate(self):
...
+        self.socket_lock = asyncio.Lock()
```
rakshith91, line 431:
```suggestion
```

swathipil (commented):
I think you'll need to update the changelog + version (to a4 not b4, which is what I bumped the version to 🤦)

sdk/eventhub/azure-eventhub/tests/pyamqp_tests/synctests/test_websocket.py
rakshith91: for proxy, CI doesn't support it - running this sample locally using https://github.com/abhinavsingh/proxy.py
https://github.com/Azure/azure-sdk-for-python/blob/feature/eventhub/pyproto/sdk/eventhub/azure-eventhub/samples/sync_samples/proxy.py

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_transport.py
```diff
@@ -647,11 +646,82 @@ def _read(self, n, initial=False, _errnos=(errno.EAGAIN, errno.EINTR)):
...
+        while n:
+            data = self.ws.recv()
+
+            if len(data) <= n:
+                view[length: length + len(data)] = data
+                n -= len(data)
+            else:
+                view[length: length + n] = data[0:n]
+                self._read_buffer = BytesIO(data[n:])
+                n = 0
```
swathipil, line 714: I might be missing some understanding of logic here. Does `length` need to be updated at the end of the `if` block here? (It looks like loop exits after going into the else block, so I'm assuming not there.)

same question for async.

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_connection_async.py
```diff
@@ -58,24 +59,33 @@ class Connection(object):
...
+    :keyword Dict http_proxy: HTTP proxy settings. This must be a dictionary with the following
+     keys: `'proxy_hostname'` (str value) and `'proxy_port'` (int value). When using these settings,
+     the transport_type would be AmqpOverWebSocket.
+     Additionally the following keys may also be present: `'username', 'password'`.
     """
 
     def __init__(self, endpoint, **kwargs):
         parsed_url = urlparse(endpoint)
         self.hostname = parsed_url.hostname
+        self._transport_type = kwargs.pop('transport_type', TransportType.Amqp)
```
swathipil, line 73: same comment about whether we need to change default if http_proxy is passed in.

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_connection_async.py
```diff
@@ -58,24 +59,33 @@ class Connection(object):
...
-
         transport = kwargs.get('transport')
         if transport:
             self.transport = transport
         elif 'sasl_credential' in kwargs:
-            self.transport = SASLTransport(
+            sasl_transport = SASLWithWebSocket if (
+                self._transport_type is TransportType.AmqpOverWebsocket or kwargs.get("http_proxy")
+                ) else SASLTransport
+            self.transport = sasl_transport(
```
swathipil, line 88: It looks like the `AsyncTransport` on line 94 does not take self._transport_type the way the sync Transport does. Do you know why this is? (or have links to docs that explain why :P)

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/client.py
```diff
@@ -155,6 +156,12 @@ def __init__(self, hostname, auth=None, **kwargs):
         self._receive_settle_mode = kwargs.pop('receive_settle_mode', ReceiverSettleMode.Second)
         self._desired_capabilities = kwargs.pop('desired_capabilities', None)
 
+        # transport
+        if kwargs.get('transport_type') is TransportType.Amqp and kwargs.get('http_proxy') is not None:
+            raise ValueError("Http proxy settings can't be passed if transport_type is explicitly set to Amqp")
+        self._transport_type = kwargs.pop('transport_type', TransportType.Amqp)
+        self._http_proxy = kwargs.pop('http_proxy', None)
```
swathipil, line 163: should the default transport type change if http_proxy is passed in? (same comment as above).

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_connection.py
```diff
@@ -92,16 +99,20 @@ def __init__(self, endpoint, **kwargs):
         self.state = None  # type: Optional[ConnectionState]
 
         transport = kwargs.get('transport')
+        self._transport_type = kwargs.pop('transport_type', TransportType.Amqp)
```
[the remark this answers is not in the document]
rakshith91: the transport type will eventually be passed into transport where the this condition is checked. So, this won't be necessary

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_transport.py
rakshith91: nope - we don't need to

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_connection_async.py
rakshith91: AsyncTransport works a little bit differently than the sync scenario - there is no Abstract Transport in async which is why we don't need to set which transport (among TCP, SSL) explicitly is my undesrtanding

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/client.py
rakshith91: same answer as above :P
the transport type will eventually be passed into transport where the this condition is checked. So, this won't be necessary

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_transport.py
swathipil: cool! will make a note to look into memoryview + come back to this code to understand.

swathipil (approved):
LGTM!

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_transport_async.py
annatisch: Okay - can we open an issue to track this? At a minimum we should refactor this so that it will be compatible with an appropriate async implementation when it lands.

sdk/eventhub/azure-eventhub/tests/pyamqp_tests/synctests/test_websocket.py
annatisch: We can also use mocked offline tests for this kind of thing to at least ensure the correct code path is followed with the correct parameters.

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_transport_async.py
rakshith91: Tracking it here https://github.com/Azure/azure-sdk-for-python/issues/24315

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_transport.py
annatisch: An opportunity to improve ;)

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_transport_async.py
```diff
@@ -81,8 +82,77 @@ def get_running_loop():
             loop = asyncio.get_event_loop()
         return loop
 
+class AsyncTransportMixin():
+    def __init__(self):
```
annatisch, line 86: A mixin should not have a constructor - and these attributes are only used by websockets anyway - so let's move it to that child constructor

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_transport_async.py
```diff
@@ -412,3 +418,74 @@ async def negotiate(self):
...
+    async def _read(self, n, buffer=None, **kwargs): # pylint: disable=unused-arguments
+        """Read exactly n bytes from the peer."""
+        
+        length = 0
+        view = buffer or memoryview(bytearray(n))
+        nbytes = self._read_buffer.readinto(view)
+        length += nbytes
+        n -= nbytes
+        while n:
+            data = await asyncio.get_event_loop().run_in_executor(
```
annatisch, line 466: You already assigned the loop to self.loop?

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_transport_async.py
rakshith91: done

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_transport_async.py
rakshith91: leveraging it

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_transport.py
rakshith91: done

rakshith91:
Clicked merge by mistake - https://github.com/Azure/azure-sdk-for-python/pull/24345

made one small change in this PR

Issue #23063, opened 2022-02-15
[EventHubs] Preview BufferedProducer


yunhaoling:
- Implement the buffered producer feature

jsquire:
@kashifkhan - Is it safe to close this one, since the buffered producer has been released?

kashifkhan:
yes this one can be closed

Issue #23062, opened 2022-02-15
[EventHubs] Buffered Producer API Review


yunhaoling:
aggregate client design for the event hubs buffered producer feature

- gist: https://gist.github.com/yunhaoling/c300396e293c06beeb73c769012a9d2d
- apiview: https://apiview.dev/Assemblies/Review/e20c9d6e94a244ffbe448faaade3050f

feedbacks from api review meeting:

- action items for behaviors/implementation:
  - investigate into callback reentrancy behavior questions and define the behavior
    - what happens if calling close/flush in on_error callback (e.g. think about infinite error case)
  - definition for various timeout behavior and how users could handle
    - send_event/send_batch times out
    - flush/close times out
  - investigation of the buffer max behavior (limitation vs timing control? back pressure?)

- action items for API:
  - on_error/on_success should be aware of input message types
  - Python prefers shorter parameter names and description in docstring

PR #23748, opened 2022-03-30
[EventHubs] Buffered Producer Implementation

yunhaoling:
## Overall architecture:

**EHProducerClient 1 -> 1 EHBufferedProducerDispatcher 1 -> n EHBufferedProducer (for single partition)**

### Duty of EHBufferedProducerDispatcher:
1. life cycle management of all single partition buffered producer
2. dispatch events (partition determination) to single partition buffered producer
3. flush all partitions
4. statics information from all single partition buffered producer

### Duty of EHBufferedProducer
1. flush(send) events
2. max_wait_time flush timing control
3. backpressure handling
4. buffering
5. batching

### what happens if calling close/flush in on_error callback (e.g. think about infinite error case)

deadlock in current implementation, we should be explicit that users should not perform flush in on_error/on_success.

### definition for various timeout behavior and how users could handle
- send_event/send_batch times out: retry
- flush/close times out: flush timing out would raise error, close time out would log warning

### investigation of the buffer max behavior (limitation vs timing control? back pressure?)

in current implementation, if the remaining buffer isn't enough for the incoming events, call flush to clear the buffer first

## Tasks

- [x] sync implementation
- [x] async implementation
- [x] test
- [x] doc
- [x] mypy/pylint
- [x] sample

## Related issue:

https://github.com/Azure/azure-sdk-for-python/issues/23062
https://github.com/Azure/azure-sdk-for-python/issues/23063


sdk/eventhub/azure-eventhub/samples/sync_samples/send_buffered_mode.py
```diff
@@ -0,0 +1,48 @@
...
+    on_error=on_error
+)
+
+start_time = time.time()
+with producer:
+    for i in range(10):
+        producer.send_event(EventData('Single data {}'.format(i)))
+
+    batch = producer.create_batch()
+    for i in range(10):
```
swathipil, line 44: maybe a comment above each of these explaining that you can either: 1) send events without batching or 2) can still send batches?

sdk/eventhub/azure-eventhub/samples/sync_samples/send_buffered_mode.py
```diff
@@ -0,0 +1,48 @@
...
+    eventhub_name=EVENTHUB_NAME,
+    buffered_mode=True,
+    on_success=on_success,
+    on_error=on_error
+)
+
+start_time = time.time()
+with producer:
+    for i in range(10):
+        producer.send_event(EventData('Single data {}'.format(i)))
```
swathipil, line 41: is there a sample using flush? or is that not a necessary/common scenario?

sdk/eventhub/azure-eventhub/azure/eventhub/_buffered_producer/_buffered_producer.py
```diff
@@ -0,0 +1,221 @@
...
+                raise OperationTimeoutError("Failed to enqueue events into buffer due to timeout.")
+
+            try:
+                # add single event into current batch
+                self._cur_batch.add(events)
+            except AttributeError:
+                # if the input events is a EventDataBatch, put the whole into the buffer
+                self._buffered_queue.put(events)
+                # create a new batch for incoming events
+                self._cur_batch = EventDataBatch(self._max_message_size_on_link)
```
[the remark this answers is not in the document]
annatisch: Actually nevermind - reading down a little further it looks like new Batches are added to the queue as they are created - so if I had a half-full one, it's already on the queue. Looks like that's fine :)

sdk/eventhub/azure-eventhub/azure/eventhub/_buffered_producer/_buffered_producer.py
```diff
@@ -0,0 +1,221 @@
...
+            with self._not_empty:
+                if not self._buffered_queue.qsize():
+                    _LOGGER.info("Partition {} worker is awaiting data.".format(self.partition_id))
+                    self._not_empty.wait()
+                now_time = time.time()
+                _LOGGER.info("Partition {} worker is checking max_wait_time.".format(self.partition_id))
+                if now_time - self._last_send_time > self._max_wait_time:
+                    # in the worker, not raising error for flush, users can not handle this
+                    self.flush(raise_error=False)
+                    self._last_send_time = now_time
```
annatisch, line 216: I think this is already being set in `flush`

sdk/eventhub/azure-eventhub/azure/eventhub/_buffered_producer/_buffered_producer.py
```diff
@@ -0,0 +1,221 @@
...
+        except Exception as exc:
+            self._on_error(events, exc, self.partition_id)
+        finally:
+            self._cur_buffered_len -= len(events)
+            self._max_concurrent_sends_semaphore.release()
+
+    def flush(self, timeout=None, raise_error=True):
+        # try flushing all the buffered batch within given time
+        _LOGGER.info("Partition: {} started flushing.".format(self.partition_id))
+        with self._not_empty:
```
annatisch, line 156: Can I use this Condition of the max_wait_time_worker is already holding/waiting on it?

sdk/eventhub/azure-eventhub/azure/eventhub/_buffered_producer/_buffered_producer.py
```diff
@@ -0,0 +1,221 @@
...
+            timeout_time = time.time() + timeout if timeout else None
+            while not self._buffered_queue.empty():
+                remaining_time = timeout_time - time.time() if timeout_time else None
+                # If flush could get the semaphore, perform sending
+                if ((remaining_time and remaining_time > 0) or remaining_time is None) and \
+                        self._max_concurrent_sends_semaphore.acquire(timeout=remaining_time):
+                    batch = self._buffered_queue.get()
+                    if not batch:
```
annatisch, line 164: This means the queue was empty?

sdk/eventhub/azure-eventhub/azure/eventhub/_buffered_producer/_buffered_producer.py
```diff
@@ -0,0 +1,221 @@
...
+    def enhanced_callback_decorator(self, callback):
+        def wrapper_callback(*args, **kwargs):
+            try:
+                callback(*args, **kwargs)
+            except Exception as exc:
+                _LOGGER.warning("callback exception", callback.__name__, exc, self.partition_id)
+
+        return wrapper_callback
+
+    def sent_callback(self, events, future: Future):
```
annatisch, line 143: Where is this used?

sdk/eventhub/azure-eventhub/azure/eventhub/_buffered_producer/_buffered_producer.py
```diff
@@ -0,0 +1,221 @@
...
+        # This method would raise OperationTimeout if the queue does not have enough space for the input and
+        # flush cannot finish in timeout.
+        with self._not_full:
+            try:
+                new_events_len = len(events)
+            except TypeError:
+                new_events_len = 1
+            timeout_time = time.time() + timeout if timeout else None
+
+            if self._max_buffer_len - self._cur_buffered_len < new_events_len:
```
annatisch, line 102: We don't seem to be using the `not_full` condition?

sdk/eventhub/azure-eventhub/azure/eventhub/_buffered_producer/_buffered_producer.py
```diff
@@ -0,0 +1,221 @@
...
+                # If flush could get the semaphore, perform sending
+                if ((remaining_time and remaining_time > 0) or remaining_time is None) and \
+                        self._max_concurrent_sends_semaphore.acquire(timeout=remaining_time):
+                    batch = self._buffered_queue.get()
+                    if not batch:
+                        self._max_concurrent_sends_semaphore.release()
+                        continue
+                    try:
+                        _LOGGER.info("Partition {} is sending.".format(self.partition_id))
+                        self._producer.send(
```
annatisch, line 169: Shouldn't this be a future added to the executor, rather than sent synchronously?
Otherwise where does the 'max_concurrency' take effect?

sdk/eventhub/azure-eventhub/azure/eventhub/_buffered_producer/_buffered_producer_dispatcher.py
```diff
@@ -0,0 +1,147 @@
...
+            _LOGGER.info('Flushing all partitions partially succeeded with result {!r}.'.format(exc_results))
+            # TODO: better error?
+            raise EventDataSendError(
+                message="Flushing all partitions partially succeeded, failed partitions are {!r}.".format(
+                    exc_results.keys()
+                ),
+                details=exc_results
+            )
+
+    def close(self, flush=True, timeout=None, raise_error=False):
```
annatisch, line 112: Nit: Even though it's internal, I feel keyword-only syntax would be nice here....

sdk/eventhub/azure-eventhub/azure/eventhub/_producer.py
```diff
@@ -199,6 +199,8 @@ def _wrap_eventdata(
             if isinstance(
                 event_data, EventDataBatch
             ):  # The partition_key in the param will be omitted.
+                if not event_data:
```
annatisch, line 202: Does this work? From what I can see `EventDataBatch` has no `__bool__`?
Or will it fallback to checking `__len__`?

sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py
```diff
@@ -41,6 +53,38 @@ class EventHubProducerClient(ClientBase):
...
+    :keyword bool buffered_mode: Set to True to get the producer client work in buffered mode. Default is False.
+    :keyword on_success: The callback to be called once a batch has successfully published.
+     The callback takes two parameters: `events` which are the list of events that have been successfully published
+     and `partition_id` which is the partition id that the events in the list have been published to.
+     The callback function should be defined like: `on_success(events, partition_id)`.
+    :paramtype on_success: Optional[Callable[[List[EventData], Optional[str]], None]]
+    :keyword on_error: The callback to be called once a batch has failed to publish.
+     The callback takes three parameters: `events` which are the list of events that failed to be published,
+     `partition_id` which is the partition id that the events in the list have been tried to be published to and
+     `error` being the exception.
```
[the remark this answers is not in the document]
annatisch: We probably need to add that the two callbacks are required when running in buffered mode.

sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py
```diff
@@ -41,6 +53,38 @@ class EventHubProducerClient(ClientBase):
...
+     The callback function should be defined like: `on_error(events, error, partition_id)`.
+     If on_error is passed in non-buffered mode, callback will be called with the error instead of
+     error being raised from the send method.
+    :paramtype on_error: Optional[Callable[[List[EventData], Exception, Optional[str]], None]]
+    :keyword int max_buffer_length: Buffered mode only.
+     The total number of events that can be buffered for publishing at a given time for a given partition.
```
annatisch, line 71: ..."before a flush will be triggered."?

sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py
```diff
@@ -41,6 +53,38 @@ class EventHubProducerClient(ClientBase):
...
+     The default value is 1500 in buffered mode.
+    :keyword Optional[float] max_wait_time: Buffered mode only.
+     The amount of time to wait for a batch to be built with events in the buffer before publishing.
+     The default value is 1 in buffered mode.
+    :keyword Optional[int] max_concurrent_sends: Buffered mode only. The total number of batches that may be sent
+     concurrently, across all partitions. This limit takes precedence over the value specified in
+     max_concurrent_sends, ensuring this maximum is respected. By default, this will be set to
+     twice the number of processors available in the host environment.
+    :keyword Optional[int] max_concurrent_sends: Buffered mode only. The number of batches that may
```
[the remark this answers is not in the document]
annatisch: Yeah it looks like one of the used to be the max concurrency per partition. Not sure which parameter we settled on?

sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py
```diff
@@ -232,12 +451,64 @@ def from_connection_string(cls, conn_str, **kwargs):
...
         """
-        constructor_args = cls._from_connection_string(conn_str, **kwargs)
+        constructor_args = cls._from_connection_string(conn_str, eventhub_name=eventhub_name, **kwargs)
         return cls(**constructor_args)
 
+    def send_event(self, event_data, **kwargs):
+        # type: (Union[EventData, AmqpAnnotatedMessage], Any) -> None
+        """
+        Sends an event data.
+        By default the method will blocks until acknowledgement is received or operation times out.
```
annatisch, line 461:
```suggestion
        By default the method will block until acknowledgement is received or operation times out.
```

sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py
```diff
@@ -232,12 +451,64 @@ def from_connection_string(cls, conn_str, **kwargs):
...
+        If the `EventHubProducerClient` is configured to run in buffered mode, the method will enqueue the event
+        into local buffer and return. The producer will do automatic batching and sending in the background.
```
annatisch, line 463: Do we also want to add a sentence here about where/how errors will be raised depending on the `on_error` handler?

sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py
```diff
@@ -232,12 +451,64 @@ def from_connection_string(cls, conn_str, **kwargs):
...
+        :param event_data: The `EventData` object to be sent.
+        :type event_data: Union[~azure.eventhub.EventData, ~azure.eventhub.amqp.AmqpAnnotatedMessage]
+        :keyword float timeout: The maximum wait time to send the event data.
+         If not specified, the default wait time specified when the producer was created will be used.
```
annatisch, line 467: How does this parameter work in buffered mode?

sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py
```diff
@@ -232,12 +451,64 @@ def from_connection_string(cls, conn_str, **kwargs):
...
+                self._on_error(event_data, exc, partition_id)
+            else:
+                raise
+
     def send_batch(self, event_data_batch, **kwargs):
         # type: (Union[EventDataBatch, SendEventTypes], Any) -> None
-        """Sends event data and blocks until acknowledgement is received or operation times out.
+        """
+        Sends a batch of event data.
+        By default the method will blocks until acknowledgement is received or operation times out.
```
annatisch, line 509:
```suggestion
        By default the method will block until acknowledgement is received or operation times out.
```

sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py
```diff
@@ -232,12 +451,64 @@ def from_connection_string(cls, conn_str, **kwargs):
...
+        If the `EventHubProducerClient` is configured to run in buffered mode, the method will enqueue the events
+        into local buffer and return. The producer will do automatic sending in the background.
```
annatisch, line 511: Maybe we should add a comment that in buffered mode, sending a batch will remain intact, and not be 'repackaged'. This may have implications to the efficiency of sending events.

sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py
```diff
@@ -434,8 +718,32 @@ def close(self):
...
+                self._buffered_producer_dispatcher.close(flush=flush, timeout=timeout)
+                self._buffered_producer_dispatcher = None
+
             for pid in self._producers:
                 if self._producers[pid]:
                     self._producers[pid].close()  # type: ignore
                 self._producers[pid] = None
         super(EventHubProducerClient, self)._close()
+
+    def get_buffered_event_count(self, partition_id: str) -> Optional[int]:
```
annatisch, line 731: Missing docstring

sdk/eventhub/azure-eventhub/azure/eventhub/_buffered_producer/_buffered_producer_dispatcher.py
```diff
@@ -0,0 +1,147 @@
...
+    def _get_partition_id(self, partition_id, partition_key):
+        if partition_id:
+            if partition_id not in self._partition_ids:
+                raise ConnectError(
+                    "Invalid partition {} for the event hub {}".format(
+                        partition_id, self._eventhub_name
+                    )
+                )
+            return partition_id
+        if isinstance(partition_key, str):
```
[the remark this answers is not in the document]
yunhaoling: yes, this is the service behavior, we have documented it in the `send_batch` API:

```
**WARNING: Setting partition_key of non-string value on the events to be sent is discouraged
         as the partition_key will be ignored by the Event Hub service and events will be assigned
         to all partitions using round-robin. Furthermore, there are SDKs for consuming events which expect
         partition_key to only be string type, they might fail to parse the non-string value.**
```

sdk/eventhub/azure-eventhub/azure/eventhub/_buffered_producer/_buffered_producer.py
```diff
@@ -0,0 +1,221 @@
...
+                                self.partition_id, len(batch)
+                            )
+                        )
+                    except Exception as exc:
+                        self._on_error(batch._internal_events, exc, self.partition_id)
+                        _LOGGER.info(
+                            "Partition {} sending {} events failed due to exception: {!r} ".format(
+                                self.partition_id, len(batch), exc
+                            )
+                        )
```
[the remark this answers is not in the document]
yunhaoling: thanks for the suggestion, you're better that it would better be logged before calling user's callback.

btw self._on_error is a decorated callback which will log.warning user's callback exception

sdk/eventhub/azure-eventhub/azure/eventhub/_buffered_producer/_buffered_producer_dispatcher.py
```diff
@@ -0,0 +1,147 @@
...
+            # TODO: better error?
+            _LOGGER.info('Stopping all partitions succeeded with result {!r}.'.format(exc_results))
+            if raise_error:
+                raise EventDataSendError(
+                    message="Stopping partially succeeded, failed partitions are {!r}.".format(exc_results.keys()),
+                    details=exc_results
+                )
+
+        if not self._existing_executor:
+            self._executor.shutdown()
```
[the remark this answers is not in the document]
yunhaoling: yes, if users choose to pass in their own executor, we'd better not shutodwn as it might be used for other tasks.

sdk/eventhub/azure-eventhub/azure/eventhub/_buffered_producer/_buffered_producer.py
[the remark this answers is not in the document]
swathipil: oh cool, didn't know that!

sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py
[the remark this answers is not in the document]
yunhaoling: updated

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_producer_client_async.py
```diff
@@ -184,12 +314,81 @@ def _create_producer(
...
+        """
+        ...
+
     @classmethod
     def from_connection_string(
         cls,
         conn_str: str,
         *,
         eventhub_name: Optional[str] = None,
+        buffered_mode: bool =False,
```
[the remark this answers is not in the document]
yunhaoling: fixed

sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py
```diff
@@ -179,15 +336,77 @@ def _create_producer(self, partition_id=None, send_timeout=None):
...
+        conn_str: str,
+        *,
+        eventhub_name: Optional[str] = None,
+        **kwargs: Any
+    ) -> "EventHubProducerClient":
         """Create an EventHubProducerClient from a connection string.
 
         :param str conn_str: The connection string of an Event Hub.
         :keyword str eventhub_name: The path of the specific Event Hub to connect the client to.
+        :keyword bool buffered_mode: Set to True to get the producer client work in buffered mode. Default is False.
```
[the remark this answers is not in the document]
yunhaoling: fixed

sdk/eventhub/azure-eventhub/samples/async_samples/send_buffered_mode_async.py
```diff
@@ -0,0 +1,60 @@
...
+            # the method returning indicates the event has been enqueued to the buffer
+            await producer.send_event(EventData('Single data {}'.format(i)))
+
+        batch = await producer.create_batch()
+        for i in range(10):
+            batch.add(EventData('Single data in batch {}'.format(i)))
+        # alternatively, you can enqueue an EventDataBatch object to the buffer
+        await producer.send_batch(batch)
+
+        # calling flush sends out the events in the buffered immediately
```
annatisch, line 55:
```suggestion
        # calling flush sends out the events in the buffer immediately
```

sdk/eventhub/azure-eventhub/samples/sync_samples/send_buffered_mode.py
```diff
@@ -0,0 +1,54 @@
...
+        # the method returning indicates the event has been enqueued to the buffer
+        producer.send_event(EventData('Single data {}'.format(i)))
+
+    batch = producer.create_batch()
+    for i in range(10):
+        batch.add(EventData('Single data in batch {}'.format(i)))
+    # alternatively, you can enqueue an EventDataBatch object to the buffer
+    producer.send_batch(batch)
+
+    # calling flush sends out the events in the buffered immediately
```
annatisch, line 51:
```suggestion
    # calling flush sends out the events in the buffer immediately
```

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_producer_client_async.py
```diff
@@ -106,20 +182,107 @@ def __init__(
...
+            await self._get_partitions()
+            await self._get_max_message_size()
+            self._buffered_producer_dispatcher = BufferedProducerDispatcher(
+                self._partition_ids,
+                self._on_success,
+                self._on_error,
+                self._create_producer,
+                self.eventhub_name,
+                self._max_message_size_on_link,
+                max_wait_time=cast(int, self._max_wait_time),
```
annatisch, line 230: Shouldn't this be float?

sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py
yunhaoling: updated

sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py
yunhaoling: fixed

sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py
yunhaoling: fixed

sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py
yunhaoling: added

sdk/eventhub/azure-eventhub/samples/async_samples/send_buffered_mode_async.py
yunhaoling: fixed

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_producer_client_async.py
yunhaoling: fixed

sdk/eventhub/azure-eventhub/samples/sync_samples/send_buffered_mode.py
yunhaoling: fixed

sdk/eventhub/azure-eventhub/azure/eventhub/_buffered_producer/_buffered_producer.py
```diff
@@ -0,0 +1,221 @@
...
+            except ValueError:
+                # add single event exceeds the cur batch size, create new batch
+                self._cur_batch = EventDataBatch(self._max_message_size_on_link)
+                self._buffered_queue.put(self._cur_batch)
+                self._cur_batch.add(events)
+            self._cur_buffered_len += new_events_len
+            # notify the max_wait_time worker
+            self._not_empty.notify()
+
+    def enhanced_callback_decorator(self, callback):
```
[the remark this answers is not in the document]
yunhaoling: naming suggestion taken :)

sdk/eventhub/azure-eventhub/azure/eventhub/_buffered_producer/_buffered_producer.py
yunhaoling: you're right! removed

sdk/eventhub/azure-eventhub/azure/eventhub/_producer.py
yunhaoling: the latter one, it fallbacks to `__len__`

sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py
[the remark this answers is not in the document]
yunhaoling: I've logged an issue: https://github.com/Azure/azure-sdk-for-python/issues/23844

sdk/eventhub/azure-eventhub/samples/sync_samples/send_buffered_mode.py
yunhaoling: I have added a sentence, but I have logged a separate issue for big chunk of sample updates:

https://github.com/Azure/azure-sdk-for-python/issues/23845

sdk/eventhub/azure-eventhub/samples/sync_samples/send_buffered_mode.py
yunhaoling: I have added the flush calling, but I would think about other complicate samples for flushing, also tracked in the issue: https://github.com/Azure/azure-sdk-for-python/issues/23845

sdk/eventhub/azure-eventhub/azure/eventhub/_buffered_producer/_buffered_producer.py
yunhaoling: `max_wait_time_worker ` will 
1. acquire `self._not_empty` lock/condition
2. calling `self._not_empty.wait()` will release the lock temporarily, waiting to be notified
3. it will call flush again which within flush it acquires the lock

so in stage 1/3, flush could no use, but in stage2 when `wait`, flush could use.

also underneath this condition is a re-entrant lock, so `max_wait_time_worker` acquire the lock and call flush will work.

sdk/eventhub/azure-eventhub/azure/eventhub/_buffered_producer/_buffered_producer.py
yunhaoling: this is originally used for the concurrent sending feature, but I'm removing it for now as we've decided to pull that out of the scope

sdk/eventhub/azure-eventhub/azure/eventhub/_buffered_producer/_buffered_producer_dispatcher.py
yunhaoling: updated!

sdk/eventhub/azure-eventhub/azure/eventhub/_buffered_producer/_partition_resolver.py
```diff
@@ -0,0 +1,198 @@
...
+
+class PartitionResolver:
+    def __init__(self, partitions):
+        self._idx = -1
+        self._partitions = partitions
+        self._partitions_cnt = len(self._partitions)
+        self._lock = Lock()
+
+    @property
+    def next_partition_id(self):
```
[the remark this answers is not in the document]
yunhaoling: updated method signature and added the async version

sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py
yunhaoling: I added one line of sentences to explain the behavioral differences

sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py
yunhaoling: comment added, thanks for the suggestion.

sdk/eventhub/azure-eventhub/azure/eventhub/_buffered_producer/_buffered_producer.py
yunhaoling: yes, that was originally I thought of to implement the feature. this will be addressed in the issue: https://github.com/Azure/azure-sdk-for-python/issues/23844

sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py
yunhaoling: it's referring to the timeout to enqueue the events to the queue. I'll add comment on this.

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_producer_client_async.py
```diff
@@ -38,6 +55,31 @@ class EventHubProducerClient(ClientBaseAsync):
...
+     It is required when buffered_mode is True while optional if buffered mode is False.
+    :paramtype on_success: Optional[Callable[[SendEventTypes, Optional[str]], Awaitable[None]]]
+    :keyword on_error: The callback to be called once a batch has failed to be published.
+     The callback takes three parameters:
+        - `events`: The list of events that failed to be published,
+        - `partition_id`: The partition id that the events in the list have been tried to be published to and
+        - `error`: The exception related to the sending failure.
+     The callback function should be defined like: `on_error(events, partition_id, error)`.
+     It is required when `buffered_mode` is True while optional if `buffered_mode` is False.
+     If `on_error is` passed when `buffered_mode`, instead of error being raised from the send methods,
```
swathipil, line 74:
```suggestion
     If `on_error is` passed when in buffered mode, instead of error being raised from the send methods,
```

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_producer_client_async.py
```diff
@@ -84,12 +126,46 @@ class EventHubProducerClient(ClientBaseAsync):
...
+        fully_qualified_namespace: str,
+        eventhub_name: str,
+        credential: "CredentialTypes",
+        *,
+        buffered_mode: bool = False,
+        on_error: Optional[Callable[[SendEventTypes, Optional[str], Exception], Awaitable[None]]] = None,
+        on_success: Optional[Callable[[SendEventTypes, Optional[str]], Awaitable[None]]] = None,
+        max_buffer_length: Optional[int] = None,
+        max_wait_time: Optional[float] = None,
+        **kwargs: Any
```
swathipil, line 168: wondering why typehints specified for the constructor implementation here but not in sync?

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_producer_client_async.py
```diff
@@ -198,8 +394,33 @@ def from_connection_string(
...
+         It is required when buffered_mode is True while optional if buffered mode is False.
+        :paramtype on_success: Optional[Callable[[SendEventTypes, Optional[str]], Awaitable[None]]]
+        :keyword on_error: The callback to be called once a batch has failed to be published.
+         The callback takes three parameters:
+            - `events`: The list of events that failed to be published,
+            - `partition_id`: The partition id that the events in the list have been tried to be published to and
+            - `error`: The exception related to the sending failure.
+         The callback function should be defined like: `on_error(events, partition_id, error)`.
+         It is required when `buffered_mode` is True while optional if `buffered_mode` is False.
+         If `on_error is` passed when `buffered_mode`, instead of error being raised from the send methods,
```
swathipil, line 413:
```suggestion
         If `on_error is` passed when in buffered mode, instead of error being raised from the send methods,
```

sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py
```diff
@@ -4,17 +4,29 @@
...
+    Union,
+    TYPE_CHECKING,
+    Dict,
+    List,
+    Optional,
+    Callable,
+    cast,
+    overload,
+)
+from typing_extensions import Literal
```
swathipil, line 18: I think Johan/Laurent mentioned that if we want to import from typing_extensions from now on, we need to include it in the required installs here: https://github.com/Azure/azure-sdk-for-python/blob/bc31f1674fd6e20d5c6eb00ab7c012896afd5ca8/sdk/eventhub/azure-eventhub/setup.py#L73

I believe they said that it's a dependency for azure-core right now, but might not be forever.

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_producer_client_async.py
```diff
@@ -181,11 +344,44 @@ def _create_producer(
...
+    def from_connection_string(
+        cls,
+        conn_str: str,
+        *,
+        eventhub_name: Optional[str] = None,
+        buffered_mode: bool = False,
+        on_error: Optional[Callable[[SendEventTypes, Optional[str], Exception], Awaitable[None]]] = None,
+        on_success: Optional[Callable[[SendEventTypes, Optional[str]], Awaitable[None]]] = None,
+        max_buffer_length: Optional[int] = None,
+        max_wait_time: Optional[float] = None,
```
swathipil, line 384: same question about why no typehints for sync implementation.

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_producer_client_async.py
```diff
@@ -257,22 +483,86 @@ def from_connection_string(
...
+            if self._on_error:
+                await self._on_error([event_data], partition_id, exc)
+            else:
+                raise
+
     async def send_batch(
         self,
         event_data_batch: Union[EventDataBatch, SendEventTypes],
-        *,
-        timeout: Optional[Union[int, float]] = None,
```
swathipil, line 264: just curious about why this was removed?

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_producer_client_async.py
yunhaoling: updated, thanks for pointing out!

sdk/eventhub/azure-eventhub/azure/eventhub/_buffered_producer/_buffered_producer.py
yunhaoling: In the previous implementation, I would enqueue empty batch into the queue.

however, you pushed me to think harder on this, so I've improved the impl such that empty batch would not be enqueued.

sdk/eventhub/azure-eventhub/azure/eventhub/_buffered_producer/_buffered_producer.py
yunhaoling: no it's not used, originally I was thinking of awaiting `check_max_wait_time_worker` to flush and to notify not full, however, I find it's easier to perform flush within `put_events` to get room as `check_max_wait_time_worker` might wait for a while before flushing.

yunhaoling: here's it's used as a lock, but I think in the future, we could introduce more sophisticated back pressure control strategy.

sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py
```diff
@@ -37,6 +50,31 @@ class EventHubProducerClient(ClientBase):
...
+     It is required when buffered_mode is True while optional if buffered mode is False.
+    :paramtype on_success: Optional[Callable[[SendEventTypes, Optional[str]], None]]
+    :keyword on_error: The callback to be called once a batch has failed to be published.
+     The callback takes three parameters:
+        - `events`: The list of events that failed to be published,
+        - `partition_id`: The partition id that the events in the list have been tried to be published to and
+        - `error`: The exception related to the sending failure.
+     The callback function should be defined like: `on_error(events, partition_id, error)`.
+     It is required when `buffered_mode` is True while optional if `buffered_mode` is False.
+     If `on_error is` passed when `buffered_mode`, instead of error being raised from the send methods,
```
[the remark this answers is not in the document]
yunhaoling: fixed

sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py
```diff
@@ -37,6 +50,31 @@ class EventHubProducerClient(ClientBase):
...
+     the `on_error` callback will be called with the error information related to sending.
+    :paramtype on_error: Optional[Callable[[SendEventTypes, Optional[str], Exception], None]]
+    :keyword int max_buffer_length: Buffered mode only.
+     The total number of events per partition that can be buffered before a flush will be triggered.
+     The default value is 1500 in buffered mode.
+    :keyword Optional[float] max_wait_time: Buffered mode only.
```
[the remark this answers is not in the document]
yunhaoling: yes, they're ignored

sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py
```diff
@@ -176,14 +334,72 @@ def _create_producer(self, partition_id=None, send_timeout=None):
...
+         It is required when buffered_mode is True while optional if buffered mode is False.
+        :paramtype on_success: Optional[Callable[[SendEventTypes, Optional[str]], None]]
+        :keyword on_error: The callback to be called once a batch has failed to be published.
+         The callback takes three parameters:
+            - `events`: The list of events that failed to be published,
+            - `partition_id`: The partition id that the events in the list have been tried to be published to and
+            - `error`: The exception related to the sending failure.
+         The callback function should be defined like: `on_error(events, partition_id, error)`.
+         It is required when `buffered_mode` is True while optional if `buffered_mode` is False.
+         If `on_error is` passed when `buffered_mode`, instead of error being raised from the send methods,
```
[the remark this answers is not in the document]
yunhaoling: fixed

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_producer_client_async.py
yunhaoling: updated, I forgot to update it as previously we are using comment style typing :P

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_producer_client_async.py
yunhaoling: updated

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_producer_client_async.py
yunhaoling: it is not removed completely -- it's being put into the kwargs
I remember Anna has talked about putting trivial key word argument into the kwargs bag such that the signature is cleaner.

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_producer_client_async.py
yunhaoling: fixed

sdk/eventhub/azure-eventhub/azure/eventhub/_buffered_producer/_buffered_producer.py
```diff
@@ -0,0 +1,215 @@
...
+            try:
+                self._check_max_wait_time_future.result(timeout)
+            except Exception as exc:  # pylint: disable=broad-except
+                _LOGGER.warning(
+                    "Partition %r stopped with error %r",
+                    self.partition_id,
+                    exc
+                )
+        if flush:
+            self.flush(timeout=timeout, raise_error=raise_error)
```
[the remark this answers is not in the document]
yunhaoling: I've re-ordered the two step and I realize that it's actually the right thing to flush first then wait for `check_max_wait_time_future` to finish.

otherwise the close turns out to be a double flush operation which is unnecessary.

sdk/eventhub/azure-eventhub/azure/eventhub/_buffered_producer/_buffered_producer.py
yunhaoling: I have logged an issue for this: https://github.com/Azure/azure-sdk-for-python/issues/23909

sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py
```diff
@@ -37,6 +50,31 @@ class EventHubProducerClient(ClientBase):
...
+        - `events`: The list of events that failed to be published,
+        - `partition_id`: The partition id that the events in the list have been tried to be published to and
+        - `error`: The exception related to the sending failure.
+     The callback function should be defined like: `on_error(events, partition_id, error)`.
+     It is required when `buffered_mode` is True while optional if `buffered_mode` is False.
+     If `on_error` is passed when in non-buffered mode, instead of error being raised from the send methods,
+     the `on_error` callback will be called with the error information related to sending.
+    :paramtype on_error: Optional[Callable[[SendEventTypes, Optional[str], Exception], None]]
+    :keyword int max_buffer_length: Buffered mode only.
+     The total number of events per partition that can be buffered before a flush will be triggered.
```
annatisch, line 73: What's interesting is that triggering this flush will be blocking....
I wonder if the background worker should be monitoring the buffer size and trigger a background flush when this size is hit - rather than waiting until the next `put_events` request to see that it's full and block until it's flushed?

What if I didn't set a `max_wait_time`, only a `max_buffer_length` - but then only add an event once per minute? We wouldn't check if the buffer was full until a minute after it had been full?

sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py
```diff
@@ -84,14 +122,47 @@ class EventHubProducerClient(ClientBase):
...
+        self,
+        fully_qualified_namespace: str,
+        eventhub_name: str,
+        credential: "CredentialTypes",
+        *,
+        buffered_mode: Literal[True],
+        on_error: Callable[[SendEventTypes, Optional[str], Exception], None],
+        on_success: Callable[[SendEventTypes, Optional[str]], None],
+        max_buffer_length: int = 1500,
+        max_wait_time: float = 1,
```
annatisch, line 148: These are documented as `Optional` - i.e. `Union[int, None]`. What would happen if I passed None in? Or is that invalid in a buffered scenario?

annatisch: hmm nevermind - I see the code below means that None would indeed be invalid.

sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py
annatisch: Looks like this can't happen as both `max_wait_time` and `max_buffer_length` must both have values - though depending on those values, we could end up waiting before we realize the buffer has been sitting full since the last event was added.

sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py
```diff
@@ -228,12 +454,85 @@ def from_connection_string(cls, conn_str, **kwargs):
...
+         `EventDataBatch` itself has partition_id.
+        :keyword str partition_key: With the given partition_key, event data will be sent to
+         a particular partition of the Event Hub decided by the service.
+         A `TypeError` will be raised if partition_key is specified and event_data_batch is an `EventDataBatch` because
+         `EventDataBatch` itself has partition_key.
+         If both partition_id and partition_key are provided, the partition_id will take precedence.
+         **WARNING: Setting partition_key of non-string value on the events to be sent is discouraged
+         as the partition_key will be ignored by the Event Hub service and events will be assigned
+         to all partitions using round-robin. Furthermore, there are SDKs for consuming events which expect
+         partition_key to only be string type, they might fail to parse the non-string value.**
```
annatisch, line 497: We should add an exception tag:
`:raises OperationTimeoutError: If the value specified by the timeout parameter elapses before the event can be sent and not running in buffered mode.`
Otherwise we can add this to the description for the timeout parameter itself.

annatisch: Same comment applies to `send_batch`

sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py
```diff
@@ -413,10 +704,29 @@ def get_partition_properties(self, partition_id):
...
 
-    def close(self):
-        # type: () -> None
+    def flush(self, **kwargs: Any) -> None:
+        """
+        Buffered mode only.
+        Flush events in the buffer to be sent immediately if the client is working in buffered mode.
+
+        :keyword Optional[float] timeout: Timeout to flush the buffered events, default is None which means no timeout.
+        :rtype: None
```
annatisch, line 713: We should also have a description of the error behaviour here, or add a `:raises:` tag.

sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py
```diff
@@ -105,21 +176,113 @@ def __init__(
...
         self._partition_ids = None  # Optional[List[str]]
         self._lock = threading.Lock()
+        self._buffered_mode = buffered_mode
+        self._on_success = on_success
+        self._on_error = on_error
+        self._buffered_producer_dispatcher = None
+        self._max_wait_time = max_wait_time
+        self._max_buffer_length = max_buffer_length
+        self._executor = kwargs.get("executor")
+        self._max_worker = kwargs.get("max_worker")
```
annatisch, line 186: We're not documenting these two for now?

sdk/eventhub/azure-eventhub/azure/eventhub/_buffered_producer/_buffered_producer_dispatcher.py
```diff
@@ -0,0 +1,152 @@
...
+        self._on_error = on_error
+        self._create_producer = create_producer
+        self._eventhub_name = eventhub_name
+        self._max_message_size_on_link = max_message_size_on_link
+        self._partition_resolver = PartitionResolver(self._partition_ids)
+        self._max_wait_time = max_wait_time
+        self._max_buffer_length = max_buffer_length
+        self._max_concurrent_sends = max_concurrent_sends
+        self._existing_executor = bool(executor)
+        self._executor = executor or ThreadPoolExecutor(max_worker)
```
annatisch, line 51: Wondering if we should use the number of partitions as the default worker count?

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_buffered_producer/_buffered_producer_dispatcher_async.py
```diff
@@ -0,0 +1,141 @@
...
+        for pid, future in futures:
+            try:
+                await future
+            except Exception as exc:  # pylint: disable=broad-except
+                exc_results[pid] = exc
+
+        if exc_results:
+            _LOGGER.warning('Stopping all partitions failed with result %r.', exc_results)
+            if raise_error:
+                raise EventDataSendError(
```
annatisch, line 128: Should this be an EventDataSendError?

sdk/eventhub/azure-eventhub/samples/async_samples/send_buffered_mode_async.py
```diff
@@ -0,0 +1,60 @@
...
+            await producer.send_event(EventData('Single data {}'.format(i)))
+
+        batch = await producer.create_batch()
+        for i in range(10):
+            batch.add(EventData('Single data in batch {}'.format(i)))
+        # alternatively, you can enqueue an EventDataBatch object to the buffer
+        await producer.send_batch(batch)
+
+        # calling flush sends out the events in the buffer immediately
+        await producer.flush()
```
annatisch, line 56: Remind me - does `flush` happen automatically when the context that's opened on line 43 closes?
If so - we should add that to the comment.

annatisch (approved):

sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py
yunhaoling: yeah, this is part of the issue: https://github.com/Azure/azure-sdk-for-python/issues/23909 how we deal with backpressure.

we can indeed have a background worker monitoring the buffer size, but it becomes a problem of the threshold of capacity we want to flush.

I feel ultimately, we need do it in a hybrid way -- we have the background worker, we also do flush if background worker doesn't clean up the queue for us.

I have put the notes in the issue.

sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py
yunhaoling: if max_wait_time is passed as None, then it defaults to 1 in the buffered mode.

sdk/eventhub/azure-eventhub/azure/eventhub/_buffered_producer/_buffered_producer.py
```diff
@@ -0,0 +1,216 @@
...
+                    self.partition_id,
+                    self._cur_buffered_len
+                )
+        if self._check_max_wait_time_future:
+            try:
+                with self._not_empty:
+                    # in the stop procedure, calling notify to give check_max_wait_time_future a chance to stop
+                    # as it is waiting for Condition self._not_empty
+                    self._not_empty.notify()
+                self._check_max_wait_time_future.result(timeout)
```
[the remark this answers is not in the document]
yunhaoling: yes, I think you're right, I have updated the implementation

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_buffered_producer/_buffered_producer_dispatcher_async.py
yunhaoling: I changed this to `EventHubError`.

I've logged a separate issue to track the error experience.

sdk/eventhub/azure-eventhub/azure/eventhub/_buffered_producer/_buffered_producer_dispatcher.py
yunhaoling: that's a good point, we can but I'm thinking of the cases that partition count is kinda large e.g. 32/64 -- not sure how the perf would go when there're excessive amount of threads..

logged an issue for it

sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py
yunhaoling: updated!

sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py
yunhaoling: updated!

sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py
yunhaoling: yeah, I decided to pull it out of our public api surface.

I had left comment in the code as well as opened separate issues for tracking.
we can expose them in the futuer.

sdk/eventhub/azure-eventhub/azure/eventhub/_buffered_producer/_buffered_producer_dispatcher.py
yunhaoling: https://github.com/Azure/azure-sdk-for-python/issues/24013

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_buffered_producer/_buffered_producer_dispatcher_async.py
yunhaoling: https://github.com/Azure/azure-sdk-for-python/issues/24011

sdk/eventhub/azure-eventhub/samples/async_samples/send_buffered_mode_async.py
yunhaoling: comment added

sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py
swathipil: we? 👀

PR #23764, opened 2022-03-31
Update testing env variables

mshaban-msft:
# Description

### DIFF
--- a/sdk/cognitivelanguage/tests.yml
+++ b/sdk/cognitivelanguage/tests.yml
@@ -12,12 +12,12 @@ stages:
         AZURE_QUESTIONANSWERING_KEY: $(qna-key)
         AZURE_QUESTIONANSWERING_PROJECT: 190a9e13-8ede-4e4b-a8fd-c4d7f2aeab6c
         AZURE_QUESTIONANSWERING_ENDPOINT: $(qna-uri)
-        AZURE_CONVERSATIONS_ENDPOINT: $(conversations-endpoint) # clu endpoint
-        AZURE_CONVERSATIONS_KEY: $(conversations-key) # clu key
-        AZURE_CONVERSATIONS_PROJECT_NAME: $(conversations-project) # conversation project - name
-        AZURE_CONVERSATIONS_DEPLOYMENT_NAME: $(conversations-conversations-project-deployment) # conversation project - deployment name
-        AZURE_CONVERSATIONS_WORKFLOW_PROJECT_NAME: $(conversations-workflow-project) # conversation project - deployment name
-        AZURE_CONVERSATIONS_WORKFLOW_DEPLOYMENT_NAME: $(conversations-workflow-project-deployment) # conversation project - deployment name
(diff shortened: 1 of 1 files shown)


sdk/cognitivelanguage/tests.yml
```diff
@@ -12,10 +12,12 @@ stages:
...
-        AZURE_CONVERSATIONS_ENDPOINT: $(conversations-endpoint)
-        AZURE_CONVERSATIONS_KEY: $(conversations-key)
-        AZURE_CONVERSATIONS_PROJECT: $(conversations-project)
-        AZURE_CONVERSATIONS_WORKFLOW_PROJECT: $(conversations-workflow-project)
+        AZURE_CONVERSATIONS_ENDPOINT: $(conversations-endpoint) # clu endpoint
+        AZURE_CONVERSATIONS_KEY: $(conversations-key) # clu key
+        AZURE_CONVERSATIONS_PROJECT_NAME: $(conversations-project) # conversation project - name
+        AZURE_CONVERSATIONS_DEPLOYMENT_NAME: $(conversations-conversations-project-deployment) # conversation project - deployment name
+        AZURE_CONVERSATIONS_WORKFLOW_PROJECT_NAME: $(conversations-workflow-project) # conversation project - deployment name
+        AZURE_CONVERSATIONS_WORKFLOW_DEPLOYMENT_NAME: $(conversations-workflow-project-deployment) # conversation project - deployment name
```
heaths, line 20: Your pipeline should be using the "Test Secrets for Cognitive Language Services Live Tests" variable group.I created this for use across languages. The secrets I currently have stored are as follows, which don't match what you have. It's best we don't store these in multiple places:

* qna-uri
* qna-key
* clu-uri
* clu-key

I'll update / add the secrets you sent me offline, but will use "clu" for consistency over "conversations". Please update your pipeline to match what we have in Azure/azure-sdk-for-net for sdk/cognitivelanguage/tests.yml.

Ideally, we should also align on names. I could change ours to `CONVERSATIONS_PROJECT_NAME` rather than just "PROJECT". To note: our test infra automatically checks for those variables names then prepends "AZURE_" if needed, so what you have here is fine if that fits your test infra better.

heaths: I'm also not convinced all these are secrets, but can still add them. It's generally bad practice to put not-secrets in Key Vault as it can unnecessary waste time and, in some scenarios, throttle the app service.

mshaban-msft: i totally agree!
i think `endpoint` and `subscription_key` are the only secrets, and `project_name` and `deployment_name` can be just variables.
i'll update tests and samples and variables to reflect this.

heaths: See .NET's variable names. The environment variables might align except for the one that doesn't start with "AZURE_", but @ZulaMostafa or I will fix that to be consistent. Ideally we'll get to a point we can just share most of our test resource files if/when one language makes a change like we do for Key Vault and some other services.

heaths (approved):

PR #23776, opened 2022-03-31
[pylint] ListReturnType checker

l0lawrence:
Still a draft - but tried to check return type, against the docstring rtype and take into account custom paging classes that have by_page.


scripts/pylint_custom_plugin/pylint_guidelines_checker.py
```diff
@@ -890,10 +891,14 @@ def visit_functiondef(self, node):
                     try:
                         # infer_call_result gives the method return value as a string
                         returns = next(node.infer_call_result()).as_string()
-                        if returns.find("ItemPaged") == -1 and returns.find("AsyncItemPaged") == -1:
-                            self.add_message(
-                                msgid="client-list-methods-use-paging", node=node, confidence=None
-                            )
+  
+                        # Also check the docstring return type and make sure rtype is there
+                        if node.doc.split("rtype:")[-1] not in returns:
```
kristapratico, line 896: Not 100% on this,  but won't `node.doc.split("rtype:")[-1]` just return everything in the docstring after `rtype`? Which might include sample snippet admonitions and versionadded directives?

scripts/pylint_custom_plugin/pylint_guidelines_checker.py
```diff
@@ -890,10 +891,14 @@ def visit_functiondef(self, node):
                     try:
                         # infer_call_result gives the method return value as a string
                         returns = next(node.infer_call_result()).as_string()
```
kristapratico, line 893: just a small note - I remember this not being a super reliable way to get the return type for client method. If there is a better way that you've found, we should reconsider :)

scripts/pylint_custom_plugin/pylint_guidelines_checker.py
```diff
@@ -1947,7 +1952,7 @@ def register(linter):
     # linter.register_checker(ClientHasApprovedMethodNamePrefix(linter))
     # linter.register_checker(ClientMethodsHaveTracingDecorators(linter))
     # linter.register_checker(ClientDocstringUsesLiteralIncludeForCodeExample(linter))
-    # linter.register_checker(ClientListMethodsUseCorePaging(linter))
+    linter.register_checker(ClientListMethodsUseCorePaging(linter))
```
kristapratico, line 1955: nit: let's move this up near the registered checkers

scripts/pylint_custom_plugin/tests/test_pylint_custom_plugins.py
```diff
@@ -1860,6 +1860,50 @@ def test_guidelines_link_active(self):
 class TestClientListMethodsUseCorePaging(pylint.testutils.CheckerTestCase):
     CHECKER_CLASS = checker.ClientListMethodsUseCorePaging
 
+    def test_list_return_type_acceptable(self):
+        class_node, function_node = astroid.extract_node("""
+        from azure.core.paging import ItemPaged
+
+        class SomeClient(): #@
+            def _list_thing(self): #@
```
kristapratico, line 1868: should this have the underscore prefix? I think the linter rule will ignore if it does, right?

scripts/pylint_custom_plugin/tests/test_pylint_custom_plugins.py
```diff
@@ -1886,9 +1930,15 @@ def test_ignores_methods_return_ItemPaged(self):
         
         class SomeClient(): #@
             def list_thing(self): #@
+                '''
+                :rtype: ItemPaged()
```
kristapratico, line 1934: I think more commonly we'd see `~azure.core.paging.ItemPaged` and `~azure.core.async_paging.AsyncItemPaged`. We also probably want to check for when the result type is included, like `~azure.core.paging.ItemPaged[ModelOperationInfo]`

scripts/pylint_custom_plugin/pylint_guidelines_checker.py
l0lawrence: I added in a check after calling infer_call_result to see if it was inferred or not and if it isn't inferred, then I look for the astroid.Return in the body of the function node and grab the name of what is being returned

scripts/pylint_custom_plugin/tests/test_pylint_custom_plugins.py
l0lawrence: Right now the rule just checks if the function is a method, node.is_method(). I will make it ignore private functions

PR #23777, opened 2022-03-31
[Storage] Bump azure-core mindep to latest version

jalauzon-msft:
Version `1.23.0` of `azure-core` contains a breaking change for Storage with the latest auto-generated code which was fixed in `1.23.1`. Bumping the mindep up to latest version to ensure we have the fix.


vincenttran-msft (approved):
LGTM!

PR #23778, opened 2022-03-31
[Storage] Remove latest service version from blob as it's not released yet

jalauzon-msft:
We had previously added support for the latest service version in `main` but it's not released yet so Live tests were failing. Removing this from `main` and will leave in STG82 feature branch.


vincenttran-msft (approved):
LGTM!

PR #23781, opened 2022-03-31
[formrecognizer] Fix failing pipeline tests

catalinaperalta:
These tests were failing in the live test runs the last couple of days. 

Added a new sub-folder in trainingdata-v3 in the container and reference that in the DMAC tests. Simply updating the recordings for the signature tests (the error doesn't appear when running locally)


sdk/formrecognizer/azure-ai-formrecognizer/tests/test_dac_analyze_custom_model.py
```diff
@@ -209,10 +209,6 @@ def test_pages_kwarg_specified(self, client, formrecognizer_storage_container_sa
     @DocumentModelAdministrationClientPreparer()
     @recorded_by_proxy
     def test_custom_document_signature_field(self, client, formrecognizer_storage_container_sas_url, **kwargs):
-        # this can be reverted to set_bodiless_matcher() after tests are re-recorded and don't contain these headers
-        set_custom_default_matcher(
-            compare_bodies=False, excluded_headers="Authorization,Content-Length,x-ms-client-request-id,x-ms-request-id"
-        )
```
kristapratico, line 215: just checking - the comment says it can be reverted to set_bodiless_matcher(). Do we need to add that when removing the custom matcher?

catalinaperalta: The add matcher endpoint on the test proxy seems to be failing at the moment :( no matter what matcher I set. I've reached out to McCoy to see what we can do about it. I'm waiting to see how the pipeline runs go without adding that back in.

catalinaperalta: Ok yes, set_bodiless_matcher() is still required for regular CI :( I'll have to wait for a fix in the test proxy then to update that. I'll add those back in and run the tests again

mccoyp: To provide context here for what I'm talking about with Cat offline:

The issue with the matcher setup is that the test proxy is failing to process multiple matcher requests in quick succession. As part of https://github.com/Azure/azure-sdk-for-python/pull/23148, I had to add a `set_custom_default_matcher` call to `proxy_startup.py`, which is run during test session startup, in order to make tests backwards-compatible after removing some headers from recordings. What's happening here is that these tests are making another request to set a matcher right after `proxy_startup.py` is run, and the test proxy can't handle that request yet.

The solution will require either waiting to make a matcher set request until the proxy is ready to handle it, or updating old recordings so that we don't need the test proxy tooling to be backwards-compatible and set this initial matcher. I'm taking a look at potential solutions now and will give an update when I have one

catalinaperalta: Thanks for looking into it McCoy!

kristapratico (approved):

sdk/formrecognizer/azure-ai-formrecognizer/tests/test_dac_analyze_custom_model.py
mccoyp: Providing the update here for record-keeping and as punishment for myself for being so confident in my evaluation of the issue:

I was wrong! It looks like the issue with the matcher request wasn't actually making requests in quick succession, but rather making matcher requests in live mode with different recording ID scopes. In short, the matcher set request from `proxy_startup.py` applies to all tests, but the bodiless matcher requests made in these tests are specific to the test they're called from. The test proxy seems to be upset about that for reasons I don't yet fully understand.

The good news is that we were doing the wrong thing anyway, in a very-easy-to-fix way: we were sending matcher requests in recording mode when we don't actually use matchers in that case, regardless. This wasn't an issue in live test pipelines because recording is turned off in that environment, but when tests are being recorded locally we would hit the error mentioned above. Since we don't actually need these requests anyway, we can fix the issue by only sending matcher requests in playback mode (which is the only time matchers have any effect). And in playback mode, the whole recording ID scoping issue doesn't crop up.

Cat made the appropriate fix [here](https://github.com/Azure/azure-sdk-for-python/pull/23781/commits/6865341b0c076446fd73982cd4c3b091126014e5), so we should now be good to go 🙂

mccoyp (approved):
Thank you for bringing this to my attention and fixing the matcher bug!

catalinaperalta:
JFYI that the signature tests are still failing in the pipeline although they pass locally (not related to test proxy issues). Will merge this PR and add a new one with the resolution for those, once I find the issue that's affecting the pipeline tests.

Issue #23543, opened 2022-03-16
[Storage] Fix sending empty headers to service on certain operations


jalauzon-msft:
In the following operations, the SDK is sending an empty header to the service. While this is not technically a problem, we should not be sending these empty headers if there is no information.
- file-share
    - `start_copy_from_url()` - We send empty `x-ms-file-create-time` and `x-ms-file-last-write-time` if they are not provided
- file-datalake
    - `_create_path()` (`create_directory()`, `create_sub_directory()`, `create_file()`) - We send empty `x-ms-properties` if none are provided.

There could be others.

PR #23796, opened 2022-04-02
[Storage] Fix Empty Headers in File Share and Data Lake

vincenttran-msft:
As raised in #23543, this PR aims to fix the empty headers being sent on certain requests in both `file-share` and `file-datalake`. Some tests were re-recorded to verify non-breaking.

Follows-ups:

1. Change the name of `add_metadata_headers()` to better represent what the method is doing
2. in .NET, the closest thing we have to `properties` in Python's SDK library is `httpHeaders`, so possibly rename to allow for more seamless transitioning between .NET and Python workflows
3. While empty headers issue was eliminated, `_datetime_to_str()` should be refactored once new library is introduced

### DIFF
--- a/sdk/storage/azure-storage-file-share/tests/recordings/test_file.test_copy_file_with_specifying_acl_copy_behavior_attributes.yaml
+++ b/sdk/storage/azure-storage-file-share/tests/recordings/test_file.test_copy_file_with_specifying_acl_copy_behavior_attributes.yaml
@@ -3,7 +3,7 @@ interactions:
     body: null
     headers:
       Accept:
-      - '*/*'
+      - application/xml
       Accept-Encoding:
       - gzip, deflate
       Connection:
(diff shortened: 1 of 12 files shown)


sdk/storage/azure-storage-file-share/tests/recordings/test_file.test_copy_file_with_specifying_acl_copy_behavior_attributes.yaml
```diff
@@ -137,51 +137,49 @@ interactions:
...
       x-ms-copy-source:
-      - https://emilystageaccount.file.core.windows.net/utsharedd3c1c70/filedd3c1c70
+      - https://vincenttrancanary.file.core.windows.net/utsharedd3c1c70/filedd3c1c70
       x-ms-date:
-      - Tue, 28 Jan 2020 05:13:22 GMT
+      - Fri, 01 Apr 2022 02:17:43 GMT
       x-ms-file-attributes:
       - Temporary|NoScrubData
-      x-ms-file-copy-ignore-read-only:
+      x-ms-file-copy-ignore-readonly:
```
annatisch, line 173: This is curious - did the header name change between API versions?
Or did something change in the swagger-generated layer?

PR #23806, opened 2022-04-04
Import OS in pylint rcfile.

tjprescott:
See: https://github.com/Azure/azure-sdk-tools/pull/2933

In order for the above PR to pass, the pylint rcfile needs the `import os` statement. The file in the SDK tools repo and the SDK repo _must_ match or the CI will fail. Thus, updating this repo's copy to match.


scbedd (approved):
Given that this code actually has `os.` calls within it, I'm legitimately blown away that this worked before this. Thanks @tjprescott

tjprescott:
@scbedd I was surprised too 😆

Issue #23169, opened 2022-02-22
[Cosmos] move `is_system_key` logic to be computed only once on container usage


simorenoh:
A user reported the fact that the async client couldn't properly deal with single partition queries since the result of the call to `is_system_key` was not being awaited properly. Due to this, we have a mitigation, but it is not ideal: https://github.com/Azure/azure-sdk-for-python/pull/22893

We are currently awaiting the result of this network call once for every query being executed with a partition key, adding overhead for each query when it is only needed once in order to populate the container settings and then be used from there. Have tried a couple different things but haven't gotten it to work yet.

Expected outcome would be to run this network call only once as stated before - need to find a way to do so.

User issue in question: https://github.com/Azure/azure-sdk-for-python/issues/22861

PR #23821, opened 2022-04-05
[Cosmos] Improve system key computation logic

simorenoh:
Follow up to https://github.com/Azure/azure-sdk-for-python/pull/22893
Previously, we were using this call to set the partition key value for a query request if it was passed in:
image

This awaitable being passed would then be resolved by the `_query_iterable_async` class once for every query being executed with a partition key, which is not ideal.

This small change in the function's return value makes it so we only await/do the network call if the user passed in a `NonePartitionKeyValue`, effectively solving the issue for all other queries and 99% of user cases.

Fixes https://github.com/Azure/azure-sdk-for-python/issues/23169


xinlian12 (approved):
LGTM, thanks

kushagraThapar (approved):

PR #23853, opened 2022-04-06
[Storage] Address feedback from API View review for STG82 preview

jalauzon-msft:
Addressing feedback from STG82 API View review.


vincenttran-msft (approved):
LGTM! Great work on addressing the feedback :)

Issue #23804, opened 2022-04-04
DataLake async file client cannot download using SAS tokens restricted to specific folders


simebg:
- **Package Name**: azure-storage-file-datalake
- **Package Version**: 12.6.0
- **Operating System**: Ubuntu 20.04
- **Python Version**: 3.9.7

**Describe the bug**
SAS tokens for specific folders in a storage account don't work in the async client, however they do in the sync one.

SAS tokens generated on the whole container do work on both async and sync clients.

**To Reproduce**
Replace relevant variables in the following code, then execute it:

```
path = '...'
container = '...'
account_url = '...'

# sas token on specific folder (works on sync, doesn't work on async)
sas_token = '...'
# sas token on container (works on both sync and async)
# sas_token = '...'

def sync_run():
    print('Trying synchronous API...')

    from azure.storage.filedatalake import DataLakeFileClient

    with DataLakeFileClient(account_url, container, path, sas_token) as client:
        data = client.download_file().readall()
        print(f'Read {len(data)} bytes.')

def async_run():
    async def run():
        print('Trying asynchronous API...')

        from azure.storage.filedatalake.aio import DataLakeFileClient

        async with DataLakeFileClient(account_url, container, path, sas_token) as client:
            data = await (await client.download_file()).readall()
            print(f'Read {len(data)} bytes.')

    import asyncio
    asyncio.get_event_loop().run_until_complete(run())


if __name__ == '__main__':
    sync_run()
    async_run()
```

**Expected behavior**
Output should be something similar to:
```
$ python3 test.py 
Trying synchronous API...
Read 3060 bytes.
Trying asynchronous API...
Read 3060 bytes.
```

However, I get the following:
```
$ python3 test.py 
Trying synchronous API...
Read 3060 bytes.
Trying asynchronous API...
Traceback (most recent call last):
  File "/home/azure-async-test/test.py", line 64, in <module>
    async_run()
  File "/home/azure-async-test/test.py", line 59, in async_run
    asyncio.get_event_loop().run_until_complete(run())
  File "/usr/lib/python3.9/asyncio/base_events.py", line 642, in run_until_complete
    return future.result()
  File "/home/azure-async-test/test.py", line 55, in run
    data = await (await client.download_file()).readall()
  File "/home/azure-async-test/.venv/lib/python3.9/site-packages/azure/storage/filedatalake/aio/_data_lake_file_client_async.py", line 481, in download_file
[...]

PR #23854, opened 2022-04-06
[Storage] Fixed parsing of path_name and file_system_name in async clients

vincenttran-msft:
A customer raised issue #23804 in where they were having inconsistent behavior with `DataLakeFileClient` where the operation is was working on `sync` clients but not on `async` clients. 

Upon further investigation, we found the source of their issue being how the `sync` and `async` client handle the parsing of `path_name` and `file_system_name`, and this PR aims to make `async` client behavior consistent with `sync` client behavior (and thus, working).


sdk/storage/azure-storage-file-datalake/tests/test_file.py
```diff
@@ -98,6 +98,22 @@ def test_create_file(self, datalake_storage_account_name, datalake_storage_accou
         # Assert
         self.assertIsNotNone(response)
 
+    @pytest.mark.live_test_only
```
jalauzon-msft, line 101: Why Live only here? You also generated recordings which would not be needed if this was Live only, but I think it can be recorded too? This applies for all tests.

vincenttran-msft: Good point- something tells me the reference test I was modeling from might have had that because of some SAS permissions that require live-testing only. Fixed!

jalauzon-msft (approved):

PR #23856, opened 2022-04-06
[ServiceBus] fix typehints

swathipil:
based on APIView here: https://apiview.dev/Assemblies/Review/fb319de65d2b4eca86b215c62d0cbe7b?revisionId=2c875b1de85e4ec58f0c1b2994cee20d&diffRevisionId=d2bf9a264ba84047a667011c9ae2d41d&doc=False&diffOnly=False#comment-thread-5

### DIFF
--- a/sdk/servicebus/azure-servicebus/azure/servicebus/_common/auto_lock_renewer.py
+++ b/sdk/servicebus/azure-servicebus/azure/servicebus/_common/auto_lock_renewer.py
@@ -258,8 +258,8 @@ class AutoLockRenewer(object):  # pylint:disable=too-many-instance-attributes
         :type receiver: ~azure.servicebus.ServiceBusReceiver
         :param renewable: A locked entity that needs to be renewed.
         :type renewable: Union[~azure.servicebus.ServiceBusReceivedMessage, ~azure.servicebus.ServiceBusSession]
-        :param max_lock_renewal_duration: A time in seconds that the lock should be maintained for. If set to None,
-         `max_lock_renewal_duration` value passed to constructor will be used. Default value is None.
+        :param max_lock_renewal_duration: A time in seconds that the lock should be maintained for.
+         Default value is None. If specified, this value will override the default value specified at the constructor.
         :type max_lock_renewal_duration: Optional[float]
         :param on_lock_renew_failure: A callback may be specified to be called when the lock is lost on the renewable
          that is being registered. Default value is None (no callback).
--- a/sdk/servicebus/azure-servicebus/azure/servicebus/_common/auto_lock_renewer.py
+++ b/sdk/servicebus/azure-servicebus/azure/servicebus/_common/auto_lock_renewer.py
@@ -258,8 +258,8 @@ class AutoLockRenewer(object):  # pylint:disable=too-many-instance-attributes
         :type receiver: ~azure.servicebus.ServiceBusReceiver
         :param renewable: A locked entity that needs to be renewed.
         :type renewable: Union[~azure.servicebus.ServiceBusReceivedMessage, ~azure.servicebus.ServiceBusSession]
-        :param max_lock_renewal_duration: A time in seconds that the lock should be maintained for.
-          Default value is 300 (5 minutes).
+        :param max_lock_renewal_duration: A time in seconds that the lock should be maintained for. If set to None,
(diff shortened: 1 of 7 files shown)


sdk/servicebus/azure-servicebus/azure/servicebus/_common/auto_lock_renewer.py
```diff
@@ -258,8 +258,8 @@ def register(
         :type receiver: ~azure.servicebus.ServiceBusReceiver
         :param renewable: A locked entity that needs to be renewed.
         :type renewable: Union[~azure.servicebus.ServiceBusReceivedMessage, ~azure.servicebus.ServiceBusSession]
-        :param max_lock_renewal_duration: A time in seconds that the lock should be maintained for.
-          Default value is 300 (5 minutes).
+        :param max_lock_renewal_duration: A time in seconds that the lock should be maintained for. If set to None,
+         `max_lock_renewal_duration` value passed to constructor will be used. Default value is None.
```
swathipil, line 262: I updated the default value in the docstring, rather than the actual default value from None to 300. I think doing so might be a breaking change for the following scenario:
For example, if user passes in `max_lock_renewal_duration=500` and doesn't pass in anything in `register` here (so `max_lock_renewal_duration=None`), then `500` would be passed in on line 312. However, changing the default to 300 here means that the value passed in to line 312 will be 300, if the customer doesn't touch their code.

annatisch: You could probably just state something like:
```python
If specified, this value will override the default value specified at the constructor.
```

yunhaoling: yep, I agree that we should only change the docstring and your description about the behavior is precise :)

annatisch: We don't need to update the docs for the async auto-renewer?

yunhaoling (approved):
LGTM, thanks!

sdk/servicebus/azure-servicebus/azure/servicebus/_common/message.py
```diff
@@ -70,7 +70,7 @@ class ServiceBusMessage(
     :param body: The data to send in a single message.
     :type body: Optional[Union[str, bytes]]
 
-    :keyword Optional[Dict] application_properties: The user defined properties on the message.
+    :keyword Optional[Dict[str, Any]] application_properties: The user defined properties on the message.
```
annatisch, line 73: User property values can by any primitive type? Is there maybe something more specific than `Any` we can use here?

sdk/servicebus/azure-servicebus/azure/servicebus/_common/auto_lock_renewer.py
yunhaoling: out of curiosity: is it that the api generator would collect information (default value 300) from docstring for api view?

swathipil: I thought there was only a shared auto renewer (since it's in the _common folder), but I might be mistaken?

sdk/servicebus/azure-servicebus/azure/servicebus/_common/message.py
swathipil: I checked in the AmqpAnnotatedMessage here and in EH EventData as well, and it looks like we use `Dict[str, Any]`, `Dict[Union[str, bytes], Any]`, or just `Dict` everywhere.

@yunhaoling - do you know if we can restrict the type here at all?

sdk/servicebus/azure-servicebus/azure/servicebus/_common/auto_lock_renewer.py
swathipil: Travis confirmed that default can be parsed from docstring if not specified in the signature.

sdk/servicebus/azure-servicebus/azure/servicebus/_common/message.py
swathipil: checked the amqp spec: `the values are restricted to be of simple types only, that is (excluding map, list, and array
types)`.

PR #23857, opened 2022-04-06
[Storage] Revert addition of 'create_if_not_exists()' for all Python clients

vincenttran-msft:
As raised by a customer in #21044 and eventually implemented and merged in #23574, this PR aims to revert the addition of `create_if_not_exists()` for all Python clients, as discussed and decided on by the architecture team and others.


sdk/storage/azure-storage-blob/CHANGELOG.md
```diff
@@ -2,9 +2,6 @@
 
 ## 12.12.0b1 (Unreleased)
 
-### Features Added
```
jalauzon-msft, line 5: Nit (but also I think ci might fail for this?): Can you leave the `Features Added` section and just delete this line? This will help with merging into feature branch as well as I think ci may fail if this isn't present?

Applies for other CHANGELOGS as well (expect Queue)

jalauzon-msft (approved):
One minor comment for CHANGELOGs, otherwise looks good!

Sorry again this got pulled :(

sdk/storage/azure-storage-blob/CHANGELOG.md
vincenttran-msft: Can do! I didn't know the procedure of whether to leave it or not, so I guessed, and as it turns out, the coinflip wasn't in my favor :)

PR #23890, opened 2022-04-07
[pylint] checking that list_ method return a Paging class

l0lawrence:
Just wanted some review for the changing of inferring return type to inferring directly on the return node if it is a call to a function.

I have a TODO to flag methods that return ItemPaged, but are not named with list (i.e Search). 

No rush on this - not to be merged until after repo shift


scripts/pylint_custom_plugin/pylint_guidelines_checker.py
```diff
@@ -886,17 +887,30 @@ def visit_functiondef(self, node):
...
-                        logger.debug("Pylint custom checker failed to check if client list method uses core paging.")
-                        pass
+                    paging_class = False
+
+                    for inner_node in node.body:
+                        # If this is a return node
+                        if isinstance(inner_node, astroid.Return):
+                            try:
+                                # Checking that what we are returning is a function call
+                                if isinstance(inner_node.value, astroid.Call):
```
kristapratico, line 902: does this work for either an ItemPaged or a function call which returns an ItemPaged? If the return type is very incorrect, like returning `None` or something, will that get caught?

scripts/pylint_custom_plugin/tests/test_files/list_return_type_custom_class_acceptable.py
```diff
@@ -0,0 +1,34 @@
...
+        query = None
+        return SearchItemPaged(
+            self._client, query, kwargs, page_iterator_class=None
+        )
+    
+    def list_this(self):
+        return azure.core.paging.ItemPaged()
+    
+    def list_something(self):
+        return [1,2,3,4,5]
```
kristapratico, line 34: files says acceptable, but shouldn't the linter error on this?

scripts/pylint_custom_plugin/tests/test_pylint_custom_plugins.py
```diff
@@ -1886,27 +1921,42 @@ def test_ignores_methods_return_ItemPaged(self):
...
             async def list_thing(self): #@
+                '''
+                :rtype: AsyncItemPaged()
+                '''
                 return AsyncItemPaged()
+
             @distributed_trace
             def list_thing2(self): #@
+                '''
+                :rtype: AsyncItemPaged()
```
kristapratico, line 1958: maybe do one with the full path? ~azure.core.async_paging.AsyncItemPaged

scripts/pylint_custom_plugin/tests/test_files/list_return_type_custom_class_acceptable.py
l0lawrence: I updated it to a file of both errors and non-errors, I will update the title!

scripts/pylint_custom_plugin/pylint_guidelines_checker.py
l0lawrence: Good point, I will test with those!

PR #23896, opened 2022-04-07
Pylint guidelines checker move - stage 1

scbedd:
The eventual goal is to have APIView and python repo share the same linter. As part of that, we need to....

- [x] [Copy pylint plugin code to azure-sdk-tools and enable CI ](https://github.com/Azure/azure-sdk-tools/pull/3078)
- [x] Consume from tool for local linting (this PR)
- [ ] Remove code from azure-sdk-for-python (upcoming)

@l0lawrence FYI as impacting your feature branch


pylintrc
```diff
@@ -5,7 +5,6 @@ reports=no
 # PYLINT DIRECTORY BLACKLIST.
 ignore=_vendor,_generated,samples,examples,test,tests,doc,.tox
 
-init-hook='import sys; import os; sys.path.insert(0, os.path.abspath(os.getcwd().rsplit("azure-sdk-for-python", 1)[0] + "azure-sdk-for-python/scripts/pylint_custom_plugin"))'
```
scbedd, line 8: no longer necessary as it will be installed.

l0lawrence (approved):

PR #23897, opened 2022-04-08
Update tox so all ci steps can use local source if package reference not yet released

jalauzon-msft:
Thanks to @scbedd making these changes in a Storage feature branch to unblock our ci. Just bringing them to `main` to separate out from Storage feature branch.

These changes allow all ci steps to pull a package from local source if it's not yet released (and therefore not available via pip). The was needed for Storage since we have one package which directly depends on another, and we often need to update them together.


scbedd:
@jalauzon-msft I realized that with this change, we may have the risk of releasing a package with a dependency that doesn't yet exist on PyPI. Giving some thought to how I want to prevent that, and will update this PR with additional commits. Thanks for backporting!

scbedd:
I decided

> I'm already in this code, will add docstrings/type only on functions being used for this new feature.

facepalm

[But Publishing Still Works](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=1501189&view=logs&s=6884a131-87da-5381-61f3-d7acc3b91d76&j=2e8bded0-b06c-5a9b-286e-efb0e0fd18fa)

Am working through final analyze failure, I should have it with this last commit.

PR #23933, opened 2022-04-11
[Cosmos] use python3 typehints and move options to kwargs on async client

simorenoh:
With these changes we tackle the python-technical aspects of the async client. 

This PR translates the current type hints to their python3 versions for the public surface area of the async client and moves many of the options for several methods to be part of the kwargs instead. This PR effectively knocks out two of the bullet points in the list of to-do items for async GA.

Non-public surface area type hints and comments still missing - if this is also needed, I can do it as part of this PR or a different one in order to not make this one excessively large.

https://mypy.readthedocs.io/en/stable/cheat_sheet_py3.html


sdk/cosmos/azure-cosmos/azure/cosmos/aio/cosmos_client.py
```diff
@@ -88,15 +89,18 @@ def _build_connection_policy(kwargs):
...
 
+
 class CosmosClient(object):
     """A client-side logical representation of an Azure Cosmos DB account.
 
     Use this client to configure and execute requests to the Azure Cosmos DB service.
 
-    :param str url: The URL of the Cosmos DB account.
+    :param url: The URL of the Cosmos DB account.
+    :type url: str
```
tjprescott, line 99: You can also simplify this to:
```
:param str url: The URL of the Cosmos DB account.
```

sdk/cosmos/azure-cosmos/azure/cosmos/aio/cosmos_client.py
```diff
@@ -109,8 +113,13 @@ class CosmosClient(object):
             :name: create_client
     """
 
-    def __init__(self, url, credential, consistency_level=None, **kwargs):
-        # type: (str, Any, Optional[str], Any) -> None
+    def __init__(
+            self,
+            url: str,
+            credential: Any,
```
tjprescott, line 119: The use of `Any` is discouraged as it does not provide much (if any?) value. This should be the class of your credential object.

sdk/cosmos/azure-cosmos/azure/cosmos/aio/cosmos_client.py
```diff
@@ -134,14 +142,18 @@ async def __aenter__(self):
...
-        # type: () -> None
+    async def close(self) -> None:
         """Close this instance of CosmosClient."""
         await self.__aexit__()
 
     @classmethod
-    def from_connection_string(cls, conn_str, credential=None, consistency_level=None, **kwargs):
-        # type: (str, Optional[Union[str, Dict[str, str]]], Optional[str], Any) -> CosmosClient
+    def from_connection_string(
+            cls: ClassType,
```
tjprescott, line 151: I would think this should be `CosmosClient` since that is logically the only type that could be returned here.

sdk/cosmos/azure-cosmos/azure/cosmos/aio/cosmos_client.py
```diff
@@ -134,14 +142,18 @@ async def __aenter__(self):
...
+            conn_str: str,
+            credential: Union[str, Dict[str, Any]] = None,
```
tjprescott, line 153: This should also be wrapped in `Optional[...]` since per the type comment, `None` is not a valid value.

sdk/cosmos/azure-cosmos/azure/cosmos/aio/cosmos_client.py
tjprescott: Your docstring :rtype: shows CosmosClient, so this should match.

sdk/cosmos/azure-cosmos/azure/cosmos/aio/cosmos_client.py
```diff
@@ -166,8 +180,7 @@ def from_connection_string(cls, conn_str, credential=None, consistency_level=Non
         )
 
     @staticmethod
-    def _get_database_link(database_or_id):
-        # type: (Union[DatabaseProxy, str, Dict[str, str]]) -> str
+    def _get_database_link(database_or_id: Union[DatabaseProxy, str, Dict[str, str]]) -> str:
```
tjprescott, line 183: Sometimes you are using `Dict[str, str]` and others `Dict[str, Any]`.  Is this intentional?

sdk/cosmos/azure-cosmos/azure/cosmos/aio/cosmos_client.py
```diff
@@ -261,12 +286,11 @@ async def create_database_if_not_exists(  # pylint: disable=redefined-builtin
...
             )
 
-    def get_database_client(self, database):
-        # type: (Union[str, DatabaseProxy, Dict[str, Any]]) -> DatabaseProxy
+    def get_database_client(self, database: Union[str, DatabaseProxy, Dict[str, Any]]) -> DatabaseProxy:
         """Retrieve an existing database with the ID (name) `id`.
 
         :param database: The ID (name) representing the properties of the database to read.
-        :type database: str or dict(str, str) or ~azure.cosmos.DatabaseProxy
+        :type database: str or Dict[str, Any] or ~azure.cosmos.DatabaseProxy
```
tjprescott, line 293: The description here seems inaccurate. It sounds like it is either the name or the actual properties themselves (as a dict or object).

sdk/cosmos/azure-cosmos/azure/cosmos/aio/cosmos_client.py
```diff
@@ -343,22 +375,26 @@ def query_databases(
...
 
         :param database: The ID (name), dict representing the properties, or :class:`DatabaseProxy`
             instance of the database to delete.
-        :type database: str or dict(str, str) or ~azure.cosmos.DatabaseProxy
-        :keyword str session_token: Token for use with Session consistency.
-        :keyword dict[str,str] initial_headers: Initial headers to be sent as part of the request.
-        :keyword str etag: An ETag value, or the wildcard character (*). Used to check if the resource
+        :type database: str or Dict[str, str] or ~azure.cosmos.DatabaseProxy
+        :keyword session_token: Token for use with Session consistency.
+        :paramtype session_token: str
```
tjprescott, line 388:
```suggestion
        :keyword str session_token: Token for use with Session consistency.
```

This can simplify a lot of your docstrings.

sdk/cosmos/azure-cosmos/azure/cosmos/aio/database.py
```diff
@@ -144,40 +149,52 @@ async def read(self, **kwargs):
...
-        default_ttl=None,  # type: Optional[int]
-        offer_throughput=None,  # type: Optional[int]
-        unique_key_policy=None,  # type: Optional[Dict[str, Any]]
-        conflict_resolution_policy=None,  # type: Optional[Dict[str, Any]]
-        **kwargs  # type: Any
-    ):
-        # type: (...) -> ContainerProxy
+            self,
+            id: str,  # pylint: disable=redefined-builtin
+            partition_key: Any,
```
tjprescott, line 154: This probably shouldn't be `Any`

sdk/cosmos/azure-cosmos/azure/cosmos/aio/database.py
```diff
@@ -630,29 +694,30 @@ async def upsert_user(self, body, **kwargs):
...
-        self,
-        user,  # type: Union[str, UserProxy, Dict[str, Any]]
-        body,  # type: Dict[str, Any]
-        **kwargs  # type: Any
-    ):
-        # type: (...) -> UserProxy
+            self,
+            user: Union[str, UserProxy, Dict[str, Any]],
+            body: Dict[str, Any],
+            **kwargs  # type: Any
```
tjprescott, line 700: Still have a `# type` here.

tjprescott (changes requested):
Overall LGTM.

sdk/cosmos/azure-cosmos/azure/cosmos/aio/cosmos_client.py
simorenoh: yep, depends on what's being returned; in this case we know for sure it's a string to string mapping, while for most of the others the body could also contain str to int or bool mappings.

sdk/cosmos/azure-cosmos/azure/cosmos/aio/scripts.py
```diff
@@ -189,31 +199,34 @@ async def delete_stored_procedure(self, sproc, **kwargs):
...
-        :rtype: dict[str, Any]
+        :returns: Result of the executed stored procedure for the given parameters.
+        :rtype: Dict[str, Any]
         """
 
         request_options = _build_options(kwargs)
         if partition_key is not None:
             request_options["partitionKey"] = (
-                _cosmos_client_connection._return_undefined_or_empty_partition_key(
+                _CosmosClientConnection._return_undefined_or_empty_partition_key(
```
annatisch, line 229: Is this a staticmethod?

sdk/cosmos/azure-cosmos/azure/cosmos/aio/user.py
```diff
@@ -148,16 +154,21 @@ def query_permissions(
         return result
 
     @distributed_trace_async
-    async def get_permission(self, permission, **kwargs):
-        # type: (Union[str, Dict[str, Any], Permission], Any) -> Permission
+    async def get_permission(
+            self,
+            permission: Union[str, Dict[str, Any], _Permission],
+            **kwargs: Any
+    ) -> _Permission:
```
annatisch, line 161: Unfortunately underscored typehints don't render as we'd like - is the reason we've imported this with an underscore to prevent it becoming duplicate API?
The best way to fix this - which we should do before GA regardless - is to make all these files private (i.e. `_user.py`) an only surface the parts we want through `__init__.py`. I know that will be inconsistent with the sync client... but we can't easily fix that, though I don't think it prevents us using a cleaner file structure for async.

In the short term, you can work around this with something like:
```python
from typing import TYPE_CHECKING:
if TYPE_CHECKING:
    from ..permission import Permission

async def get_permission(self, permission: "Permission") -> "Permission":
    ...
```

annatisch: Ah it looks like it's used in code as well as type hints - in which case I refer to my first suggestion :D

sdk/cosmos/azure-cosmos/azure/cosmos/aio/database.py
```diff
@@ -144,40 +149,52 @@ async def read(self, **kwargs):
...
-        offer_throughput=None,  # type: Optional[int]
-        unique_key_policy=None,  # type: Optional[Dict[str, Any]]
-        conflict_resolution_policy=None,  # type: Optional[Dict[str, Any]]
-        **kwargs  # type: Any
-    ):
-        # type: (...) -> ContainerProxy
+            self,
+            id: str,  # pylint: disable=redefined-builtin
+            partition_key: Any,
+            indexing_policy: Optional[Dict[str, Any]] = None,
```
annatisch, line 155: Lots of keyword-only candidates here...

sdk/cosmos/azure-cosmos/azure/cosmos/aio/database.py
```diff
@@ -532,26 +584,34 @@ async def create_user(self, body, **kwargs):
             client_connection=self.client_connection, id=user["id"], database_link=self.database_link, properties=user
         )
 
-    def get_user_client(self, user_id):
-        # type: (str) -> UserProxy
+    def get_user_client(self, user: Union[str, UserProxy, Dict[str, Any]]) -> UserProxy:
```
annatisch, line 587: I see this param was always documented as `user`, although it was previously `user_id` - what did the sync client have?

sdk/cosmos/azure-cosmos/azure/cosmos/aio/container.py
```diff
@@ -88,20 +90,17 @@ async def is_system_key(self):
...
-    def _get_document_link(self, item_or_link):
-        # type: (Union[Dict[str, Any], str]) -> str
+    def _get_document_link(self, item_or_link: (Union[Dict[str, Any], str])) -> str:
         if isinstance(item_or_link, str):
             return u"{}/docs/{}".format(self.container_link, item_or_link)
         return item_or_link["_self"]
 
-    def _get_conflict_link(self, conflict_or_link):
-        # type: (Union[Dict[str, Any], str]) -> str
+    def _get_conflict_link(self, conflict_or_link: (Union[Dict[str, Any], str])) -> str:
```
annatisch, line 103:
```suggestion
    def _get_conflict_link(self, conflict_or_link: Union[Dict[str, Any], str]) -> str:
```

sdk/cosmos/azure-cosmos/azure/cosmos/aio/container.py
```diff
@@ -88,20 +90,17 @@ async def is_system_key(self):
...
-    def scripts(self):
-        # type: () -> ScriptsProxy
+    def scripts(self) -> ScriptsProxy:
         if self._scripts is None:
             self._scripts = ScriptsProxy(self, self.client_connection, self.container_link)
         return cast('ScriptsProxy', self._scripts)
 
-    def _get_document_link(self, item_or_link):
-        # type: (Union[Dict[str, Any], str]) -> str
+    def _get_document_link(self, item_or_link: (Union[Dict[str, Any], str])) -> str:
```
annatisch, line 98:
```suggestion
    def _get_document_link(self, item_or_link: Union[Dict[str, Any], str]) -> str:
```

sdk/cosmos/azure-cosmos/azure/cosmos/aio/container.py
```diff
@@ -199,25 +214,25 @@ async def create_item(
     @distributed_trace_async
     async def read_item(
             self,
-            item,  # type: Union[str, Dict[str, Any]]
-            partition_key,  # type: Any
-            **kwargs  # type: Any
+            item: Union[str, Dict[str, Any]],
+            partition_key: Any,
```
annatisch, line 218: As Travis mentioned - we should revisit the classification of `partition_key` as Any throughout.

sdk/cosmos/azure-cosmos/azure/cosmos/aio/cosmos_client.py
```diff
@@ -282,19 +306,22 @@ def get_database_client(self, database):
 
     @distributed_trace
     def list_databases(
-        self,
-        max_item_count=None,  # type: Optional[int]
-        **kwargs  # type: Any
-    ):
-        # type: (...) -> AsyncItemPaged[Dict[str, Any]]
+            self,
+            max_item_count: Optional[int] = None,
```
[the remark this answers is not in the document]
simorenoh: Aligning the SDK to use keyword-only params was going to be part of another bullet point, but I can go ahead and do that work here too then.

sdk/cosmos/azure-cosmos/azure/cosmos/aio/scripts.py
simorenoh: _return_undefined_or_empty_partition_key is a static method in the client connection class

sdk/cosmos/azure-cosmos/azure/cosmos/aio/user.py
simorenoh: makes sense, it seems like regardless another point was to make as many things as we can private so this aligns with that too

sdk/cosmos/azure-cosmos/azure/cosmos/aio/scripts.py
```diff
@@ -49,21 +50,27 @@ class ScriptsProxy(object):
...
     """
 
-    def __init__(self, container, client_connection, container_link):
-        # type: (ContainerProxy, CosmosClientConnection, str) -> None
+    def __init__(
+            self,
+            container: _ContainerProxy,
+            client_connection: _CosmosClientConnection,
+            container_link: str
+    ) -> None:
```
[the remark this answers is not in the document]
simorenoh: they access it via the container class by calling `container.scripts.function_name_here()`

sdk/cosmos/azure-cosmos/azure/cosmos/aio/cosmos_client.py
simorenoh: The only way to get this to work for me has been to import CosmosClient from it's same file, which seems redundant  - is that the way to do it? I haven't been able to resolve the definition otherwise

sdk/cosmos/azure-cosmos/azure/cosmos/aio/database.py
simorenoh: user as well - this was a change we reverted since it didn't really make sense to edit to begin with, even if we only end up using the id

sdk/cosmos/azure-cosmos/azure/cosmos/aio/database.py
```diff
@@ -69,24 +73,28 @@ class DatabaseProxy(object):
...
+            id: str,  # pylint: disable=redefined-builtin
+            properties: Dict[str, Any] = None
+    ) -> None:
         """
-        :param ClientSession client_connection: Client from which this database was retrieved.
-        :param str id: ID (name) of the database.
+        :param client_connection: Client from which this database was retrieved.
+        :type client_connection: ~azure.cosmos.aio.CosmosClientConnection
+        :param id: ID (name) of the database.
+        :type id: str
```
tjprescott, line 86:
```suggestion
        :param str id: ID (name) of the database.
```

sdk/cosmos/azure-cosmos/azure/cosmos/aio/database.py
```diff
@@ -144,40 +150,38 @@ async def read(self, **kwargs):
...
-        default_ttl=None,  # type: Optional[int]
-        offer_throughput=None,  # type: Optional[int]
-        unique_key_policy=None,  # type: Optional[Dict[str, Any]]
-        conflict_resolution_policy=None,  # type: Optional[Dict[str, Any]]
-        **kwargs  # type: Any
-    ):
-        # type: (...) -> ContainerProxy
+            self,
+            id: str,  # pylint: disable=redefined-builtin
+            partition_key: PartitionKey,
```
tjprescott, line 155: Curious why this isn't the `Union[str, int, bool, float]` like in other places?

sdk/cosmos/azure-cosmos/azure/cosmos/aio/scripts.py
```diff
@@ -49,21 +50,27 @@ class ScriptsProxy(object):
...
         if isinstance(script_or_id, str):
             return u"{}/{}/{}".format(self.container_link, typ, script_or_id)
         return script_or_id["_self"]
 
     @distributed_trace
-    def list_stored_procedures(self, max_item_count=None, **kwargs):
-        # type: (Optional[int], Any) -> AsyncItemPaged[Dict[str, Any]]
+    def list_stored_procedures(
+            self,
+            max_item_count: Optional[int] = None,
```
tjprescott, line 71: Curious why you left this `max_item_count` alone instead of making kword-only

sdk/cosmos/azure-cosmos/azure/cosmos/aio/scripts.py
```diff
@@ -143,21 +150,22 @@ async def create_stored_procedure(self, body, **kwargs):
     @distributed_trace_async
     async def replace_stored_procedure(
         self,
-        sproc,
-        body,
-        **kwargs
-    ):
-        # type: (Union[str, Dict[str, Any]], Dict[str, Any], Any) -> Dict[str, Any]
+        sproc: Union[str, Dict[str, Any]],
```
tjprescott, line 153: Oh wow. I'm assuming this name is intuitive to stored procedure users?

tjprescott (changes requested):
Overall LGTM. Is there any APIVIew for this? It seems this is doing more than type hints--it's also converting things from regular args to kwarg-only.

sdk/cosmos/azure-cosmos/azure/cosmos/aio/container.py
```diff
@@ -199,25 +209,23 @@ async def create_item(
...
 
         :param item: The ID (name) or dict representing item to retrieve.
+        :type item: Union[str, Dict[str, Any]]
         :param partition_key: Partition key for the item to retrieve.
+        :type partition_key: Union[str, int, float, bool]
         :keyword str session_token: Token for use with Session consistency.
-        :keyword dict[str,str] initial_headers: Initial headers to be sent as part of the request.
-        :keyword Callable response_hook: A callable invoked with the response metadata.
-        **Provisional** keyword argument max_integrated_cache_staleness_in_ms
-        :keyword int max_integrated_cache_staleness_in_ms:
```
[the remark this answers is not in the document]
simorenoh: good catch thank you, seems I deleted more than I meant to when changing the typehints

sdk/cosmos/azure-cosmos/azure/cosmos/aio/cosmos_client.py
simorenoh: It is unable to find the reference to CosmosClient when it is just put like this, so I went with ClassType the way I saw other SDKs do

sdk/cosmos/azure-cosmos/azure/cosmos/aio/database.py
simorenoh: For creating the container you have to pass in a PartitionKey definition, which requires the PartitionKey object in order to specify a partition key path as well

The other methods were dealing with finding an item through a partition key value, at which point you only need to know what the item's partition key value is as opposed to the definition

sdk/cosmos/azure-cosmos/azure/cosmos/aio/scripts.py
simorenoh: oops, because I missed it 🤦 thank you for catching this!

sdk/cosmos/azure-cosmos/azure/cosmos/aio/scripts.py
simorenoh: yeah, it's also the naming present in the sync client

sdk/cosmos/azure-cosmos/azure/cosmos/aio/container.py
```diff
@@ -199,25 +209,23 @@ async def create_item(
     @distributed_trace_async
     async def read_item(
             self,
-            item,  # type: Union[str, Dict[str, Any]]
-            partition_key,  # type: Any
-            **kwargs  # type: Any
+            item: Union[str, Dict[str, Any]],
+            partition_key: Union[str, int, float, bool],
```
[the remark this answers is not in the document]
simorenoh: I tested it to ensure I could pinpoint what the previous "Any" was and it looks like it can be any of these types as well

sdk/cosmos/azure-cosmos/azure/cosmos/aio/container.py
```diff
@@ -115,27 +114,28 @@ def _set_partition_key(self, partition_key) -> Union[str, Awaitable]:
     @distributed_trace_async
     async def read(
             self,
-            populate_partition_key_range_statistics=None,  # type: Optional[bool]
-            populate_quota_info=None,  # type: Optional[bool]
-            **kwargs  # type: Any
+            **kwargs: Any
     ) -> Dict[str, Any]:
```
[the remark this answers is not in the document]
simorenoh: isn't a breaking change a change that causes a difference in behavior experienced by the end user? I don't believe any of these changes would do that

sdk/cosmos/azure-cosmos/azure/cosmos/aio/cosmos_client.py
tjprescott: You can use a forward reference for `"CosmosClient"` if you need to. It would seem to me that if other SDKs are just using `ClassType` then they are incorrect.

sdk/cosmos/azure-cosmos/azure/cosmos/aio/container.py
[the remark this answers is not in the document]
tjprescott: That is a behavioral breaking change, but anything that forces customers to change their code is the classical breaking change... and this is definitely that.

sdk/cosmos/azure-cosmos/azure/cosmos/aio/container.py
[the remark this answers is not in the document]
tjprescott: I guess I assumed a partition key to be a string or object. But okie dokie :)

tjprescott:
On a related note, now that the SDK no longer supports Python 2.7, you could actually use Py3-style type hinting throughout your sync client. For another PR :grin:

tjprescott:
Is there an APIView for this?

sdk/cosmos/azure-cosmos/azure/cosmos/aio/container.py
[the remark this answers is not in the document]
annatisch: Well, it's not _breaking_ because aio was never GA.
If we did it on the sync client it would be breaking - but there are ways we can guide users towards this in a non-breaking way for sync, while fully embracing keyword-only for async.

sdk/cosmos/azure-cosmos/azure/cosmos/aio/user.py
```diff
@@ -115,23 +118,24 @@ def list_permissions(self, max_item_count=None, **kwargs):
...
     def query_permissions(
         self,
-        query,
-        parameters=None,
-        max_item_count=None,
-        **kwargs
-    ):
-        # type: (str, Optional[List[str]], Optional[int], Any) -> AsyncItemPaged[Dict[str, Any]]
+        query: str,
+        parameters: Optional[List[Dict[str, Any]]] = None,
```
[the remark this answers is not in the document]
simorenoh: I left parameters outside of the kwargs since it's a part of the query methods that users should be aware of and that users constantly use along with their queries

sdk/cosmos/azure-cosmos/azure/cosmos/aio/scripts.py
```diff
@@ -189,35 +196,38 @@ async def delete_stored_procedure(self, sproc, **kwargs):
...
-        sproc,  # type: Union[str, Dict[str, Any]]
-        partition_key=None,  # type: Optional[str]
-        params=None,  # type: Optional[List[Any]]
-        enable_script_logging=None,  # type: Optional[bool]
-        **kwargs  # type: Any
-    ):
-        # type: (...) -> Any
+        sproc: Union[str, Dict[str, Any]],
+        partition_key: Optional[Union[str, int, float, bool]] = None,
+        params: Optional[List[Dict[str, Any]]] = None,
```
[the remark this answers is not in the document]
simorenoh: like with the parameters in the query method, I opted into leaving these two variables in since they play an important part in the method's functionality

sdk/cosmos/azure-cosmos/azure/cosmos/aio/user.py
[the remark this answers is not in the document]
simorenoh: nvm im silly

sdk/cosmos/azure-cosmos/azure/cosmos/aio/__init__.py
```diff
@@ -19,16 +19,6 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 
-from .container import ContainerProxy
 from .cosmos_client import CosmosClient
-from .database import DatabaseProxy
-from .user import UserProxy
-from .scripts import ScriptsProxy
```
annatisch, line 26: All of these proxy objects should remain here - otherwise they're not public, and will never be documented or show up in APIview :)

annatisch: And also added to `__all__`

sdk/cosmos/azure-cosmos/azure/cosmos/aio/container.py
```diff
@@ -22,7 +22,7 @@
 """Create, read, update and delete items in the Azure Cosmos DB SQL API service.
```
annatisch, line 22: Were you going to rename these files in this PR? Or separately? (i.e. -> _container.py). I don't mind if we do it separately if we want to get this PR in.

sdk/cosmos/azure-cosmos/azure/cosmos/aio/container.py
```diff
@@ -199,25 +209,27 @@ async def create_item(
...
+        :type item: Union[str, Dict[str, Any]]
         :param partition_key: Partition key for the item to retrieve.
+        :type partition_key: Union[str, int, float, bool]
         :keyword str session_token: Token for use with Session consistency.
-        :keyword dict[str,str] initial_headers: Initial headers to be sent as part of the request.
-        :keyword Callable response_hook: A callable invoked with the response metadata.
+        :keyword Dict[str, str] initial_headers: Initial headers to be sent as part of the request.
+        :keyword response_hook: A callable invoked with the response metadata.
+        :paramtype response_hook: Callable[[Dict[str, str], Dict[str, Any]], None]
         **Provisional** keyword argument max_integrated_cache_staleness_in_ms
```
annatisch, line 226: Not relevant to this PR - but we should really try to find a better name for this before it releases.
Lastly, I don't think this notation is going to render properly - so we should double check the sphinx docs.

sdk/cosmos/azure-cosmos/azure/cosmos/aio/container.py
```diff
@@ -374,36 +388,38 @@ def query_items(
     @distributed_trace
     def query_items_change_feed(
             self,
-            partition_key_range_id=None,  # type: Optional[str]
-            is_start_from_beginning=False,  # type: bool
-            continuation=None,  # type: Optional[str]
-            max_item_count=None,  # type: Optional[int]
-            **kwargs  # type: Any
+            is_start_from_beginning: bool = False,
```
annatisch, line 391: We missed one.

sdk/cosmos/azure-cosmos/azure/cosmos/aio/scripts.py
```diff
@@ -22,19 +22,20 @@
...
 
-from azure.cosmos.aio._cosmos_client_connection_async import CosmosClientConnection as _cosmos_client_connection
+from ._cosmos_client_connection_async import CosmosClientConnection as _CosmosClientConnection
 from .._base import build_options as _build_options
 from ..partition_key import NonePartitionKeyValue
 
 # pylint: disable=protected-access
 # pylint: disable=missing-client-constructor-parameter-credential,missing-client-constructor-parameter-kwargs
 
+ClassType = TypeVar("ClassType")
```
annatisch, line 38: We can get rid of this now right?

sdk/cosmos/azure-cosmos/azure/cosmos/aio/scripts.py
```diff
@@ -189,38 +196,41 @@ async def delete_stored_procedure(self, sproc, **kwargs):
...
-        :param partition_key: Specifies the partition key to indicate which partition the sproc should execute on.
-        :param params: List of parameters to be passed to the stored procedure to be executed.
-        :param bool enable_script_logging: Enables or disables script logging for the current request.
-        :returns: Result of the executed stored procedure for the given parameters.
+        :type sproc: Union[str, Dict[str, Any]]
+        :keyword partition_key: Specifies the partition key to indicate which partition the stored procedure should
+            execute on.
+        :paramtype partition_key: Union[str, int, float, bool]
+        :keyword params: List of parameters to be passed to the stored procedure to be executed.
+        :paramtype params: List[Dict[str, Any]]
```
annatisch, line 212: We had discussed renaming this to parameters - are we gonna keep that for a separate PR?

sdk/cosmos/azure-cosmos/azure/cosmos/aio/cosmos_client.py
```diff
@@ -135,40 +141,39 @@ async def __aenter__(self):
...
         return await self.client_connection.pipeline_client.__aexit__(*args)
 
-    async def close(self):
-        # type: () -> None
+    async def close(self) -> None:
         """Close this instance of CosmosClient."""
         await self.__aexit__()
 
     @classmethod
-    def from_connection_string(cls, conn_str, credential=None, consistency_level=None, **kwargs):
```
annatisch, line 144: Looks like we dropped optionally passing in a credential here?
We usually support that for cases where customers want to use a TokenCredential, in place of the key in the connection string. Agreed it would be unusual, so it can be hidden away in kwargs.

sdk/cosmos/azure-cosmos/azure/cosmos/aio/cosmos_client.py
```diff
@@ -258,16 +265,15 @@ async def create_database_if_not_exists(  # pylint: disable=redefined-builtin
         except CosmosResourceNotFoundError:
             return await self.create_database(
                 id,
-                offer_throughput=offer_throughput,
```
annatisch, line 261: Why are we not passing this through any more?

sdk/cosmos/azure-cosmos/azure/cosmos/aio/database.py
```diff
@@ -358,31 +371,31 @@ def list_containers(
...
     def query_containers(
-        self,
-        query=None,  # type: Optional[str]
-        parameters=None,  # type: Optional[List[Dict[str, Any]]]
-        max_item_count=None,  # type: Optional[int]
-        **kwargs  # type: Any
-    ):
-        # type: (...) -> AsyncItemPaged[Dict[str, Any]]
+            self,
+            query: Union[str, Dict[str, Any]],
```
annatisch, line 375: This was previously optional?

sdk/cosmos/azure-cosmos/azure/cosmos/aio/__init__.py
simorenoh: I'll revert these then but I'm confused with how we would want to make it so we only expose the client in that case? I had thought this was the outcome of what we had discussed in the meeting but seems like I'll need some clarification

sdk/cosmos/azure-cosmos/azure/cosmos/aio/scripts.py
simorenoh: changed

sdk/cosmos/azure-cosmos/azure/cosmos/aio/database.py
simorenoh: changed back and packaged into kwargs since it's optional

tjprescott (approved):

annatisch (approved):

tjprescott:
Okay, I'm clicking this thing folks.

Issue #23934, opened 2022-04-11
Azure CLI progress tracking for concurrent blob upload/download not working


jalauzon-msft:
As reported in https://github.com/Azure/azure-cli/issues/21781

With the CLI's recent switch to the Track2 SDK for blob uploads, the progress bar does not work properly. It jumps around and does not usually reach 100%. This is due to an issue where the SDK does not properly report progress on **concurrent** uploads.

This is likely also broken for chunked, concurrent download as well as both operations Datalake/File Share.

jalauzon-msft:
Both of the fixes have been released in version 12.12.0.

PR #23965, opened 2022-04-12
[Storage] Add progress tracking callback for blob upload

jalauzon-msft:
This aims to address #23934 by introducing an optional callback on blob upload that can be used to track progress. The existing method of tracking progress is to pass a response hook and examine special properties we add to the response context, but this will not work with concurrent upload/download since there is no way to aggregate the results from other threads.

This change introduces a new callback specifically for tracking progress that is called in the correct place where the upload progress is being aggregated (in a thread safe manner).

Some notes:
- This does not actually fix the existing method for progress tracking, but rather introduces a more robust solution. It is likely not possible to "fix" the existing method for concurrent operations.
- This issue will exist in datalake and file-share as well, but this change is scoped to just blob for now. A similar pattern can be followed for other services though.


vincenttran-msft (approved):
LGTM! Great work on this, as I know it was a high-priority issue 👍

PR #23975, opened 2022-04-13
[qna] regenerate with latest autorest to pass autorest ci

iscai-msft:


kristapratico (approved):

annatisch:
Thanks @iscai-msft - there's actually two clients in the QnA SDK, could you also regenerate for swagger/authoring?

Also, while looking at your PR, I noticed this - which I would also appreciate your review on ;)
https://github.com/Azure/azure-sdk-for-python/pull/23983

PR #23977, opened 2022-04-13
[EventHub] fix pylint plugin errors

swathipil:
fix new pylint rules:
- C4747: enum-must-inherit-case-insensitive-enum-meta
- C4748: client-accepts-api-version-keyword


kristapratico (approved):

l0lawrence:
azure/eventhub/_consumer_client.py:34: [E0012(bad-option-value), ] Bad option value 'client-accepts-api-version-keyword'
 This is popping up b/c that specific checker doesn't exist yet in main, once it is merged you can #pylint: disable

PR #23982, opened 2022-04-13
[Storage] Fix Queue account SAS generation with latest service version

jalauzon-msft:
In a previous change the value of `X_MS_VERSION` was updated to use the latest supported service version rather than the version from the generated code (this was done to reflect what our clients do by default). This also updated the version used for SAS generation by default and therefore broke account SAS generation due to the not including the recent change to account SAS generation that only affects later service versions.

This change adds that required change to SAS generation as well as updates our version tests to use the `X_MS_VERSION` constant.


vincenttran-msft (approved):
LGTM! Good find 😄

PR #23996, opened 2022-04-14
FIX: give useful error msg when bad URL passed to BlobClient.from_blob_url()

pat-hearps:
# Description

**Problem:** A user tries to use `BlobClient.from_blob_url()` with an incorrect URL that does not have a container name and blob path, e.g. `BlobClient.from_blob_url("https://testaccount.blob.core.windows.net/containername/")` - note the missing blobname after the containername
**Current Behaviour:** User gets obscure error message: Because the list `path_blob` only has length of 1, i.e. `["containername"]`.
```
>       container_name, blob_name = unquote(path_blob[-2]), unquote(path_blob[-1])
E       IndexError: list index out of range
```

**Fixed Behaviour**: raises more helpful error: `ValueError("Invalid URL. Provide a blob_url with a valid blob and container name.")`

- [X] **The pull request does not introduce [breaking changes]**

- Test 'test_from_blob_url_too_short_url()` added to `test_blob_client.py`

### DIFF
--- a/sdk/storage/azure-storage-blob/azure/storage/blob/_blob_client.py
+++ b/sdk/storage/azure-storage-blob/azure/storage/blob/_blob_client.py
@@ -241,23 +241,24 @@ class BlobClient(StorageAccountHostsMixin):  # pylint: disable=too-many-public-m
         account_path = ""
         if ".core." in parsed_url.netloc:
             # .core. is indicating non-customized url. Blob name with directory info can also be parsed.
-            path_blob = parsed_url.path.lstrip('/').split('/', 1)
+            path_blob = parsed_url.path.lstrip('/').split('/', maxsplit=1)
         elif "localhost" in parsed_url.netloc or "127.0.0.1" in parsed_url.netloc:
-            path_blob = parsed_url.path.lstrip('/').split('/', 2)
+            path_blob = parsed_url.path.lstrip('/').split('/', maxsplit=2)
             account_path += '/' + path_blob[0]
         else:
             # for customized url. blob name that has directory info cannot be parsed.
             path_blob = parsed_url.path.lstrip('/').split('/')
             if len(path_blob) > 2:
                 account_path = "/" + "/".join(path_blob[:-2])
+
         account_url = "{}://{}{}?{}".format(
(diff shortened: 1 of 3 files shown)


sdk/storage/azure-storage-blob/azure/storage/blob/_blob_client.py
```diff
@@ -241,23 +241,24 @@ def from_blob_url(cls, blob_url, credential=None, snapshot=None, **kwargs):
...
             if len(path_blob) > 2:
                 account_path = "/" + "/".join(path_blob[:-2])
+
         account_url = "{}://{}{}?{}".format(
             parsed_url.scheme,
             parsed_url.netloc.rstrip('/'),
             account_path,
             parsed_url.query)
-        container_name, blob_name = unquote(path_blob[-2]), unquote(path_blob[-1])
-        if not container_name or not blob_name:
```
jalauzon-msft, line 259: Can we leave this check in after line 261 in case the user passes an empty string for blob or container?

jalauzon-msft (commented):
Thanks for the contribution! Just one minor comment and then we should be able to get this merged.

sdk/storage/azure-storage-blob/azure/storage/blob/_blob_client.py
pat-hearps: Ok sure, I've put it back in with the duplicated error message refactored into a string variable.

I have not added a test for this scenario (empty blob/container string).
I tried a few different ways to feed in a bad blob_url, but couldn't find one that would:
- make it past the new check I put in above `unquote(path_blob[-2|-1]` call;
- then fail at `if not container_name|blob_name`.

jalauzon-msft: Thanks! Yeah, I'm not entirely sure if/how we could hit that case, but I think it's best to leave it in to be safe.

jalauzon-msft (approved):

jalauzon-msft:
Thanks again @pat-hearps Patrick for your contribution! I've merged your change in and it will be part of the next release.

Issue #23643, opened 2022-03-24
[ACR] Support pull/push OCI blobs and manifests


YalinLi0312:
This is REST spec: [link](https://github.com/Azure/azure-rest-api-specs/blob/2c33d5572dab4c6f52faf31004f0561205737107/specification/containerregistry/data-plane/Azure.ContainerRegistry/stable/2021-07-01/containerregistry.json).

.NET already finished some implements, this is the feature branch history in .NET: [link](https://github.com/Azure/azure-sdk-for-net/commits/feature/containerregistry/oci-blob-api/sdk/containerregistry/Azure.Containers.ContainerRegistry)

The commits of interest are:

- [x] [[ACR] Add an overload for UploadManifest() that takes a Stream as…](https://github.com/Azure/azure-sdk-for-net/commit/0f1dcb8cbfb7a964527ee39744e670af9f3add97#diff-7928b67f65311e73a11ae0b06bae59bdb349280e268fabb26b076a37fc19b31d)

    - [x] Added an overload for `UploadManifest(Async)` method that takes the manifest `Stream` as an input.
    - [x] Added methods in `ContainerRegistryModelFactory` that create instances of `DownloadBlobResult`, `DownloadManifestResult`, `UploadBlobResult` and `UploadManifestResult` for mocking.
    - [x] Added `DownloadManifestOptions` type to allow callers to  pass-in either a tag or a digest in `DownloadManifest(Async)`.
    - [x] Added `ManifestStream` as a property in `DownloadManifestResult` that contains the raw manifest stream from the service response.
    - [x] Changed `DownloadManifest(Async)` method to take `DownloadManifestOptions` as an input parameter. This allows callers to pass-in either a tag or a digest as the manifest identifier.
- [x] [[ACR] Verify Audience must be set in Options](https://github.com/Azure/azure-sdk-for-net/commit/a89b06dfff1aa7928b70cb2d543e10407983338b#diff-7928b67f65311e73a11ae0b06bae59bdb349280e268fabb26b076a37fc19b31d)
- [x] [[ACR] Support anon access in ACR blob client](https://github.com/Azure/azure-sdk-for-net/commit/9a2c067f94bedfdc007653a1796aa58139d9aa3f#diff-7928b67f65311e73a11ae0b06bae59bdb349280e268fabb26b076a37fc19b31d)
[...]

PR #24004, opened 2022-04-14
[ACR] Pull/push support for OCI manifest and blob

YalinLi0312:
For issue: https://github.com/Azure/azure-sdk-for-python/issues/23643


sdk/containerregistry/azure-containerregistry/azure/containerregistry/aio/_async_container_registry_client.py
```diff
@@ -762,3 +749,124 @@ async def update_tag_properties(
...
+        :param str repository: Name of the repository
+        :param manifest: The manifest to upload.
+        :type manifest: ~azure.containerregistry.models.OCIManifest or IO
+        :keyword tag: Tag of the manifest.
+        :paramtype tag: str
+        :returns: None
+        :rtype: None
+        """
+        tag_or_digest = tag
+        if tag:
```
annatisch, line 768: Should be `if tag is None`?

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_registry_client.py
```diff
@@ -338,33 +349,6 @@ def get_next(next_link=None):
 
         return ItemPaged(get_next, extract_data)
 
-    @distributed_trace
-    def delete_manifest(self, repository, tag_or_digest, **kwargs):
```
annatisch, line 342: Why did we lose `delete_manifest`?

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_registry_client.py
```diff
@@ -764,3 +748,146 @@ def update_repository_properties(self, *args, **kwargs):
...
+        :param manifest: The manifest to upload.
+        :type manifest: ~azure.containerregistry.models.OCIManifest or IO
+        :keyword tag: Tag of the manifest.
+        :paramtype tag: str or None
+        :returns: None
+        :rtype: None
+        :raises ValueError: If the parameter repository or manifest is None.
+        """
+        if (repository or manifest) is None:
+            raise ValueError("The parameter repository and manifest cannot be None.")
```
annatisch, line 768: Is there a reason we're doing client-side validation here?
Normally we would just let these values go to the service and let the service raise an error? Though sometimes if that error is user-unfriendly, we might do this...

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_registry_client.py
```diff
@@ -764,3 +748,146 @@ def update_repository_properties(self, *args, **kwargs):
...
+
+        tag_or_digest = tag
+        if tag is None:
```
annatisch, line 771: Is an empty string a valid tag?

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_registry_client.py
```diff
@@ -764,3 +748,146 @@ def update_repository_properties(self, *args, **kwargs):
...
+            if isinstance(manifest, OCIManifest):
+                stream = _serialize_manifest(manifest)
+            tag_or_digest = _compute_digest(stream)
+
+        self._client.container_registry.create_manifest(
+            name=repository, reference=tag_or_digest, payload=stream, content_type=OCI_MANIFEST_MEDIA_TYPE, **kwargs
+        )
+
+    @distributed_trace
+    def upload_blob(self, repository, stream, **kwargs):
```
annatisch, line 782: I'd prefer we rename the `stream` parameter to something like `data`.
What's the maximum size supported here? Or minimum size? Would I ever want to pass in bytes directly? Or will always be a stream/file handle/etc?

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_registry_client.py
```diff
@@ -764,3 +748,146 @@ def update_repository_properties(self, *args, **kwargs):
...
+        # type: (str, IO, **Any) -> None
+        """Upload an artifact blob.
+
+        :param str repository: Name of the repository
+        :param stream: The manifest to upload.
```
annatisch, line 787: This data is a manifest?

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_registry_client.py
```diff
@@ -764,3 +748,146 @@ def update_repository_properties(self, *args, **kwargs):
...
+        :raises ValueError: If the parameter repository or stream is None.
+        """
+        if (repository or stream) is None:
+            raise ValueError("The parameter repository and stream cannot be None.")
+
+        start_upload_respose = self._client.container_registry_blob.start_upload(
+            repository, cls=_return_response, **kwargs
+        )
+        upload_chunk_response = self._client.container_registry_blob.upload_chunk(
+            start_upload_respose['Location'], stream, cls=_return_response, **kwargs
```
annatisch, line 800: What's the maximum data size supported by a single chunk? Could I ever need to upload multiple chunks in a single blob?

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_registry_client.py
```diff
@@ -764,3 +748,146 @@ def update_repository_properties(self, *args, **kwargs):
...
+        )
+        digest = _compute_digest(stream)
+        self._client.container_registry_blob.complete_upload(
+            digest=digest, next_link=upload_chunk_response['Location'], value=stream, cls=_return_response, **kwargs
```
annatisch, line 804: The data in the `stream` IO has likely already been read, so not sure if passing it again here will do much, as it can't be consumed again?

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_registry_client.py
```diff
@@ -764,3 +748,146 @@ def update_repository_properties(self, *args, **kwargs):
...
+        )
+
+    @distributed_trace
+    def download_manifest(self, repository, tag_or_digest, **kwargs):
+        # type: (str, str, **Any) -> ManifestWrapper
```
annatisch, line 809: `ManifestWrapper`? Was this already a public model in the API?

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_registry_client.py
```diff
@@ -764,3 +748,146 @@ def update_repository_properties(self, *args, **kwargs):
...
+        :param str tag_or_digest: The manifest to upload.
+        :returns: ManifestWrapper
+        :rtype: ~container_registry.models.ManifestWrapper
+        :raises ValueError: If the parameter repository or tag_or_digest is None.
+        """
+        if repository or tag_or_digest is None:
+            raise ValueError("The parameter repository and tag_or_digest cannot be None.")
+
+        return self._client.container_registry.get_manifest(
+            name=repository, reference=tag_or_digest, accept=OCI_MANIFEST_MEDIA_TYPE, **kwargs
```
annatisch, line 822: How big will this be? i.e. will this method block while downloading a large payload?

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_registry_client.py
```diff
@@ -764,3 +748,146 @@ def update_repository_properties(self, *args, **kwargs):
...
+        :param str repository: Name of the repository
+        :param str digest: The digest of the blob to download.
+        :returns: IO or None
+        :rtype: IO or None
+        :raises ValueError: If the parameter repository or digest is None.
+        """
+        if repository or digest is None:
+            raise ValueError("The parameter repository and digest cannot be None.")
+
+        return self._client.container_registry_blob.get_blob(repository, digest, **kwargs)
```
annatisch, line 839: So this method is returning a readable stream?

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_registry_client.py
annatisch: Nevermind - I see it reappeared below :)

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_registry_client.py
```diff
@@ -764,3 +748,146 @@ def update_repository_properties(self, *args, **kwargs):
...
+        if _is_tag(tag_or_digest):
+            tag_or_digest = self._get_digest_from_tag(repository, tag_or_digest)
+
+        self._client.container_registry.delete_manifest(repository, tag_or_digest, **kwargs)
+
+    @distributed_trace
+    def delete_blob(self, repository, tag_or_digest, **kwargs):
+        # type: (str, str, **Any) -> IO
+        """Delete a blob. If the blob cannot be found or a response status code of
+        404 is returned an error will not be raised.
```
annatisch, line 872: We don't really like raising on 404 for delete.

annatisch: Though I guess if that was the existing behaviour of `delete_manifest` we should probably be consistent...

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_registry_client.py
```diff
@@ -764,3 +748,146 @@ def update_repository_properties(self, *args, **kwargs):
...
+        :param stream: The manifest to upload.
+        :type stream: IO
+        :returns: None
+        :rtype: None
+        :raises ValueError: If the parameter repository or stream is None.
+        """
+        if (repository or stream) is None:
+            raise ValueError("The parameter repository and stream cannot be None.")
+
+        start_upload_respose = self._client.container_registry_blob.start_upload(
```
annatisch, line 796: Linking together 3 REST calls into a single method can have complications - is this `upload_blob` method idempotent? If the `complete_upload` call on line 803 fails due to flaky network and I call `upload_blob` a second time with the same inputs - will it succeed?

sdk/containerregistry/azure-containerregistry/azure/containerregistry/aio/_async_container_registry_client.py
```diff
@@ -762,3 +749,124 @@ async def update_tag_properties(
...
+    async def download_blob(self, repository, digest, **kwargs):
+        # type: (str, str, **Any) -> IO | None
+        """Download a blob that is part of an artifact.
+
+        :param str repository: Name of the repository
+        :param str digest: The digest of the blob to download.
+        :returns: IO or None
+        :rtype: IO or None
+        """
+        return await self._client.container_registry_blob.get_blob(repository, digest, **kwargs)
```
annatisch, line 819: We're not supporting the `get_digest_from_tag` conversion here?

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_registry_client.py
YalinLi0312: Nope, this is the first time using in the API surface.

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_registry_client.py
YalinLi0312: This is a mistake, should be a blob

sdk/containerregistry/azure-containerregistry/azure/containerregistry/aio/_async_container_registry_client.py
YalinLi0312: get_blob() consumes the value of digest, not tag. I don't think it's necessary to support tag(include digest value) as input, because users can do the convention.

sdk/containerregistry/azure-containerregistry/azure/containerregistry/aio/_async_container_registry_client.py
YalinLi0312: updated

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_registry_client.py
YalinLi0312: That's not valid, it will fail in the service end. We don't validate the tag value as in other languages. I'm thinking to pass cls and return the http response.

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_registry_client.py
```diff
@@ -764,44 +764,47 @@ def upload_manifest(
...
         upload_chunk_response = self._client.container_registry_blob.upload_chunk(
-            start_upload_respose['Location'], stream, cls=_return_response, **kwargs
+            start_upload_respose['Location'], data, cls=_return_response, **kwargs
         )
-        digest = _compute_digest(stream)
+        digest = _compute_digest(data)
+        data.seek(0)
         self._client.container_registry_blob.complete_upload(
-            digest=digest, next_link=upload_chunk_response['Location'], value=stream, cls=_return_response, **kwargs
+            digest=digest, next_link=upload_chunk_response['Location'], value=data, cls=_return_response, **kwargs
```
annatisch: I'm still not sure why the service needs the data to be passed in twice - once for `upload_chunk` and once for `complete_upload`?

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_helpers.py
```diff
@@ -133,26 +124,13 @@ def _parse_exp_time(raw_token):
...
-def _compute_digest(stream):
+def _compute_digest(data):
     # type: (IO) -> str
-    # bytes_value = stream.getvalue()
-    # dict_val = json.loads(str_value)
-    # res = hashlib.sha256(bytes_value)
-    return "sha256:" + hashlib.sha256(stream.read()).hexdigest()
+    data.seek(0)
+    # return "sha256:" + hashlib.sha256(data.read()).hexdigest()
+    return "sha256:e2fffb656966c98a0b7712672e996a58327eff7705fb3d8d9581fc16ef412d13"
```
annatisch: We should be able to remove this hardcoded value now.

sdk/containerregistry/azure-containerregistry/tests/test_container_registry_client.py
```diff
@@ -603,9 +604,50 @@ def test_get_misspell_property(self, containerregistry_endpoint):
...
         client = self.create_registry_client(containerregistry_endpoint)
         
         client.upload_manifest(repo, manifest)
+    
+    @acr_preparer()
+    def test_upload_oci_manifest_stream(self, containerregistry_endpoint):
+        repo = self.get_resource_name("repo")
+        manifest = self.create_oci_manifest()
+        manifest_bytes = json.dumps(manifest.serialize()).encode("utf-8")
+        manifest_stream = BytesIO(manifest_bytes)
```
annatisch: Rather than creating a BytesIO this way - maybe a good test to do would be reading the manifest from a file? That would be a more likely customer scenario.

sdk/containerregistry/azure-containerregistry/tests/test_container_registry_client.py
```diff
@@ -603,9 +604,50 @@ def test_get_misspell_property(self, containerregistry_endpoint):
...
+        
+        client = self.create_registry_client(containerregistry_endpoint)
+        
+        client.upload_manifest(repo, manifest_stream)
```
annatisch: How can we validate that this test succeeded?
Is there a `get` call that we can do to validate that the manifest uploaded successfully?

sdk/containerregistry/azure-containerregistry/tests/test_container_registry_client.py
```diff
@@ -603,9 +604,50 @@ def test_get_misspell_property(self, containerregistry_endpoint):
...
+    
+    @acr_preparer()
+    def test_upload_oci_manifest_with_tag(self, containerregistry_endpoint):
+        repo = self.get_resource_name("repo")
+        manifest = self.create_oci_manifest()
+        client = self.create_registry_client(containerregistry_endpoint)
+        
+        self.upload_manifest_prerequisites()
+        client.upload_manifest(repo, manifest, tag="")
```
annatisch: Didn't we discuss that an empty string is an invalid tag? So this test should be rejected by the service?

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_registry_client.py
```diff
@@ -764,3 +760,188 @@ def update_repository_properties(self, *args, **kwargs):
...
+            payload=data,
+            content_type=OCI_MANIFEST_MEDIA_TYPE,
+            headers={"Accept": OCI_MANIFEST_MEDIA_TYPE},
+            cls=_return_response_header,
+            **kwargs
+        )
+        
+        digest = response['Docker-Content-Digest']
+        if not _validate_digest(data, digest):
+            raise HttpResponseError("The digest in the response does not match the digest of the uploaded manifest.")
```
annatisch, line 803: This shouldn't be an HttpResponseError unless it's actually based on an Http Response.
Possibly this should be a ValueError?

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_registry_client.py
```diff
@@ -764,3 +760,188 @@ def update_repository_properties(self, *args, **kwargs):
...
+        return digest
+
+    @distributed_trace
+    def upload_blob(self, repository, data, **kwargs):
+        # type: (str, IO, **Any) -> str
+        """Upload an artifact blob.
+
+        :param str repository: Name of the repository
+        :param data: The blob to upload.
+        :type data: IO
```
annatisch, line 814: We probably need to document that this _must_ be a seekable stream.

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_registry_client.py
```diff
@@ -764,3 +760,188 @@ def update_repository_properties(self, *args, **kwargs):
...
+            repository, cls=_return_response_header, **kwargs
+        )
+        upload_chunk_response = self._client.container_registry_blob.upload_chunk(
+            start_upload_response['Location'], data, cls=_return_response_header, **kwargs
+        )
+        digest = _compute_digest(data)
+        complete_upload_response = self._client.container_registry_blob.complete_upload(
+            digest=digest, next_link=upload_chunk_response['Location'], cls=_return_response_header, **kwargs
+        )
+        return complete_upload_response['Docker-Content-Digest']
```
annatisch, line 832: We don't validate the digest here like we did with upload_manifest?

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_registry_client.py
```diff
@@ -764,3 +760,188 @@ def update_repository_properties(self, *args, **kwargs):
...
+            accept=OCI_MANIFEST_MEDIA_TYPE,
+            headers={"Accept": OCI_MANIFEST_MEDIA_TYPE},
+            cls=_return_response,
+            **kwargs
+        )
+        digest = response.http_response.headers['Docker-Content-Digest']
+        content = response.http_response.internal_response._content
+        content_stream = BytesIO(content)
+        
+        if not _validate_digest(content_stream, digest):
```
annatisch, line 862: If the data in the response here actually streamed?
Or is all the data downloaded by the first HTTP call and already sitting in memory?

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_registry_client.py
```diff
@@ -764,3 +760,188 @@ def update_repository_properties(self, *args, **kwargs):
...
+        response = self._client.container_registry.get_manifest(
+            name=repository,
+            reference=tag_or_digest,
+            accept=OCI_MANIFEST_MEDIA_TYPE,
+            headers={"Accept": OCI_MANIFEST_MEDIA_TYPE},
+            cls=_return_response,
+            **kwargs
+        )
+        digest = response.http_response.headers['Docker-Content-Digest']
+        content = response.http_response.internal_response._content
```
annatisch, line 859: This is going to break if the transport is swapped out. We should not use the `internal_response` object at all - only the `http_response`.

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_registry_client.py
```diff
@@ -764,3 +760,188 @@ def update_repository_properties(self, *args, **kwargs):
...
+        """
+        if repository is None or digest is None:
+            raise ValueError("The parameter repository and digest cannot be None.")
+
+        result = self._client.container_registry_blob.get_blob(repository, digest, **kwargs)
+        
+        if not result:
+            return result
+        
+        data = BytesIO(result.response.internal_response.content)
```
annatisch, line 887: Same comment about internal_response.

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_registry_client.py
```diff
@@ -764,3 +760,188 @@ def update_repository_properties(self, *args, **kwargs):
...
+        digest = _compute_digest(data)
```
annatisch, line 888: Is this going to read the entire result stream? Or is this data not actually 'streamed' from the service?
That's possible going to cause other problems - it seems inefficient to 'download' the data twice - how big could it possible get?
Additionally, the connection will be closed after the stream has been read, so unless we're saving that data in memory it will be lost.
Is the digest not returned in a header we could use?

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_registry_client.py
```diff
@@ -764,3 +760,188 @@ def update_repository_properties(self, *args, **kwargs):
...
+            client = ContainerRegistryClient(endpoint, DefaultAzureCredential(), audience="my_audience")
+            client.delete_manifest("my_repository", "my_tag_or_digest")
+        """
+        if _is_tag(tag_or_digest):
+            tag_or_digest = self._get_digest_from_tag(repository, tag_or_digest)
+
+        response = self._client.container_registry.delete_manifest(
+            repository, tag_or_digest, cls=_return_response, **kwargs
+        )
+        return response.http_response
```
annatisch, line 920: We are we returning the http response? This method shouldn't be returning anything.

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_registry_client.py
```diff
@@ -764,3 +760,188 @@ def update_repository_properties(self, *args, **kwargs):
...
+    @distributed_trace
+    def delete_blob(self, repository, tag_or_digest, **kwargs):
+        # type: (str, str, **Any) -> IO
+        """Delete a blob. If the blob cannot be found or a response status code of
+        404 is returned an error will not be raised.
+
+        :param str repository: Name of the repository the manifest belongs to
+        :param str tag_or_digest: Tag or digest of the blob to be deleted
+        :returns: IO
+        :rtype: IO
```
annatisch, line 931: Why would delete_blob return an IO?

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_models.py
```diff
@@ -312,6 +313,60 @@ def repository_name(self):
...
+    def __init__(self, **kwargs):
+        self._data = kwargs.get("data")
+        self._digest = kwargs.get("digest")
+
+    @property
+    def data(self):
+        # type: () -> IO
+        return self._data
+
+    @property
```
annatisch, line 333: Why have we wrapped everything in `@property`?

annatisch (changes requested):

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_registry_client.py
YalinLi0312: * Have changed `stream` to `data`
* We haven't talked about the blob size, I didn't find there's any size check in other languages. I think that's a good question to ask in arch board review(5/10).
* We accept consuming bytes IO instead of bytes here.

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_registry_client.py
YalinLi0312: * Haven't talked about it yet, same as the blob size above.
* No, we passed in the whole stream data to one chunk.

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_registry_client.py
YalinLi0312: This is an error, it doesn't need to pass again. Already addressed.

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_registry_client.py
YalinLi0312: Changed to return a new type [DownloadManifestResult](https://github.com/YalinLi0312/azure-sdk-for-python/blob/yall-acr-oci/sdk/containerregistry/azure-containerregistry/azure/containerregistry/_models.py#L339). Also, changed to return a new type [DownloadBlobResult](https://github.com/YalinLi0312/azure-sdk-for-python/blob/yall-acr-oci/sdk/containerregistry/azure-containerregistry/azure/containerregistry/_models.py#L316) in donwload_blob.

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_registry_client.py
YalinLi0312: Followed other languages, not sure whether will block while the payload is large.

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_registry_client.py
YalinLi0312: Updated.

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_registry_client.py
YalinLi0312: This is an error, don't need to pass data again. Have updated.

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_helpers.py
YalinLi0312: Have removed.

sdk/containerregistry/azure-containerregistry/tests/test_container_registry_client.py
YalinLi0312: Have added download_manifest() call which include get_manifest() after.

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_registry_client.py
YalinLi0312: Now changed to return a new type [DownloadBlobResult](https://github.com/YalinLi0312/azure-sdk-for-python/blob/yall-acr-oci/sdk/containerregistry/azure-containerregistry/azure/containerregistry/_models.py#L316).

sdk/containerregistry/azure-containerregistry/tests/test_container_registry_client.py
YalinLi0312: +1, this is a good point! We can discuss in arch board review. To make it consistent with other languages(they all consume "stream" directly), I'll leave it as it is.

sdk/containerregistry/azure-containerregistry/tests/test_container_registry_client.py
```diff
@@ -603,9 +604,50 @@ def test_get_misspell_property(self, containerregistry_endpoint):
...
+    
+    @acr_preparer()
+    def test_upload_blob(self, containerregistry_endpoint):
+        repo = self.get_resource_name("repo")
+        client = self.create_registry_client(containerregistry_endpoint)
+        
+        blob = "654b93f61054e4ce90ed203bb8d556a6200d5f906cf3eca0620738d6dc18cbed"
+        path = os.path.join(os.getcwd(), "tests", "data", "oci_artifact", blob)
+        
+        client.upload_blob(repo, open(path, "rb"))
```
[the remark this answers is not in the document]
YalinLi0312: Updated.

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_registry_client.py
YalinLi0312: The data in response(response.http_response.internal_response._content) is in bytes.

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_registry_client.py
YalinLi0312: Why python does not prefer returning while it does return in other languages? E.g. [.Net](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Blob/ContainerRegistryBlobClient.cs#L575)

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_registry_client.py
YalinLi0312: It returns from the call in generated call: https://github.com/Azure/azure-sdk-for-python/blob/d265b84c9cac5628b41dd29a0d001a37c9e7d1f5/sdk/containerregistry/azure-containerregistry/azure/containerregistry/_generated/operations/_container_registry_blob_operations.py#L545

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_registry_client.py
YalinLi0312: Yes, upload_blob doesn't need to validate digest.

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_registry_client.py
annatisch: Delete shouldn't return anything - so this seems like a bug in the swagger?

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_registry_client.py
YalinLi0312: - Yes, it is reading the entire result stream in _compute_digest().
- The deserialized data returned from service is in type of StreamDownloadGenerator
- The digest is not returned in headers

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_registry_client.py
YalinLi0312: I'm not sure, [delete manifest](https://github.com/Azure/azure-rest-api-specs/blob/c8d9a26a2857828e095903efa72512cf3a76c15d/specification/containerregistry/data-plane/Azure.ContainerRegistry/stable/2021-07-01/containerregistry.json#L178) and [delete blob](https://github.com/Azure/azure-rest-api-specs/blob/c8d9a26a2857828e095903efa72512cf3a76c15d/specification/containerregistry/data-plane/Azure.ContainerRegistry/stable/2021-07-01/containerregistry.json#L336) both have return in swagger.

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_registry_client.py
YalinLi0312: Have changed to return None.

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_registry_client.py
YalinLi0312: Updated.

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_registry_client.py
pallavit: Today there is no size limitation in the SDK. We assume the bytes given to us our repayable. Our current thinking is to build an experience on re-playability and size management once we get customer asks. (As of now the only customer for the library is Bicep which does not need this support).

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_registry_client.py
pallavit: The service supports both chunked and resumable/range APIs here. We have decided to use chunked w/o range so we basically upload everything as a single blob. Every API call here can be retried. The idea is to scope it currently to only support wasteful upload (every time we start afresh if something goes wrong) however as we get more customers, we can use the range feature and become smart about it.

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_registry_client.py
pallavit: Even though we use 'uploadChunk' as the first scope we are not chunking at all - we are basically doing a monolithic upload of the contents.  So there is no size limitation.

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_registry_client.py
pallavit: Delete calls succeed on other clients on ACR, we should update the 404 to success here as well.

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_models.py
YalinLi0312: Have updated to attributes.

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_registry_client.py
annatisch: This is probably one to consult with the feature crew/service team on.
Normally a delete API wouldn't return anything, because if the result isn't returned successfully (e.g. intermittent network failure), there's no way to retry and the response data is lost.
It would be good to confirm with the service team/feature crew:
- Does the service actually return a response here or is it a bug in the swagger spec
- If so, what is this data and how is a customer expected to use it?

To unblock this PR, I would just return None from a delete for now.

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_registry_client.py
annatisch: Okay - we probably shouldn't block the PR on this one right now - but we will need to dig into this more closely before GA.
It sounds like there's no real limits, expectations around data size and I'm a bit concerned that we're making decisions based on small data sizes that are going to be inefficient or difficult to use if the payload is ever larger. We can open an issue for this for now and big in more deeply before GA.

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_registry_client.py
johanste: From what I can recall from earlier reviews, these APIs _should_ not have returned anything from the server (for exactly the reasons stated above). @annatisch's suggestion is the right way to go.

sdk/containerregistry/azure-containerregistry/tests/test_container_registry_client.py
YalinLi0312: Have discussed in the ACR group, I think
- Creating stream data from a file is pretty trivial, not a burden in Python.
- Consuming IO directly would have more flexibility. It allows customers to call the API after obtaining stream data from somewhere.

So, I would prefer to only support stream as an input data type. Do you still think it's necessary to have separate methods, i.e. upload_manifest_from_file/upload_blob_from_file, that support file path?

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_registry_client.py
YalinLi0312: Thanks for the feedback! Have changed to return None here, and have sent out the email to the contact in service team.

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_registry_client.py
```diff
@@ -764,3 +756,190 @@ def update_repository_properties(self, *args, **kwargs):
...
+
+        Example
+
+        .. code-block:: python
+
+            from azure.containerregistry import ContainerRegistryClient
+            from azure.identity import DefaultAzureCredential
+            endpoint = os.environ["CONTAINERREGISTRY_ENDPOINT"]
+            client = ContainerRegistryClient(endpoint, DefaultAzureCredential(), audience="my_audience")
+            response = client.delete_blob("my_repository", "my_tag_or_digest")
```
mccoyp, line 940: We should remove the response variable from the sample now that the return type is None
```suggestion
            client.delete_blob("my_repository", "my_tag_or_digest")
```

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_generated/aio/operations/_container_registry_blob_operations.py
```diff
@@ -1,17 +1,17 @@
...
-from typing import Any, Callable, Dict, Generic, IO, Optional, TypeVar
-import warnings
+from typing import Any, Callable, Dict, IO, Optional, TypeVar
 
 from azure.core.exceptions import ClientAuthenticationError, HttpResponseError, ResourceExistsError, ResourceNotFoundError, map_error
 from azure.core.pipeline import PipelineResponse
 from azure.core.pipeline.transport import AsyncHttpResponse
 from azure.core.rest import HttpRequest
 from azure.core.tracing.decorator_async import distributed_trace_async
+from azure.core.utils import case_insensitive_dict
```
mccoyp, line 14: `case_insensitive_dict` was only [made public](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/core/azure-core/CHANGELOG.md#1230-2022-03-03) in `azure-core` 1.23.0, so the version in [setup.py](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/containerregistry/azure-containerregistry/setup.py#L58) will need to be updated -- as well as the [override in `shared_requirements.txt`](https://github.com/Azure/azure-sdk-for-python/blob/main/shared_requirements.txt#L136)

sdk/containerregistry/azure-containerregistry/CHANGELOG.md
```diff
@@ -1,13 +1,9 @@
 # Release History
 
-## 1.0.1 (Unreleased)
+## 1.1.0b1 (2022/05/10)
```
mccoyp, line 3: `_version.py` should be updated to match 🙂

sdk/containerregistry/azure-containerregistry/CHANGELOG.md
```diff
@@ -1,16 +1,13 @@
...
```
mccoyp, line 3:
```suggestion
## 1.1.0b1 (2022-05-10)
```

mccoyp (approved):
Things look good to me 🙂 I think it's worth having more discussion about the result of the blob download operation, like Anna said -- whether we should read the full response and return an easy-to-use stream, or improve performance by returning the stream download generator directly -- but I think the current design makes sense for now

YalinLi0312:
> Things look good to me 🙂 I think it's worth having more discussion about the result of the blob download operation, like Anna said -- whether we should read the full response and return an easy-to-use stream, or improve performance by returning the stream download generator directly -- but I think the current design makes sense for now

Thanks for your review @mccoyp !  I leave it to return in stream type for now bc this would be consistent with other languages before we made a decision.

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_registry_client.py
```diff
@@ -764,3 +756,190 @@ def update_repository_properties(self, *args, **kwargs):
...
+        return DownloadManifestResult(digest=digest, data=manifest_stream, manifest=manifest)
+
+    @distributed_trace
+    def download_blob(self, repository, digest, **kwargs):
+        # type: (str, str, **Any) -> DownloadBlobResult | None
+        """Download a blob that is part of an artifact.
+
+        :param str repository: Name of the repository
+        :param str digest: The digest of the blob to download.
+        :returns: DownloadBlobResult or None
```
annatisch, line 874: When/why would this return None?

annatisch (approved):
Approving for preview - though one thing I would like to look into is how the streaming APIs aren't actually streaming anything, and all that data is being held in memory.

PR #24008, opened 2022-04-14
[Cosmos] disable api version check for new pylint requirements

simorenoh:
From https://azure.github.io/azure-sdk/python_design.html#python-client-constructor-api-version-argument-1

These guidelines will now begin to be enforced starting 04/15, so making this PR pre-emptively to unblock Cosmos once this check goes out (since it will block our pylint check in the pipelines). Can't merge this until those changes are out but wanted to get ahead of it.


sdk/cosmos/azure-cosmos/azure/cosmos/aio/cosmos_client.py
```diff
@@ -110,7 +110,7 @@ class CosmosClient(object):
     """
 
     def __init__(self, url, credential, consistency_level=None, **kwargs):
-        # type: (str, Any, Optional[str], Any) -> None
+        # type: (str, Any, Optional[str], Any) -> None # pylint: disable=client-accepts-api-version-keyword
```
tjprescott, line 113: Why not simply accept the api-version kwarg instead of disabling the check?

simorenoh: We generally do not provide our users with the ability to choose which API to go against and would instead prefer to ignore this option, since allowing users to pass in this option could lead to more issues for us to debug for no real reason - going against any version that's not the current one is definitely not recommended

simorenoh:
Date on this got pushed to next week

sdk/cosmos/azure-cosmos/azure/cosmos/aio/cosmos_client.py
johanste: Is the latest API version guaranteed to be available across all cloud instances at the point where a new client library is released? And what backwards compat guarantees is the service making across api versions?

simorenoh: @johanste not sure I understand what you're asking in regards to the SDK - we don't have control over the backend API or what changes they make. If newer versions affect the SDK and breaking changes need to be made, that's a separate issue than the scope of these changes - and only solidifies the fact we don't want users to have access to this setting.

That said, the current version that is to be used for the cosmos python SDK is guaranteed to be available (otherwise it wouldn't be the default).

tjprescott: The standard (hence the guideline--hence the pylint rule) is that all Azure SDKs expose api_version as a parameter as a "best effort" option for customers who want to use the latest SDK with an older API version. Forcing customers to use an older version of the SDK to use an older API version would result in us having to hotfix a bunch of old SDKs for bugs whereas this model allows us to recommend that customers use the latest version of the SDK.

The pylint disable statements are there for false alarms. This is not a false alarm and thus it is hard to approve such a change, though I'll definitely defer to @annatisch and @johanste  on this one.

kushagraThapar: @tjprescott @johanste  - This concept does not applies to Cosmos DB at all. Our SDK has more functionality than just serving as a wrapper for the REST calls - we don't support older API versions in our later SDKs. 

Also when we introduce direct mode for the SDK these changes would even become useless for instance as at that point there is no REST API call. 

This is not something that we do in any of our other SDKs.. (Java, .Net, NodeJS, GO SDK)

kushagraThapar: @tjprescott @johanste  - This concept does not applies to Cosmos DB at all. Our SDK has more functionality than just serving as a wrapper for the REST calls - we don't support older API versions in our later SDKs. 

Also when we introduce direct mode for the SDK these changes would even become useless for instance as at that point there is no REST API call. 

This is not something that we do in any of our other SDKs.. (Java, .Net, NodeJS, GO SDK)

kushagraThapar (dismissed):
This concept does not applies to Cosmos DB at all. Our SDK has more functionality than just serving as a wrapper for the REST calls - we don't support older API versions in our later SDKs.

Also when we introduce direct mode for the SDK these changes would even become useless for instance as at that point there is no REST API call.

This is not something that we do in any of our other SDKs.. (Java, .Net, NodeJS, GO SDK)

sdk/cosmos/azure-cosmos/azure/cosmos/http_constants.py
```diff
@@ -269,9 +269,30 @@ class CookieHeaders(object):
 class Versions(object):
     """Constants of versions.
     """
-    CurrentVersion = "2018-12-31"
-    SDKName = "azure-cosmos"
-    QueryVersion = "1.0"
+    current_version = "2018-12-31"
+    SDK_name = "azure-cosmos"
+    query_version = "1.0"
+    api_versions = [
```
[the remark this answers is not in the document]
j82w: We should only expose the version the python SDK targets. Supporting any other version will likely break the SDK.

sdk/cosmos/azure-cosmos/azure/cosmos/aio/cosmos_client.py
ealsur: Agree. Cosmos SDK due to their "thickness" contain logic that is compatible with the current service APIs released in prod. Before any new service API release, the team performs validations that the current existing SDKs will still function properly with any new service release. Some features in the current version in prod on the service might map to some SDK capability that if the customer decides to change the `api_version` would be broken. Breaking the SDK by setting a random lower `api_version` makes little sense to me.
Cosmos DB has no guidance anywhere that guides users to set X or Y `api_version` nor there is any need for such a guidance as far as I understand.

sdk/cosmos/azure-cosmos/azure/cosmos/http_constants.py
```diff
@@ -269,9 +269,18 @@ class CookieHeaders(object):
...
+        "2018-06-18",  # multi-master and partition migration
+        "2018-08-31",  # throughput-split
+        "2018-09-17",  # IndexV2/ updated policy (composite and additional spatial indexes)
+        "2018-12-31",  # non-partitioned collection deprecation and migration of existing collections
+        "2019-10-14",  # Merge support in SDK
```
j82w, line 280: What is the point having these versions when the SDK does not support them?

simorenoh: @j82w Thinking from the point of view of someone who joined the SDK without much guidance - if in the future we want to see which api versions support different changes (specially for future versions, once the SDK is merge proof for instance), having the versions handy within the SDK seems like a good move (rather than going to the .NET SDK to check these).

The older versions I have no issues with removing, but the newer ones I would argue for keeping for the case I outlined before, and regardless we're not giving users a way to use these versions.

I guess the question would be if there is some public/ microsoft internal document where a new Cosmos developer would be able to find these, and if so then this would not be needed.

sdk/cosmos/azure-cosmos/azure/cosmos/http_constants.py
```diff
@@ -269,9 +269,18 @@ class CookieHeaders(object):
 class Versions(object):
```
johanste, line 269: Is this a public class? If so we should not make breaking changes just to make the linter happy.

johanste (changes requested):
Let's make sure we are not doing breaking changes to make the linter happy. If this is a change to something that hasn't already been released, then it is goodness...

sdk/cosmos/azure-cosmos/azure/cosmos/http_constants.py
simorenoh: Removed this to move forward, will document later on if needed.

tjprescott (approved):
Tracking issue for this topic here: https://github.com/Azure/azure-sdk-for-python/issues/24094

PR #24009, opened 2022-04-14
[Storage] Fix some datalake cpk Live tests

jalauzon-msft:
Some new datalake cpk tests started to fail in the Live test pipeline due to improper conditions when uploading data to files. I believe these tests passed during development of the feature since I was not using an HNS enabled account to run the tests.


vincenttran-msft (approved):
LGTM 🥇

PR #24010, opened 2022-04-14
Update network loss schema

ckairen:
Newest Chaos Mesh schema for "scheduler" now called [scheduling rules](https://chaos-mesh.org/docs/2.0.7/define-scheduling-rules/)

### DIFF
--- a/sdk/servicebus/azure-servicebus/stress/templates/network_loss.yaml
+++ b/sdk/servicebus/azure-servicebus/stress/templates/network_loss.yaml
@@ -1,14 +1,7 @@
+{{- include "stress-test-addons.chaos-wrapper.tpl" (list . "stress.python-sb-network") -}}
+{{- define "stress.python-sb-network" -}}
 apiVersion: chaos-mesh.org/v1alpha1
 kind: Schedule
-metadata:
-  name: '{{ .Release.Name }}-{{ .Release.Revision }}'
-  namespace: {{ .Release.Namespace }}
-  annotations:
-    experiment.chaos-mesh.org/pause: 'true'
-  labels:
-    scenario: 'stress'
-    release: '{{ .Release.Name }}'
-    revision: '{{ .Release.Revision }}'
 spec:
   schedule: '@every 30s'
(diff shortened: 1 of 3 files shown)


sdk/servicebus/azure-servicebus/stress/templates/network_loss.yaml
```diff
@@ -1,5 +1,5 @@
 apiVersion: chaos-mesh.org/v1alpha1
-kind: NetworkChaos
+kind: Schedule
```
benbp, line 2: If you don't mind while you're in this file, can you update it to use the new [chaos wrapper](https://github.com/Azure/azure-sdk-tools/blob/main/tools/stress-cluster/cluster/kubernetes/stress-test-addons/templates/_chaos.tpl)? This should replace the metadata and annotations with the addons boilerplate version for consistency.

benbp: Also @richardpark-msft @m-redding FYI the chaos resources are migrating to the `Schedule` kind going forward, since that's how it's structured in chaos mesh 2.x

sdk/servicebus/azure-servicebus/stress/templates/network_loss.yaml
```diff
@@ -1,31 +1,26 @@
...
+        testInstance: "servicebus-{{ .Release.Name }}-{{ .Release.Revision }}"
+        chaos: 'true'
+      namespaces:
+        - {{ .Release.Namespace }}
+      podPhaseSelectors:
+        - 'Running'
+    duration: '10s'
+    direction: to
+    externalTargets:
+      - 'servicebus.windows.net'
```
benbp, line 22: As we found, this doesn't work. Maybe in this update or a subsequent one, we can update this and the ARM template to use the same `{{ .Stress.ResourceGroupName }}` value?

richardpark-msft: CC: @swathipil

benbp (approved):

Issue #23613, opened 2022-03-23
Move multi-tenant authentication-supporting code into azure-core


mccoyp:
**Context:** Key Vault currently supports multi-tenant authentication during tenant discovery, in which the home tenant of a resource is returned by the service during challenge authentication. The resource's tenant is included as part of an [authorization URI](https://github.com/Azure/azure-sdk-for-python/blob/be9bc2ebd94b204d4939b848ec9b23688eeb12c3/sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_shared/http_challenge.py#L44-L46) returned by the service -- in KV, this tenant ID is parsed and [given to token requests](https://github.com/Azure/azure-sdk-for-python/blob/0a0cc97f178a7476ec79f29c090b8c93ad5d4955/sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_shared/challenge_auth_policy.py#L102). This allows a user to authorize requests to a resource even if the credential provided to the current client was constructed to authenticate in a different tenant.

Other services want to support this scenario, and the challenge auth flow should be service-agnostic since this is an AAD feature. So, it makes sense to support multi-tenant auth with an `azure-core` authentication policy instead of having each service use custom code to parse out and use the returned tenant ID. This issue tracks moving relevant code out of KV and into Core to support this.

mccoyp:
Closing because having a unified policy would be difficult across services; behavior can differ significantly between different service implementations (this feedback was provided in an archboard review of a Core policy proposal).

PR #24019, opened 2022-04-15
[Core] Add multi-tenant authentication policy

mccoyp:
# Description

Resolves https://github.com/Azure/azure-sdk-for-python/issues/23613. For cross-language reference, [here](https://github.com/Azure/azure-sdk-for-net/blob/1722fc8329bd848d400faf73daeba06d9d718cc1/sdk/tables/Azure.Data.Tables/src/TableBearerTokenChallengeAuthorizationPolicy.cs#L28) is .NET's Tables-internal implementation, [here](https://github.com/Azure/azure-sdk-for-java/blob/main/sdk/tables/azure-data-tables/src/main/java/com/azure/data/tables/implementation/TableBearerTokenChallengeAuthorizationPolicy.java) is Java's Tables-internal implementation, and [here](https://github.com/Azure/azure-sdk-for-js/blob/bf7f3c18b6e43272b8a6db0defccd09d551923d9/sdk/core/core-client/src/authorizeRequestOnTenantChallenge.ts) is JavaScript's challenge-handling callback method in Core.

**Context:** AAD supports [authentication claims challenges](https://docs.microsoft.com/azure/active-directory/develop/claims-challenge), which are exposed in service requests as 401 responses when authentication fails. These 401 responses contain a header that includes the tenant ID where the requested resource lives -- this tenant ID can be passed along to token requests in order to get a valid token for the resource, even if the credential we initially provided to the client targeted a different tenant.

For example:

```python
from azure.data.tables import TableServiceClient
from azure.identity import ClientSecretCredential

endpoint = "https://storage-account.table.core.windows.net"
credential = ClientSecretCredential(
    tenant_id=tenant_A,  # our storage account lives in tenant_B
    client_id=...,
    client_secret=...,
)

# Note: Tables only supports auth challenges in API version 2020-12-06 and after (this will require new Tables SDK support)
client = TableServiceClient(credential=credential, endpoint=endpoint, api_version="2020-12-06")
for table in client.list_tables():
    print(table.name)
```

**Current behavior:** we call `client.list_tables()` with a `BearerTokenCredentialPolicy`:
1. We fetch a token for `tenant_A` because of our credential
2. The service rejects this token because our storage resource is in `tenant_B`
3. We return the error response in an exception

**New behavior:** we call `client.list_tables()` with a `BearerTokenChallengePolicy` (from this PR):
1. We fetch a token for `tenant_A` because of our credential
2. The service rejects this token because our storage resource is in `tenant_B`
3. We parse the `WWW-Authenticate` header of the response and fetch a token valid in `tenant_B` (for the scope provided in the same header)
4. We re-send the original request, with the new token for `tenant_B`
5. We get a successful response from the service

This policy is based on the [ChallengeAuthPolicy](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/keyvault/azure-keyvault-secrets/azure/keyvault/secrets/_shared/challenge_auth_policy.py) used by Key Vault, which currently supports multi-tenant authentication. The most significant change between this policy and KV's is that this policy doesn't use its `ChallengeCache` to remove the request body when a request is expected to prompt an auth challenge. Changing this behavior should reduce the number of requests we make:

1. If the resource we're making a request for exists in our tenant, we already provide a valid token. The request succeeds.
  1a. In KV, we would send an empty request the first time in order to prompt an auth challenge, even though our tenant matches that of the resource. Our second request succeeds.
2. If the resource exists in a different tenant, our request fails and prompts a 401. We make a second request with the correct tenant.
  2a. In KV, we would have initially sent an empty request to prompt a 401. We still send a second request with the correct tenant.
  
For context, Key Vault drops the body of requests made to a new endpoint for security reasons: we want to ensure that the endpoint we're communicating with can accept our auth flow; that it is a correct KV endpoint (at least, that it follows the expected challenge auth flow); and that we don't send sensitive information (like key or secret values) in request bodies until the former conditions are met. Tables doesn't follow these security requirements today, so I decided to leave request bodies in initial requests by default -- KV's use of the policy will just have to be modified. This is prototyped in https://github.com/mccoyp/azure-sdk-for-python/pull/1.
  
  
By default, this policy will fetch tokens after a challenge response that are valid for the tenant and scope we get back in a challenge response (and _only_ the scope provided in the response). There are two toggles to disable either part or all of this behavior: `discover_tenant` and `discover_scopes`. If both of these kwargs are set to `False`, this policy becomes functionally equivalent to the base `BearerTokenCredentialPolicy`.

d

### DIFF
--- a/sdk/core/azure-core/azure/core/pipeline/policies/__init__.py
+++ b/sdk/core/azure-core/azure/core/pipeline/policies/__init__.py
@@ -25,7 +25,12 @@
 # --------------------------------------------------------------------------
 
 from ._base import HTTPPolicy, SansIOHTTPPolicy, RequestHistory
-from ._authentication import BearerTokenCredentialPolicy, AzureKeyCredentialPolicy, AzureSasCredentialPolicy
+from ._authentication import (
+    AzureKeyCredentialPolicy,
+    AzureSasCredentialPolicy,
+    BearerTokenCredentialPolicy,
+    MultitenantTokenCredentialPolicy,
+)
 from ._custom_hook import CustomHookPolicy
 from ._redirect import RedirectPolicy
 from ._retry import RetryPolicy, RetryMode
@@ -43,9 +48,10 @@ from ._universal import (
 __all__ = [
     'HTTPPolicy',
     'SansIOHTTPPolicy',
-    'BearerTokenCredentialPolicy',
     'AzureKeyCredentialPolicy',
     'AzureSasCredentialPolicy',
+    'BearerTokenCredentialPolicy',
+    'MultitenantTokenCredentialPolicy',
     'HeadersPolicy',
     'UserAgentPolicy',
     'NetworkTraceLoggingPolicy',
--- a/sdk/core/azure-core/azure/core/pipeline/policies/__init__.py
+++ b/sdk/core/azure-core/azure/core/pipeline/policies/__init__.py
@@ -29,7 +29,7 @@ from ._authentication import (
     AzureKeyCredentialPolicy,
     AzureSasCredentialPolicy,
     BearerTokenCredentialPolicy,
-    MultitenantCredentialPolicy,
+    BearerTokenChallengePolicy,
 )
 from ._custom_hook import CustomHookPolicy
 from ._redirect import RedirectPolicy
@@ -51,7 +51,7 @@ __all__ = [
     'AzureKeyCredentialPolicy',
     'AzureSasCredentialPolicy',
     'BearerTokenCredentialPolicy',
-    'MultitenantCredentialPolicy',
+    'BearerTokenChallengePolicy',
     'HeadersPolicy',
     'UserAgentPolicy',
     'NetworkTraceLoggingPolicy',
@@ -71,13 +71,13 @@ __all__ = [
 
 try:
     from ._base_async import AsyncHTTPPolicy
-    from ._authentication_async import AsyncBearerTokenCredentialPolicy, AsyncMultitenantCredentialPolicy
+    from ._authentication_async import AsyncBearerTokenCredentialPolicy, AsyncBearerTokenChallengePolicy
     from ._redirect_async import AsyncRedirectPolicy
     from ._retry_async import AsyncRetryPolicy
     __all__.extend([
         'AsyncHTTPPolicy',
         'AsyncBearerTokenCredentialPolicy',
-        'AsyncMultitenantCredentialPolicy',
+        'AsyncBearerTokenChallengePolicy',
         'AsyncRedirectPolicy',
         'AsyncRetryPolicy'
     ])
--- a/sdk/core/azure-core/azure/core/pipeline/policies/__init__.py
+++ b/sdk/core/azure-core/azure/core/pipeline/policies/__init__.py
@@ -29,7 +29,7 @@ from ._authentication import (
     AzureKeyCredentialPolicy,
     AzureSasCredentialPolicy,
     BearerTokenCredentialPolicy,
-    MultitenantTokenCredentialPolicy,
+    MultitenantCredentialPolicy,
 )
 from ._custom_hook import CustomHookPolicy
 from ._redirect import RedirectPolicy
@@ -51,7 +51,7 @@ __all__ = [
     'AzureKeyCredentialPolicy',
     'AzureSasCredentialPolicy',
     'BearerTokenCredentialPolicy',
-    'MultitenantTokenCredentialPolicy',
+    'MultitenantCredentialPolicy',
     'HeadersPolicy',
     'UserAgentPolicy',
     'NetworkTraceLoggingPolicy',
@@ -71,12 +71,13 @@ __all__ = [
 
 try:
     from ._base_async import AsyncHTTPPolicy
-    from ._authentication_async import AsyncBearerTokenCredentialPolicy
+    from ._authentication_async import AsyncBearerTokenCredentialPolicy, AsyncMultitenantCredentialPolicy
     from ._redirect_async import AsyncRedirectPolicy
     from ._retry_async import AsyncRetryPolicy
     __all__.extend([
         'AsyncHTTPPolicy',
         'AsyncBearerTokenCredentialPolicy',
+        'AsyncMultitenantCredentialPolicy',
         'AsyncRedirectPolicy',
         'AsyncRetryPolicy'
     ])
--- a/sdk/core/azure-core/azure/core/pipeline/policies/_authentication.py
+++ b/sdk/core/azure-core/azure/core/pipeline/policies/_authentication.py
@@ -22,7 +22,7 @@ if TYPE_CHECKING:
     from azure.core.pipeline import PipelineRequest, PipelineResponse
 
 
-class _HttpChallenge(object):  # pylint:disable=too-few-public-methods
+class HttpChallenge(object):  # pylint:disable=too-few-public-methods
     """Represents a parsed HTTP WWW-Authentication Bearer challenge from a server."""
 
     def __init__(self, challenge):
@@ -255,7 +255,7 @@ class BearerTokenChallengePolicy(BearerTokenCredentialPolicy):
             return False
 
         try:
-            challenge = _HttpChallenge(response.http_response.headers.get("WWW-Authenticate"))
+            challenge = HttpChallenge(response.http_response.headers.get("WWW-Authenticate"))
             # azure-identity credentials require an AADv2 scope but the challenge may specify an AADv1 resource
             # if no scopes are included in the challenge, challenge.scope and challenge.resource will both be ''
             scope = challenge.scope or challenge.resource + "/.default" if self._discover_scopes else self._scopes
--- a/sdk/core/azure-core/azure/core/pipeline/policies/_authentication.py
+++ b/sdk/core/azure-core/azure/core/pipeline/policies/_authentication.py
@@ -4,8 +4,8 @@
 # license information.
 # -------------------------------------------------------------------------
 import time
-import six
 import urllib.parse
+import six
 
 from . import HTTPPolicy, SansIOHTTPPolicy
 from ...exceptions import ServiceRequestError
--- a/sdk/core/azure-core/azure/core/pipeline/policies/_authentication.py
+++ b/sdk/core/azure-core/azure/core/pipeline/policies/_authentication.py
@@ -214,7 +214,7 @@ class BearerTokenCredentialPolicy(_BearerTokenCredentialPolicyBase, HTTPPolicy):
         return
 
 
-class MultitenantCredentialPolicy(BearerTokenCredentialPolicy):
+class BearerTokenChallengePolicy(BearerTokenCredentialPolicy):
     """Adds a bearer token Authorization header to requests, for the tenant provided in authentication challenges.
 
     See https://docs.microsoft.com/azure/active-directory/develop/claims-challenge for documentation on AAD
--- a/sdk/core/azure-core/azure/core/pipeline/policies/_authentication.py
+++ b/sdk/core/azure-core/azure/core/pipeline/policies/_authentication.py
@@ -25,22 +25,22 @@ if TYPE_CHECKING:
 class _HttpChallenge(object):  # pylint:disable=too-few-public-methods
     """Represents a parsed HTTP WWW-Authentication Bearer challenge from a server."""
 
-    def __init__(self, request_uri, challenge):
-        self.source_authority = self._get_request_authority(request_uri)
-        self.source_uri = request_uri
+    def __init__(self, challenge):
+        if not challenge:
+            raise ValueError("Challenge cannot be empty")
+
         self._parameters = {}
 
-        # Get the scheme of the challenge and remove it from the challenge string
-        trimmed_challenge = self._get_challenge_parts(challenge)
+        # Split the scheme ("Bearer") from the challenge parameters
+        trimmed_challenge = challenge.strip()
         split_challenge = trimmed_challenge.split(" ", 1)
-        self.scheme = split_challenge[0]
         trimmed_challenge = split_challenge[1]
 
         # Split trimmed challenge into name=value pairs; these pairs are expected to be split by either commas or spaces
         # Values may be surrounded by quotes, which are stripped here
         separator = "," if "," in trimmed_challenge else " "
         for item in trimmed_challenge.split(separator):
-            # process name=value pairs
+            # Process 'name=value' pairs
             comps = item.split("=")
             if len(comps) == 2:
                 key = comps[0].strip(' "')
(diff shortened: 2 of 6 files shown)


sdk/core/azure-core/azure/core/pipeline/policies/_authentication.py
```diff
@@ -172,6 +252,41 @@ def on_exception(self, request):
...
+        :returns: a bool indicating whether the policy should send the request
+        """
+        try:
+            challenge = _HttpChallenge(request.http_request.url, response.http_response.headers.get("WWW-Authenticate"))
+            # azure-identity credentials require an AADv2 scope but the challenge may specify an AADv1 resource
+            scope = challenge.get_scope() or challenge.get_resource() + "/.default"
+        except ValueError:
+            return False
+
+        body = request.context.pop("key_vault_request_data", None)
```
xiangyan99, line 283: Is it only for key vault?

sdk/core/azure-core/azure/core/pipeline/policies/__init__.py
```diff
@@ -43,9 +48,10 @@
 __all__ = [
     'HTTPPolicy',
     'SansIOHTTPPolicy',
-    'BearerTokenCredentialPolicy',
     'AzureKeyCredentialPolicy',
     'AzureSasCredentialPolicy',
+    'BearerTokenCredentialPolicy',
+    'MultitenantTokenCredentialPolicy',
```
xiangyan99, line 54: Do we need an async one?

mccoyp: We will! That is on its way, along with unit tests 🙂

xiangyan99: And you will have a separate PR for it or will merge it into this PR?

sdk/core/azure-core/azure/core/pipeline/policies/_authentication.py
mccoyp: This is a general-purpose policy that will work for Key Vault, Tables, and any other service that opts in to challenge authentication with tenant discovery. This KV-specific bit is coming from the current KV implementation, but I may be able to remove this if it turns out to be unnecessary. For non-KV scenarios it wouldn't cause an error, and would just be superfluous.

sdk/core/azure-core/azure/core/pipeline/policies/__init__.py
mccoyp: That will be part of this PR

sdk/core/azure-core/azure/core/pipeline/policies/_authentication.py
mccoyp: Update: this did turn out to be unnecessary, and was part of KV's strategy to remove request information before intentionally triggering a 401. The original body content would be removed and stored in the request context, and then brought back in for the authenticated request. Since we're no longer making an initial, empty request to trigger 401s, we can remove this.

sdk/core/azure-core/azure/core/pipeline/policies/_authentication.py
```diff
@@ -21,6 +22,85 @@
...
+    def _get_request_authority(self, uri):
+        """ Extracts the host authority from the given URI. """
+        if not uri:
+            raise ValueError("request_uri cannot be empty")
+
+        uri = urllib.parse.urlparse(uri)
+        if not uri.netloc:
+            raise ValueError("request_uri must be an absolute URI")
+
+        if uri.scheme.lower() not in ["http", "https"]:
```
annatisch, line 98: Realistically would it ever not be one of these? And if it isn't... something probably would have failed sooner?
Not really requesting any changes, but I'm curious if this validation was actually driven by a particular scenario?

sdk/core/azure-core/azure/core/pipeline/policies/_authentication.py
```diff
@@ -172,6 +252,55 @@ def on_exception(self, request):
...
+    :param str scopes: Lets you specify the type of access needed.
+    :keyword bool enable_tenant_discovery: Determines if tenant discovery should be enabled. Defaults to True.
+    :keyword bool enable_scopes_discovery: Determines if scopes from authentication challenges should be provided to
+        token requests, instead of the scopes given to the policy's constructor. Defaults to True.
+    :raises: :class:`~azure.core.exceptions.ServiceRequestError`
+    """
+
+    def __init__(self, credential: "TokenCredential", *scopes: str, **kwargs: "Any") -> None:
+        self._discover_tenant = kwargs.pop("enable_tenant_discovery", True)
+        self._discover_scopes = kwargs.pop("enable_scopes_discovery", True)
```
annatisch, line 272: Have we used the `enable_` prefix elsewhere in credentials/identity?
We could potentially use something shorter like `discover_scopes`/`discover_tenant`. But if we've used `enable_` elsewhere we can stick with the pattern.

sdk/core/azure-core/azure/core/pipeline/policies/_authentication.py
```diff
@@ -172,6 +252,55 @@ def on_exception(self, request):
...
+            scope = challenge.get_scope() or challenge.get_resource() + "/.default"
+        except ValueError:
+            return False
+
+        if self._discover_tenant:
+            self.authorize_request(
+                request, scope if self._discover_scopes else self._scopes, tenant_id=challenge.tenant_id
+            )
+        else:
+            if self._discover_scopes:
```
annatisch, line 296: If looks like `discover_scopes` will only be used is `discover_tenant` is false. But they're both true by default?

sdk/core/azure-core/azure/core/pipeline/policies/_authentication.py
```diff
@@ -21,6 +22,85 @@
...
+        for item in trimmed_challenge.split(separator):
+            # process name=value pairs
+            comps = item.split("=")
+            if len(comps) == 2:
+                key = comps[0].strip(' "')
+                value = comps[1].strip(' "')
+                if key:
+                    self._parameters[key] = value
+
+        if not self._parameters:
```
johanste, line 51: Not needed since you are looking for the `authorization` or `authorization_uri` on line 55.

sdk/core/azure-core/azure/core/pipeline/policies/_authentication.py
```diff
@@ -21,6 +22,85 @@
...
+            raise ValueError("Invalid challenge parameters")
+
+        # Challenge must specify authorization or authorization_uri
+        if "authorization" not in self._parameters and "authorization_uri" not in self._parameters:
+            raise ValueError("Invalid challenge parameters")
```
johanste, line 56: I would add a comment specifying why the challenge parameters were invalid (which keys I/we expected)

sdk/core/azure-core/azure/core/pipeline/policies/_authentication.py
```diff
@@ -21,6 +22,85 @@
...
+
+        authorization_uri = self.get_authorization_server()
+        # the authorization server URI should look something like https://login.windows.net/tenant-id[/oauth2/authorize]
+        uri_path = urllib.parse.urlparse(authorization_uri).path.lstrip("/")
+        self.tenant_id = uri_path.split("/")[0] or None
+
+    def get_authorization_server(self):
```
johanste, line 63: Is this method supposed to be internal-public?

sdk/core/azure-core/azure/core/pipeline/policies/_authentication.py
```diff
@@ -21,6 +22,85 @@
...
+        """ Returns the URI for the authorization server if present, otherwise an empty string. """
+        value = ""
+        for key in ["authorization_uri", "authorization"]:
+            value = self._parameters.get(key) or ""
+            if value:
+                break
+        return value
+
+    def get_resource(self):
```
johanste, line 72: Why is this a function and not a property?

sdk/core/azure-core/azure/core/pipeline/policies/_authentication.py
johanste: Why `kwargs.pop()` with default value and not keyword-only arguments?

sdk/core/azure-core/azure/core/pipeline/policies/_authentication.py
```diff
@@ -21,6 +22,85 @@
...
+        self._parameters = {}
+
+        # Get the scheme of the challenge and remove it from the challenge string
+        trimmed_challenge = self._get_challenge_parts(challenge)
+        split_challenge = trimmed_challenge.split(" ", 1)
+        self.scheme = split_challenge[0]
+        trimmed_challenge = split_challenge[1]
+
+        # Split trimmed challenge into name=value pairs; these pairs are expected to be split by either commas or spaces
+        # Values may be surrounded by quotes, which are stripped here
```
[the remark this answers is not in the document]
mccoyp: They do use different separating characters: Key Vault uses commas and quote-surrounded values, but Tables uses spaces and no quotes. This works for both formats, though only the first is described in [AAD docs](https://docs.microsoft.com/azure/active-directory/develop/claims-challenge#claims-challenge-header-format)

sdk/core/azure-core/azure/core/pipeline/policies/_authentication.py
mccoyp: This was intended to avoid a `NoneType` iteration error on line 55, but I can do that on one line instead

sdk/core/azure-core/azure/core/pipeline/policies/_authentication.py
mccoyp: This (and other methods on the class) came directly from Key Vault's implementation of a [similar class](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_shared/http_challenge.py). This doesn't need to be public -- I can make this private

sdk/core/azure-core/azure/core/pipeline/policies/_authentication.py
mccoyp: For this and `get_scope`, again an instance of a decision made based on precedence. I can definitely make this a property instead

sdk/core/azure-core/azure/core/pipeline/policies/_authentication.py
mccoyp: This was written [three years ago](https://github.com/Azure/azure-sdk-for-python/blame/be9bc2ebd94b204d4939b848ec9b23688eeb12c3/sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_shared/http_challenge.py#L107) for Key Vault, so I doubt even Charles would remember at this point... I can't personally think of a real scenario where this would come up, though.

sdk/core/azure-core/azure/core/pipeline/policies/_authentication.py
mccoyp: `discover_scopes` is also used on line 293 with an in-line `if` statement -- they are both true by default. They would only both be `False` if someone explicitly set them that way, which I would imagine would be the case if they wanted or needed to restrict any token requests to only use the information provided to the credential and policy. Having them both be `False` is equivalent to using the basic `BearerTokenCredentialPolicy`, so realistically, I would only see that happening if they wanted to always use this policy but toggle these options depending on something like an API version at runtime

sdk/core/azure-core/azure/core/pipeline/policies/_authentication.py
mccoyp: On second thought, I'm removing this function entirely. Doing this on line 58 is easier to read and more efficient

sdk/core/azure-core/azure/core/pipeline/policies/__init__.py
```diff
@@ -25,7 +25,12 @@
 # --------------------------------------------------------------------------
 
 from ._base import HTTPPolicy, SansIOHTTPPolicy, RequestHistory
-from ._authentication import BearerTokenCredentialPolicy, AzureKeyCredentialPolicy, AzureSasCredentialPolicy
+from ._authentication import (
+    AzureKeyCredentialPolicy,
+    AzureSasCredentialPolicy,
+    BearerTokenCredentialPolicy,
+    MultitenantTokenCredentialPolicy,
```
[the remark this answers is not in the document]
mccoyp: AAD docs use "Multitenant", but there was (at one point) an option in .NET's Identity library for [AllowMultiTenantAuthentication](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/identity/Azure.Identity/CHANGELOG.md#breaking-changes-from-150-beta4). This never got released in a stable version though, so I don't know if there was any architect input on the naming.

Technically, .NET's policy name for this feature is [TableBearerTokenChallengeAuthorizationPolicy](https://github.com/Azure/azure-sdk-for-net/blob/1722fc8329bd848d400faf73daeba06d9d718cc1/sdk/tables/Azure.Data.Tables/src/TableBearerTokenChallengeAuthorizationPolicy.cs#L28) -- obviously that's Tables-specific, but it does suggest the idea of something like `BearerTokenChallengePolicy`. I don't think that's a bad name either, and honestly, I think using "Challenge" instead of "Multitenant" makes sense. The policy is specifically about handling auth challenges, and can be used for more (or less than) cross-tenant auth.

sdk/core/azure-core/azure/core/pipeline/policies/_authentication.py
```diff
@@ -172,6 +214,60 @@ def on_exception(self, request):
...
+        :returns: a bool indicating whether the policy should send the request
+        """
+        if not self._discover_tenant and not self._discover_scopes:
+            # We can't discover the tenant or use a different scope; the request will fail because it hasn't changed
+            return False
+
+        try:
+            challenge = _HttpChallenge(response.http_response.headers.get("WWW-Authenticate"))
+            # azure-identity credentials require an AADv2 scope but the challenge may specify an AADv1 resource
+            scope = challenge.scope or challenge.resource + "/.default" if self._discover_scopes else self._scopes
```
annatisch, line 260: It looks like there's a possibility the neither `challenge.scope` _or_ `challenge.resource` are populated, in which case I think we'd just be sending "/.default" if `discover_scopes` is true. In this scenario would sending that be the best/correct thing to do and let it fail at the service?
Or should we be falling back to `self._scopes` in the case that the information in the challenge was insufficient?

sdk/core/azure-core/azure/core/pipeline/policies/__init__.py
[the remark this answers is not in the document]
annatisch: I would be okay with something like BeaerTokenChallenge[Credential]Policy.
I agree I think it makes more sense than Multitenant - as that's more of an outcome rather than a specific part of the design.

sdk/core/azure-core/azure/core/pipeline/policies/_authentication.py
mccoyp: That's a good point -- I was thinking that we could let the invalid scope error during the service request in that case, but I don't think it would be a useful error experience. We would get something like
`azure.core.exceptions.ClientAuthenticationError: DefaultAzureCredential failed to retrieve a token from the included credentials.`
with further details of
`Authentication failed: AADSTS70011: The provided request must include a 'scope' input parameter. The provided value for the input parameter 'scope' is not valid. The scope /.default is not valid.`

That would indicate that there's an issue with scopes, but a user would have to understand what's going on under the hood to figure out that they should set `discover_scopes` to False. I guess the question is, when we detect that `challenge.scope` and `challenge.resource` are empty, should we:
1. Raise an error with a useful message? Or,
2. Default to using `self._scopes`?

mccoyp: For now, I'll go with the latter approach (and make this behavior clear in the class docstring)

sdk/core/azure-core/azure/core/pipeline/policies/_authentication.py
```diff
@@ -21,6 +22,47 @@
     from azure.core.pipeline import PipelineRequest, PipelineResponse
 
 
+class _HttpChallenge(object):  # pylint:disable=too-few-public-methods
```
xiangyan99, line 25: HttpChallenge?

mccoyp: As in, remove the underscore? I can do that -- wasn't sure if we should since this class isn't exposed publicly, but it's imported in `_authentication_async.py` as well so that sounds reasonable

xiangyan99: It is in_authentication.py (has underscore) and used in another file.

Removing the underscore from the class name may be more reasonable :)

mccoyp:
Closing this PR in favor of a new, forthcoming PR that incorporates feedback from the review in https://github.com/Azure/azure-sdk/issues/4302.

PR #24029, opened 2022-04-15
[metricsadvisor] remove rename imports

iscai-msft:
create new wrapped models for renamed impots from generated code for @l0lawrence 's new pylint checker


kristapratico (approved):
thank you!

PR #24031, opened 2022-04-15
[Storage] Update documentation for socket timeouts

### LINKED ISSUE #23049: Wrong information about `connection_timeout`
Docs:

> connection_timeout (int): Optionally sets the connect and read timeout value, in seconds.

Actually, there is a separate `read_timeout` parameter and `connection_timeout` has no influence over the read timeout.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: f60c47f1-e4d8-f68e-5f9c-460c4fdc6da8
* Version Independent ID: 172e9f9b-25fb-2121-cb72-d383cc21763a
* Content: [Azure Storage Blobs client library for Python](https://docs.microsoft.com/en-us/python/api/overview/azure/storage-blob-readme?view=azure-python)
* Content Source: [docs-ref-services/latest/storage-blob-readme.md](https://github.com/MicrosoftDocs/azure-docs-sdk-python/blob/master/docs-ref-services/latest/storage-blob-readme.md)
* Service: **storagestorage**
* Product: **azure**
* Technology: **azure**
* GitHub Login: @ramya-rao-a
* Microsoft Alias: **ramyar**

jalauzon-msft:
Resolves #23049.

Updating documentation for different socket timeouts. Also took the opportunity to remove Python 2 disclaimer and update timeout constants since we no longer support Python 3.5.

### DIFF
--- a/sdk/storage/azure-storage-blob/README.md
+++ b/sdk/storage/azure-storage-blob/README.md
@@ -1,6 +1,3 @@
-## _Disclaimer_
-_Azure SDK Python packages support for Python 2.7 has ended 01 January 2022. For more information and questions, please refer to https://github.com/Azure/azure-sdk-for-python/issues/20691_
-
 # Azure Storage Blobs client library for Python
 Azure Blob storage is Microsoft's object storage solution for the cloud. Blob storage is optimized for storing massive amounts of unstructured data, such as text or binary data.
 
@@ -347,7 +344,8 @@ Other optional configuration keyword arguments that can be specified on the clie
 
 **Client keyword arguments:**
 
-* __connection_timeout__ (int): Optionally sets the connect and read timeout value, in seconds.
+* __connection_timeout__ (int): The number of seconds the client will wait to establish a connection to the server.
+* __read_timeout__ (int): The number of seconds the client will wait, after the connections has been established, for the server to send a response.
 * __transport__ (Any): User-provided transport to send the HTTP request.
 
 **Per-operation keyword arguments:**
--- a/sdk/storage/azure-storage-blob-changefeed/README.md
+++ b/sdk/storage/azure-storage-blob-changefeed/README.md
@@ -1,6 +1,3 @@
-## _Disclaimer_
-_Azure SDK Python packages support for Python 2.7 has ended 01 January 2022. For more information and questions, please refer to https://github.com/Azure/azure-sdk-for-python/issues/20691_
-
 # Azure Storage Blob ChangeFeed client library for Python
 
 This preview package for Python enables users to get blob change feed events. These events can be lazily generated, iterated by page, retrieved for a specific time interval, or iterated from a specific continuation token.
--- a/sdk/storage/azure-storage-blob/azure/storage/blob/_shared/constants.py
+++ b/sdk/storage/azure-storage-blob/azure/storage/blob/_shared/constants.py
@@ -4,23 +4,14 @@
 # license information.
 # --------------------------------------------------------------------------
 
-import sys
 from .._serialize import _SUPPORTED_API_VERSIONS
 
 
 X_MS_VERSION = _SUPPORTED_API_VERSIONS[-1]
 
-# Socket timeout in seconds
+# Default socket timeouts, in seconds
 CONNECTION_TIMEOUT = 20
-READ_TIMEOUT = 20
-
-# for python 3.5+, there was a change to the definition of the socket timeout (as far as socket.sendall is concerned)
-# The socket timeout is now the maximum total duration to send all data.
-if sys.version_info >= (3, 5):
-    # the timeout to connect is 20 seconds, and the read timeout is 80000 seconds
-    # the 80000 seconds was calculated with:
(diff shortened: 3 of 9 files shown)


vincenttran-msft (approved):
LGTM 👍

sdk/storage/azure-storage-blob/README.md
```diff
@@ -347,7 +344,8 @@ Other optional configuration keyword arguments that can be specified on the clie
 
 **Client keyword arguments:**
 
-* __connection_timeout__ (int): Optionally sets the connect and read timeout value, in seconds.
+* __connection_timeout__ (int): The number of seconds the client will wait to establish a connection to the server.
+* __read_timeout__ (int): The number of seconds the client will wait, after the connections has been established, for the server to send a response.
```
selimb, line 348: Hi @jalauzon-msft 
I came across this because I've been following #23049. I don't believe the description for `read_timeout` matches what either `requests` or `aiohttp` (the only possible underlying transport libraries, as far as I can tell) do with that argument (see below). It's not the total number of seconds until the full response is received, but rather the number of seconds between each response chunk.

## requests

When using `requests`, `read_timeout` is passed as the second element of the `timeout` 2-tuple ([in _requests_basic.py](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/core/azure-core/azure/core/pipeline/transport/_requests_basic.py#L325)), i.e. the read timeout. The [docs](https://docs.python-requests.org/en/latest/user/advanced/#timeouts) mention (bold is mine):

> Once your client has connected to the server and sent the HTTP request, the read timeout is the number of seconds the client will wait for the server to send a response. (**Specifically, it’s the number of seconds that the client will wait between bytes sent from the server. In 99.9% of cases, this is the time before the server sends the first byte**).

`requests` use `urllib3` under the hood, which reinforces the point in [its docs](https://urllib3.readthedocs.io/en/stable/reference/urllib3.util.html#urllib3.util.Timeout).

> In addition, the read and total timeouts only measure the time between read operations on the socket connecting the client and the server, not the total amount of time for the request to return a complete response. For most requests, the timeout is raised because the server has not sent the first byte in the specified time. This is not always the case; if a server streams one byte every fifteen seconds, a timeout of 20 seconds will not trigger, even though the request will take several minutes to complete.

Ultimately, the read timeout is passed to [socket.settimeout](https://docs.python.org/3/library/socket.html#socket.socket.settimeout). This is an OS-level timeout and not an application-level timeout, meaning that even large `time.sleep` between `recv()` calls won't trigger the timeout.

## aiohttp

When using `aiohttp`, `read_timeout` is used as the `sock_read`  parameter to `ClientTimeout` ([in _aiohttp.py](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/core/azure-core/azure/core/pipeline/transport/_aiohttp.py#L228)). The [docs](https://docs.aiohttp.org/en/stable/client_quickstart.html#timeouts) mention:

> The maximal number of seconds allowed for period between reading a new data portion from a peer.

`aiohttp`'s `ClientTimeout` does have a `total` parameter, but `requests` has no equivalent (although urllib3 does, so there may be a workaround there if desired).

## READ_TIMEOUT

Assuming the above is correct, then I believe `READ_TIMEOUT` is **way** too large and the calculation in the comment is ... not relevant. 400MB / 50KB/s gives you an estimate of the total time it takes to read the response, but AFAIK that response will never come in as a single chunk at the socket level (which is where the timeout is applied).

I'm not an expert on TCP and sockets, but I think a client-side read timeout would occur if:

- there's very high latency and packet loss.
- the server (azure blob storage) is either slow (at the application level) or overloaded (at the network level).
- maybe if the client's network is overloaded?

In other words, low bandwidth should not result in read timeouts. 
Even large file sizes shouldn't result in read timeouts (unless they result in slow processing times on the server side).

## Conclusion

I think the description for `read_timeout` should be changed and clarified, and that the default `READ_TIMEOUT` should be lowered -- it looks like the [Azure SDK for Java uses 60 seconds](https://javadoc.io/static/com.azure/azure-core/1.16.0/com/azure/core/util/HttpClientOptions.html#readTimeout-java.time.Duration-), why not use that?

Alternatively, a `total_timeout` parameter could be exposed either instead or alongside `read_timeout`.

Apologies for the long comment.

jalauzon-msft: Hi @selimb Selim, I know it's been quite some time since you left this comment, but I wanted to thank you for your insight and write-up here and let you know it hasn't gone unnoticed! We've been quite busy with other things but I'm finally planning to implement a change to the docs and the default `read_timeout` value soon! We are going to go with a default timeout of 60 seconds and docs similar to `urllib`. Thanks again!

jalauzon-msft: Here is the PR: #25950

selimb: @jalauzon-msft Thanks for getting back to me! I'm glad this was useful 🙂 👍️

PR #24034, opened 2022-04-15
[Storage] Fix documentation for `start_copy_from_url`

### LINKED ISSUE #23843: Fix documentation for `start_copy_from_url()` to better reflect its behavior.
The documentation for `start_copy_from_url()` states:
"For all blob types, you can call status() on the returned polling object to check the status of the copy operation, or wait() to block until the operation is complete. The final blob will be committed when the copy completes."

This is not accurate, we do not return any sort of polling object and rather just some properties of the destination blob. We should remove this and maybe document the expected way to check for the copy to complete. Additionally, `requires_sync`should probably be mentioned in the main documentation.

jalauzon-msft:
The documentation for `start_copy_from_url()` was specific to asynchronous copy and mentioned a return object that could be used to block until the operation was complete. I believe this was carried over from Track1 but is not accurate for Track2, especially now that synchronous copy has been added. Updating the documentation to discuss what is actually returned in sync/async case.

Resolves #23843


vincenttran-msft (approved):
Good catch 💯

PR #24052, opened 2022-04-16
[Storage] Fix changefeed import for local testing

jalauzon-msft:
There is an issue with running ChangeFeed tests locally due to the direct dependency on blob package which has a conflicting namespace. This change adds a workaround for this by importing from the `changefeed` folder directly rather than `azure-storage-blob-changefeed` and explains how to set up sources root in PyCharm for this to work. The original import is left as-is in an `except` block for pipelines.


vincenttran-msft (approved):
What a pesky import issue.. Thanks for tracking this down 😄

PR #24068, opened 2022-04-18
[pylint] fixing enum checker

l0lawrence:
Only str enum classes inherit metaclass

### DIFF
--- a/scripts/pylint_custom_plugin/pylint_guidelines_checker.py
+++ b/scripts/pylint_custom_plugin/pylint_guidelines_checker.py
@@ -1763,7 +1763,7 @@ class CheckEnum(BaseChecker):
                     self._enum_uppercase(node)   
             # Else if it does not have a metaclass, but it is an enum class
             # Check both capitalization and throw pylint error for metaclass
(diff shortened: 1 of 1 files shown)


scripts/pylint_custom_plugin/pylint_guidelines_checker.py
```diff
@@ -1763,7 +1763,7 @@ def visit_classdef(self, node):
                     self._enum_uppercase(node)   
             # Else if it does not have a metaclass, but it is an enum class
             # Check both capitalization and throw pylint error for metaclass
-            elif node.bases[1].name == "Enum":
+             elif node.bases[0].name == "str" and node.bases[1].name == "Enum":
```
kristapratico, line 1766: seems like this line starts with an extra space that's not needed?

kristapratico (approved):

scripts/pylint_custom_plugin/pylint_guidelines_checker.py
catalinaperalta, line 1766: Nit: extra space

catalinaperalta (approved):

PR #24070, opened 2022-04-18
[Key Vault] Update default value of lifetime_actions

mccoyp:
# Description

Context for supporting a value of `None` for `lifetime_actions` is described in https://github.com/Azure/azure-sdk-for-python/pull/24059. After looking over Managed HSM responses and discussing with Laurent, we decided that using an empty list instead of `None` makes the most sense for `KeyRotationPolicy.lifetime_actions`, since the assumption in the regular case is that this attribute is iterable, even if empty. Other attributes have had their type hints updated, since we can't rely on the values we _expect_ to get back from Key Vault.

  - No tests for Managed HSM since this feature isn't in preview yet for the service, and the changes here aren't relevant for Key Vault.

### DIFF
--- a/sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_models.py
+++ b/sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_models.py
@@ -291,12 +291,14 @@ class KeyRotationLifetimeAction(object):
     :param action: The action that will be executed.
     :type action: ~azure.keyvault.keys.KeyRotationPolicyAction or str
 
-    :keyword str time_after_create: Time after creation to attempt the specified action, as an ISO 8601 duration.
+    :keyword time_after_create: Time after creation to attempt the specified action, as an ISO 8601 duration.
         For example, 90 days is "P90D". See `Wikipedia <https://wikipedia.org/wiki/ISO_8601#Durations>`_ for more
         information on ISO 8601 durations.
-    :keyword str time_before_expiry: Time before expiry to attempt the specified action, as an ISO 8601 duration.
+    :paramtype time_after_create: Optional[str]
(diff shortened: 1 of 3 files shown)


sdk/keyvault/azure-keyvault-keys/azure/keyvault/keys/_models.py
```diff
@@ -319,30 +321,32 @@ def _from_generated(cls, lifetime_action):
...
     :ivar lifetime_actions: Actions that will be performed by Key Vault over the lifetime of a key.
     :vartype lifetime_actions: List[~azure.keyvault.keys.KeyRotationLifetimeAction]
-    :ivar str expires_in: The expiry time of the policy that will be applied on new key versions, defined as an ISO 8601
+    :ivar expires_in: The expiry time of the policy that will be applied on new key versions, defined as an ISO 8601
         duration. For example, 90 days is "P90D".  See `Wikipedia <https://wikipedia.org/wiki/ISO_8601#Durations>`_ for
         more information on ISO 8601 durations.
+    :vartype expires_in: Optional[str]
     :ivar created_on: When the policy was created, in UTC
-    :vartype created_on: ~datetime.datetime
+    :vartype created_on: Optional[~datetime.datetime]
```
iscai-msft, line 333: The ivar won't be optional

mccoyp: Synced about this offline -- this attribute (and others now marked as `Optional`) are actually optional, because these could be set as `None` at model creation time. `lifetime_actions` still isn't marked as `Optional` since it will always default to an empty list, even if the service returns `None` for the property

kristapratico (approved):
discussed offline

PR #24074, opened 2022-04-18
[pylint] Fix remoterender

l0lawrence:


catalinaperalta (approved):
Just double check that the changelog error should be resolved with this change.

PR #24104, opened 2022-04-20
[pylint] Core Pylint 

l0lawrence:
Fixes Core pylint issues with Enums

Bypassing pylint on Cosmos files with api version keyword questions so that core will be green, follow up with issue #24117

### DIFF
--- a/sdk/core/azure-core/azure/core/pipeline/policies/_retry.py
+++ b/sdk/core/azure-core/azure/core/pipeline/policies/_retry.py
@@ -45,11 +45,12 @@ from azure.core.exceptions import (
 
 from ._base import HTTPPolicy, RequestHistory
 from . import _utils
-
+from ..._enum_meta import CaseInsensitiveEnumMeta
 
 _LOGGER = logging.getLogger(__name__)
 
-class RetryMode(str, Enum):
+class RetryMode(str, Enum, metaclass=CaseInsensitiveEnumMeta):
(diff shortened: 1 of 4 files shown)


sdk/core/azure-core/azure/core/pipeline/policies/_retry.py
```diff
@@ -45,11 +45,12 @@
...
-
+from ..._enum_meta import CaseInsensitiveEnumMeta
 
 _LOGGER = logging.getLogger(__name__)
 
-class RetryMode(str, Enum):
+class RetryMode(str, Enum, metaclass=CaseInsensitiveEnumMeta):
+    # pylint: disable=enum-must-be-uppercase
     Exponential = 'exponential'
     Fixed = 'fixed'
```
tjprescott, line 55: Why is this disabled if we are in fact following the rule?

l0lawrence: I wasn't sure if this would be a breaking change to make it EXPONENTIAL and FIXED

tjprescott: Ooooh, it's saying the entire thing should be all caps. Okay, yeah that would be. I presume the case insensitivity only applies to the string value. If it actually applied to the attribute label I'd be surprised (and impressed honestly)

iscai-msft (approved):

PR #24118, opened 2022-04-20
[Storage] Fix test_quick_query_output_in_arrow_format

vincenttran-msft:
As seen in [failing pipeline example](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=1518966&view=ms.vss-test-web.build-test-results-tab&runId=31021430&resultId=101466&paneView=debug), we have had this show up in CI and I believe it has been there for awhile now. 

After confirming with the service team that the output has changed due to a change on their side, this PR aims to resolve this test failure by putting the **correct** output as the expected output.


jalauzon-msft (approved):
Thanks for jumping on this!

PR #24137, opened 2022-04-22
[AutoRelease] t2-education-2022-04-22-33298(Do not merge)

azclibot:
https://github.com/Azure/sdk-release-request/issues/2730 
Live test success


sdk/education/azure-mgmt-education/azure/mgmt/education/aio/operations/_create_operations.py
```diff
@@ -0,0 +1,196 @@
...
+        request.url = self._client.format_url(request.url)
+
+        pipeline_response = await self._client._pipeline.run(  # pylint: disable=protected-access
+            request,
+            stream=False,
+            **kwargs
+        )
+        response = pipeline_response.http_response
+
+        if response.status_code not in [200, 201]:
```
yuhhuo, line 100: eduRP create and update endpoint is going to be returning 201 and 202, this will be failing is 202 is returned.
How do we fix this?

BigCat20196: Hi Yuheng, Python SDK is generated based on the [rest API](https://github.com/Azure/azure-rest-api-specs), so to add 202 response, we need to define it [here](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/education/resource-manager/Microsoft.Education/preview/2021-12-01-preview/education.json#L410) first, and then we regenerate it.

PR #24150, opened 2022-04-22
[Cosmos] bugfix: set automaticIdGeneration to false properly for `upsert_item()` calls

simorenoh:
Seems like this option wasn't being parsed properly due to wrong naming, which was allowing for the creation of items with auto-generated id's through the use of `upsert_item()` method. This is wrong behavior, and with these changes the backend successfully throws errors when users upsert items without an 'id' field in the body. The parsing was happening in _cosmos_client_connection.py, _GetContainerIdWithPathForItem method, and was always being ignored since the option was not being properly set in the upsert_item call.

Marked it as a breaking change since this may have allowed users to create items in the past without passing in an 'id' field in their item bodies - which would now no longer be possible.


tjprescott (approved):
LGTM. I'm somewhat curious why if the service supports an auto-generated ID the SDK doesn't, but clearly this was not the intended behavior in the SDK.

PR #24153, opened 2022-04-22
[ServiceBus] Sample typo

l0lawrence:


swathipil (approved):
this is a byootiful catch

PR #24165, opened 2022-04-25
[KeyVault] Keyvault Keys to Test Proxy

### LINKED ISSUE #23677: Move KeyVault Keys to Test Proxy

kashifkhan:
closes #23677


sdk/keyvault/azure-keyvault-keys/tests/_shared/test_case.py
```diff
@@ -2,28 +2,46 @@
...
-    def setUp(self):
-        self.list_test_size = 7
-        super(KeyVaultTestCase, self).setUp()
 
+class KeyVaultTestCase(AzureRecordedTestCase):
     def get_resource_name(self, name):
         """helper to create resources with a consistent, test-indicative prefix"""
         return super(KeyVaultTestCase, self).get_resource_name("livekvtest{}".format(name))
 
+    def _get_attestation_uri(self):
```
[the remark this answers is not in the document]
mccoyp: I forgot that the class in `/tests/_test_case.py` is now a preparer (which does make sense) -- maybe we should also create a `KeysTestCase` class (or similar) that has these necessary test instance methods, and either have it inherit from the shared test class, or have test file classes inherit from both classes like they currently do

sdk/keyvault/azure-keyvault-keys/tests/_async_test_case.py
```diff
@@ -0,0 +1,134 @@
...
+        
+        return self.create_client_from_credential(KeyClient, credential=credential, vault_url=vault_uri, **kwargs)
+
+    def create_crypto_client(self, key, **kwargs):
+        
+        from azure.keyvault.keys.crypto.aio import CryptographyClient
+
+        credential = self.get_credential(CryptographyClient, is_async=True)
+        return self.create_client_from_credential(CryptographyClient, credential=credential, key=key, **kwargs)
+
```
mccoyp, line 113: We can remove this since it's not needed in the preparer
```suggestion
```

sdk/keyvault/azure-keyvault-keys/tests/_async_test_case.py
```diff
@@ -0,0 +1,134 @@
...
+    def _get_attestation_uri(self):
+        playback_uri = "https://fakeattestation.azurewebsites.net"
+        if self.is_live:
+            real_uri = os.environ.get("AZURE_KEYVAULT_ATTESTATION_URL")
+            if real_uri is None:
+                pytest.skip("No AZURE_KEYVAULT_ATTESTATION_URL environment variable")
+            self._scrub_url(real_uri, playback_uri)
+            return real_uri
+        return playback_uri
```
mccoyp, line 122: We can also remove this method:
```suggestion
```

sdk/keyvault/azure-keyvault-keys/tests/_keys_test_case.py
```diff
@@ -0,0 +1,36 @@
...
+        if kwargs.pop("is_async", False):
+            from azure.keyvault.keys.crypto.aio import CryptographyClient
+            credential = self.get_credential(CryptographyClient,is_async=True)
+        else:
+            from azure.keyvault.keys.crypto import CryptographyClient
+            credential = self.get_credential(CryptographyClient)
+
+        return self.create_client_from_credential(CryptographyClient, credential=credential, key=key, **kwargs)
+
+    def create_key_client(self, vault_uri, **kwargs):
```
mccoyp, line 28: It looks like this method is only used in the single `test_parse_id.py` live test -- I wonder if we should restructure that test to use a client preparer instead and remove this method. It probably wouldn't hurt to test `test_parse_key_id_with_version` against all API versions; I wouldn't expect the behavior to be API version-dependent, but technically it depends on a response from the service so it's not unreasonable

sdk/keyvault/azure-keyvault-keys/tests/_test_case.py
```diff
@@ -87,79 +62,69 @@ def get_test_parameters(only_hsm=False, only_vault=False, api_versions=None):
...
-            credential = self.get_credential(KeyClient)
+    def create_key_client(self, vault_uri, **kwargs):
+       
+        from azure.keyvault.keys import KeyClient
+
+        credential = self.get_credential(KeyClient)
+        
         return self.create_client_from_credential(KeyClient, credential=credential, vault_url=vault_uri, **kwargs)
 
     def create_crypto_client(self, key, **kwargs):
```
mccoyp, line 110: Same comments from the async version: we can remove `create_crypto_client` and `_get_attestation_uri`

mccoyp (approved):
🚀🚀🚀

PR #24170, opened 2022-04-25
[Storage] Fix progress_hook types for API View

jalauzon-msft:
API View did not like the way the types for the new `progress_hook` were defined. Updating them to use the `paramtype` docstring instead. Also including the `progress_hook` in the docstring for `upload_blob` in `ContainerCleint` which was previously missed.


vincenttran-msft (approved):
LGTM 👍

PR #23005, opened 2022-02-10
Update Search to enable live testing in sovereign clouds for multiple…

v-xuto:
1.Fixes https://github.com/Azure/azure-sdk-for-python/issues/22867. These changes enable Search to run live tests against Public, UsGov and China.

2.File changed description:
- Match different endpoint suffix in different clouds.
image

- Enable UsGov and China clouds.
image

3.Pipeline results:
https://dev.azure.com/azure-sdk/internal/_build/results?buildId=1358501&view=results

@jameszliao-msft, @lmazuel, @meeraharidasa, @AlexGhiondea and @kashifkhan for notification.

PR #24198, opened 2022-04-26
[Cosmos] Infra: use Bicep for resource provisioning

simorenoh:
This PR takes care of updating the Python Cosmos SDK to utilize Bicep to provision test resources. Also has changes to the test config files to ensure these changes work and that the Cosmos test pipelines can be enabled again.


sdk/cosmos/test-resources.bicep
```diff
@@ -0,0 +1,66 @@
...
+param location string = resourceGroup().location
+
+@description('The api version to be used by Bicep to create resources')
+param apiVersion string = '2020-04-01'
+
+var accountName = toLower(baseName)
+var resourceId = cosmosAccount.id
+var singleRegionConfiguration = [
+  {
+    locationName: 'East US 2'
```
benbp, line 22: I'm not sure yet how we'll solve this for multi-region, but for single region you could also set the location name to `resourceGroup().location` (unless it gives the short name, if that won't work).

Also one other thing to consider is hardcoding regions will break the tests if/when we try to add sovereign cloud support. We'll have to figure this out anyway in more detail for multi-region though.

benbp: Also, which region configuration will be used in the tests? Single? Is the idea that locally you may do multi-region, or vice versa?

Would we want to support testing BOTH configurations in the pipeline?

sdk/cosmos/test-resources.bicep
```diff
@@ -0,0 +1,66 @@
...
+    consistencyPolicy: {
+      defaultConsistencyLevel: defaultConsistencyLevel
+    }
+    locations: locationsConfiguration
+  }
+}
+
+
+output ACCOUNT_HOST string = reference(resourceId, apiVersion).documentEndpoint
+output ACCOUNT_KEY string = listKeys(resourceId, apiVersion).primaryMasterKey
```
benbp, line 66: Not sure if these are magic variable names or not, but it may be better to do:

```
output COSMOS_ACCOUNT_HOST string = reference(resourceId, apiVersion).documentEndpoint
output COSMOS_ACCOUNT_KEY string = listKeys(resourceId, apiVersion).primaryMasterKey
```

benbp:
Some nits and considerations but LGTM.

sdk/cosmos/test-resources.bicep
tjprescott: @v-xuto has experience converting tests to work with sovereign cloud. Here's an example PR for Search: https://github.com/Azure/azure-sdk-for-python/pull/23005

tjprescott:
@benbp does @simorenoh need to make a nominal change to the Cosmos code path to trigger the CI to run the tests?

tjprescott:
@simorenoh does this resolve the live test failures or is this a stepping stone on that path?

sdk/cosmos/test-resources.bicep
simorenoh: I'll keep in mind this suggestion for the next iteration, but currently we use 'ACCOUNT_HOST' and 'ACCOUNT_KEY' respectively in all our tests so keeping this like this for now

sdk/cosmos/test-resources.bicep
simorenoh: @benbp we do seem to use location for the `cosmosAccount` so it can probably be used for the single region - these settings are based off the Java SDK's own configs for test-resources.json and are supposed to be used in our multimaster tests, but not sure exactly how this gets set since this deployment only runs once right?

PR #24206, opened 2022-04-27
[textanalytics] remove v3.2-preview.2

kristapratico:
Removes v3.2-preview.2 now that all client methods are routed to the language API. Also regenerates with `python3-only` and fixes some casing for properties in the models file.


iscai-msft (approved):

maririos:
so this means that the library now supports 3.0, 3.1, and then language api?

kristapratico:
> so this means that the library now supports 3.0, 3.1, and then language api?

Yes, although it supports 2022-03-01-preview currently (haven't yet regenerated on the April swagger). Here's the PR that merges the feature branch and includes all the PRs added to support language API: https://github.com/Azure/azure-sdk-for-python/pull/24195

PR #24216, opened 2022-04-27
[Storage] Tag indexing for versions

vincenttran-msft:
Add support for `include=['versions']` for `find_blobs_by_tags`.


sdk/storage/azure-storage-blob/azure/storage/blob/_models.py
```diff
@@ -679,11 +679,17 @@ class FilteredBlob(DictMixin):
     :type container_name: str
     :ivar tags: Key value pairs of blob tags.
     :type tags: Dict[str, str]
+    :ivar version_id: Version ID
+    :vartype version_id: str
```
jalauzon-msft, line 683: Can you change `vartype` to `type` to match others?

sdk/storage/azure-storage-blob/azure/storage/blob/aio/_models.py
```diff
@@ -138,7 +137,10 @@ async def _extract_data_cb(self, get_next_return):
     def _build_item(item):
         if isinstance(item, FilterBlobItem):
             tags = parse_tags(item.tags)
-            blob = FilteredBlob(name=item.name, container_name=item.container_name, tags=tags)
+            blob = FilteredBlob(name=item.name,
+                                container_name=item.container_name,
+                                tags=tags, version_id=item.version_id,
```
jalauzon-msft, line 142: nit: One more newline to split these two params?

sdk/storage/azure-storage-blob/tests/test_blob_tags.py
```diff
@@ -384,6 +384,34 @@ def test_filter_blobs(self, storage_account_name, storage_account_key):
...
+        where = "\"tag1\"='firsttag'"
+
+        # Act
+        blob_client.create_append_blob(tags=tags1)
+        blob_client.create_append_blob(tags=tags1)
+
+        if self.is_live:
+            sleep(10)
+
+        blob_list = self.bsc.find_blobs_by_tags(filter_expression=where, results_per_page=2,
```
jalauzon-msft, line 404: Sorry I probably should have mentioned this while we were working on it, but would you mind refactoring the tests to not use `by_page`? That's really only to test pagination which is covered by other tests so this test should be simplified and only test the version support. For sync it would be pretty simple. Just make this line more like (just wrapping the result in `list()` and not including the `results_per_page` arg):

`blob_list = list(self.bsc.find_blobs_by_tags(filter_expression=where, include=include_list))`

Then just minor refactoring of the asserts.

sdk/storage/azure-storage-blob/tests/test_blob_tags_async.py
```diff
@@ -376,4 +376,36 @@ async def test_filter_blobs(self, storage_account_name, storage_account_key):
...
+        where = "\"tag1\"='firsttag'"
+
+        # Act
+        await blob_client.create_append_blob(tags=tags1)
+        await blob_client.create_append_blob(tags=tags1)
+
+        if self.is_live:
+            sleep(10)
+
+        blob_list = self.bsc.find_blobs_by_tags(filter_expression=where, results_per_page=2,
```
jalauzon-msft, line 397: Same comment about refactoring to not use `by_page`. It will be slightly different for async, you'll have to use a separate list variable an `async for` to append all items to the list then do the asserts. There should be plenty of examples around, you can look at `list_blobs` examples.

sdk/storage/azure-storage-blob/tests/test_blob_tags.py
jalauzon-msft: Same comment applies to async and container versions of the tests.

jalauzon-msft (approved):

PR #24217, opened 2022-04-28
[Storage] Fix service operations with account SAS

### LINKED ISSUE #24156: `BlobServiceClient.list_containers()` fails when using a sas token generated using `generate_account_sas`
- **Package Name**: azure-storage-blob
- **Package Version**: 12.12.0b1
- **Operating System**: OSX
- **Python Version**: 3.9

**Describe the bug**

`BlobServiceClient.list_containers()` fails when using an account SAS token generated using `generate_account_sas`

**To Reproduce**

If I create an account SAS token like this: 

```py
from datetime import datetime, timezone, timedelta
from azure.storage.blob import BlobServiceClient, generate_account_sas

ACCOUNT_KEY = "<< account key >>"
ACCOUNT_NAME = "<< account name >>"
start_time = datetime.now(timezone.utc) - timedelta(minutes=1)
expiry_time = datetime.now(timezone.utc) + timedelta(hours=1)
ACCOUNT_SAS_TOKEN = generate_account_sas(
    account_key=ACCOUNT_KEY,
    account_name=ACCOUNT_NAME,
    start=start_time,
    expiry=expiry_time,
    permission="rl",
    resource_types="sco",
)
BLOB_ACCOUNT_URL = "https://{account}.blob.core.windows.net/".format(
    account=ACCOUNT_NAME
)
```

`BlobServiceClient.list_containers()` fails when using the account sas token like so:

```py
BSC = BlobServiceClient(account_url=BLOB_ACCOUNT_URL, credential=ACCOUNT_SAS_TOKEN)
container_list = BSC.list_containers()
container_list.next()
# *** azure.core.exceptions.ClientAuthenticationError: Server failed to authenticate the request. Make sure the value of Authorization header is formed correctly including the signature.
```

Even though it works just fine with the account key:

```py
BSC = BlobServiceClient(account_url=BLOB_ACCOUNT_URL, credential=ACCOUNT_KEY)
container_list = BSC.list_containers()
container_list.next() # no errors
```

**Expected behavior**

`BlobServiceClient.list_containers()` should work with an account key generated by `generate_account_sas`

**Screenshots**
N/A

**Additional context**
N/A

jalauzon-msft:
Resolves #24156

As part of the latest beta release, we introduced a regression that caused certain service operations (e.g., `list_containers()`) to fail when using an account SAS as the authentication method. This was due to the modifications that were made to the generated code as part of the move to the latest version of `autorest/python`. One of the custom directives we added wasn't fully correct and caused URLs to be malformed with certain service operations using account SAS.

This change modifies the custom directive and re-generates the service operations code to address the issue. This change also introduces a series of tests that would have helped to catch this issue before release and can help prevent future, similar regressions.

### DIFF
--- a/sdk/storage/azure-storage-blob/CHANGELOG.md
+++ b/sdk/storage/azure-storage-blob/CHANGELOG.md
@@ -8,6 +8,8 @@
 ### Bugs Fixed:
 - Fixed a bug in `BlobClient.from_blob_url()` such that users will receive a more helpful error
 message if they pass an incorrect URL without a full `/container/blob` path.
+- Fixed a bug, introduced in the previous beta release, that caused Authentication errors when attempting to use
+an Account SAS with certain service level operations.
 
 ## 12.12.0b1 (2022-04-14)
 
--- a/sdk/storage/azure-storage-blob/azure/storage/blob/_generated/aio/operations/_service_operations.py
+++ b/sdk/storage/azure-storage-blob/azure/storage/blob/_generated/aio/operations/_service_operations.py
@@ -86,6 +86,7 @@ class ServiceOperations:
         _content = self._serialize.body(storage_service_properties, 'StorageServiceProperties', is_xml=True)
 
         request = build_set_properties_request(
+            url=self._config.url,
             restype=restype,
             comp=comp,
             version=self._config.version,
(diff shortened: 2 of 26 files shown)


vincenttran-msft (approved):
Great work, and thanks for increasing test coverage 👍

PR #24234, opened 2022-04-28
[Cosmos] Bugfix: Get database account properly raises 401s

simorenoh:
This is a bugfix I found while working on separate things, but our SDK was not throwing 401 errors for failed calls to the GetDatabaseAccount method that gets invoked during client initialization. 

This would throw errors later regardless:
- If the user **did not** provide an explicit consistency level, in which case the SDK would throw an error when attempting to get the database account's consistency level (NoneType does not have attribute ConsistecyLevel)
- If the user **did** provide an explicit consistency level, in which case the SDK would throw an Unauthorized error on any other IO operation after initializing their client.

With this we fix both and ensure that DatabaseAccount information has to be populated properly to use the client.


simorenoh:
Changing this to draft and adding DO NOT MERGE label since it might actually break a lot of our tests that mock client behavior and don't count on this check happening - tests should be fixed along with the bugfix if possible.

Issue #24221, opened 2022-04-28
[pipeline] upgrade autorest.python from `5.13.0` to `5.16.0` 


msyyc:
To fix the [content_type](https://github.com/Azure/autorest.python/blob/autorestv3/ChangeLog.md#2022-04-18---5160), we have to upgrade version(>= 5.16.0). Now we find that [compute](https://github.com/Azure/azure-rest-api-specs/pull/17988/checks?check_run_id=6197650799) has the bug, we must upgrade it before next release for compute.

5.16.0 needs min version of `azure-core` is 1.23.0, so we have to update the following things:

1. update mindep and release new version for [azure-mgmt-core](https://github.com/Azure/azure-sdk-for-python/blob/d69e517311f1878d52849a6973b6ebd521333497/sdk/core/azure-mgmt-core/setup.py#L70)
2. [autorest configuration]()
3. [template](https://github.com/Azure/azure-sdk-for-python/blob/b580e4912ab633d5705f0c60c2616d356e3620f2/tools/azure-sdk-tools/packaging_tools/templates/setup.py#L76) (to confirm)
4. [auto-release pipeline](https://github.com/Azure/azure-sdk-for-python/blob/b580e4912ab633d5705f0c60c2616d356e3620f2/scripts/auto_release/main.py#L254-L255) (to confirm)

steps:
1 first, then 2, 3, 4 at the same time

BigCat20196:
Resources pr:
5.13.0 https://github.com/Azure/azure-sdk-for-python/pull/24325
5.16.0 https://github.com/Azure/azure-sdk-for-python/pull/24326

BigCat20196:
Compute pr:
5.13.0: https://github.com/Azure/azure-sdk-for-python/pull/24346
5.16.0: https://github.com/Azure/azure-sdk-for-python/pull/24347

BigCat20196:
Network pr:
5.13.0: https://github.com/Azure/azure-sdk-for-python/pull/24354
5.16.0: https://github.com/Azure/azure-sdk-for-python/pull/24353

PR #24245, opened 2022-04-29
[mgmt][core] upgrade min dependency for `azure-core`

msyyc:
https://github.com/Azure/azure-sdk-for-python/issues/24221


sdk/core/azure-mgmt-core/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Release History
 
+## 1.3.1 (2022-04-28)
+
+### Other Changes
+
+- Updated required `azure-core` version
```
iscai-msft, line 7: v tiny nit, don't even have to change it for this pr, but it would be nice to say the azure-core min version

iscai-msft (approved):

PR #24258, opened 2022-04-29
Share APIView Pins of pylint-guidelines-checker

scbedd:
@tjprescott @praveenkuttappan 

This will remain draft until after release week. I know that Travis didn't want to accidentally turn the world on its head with new apiview errors the week of :P 

[sdk-tools PR](https://github.com/Azure/azure-sdk-tools/pull/3189) that does the same thing

So, process if we want to add more guideline checks to APIView.

1. bump `api-stub-generator` pylint-guidelines-checker pin.
2. Release new api-stub-generator
3. Pin new apistub generator version in APIView UI Deployment from sdk-tools
4. Pin new apistub generator in the eng/apiview_reqs.txt from sdk-for-python

Should be very straightfoward.

This PR will be **red** until apiview version `0.3` releases.


tjprescott:
@scbedd as long as both the SDK and the web installation use the same version of pylint-guidelines-checker, I would prefer to not have to release a new version of `api-stub-generator` just to bump the guidlines checker version. Would this unified pipeline permit that?

scbedd:
cc @praveenkuttappan for further discussion

@tjprescott no. My entire goal with these PRs has been to _lock down_ on the pylint-guidelines-checker version. 

I want to:

- Unify the version of `pylint-guidelines-checker` that both Azure SDK CI StubGen and APIView are consuming, without relying on timing of release to ensure we're using the same one. This means that we _pin_ the version of `pylint-guidelines-checker`. In both APIView and our CI.
   - When I spoke to you and Praveen, we discussed pinning the dependency in the `api-stub-generator` and then just bumping to a new version of `api-stub-generator` whenever we want to update the consumed version. 
- If we want to leave `pylint-guidelines-checker` unpinned in `api-stub-generator`, I'm fine with that, but we will still need to do the same amount of work (updating a pin some place instead of a setup.py) and release APIView UI/sdk-for-python CI update when we release a new version. You won't have to release a new version of `api-stub-generator` though.

However, the one piece I am not ok with is just floating the version of `pylint-guidelines-checker` we consume.

tjprescott (approved):
I'm happy with it if it does what you want it to and passes CI :)

praveenkuttappan (approved):

PR #24276, opened 2022-05-03
[Storage] Add progress callback to `download_blob` methods

jalauzon-msft:
This change adds a progress callback, `progress_hook`, to `download_blob` which can be used to track the progress of a long running download. This is similar to the existing `progress_hook` that was added for `upload_blob` recently.

This addresses the download piece of #23934


sdk/storage/azure-storage-blob/azure/storage/blob/aio/_blob_client_async.py
```diff
@@ -465,6 +465,11 @@ async def download_blob(self, offset=None, length=None, **kwargs):
             The number of parallel connections with which to download.
         :keyword str encoding:
             Encoding to decode the downloaded bytes. Default is None, i.e. no decoding.
+        :keyword progress_hook:
```
vincenttran-msft, line 468: Is this an optional kwarg? If so, just curious as to why this convention as opposed to:
`:keyword Callable[[int, int], Awaitable[None]] progress_hook`
Same feedback on non-async client

vincenttran-msft (approved):
LGTM! Just possible change on docstring unless I'm misunderstanding 🥇

sdk/storage/azure-storage-blob/azure/storage/blob/aio/_blob_client_async.py
jalauzon-msft: The API View generator did not like this type listed in that format for some reason. It would cause this keyword to not show up at all in the View. So, I moved it to a `paramtype` to appease API View. Good question though as I would normally like to have it how you mentioned.

PR #24281, opened 2022-05-03
[textanalytics] decorator to validate multiapi 

kristapratico:
This decorator helps... 
- move some of the multiapi param validation out of the client methods
- validate args/action type support between the TA service vs Language... since the input shaped has changed between API versions and many of the args went from query params -> body params, we validate so that users targeting an older API version don't pass unsupported params which get silently ignored. We plan to let the service validate between Language API versions in the future.


iscai-msft (approved):
lowkey a ba but looks good from the skim!

PR #24296, opened 2022-05-04
Only MCR Images

scbedd:
I have made these updates as appropriately as I can tell. These all build and `python` still resolves to correct version, etc. As of yet there are [no python images](https://eng.ms/docs/more/containers-secure-supply-chain/approved-images) within MCR that we can just crib off of. Expected to come in copper though.

@xiangyan99 can you take a look at the identity tests docker image changes?
@swathipil do you see any gotchas with my changes to eventhub and service-bus stress containers?
@benbp are the `stress` images actively being used? How should I test these?

I am fairly confident in `tools` dockerfile, so I won't look for a specific reviewer on that one.


tools/Dockerfile
```diff
@@ -1,19 +1,9 @@
-FROM ubuntu
-MAINTAINER zikalino
+FROM mcr.microsoft.com/dotnet/sdk:5.0.407-focal
+LABEL authors="zikalino,scbedd"
 
 RUN apt-get update
 RUN apt-get install -y git curl gnupg vim python3 python3-pip git software-properties-common apt-transport-https wget python3-venv nodejs npm libunwind-dev
 
-# install dotnet
```
scbedd, line 7: We can simply consume from a dotnet/sdk image instead of installing this entire thing for ourselves.

sdk/eventhub/azure-eventhub/stress/Dockerfile
```diff
@@ -1,8 +1,16 @@
 # syntax=docker/dockerfile:1
 
-FROM python:3.8-slim-buster
+ARG PYTHON_VERSION=3.8
+FROM mcr.microsoft.com/dotnet/runtime:5.0.9-focal
```
weshaggard, line 4: I know I suggested this but it does look a little odd out of context. It might be worth at least a comment note about why we are using this image.

scbedd: I'll publish a base image for `python3.9` and `python3.10`, then rely on that. We'll be able to cut all these install steps anyway.

scbedd:
Closing this PR to follow up on #24299 before swapping these images over. Will be disabling the warning temporarily.

Issue #24299, opened 2022-05-04
Produce `cbl-mariner` base image for python


scbedd:
To be consumed by

- [ ] sdk/eventhub/azure-eventhub/stress/Dockerfile
- [ ] sdk/identity/azure-identity/tests/managed-identity-live/Dockerfile
- [ ] sdk/servicebus/azure-servicebus/stress/Dockerfile
- [ ] tools/Dockerfile

PR #24300, opened 2022-05-04
Disable Azure Artifact Docker Warnings

scbedd:
Addressing in #24299 with priority.


eng/pipelines/templates/variables/globals.yml
```diff
@@ -3,3 +3,5 @@ variables:
   skipComponentGovernanceDetection: true
   AzureSDKCondaChannel: https://azuresdkconda.blob.core.windows.net/channel1/
   Package.EnableSBOMSigning: true
+  # this will be disabled after issue#24299 is merged.
```
weshaggard, line 6:
```suggestion
  # this will be disabled until issue#24299 is addressed.
```

weshaggard (approved):
Looks like you have an interesting merge but it will likely go away once you squash merge.

PR #24335, opened 2022-05-06
[Storage] Adjust connection string parser test cases

vincenttran-msft:
This PR adjusts the test cases added in #23854
While the original test cases were also sufficient as a failed request would result in:
 `ErrorCode:AuthenticationFailed` , `authenticationerrordetail:Invalid resource path` on the creation of a `DataLakeFileClient` object

The intended logic was to create a `file_client` using the test information prepared in `self._setUp` which shows that that `file_system_name` and `path_name` are valid, and then to create a `new_file_client` while appending extraneous `/` symbols to test the parser.

The current logic creates a `new_file_client` and functionality is tested as intended, but the final assert is whether operations are valid on the original `file_client` and not the `new_file_client`.


jalauzon-msft (approved):
Good catch on this!

Issue #24307, opened 2022-05-05
[tools] optimize comparison strategy of changelog


msyyc:
When generating changelog, python compares current code with last SDK version no matter the version is beta version or not. For other languages:
- Java: If current version is beta, compare with last version; if stable, compare with last **stable** version
- Go: same with Java
- Js: always compare with last **stable** version

Python shall adopt one of up categories.

msyyc:
https://github.com/Azure/azure-sdk-for-python/issues/24307

PR #24349, opened 2022-05-07
[auto_release] remove surplus code

msyyc:
After https://github.com/Azure/azure-sdk-for-python/pull/24348 merged, we could get tag kind from middle result so there is no need to keep the surplus code anymore


msyyc:
https://github.com/Azure/azure-sdk-for-python/issues/24307

BigCat20196 (approved):

Issue #23909, opened 2022-04-08
[EventHubs] Buffered Producer back-pressure handling strategy


yunhaoling:
### Issue: Back-pressure

How should the `EventHubProducerClient` in buffered mode behave against the scenario of enqueuing events when the buffer is full/not having enough room for the coming events.

(Originally from [comment in PR](https://github.com/Azure/azure-sdk-for-python/pull/23748#discussion_r839924407))

Anna's comment:

```
What's interesting is that triggering this flush will be blocking....
I wonder if the background worker should be monitoring the buffer size and trigger a background flush when this size is hit - rather than waiting until the next put_events request to see that it's full and block until it's flushed?
What if I didn't set a max_wait_time, only a max_buffer_length - but then only add an event once per minute? We wouldn't check if the buffer was full until a minute after it had been full?
Looks like this can't happen as both max_wait_time and max_buffer_length must both have values - though depending on those values, we could end up waiting before we realize the buffer has been sitting full since the last event was added.
```

#### sample cases:
max_buffer_length is 100, there's 50 events in the buffer, and 60 coming

----------

### Scope

1. the timeout behavior
2. the back-pressure handling strategy

----------

### Current implementation/behavior:

when EventHubProducerClient works in buffered mode, and.send_event/batch is called with timeout, timeout is used to control the enqueue operation, not the send operation.

if there're fewer slots in the buffer than the incoming events, flush first with the given timeout, then check timeout, if times out then we raise error, else we put events into buffer.

#### pseudocode: 

```python
if there_is_room_in_buffer:
    enqueue_events()
else:
    flush(timeout)  # clear the buffer
    if check_time_out_after_flush:
        raise timeout
    else:
        enqueue_events()
```

----------

### Other option

1. In general, we need a way for `EventHubProducerClient.send_event/batch` in buffered mode to flush the buffer queue.
if not via proactive call, we could use flag/condition to decouple `send_event/batch` and the flush operation.

we could add another flag and reuse the `check_max_wait_time_future` thread for flushing, it's almost the same effect as the current implementation, but decouple `flush` from `enqueue_events`. but this way, the timing is not perfectly controlled as `check_max_wait_time_future` will sleep periodically or we change the sleep period simultaneously.

```python
self._need_flush = threading.Event()  # or Condition

if there_is_room_in_buffer:
    enqueue_events()
else:
    try:
        self._need_flush.set()
        self._not_full.wait()
        if check_time_out_after_flush:
[...]

Issue #23844, opened 2022-04-06
[EventHubs] Buffered Producer concurrent sending support


yunhaoling:
### Issue
support concurrent sending in the EventHubProducerClient buffered mode

### Background

there are two parameters in .NET:
- `maximumConcurrentSends`: The number of batches that may be sent concurrently across all partitions.
- `maximumConcurrentSendsPerPartition`: The number of batches that may be sent concurrently to each partition.

However, those two parameters are pulled out of the scopes of the preview release due to following reasons:

- Our current producer implementation is preventing concurrency as we [lock the whole producer's send method](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/eventhub/azure-eventhub/azure/eventhub/_producer.py#L250):  which makes this `maximumConcurrentSendsPerPartition` useless.
- The `maximumConcurrentSends` would also bring confusion if we cannot support concurrent sending for each partition producer. e.g. set maximumConcurrentSends to 10 but we only have 4 partition producer for each partition.
- This is an advanced perf configuration; I think only few people would be using it. We could add it incrementally.

To sum up, the concurrency story is an advanced feature which needs more time for investigation and experimenting, I prefer pulling it out of the scopes for the first preview.

_Originally posted by @yunhaoling in https://github.com/Azure/azure-sdk-for-python/pull/23748#discussion_r844298172_

### Scopes:

- understand the necessity of this parameter, understand whether/how customer could benefit from this parameter.
- propose a technical solution to support this feature as the current implementation of EventHubProducer could not accommodate concurrent model well (there's a big lock)
  - to enable concurrent sending, there're two options (and more) that I think we could do:
    - 1. refactor the EventHubProducer implementation to embrace concurrency with fine-grained lock
    - 2. or create multiple partition producers

PR #24362, opened 2022-05-09
[EventHubs] Buffered Producer 

kashifkhan:
This PR is looking to address the following issues:
* Concurrency strategy - #23844
* Back-pressure handling strategy - #23909

How BufferedProducer works and its place within EventHub #23748

After an initial sync with Anna, we agreed on the following changes to BufferedProducer:
* The `ThreadPoolExecutor` would give us the concurrency we needed given the design of n number of `BufferedProducers` per partition. We would expose a `buffer_concurrency` kwarg that would allow a user to send in their own `ThreadPoolExecutor` or maximum number of workers.
* Document the Python specific behavior for how many workers are created by default
* Clean up `BufferedProducer` to not use semaphores & conditions, given that each `Queue` belongs to a particular partition, is independent and is not shared/modified between partitions.
* The background worker `check_max_wait_time_worker` in `BufferedProducer` would proactively flush the queue if it was at max capacity bringing the behavior inline with the documentation.

### DIFF
--- a/sdk/eventhub/azure-eventhub/azure/eventhub/_buffered_producer/_buffered_producer.py
+++ b/sdk/eventhub/azure-eventhub/azure/eventhub/_buffered_producer/_buffered_producer.py
@@ -168,7 +168,8 @@ class BufferedProducer:
             if self._cur_buffered_len > 0:
                 now_time = time.time()
                 _LOGGER.info("Partition %r worker is checking max_wait_time.", self.partition_id)
-                # flush the partition if the producer is running beyond the waiting time or the buffer is at max capacity
+                # flush the partition if the producer is running beyond the waiting time 
+                # or the buffer is at max capacity
                 if (now_time - self._last_send_time > self._max_wait_time) or (
                     self._cur_buffered_len >= self._max_buffer_len
                 ):
--- a/sdk/eventhub/azure-eventhub/azure/eventhub/_buffered_producer/_buffered_producer.py
+++ b/sdk/eventhub/azure-eventhub/azure/eventhub/_buffered_producer/_buffered_producer.py
@@ -30,7 +30,6 @@ class BufferedProducer:
             executor: ThreadPoolExecutor,
             *,
             max_wait_time: float = 1,
-            max_concurrent_sends: int = 1,
             max_buffer_length: int = 1500
     ):
         self._buffered_queue: queue.Queue = queue.Queue()
@@ -39,7 +38,6 @@ class BufferedProducer:
         self._executor: ThreadPoolExecutor = executor
         self._producer: EventHubProducer = producer
         self._lock = RLock()
-        self._max_concurrent_sends = max_concurrent_sends
         self._max_wait_time = max_wait_time
         self._on_success = self.failsafe_callback(on_success)
         self._on_error = self.failsafe_callback(on_error)
--- a/sdk/eventhub/azure-eventhub/azure/eventhub/_buffered_producer/_buffered_producer.py
+++ b/sdk/eventhub/azure-eventhub/azure/eventhub/_buffered_producer/_buffered_producer.py
@@ -57,6 +57,7 @@ class BufferedProducer:
                 self._check_max_wait_time_future = self._executor.submit(self.check_max_wait_time_worker)
 
     def stop(self, flush=True, timeout_time=None, raise_error=False):
+        
         self._running = False
         if flush:
             with self._lock:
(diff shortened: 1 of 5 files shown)


sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py
```diff
@@ -53,6 +54,9 @@ class EventHubProducerClient(ClientBase):   # pylint: disable=client-accepts-api
      or ~azure.core.credentials.AzureNamedKeyCredential
     :keyword bool buffered_mode: If True, the producer client will collect events in a buffer, efficiently batch,
      then publish. Default is False.
+    :keyword Union[ThreadPoolExecutor, int] buffer_concurrency: The ThreadPoolExecutor to be used for publishing events
+    or the number of workers for the ThreadPoolExecutor. Default is none and a ThreadPoolExecutor with the default number of workers
+    will be created https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ThreadPoolExecutor
```
annatisch, line 59: I think you need a proceeding space character on both lines 58 and 59 (i.e. multi-lines should be aligned to after the ":")

annatisch (commented):
Looking better - thanks!

sdk/eventhub/azure-eventhub/azure/eventhub/_buffered_producer/_buffered_producer.py
```diff
@@ -31,18 +31,15 @@ def __init__(
             *,
             max_wait_time: float = 1,
             max_concurrent_sends: int = 1,
-            max_buffer_length: int = 10
+            max_buffer_length: int = 1500
```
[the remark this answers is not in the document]
kashifkhan: yes, it gets defaulted at the dispatcher level.

sdk/eventhub/azure-eventhub/azure/eventhub/_buffered_producer/_buffered_producer.py
```diff
@@ -31,18 +31,15 @@ def __init__(
             *,
             max_wait_time: float = 1,
             max_concurrent_sends: int = 1,
```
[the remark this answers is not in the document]
kashifkhan: no we don't need it anymore and definitely not the producer level.

sdk/eventhub/azure-eventhub/azure/eventhub/_buffered_producer/_buffered_producer.py
```diff
@@ -147,64 +134,57 @@ def flush(self, timeout_time=None, raise_error=True):
...
+                _LOGGER.info(
+                    "Partition %r fails to flush due to timeout.",
+                    self.partition_id
+                )
+                if raise_error:
+                    raise OperationTimeoutError("Failed to flush {!r}".format(self.partition_id))
+                break
         # after finishing flushing, reset cur batch and put it into the buffer
         self._last_send_time = time.time()
         self._cur_batch = EventDataBatch(self._max_message_size_on_link)
```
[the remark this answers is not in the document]
kashifkhan: Yes we do, Ill remove the reset on line 139 and leave this one so it always resets since that is the case as it is

sdk/eventhub/azure-eventhub/azure/eventhub/_buffered_producer/_buffered_producer_dispatcher.py
```diff
@@ -46,9 +45,15 @@ def __init__(
...
-        self._existing_executor = bool(executor)
-        self._executor = executor or ThreadPoolExecutor(max_worker)
+        self._existing_executor = False
+
+        if not executor:
+            self._executor = ThreadPoolExecutor(max_worker)
+        elif isinstance(executor, ThreadPoolExecutor):
+            self._executor = executor
+        elif isinstance(executor, int):
+            self._existing_executor = True
```
annatisch, line 55: This should be false here, the only scenario where it would be true is line 52, i.e. the customer instantiated the ThreadPoolExecutor themselves. The reason for tracking this is so that we don't shutdown the executor if we didn't own it to start with (i.e. line 146)

sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py
```diff
@@ -234,7 +237,7 @@ def _buffered_send(self, events, **kwargs):
                 max_wait_time=self._max_wait_time,
                 max_buffer_length=self._max_buffer_length,
                 executor=self._executor,
-                max_worker=self._max_worker
+                max_worker=self._max_worker,
```
annatisch, line 240: Looks like passing in max_worker will always be None - so we could probably remove this parameter entirely?

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_buffered_producer/_buffered_producer_async.py
```diff
@@ -147,69 +127,50 @@ async def flush(self, timeout_time=None, raise_error=True):
...
-                await self.flush(raise_error=False)
+            if self._max_buffer_len > 0:
+                now_time = time.time()
+                _LOGGER.info("Partition %r worker is checking max_wait_time.", self.partition_id)
+                # flush the partition if its beyond the waiting time or the buffer is at max capacity
+                if (now_time - self._last_send_time > self._max_wait_time and self._running) or (
+                    self._cur_buffered_len >= self._max_buffer_len and self._running
+                ):
+                    # in the worker, not raising error for flush, users can not handle this
+                    await self.flush(raise_error=False)
```
annatisch, line 173: Missing the lock here?

annatisch (approved):
Just a few small comments - it's looking really good thanks :)

PR #24548, opened 2022-05-20
add missing type hints for ItemPaged

kristapratico:
The async implementation has type hints, but it was missing/type commented for sync.


tjprescott (approved):

iscai-msft (approved):
should we add a changelog entry?

xiangyan99 (approved):

johanste (approved):

Issue #24399, opened 2022-05-11
Investigate all skipped Storage tests


jalauzon-msft:
We have quite a few tests marked with `pytest.skip`, `pytest.mark.skip`, or `pytest.mark.skipif` that have various comments about past Issues or things that need to be "investigated later". Some have GitHub issues mentioned in the comments and some do not. It looks like some were also skipped because they don't run Live/Recorded but we have other mechanisms for that. Ignored tests have a huge potential to let regressions slip through and ideally, we should not have any.

We need to audit all instances of skipped tests and ideally eliminate them. My hunch is that many of them will just work now but those that don't we should fix.

To find tests, search the Storage packages for `pytest.skip`, `pytest.mark.skip`, or `pytest.mark.skipif`.

✔Checkmark = Resolved
☐ No checkmark = **Not resolved, comment to the right**

**pytest.skip**
- [x] test_copy_blob_with_blob_tier_specified **| Skipped because doesn't work on premium, but test case works fine with standard account (begs the Q: where was premium account implied [and the Err msg is about Block blob not prem. tier]**
- [x] test_copy_blob_with_rehydrate_priority **| Skipped because doesn't work on premium, but works fine with standard account**
- [x] test_list_blobs_include_deletedwithversion_async **| Pytest skip comment was already commented out, just removing now**
- [x] test_list_blobs_include_deletedwithversion **| Pytest skip comment was already commented out, just removing now**
- [x] test_list_blobs_with_include_metadata **| Pytest skip comment was already commented out, just removing now**
- [x] test_list_blobs_include_deletedwithversion **| Pytest skip comment was already commented out, just removing now**
- [x] test_list_blobs_with_include_multiple **| Metadata XML Fix in MSRest, removed skip**
- [x] test_unicode_get_messages_unicode_data **| Metadata XML Fix in MSRest, removed skip**
- [x] test_unicode_update_message_unicode_data **| Metadata XML Fix in MSRest, removed skip**
- [x] test_no_server_encryption (async) **| Resolved, removed AiohttpTestTransport workaround**
- [x] test_upload_large_stream_without_network (async) **| Continue skipping, changed to decorator**
- [x] test_upload_large_stream_without_network **| Continue skipping, changed to decorator**
- [x] test_rename_file_with_account_sas **| Continue to skip, requires further investigation**
- [ ] test_rename_directory_to_non_empty_directory **| Continue to skip, requires further investigation**
[...]

PR #24550, opened 2022-05-20
[Storage] Triaging Skipped Storage Tests (33/54)

vincenttran-msft:
This PR completes the work of triaging several skipped storage tests for various reasons. All details can be found in #24399. A follow-up is required to complete triaging of the remaining 21 tests, but this PR lays the groundwork for getting towards the goal of completely triaging and resolving all currently skipped test in our testing pipeline.

Remaining skipped Storage test tracker: #24552

### DIFF
--- a/sdk/storage/azure-storage-blob/tests/test_container.py
+++ b/sdk/storage/azure-storage-blob/tests/test_container.py
@@ -1811,7 +1811,6 @@ class StorageContainerTest(StorageTestCase):
     @BlobPreparer()
     def test_list_blobs_with_include_multiple(self, storage_account_name, storage_account_key):
         bsc = BlobServiceClient(self.account_url(storage_account_name, "blob"), storage_account_key)
-        # pytest.skip("Waiting on metadata XML fix in msrest")
         container = self._create_container(bsc)
         data = b'hello world'
         blob1 = container.get_blob_client('blob1')
--- a/sdk/storage/azure-storage-blob/tests/test_container.py
+++ b/sdk/storage/azure-storage-blob/tests/test_container.py
@@ -961,7 +961,6 @@ class StorageContainerTest(StorageTestCase):
                          'application/octet-stream')
         self.assertIsNotNone(blobs[0].creation_time)
 
-    @pytest.mark.skipif(sys.version_info < (3, 0), reason="quote and unquote behave differently on py2 for recordings")
     @BlobPreparer()
     def test_list_encoded_blobs(self, storage_account_name, storage_account_key):
         bsc = BlobServiceClient(self.account_url(storage_account_name, "blob"), storage_account_key)
@@ -1087,7 +1086,6 @@ class StorageContainerTest(StorageTestCase):
     @BlobPreparer()
     def test_list_blobs_with_include_metadata(self, storage_account_name, storage_account_key):
         bsc = BlobServiceClient(self.account_url(storage_account_name, "blob"), storage_account_key)
-        # pytest.skip("Waiting on metadata XML fix in msrest")
         container = self._create_container(bsc)
         data = b'hello world'
         content_settings = ContentSettings(
@@ -1116,7 +1114,6 @@ class StorageContainerTest(StorageTestCase):
     @BlobPreparer()
     def test_list_blobs_include_deletedwithversion(self, versioned_storage_account_name, versioned_storage_account_key):
         bsc = BlobServiceClient(self.account_url(versioned_storage_account_name, "blob"), versioned_storage_account_key)
-        # pytest.skip("Waiting on metadata XML fix in msrest")
         container = self._create_container(bsc)
         data = b'hello world'
         content_settings = ContentSettings(
@@ -1814,7 +1811,7 @@ class StorageContainerTest(StorageTestCase):
     @BlobPreparer()
     def test_list_blobs_with_include_multiple(self, storage_account_name, storage_account_key):
         bsc = BlobServiceClient(self.account_url(storage_account_name, "blob"), storage_account_key)
-        pytest.skip("Waiting on metadata XML fix in msrest")
+        # pytest.skip("Waiting on metadata XML fix in msrest")
         container = self._create_container(bsc)
         data = b'hello world'
         blob1 = container.get_blob_client('blob1')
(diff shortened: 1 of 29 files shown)


sdk/storage/azure-storage-blob/tests/test_container.py
```diff
@@ -1814,7 +1811,7 @@ def recursive_walk(prefix):
     @BlobPreparer()
     def test_list_blobs_with_include_multiple(self, storage_account_name, storage_account_key):
         bsc = BlobServiceClient(self.account_url(storage_account_name, "blob"), storage_account_key)
-        pytest.skip("Waiting on metadata XML fix in msrest")
+        # pytest.skip("Waiting on metadata XML fix in msrest")
```
jalauzon-msft, line 1814: Was this intentional? Seems odd the other tests would work but this one would not if whatever "fix" was fixed.

jalauzon-msft, line 1814: Was this intentional? Seems odd the other tests would work but this one would not if whatever "fix" was fixed.

sdk/storage/azure-storage-blob/tests/test_page_blob.py
```diff
@@ -1792,7 +1792,7 @@ def test_create_blob_with_md5_large(self, storage_account_name, storage_account_
 
         # Assert
 
-    @pytest.mark.skip(reason="Failing live test https://github.com/Azure/azure-sdk-for-python/issues/10473")
+    # @pytest.mark.skip(reason="Failing live test https://github.com/Azure/azure-sdk-for-python/issues/10473")
```
jalauzon-msft, line 1795: Remove?

jalauzon-msft, line 1795: Remove?

sdk/storage/azure-storage-file-datalake/tests/test_file.py
```diff
@@ -859,10 +859,10 @@ def test_rename_file_with_file_sas(self, datalake_storage_account_name, datalake
         self.assertEqual(data, data_bytes)
         self.assertEqual(new_client.path_name, "newname")
 
+    @pytest.mark.skip(reason="Service bug")
```
jalauzon-msft, line 862: Do we know what the bug is? Or does this still just need more investigation?

jalauzon-msft, line 862: Do we know what the bug is? Or does this still just need more investigation?

sdk/storage/azure-storage-file-datalake/tests/test_file_system.py
```diff
@@ -203,9 +203,9 @@ def test_rename_file_system(self, datalake_storage_account_name, datalake_storag
             self.dsc._rename_file_system(name="badfilesystem", new_name="filesystem")
         self.assertEqual(new_name, new_filesystem.get_file_system_properties().name)
 
+    @pytest.mark.skip(reason="Feature not yet enabled. Record when enabled.")
```
jalauzon-msft, line 206: Hmm, at this point I find it hard to believe it will be enabled if it's not yet. Possibly this feature was cut and never actually shipped. We should check with Sean probably.

jalauzon-msft, line 206: Hmm, at this point I find it hard to believe it will be enabled if it's not yet. Possibly this feature was cut and never actually shipped. We should check with Sean probably.

sdk/storage/azure-storage-file-share/tests/test_handle_async.py
```diff
@@ -61,10 +61,11 @@ def _validate_handles(self, handles):
         self.assertIsNotNone(handles[0].client_ip)
         self.assertIsNotNone(handles[0].open_time)
 
+    @pytest.mark.skip(reason="Needs further investigation.")
```
jalauzon-msft, line 64: I don't see this on your list for further investigation :). If you don't plan to investigate it now, we should at least open an Issue to track all the "needs further investigation" skips.

jalauzon-msft, line 64: I don't see this on your list for further investigation :). If you don't plan to investigate it now, we should at least open an Issue to track all the "needs further investigation" skips.

sdk/storage/azure-storage-blob/tests/test_page_blob.py
vincenttran-msft: Good catch (and duplicated across async version). Originally cited as failing for a different reason, GH Issue marked as resolved, but now failing for `TypeError: Session.request() got an unexpected keyword argument 'seal_blob'`. Going to mark as needs further investigation for now and will be caught in 2nd pass-through.

sdk/storage/azure-storage-file-datalake/tests/test_file.py
vincenttran-msft: Added more descriptive skip; was originally a `pytest.skip` inside so I just moved it outside to use the decorator `pytest.mark.skip`. No idea on what the bug is unfortunately 😞

sdk/storage/azure-storage-file-datalake/tests/test_file_system.py
vincenttran-msft: Will leave this open for now and check back with Sean tomorrow when he is back 😄

sdk/storage/azure-storage-file-share/tests/test_handle_async.py
vincenttran-msft: Thanks, I have a list of the remaining ones #24552, but this slipped through the cracks 😞 added now :)

jalauzon-msft (approved):
Thanks for tackling this! A great step towards better test coverage!

sdk/storage/azure-storage-file-datalake/tests/test_file_system.py
vincenttran-msft: Update: This feature has been problematic and is pending to be enabled (possibly) in the future.. so continuing to skip for now 😄

PR #24561, opened 2022-05-23
[Cosmos] add post_trigger_include to async read_item method and docs updates

simorenoh:
Small gap was missed with this parameter, adding it in this PR so functionality mirrors the sync client's read_item method.

Also has some small changes to the comments, including the Dict types.
Also has other small changes that were missing, like making cosmos_client.py private.


sdk/cosmos/azure-cosmos/azure/cosmos/aio/_container.py
```diff
@@ -161,22 +161,16 @@ async def create_item(
...
-        :keyword enable_automatic_id_generation: Enable automatic id generation if no id present.
-        :paramtype enable_automatic_id_generation: bool
-        :keyword session_token: Token for use with Session consistency.
-        :paramtype session_token: str
-        :keyword initial_headers: Initial headers to be sent as part of the request.
-        :paramtype initial_headers: dict[str,str]
+        :paramtype indexing_directive: Union[int, ~azure.cosmos.documents.IndexingDirective]
+        :keyword bool enable_automatic_id_generation: Enable automatic id generation if no id present.
+        :keyword str session_token: Token for use with Session consistency.
+        :keyword Dict[str, str] initial_headers: Initial headers to be sent as part of the request.
```
tjprescott, line 171: Should this be `dict[str, str]` to make sphinx happy?

sdk/cosmos/azure-cosmos/azure/cosmos/aio/_database.py
```diff
@@ -132,7 +132,7 @@ async def read(self, **kwargs: Any) -> Dict[str, Any]:
         :rtype: Dict[str, Any]
         """
         # TODO this helper function should be extracted from CosmosClient
```
tjprescott, line 134: Think this TODO can be deleted now?

simorenoh: We're still using the CosmosClient method in the line right below (137) so not yet unfortunately.

tjprescott (approved):

PR #24581, opened 2022-05-24
[pylint] removing ignore-paths

l0lawrence:
removing the ignore-path from the pylintrc file as they have bypassed api keyword for now


catalinaperalta:
Just double checking actually, the core ci passes still?

PR #24582, opened 2022-05-24
[servicebus] uamqp custom endpoint support

### LINKED ISSUE #24152: Service Bus: Custom Endpoint Address Support
# Summary

When using an Express Route circuit, there are some configurations which are not able to open the initial connection to the Service Bus service using the fully qualified namespace as the connection endpoint.  In these cases, a custom endpoint address is needed to ensure the correct routing for the connection but should only be used when first establishing the connection; after that point, the fully qualified namespace is expected by the service for normal operations.  

In order to support this scenario, an approach is needed to allow for the endpoint address for connections to be overridden with an explicit value, taking precedence over the default behavior of assuming the fully qualified namespace. 

# Scope of Work

- Implement the design for supporting a custom endpoint address.

# Success Criteria

- The design has been implemented and the targeted clients are able to provide a custom endpoint address for connecting to the Service Bus service which is respected by the client library types.

- The tests necessary for its validation have been created or adjusted and pass reliably.

- The existing test suite continues to produce deterministic results and pass reliably.

# Related Issues and References

- Design Proposal for Custom Endpoint Address Support (https://gist.github.com/jsquire/f0948c9a5ffe2741b03d312dcf8d899e)
- [Pull Request] Event Hubs: Support for Custom endpoint address and custom certificate ([#16295](https://github.com/Azure/azure-sdk-for-python/pull/16295))

l0lawrence:
closes #24152


sdk/servicebus/azure-servicebus/CHANGELOG.md
```diff
@@ -6,6 +6,15 @@
 
 - Fixed bug to make AMQP exceptions retryable by default, if condition is not non-retryable, to ensure that InternalServerErrors are retried.
 
+### Features Added
+
+- The `ServiceBusClient` constructor now accepts optional `custom_endpoint_address` argument
+which allows for specifying a custom endpoint to use when communicating with the Event Hubs service,
```
annatisch, line 12: Service Bus service? :)

sdk/servicebus/azure-servicebus/azure/servicebus/_common/_configuration.py
```diff
@@ -30,3 +31,24 @@ def __init__(self, **kwargs):
...
+            self.connection_port = DEFAULT_AMQP_WSS_PORT
+
+        # custom end point
+        if self.custom_endpoint_address:
+            # if the custom_endpoint_address doesn't include the schema,
+            # we prepend a default one to make urlparse work
+            if self.custom_endpoint_address.find("//") == -1:
+                self.custom_endpoint_address = "sb://" + self.custom_endpoint_address
+            endpoint = urlparse(self.custom_endpoint_address)
+            self.transport_type = TransportType.AmqpOverWebsocket
```
annatisch, line 51: If someone has explicitly set the transport type to AMQP (as in, actually passed in that value, rather than just the default), and also supplied a custom_endpoint_address - we might want to raise a ValueError

sdk/servicebus/azure-servicebus/CHANGELOG.md
l0lawrence: Ah! Sorry, Anna!

sdk/servicebus/azure-servicebus/azure/servicebus/_common/_configuration.py
annatisch: Discussed OOB - we'll address this separately - as it may also apply to http_proxy if the change is non-breaking.

Issue #20282, opened 2021-08-14
`azure.data.tables` `query_entities()` JSON example?


SeaDude:
Hello,

I need to create a JSON object or array from the `table_client.query_entities()` response. All of the examples I've found,  use `print()` ([example](https://github.com/Azure/azure-sdk-for-python/blob/c2524442d71320a9e020cf6f8d7f758be86f3fea/sdk/tables/azure-data-tables/samples/sample_query_table.py#L105)): 

**Instead of something like:**

```
for entity in list(table_client.query_entities(query)):
    print(entity)
```

**How do I get a JSON object from the `<class 'azure.core.paging.ItemPaged'>` response**?

SeaDude:
This gets close-ish, but returns `timedate` data. 

```
test = []
for entity in list(table_client.query_entities(query)):
    test.append(entity)
```

**Example:** 
- `'dl_end_time': TablesEntityDatetime(2021, 8, 10, 0, 23, 19, tzinfo=datetime.timezone.utc)`

SeaDude:
I _think_ this is correct. Found the [Azure Data Table REST API docs](https://docs.microsoft.com/en-us/rest/api/storageservices/query-entities#request-headers) and added the `nometadata` header to a Python SDK request. 

```
test = []
for entity in table_client.query_entities(query, headers ={'Accept' : 'application/json;odata=nometadata'}):
    test.append(entity)

test = json.dumps(test)
```

Can someone validate this is the correct way to return clean JSON?

SeaDude:
Hm. The above doesn't quite work because one of the columns in the data table is JSON itself. 

All of the `"` 's in the column are escaped and not easily parsed. 

```
[...]

PR #24607, opened 2022-05-26
Update query_entities() samples

YalinLi0312:
For issue: https://github.com/Azure/azure-sdk-for-python/issues/20282

Add headers in query_entities() to get a cleaner JSON response.

### DIFF
--- a/sdk/tables/azure-data-tables/samples/sample_query_table.py
+++ b/sdk/tables/azure-data-tables/samples/sample_query_table.py
@@ -68,27 +68,81 @@ class SampleTablesQuery(object):
         from azure.data.tables import TableClient
         from azure.core.exceptions import HttpResponseError
 
+        print("Entities with name: marker")
         # [START query_entities]
         with TableClient.from_connection_string(self.connection_string, self.table_name) as table_client:
             try:
-                print("Entities with name: marker and brand: Crayola")
-                queried_entities1 = table_client.query_entities(
-                    query_filter={"name": "marker", "brand": "Crayola"},
-                    select=["Brand", "Color"],
-                    parameters="Name eq @name and Brand eq @brand",
-                    headers ={"Accept" : "application/json;odata=nometadata"}
+                parameters = {"name": "marker"}
+                name_filter = "Name eq @name"
+                queried_entities = table_client.query_entities(
+                    query_filter=name_filter, select=["Brand", "Color"], parameters=parameters
                 )
-                for entity_chosen in queried_entities1:
+
+                for entity_chosen in queried_entities:
                     print(entity_chosen)
 
-                print("Entities with 25 < Value < 50")
-                queried_entities2 = table_client.query_entities(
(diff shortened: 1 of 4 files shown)


sdk/tables/azure-data-tables/samples/sample_query_table.py
```diff
@@ -139,7 +107,5 @@ def clean_up(self):
     try:
         stq.insert_random_entities()
         stq.sample_query_entities()
-        stq.sample_query_entities_multiple_params()
-        stq.sample_query_entities_values()
```
annatisch, line 143: Why were these removed?

sdk/tables/azure-data-tables/samples/sample_query_table.py
```diff
@@ -68,59 +68,27 @@ def sample_query_entities(self):
...
-                parameters = {"name": "marker"}
-                name_filter = "Name eq @name"
-                queried_entities = table_client.query_entities(
-                    query_filter=name_filter, select=["Brand", "Color"], parameters=parameters
+                print("Entities with name: marker and brand: Crayola")
+                queried_entities1 = table_client.query_entities(
+                    query_filter={"name": "marker", "brand": "Crayola"},
+                    select=["Brand", "Color"],
+                    parameters="Name eq @name and Brand eq @brand",
+                    headers ={"Accept" : "application/json;odata=nometadata"}
```
annatisch, line 79: We should not add this to an existing sample - let's leave the existing samples untouched as adding this header should be very uncommon.
I would add a new method called something like `sample_query_entities_without_metadata` and add it the bottom of the file.

sdk/tables/azure-data-tables/samples/sample_query_table.py
```diff
@@ -68,59 +68,27 @@ def sample_query_entities(self):
...
             try:
-                parameters = {"name": "marker"}
-                name_filter = "Name eq @name"
-                queried_entities = table_client.query_entities(
-                    query_filter=name_filter, select=["Brand", "Color"], parameters=parameters
+                print("Entities with name: marker and brand: Crayola")
+                queried_entities1 = table_client.query_entities(
+                    query_filter={"name": "marker", "brand": "Crayola"},
+                    select=["Brand", "Color"],
+                    parameters="Name eq @name and Brand eq @brand",
```
annatisch, line 78: This is incorrect - the `query_filter` and `parameters` values are backwards.

sdk/tables/azure-data-tables/samples/sample_query_table.py
YalinLi0312: I moved these three functions into one.

sdk/tables/azure-data-tables/samples/sample_query_table.py
YalinLi0312: Updated, added a new function for `no metadata` headers.

annatisch (approved):

PR #24620, opened 2022-05-26
Disable Verify Dependencies Step for azure-ml (in code)

scbedd:
We disabled this check to get them into main, but that same variable is not set for everyone else.

@iscai-msft @kristapratico my bad!


scbedd:
This is a local run. 

image

Given that all PRs are broken by this check, I am going to override and merge.

hallipr (approved):

Issue #24690, opened 2022-06-02
[Storage] Investigate test_quick_query_output_in_arrow_format


vincenttran-msft:
This has been a long withstanding issue that keeps coming up in the pipeline every once and awhile (due to changes on the Service end). For the time being, the test only does the following:
- **Test that the Service accepts our SQL-like query** 
- **Test that our SQL-like query is well-formatted and does not return an exception**

We would like to eventually extend the test to also do the following:
- **Verify the output matches the expected output.** Due to changes in the Service end, byte-by-byte comparison has proven problematic. To alleviate this, we should probably get a better understanding of what this API call does so that we can compare the properties of the output rather than byte by byte (as any difference may just be due to formatting of the output, not an actual difference in the data being communicated).

PR #24691, opened 2022-06-02
[Storage] Modify test_quick_query_output_in_arrow_format assert

vincenttran-msft:
This has been back and forth and keeps causing red in our release pipeline, so we are going to comment this out for now (since any changes in Service can cause this fragile flaky test to break). Will be further investigated at a later date to figure out a better way to check the output rather than byte-by-byte comparison (tracked in #24690)


jalauzon-msft (approved):
Be gone, flaky test!

PR #24693, opened 2022-06-03
[Storage] Add client-side encryption V2 for Queue

jalauzon-msft:
Adding a new version of client-side encryption supporting the use of AES-GCM.


sdk/storage/azure-storage-blob/azure/storage/blob/_shared/encryption.py
```diff
@@ -115,23 +154,32 @@ def __init__(self, content_encryption_IV, encryption_agent, wrapped_content_key,
...
         _validate_not_none('encryption_agent', encryption_agent)
         _validate_not_none('wrapped_content_key', wrapped_content_key)
 
+        # Validate we have the right info for the specified algorithm
+        if encryption_agent.encryption_algorithm == _EncryptionAlgorithm.AES_CBC_256:
+            _validate_not_none('content_encryption_IV', content_encryption_IV)
+        elif encryption_agent.encryption_algorithm == _EncryptionAlgorithm.AES_GCM_256:
+            _validate_not_none('encrypted_region_info', encrypted_region_info)
+        else:
+            raise ValueError("Invalid encryption algorithm.")
```
vincenttran-msft, line 166: This is probably fine, but just a thought -- maybe a more specific ValueError? Since technically they could have specified a valid encryption algorithm but just didn't provide the matching required parameter (but the ValueError suggests otherwise)

sdk/storage/azure-storage-blob/azure/storage/blob/_shared/encryption.py
```diff
@@ -115,23 +154,32 @@ def __init__(self, content_encryption_IV, encryption_agent, wrapped_content_key,
             A dict containing metadata related to the key wrapping.
         '''
 
-        _validate_not_none('content_encryption_IV', content_encryption_IV)
         _validate_not_none('encryption_agent', encryption_agent)
         _validate_not_none('wrapped_content_key', wrapped_content_key)
 
+        # Validate we have the right info for the specified algorithm
```
vincenttran-msft, line 160: Nit: Mostly no one is looking here anyways, but maybe we can put "matching required parameter"

sdk/storage/azure-storage-blob/azure/storage/blob/_shared/encryption.py
```diff
@@ -101,11 +131,20 @@ class _EncryptionData:
...
+            content_encryption_IV,
+            encrypted_region_info,
+            encryption_agent,
+            wrapped_content_key,
+            key_wrapping_metadata):
         '''
-        :param bytes content_encryption_IV:
+        :param Optional[bytes] content_encryption_IV:
             The content encryption initialization vector.
+            Required for AES-CBC.
```
vincenttran-msft, line 144: Nit: After looking futher in the review, I see that the correspondence for selecting between AES-CBC and AES-GCM is whether `version` is V1 or V2. May be helpful to sort of make this obvious to cx when referring to the algorithms, something like this:
AES-CBC (V1)
AES-GCM (V2)
So it is painfully obvious whenever you decide to refer to V1 or V2 what that exactly means!

sdk/storage/azure-storage-queue/azure/storage/queue/_queue_client.py
```diff
@@ -477,10 +476,20 @@ def send_message(
...
-            resolver=self.key_resolver_function)
+        try:
+            self._config.message_encode_policy.configure(
+                require_encryption=self.require_encryption,
+                key_encryption_key=self.key_encryption_key,
+                resolver=self.key_resolver_function,
+                encryption_version=self.encryption_version)
+        except TypeError:
+            warnings.warn(
+                "message_encode_policy.configure is likely missing the encryption_version parameter."
```
vincenttran-msft, line 487: I like this nuanced warning! Perhaps something more general at the start (just to let them know generally what happened, just in case `message_encode_policy.configure` is not their issue). 
Not 100% sure what to put here so open to your input, but something like "Encryption failed" or whatever happened, and then this warning would be nice.

sdk/storage/azure-storage-blob/azure/storage/blob/_shared/encryption.py
jalauzon-msft: This `else` actually DOES mean the actual encryption algorithm was invalid, not that we had missing parameters for it, so I think this error is appropriate. Also, this is an internal method, and I don't think a customer would ever actually see this, more here to be extra cautious, I guess. A customer would have to modify their metadata in such a way to get this.

sdk/storage/azure-storage-blob/azure/storage/blob/_shared/encryption.py
jalauzon-msft: Sure, I can do that. This is an internal only method that a customer would never see/call but I can add it to help us remember :)

sdk/storage/azure-storage-queue/azure/storage/queue/_queue_client.py
jalauzon-msft: Yeah, that's a good suggestion. Will update.

sdk/storage/azure-storage-queue/azure/storage/queue/_shared/base_client.py
```diff
@@ -104,6 +104,7 @@ def __init__(
             self._hosts = {LocationMode.PRIMARY: primary_hostname, LocationMode.SECONDARY: secondary_hostname}
 
         self.require_encryption = kwargs.get("require_encryption", False)
+        self.encryption_version = kwargs.get("encryption_version", "1.0")
```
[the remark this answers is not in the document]
jalauzon-msft: This feature is completely undocumented for Queue and I think we may be leaving it like that. I think we will be adding a warning if the user configures encryption but does not set this  to 2.0. Still sort of an open dicussion on that though so will leave it out for now.

@annatisch curious on your thoughts of adding a warning here if the user adds a key_encryption_key (enables encryption) but does not set this version to 2.0? Sort of a deprecation warning for 1.0. This may be something we want to do after preview? There is an open mail thread on this as well.

sdk/storage/azure-storage-queue/azure/storage/queue/_queue_client.py
```diff
@@ -477,10 +476,22 @@ def send_message(
...
+            self._config.message_encode_policy.configure(
+                require_encryption=self.require_encryption,
+                key_encryption_key=self.key_encryption_key,
+                resolver=self.key_resolver_function,
+                encryption_version=self.encryption_version)
+        except TypeError:
+            warnings.warn(
+                "TypeError when calling message_encode_policy.configure. \
+                It is likely missing the encryption_version parameter. \
+                Retrying without encryption_version."
```
annatisch, line 489: Maybe add a message here to recommend they upgrade their encryption implementation?

sdk/storage/azure-storage-queue/azure/storage/queue/_queue_client.py
```diff
@@ -711,10 +722,21 @@ def update_message(self,
...
-                self.key_resolver_function)
+            try:
+                self._config.message_encode_policy.configure(
+                    self.require_encryption,
+                    self.key_encryption_key,
+                    self.key_resolver_function,
+                    encryption_version=self.encryption_version)
+            except TypeError:
+                warnings.warn(
+                    "message_encode_policy.configure does not accept encryption_version parameter.",
```
annatisch, line 733: Why is this warning different to the one above?

sdk/storage/azure-storage-queue/azure/storage/queue/_shared/base_client.py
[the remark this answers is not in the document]
annatisch: Yes - we should add a warning if customers are passing in the other encryption parameters but not setting the version to 2.0.

annatisch (commented):
Looking good to me! Just a few minor things :)

sdk/storage/azure-storage-queue/azure/storage/queue/_queue_client.py
jalauzon-msft: Oops, I had updated this at one point and missed the second one. Will update.

sdk/storage/azure-storage-queue/azure/storage/queue/_shared/base_client.py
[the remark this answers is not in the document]
jalauzon-msft: I've added a warning.

PR #24705, opened 2022-06-03
Add Build Definition For `Azure-CLI` tests

### LINKED ISSUE #24722: Create Integrated `Run CLI Tests` Build
- [x] Choose Python Version
- [x] Choose which packages from the azure-sdk-for-python repo should be installed
- [x] Choose which additional packages should be installed
- [x] Invoke CLI recorded CLI tests

scbedd:
Running this build allows us to invoke the CLI tests against any package within our repo. Also parametrized are:

- `PythonVersion`: which python version is used
- `TargetModules`: which cli modules to invoke tests for
- `TargetRepoPackages`: list of repo paths that should be installed`
- `InjectedPackages`: list of pip installable requirements. These can be a VCS url, a raw URI to a blob stored wheel, a python version specifier, or any other pip installable component.

Mac/Windows both hit issues in the azdev_setup.yml. More work is necessary to enable those. Probably will need to roll our own setup code if we want crossplat.

Resolves #24722 24722

### DIFF
--- a/eng/pipelines/templates/jobs/run-cli-tests.yml
+++ b/eng/pipelines/templates/jobs/run-cli-tests.yml
@@ -75,3 +75,7 @@ jobs:
       - template: /eng/pipelines/templates/steps/use-python-version.yml
         parameters:
           versionSpec: '${{ parameters.PythonVersion }}'
+
+      - template: .azure-pipelines/templates/azdev_setup.yml@AzureCLI
+        parameters:
+          CLIRepoPath: $(Build.SourcesDirectory)/cli
--- a/eng/pipelines/templates/jobs/run-cli-tests.yml
+++ b/eng/pipelines/templates/jobs/run-cli-tests.yml
@@ -95,5 +95,5 @@ jobs:
           source env/bin/activate
 
           serial_modules="appservice botservice cloud network azure-cli-core azure-cli-telemetry"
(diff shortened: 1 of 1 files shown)


eng/pipelines/templates/jobs/run-cli-tests.yml
```diff
@@ -0,0 +1,87 @@
...
+    # - <packagename>==<version>
+  - name: InjectedPackages
+    type: object
+    default: []
+  - name: PythonVersion
+    type: string
+    default: "3.9"
+  - name: TargetModules
+    type: string
+    default: "appservice botservice cloud network azure-cli-core azure-cli-telemetry"
```
scbedd, line 25: This is the default list from [here](https://github.com/Azure/azure-cli/blob/cb3793acfabada7815fae83686d2b6324fa062a3/.azure-pipelines/templates/automation_test.yml#L43).

eng/pipelines/templates/jobs/run-cli-tests.yml
```diff
@@ -0,0 +1,87 @@
+resources:
+  repositories:
+    - repository: AzureCLI
+      type: github
+      name: azure/azure-cli
+      endpoint: Azure
+      ref: refs/heads/dev
```
weshaggard, line 7: Do we want a moving target or should this be a release tag?

PR #24706, opened 2022-06-03
[Storage] Add client-side encryption V2 to Blob SDK

jalauzon-msft:
Adding client-side encryption V2.0 with support for AES-GCM-256 encryption.

NOTE: This PR depends on #24693, and will contain changes from that PR until its merged.


sdk/storage/azure-storage-blob/azure/storage/blob/_shared/encryption.py
```diff
@@ -115,23 +158,244 @@ def __init__(self, content_encryption_IV, encryption_agent, wrapped_content_key,
...
+    will use the same encryption key and will generate a guaranteed unique
+    nonce for each encryption region.
+    """
+    def __init__(
+            self,
+            content_encryption_key: bytes,
+            data_stream: BinaryIO,
+            ):
+        """
+        :param content_encryption_key: The encryption key to use.
```
vincenttran-msft, line 192: Maybe because this is `bytes` and it is convention to not list `bytes` as the type -- but if not, perhaps it should be `:param bytes content_encryption_key`? Duplicate this comment for `data_stream` as well for possibly missing a paramtype?

sdk/storage/azure-storage-blob/azure/storage/blob/_shared/encryption.py
```diff
@@ -115,23 +158,244 @@ def __init__(self, content_encryption_IV, encryption_agent, wrapped_content_key,
...
+
+        self.offset = 0
+        self.current = b''
+        self.nonce_counter = 0
+
+    def read(self, size: int = -1) -> bytes:
+        """
+        Read data from the stream. Specify -1 to read all available data.
+
+        :param size: The amount of data to read. Defaults to -1 for all data.
```
vincenttran-msft, line 206: Similar comment to previous comment -- should this be `:param int size`?

sdk/storage/azure-storage-blob/azure/storage/blob/_shared/encryption.py
```diff
@@ -115,23 +158,244 @@ def __init__(self, content_encryption_IV, encryption_agent, wrapped_content_key,
...
+                data = self.data_stream.read(_GCM_REGION_DATA_LENGTH)
+                if len(data) == 0:
+                    # No more data to read
+                    break
+
+                self.current = self._encrypt_region(data)
+
+        return result.getvalue()
+
+    def _encrypt_region(self, data: bytes) -> bytes:
```
vincenttran-msft, line 232: Possibly missing a `:param bytes data:`?

sdk/storage/azure-storage-blob/azure/storage/blob/_shared/encryption.py
```diff
@@ -115,23 +158,244 @@ def __init__(self, content_encryption_IV, encryption_agent, wrapped_content_key,
...
+        # Returns ciphertext + tag
+        cipertext_with_tag = aesgcm.encrypt(nonce, data, None)
+        return nonce + cipertext_with_tag
+
+
+def is_encryption_v2(encryption_data: Optional[_EncryptionData]) -> bool:
+    """
+    Determine whether the given encryption data signifies version 2.0.
+
+    :param encryption_data: The encryption data. Will return False if this is None.
```
vincenttran-msft, line 252: Okay this is my last comment on this since it's recurring so I'm sure you get the gist 😛, but I see, is the convention here to list the paramtype above and then omit from docstring? If so then continue with that -- I think I am just unfamiliar with that pattern 😄

sdk/storage/azure-storage-blob/azure/storage/blob/_shared/encryption.py
```diff
@@ -115,23 +158,244 @@ def __init__(self, content_encryption_IV, encryption_agent, wrapped_content_key,
...
     '''
     Generates and returns the encryption metadata as a dict.
 
     :param object kek: The key encryption key. See calling functions for more information.
     :param bytes cek: The content encryption key.
-    :param bytes iv: The initialization vector.
+    :param Optional[bytes] iv: The initialization vector. Only required for AES-CBC.
+    :param str version: The client encryption version used.
     :return: A dict containing all the encryption metadata.
     :rtype: dict
```
vincenttran-msft, line 401: Nit: Just curious if API View will be happy with this `:rtype: dict` or if it would be expecting something like `:rtype: dict[Any,Any]`, but fine as-is

sdk/storage/azure-storage-blob/azure/storage/blob/_shared/encryption.py
```diff
@@ -187,7 +466,21 @@ def _dict_to_encryption_data(encryption_data_dict):
...
+    encryption_iv = None
+    if 'ContentEncryptionIV' in encryption_data_dict:
+        encryption_iv = decode_base64_to_bytes(encryption_data_dict['ContentEncryptionIV'])
+
+    # AES-GCM only
+    region_info = None
+    if 'EncryptedRegionInfo' in encryption_data_dict:
+        encrypted_region_info = encryption_data_dict['EncryptedRegionInfo']
+        region_info = _EncryptedRegionInfo(encrypted_region_info['EncryptedRegionDataLength'],
+                                          encrypted_region_info['NonceLength'],
```
vincenttran-msft, line 479: Super nit: Line 479 and 480 could be moved up 1 whitespace to line up with 478 😛

vincenttran-msft (commented):
Great work! LGTM, just some small nits that might come back to you from API View or Pylint 😄

sdk/storage/azure-storage-blob/azure/storage/blob/_shared/encryption.py
jalauzon-msft: I had left it out since this is a "private" method. Will add it anyway.

sdk/storage/azure-storage-blob/azure/storage/blob/_shared/encryption.py
jalauzon-msft: This is an internal, non-exposed method. Should be fine.

sdk/storage/azure-storage-blob/README.md
```diff
@@ -331,6 +331,8 @@ Defaults to `False`.
 Use the following keyword arguments when instantiating a client to configure encryption:
```
annatisch, line 331: Do we want to add a message here that we recommend _not_ using client-side encryption, and link to docs or a guide to migrate to server side encryption? I would also probably put a more detailed note here about the deprecation, you could even reuse the note in the changelog.

sdk/storage/azure-storage-blob/azure/storage/blob/_download.py
```diff
@@ -376,6 +381,21 @@ def __init__(
     def __len__(self):
         return self.size
 
+    def _get_encryption_data_request(self):
+        # Save current request cls
+        download_cls = self._request_options.pop('cls', None)
+        # Adjust cls for get_properties
+        self._request_options['cls'] = deserialize_blob_properties
```
annatisch, line 388: Modifying `self` makes this object not threadsafe - is there any possibility this method could be run concurrently?

sdk/storage/azure-storage-blob/azure/storage/blob/_download.py
```diff
@@ -376,6 +381,21 @@ def __init__(
...
+
+        properties = self._clients.blob.get_properties(**self._request_options)
```
annatisch, line 390: Looks like we decided on the pre-emptive `get_properties` approach - I'm happy so long as non-encryption path is unaffected. As we discussed - I don't think much sleep will be lost if the client-side encryption path is not so performant.

sdk/storage/azure-storage-blob/azure/storage/blob/aio/_download_async.py
```diff
@@ -231,26 +237,53 @@ def __init__(
...
+        # This will return None if there is no encryption metadata or there are parsing errors.
+        # That is acceptable here, the proper error will be caught and surfaced hen attempting
+        # to decrypt the blob.
+        self._encryption_data = parse_encryption_data(properties.metadata)
+
+        # Restore cls for download
+        self._request_options['cls'] = download_cls
+
+    async def _setup(self):
+        if self._encryption_options.get("key") is not None or self._encryption_options.get("resolver") is not None:
```
annatisch, line 270: Do we also need to check `requires_encryption=True`? Or is that assessed elsewhere?

sdk/storage/azure-storage-blob/azure/storage/blob/_shared/uploads.py
```diff
@@ -149,7 +159,6 @@ def __init__(
         self.parallel = parallel
 
         # Stream management
-        self.stream_start = stream.tell() if parallel else None
```
annatisch, line 152: We don't need this any more?

sdk/storage/azure-storage-blob/README.md
jalauzon-msft: Yeah, that might be a good idea. We are still internally discussing all messaging around this feature, and I will likely be changing CHANGELOG/warnings/etc. so I'll make another pass at this section.

sdk/storage/azure-storage-blob/azure/storage/blob/_download.py
jalauzon-msft: Yep, felt like the easiest/least error-prone approach. We can always revisit if this starts to cause a problem, but I think this should be acceptable. And yep, this will only be triggered for requests where the user has provided encryption info for decryption. There is of course the chance the blob is not actually encrypted in which case this is a useless call, but I think that's also justifiable if encryption information is provided.

The one note is this applies to V1 now as well which is a bit of a change in behavior/performance of V1. I don't think this should be a problem since we are deprecated V1 but let me know if you feel otherwise.

sdk/storage/azure-storage-blob/azure/storage/blob/_shared/uploads.py
jalauzon-msft: This `self.stream_start` was not used anywhere as far as I could tell. It's definitely not used internally, and I couldn't see anywhere outside of the class it was being used. It's also a private internal class. Seems like it was left over from some old implementation?
I needed to remove it since the new wrapper stream I created for encryption doesn't support seeking or `tell`.

sdk/storage/azure-storage-blob/azure/storage/blob/aio/_download_async.py
jalauzon-msft: `requires_encryption` is sort of an odd option. In almost all cases its mostly ignored and whether we encrypt/decrypt is controlled by whether the user passes `key_encryption_key` or `key_resolver_function` on client construction. Even if you set this to `False` but provide a `key_encryption_key`, we would still perform encryption.

The only case this value is taken into account is if the blob isn't actually encrypted (no encryption metadata is found) or there is an error in parsing the metadata. If `requires_encryption` is `True` these cases will cause an exception, if its `False`, no exception will occur, and the blob will be treated as unencrypted.

So to actually answer your question, no I don't think we need/want to assess that value here since if someone sets this to `True` for some reason but doesn't provide the other arguments, we don't want to make this extra request.

Issue #24496, opened 2022-05-17
Inconsistent dependency is crashing `apistubgen` in specific circumstances


scbedd:
[Build Failure](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=1574557&view=logs&j=3dc8fd7e-4368-5a92-293e-d53cefc8c4b3&t=651a14cc-88df-515d-1311-e0e8b23b578a&l=166)

Unfortunately this error did not repro on the `core` build that tested the change that is causing this to throw. Investigating that.

scbedd:
This is turning out to be an issue specific to `azure-agrifood-farming`. This makes me think there is actually something going wrong, and we're just not certain of why.

Here's what I've investigated. 

It is _some_ interaction of our specific agent image, `tox`, and `astroid`.

How can we tell that it's a specific combination of problems? Well, the problem refuses to repro anywhere _except_ the CI agents. Manually created Ubuntu 20.04 VMs? Good. Local WSL? Still good. Export and re-use of venv site-packages from failing agent? **Still good.** The problem literally only manifests in _this specific use case on the build agent itself_.

- Ubuntu 20.04 agent invoking `tox -e apistub` **fails** [check any agrifood build to see a failure]
- Ubuntu 18.04 agent invoking `tox -e apistub` **works** [example build](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=1628254&view=logs&j=3dc8fd7e-4368-5a92-293e-d53cefc8c4b3&t=76d9b154-b861-59e4-d8e3-dfd06fc0fc10)
- Ubuntu 20.04 agent invoking via manual venv creation, install, and invoke **works in some cases** [example build](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=1628202&view=logs&j=3dc8fd7e-4368-5a92-293e-d53cefc8c4b3&t=44e6c678-cb87-52d9-8547-bcdbd0ad6ae4)
- Ubuntu 18.04 agent invoking via manual venv creation, install, and invoke **works**

Other things I've checked out on my way to the above:

- Trim Installation down to just apistubgen reqs from private feed 
- Add force reinstall to above installation
- Update tox call into apistubgen to directly call into the console entry point using `{envbindir}/python -m <blah>` to ensure that we aren't accidentally breaking out of the venv
[...]

PR #24725, opened 2022-06-06
Adjust APIStubGen Installation

scbedd:
Hopefully resolving #24496


eng/tox/tox.ini
```diff
@@ -326,8 +326,9 @@ changedir = {envtmpdir}
 setenv =
   {[testenv]setenv}
   PROXY_URL=http://localhost:5014
+  PIP_EXTRA_INDEX_URL=
 deps =
-  {[base]deps}
+  -rdev_requirements.txt
```
scbedd, line 331: dev_requirements, no longer including `test_tools.txt`.

The install of apiview_reqs.txt is where stub generator comes in.

eng/tox/run_apistubgen.py
```diff
@@ -63,7 +64,7 @@ def get_package_wheel_path(pkg_root):
     if not pkg_path:
         pkg_path = args.target_package
 
-    cmds = ["apistubgen", "--pkg-path", pkg_path]
```
scbedd, line 66: Perhaps the wrong executable is being called in? This ensures that we are using the correct `site-packages` folder by invoking the module from directly.

eng/pipelines/templates/steps/build-artifacts.yml
```diff
@@ -70,6 +70,10 @@ steps:
       twine check $(Build.ArtifactStagingDirectory)/**/*.zip
     displayName: 'Verify Readme'
 
+  - template: /eng/pipelines/templates/steps/run_apistub.yml
```
scbedd, line 73: I myself can't see why this APIStubGen step is failing. I'm moving it up above GenerateDocs just in case we're downloading a wrong version of `pylint` or the like earlier in the pipeline.

scbedd: [No luck](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=1622691&view=logs&j=3dc8fd7e-4368-5a92-293e-d53cefc8c4b3&t=76d9b154-b861-59e4-d8e3-dfd06fc0fc10)

scbedd:
We got green by swapping to `Ubuntu 18.04`.

Given that we won't be downgrading the build agent, I will be creating an alternative to `tox` that we can use in CI until upgrading to `Ubuntu 22.04`.

Going to close this PR in favor of a new one that doesn't include all the debug commits.

PR #24749, opened 2022-06-07
[formrecognizer] Remove document captions and footnotes

catalinaperalta:


sdk/formrecognizer/azure-ai-formrecognizer/azure/ai/formrecognizer/_models.py
```diff
@@ -3294,8 +3290,6 @@ def __init__(self, **kwargs):
         self.row_count = kwargs.get("row_count", None)
         self.column_count = kwargs.get("column_count", None)
         self.cells = kwargs.get("cells", None)
-        self.caption = kwargs.get("caption", None)
```
catalinaperalta, line 3297: per offline discussion, removing caption and footnote implementation, leaving the response hook as an option is ever needed to access these fields.

PR #24754, opened 2022-06-07
[Servicebus] message settlement and pyamqp

l0lawrence:
Main thing on this branch:

- I ran send_queue.py (only the single_message() method, message batch I have not set up send_messages() for SB pyamqp yet)
- On running receive_queue.py, after the last outgoing_flow frame, we wait for Transfer frames back from the Service. These are never received. 

Tested Scenarios:

- send_queue.py running on pyamqp, confirm messages sent, then receive_queue.py on uamqp -> WORKS 
- send_queue.py running on uamqp,  confirm messages sent, then receive_queue.py on pyamqp -> DOESNT WORK
- send_queue.py running on pyamqp, confirm messages sent, then receive_queue.py on pyamqp -> DOESNT WORK
- send_queue.py running on uamqp, confirm messages sent, then receive_queue.py on uamqp -> WORKS

The issue seems to be solely on the receiving side of SB pyamqp. I am unsure where the messages are going because the Service shows that they are being sent out of the queue when requested, they are just not appearing on the Wireshark trace or being received by the socket.  

The interesting case here is:
If you run pyamqp receive_queue.py, and then simultaneously run send_queue.py on either pyamqp or uamqp you will receive the messages as expected.

Also, this branch is without message settlement right now, it is just trying to get the messages in order to settle them. 

Edit: We are receiving frames as expected, we just are running receive twice. Need to investigate why.

### DIFF
--- a/sdk/servicebus/azure-servicebus/azure/servicebus/_common/utils.py
+++ b/sdk/servicebus/azure-servicebus/azure/servicebus/_common/utils.py
@@ -31,7 +31,8 @@ try:
 except ImportError:
     from urllib.parse import urlparse
 
-from uamqp import authentication, types
+from .._pyamqp.authentication import JWTTokenAuth
+from .._pyamqp import types
 
 from azure.core.settings import settings
 from azure.core.tracing import SpanKind, Link
@@ -110,14 +111,14 @@ def create_properties(user_agent=None):
     :rtype: dict
     """
     properties = {}
-    properties[types.AMQPSymbol("product")] = USER_AGENT_PREFIX
-    properties[types.AMQPSymbol("version")] = VERSION
+    properties["product"] = USER_AGENT_PREFIX
+    properties["version"] = VERSION
     framework = "Python/{}.{}.{}".format(
         sys.version_info[0], sys.version_info[1], sys.version_info[2]
     )
-    properties[types.AMQPSymbol("framework")] = framework
+    properties["framework"] = framework
     platform_str = platform.platform()
-    properties[types.AMQPSymbol("platform")] = platform_str
+    properties["platform"] = platform_str
 
     final_user_agent = "{}/{} {} ({})".format(
         USER_AGENT_PREFIX, VERSION, framework, platform_str
@@ -125,7 +126,7 @@ def create_properties(user_agent=None):
     if user_agent:
         final_user_agent = "{} {}".format(user_agent, final_user_agent)
 
-    properties[types.AMQPSymbol("user-agent")] = final_user_agent
+    properties["user-agent"] = final_user_agent
     return properties
 
 
@@ -165,7 +166,7 @@ def create_authentication(client):
     except AttributeError:
         token_type = TOKEN_TYPE_JWT
     if token_type == TOKEN_TYPE_SASTOKEN:
-        auth = authentication.JWTTokenAuth(
+        auth =JWTTokenAuth(
             client._auth_uri,
             client._auth_uri,
             functools.partial(client._credential.get_token, client._auth_uri),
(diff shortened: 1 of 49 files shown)


sdk/servicebus/azure-servicebus/azure/servicebus/_common/utils.py
```diff
@@ -110,22 +111,22 @@ def create_properties(user_agent=None):
     :rtype: dict
     """
     properties = {}
-    properties[types.AMQPSymbol("product")] = USER_AGENT_PREFIX
-    properties[types.AMQPSymbol("version")] = VERSION
```
annatisch, line 114: I don't think we can just drop the AMQPSymbol from these.... otherwise I think this is just going to be encoded as strings. I see this in a couple of places, so we should double check the payloads are being encoded as we think they are.

sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_receiver.py
```diff
@@ -355,18 +362,44 @@ def _create_handler(self, auth):
...
-
-        auth = None if self._connection else create_authentication(self)
+        if not self._running:
+            if self._handler:
+                self._handler.close()
+        auth = self._create_auth()
         self._create_handler(auth)
         try:
-            self._handler.open(connection=self._connection)
+            self._handler.open()
```
annatisch, line 402: Why are we no longer passing in the connection? Does this happen somewhere else now?

sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_receiver.py
```diff
@@ -407,12 +441,14 @@ def _receive(self, max_message_count=None, timeout=None):
             # Dynamically issue link credit if max_message_count > 1 when the prefetch_count is the default value 1
             if max_message_count and self._prefetch_count == 1 and max_message_count > 1:
                 link_credit_needed = max_message_count - len(batch)
-                amqp_receive_client.message_handler.reset_link_credit(link_credit_needed)
+                amqp_receive_client._link_credit = link_credit_needed
```
annatisch, line 444: This looks problematic - we should revisit this one.

sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_sender.py
```diff
@@ -230,35 +236,65 @@ def _from_connection_string(cls, conn_str, **kwargs):
...
         try:
-            self._handler.open(connection=self._connection)
+            self._handler.open()
             while not self._handler.client_ready():
                 time.sleep(0.05)
             self._running = True
             self._max_message_size_on_link = (
-                self._handler.message_handler._link.peer_max_message_size
-                or uamqp.constants.MAX_MESSAGE_LENGTH_BYTES
+                # self._handler.message_handler._link.peer_max_message_size
```
annatisch, line 296: We might need to add a todo for this one too

sdk/servicebus/azure-servicebus/azure/servicebus/amqp/_amqp_message.py
```diff
@@ -219,48 +232,46 @@ def _from_amqp_message(self, message):
...
             priority=message.header.priority
         ) if message.header else None
-        self._footer = message.footer
-        self._annotations = message.annotations
-        self._delivery_annotations = message.delivery_annotations
-        self._application_properties = message.application_properties
+        self._footer = message.footer if message.footer else {}
+        self._annotations = message.message_annotations if message.message_annotations else {}
+        self._delivery_annotations = message.delivery_annotations if message.delivery_annotations else {}
+        self._application_properties = message.application_properties if message.application_properties else {}
```
annatisch, line 244: Did these previously default to empty dictionaries?
In terms of an AMQP message, an empty dict != None

annatisch (commented):
Couple of preliminary comments I noticed :)

PR #24755, opened 2022-06-07
[Storage] Fix page blob tier on `upload_blob`

jalauzon-msft:
The recent changes to the generated code as part of STG83 included a change to creating page blobs which requires that we no longer set the `x-ms-access-tier` header manually and instead pass it to the generated code. This was partly addressed in #24668 but `upload_blob` was missed in that change. This change applies the same fix to `upload_blob`.


vincenttran-msft (approved):
Thanks for finding this and the speedy turnaround time!

PR #24772, opened 2022-06-08
[Storage] Tag indexing for versions in STG84

vincenttran-msft:
Adding back in tag indexing for versions for the upcoming release (STG84) 😄


sdk/storage/azure-storage-blob/azure/storage/blob/_blob_service_client.py
```diff
@@ -462,6 +462,13 @@ def find_blobs_by_tags(self, filter_expression, **kwargs):
             To specify a container, eg. "@container='containerName' and \"Name\"='C'"
         :keyword int results_per_page:
             The max result per page when paginating.
+        :keyword list[str] include:
+            Specifies one or more additional datasets to include in the response.
+            Options include: 'versions'.
+
+            .. versionadded:: 12.13.0
+                This parameter was introduced in API version '2021-08-06'.
```
vincenttran-msft, line 470: Is this accurate for STG84? Can change this to reflect the correct API version (since I think the version would be bumped up for this release? but just don't know the value)

sdk/storage/azure-storage-blob/CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Release History
 
+## 12.13.0 (Unreleased)
```
jalauzon-msft, line 3: Let's leave out the CHANGELOG entry for now since this will be in version 12.14.0 but we have to add/release 12.13.0 first :).

sdk/storage/azure-storage-blob/azure/storage/blob/_blob_service_client.py
jalauzon-msft: The STG84 version should be 12.14.0 barring any unforeseen patch releases we end up needing.

jalauzon-msft (approved):

Issue #21416, opened 2021-10-26
Exception during tables error handling: 'NoneType' object has no attribute 'headers'


Indy2222:
- **Package Name**: azure-data-tables
- **Package Version**: 12.1.0
- **Operating System**: Linux
- **Python Version**: 3.9

Loading entities from a table raises this exception:

`Entrypoint raised exception <class 'AttributeError'>: 'NoneType' object has no attribute 'headers'`

with the following traceback:

```
File "/app/bin/.venv/lib/python3.9/site-packages/joe_runtime/invocation.py", line 182, in _process_input_batch_inner
  entrypoint_yielded_object = next(entrypoint_output)
File "/app/bin/.venv/lib/python3.9/site-packages/pyresources.py", line 25, in wrapper
  yield from entrypoint_output
File "/app/bin/./main.py", line 163, in entrypoint
  self.reload_farms_in_optimization_license_keys()
File "/app/bin/./main.py", line 206, in reload_farms_in_optimization_license_keys
  self.farms_service.load_farms()}
File "/app/bin/.venv/lib/python3.9/site-packages/pto_farms/farms.py", line 138, in load_farms
  return [Farm._from_entity(farm) for farm in farms]
File "/app/bin/.venv/lib/python3.9/site-packages/pto_farms/farms.py", line 138, in <listcomp>
  return [Farm._from_entity(farm) for farm in farms]
File "/app/bin/.venv/lib/python3.9/site-packages/azure/core/paging.py", line 129, in __next__
  return next(self._page_iterator)
File "/app/bin/.venv/lib/python3.9/site-packages/azure/core/paging.py", line 76, in __next__
  self._response = self._get_next(self.continuation_token)
File "/app/bin/.venv/lib/python3.9/site-packages/azure/data/tables/_models.py", line 372, in _get_next_cb
  _process_table_error(error)
File "/app/bin/.venv/lib/python3.9/site-packages/azure/data/tables/_error.py", line 152, in _process_table_error
  decoded_error = _decode_error(storage_error.response, storage_error.message)
File "/app/bin/.venv/lib/python3.9/site-packages/azure/data/tables/_error.py", line 71, in _decode_error
  error_code = response.headers.get("x-ms-error-code")
```

The following warning was logged just before the exception was raised: `DefaultAzureCredential failed to retrieve a token from the included credentials.`

`DefaultAzureCredential` from package `azure-identity` of version `1.7.0` was used.

annatisch:
[...]

PR #24778, opened 2022-06-09
[Tables] Reraise an error with no response

annatisch:
Hi @YalinLi0312 - I believe this is a potential fix for #21416 
Would you be able to try and construct a test case for a scenario where `HttpResponseError.response is None`
It might be worth checking with McCoy to see where this azure-identity AuthenticationError is being raised, and why it has an empty `.response` attribute.

### DIFF
--- a/sdk/tables/azure-data-tables/azure/data/tables/_error.py
+++ b/sdk/tables/azure-data-tables/azure/data/tables/_error.py
@@ -204,6 +204,8 @@ def _reraise_error(decoded_error):
 
 
 def _process_table_error(storage_error, table_name=None):
+    if not storage_error.response:
+        _reraise_error(storage_error)
     decoded_error = _decode_error(storage_error.response, storage_error.message)
     if table_name:
         _validate_tablename_error(decoded_error, table_name)


mccoyp:
For context on the `azure-identity` error, it's coming from [this line](https://github.com/Azure/azure-sdk-for-python/blob/81c6aa0a21c542f65827304b6aa313d7af7e4910/sdk/identity/azure-identity/azure/identity/_credentials/chained.py#L108) in ChainedTokenCredential (which DefaultAzureCredential inherits from). The scenario in https://github.com/Azure/azure-sdk-for-python/issues/21416 occurs because none of the credentials in the credential chain succeed in their `get_token` call preceding our service request -- so, the service request never actually gets made and the error is raised from `azure-identity` instead. That's why there's no `response` property of the error. Yalin will follow up in a bit, since we got a chance to look at this today!

annatisch:
Great - thanks @mccoyp!
@YalinLi0312 - we should be able to inject a failure a here to repro the behaviour in tables and add some test cases to this PR and validate the fix :)
(Let me know if you have any trouble pushing to my branch)

annatisch:
Closing this PR because @YalinLi0312 has a better idea to do try/except around the decode method.

PR #24788, opened 2022-06-09
[Tables] Raise an error while failed to decode response

YalinLi0312:
For issue: https://github.com/Azure/azure-sdk-for-python/issues/21416
Raise an error while it's failing before a service request is sent out.

### DIFF
--- a/sdk/tables/azure-data-tables/CHANGELOG.md
+++ b/sdk/tables/azure-data-tables/CHANGELOG.md
@@ -7,6 +7,7 @@
 ### Breaking Changes
 
 ### Bugs Fixed
+* Fix the error handling bug while the `DefaultAzureCredential` failed to get a token in an incorrect environment before the actual service request sending out.([#24788](https://github.com/Azure/azure-sdk-for-python/pull/24788))
 
 ### Other Changes
 
--- a/sdk/tables/azure-data-tables/CHANGELOG.md
+++ b/sdk/tables/azure-data-tables/CHANGELOG.md
@@ -7,7 +7,7 @@
 ### Breaking Changes
 
 ### Bugs Fixed
-* Fix the error handling bug while the `DefaultAzureCredential` failed to get a token in an incorrect environment before the actual service request sending out.([#24788](https://github.com/Azure/azure-sdk-for-python/pull/24788))
+* Fix handling of client-side exceptions that get raised during service requests (such as [#21416](https://github.com/Azure/azure-sdk-for-python/issues/21416)) ([#24788](https://github.com/Azure/azure-sdk-for-python/pull/24788))
 
 ### Other Changes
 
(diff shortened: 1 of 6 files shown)


sdk/tables/azure-data-tables/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 ### Breaking Changes
 
 ### Bugs Fixed
+* Fix the error handling bug while the `DefaultAzureCredential` failed to get a token in an incorrect environment before the actual service request sending out.([#24788](https://github.com/Azure/azure-sdk-for-python/pull/24788))
```
mccoyp, line 10: The `DefaultAzureCredential` is what brought this error handling issue to our attention, but since it's actually an issue general to any internal errors that may come up during a client method call, I would suggest something like:
```suggestion
* Fix handling of client-side exceptions that get raised during service requests (such as [#21416](https://github.com/Azure/azure-sdk-for-python/issues/21416)) ([#24788](https://github.com/Azure/azure-sdk-for-python/pull/24788))
```

sdk/tables/azure-data-tables/tests/test_table_entity.py
```diff
@@ -2236,3 +2238,21 @@ def test_upsert_entity_with_invalid_key_type(self, tables_storage_account_name,
                 self.table.upsert_entity(entity2)
         finally:
             self._tear_down()
+
+    @tables_decorator
+    @recorded_by_proxy
```
mccoyp, line 2243: The recordings for these tests are empty because the service request never gets made, so we could save space in the repo by not recording these tests. The fact that these tests could run without recordings is evidence that service calls aren't being made, so I don't think there's a reason to keep the empty recordings

sdk/tables/azure-data-tables/tests/test_table_entity.py
```diff
@@ -2236,3 +2238,21 @@ def test_upsert_entity_with_invalid_key_type(self, tables_storage_account_name,
...
+    def test_list_tables_with_invalid_credential(self, tables_storage_account_name, tables_primary_storage_account_key):
```
mccoyp, line 2244: This is a (very) recent development, but per https://github.com/Azure/azure-sdk-for-python/pull/24799, the test proxy is getting an update that will require test arguments to be passed via `**kwargs` instead of with positional parameters. So this should get updated to:
```suggestion
    def test_list_tables_with_invalid_credential(self, **kwargs):
        tables_storage_account_name = kwargs.pop("tables_storage_account_name")
        tables_primary_storage_account_key = kwargs.pop("tables_primary_storage_account_key")
```
This applies to the async test as well

mccoyp: Change of plans on the above -- you can leave the positional parameters as they are 🙂

mccoyp (approved):

annatisch (approved):

Issue #24769, opened 2022-06-08
[textanalytics] adjust code for 2022-05-01 api changes


kristapratico:
TA GA swagger diffs:
- [x] Removes ExtractiveSummarization
- [x] CustomSingleLabelClassificationResult/CustomMultiLabelClassificationResult merged into CustomLabelClassificationResult
- [x] Removes FHIR
- [x] JobMetadata -> JobState
    - lastUpdateDateTime -> lastUpdatedDateTime
    - adds partiallyCompleted status

PR #24810, opened 2022-06-10
[textanalytics] remove fhir and add backwards compat for changed metadata property

kristapratico:
Resolves https://github.com/Azure/azure-sdk-for-python/issues/24769

- Removes FHIR
- Regenerates with swagger transform to add backwards compat for poller metadata property which went from `lastUpdateDateTime` to `lastUpdatedDateTime` in this version.


sdk/textanalytics/azure-ai-textanalytics/tests/test_analyze_healthcare.py
```diff
@@ -216,7 +216,7 @@ def callback(resp):
         response = client.begin_analyze_healthcare_entities(
             docs,
             show_stats=True,
-            model_version="2022-03-01",
```
deyaaeldeen, line 219: The current list of model versions does not include 2022-03-01: https://docs.microsoft.com/en-us/azure/cognitive-services/language-service/text-analytics-for-health/language-support I wonder if the model versioning policy allows removing versions this soon.

PR #24838, opened 2022-06-14
[Cosmos] Add AutoScale Header Support 

JericHunter:
Adding autoscale header support to container/database methods.


sdk/cosmos/azure-cosmos/azure/cosmos/_base.py
```diff
@@ -673,3 +676,14 @@ def validate_cache_staleness_value(max_integrated_cache_staleness):
     int(max_integrated_cache_staleness)  # Will throw error if data type cant be converted to int
     if max_integrated_cache_staleness <= 0:
         raise ValueError("Parameter 'max_integrated_cache_staleness_in_ms' can only be a positive integer.")
+
+
+def auto_scale_header(auto_scale):
```
tjprescott, line 681: This feels weird as a method. Also, I'd recommend adding type hints since I don't know what `auto_scale` is. Now that we no longer support Python 2, you can use type hints in both sync and async and for new stuff I would argue you definitely should.

sdk/cosmos/azure-cosmos/azure/cosmos/aio/_database.py
```diff
@@ -230,6 +230,11 @@ async def create_container(
         if offer_throughput is not None:
             request_options["offerThroughput"] = offer_throughput
 
+        auto_scale_settings = kwargs.pop('auto_scale_settings', None)
+        if auto_scale_settings is not None:
+            auto_scale_settings = auto_scale_header(auto_scale=auto_scale_settings)
+            request_options["autoUpgradePolicy"] = auto_scale_settings
```
tjprescott, line 236: It feels like this should be one step
```Python
if auto_scale_settings is not None:
  request_options["autoUpgradePolicy"] = str(auto_scale_settings.__dict__)
```

The `auto_scale_method` doesn't feel necessary.

sdk/cosmos/azure-cosmos/azure/cosmos/_base.py
tjprescott: Yeah, I don't think you need this method at all. 
`str(auto_scale.__dict__)` should do what you need, I would think.

sdk/cosmos/azure-cosmos/azure/cosmos/auto_scale.py
```diff
@@ -0,0 +1,33 @@
+# The MIT License (MIT)
```
annatisch, line 1: Please make this file private, i.e. rename to `_auto_scale.py`

sdk/cosmos/azure-cosmos/azure/cosmos/http_constants.py
```diff
@@ -166,6 +166,7 @@ class HttpHeaders(object):
     # Offer type.
     OfferType = "x-ms-offer-type"
     OfferThroughput = "x-ms-offer-throughput"
+    AutoscaleSettings = "x-ms-cosmos-offer-autopilot-settings"
```
annatisch, line 169: Why does this have different capitalization to the object name?

sdk/cosmos/azure-cosmos/azure/cosmos/auto_scale.py
```diff
@@ -0,0 +1,33 @@
...
+# SOFTWARE.
+
+class AutoScaleProperties(object):
+    """Represents the auto-scale throughput properties in an Azure Cosmos DB SQL API container.
+
+   :param auto_scale_max_throughput is the max autoscale throughput, it should have valid throughput
+   values between 1000 and 1000000 inclusive in increments of 1000
+
+   :param auto_upgrade_throughput_increment_percent is the auto upgrade max throughput increment percentage
+   default value is 0 the increment percent should be greater than or equal to zero
```
annatisch, line 29: Please add the types to the docstring.
Please also type hint the model

sdk/cosmos/azure-cosmos/azure/cosmos/auto_scale.py
```diff
@@ -0,0 +1,33 @@
...
+    """
+    def __init__(self, auto_scale_max_throughput, auto_upgrade_throughput_increment_percent=0):
+        self.auto_scale_max_throughput = auto_scale_max_throughput
+        self.auto_upgrade_throughput_increment_percent = auto_upgrade_throughput_increment_percent
```
annatisch, line 33: These aren't actual header names - any chance we could improve the names?
They don't need the word "auto" in them - that's already implied by the object name.
These names are long - yet their length doesn't provide me much clarity as to what I'm configuring...so I'm reliant on documentation anyway...

annatisch: Another thought - do we expect the number of properties here to increase?
I'm not _that_ opposed to using a model for this (rather than flattening it out) - but only 2 properties is a pretty fine line....
How do these two properties interact? Must I always set both of them? It looks like the 2nd one has a default value - will most people use this? If so, are we making most people instantiate an object to set a single property?

sdk/cosmos/azure-cosmos/azure/cosmos/_base.py
JericHunter: Hello Travis the reason why we needed to use this method is because we are using a nested dictionary and when doing it the way you are thinking will cause it not to work as it just places both properties in a single dictionary and we are still missing dictionaries that aren't directly linked to our object, I have linked an example image showing what it is supposed to look like. 
Screenshot 2022-06-23 113218

sdk/cosmos/azure-cosmos/azure/cosmos/auto_scale.py
JericHunter: I am beginning to feel that we should just put the properties in our Offer file that contains the ThroughputProperties but I am not sure.

sdk/cosmos/azure-cosmos/azure/cosmos/_base.py
tjprescott: I don't understand. What is is you expect the payload to look like?

JericHunter: Screenshot 2022-06-30 111929
This is what it is expected to look like but when doing it the way you are thinking, it will not work because we are not using the layered dictionary, your version will look like this Screenshot 2022-06-30 112414 and it will not run because we are not getting any of our values from the properties

tjprescott: Why is the input format different from the format needed on the wire? I just don't know of any other place in the Python SDK where we do something like this, so it seems immediately suspect. But @annatisch has a lot more experience in the Python SDK than me, so I'll gladly defer to her on this.

PR #24860, opened 2022-06-15
[Storage] Increase allowed timing difference in flaky relative expiry test

vincenttran-msft:
In a recent test added in STG83 (`test_create_file_relative_expiry`) the allowed difference was quite low (1 second) as it rarely hiccuped in local testing. But when running automated tests, we really should allow for a greater difference (within reason). This PR bumps this difference from 1 second --> 60 seconds.


jalauzon-msft (approved):
Thanks for jumping on this!

PR #24861, opened 2022-06-15
[Storage] Better plumbing for batch operations + version_id support

vincenttran-msft:
This is a draft PR. Finally got a fully working proof-of-concept, but want some input on whether this is objectively less hacky than how we are doing it now (since it seems like the `_generated` function we are calling into gives us a couple things we don't want, but alot of things we do want 😝)


sdk/storage/azure-storage-blob/azure/storage/blob/_container_client.py
```diff
@@ -1256,14 +1257,24 @@ def _generate_delete_blobs_options(self,
                     if_tags_match_condition=if_tags_match_condition
                 )
 
-            query_parameters, header_parameters = self._generate_delete_blobs_subrequest_options(**options)
+            built_url = "/{}/{}{}".format(quote(container_name), quote(blob_name, safe='/~'), self._query_str)
```
vincenttran-msft, line 1260: Needed to pull this out from how the `HttpRequest` was built previously since we need the URL to pass into `build_delete_request`

sdk/storage/azure-storage-blob/azure/storage/blob/_container_client.py
```diff
@@ -1256,14 +1257,24 @@ def _generate_delete_blobs_options(self,
...
 
-            req = HttpRequest(
-                "DELETE",
-                "/{}/{}{}".format(quote(container_name), quote(blob_name, safe='/~'), self._query_str),
-                headers=header_parameters
-            )
-            req.format_parameters(query_parameters)
+            # TODO: Shouldn't need to POP these
```
vincenttran-msft, line 1262: Several values need to be popped as I was hitting errors on every single one that I am popping here. Any ideas on if this is the right direction or if we want to refactor / other ways to get the same effect?

sdk/storage/azure-storage-blob/azure/storage/blob/_container_client.py
```diff
@@ -1256,14 +1257,24 @@ def _generate_delete_blobs_options(self,
...
+                options.pop('etag')
+            if 'lease_access_conditions' in options:
+                options.pop('lease_access_conditions')
+            if 'modified_access_conditions' in options:
+                options.pop('modified_access_conditions')
+
+            req = build_delete_request(built_url, **options)
+            if 'x-ms-version' in req.headers:
+                req.headers.pop('x-ms-version')
+            if '?' in req.url:
```
vincenttran-msft, line 1274: `build_delete_request` was popping in a delete type of `permanent` causing Storage to kickback saying permanent deletion is not enabled, so have to strip after the `?`

PR #24874, opened 2022-06-16
[Storage] Increment versions after release

jalauzon-msft:


vincenttran-msft (approved):
It was a long and difficult review, but in the end, I have concluded that this is great work!

PR #24889, opened 2022-06-16
[Storage] Update client-side encryption samples to use V2

jalauzon-msft:


vincenttran-msft (approved):
LGTM! Great work on these, should be helpful to our customers to have these samples in there 😄

PR #24896, opened 2022-06-17
Resolve Core Conda Failure

scbedd:
Version of azure core we were running tests against didn't contain necessary items.


xiangyan99 (approved):

scbedd:
The storage build is a different error. This resolves the `core` build issues. Going to check enforcer override, tackle the storage failure, then figure out why test-proxy isn't starting properly in the conda tests.

weshaggard:
@scbedd looks like you still need to fix your local/fork'ed main branch, given all the merges.

scbedd:
> @scbedd looks like you still need to fix your local/fork'ed main branch, given all the merges.

I don't quite get what you're saying here. We definitely had another follow-up to fix though. Looks like I still have one to fix for storage, but we should be entirely fine for conda after that.

weshaggard:
Orthogonal to your actual changes just pointing out that you have a lot of merge commits in your PRs:
image

That is just a sign that your fork or local repo should probably be rest to upstream main. When the PR is squashed as part of the merge these go away but the number of commits in your PRs are pretty long.

scbedd:
Aha. Yeah. I'll just go do that. Thanks

PR #24912, opened 2022-06-20
[Confidential Ledger] Enable live testing

lynshi:
# Description

Adds configuration files for live testing of `confidentialledger`. Note the tests may currently be broken as setup for live testing is not currently included for the preview SDK.

### DIFF
--- a/sdk/confidentialledger/test-resources.json
+++ b/sdk/confidentialledger/test-resources.json
@@ -1,0 +1,109 @@
+{
+  "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#",
+  "contentVersion": "1.0.0.0",
+  "parameters": {
+    "ledgerName": {
+      "type": "string",
+      "defaultValue": "[resourceGroup().name]",
+      "metadata": {
+        "description": "Ledger Name"
+      },
+      "minLength": 3,
+      "maxLength": 24
+    },
+    "resourceGroup": {
+      "type": "string",
+      "defaultValue": "[resourceGroup().name]",
+      "metadata": {
+        "description": "Resource Group containing the application."
+      }
+    },
+    "subscriptionId": {
+      "type": "string",
+      "defaultValue": "[subscription().subscriptionId]",
+      "metadata": {
+        "description": "Id of the subscription containing the application."
+      }
+    },
+    "tenantId": {
+      "type": "string",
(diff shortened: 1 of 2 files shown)


sdk/confidentialledger/tests.yml
```diff
@@ -0,0 +1,13 @@
+trigger: none
+
+stages:
+  - template: ../../eng/pipelines/templates/stages/archetype-sdk-tests.yml
```
benbp, line 4:
```suggestion
  - template: /eng/pipelines/templates/stages/archetype-sdk-tests.yml
```

sdk/confidentialledger/test-resources.json
```diff
@@ -0,0 +1,109 @@
+{
```
benbp, line 1: Preferably we start using bicep templates instead of ARM templates, if you don't mind updating. The file can be called `test-resources.bicep` and can look like below. Also we don't need to include parameters or outputs for the service principal/subscription values as the deployment script will handle that and prefix them in the way you had specified (similarly, you have an oid hardcoded default that needs to be removed). Additionally, the references to `CONFIDENTIALLEDGER_GROUP` should be changed to `CONFIDENTIALLEDGER_RESOURCE_GROUP` to follow convention (it gets set as an env var output [here](https://github.com/Azure/azure-sdk-tools/blob/main/eng/common/TestResources/New-TestResources.ps1#L251)).

```
@description('Ledger Name')
@minLength(3)
@maxLength(24)
param ledgerName string = resourceGroup().name

@description('The client OID to grant access to test resources.')
param testApplicationOid string

@description('The location of the resource. By default, this is the same as the resource group.')
param location string = resourceGroup().location

var azureConfidentialLedgerUrl = 'https://${ledgerName}.confidential-ledger.azure.com'

resource ledgerName_resource 'Microsoft.ConfidentialLedger/ledgers@2022-05-13' = {
  name: ledgerName
  location: location
  properties: {
    ledgerType: 'Public'
    aadBasedSecurityPrincipals: [
      {
        principalId: testApplicationOid
        ledgerRoleName: 'Administrator'
      }
    ]
  }
}

output CONFIDENTIALLEDGER_ENDPOINT string = azureConfidentialLedgerUrl
```

benbp: No prob. FYI bicep has a nifty `az bicep decompile -f <arm template>` function that makes it pretty easy to migrate.

benbp (approved):

PR #24913, opened 2022-06-20
Add azure monitor opentelemetry distro

lzchen:
Distro is a convenience package that will install azure monitor related components for monitoring solution.


rakshith91:
Discussed offline with @lzchen - we don't want to do a meta package at this time. Closing this for now.

PR #24930, opened 2022-06-21
[Storage] Make lease duration tests less flakey

jalauzon-msft:
This change aims to make some tests that rely on a lease expiring less flaky by slightly increasing the sleep time that is used to wait for the lease to expire. Previously the sleep time was exactly equal to the lease duration which meant sometimes the lease would not expire in time during pipeline execution (likely due to clock skew and other timing issues). This change just adds a 2 second buffer to ensure there is enough time for the lease to expire.


vincenttran-msft (approved):
Thanks for this! Be gone red!

PR #24931, opened 2022-06-21
[Storage] Remove client-side encryption code from shared

jalauzon-msft:
Previously the client-side encryption code lived in shared and therefore was present in fileshare and datalake even though those packages do not support encryption. This change moves the code out of `_shared` and into individual files inside just the Blob and Queue packages. This required a few changes as well as some mechanical changes.

- Remove `encryption.py` from `_shared` in all packages
- Create `_encryption.py` in Blob and Queue (duplicated code)
- Adjust all the imports accordingly as well as clean up imports in a bunch of files
- Move some upload logic out of the shared upload code and into the Blob specific upload code since the encryption code moved out of shared
- Remove encryption options from base client and create `StorageEncryptionMixin` and add it to all Blob and Queue clients
- Finally remove all encryption code from file-share and datalake


sdk/storage/azure-storage-blob/azure/storage/blob/_shared/uploads.py
```diff
@@ -56,24 +50,9 @@ def upload_data_chunks(
         max_concurrency=None,
         stream=None,
         validate_content=None,
-        encryption_options=None,
         progress_hook=None,
         **kwargs):
 
-    if encryption_options:
```
jalauzon-msft, line 63: This code needed to be moved to the Blob specific code (in `upload_helpers` since encryption was no longer in the shared folder. It was also good to get this encryption stuff out of shared as well.

sdk/storage/azure-storage-blob/azure/storage/blob/_upload_helpers.py
```diff
@@ -220,8 +227,12 @@ def upload_page_blob(
             except AttributeError:
                 tier = premium_page_blob_tier
 
-        if encryption_options and encryption_options.get('data'):
-            headers['x-ms-meta-encryptiondata'] = encryption_options['data']
+        if encryption_options and encryption_options.get('key'):
```
jalauzon-msft, line 230: This is the stuff that was removed from `_blob_client`.

vincenttran-msft (approved):
LGTM! Thanks for this, as it looks like it was quite a hairy process 😨

PR #24932, opened 2022-06-21
fix type of _SansIOHTTPPolicyRunner

xiangyan99:
We have 

class HTTPPolicy(ABC, Generic[HTTPRequestType, HTTPResponseType]) and 

class _SansIOHTTPPolicyRunner(HTTPPolicy, Generic[HTTPRequestType, HTTPResponseType]):

Python starts to complain such double inheritance since 3.11.0b3.

### DIFF
--- a/sdk/core/azure-core/CHANGELOG.md
+++ b/sdk/core/azure-core/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### Bugs Fixed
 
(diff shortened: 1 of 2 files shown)


annatisch (approved):

major:
@xiangyan99 I added this patch to an RPM build of `azure-core` in Fedora with Python 3.11.0b3 and I'm able to import `azure.core` now without issues. 🎊

PR #24945, opened 2022-06-22
[Eventhubs] Updated perf tests

annatisch:
Includes:
- New EventPerfTest based receiving test
- Updated existing send tests to use BatchPerfTest base
- Added perf test for buffered sender
- Added system test for uamqp perf
- Removed old T1 tests and receive tests


kashifkhan (approved):

swathipil (approved):

l0lawrence (approved):

mikeharder (changes requested):

tools/azure-devtools/src/azure_devtools/perfstress_tests/system_perfstress/uamqp_event_test.py
```diff
@@ -0,0 +1,97 @@
+# --------------------------------------------------------------------------------------------
```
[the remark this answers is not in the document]
annatisch: Sure okay - I'll move it into eventhubs (given that's the endpoint it using)

sdk/eventhub/azure-eventhub/tests/perfstress_tests/_test_base.py
```diff
@@ -3,37 +3,145 @@
...
+            load_balancing_strategy=arguments.load_balancing_strategy
+        )
+        if arguments.preload:
+            self.async_producer = AsyncEventHubProducerClient.from_connection_string(connection_string, eventhub_name=eventhub_name)
+
+    async def _preload_eventhub(self):
+        data = get_random_bytes(self.args.event_size)
+        async with self.async_producer as producer:
+            partitions = await producer.get_partition_ids()
+            for partition in partitions:
```
mikeharder, line 58: Would it be faster to populate the partitions in parallel?

sdk/eventhub/azure-eventhub/tests/perfstress_tests/_test_base.py
```diff
@@ -3,37 +3,145 @@
...
+        parser.add_argument('--prefetch-count', nargs='?', type=int, help='Number of events to receive locally per request. Defaults to 300', default=300)
+        parser.add_argument('--load-balancing-strategy', nargs='?', type=str, help="Event Processor load balancing strategy, 'greedy' or 'balanced'. Default is 'greedy'.", default='greedy')
+        parser.add_argument('--checkpoint-interval', nargs='?', type=int, help='Interval between checkpoints (in number of events).  Default is no checkpoints.', default=None)
+        parser.add_argument('--max-wait-time', nargs='?', type=float, help='Maximum time to wait for an event to be received.', default=None)
+        parser.add_argument('--processing-delay', nargs='?', type=int, help='Delay when processing each event (in ms).', default=None)
+        parser.add_argument('--processing-delay-strategy', nargs='?', type=str, help="Whether to 'sleep' or 'spin' during processing delay. Default is 'sleep'.", default='sleep')
+        parser.add_argument('--preload', nargs='?', type=int, help='Ensure the specified number of events are available in each partition. Default is 0.', default=0)
```
mikeharder, line 120: Do you think it might be more intuitive if the parameter was the total number of events across all partitions, instead of per-partition?  Once you have an estimate of your read ops/sec, you then know the total number of events you need to preload.  It's harder if you have to divide the number of events by number of partitions.

sdk/eventhub/azure-eventhub/tests/perfstress_tests/_test_base.py
mikeharder: A related question, why populate events partition by partition, rather than uploading to the gateway and letting the events be automatically partitioned?  Do we have scenarios where we want to measure direct access to partitions, as well as through the gateway?

mikeharder (approved):

sdk/eventhub/azure-eventhub/tests/perfstress_tests/_test_base.py
annatisch: The reason that it's per partition is because I can only implement the 'idempotent message preloading' per partition - i.e. there's no way to detect cumulative messages in the EH. So I could count all the messages in each partition - then send the remainder to the gateway without prescribing partition - but I figured given that I was already iterating over the partitions - and when we count on the receiving end this is also done per partition - it seemed simpler just to load messages on that basis too.
Happy to update to the "count per partition but send to all partitions" model if you think it's preferable.

mikeharder: Thanks for the explanation.  I do think it will be easier to use if we switch to "count per partition but send to all partitions", so can you please make this change?

annatisch: Yup - sure thing!

annatisch: Thanks @mikeharder - updated :)

mikeharder:
@annatisch: LGTM.  Feel free to merge at will.

Issue #24836, opened 2022-06-14
Storage clients should handle AzureNamedKeyCredential as credential


bluenote10:
**Is your feature request related to a problem? Please describe.**

The storage clients (blob, queue, ...) doesn't seem to support the type `AzureNamedKeyCredential`. This is inconsistent with many other parts of the API. For instance the table client allows it:

https://github.com/Azure/azure-sdk-for-python/blob/c8a02681b2cff658237e9ea117ee67e65a8c79ca/sdk/tables/azure-data-tables/azure/data/tables/_base_client.py#L76-L82

It is also frustrating that the `credential` is typed as `Any`.

https://github.com/Azure/azure-sdk-for-python/blob/c8a02681b2cff658237e9ea117ee67e65a8c79ca/sdk/storage/azure-storage-blob/azure/storage/blob/_shared/base_client.py#L71

This causes a lot of guessing what kind of credential the clients support 😕 

**Describe the solution you'd like**

The signature should get proper type annotations, and this function should simply handle `AzureNamedKeyCredential`.

https://github.com/Azure/azure-sdk-for-python/blob/c8a02681b2cff658237e9ea117ee67e65a8c79ca/sdk/storage/azure-storage-blob/azure/storage/blob/_shared/base_client.py#L348-L359

Perhaps the private type `SharedKeyCredentialPolicy` should be even completely be replaced by `AzureNamedKeyCredential`.

**Describe alternatives you've considered**

Type guessing and constantly converting from and to `AzureNamedKeyCredential` and the plain `{"account_name": str, "account_key": str}` that the blob storage wants.

kashifkhan:
@bluenote10 Thanks for your feedback, we'll investigate asap.

annatisch:
Thanks for this report @bluenote10!
You're 100% correct - this should be added. `AzureNamedKeyCredential` was added to Azure Core some time after the first stable Storage SDK releases - and it looks like we forgot to go back and add support for it. Thanks for calling this out. :)

Stevenjin8:
@vincenttran-msft To add onto this, `AzureKeyCredential` also isn't supported. I'd be happy to try to help out with this.

bupt-wenxiaole:
any update on this?  I also encountered issues that storage client does not support AzureNamedKeyCredential

jalauzon-msft:
[...]

PR #24967, opened 2022-06-24
[Storage] Handling AzureNamedKeyCredential in Python SDK

vincenttran-msft:
Still TODO for Blob completion:

- [x] Propagate all the changes to other places we accept `credential`

TODO for full support-complete:

- [x] Propagate changes to other 4 packages
- [x] Add a test for each package

Originally raised in #24836


sdk/storage/azure-storage-blob/azure/storage/blob/_shared/base_client.py
```diff
@@ -54,6 +54,8 @@
 from .._version import VERSION
 from .response_handlers import process_storage_error, PartialBatchErrorException
 
+if TYPE_CHECKING:
```
vincenttran-msft, line 57: Do we like this placement? I am mirroring this from tables example (one newline after the last import, then a newline after)

sdk/storage/azure-storage-blob/azure/storage/blob/_shared/base_client.py
```diff
@@ -69,7 +73,7 @@ def __init__(
         self,
         parsed_url,  # type: Any
         service,  # type: str
-        credential=None,  # type: Optional[Any]
+        credential=None,  # type: Optional[Union[AzureNamedKeyCredential, AzureSasCredential, Dict[str,str], str, "TokenCredential"]]
```
vincenttran-msft, line 76: Do we want to go with this ordering before I duplicate this effort (I made it alphabetical, but still not sure if we are in love with that or want to group the more obvious types at the front and then the `Credential` types at the back?

sdk/storage/azure-storage-blob/azure/storage/blob/_shared/base_client.py
jalauzon-msft: LGTM!

sdk/storage/azure-storage-blob/azure/storage/blob/_shared/base_client.py
jalauzon-msft: Let's go with `str` and `Dict` first since those are probably what most people will want to use (and use today).
`str, Dict[str, str], AzureNamedKeyCredential, AzureSasCredential, "TokenCredential"`?

vincenttran-msft: Will do!

sdk/storage/azure-storage-blob/azure/storage/blob/_blob_service_client.py
```diff
@@ -74,10 +74,12 @@ class BlobServiceClient(StorageAccountHostsMixin):
     :param credential:
```
vincenttran-msft, line 74: How do we like the rewriting. Open to suggestions and formatting, since I sort of just tacked it on at the end

jalauzon-msft: LGTM! Thanks

sdk/storage/azure-storage-blob/azure/storage/blob/_shared/base_client.py
```diff
@@ -69,7 +74,8 @@ def __init__(
         self,
         parsed_url,  # type: Any
         service,  # type: str
-        credential=None,  # type: Optional[Any]
+        credential= None,  # type: Optional[Union[str, Dict[str, str],
```
vincenttran-msft, line 77: Linter hasn't complained, but at the same time I don't know if this type hint is being properly parsed as a typehint (namely, line 78). Any ideas on how to check this?

jalauzon-msft: Hmmm yeah, don't love that multi-line type hint. There are two things that can check it. Intelisense in VS code and then you can generate a `.whl` and upload it to API View and see if it got it. You can just create a new review with just one iteration and search for it then delete the review when you're done.

That being said, I don't love the way it looks :). The line is too long for lint?

vincenttran-msft: Will try to make it happy with a pylint disable 😄

sdk/storage/azure-storage-blob-changefeed/tests/test_change_feed.py
```diff
@@ -333,3 +334,14 @@ def test_list_3_shards_events_works_with_1_shard_cursor(self, storage_account_na
         dict_token_with_1_shard = json.loads(token_with_1_shard)
         self.assertEqual(len(dict_token_with_1_shard['CurrentSegmentCursor']['ShardCursors']), 1)
         self.assertEqual(len(dict_token['CurrentSegmentCursor']['ShardCursors']), 3)
+
+    # @ChangeFeedPreparer()
+    # def test_azure_named_key_credential_access(self, storage_account_name, storage_account_key):
```
vincenttran-msft, line 339: Test is passing when ran live -- but then doesn't generate a recording. Any idea on how to resolve this? I have checked my `settings_real.py` file and to no avail 😢

sdk/storage/azure-storage-blob-changefeed/azure/storage/blob/changefeed/_change_feed_client.py
```diff
@@ -25,10 +25,12 @@ class ChangeFeedClient(object):  # pylint: disable=too-many-public-methods
     :param credential:
```
jalauzon-msft, line 25: We should be updating the type hint for these as well. Since its a param, its listed in the function porotype below.

This applies to most of the files in this review (sorry!). Anywhere we document this, we should also update the type hint to match what you have in `base_client`. I guess that means PyLint disable and all sadly.

sdk/storage/azure-storage-blob-changefeed/tests/test_change_feed.py
```diff
@@ -21,6 +21,7 @@
 
 from devtools_testutils.storage import StorageTestCase
 from settings.testcase import ChangeFeedPreparer
+from azure.core.credentials import AzureNamedKeyCredential
```
jalauzon-msft, line 24: I think this is left over and isn't need anymore?

sdk/storage/azure-storage-blob-changefeed/azure/storage/blob/changefeed/_change_feed_client.py
vincenttran-msft: Good point! Will get on that now 😄

sdk/storage/azure-storage-blob-changefeed/tests/test_change_feed.py
vincenttran-msft: Good catch, unused import since we canned the changefeed test!

jalauzon-msft (approved):

Issue #23472, opened 2022-03-10
[AMQP Python] Message settlement


yunhaoling:
### Issue
implement the amqp protocol section 3.4, This is the feature required by service bus received message settlement (complete/defer/deadletter/abandon)

<img width="219" alt="image" src="https://user-images.githubusercontent.com/47871814/157743552-dc5779e6-ab75-4544-bf9b-610226e225c5.png">

### references
- uamqp implementation details:
  - python
    - message.py: https://github.com/Azure/azure-uamqp-python/blob/master/uamqp/message.py#L484-L573
    - receiver.py: https://github.com/Azure/azure-uamqp-python/blob/master/uamqp/receiver.py#L190-L219
  - cython
    - message_receiver.pyx: https://github.com/Azure/azure-uamqp-python/blob/master/src/message_receiver.pyx#L73-L109
    - c_message_receiver.pxd: https://github.com/Azure/azure-uamqp-python/blob/master/src/vendor/inc/c_message_receiver.pxd#L33
   - c
     - message_receiver.c: https://github.com/Azure/azure-uamqp-python/blob/master/src/vendor/azure-uamqp-c/src/message_receiver.c#L537
     - link.c https://github.com/Azure/azure-uamqp-python/blob/master/src/vendor/azure-uamqp-c/src/link.c#L1707
     - link.c https://github.com/Azure/azure-uamqp-python/blob/master/src/vendor/azure-uamqp-c/src/link.c#L161

kashifkhan:
done with release of service bus 7.10.0

PR #24975, opened 2022-06-24
[SB PyAMQP] Servicebus PyAMQP Working off of Anna's Branch

l0lawrence:
fixes: #23472


sdk/servicebus/azure-servicebus/azure/servicebus/amqp/_amqp_message.py
```diff
@@ -201,7 +220,17 @@ def __repr__(self):
         return "AmqpAnnotatedMessage({})".format(message_repr)[:1024]
 
     def _from_amqp_message(self, message):
-        # populate the properties from an uamqp message
+        # populate the properties from an pyamqp message
+        if message[5]:
```
swathipil, line 224: out of curiosity: is this more performant than calling message.data?

sdk/servicebus/azure-servicebus/azure/servicebus/amqp/_amqp_message.py
```diff
@@ -141,19 +159,16 @@ def __init__(
                 "or value_body being set as the body of the AmqpAnnotatedMessage."
             )
 
-        self._body = None
-        self._body_type = None
         if "data_body" in kwargs:
-            self._body = kwargs.get("data_body")
-            self._body_type = uamqp.MessageBodyType.Data
+            self._data_body = normalized_data_body(kwargs.get("data_body"))
+            self.body_type = AmqpMessageBodyType.DATA
```
swathipil, line 164: wondering why we want to allow users to set body_type as well now? are there cases where we would want to allow users to set this?

sdk/servicebus/azure-servicebus/azure/servicebus/_common/message.py
```diff
@@ -227,36 +218,30 @@ def _build_message(self, body):
...
-        # type: () -> ServiceBusMessage
-        # pylint: disable=protected-access
-        self.message = self.raw_amqp_message._to_outgoing_amqp_message()
+    def _to_outgoing_message(self) -> "ServiceBusMessage":
         return self
 
     @property
-    def raw_amqp_message(self):
-        # type: () -> AmqpAnnotatedMessage
+    def message(self) -> LegacyMessage:
```
swathipil, line 233: wondering if the reason there isn't a message setter is because we expect that users probably only used get? and whether, b/c message can't be set anymore, it changes the API?

annatisch: hmm that's a good question....
I mostly assumed that because it was on a ReceivedMessage, it could really only be consumed - but I guess a ReceivedMessage can be resent right?
I think it would be unlikely that someone was building a uamqp message and setting this property, and more likely that they updated the existing message... but could be worth looking into.

sdk/servicebus/azure-servicebus/azure/servicebus/amqp/_amqp_message.py
annatisch: I think this is probably my mistake.... will fix.

sdk/servicebus/azure-servicebus/azure/servicebus/amqp/_amqp_message.py
annatisch: Generally yes - accessing a tuple by index is the tiniest bit faster.

sdk/servicebus/azure-servicebus/azure/servicebus/amqp/_amqp_message.py
```diff
@@ -273,45 +304,38 @@ def _to_outgoing_amqp_message(self):
...
-        return uamqp.message.Message(
-            body=amqp_body,
-            body_type=amqp_body_type,
+        annotations = None
+        if self.annotations:
+            # TODO: Investigate how we originally encoded annotations.
+            annotations = dict(self.annotations)
+            for key in _LONG_ANNOTATIONS:
+                if key in self.annotations:
+                    annotations[key] = amqp_long_value(self.annotations[key])
```
swathipil, line 320: are _LONG_ANNOTATIONS only relevant for SB? I don't see these in EH, so am currently passing in `message_annotations=self.annotations` to the pyamqp Message there.

sdk/servicebus/azure-servicebus/azure/servicebus/_common/message.py
```diff
@@ -227,36 +218,30 @@ def _build_message(self, body):
...
+        if not self._uamqp_message:
+            self._uamqp_message = LegacyMessage(self._raw_amqp_message)
```
swathipil, line 235: in order to remove the amqp Message creation/components from EventData and AmqpAnnotatedMessage, I removed the `to_outgoing_amqp_message` method from AmqpAnnotatedMessage. So, I was thinking that we could pass in PyamqpTransport().to_outgoing_amqp_message here:
```suggestion
            self._uamqp_message = LegacyMessage(self._raw_amqp_message, to_outgoing_amqp_message=PyamqpTransport().to_outgoing_amqp_message)
```
and in `_pyamqp/_message_backcompat.py`, we could create the amqp message like:
`to_outgoing_amqp_message(self._message)` rather than `self._message._to_outgoing_amqp_message`.
Thoughts/concerns?

sdk/servicebus/azure-servicebus/azure/servicebus/_common/message.py
```diff
@@ -125,15 +120,13 @@ def __init__(
         # Although we might normally thread through **kwargs this causes
         # problems as MessageProperties won't absorb spurious args.
         self._encoding = kwargs.pop("encoding", "UTF-8")
+        self._uamqp_message = None
```
swathipil, line 123: In EH, since we moved all of the uamqp/pyamqp Message creation out to right before sending (so that the user doesn't have to specify the transport to the EventData object each time they create one), how will we know whether the message property should return a uamqp or pyamqp Message?

Should we always have the message property return a pyamqp Message for now, since it will essentially provide the same info as the uamqp Message?

sdk/servicebus/azure-servicebus/azure/servicebus/_common/message.py
swathipil, line 235: Since I've moved the Message creation in AmqpAnnotatedMessage (specifically, the `to_outgoing_amqp_message` method) out to a Amqp Transport object, I think we'll need to call that method like:
`to_outgoing_amqp_message(self._message)` rather than: `self._message._to_outgoing_amqp_message()`. So I was thinking we could just do the following here:
```suggestion
            self._uamqp_message = LegacyMessage(self._raw_amqp_message, to_outgoing_amqp_message=PyamqpTransport().to_outgoing_amqp_message)
```
Thoughts/concerns?

sdk/servicebus/azure-servicebus/azure/servicebus/_common/message.py
```diff
@@ -722,32 +674,36 @@ def _add(self, add_message, parent_span=None):
...
         self._size = size_after_add
         self._count += 1
         self._messages.append(message)
 
     @property
-    def max_size_in_bytes(self):
-        # type: () -> int
+    def message(self) -> LegacyBatchMessage:
+        if not self._uamqp_message:
+            raise Exception("Attempting to use legacy batch")
```
swathipil, line 685: just checking that this line will be removed after testing?

annatisch: Absolutely - sorry this was my local debugging and I committed by accident >_<

sdk/servicebus/azure-servicebus/azure/servicebus/_pyamqp/_message_backcompat.py
```diff
@@ -0,0 +1,238 @@
...
+        self._settler = kwargs.get('settler')
+        self._encoding = kwargs.get('encoding')
+        self.delivery_no = kwargs.get('delivery_no')
+        self.delivery_tag = kwargs.get('delivery_tag') or None
+        self.on_send_complete = None
+        self.properties = LegacyMessageProperties(self._message.properties)
+        self.application_properties = self._message.application_properties
+        self.annotations = self._message.annotations
+        self.header = LegacyMessageHeader(self._message.header)
+        self.footer = self._message.footer
```
swathipil, line 56: for properties and header, these result in an error if self._message.header/properties is None.
```suggestion
        self.properties = LegacyMessageProperties(self._message.properties) if self._message.properties else None
        self.header = LegacyMessageHeader(self._message.header) if self._message.header else None
```
for the other ones, wondering if they should be like in AmqpAnnotatedMessage, like this? or if they should just be None, as they are here?
```
self.application_properties = self._message.application_properties if self._message.application_properties else {}
self.annotations = self._message.annotations if self._message.annotations else {}
self.footer = self._message.footer if self._message.footer else {}
self.delivery_annotations = self._message.delivery_annotations if self._message.delivery_annotations else {}
```

PR #24976, opened 2022-06-24
[Cosmos] Add AutoScale Header Support 1/2

JericHunter:
Adding autoscale header support to container/database methods.


sdk/cosmos/azure-cosmos/samples/container_management.py
```diff
@@ -181,6 +181,19 @@ def create_container(db, id):
     except exceptions.CosmosResourceExistsError:
         print('A container with id \'_container_analytical_store\' already exists')
 
+    print("\n2.8 Create Container - With auto scale settings")
+
+    try:
+        container = db.create_container(
+            id=id+"_container_auto_scale_settings",
+            partition_key=partition_key,
+            offer_throughput=Offer(auto_scale_max_throughput=5000, auto_upgrade_throughput_increment_percent=0)
```
annatisch, line 190: We shouldn't use the object `Offer` here - that was deprecated in favor of `ThroughputProperties`.
Also - I think the param name is `auto_scale_increment_percentage`?

sdk/cosmos/azure-cosmos/test/test_auto_scale.py
```diff
@@ -0,0 +1,177 @@
...
+        created_container = self.created_database.create_container(
+            id='container_with_auto_scale_settings',
+            partition_key=PartitionKey(path="/id"),
+            offer_throughput=Offer(auto_scale_max_throughput=5000, auto_scale_increment_percentage=0)
+
+        )
+        created_container_properties = created_container.get_throughput()
+        # Testing the input value of the max_throughput
+        self.assertEqual(
+            created_container_properties.properties['content']['offerAutopilotSettings']['maxThroughput'], 5000)
```
annatisch, line 60: What is this checking?
We should be testing the object itself:
`assert created_container_properties.auto_scale_max_throughput == 5000`
Is the object not being deserialized properly?

sdk/cosmos/azure-cosmos/test/test_auto_scale.py
```diff
@@ -0,0 +1,177 @@
...
+        # Testing the incorrect passing of an input value of the max_throughput to verify negative behavior
+        try:
+            created_container = self.created_database.create_container(
+                id='container_with_wrong_auto_scale_settings',
+                partition_key=PartitionKey(path="/id"),
+                offer_throughput=Offer(auto_scale_max_throughput=200, auto_scale_increment_percentage=0)
+
+            )
+        except exceptions.CosmosHttpResponseError as e:
+            self.assertEqual(e.status_code, http_constants.StatusCodes.BAD_REQUEST)
```
annatisch, line 73: Do we get any error message for an incorrect percentage value?
It would be pretty hard to debug the problem if all I got was "bad request"

sdk/cosmos/azure-cosmos/test/test_auto_scale.py
```diff
@@ -0,0 +1,177 @@
...
+            id='container_with_auto_scale_settings',
+            partition_key=PartitionKey(path="/id"),
+            offer_throughput=Offer(auto_scale_max_throughput=5000, auto_scale_increment_percentage=1)
+
+        )
+        created_container_properties = created_container.get_throughput()
+        # Testing the input value of the max_increment_percentage
+        self.assertEqual(
+            created_container_properties.properties['content']['offerAutopilotSettings']['autoUpgradePolicy']
+            ['throughputPolicy']['incrementPercent'], 1)
```
annatisch, line 86: Same as above - this needs to be properly deserialized:
```python
assert created_container_properties.auto_scale_increment_percent == 1
```

sdk/cosmos/azure-cosmos/test/test_auto_scale.py
```diff
@@ -0,0 +1,177 @@
...
+        # Testing for wrong attributes for the auto_scale_settings in the created container
+        try:
+            created_container = self.created_database.create_container(
+                id='container_with_wrong_auto_scale_settings',
+                partition_key=PartitionKey(path="/id"),
+                offer_throughput="wrong setting"
+
+            )
+        except exceptions.CosmosHttpResponseError:
+            print("CosmosHttpResponseError")
```
annatisch, line 112: If we expect an error to be raised, then the correct format is:
```python
with pytest.raises(CosmosHttpResponseError) as e:
    self.created_database.create_container(
        id='container_with_wrong_auto_scale_settings',
        partition_key=PartitionKey(path="/id"),
        offer_throughput="wrong setting"
    )
```

sdk/cosmos/azure-cosmos/azure/cosmos/offer.py
```diff
@@ -27,10 +27,14 @@ class ThroughputProperties(object):
     To read and update throughput properties use the associated methods on the :class:`Container`.
```
annatisch, line 27: Looks like we don't have a proper docstring for this object.
Now that we are expecting people to actually import and use this as input - we will need to fully document the constructor, and all the parameters and how they are used together.
For example - the increment_percent will be ignored if auto_scale_max_throughput is set, and if autoscale is set, I presume offer_througput will be ignored?
These combinations are all pretty confusing - so if the intention is to combine this object - it needs to be really clear how it's used.

sdk/cosmos/azure-cosmos/azure/cosmos/offer.py
```diff
@@ -27,10 +27,14 @@ class ThroughputProperties(object):
...
 
-    def __init__(self, offer_throughput, properties=None):  # pylint: disable=super-init-not-called
-        # type: (int, Dict[str, Any]) -> None
+    def __init__(self, offer_throughput=None, properties=None, auto_scale_max_throughput=None,
+                 auto_scale_increment_percentage=None):  # pylint: disable=super-init-not-called
+        # type: (int, Dict[str, Any], int, int) -> None
         self.offer_throughput = offer_throughput
         self.properties = properties
+        self.auto_scale_max_throughput = auto_scale_max_throughput
+        self.auto_scale_increment_percentage = auto_scale_increment_percentage
```
annatisch, line 36: Maybe just `percent` rather than `percentage`?

annatisch (changes requested):

sdk/cosmos/azure-cosmos/samples/container_management.py
JericHunter: We cannot reach the object without exposing **Offer** so what should we do? Do we expose it anyway? currently we are not passing in **ThroughputProperty** in our __init__ file.

annatisch: Then that would seem to be a bug that should be fixed. :)

JericHunter: so for this PR in particular do I still go with Offer or change it to ThroughputProperty and see about it being fixed after?

sdk/cosmos/azure-cosmos/test/test_auto_scale.py
JericHunter: We have not done that part yet, the plan was to make a separate PR to keep each functionality separate from each other meaning the create, replace, and read. For now, checking this way works so we want to go ahead with this way and will update it accordingly.

sdk/cosmos/azure-cosmos/test/test_auto_scale.py
JericHunter: Would you like me to print the error message there?

sdk/cosmos/azure-cosmos/azure/cosmos/offer.py
```diff
@@ -27,10 +27,14 @@ class ThroughputProperties(object):
     To read and update throughput properties use the associated methods on the :class:`Container`.
     """
 
-    def __init__(self, offer_throughput, properties=None):  # pylint: disable=super-init-not-called
-        # type: (int, Dict[str, Any]) -> None
+    def __init__(self, offer_throughput=None, properties=None, auto_scale_max_throughput=None,
+                 auto_scale_increment_percentage=None):  # pylint: disable=super-init-not-called
```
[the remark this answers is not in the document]
JericHunter: yes, and it is done

sdk/cosmos/azure-cosmos/azure/cosmos/aio/_database.py
```diff
@@ -228,7 +228,12 @@ async def create_container(
         response_hook = kwargs.pop('response_hook', None)
         offer_throughput = kwargs.pop('offer_throughput', None)
         if offer_throughput is not None:
-            request_options["offerThroughput"] = offer_throughput
+            if isinstance(offer_throughput, int):
+                request_options["offerThroughput"] = offer_throughput
+
+            if isinstance(offer_throughput, object):
```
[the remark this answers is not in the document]
annatisch: I think we need to work on this logic a bit. Maybe something like:
```
try:
    if offer.auto_scale_max_throughput or offer.auto_scale_increment_percent:
        request_options['autoUpgradePolicy'] = stringify_auto_scale(offer=offer_throughput)
    elif offer.offer_throughput:
        request_options["offerThroughput"] = offer.offer_throughput
    else:
        raise ValueError(...)
except AttributeError:
    if isinstance(offer, int):
        request_options["offerThroughput"] = offer_throughput
    else:
        ...
```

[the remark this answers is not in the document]
JericHunter: I tried what you have recommended but it doesn't really work out the same way we would want it to especially because offer doesn't exist here and I tried to use the ThroughputProperties instead but that doesn't work out either, I understand technically that is true but it is not causing any issues with wanting to set a offer throughput or setting the autoscale settings.

sdk/cosmos/azure-cosmos/test/test_auto_scale.py
JericHunter: When asserting with the message rather than the status code we get assertion errors, so what should I do with that?

sdk/cosmos/azure-cosmos/azure/cosmos/offer.py
```diff
@@ -25,12 +25,26 @@
...
+    :param offer_throughput: The provisioned throughput in request units per second as a number.
+
+    :param auto_scale_max_throughput: is the max autoscale throughput, it should have valid throughput
+   values between 1000 and 1000000 inclusive in increments of 1000.
+
+    :param auto_scale_increment_percent: is the  % from the base selected RU it increases at a given time,
+    up to a maximum default value is 0 the increment percent should be greater than or equal to zero.
+
+    Both auto_scale_max_throughput and auto_scale_increment_percent have to be used simultaneously, you cannot
+    set the offer_throughput when setting the auto_scale properties.
```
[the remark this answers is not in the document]
JericHunter: That makes sense completely and I will change it accordingly.

sdk/cosmos/azure-cosmos/azure/cosmos/offer.py
```diff
@@ -25,12 +25,26 @@
...
     """
 
-    def __init__(self, offer_throughput, properties=None):  # pylint: disable=super-init-not-called
-        # type: (int, Dict[str, Any]) -> None
+    def __init__(self, offer_throughput=None, properties=None, **kwargs):  # pylint: disable=super-init-not-called
```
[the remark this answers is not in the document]
JericHunter: Done

sdk/cosmos/azure-cosmos/CHANGELOG.md
```diff
@@ -3,7 +3,7 @@
 ### 4.3.1 (Unreleased)
 
 #### Features Added
-
+- Added the ability to create containers and databases with autoscale properties for the sync and async clients. - [See PR #24976](https://github.com/Azure/azure-sdk-for-python/pull/24976)
```
tjprescott, line 6: @annatisch do we typically link to PRs in the change log? I've seen where we might link to a particular GH issue when fixing something.

sdk/cosmos/azure-cosmos/azure/cosmos/_base.py
```diff
@@ -673,3 +676,16 @@ def validate_cache_staleness_value(max_integrated_cache_staleness):
...
+
+def stringify_auto_scale(offer):
+    auto_scale_params = None
+    max_throughput = offer.auto_scale_max_throughput
+    increment_percent = offer.auto_scale_increment_percent
+    if max_throughput is not None:
+        auto_scale_params = {"maxThroughput": max_throughput}
+        if increment_percent is not None:
+            auto_scale_params["autoUpgradePolicy"] = {"throughputPolicy": {"incrementPercent": increment_percent}}
+    auto_scale_settings = json.dumps(auto_scale_params)
```
tjprescott, line 689: What happens if increment percent is provided but not max throughput?  Does the service error? Do we need to validate that?

sdk/cosmos/azure-cosmos/azure/cosmos/database.py
```diff
@@ -28,7 +28,7 @@
 from azure.core.tracing.decorator import distributed_trace  # type: ignore
 
 from ._cosmos_client_connection import CosmosClientConnection
-from ._base import build_options
+from ._base import build_options, stringify_auto_scale
```
tjprescott, line 31: You need to either change `stringify_auto_scale` to `_stringify_auto_scale` generally, or import it here `import stringify_auto_scale as _stringify_auto_scale`. Otherwise, this leaks as public API, which we do not want. 

My recommendation would be to change the method name to have the leading underscore.

tjprescott (changes requested):
A few comments here.

sdk/cosmos/azure-cosmos/azure/cosmos/database.py
JericHunter: Done!

sdk/cosmos/azure-cosmos/azure/cosmos/_base.py
JericHunter: Well you wouldn't have an increment percent without the max throughput being provided as well it wouldn't work if you did that because if the max throughput is not provided then you will not get the option to have an increment percent.

sdk/cosmos/azure-cosmos/CHANGELOG.md
JericHunter: I got the idea from Simon so I am not too sure myself.

sdk/cosmos/azure-cosmos/azure/cosmos/_base.py
tjprescott: Well, you could just set `offer.auto_scale_increment_percent` and not set the other. As written, the SDK would silently ignore this, and simply not send any autoscale information at all. Presumably it should send that info and let the service throw an error.

JericHunter: Well we don't set the inputs that way because it will not work as it is not deserialized. You can't do `offer.auto_scale_increment_percent` to set the increment percent you would have to do something like `Offer(auto_scale_max_throughput=5000, auto_upgrade_throughput_increment_percent=0)` if you tried to set increment percent without the max throughput you will get an error and it will not work.

sdk/cosmos/azure-cosmos/azure/cosmos/aio/_database.py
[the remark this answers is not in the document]
JericHunter: nvm just needed to change 'offer' to offer_throughput and it worked thank you!

sdk/cosmos/azure-cosmos/CHANGELOG.md
tjprescott: Per our discussion, let's not extend this pattern to Python. If you are linking to a Github issue for a bug fix, that is fine (nd encouraged).

sdk/cosmos/azure-cosmos/test/test_auto_scale.py
JericHunter: This has been deserialized properly so we can do that now

sdk/cosmos/azure-cosmos/azure/cosmos/__init__.py
```diff
@@ -26,7 +26,7 @@
 from .database import DatabaseProxy
 from .user import UserProxy
 from .scripts import ScriptsProxy
-from .offer import Offer
```
simorenoh, line 29: were we not going to expose both of them?

JericHunter: from my understanding i thought it was being replaced with throughput properties

JericHunter: I can add it back if necessary though

sdk/cosmos/azure-cosmos/CHANGELOG.md
```diff
@@ -3,7 +3,7 @@
 ### 4.3.1 (Unreleased)
 
 #### Features Added
-
+- Added the ability to create containers and databases with autoscale properties for the sync and async clients.
```
simorenoh, line 6: this is just containers since we didn't include the database methods in this PR - either remove database or add the changes to those files as well please

sdk/cosmos/azure-cosmos/azure/cosmos/aio/_database.py
```diff
@@ -745,7 +756,12 @@ async def get_throughput(self, **kwargs: Any) -> ThroughputProperties:
             response_hook(self.client_connection.last_response_headers, throughput_properties)
 
         return ThroughputProperties(offer_throughput=throughput_properties[0]["content"]["offerThroughput"],
-                                    properties=throughput_properties[0])
+                                    properties=throughput_properties[0],
+                                    auto_scale_max_throughput=throughput_properties[0]['content']
+                                    ['offerAutopilotSettings']['maxThroughput'],
+                                    auto_scale_increment_percent=throughput_properties[0]['content']
+                                    ['offerAutopilotSettings']['autoUpgradePolicy']['throughputPolicy'][
+                                        'incrementPercent'])
```
simorenoh, line 765: still unsure about how this will work for containers that don't have autoscale settings

simorenoh (commented):
A couple comments on things to fix/ look at, but overall looking good 👍

sdk/cosmos/azure-cosmos/azure/cosmos/offer.py
```diff
@@ -24,13 +24,24 @@
 
 class ThroughputProperties(object):
     """Represents the throughput properties in an Azure Cosmos DB SQL API container.
-    To read and update throughput properties use the associated methods on the :class:`Container`.
+
+    To read and update throughput properties, use the associated methods on the :class:`Container`.
+    If configuring auto-scale, both `auto_scale_max_throughput` and `auto_scale_increment_percent` must be set and
```
simorenoh, line 29: this is false, you only need auto_scale_max_throughput - please make sure all the documentation says as much

sdk/cosmos/azure-cosmos/azure/cosmos/offer.py
```diff
@@ -24,13 +24,24 @@
...
+    the value of `offer_throughput` will not be allowed to be set in conjunction with auto-scale.
+
+    :keyword int offer_throughput: The provisioned throughput in request units per second as a number.
+    :keyword int auto_scale_max_throughput: The max auto-scale throughput. It should have a valid throughput
+     value between 1000 and 1000000 inclusive, in increments of 1000.
+    :keyword int auto_scale_increment_percent: is the % from the base selected RU it increases at a given time,
+     up to a maximum default value of 0. The increment percent should be greater than or equal to zero.
```
simorenoh, line 36: this wording seems weird as well - what do you mean that the maximum is 0 if you then say that it should be greater than or equal to zero? did you mean to say minimum?

JericHunter: yes I meant minimum thats my bad will change that now

sdk/cosmos/azure-cosmos/azure/cosmos/offer.py
JericHunter: will do

sdk/cosmos/azure-cosmos/azure/cosmos/_base.py
```diff
@@ -673,3 +676,16 @@ def validate_cache_staleness_value(max_integrated_cache_staleness):
...
+
+
+def _stringify_auto_scale(offer):
+    auto_scale_params = None
+    max_throughput = offer.auto_scale_max_throughput
+    increment_percent = offer.auto_scale_increment_percent
+    if max_throughput is not None:
+        auto_scale_params = {"maxThroughput": max_throughput}
+        if increment_percent is not None:
+            auto_scale_params["autoUpgradePolicy"] = {"throughputPolicy": {"incrementPercent": increment_percent}}
```
tjprescott, line 688: This is still silently ignoring `increment_percent` if `max_throughput` is not supplied. If the server response to this situation is inadequate, then this should raise a local error.

sdk/cosmos/azure-cosmos/azure/cosmos/aio/_container.py
```diff
@@ -597,6 +597,19 @@ async def get_throughput(self, **kwargs: Any) -> ThroughputProperties:
...
+                throughput_properties[0]['content']['offerAutopilotSettings']['maxThroughput'],
+                                        auto_scale_increment_percent=throughput_properties[0]['content']
+                                        ['offerAutopilotSettings']['autoUpgradePolicy']['throughputPolicy'][
+                                            'incrementPercent'])
+
+        elif 'offerAutopilotSettings' in throughput_properties[0]['content']:
+            return ThroughputProperties(properties=throughput_properties[0],
+                                        auto_scale_max_throughput=throughput_properties[0]['content']
+                                        ['offerAutopilotSettings']['maxThroughput'])
+
```
tjprescott, line 612: Why is this conditional logic necessary? It seems like this has a high chance of silently ignoring things.

sdk/cosmos/azure-cosmos/azure/cosmos/aio/_cosmos_client.py
```diff
@@ -224,7 +225,18 @@ async def create_database(  # pylint: disable=redefined-builtin
...
+                    request_options['autoUpgradePolicy'] = _stringify_auto_scale(offer=offer_throughput)
+                elif offer_throughput.auto_scale_increment_percent:
+                    raise ValueError("auto_scale_max_throughput must be supplied in "
+                                     "conjunction with auto_scale_increment_percent")
+                elif offer_throughput.offer_throughput:
+                    request_options["offerThroughput"] = offer_throughput.offer_throughput
+
+            except AttributeError:
+                if isinstance(offer_throughput, int):
+                    request_options["offerThroughput"] = offer_throughput
```
tjprescott, line 239: This accepts either an integer or complex object?  Why are we doing that? And why is `offer_throughput.offer_throughput` an acceptable input?

sdk/cosmos/azure-cosmos/azure/cosmos/aio/_cosmos_client.py
```diff
@@ -224,7 +225,18 @@ async def create_database(  # pylint: disable=redefined-builtin
         response_hook = kwargs.pop('response_hook', None)
         offer_throughput = kwargs.pop('offer_throughput', None)
         if offer_throughput is not None:
-            request_options["offerThroughput"] = offer_throughput
+            try:
+                if offer_throughput.auto_scale_max_throughput:
+                    request_options['autoUpgradePolicy'] = _stringify_auto_scale(offer=offer_throughput)
+                elif offer_throughput.auto_scale_increment_percent:
+                    raise ValueError("auto_scale_max_throughput must be supplied in "
+                                     "conjunction with auto_scale_increment_percent")
```
tjprescott, line 233: It doesn't look like we are using `auto_scale_increment_percent` in the above logic?

sdk/cosmos/azure-cosmos/azure/cosmos/aio/_database.py
```diff
@@ -228,7 +228,18 @@ async def create_container(
...
+                    request_options['autoUpgradePolicy'] = _stringify_auto_scale(offer=offer_throughput)
+                elif offer_throughput.auto_scale_increment_percent:
+                    raise ValueError("auto_scale_max_throughput must be supplied in "
+                                     "conjunction with auto_scale_increment_percent")
+                elif offer_throughput.offer_throughput:
+                    request_options["offerThroughput"] = offer_throughput.offer_throughput
+
+            except AttributeError:
+                if isinstance(offer_throughput, int):
+                    request_options["offerThroughput"] = offer_throughput
```
tjprescott, line 242: Why is this logic being duplicated? It seems this logic should be refactored into one place.

sdk/cosmos/azure-cosmos/azure/cosmos/aio/_database.py
```diff
@@ -744,10 +755,22 @@ async def get_throughput(self, **kwargs: Any) -> ThroughputProperties:
...
+                throughput_properties[0]['content']['offerAutopilotSettings']['maxThroughput'],
+                                        auto_scale_increment_percent=throughput_properties[0]['content']
+                                        ['offerAutopilotSettings']['autoUpgradePolicy']['throughputPolicy'][
+                                            'incrementPercent'])
+
+        elif 'offerAutopilotSettings' in throughput_properties[0]['content']:
+            return ThroughputProperties(properties=throughput_properties[0],
+                                        auto_scale_max_throughput=throughput_properties[0]['content']
+                                        ['offerAutopilotSettings']['maxThroughput'])
+
```
tjprescott, line 770: This also being duplicated is particularly problematic because it is very hard to follow as written.

sdk/cosmos/azure-cosmos/azure/cosmos/container.py
```diff
@@ -668,9 +668,23 @@ def get_throughput(self, **kwargs):
...
+                throughput_properties[0]['content']['offerAutopilotSettings']['maxThroughput'],
+                                        auto_scale_increment_percent=throughput_properties[0]['content']
+                                        ['offerAutopilotSettings']['autoUpgradePolicy']['throughputPolicy'][
+                                            'incrementPercent'])
+
+        elif 'offerAutopilotSettings' in throughput_properties[0]['content']:
+            return ThroughputProperties(properties=throughput_properties[0],
+                                        auto_scale_max_throughput=throughput_properties[0]['content']
+                                        ['offerAutopilotSettings']['maxThroughput'])
+
```
tjprescott, line 683: Okay, yes, three times is two too many.

tjprescott (changes requested):
There seems to be a lot of duplicated code and there still appears to be silent ignoring of parameter combinations.

sdk/cosmos/azure-cosmos/azure/cosmos/aio/_container.py
JericHunter: This logic is used for deserializing the auto scale properties. We cannot just return them all at once as we need to check if the container or database has autoscale. If we tried to return all the properties then it would fail because if we tried to use a container with no autoscale it would tell us that some of the properties we are returning are not found.

sdk/cosmos/azure-cosmos/azure/cosmos/aio/_cosmos_client.py
JericHunter: the elif statement is just for if we are only supplying the `auto_scale_increment_percent` if we only supplied that then it would raise the error. However if we supplied the `auto_scale_max_throughput` we would be able to supply both and everything would work as expected.

sdk/cosmos/azure-cosmos/azure/cosmos/aio/_database.py
JericHunter: This logic is being duplicated because this is for the asynchronous version so we wanted to have the same code being used in the sync version here as well

sdk/cosmos/azure-cosmos/azure/cosmos/aio/_database.py
JericHunter: The reason why this code is being repeated is because of the same thing I said before we want to have the async match the sync version of the code.

sdk/cosmos/azure-cosmos/azure/cosmos/container.py
JericHunter: Also to note we are adding the ability to provision autoscale with both containers and databases so this is why you will see this logic in so many different places as we need to provide the logic for both sync and async versions.

sdk/cosmos/azure-cosmos/azure/cosmos/aio/_container.py
JericHunter: I have really tested this out plenty and this doesn't ignore anything that I am aware of and everything works how it should.

sdk/cosmos/azure-cosmos/azure/cosmos/_base.py
JericHunter: We raise an error for if `increment_percent` is supplied without `max_throughput` so this is fine and we would want increment percent to not exist if we are not supplying the max throughput.

sdk/cosmos/azure-cosmos/azure/cosmos/container.py
tjprescott: This logic can and should still be extracted out so that it only exists in one place.

JericHunter: Okay sounds good I will get on that then

sdk/cosmos/azure-cosmos/azure/cosmos/aio/_cosmos_client.py
simorenoh: @tjprescott the method traditionally only accepted an int but since we want to use the same option for autoscale properties, we had to change the signature to accept the object as well

This decision was made since otherwise we would be creating a new object with only the two autoscale properties and adding additional options to the method that are throughput related - this way we instead use what we already had while still adding the new functionality

sdk/cosmos/azure-cosmos/azure/cosmos/database.py
```diff
@@ -232,9 +232,7 @@ def create_container(
                 UserWarning,
             )
             request_options["populateQueryMetrics"] = populate_query_metrics
-        if offer_throughput is not None:
```
simorenoh, line 235: you should keep the checks for not None and do _throughput_settings() if it's not None

simorenoh: nvm im silly, didn't see the check in the _throughput_settings method

simorenoh (approved):
LGTM, thanks a lot Jeric!

sdk/cosmos/azure-cosmos/azure/cosmos/_base.py
```diff
@@ -674,3 +676,53 @@ def validate_cache_staleness_value(max_integrated_cache_staleness):
...
+                request_options['autoUpgradePolicy'] = _stringify_auto_scale(offer=offer_throughput)
+            elif offer_throughput.auto_scale_increment_percent:
+                raise ValueError("auto_scale_max_throughput must be supplied in "
+                                 "conjunction with auto_scale_increment_percent")
+            elif offer_throughput.offer_throughput:
+                request_options["offerThroughput"] = offer_throughput.offer_throughput
+
+        except AttributeError:
+            if isinstance(offer_throughput, int):
+                request_options["offerThroughput"] = offer_throughput
```
tjprescott, line 709: I would write it something like this to be clearer. You should add comments to indicate edge cases and so forth. 

```Python
if offer is None:
  return

if isinstance(offer, int):
  # do the int case here...
  return
elif hasattr(offer, "offer_throughput"):
  # the offer.offer_throughput here...
  return

# then this is the "normal" case that has validation
max_throughput = offer.auto_scale_max_throughput
increment_percent  = offer.auto_scale_increment_percent
if not(max_throughput and increment_percent):
  raise ValueError("...")
else:
  options['autoUpgradePolicy'] = _stringify_auto_scale(offer=offer)
```

tjprescott (changes requested):

sdk/cosmos/azure-cosmos/azure/cosmos/_base.py
```diff
@@ -674,3 +676,53 @@ def validate_cache_staleness_value(max_integrated_cache_staleness):
...
+
+
+def _deserialize_throughput(throughput):
+    throughput_properties = throughput
+    if 'offerAutopilotSettings' in throughput_properties[0]['content'] and 'autoUpgradePolicy' in \
```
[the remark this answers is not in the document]
simorenoh: This is the way the response is returned from the service and has always been this way - this is not new functionality that we are introducing. In order to get the throughput, we query the backend and as such get a list of results. For this to break in the future it would have to be a backend change.

sdk/cosmos/azure-cosmos/azure/cosmos/_base.py
```diff
@@ -674,3 +676,53 @@ def validate_cache_staleness_value(max_integrated_cache_staleness):
...
+    offer_throughput = offer
+    request_options = options
+    if offer_throughput is not None:
+        try:
+            if offer_throughput.auto_scale_max_throughput:
+                request_options['autoUpgradePolicy'] = _stringify_auto_scale(offer=offer_throughput)
+            elif offer_throughput.auto_scale_increment_percent:
+                raise ValueError("auto_scale_max_throughput must be supplied in "
+                                 "conjunction with auto_scale_increment_percent")
+            elif offer_throughput.offer_throughput:
```
annatisch, line 704: If this was `if` instead of `elif` what would the service error be?

sdk/cosmos/azure-cosmos/azure/cosmos/_base.py
```diff
@@ -674,3 +676,53 @@ def validate_cache_staleness_value(max_integrated_cache_staleness):
...
+                                    ['offerAutopilotSettings']['autoUpgradePolicy']['throughputPolicy'][
+                                        'incrementPercent'])
+
+    if 'offerAutopilotSettings' in throughput_properties[0]['content']:
+        return ThroughputProperties(properties=throughput_properties[0],
+                                    auto_scale_max_throughput=throughput_properties[0]['content'][
+                                        'offerAutopilotSettings']['maxThroughput'])
+
+    return ThroughputProperties(offer_throughput=throughput_properties[0]["content"]["offerThroughput"],
+                                properties=throughput_properties[0])
```
annatisch, line 728: I feel like this section could be simplified a bit with:
```python
def _deserialize_throughput(throughput):
    try:
        max_throughput = throughput_properties[0]['content']['offerAutopilotSettings']['maxThroughput']
    except (KeyError, TypeError):  #  Adding TypeError just in case one of these dicts is None
        max_throughput = None
    try:
        increment_percent = throughput_properties[0]['content']['offerAutopilotSettings']['autoUpgradePolicy']['throughputPolicy']['incrementPercent']
    except (KeyError, TypeError):
        increment_percent = None
    try:
        throughput = throughput_properties[0]["content"]["offerThroughput"]
    except (KeyError, TypeError):
        throughput = None
    return ThroughputProperties(
        auto_scale_max_throughput=max_throughput,
        auto_scale_increment_percent=increment_percent,
        offer_throughput=throughput,
        properties=throughput_properties[0]
    )
```

sdk/cosmos/azure-cosmos/azure/cosmos/aio/_cosmos_client.py
```diff
@@ -223,8 +224,7 @@ async def create_database(  # pylint: disable=redefined-builtin
         request_options = _build_options(kwargs)
         response_hook = kwargs.pop('response_hook', None)
         offer_throughput = kwargs.pop('offer_throughput', None)
-        if offer_throughput is not None:
-            request_options["offerThroughput"] = offer_throughput
+        _set_throughput_options(offer=offer_throughput, options=request_options)
```
annatisch, line 227: We need to update the type in the docstring.

sdk/cosmos/azure-cosmos/azure/cosmos/aio/_database.py
```diff
@@ -227,8 +227,7 @@ async def create_container(
         request_options = _build_options(kwargs)
         response_hook = kwargs.pop('response_hook', None)
         offer_throughput = kwargs.pop('offer_throughput', None)
-        if offer_throughput is not None:
-            request_options["offerThroughput"] = offer_throughput
+        _set_throughput_options(offer=offer_throughput, options=request_options)
```
annatisch, line 230: Also need to update the type in this docstring.

sdk/cosmos/azure-cosmos/azure/cosmos/cosmos_client.py
```diff
@@ -230,7 +231,7 @@ def create_database(  # pylint: disable=redefined-builtin
         self,
         id,  # type: str
         populate_query_metrics=None,  # type: Optional[bool]
-        offer_throughput=None,  # type: Optional[int]
+        offer_throughput=None,  # type: Optional[Union[int, ThroughputProperties]]
```
annatisch, line 234: Let's add this to the docstring as well - though it will probably need to be two lines to get the right formatting (this applies to all the docstring updates - whether 'keyword' or 'param'):
```python
"""
:param offer_throughput: The provisioned throughput for this offer.
:paramtype offer_throughput: int or ~azure.cosmos.ThroughputProperties
"""
```

sdk/cosmos/azure-cosmos/azure/cosmos/_base.py
```diff
@@ -674,3 +676,53 @@ def validate_cache_staleness_value(max_integrated_cache_staleness):
...
+
+
+def _set_throughput_options(offer, options):
+    offer_throughput = offer
+    request_options = options
+    if offer_throughput is not None:
+        try:
+            if offer_throughput.auto_scale_max_throughput:
+                request_options['autoUpgradePolicy'] = _stringify_auto_scale(offer=offer_throughput)
+            elif offer_throughput.auto_scale_increment_percent:
```
annatisch, line 701: This will still allow a value of 0

sdk/cosmos/azure-cosmos/test/test_auto_scale.py
```diff
@@ -0,0 +1,141 @@
...
+        created_container = self.created_database.create_container(
+            id='container_with_auto_scale_settings',
+            partition_key=PartitionKey(path="/id"),
+            offer_throughput=ThroughputProperties(auto_scale_max_throughput=7000, auto_scale_increment_percent=0)
+
+        )
+        created_container_properties = created_container.get_throughput()
+        # Testing the input value of the max_throughput
+        self.assertEqual(
+            created_container_properties.auto_scale_max_throughput, 7000)
```
annatisch, line 60: Could we also add assert statements for the other properties? i.e.
```python
self.assertEqual(created_container_properties.auto_scale_increment_percent, 0)
self.assertEqual(created_container_properties.offer_throughput, None)
```

sdk/cosmos/azure-cosmos/test/test_auto_scale.py
```diff
@@ -0,0 +1,141 @@
...
+        created_container_properties = created_container.get_throughput()
+        # Testing the input value of the increment_percentage
+        self.assertEqual(
+            created_container_properties.auto_scale_increment_percent, 1)
+
+        self.created_database.delete_container(created_container)
+
+    def test_auto_scale_settings(self):
+        # Testing for wrong attributes for the auto_scale_settings in the created container
+        with pytest.raises(AssertionError) as e:
```
annatisch, line 88: This should raise a `TypeError` - I didn't see in the code where the `AssertionError` would be coming from?

sdk/cosmos/azure-cosmos/azure/cosmos/_base.py
JericHunter: allow a value of 0 for what? can you clarify this a bit?

sdk/cosmos/azure-cosmos/test/test_auto_scale.py
JericHunter: I separated the tests so some assert for max throughput others assert for increment percent. the methods are also pretty much named accordingly. so do you think I should just reformat the tests entirely or add in a test for offer throughput to be none?

sdk/cosmos/azure-cosmos/azure/cosmos/_base.py
JericHunter: you would get an error because if you have an increment percent at all even with the max throughput you will get the value error.

sdk/cosmos/azure-cosmos/azure/cosmos/_base.py
tjprescott: Whatever the logic actually should, the general thrust is to make it more readable and self-documenting.

JericHunter: Okay understood, will take a deeper look into it.

JericHunter: I did my best to make it more readable let me know what you think!

sdk/cosmos/azure-cosmos/azure/cosmos/_base.py
```diff
@@ -679,6 +679,7 @@ def validate_cache_staleness_value(max_integrated_cache_staleness):
 
 
 def _stringify_auto_scale(offer):
+    # type: (Dict[str, Any]) -> Any
```
tjprescott, line 682: Please use Python 3 type hints exclusively. We no longer support Python 2 at all and thus don't want to proliferate Python 2 type hints.

https://docs.python.org/3/library/typing.html

sdk/cosmos/azure-cosmos/azure/cosmos/_base.py
```diff
@@ -692,37 +693,43 @@ def _stringify_auto_scale(offer):
...
+            request_options["offerThroughput"] = offer_throughput
 
+        max_throughput = offer.auto_scale_max_throughput
+        increment_percent = offer.auto_scale_increment_percent
+
+        if max_throughput:
+            request_options['autoUpgradePolicy'] = _stringify_auto_scale(offer=offer_throughput)
+        elif increment_percent:
+            raise ValueError("auto_scale_max_throughput must be supplied in "
+                             "conjunction with auto_scale_increment_percent")
```
tjprescott, line 711:
```suggestion
    if offer is None:
      return
      
    if isinstance(offer, int):
        request_options["offerThroughput"] = offer_throughput

    max_throughput = offer.auto_scale_max_throughput
    increment_percent = offer.auto_scale_increment_percent

    if max_throughput:
        request_options['autoUpgradePolicy'] = _stringify_auto_scale(offer=offer_throughput)
    elif increment_percent:
        raise ValueError("auto_scale_max_throughput must be supplied in "
                         "conjunction with auto_scale_increment_percent")
```

sdk/cosmos/azure-cosmos/azure/cosmos/_base.py
```diff
@@ -692,37 +693,43 @@ def _stringify_auto_scale(offer):
...
+    try:
+        throughput = throughput_properties[0]["content"]["offerThroughput"]
+    except (KeyError, TypeError):
+        throughput = None
+    return ThroughputProperties(
+        auto_scale_max_throughput=max_throughput,
+        auto_scale_increment_percent=increment_percent,
+        offer_throughput=throughput,
+        properties=throughput_properties[0]
+    )
```
tjprescott, line 735: This looks much cleaner!

tjprescott (changes requested):
Some follow-up comments, but getting there!

sdk/cosmos/azure-cosmos/azure/cosmos/_base.py
JericHunter: fixed!

sdk/cosmos/azure-cosmos/azure/cosmos/_base.py
JericHunter: added these changes with recent commit

sdk/cosmos/azure-cosmos/test/test_auto_scale.py
annatisch: I would just add a little assert statement coverage to the existing tests. That should be sufficient.

sdk/cosmos/azure-cosmos/azure/cosmos/_base.py
tjprescott: This comment is still unaddressed.

tjprescott (changes requested):
One lingering comment regarding `_stringify_auto_scale`

JericHunter:
@tjprescott I don't know what that is supposed to look like from what I understand this should be fine and it is how it's done in other SDKs.

sdk/cosmos/azure-cosmos/azure/cosmos/_base.py
```diff
@@ -674,3 +676,60 @@ def validate_cache_staleness_value(max_integrated_cache_staleness):
...
+    if max_throughput is not None:
+        auto_scale_params = {"maxThroughput": max_throughput}
+        if increment_percent is not None:
+            auto_scale_params["autoUpgradePolicy"] = {"throughputPolicy": {"incrementPercent": increment_percent}}
+    auto_scale_settings = json.dumps(auto_scale_params)
+
+    return auto_scale_settings
+
+
+def _set_throughput_options(offer: int, options: Dict[str, Any]) -> Any:
```
simorenoh, line 694: offer is not necessarily an int, it can also be a ThroughputProperties object

JericHunter: changed it, thanks for pointing that out

simorenoh (commented):
one minor comment on typehint

sdk/cosmos/azure-cosmos/azure/cosmos/_base.py
simorenoh: @tjprescott I'm unsure what you mean by this or what's missing here - if the user only supplied increment_percent and not max_throughput, then the client returns an error to the user (line 709 in this same file) which is the behavior you described to begin with and the behavior we are aiming for. Could you expand on what's missing, or what type of flaw you are talking about?

tjprescott: @simorenoh @JericHunter all of that validation is provided in the setter and the SDK will emit a client error if it is violated. Thus, there is no reason to duplicate validation here. It should simply plumb everything through.

sdk/cosmos/azure-cosmos/azure/cosmos/_base.py
```diff
@@ -674,3 +676,60 @@ def validate_cache_staleness_value(max_integrated_cache_staleness):
...
+    auto_scale_params = None
+    max_throughput = offer.auto_scale_max_throughput
+    increment_percent = offer.auto_scale_increment_percent
+    if max_throughput is not None:
+        auto_scale_params = {"maxThroughput": max_throughput}
+        if increment_percent is not None:
+            auto_scale_params["autoUpgradePolicy"] = {"throughputPolicy": {"incrementPercent": increment_percent}}
+    auto_scale_settings = json.dumps(auto_scale_params)
+
+    return auto_scale_settings
```
tjprescott, line 691: This eliminates the unnecessary validation.

```suggestion
    auto_scale_params = None
    max_throughput = offer.auto_scale_max_throughput
    increment_percent = offer.auto_scale_increment_percent
    if max_throughput is not None:
        auto_scale_params = {"maxThroughput": max_throughput}
    if increment_percent is not None:
        auto_scale_params["autoUpgradePolicy"] = {"throughputPolicy": {"incrementPercent": increment_percent}}
    auto_scale_settings = json.dumps(auto_scale_params)

    return auto_scale_settings
```

JericHunter: Thank you, I committed the suggestion.

tjprescott (approved):

simorenoh (approved):

sdk/cosmos/azure-cosmos/azure/cosmos/_base.py
```diff
@@ -674,3 +676,60 @@ def validate_cache_staleness_value(max_integrated_cache_staleness):
...
+
+def _set_throughput_options(offer: Union[int, ThroughputProperties], options: Dict[str, Any]) -> Any:
+    offer_throughput = offer
+    request_options = options
+
+    if offer is None:
+        return
+
+    if isinstance(offer, int):
+        request_options["offerThroughput"] = offer_throughput
```
annatisch, line 702: Should this be followed by a return statement?
If offer _is_ an int, then line 704/705 will just raise an AttributeError

sdk/cosmos/azure-cosmos/azure/cosmos/_base.py
```diff
@@ -674,3 +676,60 @@ def validate_cache_staleness_value(max_integrated_cache_staleness):
...
+    if increment_percent is not None:
+        auto_scale_params["autoUpgradePolicy"] = {"throughputPolicy": {"incrementPercent": increment_percent}}
+    auto_scale_settings = json.dumps(auto_scale_params)
+
+    return auto_scale_settings
+
+
+def _set_throughput_options(offer: Union[int, ThroughputProperties], options: Dict[str, Any]) -> Any:
+    offer_throughput = offer
+    request_options = options
```
annatisch, line 696: These two lines are redundant.

sdk/cosmos/azure-cosmos/azure/cosmos/aio/_cosmos_client.py
```diff
@@ -197,6 +198,7 @@ async def create_database(  # pylint: disable=redefined-builtin
 
         :param str id: ID (name) of the database to create.
         :keyword int offer_throughput: The provisioned throughput for this offer.
+        :keyword type offer_throughput: int or ~azure.cosmos.ThroughputProperties.
```
annatisch, line 201:
```suggestion
        :paramtype offer_throughput: int or ~azure.cosmos.ThroughputProperties.
```

sdk/cosmos/azure-cosmos/azure/cosmos/aio/_database.py
```diff
@@ -165,6 +165,7 @@ async def create_container(
         :keyword int default_ttl: Default time to live (TTL) for items in the container.
             If unspecified, items do not expire.
         :keyword int offer_throughput: The provisioned throughput for this offer.
+        :keyword type offer_throughput: int or ~azure.cosmos.ThroughputProperties.
```
annatisch, line 168:
```suggestion
        :paramtype offer_throughput: int or ~azure.cosmos.ThroughputProperties.
```

sdk/cosmos/azure-cosmos/azure/cosmos/cosmos_client.py
```diff
@@ -239,6 +240,7 @@ def create_database(  # pylint: disable=redefined-builtin
 
         :param id: ID (name) of the database to create.
         :param int offer_throughput: The provisioned throughput for this offer.
+        :param type offer_throughput: int or ~azure.cosmos.ThroughputProperties.
```
annatisch, line 243:
```suggestion
        :paramtype offer_throughput: int or ~azure.cosmos.ThroughputProperties.
```

sdk/cosmos/azure-cosmos/azure/cosmos/database.py
```diff
@@ -170,6 +170,7 @@ def create_container(
         :param indexing_policy: The indexing policy to apply to the container.
         :param default_ttl: Default time to live (TTL) for items in the container. If unspecified, items do not expire.
         :param offer_throughput: The provisioned throughput for this offer.
+        :param type offer_throughput: int or ~azure.cosmos.ThroughputProperties.
```
annatisch, line 173:
```suggestion
        :paramtype offer_throughput: int or ~azure.cosmos.ThroughputProperties.
```

sdk/cosmos/azure-cosmos/azure/cosmos/_base.py
```diff
@@ -674,3 +676,60 @@ def validate_cache_staleness_value(max_integrated_cache_staleness):
...
+
+    if offer is None:
+        return
+
+    if isinstance(offer, int):
+        request_options["offerThroughput"] = offer_throughput
+
+    max_throughput = offer.auto_scale_max_throughput
+    increment_percent = offer.auto_scale_increment_percent
```
annatisch, line 705: I thought we were still going to send `offer.offer_throughput` if a customer had set it?

sdk/cosmos/azure-cosmos/azure/cosmos/_base.py
JericHunter: I have returned that bit of code to its original state when you first made the suggestion, I didn't realize that what Travis had suggested broke the code. So we should be good to go now.

sdk/cosmos/azure-cosmos/azure/cosmos/_base.py
JericHunter: I returned it back to its original state as I said before so with that hopefully we are all good.

sdk/cosmos/azure-cosmos/azure/cosmos/aio/_cosmos_client.py
```diff
@@ -197,6 +198,7 @@ async def create_database(  # pylint: disable=redefined-builtin
 
         :param str id: ID (name) of the database to create.
         :keyword int offer_throughput: The provisioned throughput for this offer.
+        :paramtype offer_throughput: int or ~azure.cosmos.ThroughputProperties.
```
simorenoh, line 201: this is what it should look like, the way the other keywords work when they have complex typings (look at match_condition or response_hook for instance). Please fix this in all other instances that need it.
```suggestion
        :keyword offer_throughput: The provisioned throughput for this offer.
        :paramtype offer_throughput: int or ~azure.cosmos.ThroughputProperties.
```

simorenoh (changes requested):

sdk/cosmos/azure-cosmos/azure/cosmos/_base.py
```diff
@@ -674,3 +676,61 @@ def validate_cache_staleness_value(max_integrated_cache_staleness):
...
+    return auto_scale_settings
+
+
+def _set_throughput_options(offer: Union[int, ThroughputProperties], request_options: Dict[str, Any]) -> Any:
+    if offer is not None:
+        try:
+            max_throughput = offer.auto_scale_max_throughput
+            increment_percent = offer.auto_scale_increment_percent
+
+            if max_throughput:
```
annatisch, line 700:
```suggestion
            if max_throughput is not None:
```

sdk/cosmos/azure-cosmos/azure/cosmos/_base.py
```diff
@@ -674,3 +676,61 @@ def validate_cache_staleness_value(max_integrated_cache_staleness):
     if max_integrated_cache_staleness < 0:
         raise ValueError("Parameter 'max_integrated_cache_staleness_in_ms' can only be an "
                          "integer greater than or equal to zero")
+
+
+def _stringify_auto_scale(offer: Dict[str, Any]) -> Any:
+    auto_scale_params = None
+    max_throughput = offer.auto_scale_max_throughput
+    increment_percent = offer.auto_scale_increment_percent
+    if max_throughput is not None:
```
annatisch, line 685: We should remove this line - as this check is already on line 700.
It also means if a user has specified `increment_percent` but not `max_throughput` they will get a `NameError` on line 688.

sdk/cosmos/azure-cosmos/azure/cosmos/cosmos_client.py
```diff
@@ -281,7 +282,7 @@ def create_database_if_not_exists(  # pylint: disable=redefined-builtin
         self,
         id,  # type: str
         populate_query_metrics=None,  # type: Optional[bool]
-        offer_throughput=None,  # type: Optional[int]
+        offer_throughput=None,  # type: Optional[Union[int, ThroughputProperties]]
```
annatisch, line 285: Please also fix line 300

sdk/cosmos/azure-cosmos/azure/cosmos/aio/_database.py
```diff
@@ -258,7 +258,7 @@ async def create_container_if_not_exists(
         :keyword dict[str, str] indexing_policy: The indexing policy to apply to the container.
         :keyword int default_ttl: Default time to live (TTL) for items in the container.
             If unspecified, items do not expire.
-        :keyword int offer_throughput: The provisioned throughput for this offer.
+        :keyword offer_throughput: The provisioned throughput for this offer.
```
annatisch, line 261: Please add:
`:paramtype offer_throughput: int or ~azure.cosmos.ThroughputProperties.`

sdk/cosmos/azure-cosmos/azure/cosmos/database.py
```diff
@@ -252,7 +251,7 @@ def create_container_if_not_exists(
         indexing_policy=None,  # type: Optional[Dict[str, Any]]
         default_ttl=None,  # type: Optional[int]
         populate_query_metrics=None,  # type: Optional[bool]
-        offer_throughput=None,  # type: Optional[int]
+        offer_throughput=None,  # type: Optional[Union[int, ThroughputProperties]]
```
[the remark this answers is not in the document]
annatisch: Please add the paramtype to line 272:
`:paramtype offer_throughput: int or ~azure.cosmos.ThroughputProperties.`

annatisch (approved):

sdk/cosmos/azure-cosmos/azure/cosmos/cosmos_client.py
```diff
@@ -297,6 +298,7 @@ def create_database_if_not_exists(  # pylint: disable=redefined-builtin
         :param id: ID (name) of the database to read or create.
         :param bool populate_query_metrics: Enable returning query metrics in response headers.
         :param int offer_throughput: The provisioned throughput for this offer.
+        :paramtype offer_throughput: int or ~azure.cosmos.ThroughputProperties.
```
simorenoh, line 301:
```suggestion
        :param offer_throughput: The provisioned throughput for this offer.
        :type offer_throughput: int or ~azure.cosmos.ThroughputProperties.
```

sdk/cosmos/azure-cosmos/azure/cosmos/database.py
```diff
@@ -170,6 +170,7 @@ def create_container(
         :param indexing_policy: The indexing policy to apply to the container.
         :param default_ttl: Default time to live (TTL) for items in the container. If unspecified, items do not expire.
         :param offer_throughput: The provisioned throughput for this offer.
+        :paramtype offer_throughput: int or ~azure.cosmos.ThroughputProperties.
```
simorenoh, line 173:
```suggestion
        :param offer_throughput: The provisioned throughput for this offer.
        :type offer_throughput: int or ~azure.cosmos.ThroughputProperties.
```

simorenoh (approved):

Issue #24273, opened 2022-05-02
[core] add async test for offline test_compress_compressed_no_header_offline


iscai-msft:
Right now we only have the sync test for test_compress_compressed_no_header_offline. Please see the original [here](https://github.com/Azure/azure-sdk-for-python/blob/33c845aa7674c26bf851313daeae23e2f3f5459c/sdk/core/azure-core/tests/async_tests/test_streaming_async.py#L94),

cc @marcusc2

iscai-msft:
@kristapratico assigning you because I can't assign one of the impact students

kristapratico:
Marcus is working on this issue

HimanshuBarak:
Hey @kristapratico , it this issue still up for grabs?? . If so I would love to work on this. I would just need a little help with the setup. but I am eager to contribute.

kristapratico:
@HimanshuBarak I don't believe anyone is working on this issue, so feel free to give it a try and let us know if you have any questions.

HimanshuBarak:
Hey @kristapratico  I have cloned the repo and written the test. The test is successfully raising the UnicodeDecodeError error. Can I directly raise the PR ? or are there any other tests requisite I need to test first?

kristapratico:
@HimanshuBarak feel free to open it and just tag this issue in the description. If anything else is needed by the reviewers we can cover that in the PR. Thanks!

HimanshuBarak:
Hey @kristapratico I have raised the PR https://github.com/Azure/azure-sdk-for-python/pull/33800. However I see some of the checks are failing. Are all of these checks mandatory ? Also is there a way to run these checks locally using tox?? Asking to know how I should proceed to debug them.

kristapratico:
Hey @HimanshuBarak, you can run the checks locally with tox. Here are [instructions](https://github.com/Azure/azure-sdk-for-python/blob/main/doc/dev/tests.md#tox) on how to do that. It is possible that the errors are unrelated to your PR. I believe there are currently a few core tests failing that we are working on fixing ASAP. The reviewers should be able to help distinguish the relevant errors from the unrelated ones.

HimanshuBarak:
Sounds good. I will try to debug on my end. Meanwhile I will look around to see I could contribute anywhere else. Do you have any suggestions for any other similar issues??. I am looking to be a first time contributor to open source and am super eager to learn.

kristapratico:
[...]

PR #25000, opened 2022-06-28
Issue #24273 add async test for offline test compress compressed no header offline

marcusc2:
This PR will:
- mark async test_compress_compressed_no_header as live test only


sdk/core/azure-core/tests/test_streaming.py
```diff
@@ -92,6 +92,19 @@ def test_decompress_compressed_no_header(http_request):
     except UnicodeDecodeError:
         pass
 
+@pytest.mark.parametrize("http_request", HTTP_REQUESTS)
```
iscai-msft, line 95: @marcusc2 this should be an async call, and should be in `test_streaming_async.py`

sdk/core/azure-core/tests/testserver_tests/coretestserver/coretestserver/test_routes/streams.py
```diff
@@ -31,6 +33,15 @@ def streaming_test():
 def stream_compressed_header_error():
     yield b'test'
 
+def stream_compressed_no_header():
```
iscai-msft, line 36: I think there's already a sync version of this test, so I don't think you have to add a new endpoint in the testserver, can you double check? Thanks!

PR #25028, opened 2022-06-29
Add common github action event handler for check enforcer

benbp:
Adding check enforcer v2 action to repository. See https://github.com/Azure/azure-sdk-tools/pull/3287 for details.

### DIFF
--- a/.github/workflows/event.yml
+++ b/.github/workflows/event.yml
@@ -1,0 +1,16 @@
+name: GitHub Event Handler
+
+on:
+  check_suite:
+    types: [completed]
+  issue_comment:
+    types: [created]
+
+jobs:
+  event-handler:
+    name: Handle ${{ github.event_name }} ${{ github.event.action }} event
+    runs-on: ubuntu-latest
+    steps:
+      - uses: azure/azure-sdk-actions@main
+        with:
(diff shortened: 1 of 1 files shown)


scbedd (approved):
What are your thoughts on rollout with this thing? Considered ready to go?

.github/workflows/event.yml
```diff
@@ -0,0 +1,16 @@
...
+on:
+  check_suite:
+    types: [completed]
+  issue_comment:
+    types: [created]
+
+jobs:
+  event-handler:
+    name: Handle ${{ github.event_name }} ${{ github.event.action }} event
+    runs-on: ubuntu-latest
```
weshaggard, line 12: We haven't thought much about this yet but do we want to have a rolling image? (i.e. -latest)? I know we don't do that for pipelines because we need repeatability but maybe it is OK for actions but updates could definitely break things.

weshaggard (approved):
Lets wait until releases are over.

.github/workflows/event.yml
benbp: Good point. Or maybe keep a rolling image for tools only so we catch issues early?

PR #25066, opened 2022-07-04
Move test location of ARR back to westus2, because service is not supported in westus

MichaelZp0:
# Description

The default test location changed a while ago to westus instead of westus2. Unfortunately the Azure Remote Rendering service is not supported in westus and thus all tests are failing. This PR specifies westus2 for the location of the tests of the ARR test.


sdk/remoterendering/tests.yml
```diff
@@ -4,6 +4,7 @@ stages:
   - template: ../../eng/pipelines/templates/stages/archetype-sdk-tests.yml
     parameters:
       ServiceDirectory: remoterendering
+      Location: westus2
```
benbp, line 7: FYI @mikeharder the only other options seem to be south central or east us regions, which are further from the agent VM location (https://docs.microsoft.com/en-us/azure/remote-rendering/reference/regions). I'm more inclined to use westus2, but it might be worth trying the live tests in another US region to see if they work ok.

mikeharder: Unless and until we hear otherwise, westus2 is fine with me for the limited scenarios that require it.

mikeharder (approved):

sdk/remoterendering/azure-mixedreality-remoterendering/CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # Release History
 
+## 1.0.0b3 (2022-07-04)
+
 ## 1.0.0b2 (Unreleased)
```
tjprescott, line 5: Why do you have 1.0.0b3 here when 1.0.0b2 says it is unreleased?

tjprescott (changes requested):

sdk/remoterendering/azure-mixedreality-remoterendering/CHANGELOG.md
MichaelZp0: As far as I know 1.0.0b2 was released, but simply no date was given (at least I found the PR into main with the canges from that release). I left it as unreleased as I was not sure, whether I should retroactively change it to the 'correct' date or whether I should just leave it.

I can dig up the date again and change it to the actual release date.

tjprescott (approved):

sdk/remoterendering/azure-mixedreality-remoterendering/CHANGELOG.md
MichaelZp0: Timestamps should now be fixed.

PR #25074, opened 2022-07-05
[Core] Case Insensitive Dictionary Implementation

kashifkhan:
Old PR 24824 had gotten a bit messy, so brought in all changes and other updates into this PR.


iscai-msft (approved):
don't have any comments, looks good to me!

PR #25084, opened 2022-07-05
[Storage] Revert removing aiohttp dependency for storage.blob.aio

vincenttran-msft:
Reverting the changes made in #24965 (sorry @Stevenjin8) as adequate discussion and investigation has not been done on this change before the upcoming GA release. 

In the future, if/when this work gets put back in, simply revert this PR (and probably play with the changelog merge conflicts that will arise)


lmazuel (approved):

Stevenjin8:
@vincenttran-msft @jalauzon-msft No worries! This isn't something that is very blocking, so I might not look into if for a couple weeks.

PR #24892, opened 2022-06-16
[ServiceBus] Adding in support for client identifier

kashifkhan:
closes #24486 

The identifier is used by the server for error logging etc. and comes in to play when the host specifies a client id that it can keep track of as it can help with correlation of error to client that opened the link. The identifier does not make its way back when receiving the message.

PR #25109, opened 2022-07-06
[ServiceBus]

kashifkhan:
Fixes for comments from previous PR #24892


annatisch (approved):

sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_receiver.py
```diff
@@ -126,7 +126,7 @@ class ServiceBusReceiver(
      The default value is 0, meaning messages will be received from the service and processed one at a time.
      In the case of prefetch_count being 0, `ServiceBusReceiver.receive` would try to cache `max_message_count`
      (if provided) within its request to the service.
```
swathipil, line 128: just checking that the changes in these file don't need to be made to async?

swathipil (approved):

sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_receiver.py
kashifkhan: No there was a spacing issue in the sync file that wasnt there in async. Thats why the change is only here 😊

PR #25157, opened 2022-07-10
[AzureML] Support async blob storage

annatisch:


sdk/ml/azure-ai-ml/azure/ai/ml/_artifacts/_artifact_utilities.py
```diff
@@ -40,8 +36,6 @@
 from azure.ai.ml._utils._arm_id_utils import AMLNamedArmId
 from azure.ai.ml.constants import SHORT_URI_FORMAT, STORAGE_ACCOUNT_URLS
 from azure.ai.ml.entities._datastore._constants import WORKSPACE_BLOB_STORE
-from azure.ai.ml._artifacts._blob_storage_helper import BlobStorageClient
-from azure.ai.ml._artifacts._gen2_storage_helper import Gen2StorageClient
```
annatisch, line 44: I've removed any imports from this module that would require the presence of any of the Storage dependencies - this means that only the Storage type that's needed will be loaded - and only _when_ it needed.
This means that the non-default Storage providers (i.e. datalake and fileshare) can be removed from the dependencies.

sdk/ml/azure-ai-ml/azure/ai/ml/_artifacts/_default_storage_helper.py
```diff
@@ -1,343 +0,0 @@
-# ---------------------------------------------------------
```
annatisch, line 1: I couldn't see where/how this file was being used - so I removed it. Happy to add it back if my housekeeping was a bit over-enthusiastic :)

sdk/ml/azure-ai-ml/azure/ai/ml/_artifacts/_fileshare_storage_helper.py
```diff
@@ -318,3 +322,11 @@ def recursive_download(
...
             error_category=ErrorCategory.SYSTEM_ERROR,
         )
+
+    def generate_sas(self, **kwargs):
+        """Not supported by file share."""
+        raise NotImplementedError("SAS generation not supported by FileShare storage.")
+
+    def update_metadata(self, *args) -> None:
+        """Not supported by file share"""
+        raise NotImplementedError("Metadata not supported by FileShare storage.")
```
annatisch, line 332: I've started building a unified API between the different Storage-type wrapper objects.
This would allow us to remove all the Storage-type validation throughout the code and just have the wrapper implement  the required features or raise like here where it's not supported.
(These `NotImeplementedError` could be swapped out for your validation error for example.)

sdk/ml/azure-ai-ml/azure/ai/ml/_utils/_storage_utils.py
```diff
@@ -120,15 +120,32 @@ def get_uri_type(self) -> str:
 
 
 def get_storage_client(
-    credential: str,
-    storage_account: str,
+    *,
+    credential: Optional[str] = None,
+    storage_account: Optional[str] = None,
+    storage_client: Optional[Any] = None,  # TODO: support user passing in own storage client.
```
annatisch, line 126: We will need to allow the option of passing in a pre-configured/pre-authenticated Storage client.
If you let me know the public API areas where users are expected to configure their storage setup - I can take a shot at plumbing this through more extensively.

sdk/ml/azure-ai-ml/azure/ai/ml/_utils/_storage_utils.py
```diff
@@ -140,6 +157,8 @@ def get_storage_client(
         account_url = STORAGE_ACCOUNT_URLS[storage_type].format(storage_account, storage_endpoint)
 
     if storage_type == DatastoreType.AZURE_BLOB:
+        if use_async:
+            return AsyncBlobStorageClient(credential=credential, container_name=container_name, account_url=account_url)
```
annatisch, line 161: I've only made the changes for Blob so far - so the same would need to be done for DataLake etc.

PR #25159, opened 2022-07-11
Issue #24273 add async test for offline test compress compressed no header offline

marcusc2:
This PR will:
- mark async test_compress_compressed_no_header as live test only


xiangyan99 (approved):

xiangyan99:
@marcusc2 Thanks for your contribution!

Issue #25064, opened 2022-07-04
Key Vault keys/secrets/certificates Samples names improvement


v-xuto:
**Issue Description:**
When implementing automated tests for key vault samples, the samples keys/secrets/certificates names in `sync` and `async samples` are the same, and the error is reported as:
![image](https://user-images.githubusercontent.com/57166602/177125630-204126a9-7fb1-46db-b88a-0e6bac7e2c74.png)

**Suggestion:**
Add `async` to the names of keys/secrets/certificates in the `async samples` to distinguish the names, 
which is convenient for automatic testing of samples. For examples: 
- `helloWorldSecretName` in **sync samples**.
- `helloWorldSecretNameAsync` in **async samples**.

@lmazuel , @hector-norza , @heaths , @schaabs for notification.

PR #25161, opened 2022-07-11
Key Vault keys/secrets/certificates Samples names improvement

v-xuto:
Fixes https://github.com/Azure/azure-sdk-for-python/issues/25064

@lmazuel , @heaths , @schaabs for notification.


mccoyp (approved):
Thank you for updating these!

Issue #24661, opened 2022-06-01
Typing errors in blob storage documentation code


TobiRoby:
Hi,

I am receiving typing errors from `mypy` when following the `blob storage` examples in the [official documentation](https://docs.microsoft.com/en-us/azure/storage/blobs/storage-quickstart-blobs-python?tabs=environment-variable-windows).

It would be awesome, if the documentation can provide an example with no type errors :)

This is the code I am type checking (copy & pasted from the linked documentation):

```python
import os
import uuid

from azure.storage.blob import BlobServiceClient

# Create the BlobServiceClient object which will be used to create a container client
blob_service_client: BlobServiceClient = BlobServiceClient.from_connection_string(
    "connect_str"
)
# Create a unique name for the container
container_name = str(uuid.uuid4())

# Create a local directory to hold blob data
local_path = "./data"
os.mkdir(local_path)

# Create a file in the local data directory to upload and download
local_file_name = str(uuid.uuid4()) + ".txt"
upload_file_path = os.path.join(local_path, local_file_name)

# Write text to the file
file = open(upload_file_path, "w")
file.write("Hello, World!")
file.close()

# Create a blob client using the local file name as the name for the blob
blob_client = blob_service_client.get_blob_client(
    container=container_name, blob=local_file_name
)

print("\nUploading to Azure Storage as blob:\n\t" + local_file_name)

# Upload the created file
with open(upload_file_path, "rb") as data:
    blob_client.upload_blob(data)

download_file_path = os.path.join(
    local_path, str.replace(local_file_name, ".txt", "DOWNLOAD.txt")
)
print("\nDownloading blob to \n\t" + download_file_path)

with open(download_file_path, "wb") as download_file:
    download_file.write(blob_client.download_blob().readall())
```

I receive 2 errors when applying `mypy`:

``` bash
test.py:35: error: Argument 1 to "upload_blob" of "BlobClient" has incompatible type "BufferedReader"; expected "Iterable[str]"
test.py:35: note: Following member(s) of "BufferedReader" have conflicts:
test.py:35: note:     Expected:
test.py:35: note:         def __iter__(self) -> Iterator[str]
test.py:35: note:     Got:
test.py:35: note:         def __iter__(self) -> Iterator[bytes]
[...]

PR #25173, opened 2022-07-11
[Storage] Fix upload data type hint

jalauzon-msft:
This change addresses part of #24661. The second part will come later.

There is a MyPy bug that has existed since 2017 where a `TypeVar` with multiple types inside a `Union` can lead to typing errors. See Issue for more details. Since this has not been fixed in quite some time, making this change to expand the `AnyStr` in the type hint for the data parameter to our blob/datalake upload APIs. I verified this does address the type error found in the Issue.


vincenttran-msft (approved):
LGTM! Glad that this is able to (somewhat) resolve the issue :)

PR #25176, opened 2022-07-12
[Tables] Add copy data between table and blob samples

YalinLi0312:
Resolves: https://github.com/Azure/azure-sdk-for-python/issues/19974
* Added table copy samples
* Checked and fixed typos and bugs in the rest of files under samples


sdk/tables/azure-data-tables/samples/async_samples/sample_copy_client_async.py
```diff
@@ -0,0 +1,108 @@
...
+    def __init__(self):
+        load_dotenv(find_dotenv())
+        self.access_key = os.getenv("TABLES_PRIMARY_STORAGE_ACCOUNT_KEY")
+        self.endpoint_suffix = os.getenv("TABLES_STORAGE_ENDPOINT_SUFFIX")
+        self.account_name = os.getenv("TABLES_STORAGE_ACCOUNT_NAME")
+        self.endpoint = "{}.table.{}".format(self.account_name, self.endpoint_suffix)
+        self.table_connection_string = "DefaultEndpointsProtocol=https;AccountName={};AccountKey={};EndpointSuffix={}".format(
+            self.account_name, self.access_key, self.endpoint_suffix
+        )
+        self.blob_connection_string = os.getenv("AZURE_STORAGE_CONNECTION_STRING")
```
annatisch, line 45: Maybe it's possible to use the same storage connection string for both Tables and Blobs?

sdk/tables/azure-data-tables/samples/async_samples/sample_copy_client_async.py
```diff
@@ -0,0 +1,108 @@
...
+        blob_service = BlobServiceClient.from_connection_string(self.blob_connection_string)
+        self.container = blob_service.get_container_client("mycontainer")
+        self.blob = BlobClient.from_connection_string(conn_str=self.blob_connection_string, container_name="mycontainer", blob_name="myblob")
+
+
+    async def copy_table_from_table_to_blob(self):
+        try:
+            # Upload entities to a table
+            await self.table.create_table()
+            await self.table.create_entity(self.entity)
```
annatisch, line 67: Maybe we could move the create/population of the initial table into a separate method called something like "setup"
```python
def setup_table(self):
    await self.table.create_table()
    for i in range(10):
        self.entity["RowKey"] = str(i)
        await self.table.create_entity(self.entity)
```

sdk/tables/azure-data-tables/samples/async_samples/sample_copy_client_async.py
```diff
@@ -0,0 +1,108 @@
...
+            # Download entities from table to memory
+            entities = self.table.list_entities()
+
+            self.container.create_container()
```
annatisch, line 71: What's the container called? Maybe we should create the container from the name of the table?

sdk/tables/azure-data-tables/samples/async_samples/sample_copy_client_async.py
```diff
@@ -0,0 +1,108 @@
...
+            # Upload in-memory table data to a blob that stays in a container
+            async for e in entities:
+                await self.blob.upload_blob(json.dumps(e))
```
annatisch, line 74: It would be good to test this with something like timestamps in the data as I'm not sure `json.dumps` will handle it.

sdk/tables/azure-data-tables/samples/async_samples/sample_copy_client_async.py
```diff
@@ -0,0 +1,108 @@
...
+        finally:
+            await self._tear_down()
+
+    async def copy_table_from_blob_to_table(self):
+        try:
+            self.container.create_container()
+            # Upload entities to a blob that stays in a container
+            await self.blob.upload_blob(json.dumps(self.entity))
```
annatisch, line 82: We shouldn't need to create the container/blob again. If we run these two methods back-to-back, that data should already be uploaded.

sdk/tables/azure-data-tables/samples/async_samples/sample_copy_client_async.py
```diff
@@ -0,0 +1,108 @@
...
+            # Download entities from blob to memory
+            # Note: when entities size is too big, may need to do copy by chunk
+            download_stream = await self.blob.download_blob()
```
annatisch, line 85: Don't we need loop through each blob in the container?

sdk/tables/azure-data-tables/samples/async_samples/sample_copy_client_async.py
```diff
@@ -0,0 +1,108 @@
...
+            entities = json.load(download_stream.readall())
+            # Upload entities to a table
+            await self.table.create_table()
```
annatisch, line 88: We should create the table based on the container name?

sdk/tables/azure-data-tables/samples/async_samples/sample_copy_client_async.py
YalinLi0312: I don't think so, for example, it's using different connection strings when running set-up resource scripts.

sdk/tables/azure-data-tables/samples/async_samples/sample_copy_client_async.py
YalinLi0312: The container name is passing through get_container_client() call by a BlobServiceClient client in line 59.

sdk/tables/azure-data-tables/samples/async_samples/sample_copy_client_async.py
YalinLi0312: I added a tear_down() to delete table, blob and container that created in each test to make them looks independent.

sdk/tables/azure-data-tables/samples/async_samples/sample_copy_client_async.py
YalinLi0312: No, the table name is given while creating `table` in line 57.

sdk/tables/azure-data-tables/samples/async_samples/sample_copy_client_async.py
YalinLi0312: I think we can integrate create_table() and create_container() into a initial table function if we want, the create_entity() should not be included because the other one is using upsert_entity(). So I think it's pretty clear(at least for me) now, not necessary to create a separate function. Also, I accept having a setup function if you still recommend it.

sdk/tables/azure-data-tables/samples/async_samples/sample_copy_client_async.py
YalinLi0312: I just created one blob client in the container `mycontainer` with blob name `myblob`, so I think it's not necessary to loop.

sdk/tables/azure-data-tables/samples/async_samples/sample_copy_client_async.py
YalinLi0312: There's a [key-value pair](https://github.com/Azure/azure-sdk-for-python/blob/1cb62159c9676c52d0e3dc154cce1b4c18c62def/sdk/tables/azure-data-tables/samples/async_samples/sample_copy_client_async.py#L52) about `last_updated` time in _entity_, is that what you expect to have?

YalinLi0312:
Updates since the last one:
1. Added setup functions for tables and blobs, btw, the two tests are still independent.
2. Added 10 entities/blobs to a table/container instead of one
3. Checked other files under samples, addressed some typos/imports/missing tests etc.

sdk/tables/azure-data-tables/samples/async_samples/sample_copy_client_async.py
annatisch: No, I mean if I use a datetime in my entity:
```python
entity = {
    'PartitionKey': 'foo',
    'RowKey': 'bar',
    'CreatedTime': datetime.now()
}
```

sdk/tables/azure-data-tables/samples/async_samples/sample_update_upsert_merge_entities_async.py
```diff
@@ -186,24 +186,12 @@ async def update_entities(self):
...
-
-        print("Cleaned up")
-
 
 async def main():
     sample = TableEntitySamples()
     await sample.create_and_get_entities()
     await sample.list_all_entities()
     await sample.update_entities()
-    await sample.clean_up()
```
annatisch, line 206: Why did we remove the cleanup?

sdk/tables/azure-data-tables/samples/sample_update_upsert_merge_entities.py
```diff
@@ -109,17 +110,18 @@ def list_all_entities(self):
                 # [START list_entities]
                 # Query the entities in the table
                 entities = list(table.list_entities())
-
+                i = 0
                 for i, entity in enumerate(entities):
                     print("Entity #{}: {}".format(entity, i))
+                    i += 1
```
annatisch, line 116: `i` is being set in the for loop on line 114 - why is it being incremented here?

sdk/tables/azure-data-tables/samples/sample_copy_table.py
```diff
@@ -0,0 +1,110 @@
...
+class CopyTableSamples(object):
+    def __init__(self):
+        load_dotenv(find_dotenv())
+        self.access_key = os.getenv("TABLES_PRIMARY_STORAGE_ACCOUNT_KEY")
+        self.endpoint_suffix = os.getenv("TABLES_STORAGE_ENDPOINT_SUFFIX")
+        self.account_name = os.getenv("TABLES_STORAGE_ACCOUNT_NAME")
+        self.table_connection_string = "DefaultEndpointsProtocol=https;AccountName={};AccountKey={};EndpointSuffix={}".format(
+            self.account_name, self.access_key, self.endpoint_suffix
+        )
+        self.table_client = TableClient.from_connection_string(conn_str=self.table_connection_string, table_name="mytable")
```
annatisch, line 44: I would just store the TableServiceClient - then use it to create the TableClient in each of the sample methods below. That way we can recreate the TableClient using the name of the container.

sdk/tables/azure-data-tables/samples/sample_copy_table.py
```diff
@@ -0,0 +1,110 @@
...
+        finally:
+            self._tear_down()
+
+    def copy_table_from_blob_to_table(self):
+        self._setup_blob()
+        try:
+            # Download entities from blob to memory
+            # Note: when entities size is too big, may need to do copy by chunk
+            blob_content = self.blob_client.download_blob().readall()
+            entities = json.load(blob_content)
```
annatisch, line 80: The loop in lines 68-70 are creating a unique blob for each entity.
This method is assuming all the entities are in a single blob.
The two methods must be consistent. Either we merge all the entities into one blob before uploading, or we download every blob and to recreate each entity.

sdk/tables/azure-data-tables/samples/async_samples/sample_copy_client_async.py
annatisch: It is necessary because we need to demonstrate how to create each entity.

sdk/tables/azure-data-tables/samples/async_samples/sample_copy_client_async.py
annatisch: Let's not call the container `mycontainer` and the blob `myblob`. That's not particularly useful.
I would say use the name of the table to create the container name, and name each blob in the container as the partitionkey + rowkey.

Then when we rebuild the table, use the container name to create the TableClient/table.

sdk/tables/azure-data-tables/samples/async_samples/sample_copy_client_async.py
annatisch: It should be possible, as most Storage accounts support both blobs and tables in the same account. However we can keep them separate in case someone wants to combine Storage blobs with Cosmos tables etc

sdk/tables/azure-data-tables/samples/async_samples/sample_copy_client_async.py
annatisch: Yeah but in the real scenario we will want to recreate the table from the blob backup - and we may not have the name in advance.

sdk/tables/azure-data-tables/samples/async_samples/sample_copy_client_async.py
YalinLi0312: It's saying `datetime` is not json serializable in my test, let me explore a way to make the deserialization happen.

sdk/tables/azure-data-tables/samples/sample_update_upsert_merge_entities.py
YalinLi0312: You're right, it is not expected to exist.

sdk/tables/azure-data-tables/samples/async_samples/sample_update_upsert_merge_entities_async.py
YalinLi0312: As I see it call delete_table() in the end of each test, so there's no necessary to have another clean_up() after.

sdk/tables/azure-data-tables/samples/async_samples/sample_copy_client_async.py
YalinLi0312: After another exploration, I find so far there have 3 types of objects that are not JSON serializable: `datetime`, `uuid`, `bytes`. I'm going to transfer these types values to `string` when uploading to **blob**, then will _transfer them back_ to their original types when downloading and uploading to **tables**.

YalinLi0312:
Main updates:
datetime, uuid and bytes types are not JSON serializable, to make entity object serializable to JSON string when uploading to blob, I stored iso format of datetime, uuid.hex, decoded bytes as strings in blob. They will transfer to their original types when downloading.

sdk/tables/azure-data-tables/samples/async_samples/sample_copy_table_async.py
```diff
@@ -0,0 +1,136 @@
...
+        }
+
+    async def copy_table_from_table_to_blob(self):
+        await self._setup_table()
+        try:
+            self.container_client = await self.blob_service_client.create_container(self.container_name)           
+            # Download entities from table to memory
+            entities = self.table_client.list_entities()
+            # Upload in-memory table data to a blob that stays in a container
+            for i in range(10):
```
annatisch, line 69: I think we would need to do `async for entity in self.table_client.list_entities()`?
We can't index into the list on line 70 until the entities have been iterated.

sdk/tables/azure-data-tables/samples/async_samples/sample_update_upsert_merge_entities_async.py
```diff
@@ -111,10 +111,8 @@ async def list_all_entities(self):
                 await table.create_entity(entity=entity1)
                 # [START list_entities]
                 # Query the entities in the table
-                i = 0
-                async for entity in table.list_entities():
+                async for i, entity in table.list_entities():
```
annatisch, line 114: `list_entities` doesn't return two outputs by default - I think this would need to be:
`async for i, entity in enumerate(table.list_entities())`
However I've never tried using `enumerate` with an async iterator before...

sdk/tables/azure-data-tables/samples/sample_copy_table.py
```diff
@@ -0,0 +1,129 @@
...
+
+    def copy_table_from_table_to_blob(self):
+        self._setup_table()
+        try:
+            self.container_client = self.blob_service_client.create_container(self.container_name)
+            # Download entities from table to memory
+            entities = self.table_client.list_entities()
+            # Upload in-memory table data to a blob that stays in a container
+            for i in range(10):
+                entity = entities[i]
```
annatisch, line 69: The entities need to be iterated first - so I think we need `for entity in client.list_entities()`

sdk/tables/azure-data-tables/samples/async_samples/sample_update_upsert_merge_entities_async.py
YalinLi0312: It failed with `TypeError: 'AsyncItemPaged' object is not iterable`, I'll define `i` outside.

Btw, the `list_entities()` is a **sync** function, but it use **async** keyword in most cases when iterating _listXXX()_ result, I'll ship a separate PR to address that.

annatisch: That's correct - `list_entities` is a sync function. The async behaviour is only in the iterator.
But you're right - I think `i` needs to be defined outside the loop. That seems to be a gap in Python that I hope they address some day...

sdk/tables/azure-data-tables/samples/async_samples/sample_update_upsert_merge_entities_async.py
```diff
@@ -111,8 +111,10 @@ async def list_all_entities(self):
                 await table.create_entity(entity=entity1)
                 # [START list_entities]
                 # Query the entities in the table
-                async for i, entity in table.list_entities():
+                i = 0
+                for entity in table.list_entities():
```
annatisch, line 115: This will still need to be `async for`

sdk/tables/azure-data-tables/samples/async_samples/sample_update_upsert_merge_entities_async.py
YalinLi0312: Ah, got it! Will add the `async` back.

annatisch (approved):

PR #25204, opened 2022-07-13
[Pyamqp] Thread safety for Connection, Links and Session

kashifkhan:
Adding in locks for `Connection`, `Link` and `Session`. This will come in to play when we decide to share these objects between multiple senders & receivers.

### DIFF
--- a/sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_link_async.py
+++ b/sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_link_async.py
@@ -101,6 +101,7 @@ class Link(object):
         self._received_payload = bytearray()
         self._on_link_state_change = kwargs.get('on_link_state_change')
         self._error = None
+        self._lock = asyncio.Lock()
 
     async def __aenter__(self):
         await self.attach()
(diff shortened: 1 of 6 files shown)


sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_link_async.py
```diff
@@ -115,108 +116,116 @@ def from_incoming_frame(cls, session, handle, frame):
...
+                incomplete_unsettled=self.incomplete_unsettled,
+                initial_delivery_count=self.initial_delivery_count if self.role == Role.Sender else None,
+                max_message_size=self.max_message_size,
+                offered_capabilities=self.offered_capabilities if self.state == LinkState.ATTACH_RCVD else None,
+                desired_capabilities=self.desired_capabilities if self.state == LinkState.DETACHED else None,
+                properties=self.properties
+            )
+            if self.network_trace:
+                _LOGGER.info("-> %r", attach_frame, extra=self.network_trace_params)
+            await self._session._outgoing_attach(attach_frame)
```
l0lawrence, line 192: Would this cause an issue b/c the lock for session and the lock for connection are 2 different locks?

PR #25205, opened 2022-07-13
[Storage] Fix some live tests

jalauzon-msft:
Fixing the recent test proxy test that was introduced which improperly used `variables` and was failing in the Live test pipeline.

Also sneaking a change in to further increase the wait to for soft delete to enable since we're still seeing soft delete tests fail occasionally in the nightly pipeline runs.


vincenttran-msft (approved):
Glad you were able to get this sorted out. Thanks!

PR #25207, opened 2022-07-13
Ability to specify tenant_id

tikicoder:
Updates the Azure CLI SDK To allow to create credential objects that are connected to tenants. This is EXTREMELY useful when you work with multiple tenants.

# Description
 This allows developers to specify what tenant the credentials should use. This way if you have access to multiple you do not need to worry about switching other ways.

I created the following wrapper class locally and it has fixed my issue. No reason I should have to be that hacky.

```
from azure.identity import AzureCliCredential as main_AzureCliCredential

class AzureCliCredential(main_AzureCliCredential):
    def __init__(self, tenant_id = None):
        main_AzureCliCredential.__init__(self)

        self.tenant_id = tenant_id

    def get_token(self, *scopes: str, **kwargs):
        if self.tenant_id is not None:
            if kwargs is None:
                kwargs = {}
            kwargs["tenant_id"] = self.tenant_id
            return main_AzureCliCredential.get_token(self, *scopes, **kwargs)
        
        return main_AzureCliCredential.get_token(self, *scopes, **kwargs)
```

### DIFF
--- a/sdk/identity/azure-identity/azure/identity/_credentials/azure_cli.py
+++ b/sdk/identity/azure-identity/azure/identity/_credentials/azure_cli.py
@@ -36,7 +36,7 @@ class AzureCliCredential(object):
 
     This requires previously logging in to Azure via "az login", and will use the CLI's currently logged in identity.
     """
-    def __init__(self, tenant_id: str = None):
+    def __init__(self, tenant_id: str = ""):
         object.__init__(self)
 
         self.tenant_id = tenant_id
@@ -71,7 +71,7 @@ class AzureCliCredential(object):
 
         resource = _scopes_to_resource(*scopes)
         command = COMMAND_LINE.format(resource)
-        tenant = resolve_tenant(default_tenant= "", tenant_id= self.tenant_id, **kwargs)
+        tenant = resolve_tenant(default_tenant= self.tenant_id, **kwargs)
         
         if tenant:
             command += " --tenant " + tenant
--- a/sdk/identity/azure-identity/azure/identity/_credentials/azure_cli.py
+++ b/sdk/identity/azure-identity/azure/identity/_credentials/azure_cli.py
@@ -72,7 +72,7 @@ class AzureCliCredential(object):
         resource = _scopes_to_resource(*scopes)
         command = COMMAND_LINE.format(resource)
         tenant = resolve_tenant(default_tenant= self.tenant_id, **kwargs)
-        
+
         if tenant:
             command += " --tenant " + tenant
         output = _run_command(command)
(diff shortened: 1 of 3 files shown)


tikicoder:
> Hi @tikicoder, thank you for the PR and feedback! Some folks are out of office today, but we'll discuss your proposal and get back to you as soon as possible 🙂

Thanks

sdk/identity/azure-identity/azure/identity/_credentials/azure_cli.py
```diff
@@ -67,7 +71,8 @@ def get_token(self, *scopes, **kwargs): # pylint: disable=no-self-use
 
         resource = _scopes_to_resource(*scopes)
         command = COMMAND_LINE.format(resource)
-        tenant = resolve_tenant("", **kwargs)
+        tenant = resolve_tenant(default_tenant= "", **kwargs) if self.tenant_id is None else resolve_tenant(default_tenant= "", tenant_id= self.tenant_id, **kwargs)
```
xiangyan99, line 74: Given the definition of resolve_tenant is:

def resolve_tenant(default_tenant, tenant_id=None, **_):

why not just use 

tenant = resolve_tenant(default_tenant= "", tenant_id= self.tenant_id, **kwargs)

tikicoder: With the last update I did where I was explicit about the params there isn't a reason not to. It does make it similar to read and I have updated it.

xiangyan99 (commented):
Test fails with error:

TypeError: resolve_tenant() got multiple values for keyword argument 'tenant_id'

Seems like we need:

if 'tenant_id' not in kwargs:
tenant = resolve_tenant(default_tenant= "", tenant_id= self.tenant_id, **kwargs)
else:
tenant = resolve_tenant(default_tenant= "", **kwargs)

tikicoder:
> Test fails with error:
> 
> TypeError: resolve_tenant() got multiple values for keyword argument 'tenant_id'
> 
> Seems like we need:
> 
> if 'tenant_id' not in kwargs: tenant = resolve_tenant(default_tenant= "", tenant_id= self.tenant_id, **kwargs) else: tenant = resolve_tenant(default_tenant= "", **kwargs)

Ok I updated it so instead of passing in default_tenant of empty it now passes something in. In theory, I would hope the tenant_id that is being passed in is None. I could update to include it in the logging but that seems overkill. Either this will work or it won't. It would make sense also that if you pass in a tenant_id on the class level it will become the default.

tikicoder:
> 

That should be resolved now. I edited it in the github editor so looks like it added a tab that wasn't needed.

PR #25224, opened 2022-07-14
[formrecognizer] API review feedback pt1

catalinaperalta:
Update typing to show which attributes are optional and add `__str__` for CurrencyValue

### DIFF
--- a/sdk/formrecognizer/azure-ai-formrecognizer/azure/ai/formrecognizer/_models.py
+++ b/sdk/formrecognizer/azure-ai-formrecognizer/azure/ai/formrecognizer/_models.py
@@ -1972,7 +1972,7 @@ class TextAppearance:
 class BoundingRegion:
     """The bounding region corresponding to a page.
 
-    :ivar list[~azure.ai.formrecognizer.Point] polygon:
+    :ivar Optional[list[~azure.ai.formrecognizer.Point]] polygon:
         A list of points representing the bounding polygon
         that outlines the document component. The points are listed in
         clockwise order relative to the document component orientation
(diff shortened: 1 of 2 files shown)


sdk/formrecognizer/azure-ai-formrecognizer/azure/ai/formrecognizer/_models.py
```diff
@@ -2134,6 +2134,11 @@ def _from_generated(cls, data):
             symbol=data.currency_symbol,
         )
 
+    def __str__(self):
+        if self.symbol is not None:
+            return f"{self.symbol}{self.amount}"
```
kristapratico, line 2139: is this always like "$1.00" or can it be, for example, "USD1.00"?

sdk/formrecognizer/azure-ai-formrecognizer/azure/ai/formrecognizer/_models.py
```diff
@@ -2292,11 +2297,11 @@ class AnalyzedDocument:
     :ivar doc_type: The type of document that was analyzed.
     :vartype doc_type: str
     :ivar bounding_regions: Bounding regions covering the document.
-    :vartype bounding_regions: list[~azure.ai.formrecognizer.BoundingRegion]
+    :vartype bounding_regions: Optional[list[~azure.ai.formrecognizer.BoundingRegion]]
     :ivar spans: The location of the document in the reading order concatenated content.
     :vartype spans: list[~azure.ai.formrecognizer.DocumentSpan]
     :ivar fields: A dictionary of named field values.
-    :vartype fields: dict[str, ~azure.ai.formrecognizer.DocumentField]
+    :vartype fields: Optional[dict[str, ~azure.ai.formrecognizer.DocumentField]]
```
kristapratico, line 2304: kind of nit picky, but if we're talking about Optional in a typing sense, it means it can be the value or None. I think in more cases than not, we return an empty dict for fields, right? Seeing as this is just the docstring, I think it's probably fine, though.

kristapratico (approved):

sdk/formrecognizer/azure-ai-formrecognizer/azure/ai/formrecognizer/_models.py
catalinaperalta: It should only be the symbol as in "$"

sdk/formrecognizer/azure-ai-formrecognizer/azure/ai/formrecognizer/_models.py
catalinaperalta: I had the same thought when doing this. I'm going to bring this up to Anna and Johan to see about the correctness of returning the empty dict or list.

catalinaperalta: If a change is needed I'll add it in a future PR.

PR #25230, opened 2022-07-15
[formrecognizer] Make get words a property

catalinaperalta:
Make `get_words()` a property `words` on `DocumentLine`.


sdk/formrecognizer/azure-ai-formrecognizer/CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 - Renamed `AccountInfo` model to `ResourceInfo`.
 - Renamed `DocumentModelInfo` model to `DocumentModelSummary`.
 - Renamed `DocumentModel` to `DocumentModelInfo`.
+- Updated `get_words()` method on `DocumentLine` to `words` property.
```
kristapratico, line 14: code LGTM, but just wanted to check - I recall wanting to be able to pass options like whether words overlap or something like that? Is that not part of future plans anymore?

catalinaperalta: Thanks for the reminder! I'll bring this up with Johan and Anna.

catalinaperalta: After the arch board, we will keep `get_words()` as a method because it will need to have the options for searching for words.

PR #25239, opened 2022-07-15
[EventHubs] Fix mgmt request token format + logging

annatisch:


sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py
```diff
@@ -431,9 +435,12 @@ def close(self):
...
         """
+        _LOGGER.debug("Closing ProducerClient")
         with self._lock:
             for pid in self._producers:
                 if self._producers[pid]:
                     self._producers[pid].close()  # type: ignore
-                self._producers[pid] = None
+            self._producers = {
+                ALL_PARTITIONS: self._create_producer()
+            }
```
annatisch, line 445: While we should probably add this in future - I don't think we should add it to the debug build we send out with the extra logging. As - assuming they're not adding partition id/key to their batches - the problem will go away because we're simply not hitting the code path any longer. Which won't actually help us fix the problem.

kashifkhan: agree this should not be in the build with extra logging et al. Maybe we can add a bool `close_called`. This way if someone calls close and then resends we can use that to `create_producer`. 

If someone calls close and they are done with the class, they are still left with an instantiated producer.

PR #25249, opened 2022-07-15
Enable ML Sample Submission Directory off of PR

scbedd:
This PR contains

- [x] The additional script files
- [x] Job Definition

The intent is to provide a job that can be triggered via `/azp run python - ml-samples-submission`. Which will be an internally defined job.

This job will

- Build the ml artifact from the PR/branch it was triggered against
- Upload to a blob storage location
- Submit a PR to the `azureml-examples` repo to trigger their validation process

It will not be in `public`, and will NOT autotrigger on all builds.

### DIFF
--- a/scripts/devops_tasks/generate_ml_sample_update.py
+++ b/scripts/devops_tasks/generate_ml_sample_update.py
@@ -1,4 +1,7 @@
+# pip install -r eng/ml_sample_tools.txt before invoking
 import argparse
+import glob
+import os
 
 if __name__ == "__main__":
     parser = argparse.ArgumentParser(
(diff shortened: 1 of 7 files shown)


scripts/devops_tasks/generate_ml_sample_update.py
```diff
@@ -0,0 +1,97 @@
+# pip install -r eng/ml_sample_tools.txt before invoking
+import argparse
+import glob
+import os
+import pdb
```
scbedd, line 5:
```suggestion
```

benbp (approved):

PR #25258, opened 2022-07-15
[Test Proxy] Align sanitizers, matchers, and transforms

mccoyp:
# Description

This PR initially was just going to be a fix to replace `add_request_subscription_id_sanitizer` with `add_uri_subscription_id_sanitizer` -- the former sanitizer either never existed, or was replaced with the latter sanitizer without me noticing. (In either case, my bad!)

After fixing this sanitizer though, I noticed that multiple sanitizers had been given a `condition` option since the time I added Python support for them. This PR adds that support where applicable.

Finally, this adds support for any other matchers, sanitizers, and transforms that are supported by the test proxy but haven't had Python support.

  - Changes verified to work locally with SDK tests.


tools/azure-sdk-tools/devtools_testutils/sanitizers.py
```diff
@@ -283,4 +475,25 @@ def _send_sanitizer_request(sanitizer: str, parameters: dict) -> None:
         headers={"x-abstraction-identifier": sanitizer, "Content-Type": "application/json"},
         json=parameters
     )
+    breakpoint()
```
scbedd, line 478: ?

tools/azure-sdk-tools/devtools_testutils/sanitizers.py
```diff
@@ -283,4 +475,25 @@ def _send_sanitizer_request(sanitizer: str, parameters: dict) -> None:
...
+    :param str transform: The name of the transform to add.
+    :param dict parameters: The transform constructor parameters, as a dictionary.
+    """
+
+    if is_live():
+        return
+
+    response = requests.post(
+        f"{PROXY_URL}/Admin/AddTransform",
+        headers={"x-abstraction-identifier": transform, "Content-Type": "application/json"},
```
scbedd, line 496: So this one _looks_ to me to only add `session` level sanitizer. However, further up in `set_reset_request` we actually always send for a recordingId (unless it gets called and the fixture can't find the active recordingId).

Can we have a quick chat about the PR?

scbedd (changes requested):
I'd like a quick offline conversation to discuss session vs individual recording settings.

scbedd (approved):

PR #25265, opened 2022-07-18
[Pyamqp] Client Class Clean up

kashifkhan:
Changes to the client class to ensure consistency with names, doctrings and porting over methods from uamqp

### DIFF
--- a/sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/client.py
+++ b/sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/client.py
@@ -125,7 +125,7 @@ class AMQPClient(object):
     def __init__(self, remote_address, auth=None, client_name=None, debug=False, retry_policy=None, 
     keep_alive_interval=None, **kwargs):
         self._remote_address = remote_address
-        self._auth = auth if auth else SASLAnonymousCredential()
+        self._auth = auth
         self._name = client_name if client_name else str(uuid.uuid4())
 
         self._shutdown = False
--- a/sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/client.py
+++ b/sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/client.py
@@ -19,7 +19,7 @@ from .session import Session
 from .sender import SenderLink
 from .receiver import ReceiverLink
 from .sasl import SASLAnonymousCredential, SASLTransport
-from .endpoints import Source, Target
+from .endpoints import Sopurce, Source, Target
 from .error import (
     AMQPConnectionError,
     AMQPException,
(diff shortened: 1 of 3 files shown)


sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/client.py
```diff
@@ -234,15 +239,15 @@ def open(self):
 
         :param connection: An existing Connection that may be shared between
          multiple clients.
-        :type connetion: ~uamqp.Connection
+        :type connetion: ~pyamqp.Connection
```
l0lawrence, line 242: connection spelled wrong

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/client.py
```diff
@@ -59,14 +59,14 @@ class AMQPClient(object):
...
     :param remote_address: The AMQP endpoint to connect to. This could be a send target
      or a receive source.
-    :type remote_address: str, bytes or ~uamqp.address.Address
-    :param auth: Authentication for the connection. This should be one of the subclasses of
-     uamqp.authentication.AMQPAuth. Currently this includes:
-        - uamqp.authentication.SASLAnonymous
-        - uamqp.authentication.SASLPlain
-        - uamqp.authentication.SASTokenAuth
+    :type remote_address: str, bytes or ~pyamqp.endpoint.Target or ~pyamqp.endpoint.Source
+    :param auth: Authentication for the connection. This should be one of the following:
```
l0lawrence, line 63: (self-note for my brain) Okay so we instantiate an AMQP Client only in pyamqp code or in basehandler -- we use Receive and Send in the producer/consumer. Need to check that when we hit the base_handler this gets passed through to the send/recieve client correctly so we can drop hostname in the front-end

sdk/eventhub/azure-eventhub/tests/pyamqp_tests/unittest/test_exceptions_pyamqp.py
```diff
@@ -12,7 +12,7 @@ def test_client_creation_exceptions():
         sender = SendClient(
             "fake.host.com",
         )
-        assert sender._hostname == "fake.host.com"
+        assert sender._remote_address == "fake.host.com"
```
l0lawrence, line 15: (fix) like the tests below -- target I think - or remove assert? We might not need this test anymore - I will play around with it

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/client.py
```diff
@@ -18,7 +18,7 @@
 from .session import Session
```
l0lawrence, line 18: Do we need the same on async? - will look

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/client.py
```diff
@@ -105,27 +105,28 @@ class AMQPClient(object):
...
     :param encoding: The encoding to use for parameters supplied as strings.
      Default is 'UTF-8'
     :type encoding: str
     """
 
-    def __init__(self, hostname, auth=None, **kwargs):
-        self._hostname = hostname
+    def __init__(self, remote_address, auth=None, client_name=None, debug=False, retry_policy=None, 
+    keep_alive_interval=None, **kwargs):
+        self._remote_address = remote_address.address if (isinstance(remote_address, Source) or isinstance(remote_address, Target)) else remote_address
```
l0lawrence, line 127: we format hostname in the EH SDK side - will need to do something similiar in pyamqp to remote_address otherwise wont be able to connect to socket

PR #25268, opened 2022-07-18
[formrecognizer] Add a type for target authorization for copy requests

catalinaperalta:
Address architect feedback to add a type for target authorization for copy request. This will give us some type checking when passing the target in to the copy request method.


kristapratico (approved):
yay NewType!

PR #25270, opened 2022-07-18
[Key Vault] Skip tests when vault SKU is incompatible

mccoyp:
# Description

I recently got our weekly sovereign cloud tests running again, and [test runs show](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=1713557&view=ms.vss-test-web.build-test-results-tab&runId=33466850&resultId=100365&paneView=debug) that some tests -- namely, those that deal with key release -- require a premium vault SKU in order to work. Standard SKUs are used in China cloud tests, so we need to skip these when necessary.

### DIFF
--- a/sdk/keyvault/azure-keyvault-keys/tests/test_keys_async.py
+++ b/sdk/keyvault/azure-keyvault-keys/tests/test_keys_async.py
@@ -8,6 +8,7 @@ from dateutil import parser as date_parse
 import functools
 import json
 import logging
+import os
 
 from azure.core.exceptions import HttpResponseError, ResourceExistsError, ResourceNotFoundError
 from azure.core.pipeline.policies import SansIOHTTPPolicy
@@ -515,6 +516,9 @@ class TestKeyVaultKey(KeyVaultTestCase, KeysTestCase):
     @AsyncKeysClientPreparer()
     @recorded_by_proxy_async
     async def test_key_release(self, client, **kwargs):
+        if (self.is_live and os.environ["KEYVAULT_SKU"] != "premium"):
+            pytest.skip("This test is not supported on standard SKU vaults. Follow up with service team")
+
         set_bodiless_matcher()
         attestation_uri = self._get_attestation_uri()
         attestation = await get_attestation_token(attestation_uri)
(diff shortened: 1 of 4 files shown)


sdk/keyvault/azure-keyvault-keys/tests/test_samples_keys.py
```diff
@@ -42,7 +42,7 @@ class TestExamplesKeyVault(KeyVaultTestCase, KeysTestCase):
     @recorded_by_proxy
     def test_example_key_crud_operations(self, key_client, **kwargs):
         if (self.is_live and os.environ["KEYVAULT_SKU"] != "premium"):
-            pytest.skip("This test not supprot in usgov/china region. Follow up with service team")
+            pytest.skip("This test is not supported on standard SKU vaults. Follow up with service team")
```
jeesang7, line 45:
```suggestion
            pytest.skip("This test is not supported in usgov/china region. Follow up with service team")
```

sdk/keyvault/azure-keyvault-keys/tests/test_samples_keys_async.py
```diff
@@ -46,7 +46,7 @@ class TestExamplesKeyVault(KeyVaultTestCase):
     @recorded_by_proxy_async
     async def test_example_key_crud_operations(self, key_client, **kwargs):
         if (self.is_live and os.environ["KEYVAULT_SKU"] != "premium"):
-            pytest.skip("This test not supprot in usgov/china region. Follow up with service team")
+            pytest.skip("This test is not supported on standard SKU vaults. Follow up with service team")
```
jeesang7, line 49: How about keeping some messages before the change?
```suggestion
            pytest.skip("This test is not supported in usgov/china region. Follow up with service team")
```

sdk/keyvault/azure-keyvault-keys/tests/test_keys_async.py
```diff
@@ -637,7 +647,7 @@ async def test_immutable_release_policy(self, client, **kwargs):
     async def test_key_rotation(self, client, **kwargs):
         set_bodiless_matcher()
         if (not is_public_cloud() and self.is_live):
-            pytest.skip("This test not supprot in usgov/china region. Follow up with service team.")
+            pytest.skip("This test is not supported in usgov/china region. Follow up with service team.")
```
YalinLi0312, line 650: Why not put the checks in the first step, before `set_bodiless_matcher()`?

sdk/keyvault/azure-keyvault-keys/tests/test_samples_keys.py
mccoyp: My thinking was that the check here is specifically for standard SKU vaults, rather than the cloud. China cloud tests do only use standard SKUs, so the message is technically accurate, but the incompatibility seems to come more from the SKU than from the region.

What do you think?

YalinLi0312 (approved):

PR #25272, opened 2022-07-18
[Storage] Migrate `test_blob_api_version` and `test_blob_client` to test proxy

jalauzon-msft:


sdk/storage/azure-storage-blob/tests/test_blob_client.py
```diff
@@ -3,78 +3,94 @@
...
+            permission=AccountSasPermissions(read=True, list=True),
+            start=datetime.now() - timedelta(hours=24),
+            expiry=datetime.now() + timedelta(days=8),
+        )
 
     # --Direct Parameters Test Cases --------------------------------------------
     @BlobPreparer()
-    def test_create_service_with_key(self, storage_account_name, storage_account_key):
-        # Arrange
+    def test_create_service_with_key(self, **kwargs):
```
vincenttran-msft, line 61: Are these live-only? I thought these would need a `@recorded_by_proxy` 👀

vincenttran-msft (approved):
LGTM! Two sets of files down, nice! Still curious as to why some of these don't have `@recorded_by_proxy`, but the tests pass, so all is good?

sdk/storage/azure-storage-blob/tests/test_blob_client.py
jalauzon-msft: Any test here that does not have `recorded_by_proxy` is because that test makes no network requests, so it doesn't need to be recorded. These files have quite a few instances of that where they are just testing client construction and such. They are more like unit tests.

PR #25288, opened 2022-07-19
bump python version in autorest pipeline check

iscai-msft:
Since we're dropping suport for python 3.6, bumping the version of python we use to run autorest so we don't have python version compatibility issues


scbedd (approved):
I'm not certain I'm the person who really knows though.

PR #25324, opened 2022-07-20
add needs-triage only if it is not assigned

xiangyan99:


.github/workflows/label-needs-triage.yml
```diff
@@ -16,6 +16,7 @@ jobs:
     steps:
       - name: Label issues
         uses: andymckay/labeler@e6c4322d0397f3240f0e7e30a33b5c5df2d39e90
+        if: "${{ github.event.issue.assignees == '[]' }}"
```
jsquire, line 19: Can we do the same check for labels?

jsquire (approved):

.github/workflows/label-needs-triage.yml
xiangyan99: Have not figured out how to do it yet. :(

jsquire: The format of the above would lead me to speculate that the following **_may_** work:

```yaml
if: "${{ github.event.issue.labels == '[]' }}"
```

xiangyan99: From my observation, it is always [] and we need to get the detail labels from the url:

labels: [],
labels_url: https://api.github.com/repos/xiangyan99/AzureDTL/issues/20/labels{/name},

PR #25331, opened 2022-07-20
fix proxy url in tox

iscai-msft:
The current default in tox is for a non-existent url. Trying to see if fixing this helps with running test proxy in the autorest pipelines


scbedd (approved):

mccoyp (approved):
LGTM! This variable doesn't seem to get used in most contexts (otherwise things would fail). This is affecting autorest pipelines though, which require an HTTP connection instead of HTTPS.

PR #25335, opened 2022-07-20
[Storage] Adjust the docstring for `include` in `list_blobs` to fix docs

jalauzon-msft:
When looking at the [`list_blob` docs](https://docs.microsoft.com/en-us/python/api/azure-storage-blob/azure.storage.blob.containerclient?view=azure-python#azure-storage-blob-containerclient-list-blobs) I saw the type for the `include` parameter is messed up. This was due to the docstring using `or` in the param declaration. This PR fixes that up so the docs will render correctly.


vincenttran-msft (approved):
LGTM! That pesky `or` case 😢

PR #25336, opened 2022-07-21
[Storage] Migrate test_blob_service_properties to test proxy

vincenttran-msft:
This PR converts `test_blob_service_properties.py` and ``test_blob_service_properties_async.py` to use test proxy!


jalauzon-msft (approved):
More great work! ~30 more tests down!

PR #25345, opened 2022-07-21
[formrecognizer] Arch board renames

### LINKED ISSUE #24980: [formrecognizer] Consistency issues
Look into:
- [x] checkout model naming in begin_analyze_document do we use model_id everywhere else? Should we unify?
- [x] rename begin_create_compose_model -> begin_compose_model - follow up
- [x] rename get_account_info -> get_resource_info
- [x] rename AccountInfo -> ResourceInfo
- [x] check the dict[str, str] returned from get_copy_authorization...other langs use a model should we have a typed dict?
- [x] rename documentmodelinfo to documentmodelsummary. also documentmodel to documentmodelinfo
- [x] GA blocker is the service going to support different input source types?
- [x] rename target to authorization on copy model - follow up

catalinaperalta:
Fixes #24980


sdk/formrecognizer/azure-ai-formrecognizer/CHANGELOG.md
```diff
@@ -7,10 +7,12 @@
 
 ### Breaking Changes
 - Renamed `begin_create_composed_model()` on `DocumentModelAdministrationClient` to `begin_compose_model()`.
-- Renamed `get_account_info()` on `DocumentModelAdministrationClient` to `get_resource_info()`.
-- Renamed `AccountInfo` model to `ResourceInfo`.
+- Renamed `get_account_info()` on `DocumentModelAdministrationClient` to `get_resource_details()`.
+- Renamed `AccountInfo` model to `ResourceDetails`.
```
kinelski, line 11: We have it as `ResourceDetails` too, but I think we may need to find a better name before GA (at least in .NET). Our architects complained about the name being too generic when we had it as `ResourceInfo`. `FormRecognizerResourceDetails` looks too long too.

kinelski (approved):
LGTM. .NET follows the same naming.

kristapratico (approved):
🚢

PR #25357, opened 2022-07-22
[Storage] Migrate `test_block_blob` to test proxy

jalauzon-msft:
In addition to migrating the tests, I also adjusted the chunk/blob sizes used throughout this test to decrease the size of the recordings. I tried to do this in a way that would not change the functionality being tested (still testing chunking when needed).


vincenttran-msft (approved):
LGTM! Glad you were able to resolve all those sanitizer and other fun issues like datetime 😛

Issue #25341, opened 2022-07-21
Opentelemetry span events have wrong ParentId in Azure Monitor logs


neerajanil:
- **Package Name**: azure-monitor-opentelemetry-exporter
- **Package Version**: 1.0.0b6 
- **Operating System**: Windows 11
- **Python Version**: 3.8.10

**Describe the bug**
The bug occurs when a span which has events added to it is exported to Azure Monitor through the  `AzureMonitorTraceExporter`.
The `ai.operation.parentId` tag of the Trace that is created for these events is set to `span.parent.span_id` instead of `span.context.span_id`. This causes the Trace to get associated with the span's parent instead of the span in Log Analytics.

**To Reproduce**
1. run the following code:
```python
import os

from azure.monitor.opentelemetry.exporter import AzureMonitorTraceExporter
from opentelemetry import trace
from opentelemetry.sdk.trace import TracerProvider
from opentelemetry.sdk.trace.export import BatchSpanProcessor

exporter = AzureMonitorTraceExporter.from_connection_string(
    "<Your Connection String>"
)
provider = TracerProvider()
tracer = trace.get_tracer(__name__)
span_processor = BatchSpanProcessor(exporter)
provider.add_span_processor(span_processor)
trace.set_tracer_provider(provider)


with tracer.start_as_current_span("foo") as span:
    span.add_event("bar")
    print("event added!")
    with tracer.start_as_current_span("quark") as span:
        span.add_event("boson")
        print("event added!")

provider.shutdown()

```
2) In Azure Monitor you can see that the 
    - trace "bar" has `Parent Id` set incorrectly to `Operation Id` instead of being set to `Id` of "foo"
    - trace "boson" has `Parent Id` set incorrectly to `Id` of "foo" instead of being set to `Id` of "quark"

**Expected behavior**
The Trace generated in Azure Monitor for a span event should have its `ai.operation.parentId` tag set to the `id` of the span in which it was added

**Additional context**
The root cause of the issue seems to be the code [here](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/monitor/azure-monitor-opentelemetry-exporter/azure/monitor/opentelemetry/exporter/export/trace/_exporter.py#L447-L450), which should be changed to :
```python
if span.context and span.context.span_id:
    envelope.tags["ai.operation.parentId"] = "{:016x}".format(
        span.context.span_id
    )
```

tjprescott:
[...]

PR #25369, opened 2022-07-22
Fix span event ai.operation.id to use span id instead of span parent id

lzchen:
Fixes https://github.com/Azure/azure-sdk-for-python/issues/25341#


jeremydvoss (approved):
Looks straightforward and good to me.

rakshith91 (approved):

PR #25371, opened 2022-07-22
[Cosmos] Add AutoScale Header Support 2/2

JericHunter:
Adding the ability to update the autoscale settings


JericHunter:
will be using a different PR for this

PR #25388, opened 2022-07-25
[Pylint] Bump version disabling checkers

l0lawrence:
disabling the difference between pylint --list-msgs-enabled for version 2.14.5 versus 2.9.3, until can take a closer look at what we want to keep

Diff: Total  Approx 40 differences
typevar-name-incorrect-variance  
typevar-double-variance
typevar-name-mismatch 
use-sequence-for-iteration
consider-using-f-string - yes there are quite a few
use-implicit-booleaness-not-len -
use-implicit-booleaness-not-comparison
non-ascii-module-import 
bad-file-encoding 
unnecessary-dunder-call 
unnecessary-lambda-assignment
unnecessary-direct-lambda-call 
bad-configuration-section
unrecognized-option
invalid-enum-extension 
potential-index-error
invalid-unicode-codec 
bidirectional-unicode 
invalid-character-backspace
invalid-character-carriage-return
invalid-character-sub 
invalid-character-esc 
invalid-character-nul
invalid-character-zero-width-space 
modified-iterating-dict
modified-iterating-set 
config-parse-error 
useless-option-value 
comparison-of-constants 
use-list-literal
use-dict-literal
unnecessary-list-index-lookup
unknown-option-value 
duplicate-value
overridden-final-method
subclassed-final-class 
redefined-slots-in-subclass
super-without-brackets
format-string-without-interpolation
redundant-u-string-prefix
unspecified-encoding 
forgotten-debug-statement
method-cache-max-size-none
useless-with-lock
unnecessary-ellipsis 
non-ascii-file-name
using-f-string-in-unsupported-versio 
using-final-decorator-in-unsupported-version
modified-iterating-list


tjprescott (changes requested):
Yikes that's a massive list! Is there any way to break it up so it's obvious what was "permanent" and what is "temporary"?

l0lawrence:
> Yikes that's a massive list! Is there any way to break it up so it's obvious what was "permanent" and what is "temporary"?

I can order them so that permanently/currently disabled checkers go first and then put a space before the temporarily/new version added disabled checkers are listed

tjprescott:
@l0lawrence how many violations are there of the new checks?

l0lawrence:
> @l0lawrence how many violations are there of the new checks?

It is running - I will update the violations above next to the check name ^

tjprescott (approved):
Let's work on trimming this list once we verify this fixes the Const is not iterable issue.

tjprescott:
@l0lawrence is this PR still relevant or should it be closed?

l0lawrence:
It is still relevant! I will be addressing pylint bumps during MQ

tjprescott:
@l0lawrence then should we resolve the conflicts and merge this or close it until MQ?

l0lawrence:
Close until MQ

Issue #23510, opened 2022-03-15
Send EventDataBatch with per-message partition_key


weltonrodrigo:
It would be nice to be able to send a batch where every message has it's own partition_key.

The case here is when I get a batch from an external client, but I need to repartition this batch. Send every message isolated it not ideal, so using a batch is essential.

But there is no public interface to set the partition key for a single EventData, only EventDataBatch. It used to exist, but was removed some time ago, judging for avaiable examples.

Why was it removed? Can we get it back?

Currently, I'm having to do:

```python
for message in messages:
  ev = EventData(json.dumps(message))
  ev._raw_amqp_message.annotations[b"x-opt-partition-key"] = self._custom_partitioner(message)  # hack
  event_data_batch.add(ev)
```

Where `_custom_partitioner` it's just a business-aware partitioner.

yunhaoling:
hey @weltonrodrigo , thanks for your feedback!

The `EventDataBatch` is designed to host a batch of events of the same partition_key or partition_id for high sending performance. Within the `EventHubProducerClient` there are producers for each partition id. events with no partition_id but with partition_key will be sent out by default producer.

I need to first check whether setting different partition keys of events within a batch would get events sent to the right partition.
If the service is working as we expected, then yes, we could consider adding the support.

yunhaoling:
thanks for confirmation.

this is expected by me actually because for a batch event, it's the partition id/partition key on the batch event that dominates.

Yes, so far you would need to create batch for each message (partition).

But we're working on a new feature called buffered producer in which you don't have to manage batch by yourselves and does automatically batching and sending in the background.

I think this coming feature should help resolve your issue!

swathipil:
Hi @weltonrodrigo - `azure-eventhub` v`5.10.0` has been released with the buffered producer feature. The sample for buffered sending is [[here](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/eventhub/azure-eventhub/samples/sync_samples/send_buffered_mode.py)].

Would you be able to test this to see if it resolves your issue? You can upgrade with `pip install azure-eventhub --upgrade`.

jsquire:
> It would be nice to be able to send a batch where every message has it's own partition_key.

For context, this is a service limitation.  There is no way to send a single batch of messages with heterogeneous partition keys.  

[...]

PR #25406, opened 2022-07-26
[Eventhub] Fix Blocking Behavior of Buffered Producer Flush

kashifkhan:
This is in reference to the issue #23510

Thanks to @annatisch for helping with the bug :)

- Identified that the blocking behavior of `queue.get` was not needed for both sync & async. Flush should simply empty the queue, reset the buffer length and move on. This is why `block=False` is passed in and `cur_buffered_len` is reset on Empty exception
- Added locks around `put_event` as it is modifying some shared variables such as `curr_batch`, `cur_buffered_len` as well as attempts a `flush` when the queue is full. This way we won't be adding things into the queue while flushing, leading to an infinite loop.


sdk/eventhub/azure-eventhub/samples/async_samples/send_buffered_mode_async.py
```diff
@@ -58,4 +58,4 @@ async def run():
 
 start_time = time.time()
 asyncio.run(run())
-print("Send messages in {} seconds.".format(time.time() - start_time))
```
swathipil, line 61: I think this file needs a newline at the end?

sdk/eventhub/azure-eventhub/azure/eventhub/_buffered_producer/_buffered_producer.py
```diff
@@ -85,39 +85,40 @@ def put_events(self, events, timeout_time=None):
...
+            except TypeError:
+                new_events_len = 1
+            if self._max_buffer_len - self._cur_buffered_len < new_events_len:
+                _LOGGER.info(
+                    "The buffer for partition %r is full. Attempting to flush before adding %r events.",
+                    self.partition_id,
+                    new_events_len,
+                )
+                # flush the buffer
+                with self._lock:
```
l0lawrence, line 100: Question - what does using the same lock twice do from a operational standpoint? Wouldn't we already have the lock to flush since we are in the first with statement.

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_buffered_producer/_buffered_producer_async.py
```diff
@@ -85,41 +85,43 @@ async def put_events(self, events, timeout_time=None):
...
+            except TypeError:
+                new_events_len = 1
+
+            if self._max_buffer_len - self._cur_buffered_len < new_events_len:
+                _LOGGER.info(
+                    "The buffer for partition %r is full. Attempting to flush before adding %r events.",
+                    self.partition_id,
+                    new_events_len,
+                )
+                # flush the buffer
```
l0lawrence, line 100: here we just use the 1 lock - is that just a difference between async/sync?

l0lawrence (commented):
I just had some questions about the locking -- wasn't sure how it works

sdk/eventhub/azure-eventhub/azure/eventhub/_buffered_producer/_buffered_producer.py
kashifkhan: good catch, there should only be one lock. No need for lockception :)

sdk/eventhub/azure-eventhub/azure/eventhub/_buffered_producer/_buffered_producer.py
```diff
@@ -85,39 +85,39 @@ def put_events(self, events, timeout_time=None):
...
-            new_events_len = 1
-        if self._max_buffer_len - self._cur_buffered_len < new_events_len:
-            _LOGGER.info(
-                "The buffer for partition %r is full. Attempting to flush before adding %r events.",
-                self.partition_id,
-                new_events_len,
-            )
-            # flush the buffer
-            with self._lock:
+        with self._lock:
```
annatisch, line 88: I wouldn't lock the whole method like this.
When flush is called on line 100, it will also try to take the lock and likely hang.
We only need to lock access to the queue.

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_buffered_producer/_buffered_producer_async.py
```diff
@@ -146,10 +148,16 @@ async def _flush(self, timeout_time=None, raise_error=True):
         if self._cur_batch:  # if there is batch, enqueue it to the buffer first
             self._buffered_queue.put(self._cur_batch)
             self._cur_batch = EventDataBatch(self._max_message_size_on_link)
-        while self._cur_buffered_len:
+        while self._buffered_queue.qsize() > 0:
             remaining_time = timeout_time - time.time() if timeout_time else None
             if (remaining_time and remaining_time > 0) or remaining_time is None:
-                batch = self._buffered_queue.get()
+                print("Getting an item from the queue to start flush")
```
annatisch, line 154: Whoops :)

sdk/eventhub/azure-eventhub/tests/livetest/asynctests/test_buffered_producer_async.py
```diff
@@ -490,3 +490,51 @@ async def on_error(events, pid, err):
 
     await consumer.close()
     await receive_thread
+
+@pytest.mark.liveTest
+@pytest.mark.asyncio
+async def test_long_wait_small_buffer(connection_str):
```
annatisch, line 496: Did this test actually repro the reported hanging bug before we fixed it?

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_buffered_producer/_buffered_producer_async.py
kashifkhan: doh

sdk/eventhub/azure-eventhub/tests/livetest/asynctests/test_buffered_producer_async.py
kashifkhan: No this was a test I added in after the fact, this was the repro code to catch the bug.

sdk/eventhub/azure-eventhub/azure/eventhub/_buffered_producer/_buffered_producer.py
kashifkhan: I also changed the behavior on the async side to match it.

sdk/eventhub/azure-eventhub/azure/eventhub/_buffered_producer/_buffered_producer.py
```diff
@@ -140,10 +141,15 @@ def flush(self, timeout_time=None, raise_error=True):
...
-            while self._cur_buffered_len:
+            while self._buffered_queue.qsize() > 0:
                 remaining_time = timeout_time - time.time() if timeout_time else None
                 if (remaining_time and remaining_time > 0) or remaining_time is None:
-                    batch = self._buffered_queue.get()
+                    try:
+                        batch = self._buffered_queue.get(block=False)
+                    except queue.Empty:
+                        #reset buffered count
+                        self._cur_buffered_len = 0
```
annatisch, line 151: We're only going to enter this `except` clause in the odd situation that the `while qsize() > 0` loop continued to iterate yet it still raised an Empty. So we should probably reset the `cur_buffered_len` outside of the loop?

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_buffered_producer/_buffered_producer_async.py
```diff
@@ -89,18 +89,16 @@ async def put_events(self, events, timeout_time=None):
...
                 self.partition_id,
                 new_events_len,
             )
             # flush the buffer
-            async with self._lock:
-                await self._flush(timeout_time=timeout_time)
-
+            await self.flush(timeout_time=timeout_time)
         if timeout_time and time.time() > timeout_time:
+            print("Failed to enqueue events into buffer due to timeout.")
```
annatisch, line 101: Another sneaky print~

sdk/eventhub/azure-eventhub/azure/eventhub/_buffered_producer/_buffered_producer.py
kashifkhan: I can do that, the only reason I had it over there was because if the while loop exited normally then `cur_buffered_len` would be 0 already.

annatisch (approved):

PR #25411, opened 2022-07-26
[Storage] Migrate `test_page_blob` to test proxy

jalauzon-msft:
In addition to the migration, adjusting blob sizes to reduce recording sizes.

One test is left skipped here as we still need to investigate the failure. The code has been migrated to test proxy though.


vincenttran-msft (approved):
LGTM! Thanks for getting to this 😄

PR #25437, opened 2022-07-27
Don't skip checkout exclusion in release config

benbp:
This is not necessary and is causing us to do a full checkout unnecessarily.


weshaggard (approved):

eng/pipelines/templates/stages/archetype-python-release.yml
```diff
@@ -171,7 +171,6 @@ stages:
                         steps:
                           - template: /eng/common/pipelines/templates/steps/sparse-checkout.yml
                             parameters:
-                              SkipDefaultCheckout: true
```
scbedd, line 174: I see that eng get's pulled in regardless. Concerns gone.

scbedd (approved):

PR #25455, opened 2022-07-28
[storage] remove dead retry 

Stevenjin8:
# Description

This PR is a little different from #25017 and the question is around the retry mechanism in

https://github.com/Azure/azure-sdk-for-python/blob/33acf79434ce1ab78093fc36b27b4ff7ee5f0e2a/sdk/storage/azure-storage-blob/azure/storage/blob/aio/_download_async.py#L335-L391

which was added in #18164. Notably, this retry mechanism catches `ClientPayloadError`. A few months later, #20888 changed `AioHttpTransportResponse.load_body` to catch `ClientPayloadError`'s and raise them as `IncompleteReadError`. So right now, the retry mechanism isn't doing anything (also when i remove it, all tests pass).

So far, this is looking a lot like #25017. But the crucial difference lies in where the body of the response is loaded. In #25017, the body of the response was loaded _outside_ the pipeline in the `process_content` function. For the async clients, the data is loaded inside the pipeline.

The pipeline is defined in 

https://github.com/Azure/azure-sdk-for-python/blob/33acf79434ce1ab78093fc36b27b4ff7ee5f0e2a/sdk/storage/azure-storage-blob/azure/storage/blob/_shared/base_client_async.py#L92-L106

The `AsyncStorageResponseHook` is __after__ `config.retry_policy`. If we look at `AsyncStorageResponseHook.send`, we see

https://github.com/Azure/azure-sdk-for-python/blob/33acf79434ce1ab78093fc36b27b4ff7ee5f0e2a/sdk/storage/azure-storage-blob/azure/storage/blob/_shared/policies_async.py#L65

meaning that the body of the response is loaded later in the pipeline than the retry policy. Thus, if loading the body of the response raises a `IncompleteReadError`, the pipeline will retry the request. If the retry policy runs out of retries, then it will raise a `ServiceResponseError`.

In other words, if we modify the #18164 's retry mechanism to catch `ServiceResponseError`s instead of `aiohttp.ClientPayloadError`s, we would be tripling the number of retries as @jalauzon-msft/@vincenttran-msft suggested earlier.

Given that, I think the best way forward is to revert #18164 's retry mechanism because we already have a working retry mechanism and the current code shouldn't ever run.


Stevenjin8:
currently working on tests. They are only passing when playing live. I think something funky is going on with vcr

Stevenjin8:
> currently working on tests. They are only passing when playing live. I think something funky is going on with vcr

We could also mark the test as live-only.

jalauzon-msft:
Hi @Stevenjin8, thanks for the investigation / explanation about why this retry code is no longer needed. I think you are correct in that the response streaming case does not really come into play here in async since `load_body` is in the response hook. What really sold it for me was the fact that in the `process_content` function for async we are just grabbing the repsonse body that has already been loaded:

https://github.com/Azure/azure-sdk-for-python/blob/e789c735b9c11cee94bcaef66b23e09f66e6dc86/sdk/storage/azure-storage-blob/azure/storage/blob/aio/_download_async.py#L35

So here we are really only wrapping the calls to `_client.download` in retry which is not needed since the normal download calls will go through our regular pipeline with retries. 

For me, I think there is still a question of whether `load_body` will actually perform retries if response streaming is taking place. It is in the pipeline, yes, but I'm not sure the response streaming requests will actually go through the pipeline and be retried. We may end up needing to wrap `load_body` in retries. But this does not affect this PR so we should be good.

sdk/storage/azure-storage-blob/azure/storage/blob/aio/_download_async.py
```diff
@@ -12,8 +12,8 @@
 from typing import AsyncIterator, Generic, TypeVar
 
 import asyncio
-from aiohttp import ClientPayloadError
-from azure.core.exceptions import HttpResponseError, ServiceResponseError
+
+from azure.core.exceptions import HttpResponseError, ServiceResponseError, IncompleteReadError
```
jalauzon-msft, line 16: I think `IncompleteReadError` is unused here?

sdk/storage/azure-storage-blob/tests/test_retry_async.py
```diff
@@ -471,4 +474,38 @@ async def test_invalid_account_key_async(self, storage_account_name, storage_acc
...
+    def _count_wrapper(counter, func):
+        @functools.wraps(func)
+        def wrapper(*args, **kwargs):
+            counter[0] += 1
+            return func(*args, **kwargs)
+        return wrapper
+
+    @BlobPreparer()
+    @AsyncStorageTestCase.await_prepared_test
+    async def test_streaming_retry_async(self, storage_account_name, storage_account_key):
```
jalauzon-msft, line 487: If the recording is giving a lot of trouble, we can just mark this live only as you suggested.

jalauzon-msft (approved):

sdk/storage/azure-storage-blob/tests/test_retry_async.py
Stevenjin8: marked as live-onlu

PR #25494, opened 2022-08-01
[EventHubs] uamqp switch support

swathipil:
Main changes:
- Moving code shared from SDK layer to /_transport/_uamqp_transport.py
- Remove _message property and Message construction out of AmqpAnnotatedMessage

TODO:
- [x] update uamqp mindependency version after releasing uamqp with changes from PR. Tests will fail until then.


sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
```diff
@@ -525,8 +503,18 @@ class EventDataBatch(object):
...
     """
 
-    def __init__(self, max_size_in_bytes=None, partition_id=None, partition_key=None):
-        # type: (Optional[int], Optional[str], Optional[Union[str, bytes]]) -> None
+    def __init__(
+        self,
+        max_size_in_bytes: Optional[int] = None,
+        partition_id: Optional[str] = None,
+        partition_key: Optional[Union[str, bytes]] = None,
+        **kwargs,
```
swathipil, line 511: @annatisch - adding kwargs is the only public API change (other than making the `EventData.message` a property instead of an ivar). B/c we recommend creating this object with producer_client.create_batch(), I assumed that we would be able to just pass the amqp transport to the EventDataBatch internally. However, since it's possible that users are constructing this manually, I was thinking that we could do the following in the future once we expose the `uamqp_transport` kwarg.

```
def __init__(
        self,
        max_size_in_bytes: Optional[int] = None,
        partition_id: Optional[str] = None,
        partition_key: Optional[Union[str, bytes]] = None,
        **kwargs,
):
try:
  self._amqp_transport = kwargs.pop("amqp_transport")
except KeyError:
  try:
    uamqp_transport = kwargs.pop("uamqp_transport")
    self._amqp_transport = UamqpTransport if uamqp_transport else PyamqpTransport
  except KeyError:
    raise TypeError("uamqp_transport is a required keyword.")
```

One other potential option would be to move out BatchMessage construction from here to in `send_batch`. However, I'm thinking that:
1) It might not be efficient to have to add the Messages to the BatchMessage in a loop in send instead of in add here?
2) The message sizes seem to be different b/w uamqp Message and pyamqp Message for some reason. So, when we check whether the batch is full, it might differ from when that error would occur for uamqp Message vs pyamqp Message.

sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
```diff
@@ -298,28 +288,28 @@ def enqueued_time(self):
...
-        try:
-            return self._raw_amqp_message.annotations[PROP_PARTITION_KEY_AMQP_SYMBOL]
-        except KeyError:
-            return self._raw_amqp_message.annotations.get(PROP_PARTITION_KEY, None)
+        # TODO: Ask Anna. can we remove the try and just do except? Haven't seen a case where symbol is used to get.
+        # try:
+        #    return self._raw_amqp_message.annotations[types.AMQPSymbol(PROP_PARTITION_KEY)]
+        # except KeyError:
+        #    return self._raw_amqp_message.annotations.get(PROP_PARTITION_KEY, None)
+        return self._raw_amqp_message.annotations.get(PROP_PARTITION_KEY, None)
```
swathipil, line 301: question @annatisch : can we just do the following?

sdk/eventhub/azure-eventhub/azure/eventhub/_transport/_base.py
```diff
@@ -0,0 +1,223 @@
+# --------------------------------------------------------------------------------------------
```
swathipil, line 1: this file might not be needed anymore. Just added it originally to keep track of methods I'd need for both pyamqp and uamqp.

sdk/eventhub/azure-eventhub/azure/eventhub/_constants.py
```diff
@@ -45,10 +45,33 @@
...
+     Amqp: AMQP over the default TCP transport protocol, it uses port 5671.
+     AmqpOverWebsocket: Amqp over the Web Sockets transport protocol, it uses
+     port 443.
+    """
+    Amqp = 1
+    AmqpOverWebsocket = 2
+
+DEFAULT_AMQPS_PORT = 5671
+DEFAULT_AMQP_WSS_PORT = 443
+READ_OPERATION = b"READ"
```
swathipil, line 77: API change to add TransportType to the SDK level here, rather than calling the one from uamqp.constants.

sdk/eventhub/azure-eventhub/CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Release History
 
+## 5.10.1 (Unreleased)
+
+### Bugs Fixed
+
+### Other Changes
+
+- Internal refactoring to support upcoming Pure Python AMQP-based release.
```
swathipil, line 9: not sure if we should say this at all.

sdk/eventhub/azure-eventhub/azure/eventhub/_constants.py
swathipil: should we wait to make this API change until pyamqp gets merged, and continue to use uamqp TransportType + constants for now?

sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
swathipil: Should we wait to make this API change until pyamqp gets merged, so that we don't have to bump minor version number until then? If so, I can change this so that we don't take kwargs and have: ` self._amqp_transport = UamqpTransport`.

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_producer_client_async.py
```diff
@@ -402,7 +405,7 @@ def from_connection_string(
         auth_timeout: float = 60,
         user_agent: Optional[str] = None,
         retry_total: int = 3,
-        transport_type: Optional["TransportType"] = None,
+        transport_type: TransportType = TransportType.Amqp,
```
swathipil, line 408: Default is TransportType.Amqp according to docstring. Passing in None results in error after updating uamqp to compare `TransportType.Amqp.value == transport_type.value`, rather than `TransportType.Amqp == transport_type`.

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_consumer_client_async.py
```diff
@@ -231,7 +232,7 @@ def from_connection_string(
         auth_timeout: float = 60,
         user_agent: Optional[str] = None,
         retry_total: int = 3,
-        transport_type: Optional["TransportType"] = None,
+        transport_type: TransportType = TransportType.Amqp,
```
swathipil, line 235: Default is TransportType.Amqp according to docstring. Passing in None results in error after updating uamqp to compare `TransportType.Amqp.value == transport_type.value`, rather than `TransportType.Amqp == transport_type`.

sdk/eventhub/azure-eventhub/CHANGELOG.md
kashifkhan: I think its fair game since we plan on promoting the alpha in the changelog. Changes for something coming in the future, that is known is ok

sdk/eventhub/azure-eventhub/azure/eventhub/_consumer.py
```diff
@@ -97,110 +93,86 @@ def __init__(self, client, source, **kwargs):
...
         self._keep_alive = keep_alive
         self._auto_reconnect = auto_reconnect
-        self._retry_policy = errors.ErrorPolicy(
-            max_retries=self._client._config.max_retries,
-            on_error=_error_handler,  # pylint:disable=protected-access
-        )
+        self._retry_policy = self._amqp_transport.create_retry_policy(self._client._config)
         self._reconnect_backoff = 1
-        self._link_properties = {}  # type: Dict[types.AMQPType, types.AMQPType]
+        link_properties: Dict[uamqp_types.AMQPType, uamqp_types.AMQPType] = {}
```
l0lawrence, line 98: Why is this no longer self._link?

sdk/eventhub/azure-eventhub/azure/eventhub/_consumer.py
```diff
@@ -97,110 +93,86 @@ def __init__(self, client, source, **kwargs):
...
         self._error = None
         self._timeout = 0
-        self._idle_timeout = (idle_timeout * 1000) if idle_timeout else None
+        self._idle_timeout = (idle_timeout * self._amqp_transport.IDLE_TIMEOUT_FACTOR) if idle_timeout else None
         partition = self._source.split("/")[-1]
         self._partition = partition
```
l0lawrence, line 103: Not related to your changes, is there a reason to have 102 and 103 as separate lines?

sdk/eventhub/azure-eventhub/azure/eventhub/_consumer.py
```diff
@@ -211,18 +183,17 @@ def receive(self, batch=False, max_batch_size=300, max_wait_time=None):
...
         deadline = self._receive_start_time + (
             max_wait_time or 0
-        )  # max_wait_time can be None
+        )
         if len(self._message_buffer) < max_batch_size:
             while retried_times <= max_retries:
                 try:
                     if self._open():
-                        self._handler.do_work()  # type: ignore
+                        self._handler.do_work(batch=self._prefetch)  # type: ignore
```
l0lawrence, line 191: Did we decide against doing this?

l0lawrence (commented):
Just some questions -- only went through sync

sdk/eventhub/azure-eventhub/azure/eventhub/_client_base.py
```diff
@@ -274,8 +282,16 @@ def get_token(self, *scopes, **kwargs):  # pylint:disable=unused-argument
...
-        # type: (str, str, CredentialTypes, Any) -> None
+    def __init__(
+        self,
+        fully_qualified_namespace: str,
+        eventhub_name: str,
+        credential: CredentialTypes,
+        **kwargs: Any,
+    ) -> None:
+        self._uamqp_transport = kwargs.pop("uamqp_transport", True)
+        self._amqp_transport = kwargs.pop("amqp_transport", UamqpTransport)
```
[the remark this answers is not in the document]
swathipil: No so this uamqp_transport is a bool. By default, it will be False --> amqp_transport = PyamqpTransport. If users pass in True --> amqp_transport = UamqpTransport.

sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
```diff
@@ -225,7 +224,7 @@ def _from_message(cls, message, raw_amqp_message=None):
         :rtype: ~azure.eventhub.EventData
         """
         event_data = cls(body="")
-        event_data.message = message
+        event_data._message = message
         # pylint: disable=protected-access
```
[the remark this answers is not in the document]
swathipil: this is to disable lines below it, so 229

sdk/eventhub/azure-eventhub/azure/eventhub/_consumer.py
swathipil: I pass this into amqp_transport._create_link_properties, which will then create a new dict, which I set as self._link_properties further down. The reason I do this is b/c the keys in this link properties object are passed to uamqp.types.AMQPSymbol and the values are wrapped in another uamqp.types object. For pyamqp, it wraps keys and values in a similar symbol/long pyamqp type.

sdk/eventhub/azure-eventhub/azure/eventhub/_consumer.py
swathipil: 🤔 oh I guess not. I just didn't touch it, b/c it's the same as pyamqp.

sdk/eventhub/azure-eventhub/azure/eventhub/_consumer.py
swathipil: based on our discussion, b/c pyamqp passes in batch to this method in that helper `receive_batch_messages` method, I believe we decided to keep pyamqp as is. I updated uamqp to take kwargs but disregard them, since that's an easier change. (+ In general, I think we want to adjust uamqp to fit pyamqp, rather than the other way around.)

sdk/eventhub/azure-eventhub/azure/eventhub/_client_base.py
```diff
@@ -11,15 +11,10 @@
...
-
-try:
-    from urlparse import urlparse
-    from urllib import quote_plus  # type: ignore
-except ImportError:
-    from urllib.parse import urlparse, quote_plus
-
-from uamqp import AMQPClient, Message, authentication, constants, errors, compat, utils
+from urllib.parse import urlparse, quote_plus
 import six
```
kashifkhan, line 15: We can also get rid of six here since its being used in one place below. Since for py3+ six.binary_type will always be bytes.
https://six.readthedocs.io/#six.binary_type

sdk/eventhub/azure-eventhub/azure/eventhub/_client_base.py
```diff
@@ -290,11 +306,12 @@ def __init__(self, fully_qualified_namespace, eventhub_name, credential, **kwarg
             self._credential = credential  # type: ignore
         self._keep_alive = kwargs.get("keep_alive", 30)
         self._auto_reconnect = kwargs.get("auto_reconnect", True)
-        self._mgmt_target = "amqps://{}/{}".format(
```
kashifkhan, line 293: was mgmt_target no longer needed? Don't see it re-added after the changes below

swathipil: yeah. This is only being used by uamqp and only in one place (mgmt client creation) so I re-constructed it there.

sdk/eventhub/azure-eventhub/azure/eventhub/_client_base.py
```diff
@@ -11,15 +11,10 @@
...
-try:
-    from urlparse import urlparse
-    from urllib import quote_plus  # type: ignore
-except ImportError:
-    from urllib.parse import urlparse, quote_plus
-
-from uamqp import AMQPClient, Message, authentication, constants, errors, compat, utils
+from urllib.parse import urlparse, quote_plus
 import six
+from uamqp import utils
```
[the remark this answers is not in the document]
swathipil: Ah, I was importing the create sas token method from uamqp.utils here since we're not adding the pyamqp library yet. Once pyamqp is merged, the plan was to use the pyamqp.utils.create_sas_token method. However, I can just add the pyamqp generate sas token method to _utils.py file here for now.

sdk/eventhub/azure-eventhub/azure/eventhub/_client_base.py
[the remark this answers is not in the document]
swathipil: We don't anymore, will remove it! (Originally, I was using it to build either the pyamqp-based EHSharedKeyCredential/etc. vs. uamqp-based EHSharedKeyCredential/etc. Missed removing it once we decided on just using pyamqp create_sas_token for the credentials, even with uamqp.)

sdk/eventhub/azure-eventhub/azure/eventhub/_client_base.py
```diff
@@ -385,26 +388,29 @@ def _backoff(
...
-            mgmt_client = AMQPClient(
-                self._mgmt_target, auth=mgmt_auth, debug=self._config.network_tracing
+            mgmt_client = self._amqp_transport.create_mgmt_client(
+                self._address, mgmt_auth=mgmt_auth, config=self._config
             )
             try:
-                conn = self._conn_manager.get_connection(  # pylint:disable=assignment-from-none
-                    self._address.hostname, mgmt_auth
-                )
-                mgmt_client.open(connection=conn)
```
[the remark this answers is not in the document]
swathipil: So I traced this through and the get_connection is always returning None (even before I made the change to remove the SharedConnectionManager class from the _connection_manager.py file, which I added back in). By default, the conn manager is being set to SeparateConnectionManager and I don't see SharedConnectionManager being passed in anywhere.

sdk/eventhub/azure-eventhub/azure/eventhub/_client_base.py
```diff
@@ -417,24 +423,25 @@ def _management_request(self, mgmt_msg, op_type):
...
-                        "Management authentication failed. Status code: {}, Description: {!r}".format(
-                            status_code, description
-                        )
-                    )
-                if status_code in [404]:
-                    raise ConnectError(
-                        "Management connection failed. Status code: {}, Description: {!r}".format(
-                            status_code, description
-                        )
+                    raise self._amqp_transport.get_error(
```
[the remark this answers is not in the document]
swathipil: I'm not sure why, but pyamqp is raising pyamqp.errors.AMQPConnectionError. This will probably need to be updated to a ConnectError there?

sdk/eventhub/azure-eventhub/azure/eventhub/_client_base.py
```diff
@@ -453,12 +460,13 @@ def _add_span_request_attributes(self, span):
         span.add_attribute("message_bus.destination", self._address.path)
         span.add_attribute("peer.address", self._address.hostname)
 
-    def _get_eventhub_properties(self):
-        # type:() -> Dict[str, Any]
-        mgmt_msg = Message(application_properties={"name": self.eventhub_name})
+    def _get_eventhub_properties(self) -> Dict[str, Any]:
+        mgmt_msg = self._amqp_transport.MESSAGE(
```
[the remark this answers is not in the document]
swathipil: Since MESSAGE is just set to the constant of `uamqp(/pyamqp).Message`, I thought this was convention? Would the alternative be just "Message"?

sdk/eventhub/azure-eventhub/azure/eventhub/_client_base.py
```diff
@@ -533,17 +539,13 @@ def _open(self):
                 self._handler.close()
             auth = self._client._create_auth()
             self._create_handler(auth)
-            self._handler.open(
-                connection=self._client._conn_manager.get_connection(
-                    self._client._address.hostname, auth
-                )  # pylint: disable=protected-access
-            )
+            self._handler.open()
```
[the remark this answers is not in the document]
swathipil: get_connection also returns None here, since conn_manager is set to SeparateConnectionManager.

sdk/eventhub/azure-eventhub/azure/eventhub/_client_base.py
```diff
@@ -533,17 +539,13 @@ def _open(self):
...
             while not self._handler.client_ready():
                 time.sleep(0.05)
             self._max_message_size_on_link = (
-                self._handler.message_handler._link.peer_max_message_size
-                or constants.MAX_MESSAGE_LENGTH_BYTES
-            )  # pylint: disable=protected-access
+                self._amqp_transport.get_remote_max_message_size(self._handler)
+                or self._amqp_transport.MAX_FRAME_SIZE_BYTES
```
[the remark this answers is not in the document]
swathipil: that makes sense. In pyamqp, the value here is pyamqp.constants.MAX_FRAME_SIZE_BYTES. It looks like pyamqp.constants doesn't even have a MAX_MESSAGE_LENGTH_BYTES value defined, so I wasn't sure what the difference was. To circumvent this, I did the following for the two transports:
```python
class UamqpTransport:
  MAX_FRAME_SIZE_BYTES = uamqp.constants.MAX_MESSAGE_LENGTH_BYTES

class PyamqpTransport:
  MAX_FRAME_SIZE_BYTES = pyamqp.constants.MAX_FRAME_SIZE_BYTES
```
Not sure if there's a better way to do this? Or if pyamqp should have a separate MAX_MESSAGE_LENGTH_BYTES value as well?

sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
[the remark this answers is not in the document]
swathipil: ahh, so I should prob move this line up, good catch @l0lawrence!

sdk/eventhub/azure-eventhub/azure/eventhub/_consumer.py
swathipil: updated

sdk/eventhub/azure-eventhub/azure/eventhub/_utils.py
```diff
@@ -20,32 +20,34 @@
...
     USER_AGENT_PREFIX,
     PROP_LAST_ENQUEUED_SEQUENCE_NUMBER,
     PROP_LAST_ENQUEUED_TIME_UTC,
     PROP_RUNTIME_INFO_RETRIEVAL_TIME_UTC,
     PROP_LAST_ENQUEUED_OFFSET,
     PROP_TIMESTAMP,
+    PROP_PARTITION_KEY
 )
 
+PROP_PARTITION_KEY_AMQP_SYMBOL = uamqp_types.AMQPSymbol(PROP_PARTITION_KEY)
```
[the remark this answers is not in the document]
swathipil: will move `set_event_partition_key` out to the UamqpTransport, so I can remove this.

sdk/eventhub/azure-eventhub/azure/eventhub/_utils.py
```diff
@@ -314,14 +288,16 @@ def transform_outbound_single_message(message, message_type):
     :rtype: EventData
     """
     try:
-        # EventData
         # pylint: disable=protected-access
-        return message._to_outgoing_message()  # type: ignore
+        # EventData
+        message._message = to_outgoing_amqp_message(message.raw_amqp_message)
```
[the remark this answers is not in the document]
swathipil: Yeah, so previously we had the EventData._to_outgoing_message(). In that method, we had the following:
```event_data.message = event_data.raw_amqp_message.to_outgoing_amqp_message()```.
Just to clarify what's happening, two consecutive changes were made here:
1) We no longer have the `to_outgoing_amqp_message` on the AmqpAnnotatedMessage, b/c we want to avoid adding pyamqp/uamqp specific code in AmqpAnnotatedMessage. So, we're constructing the outgoing amqp message here. At this point, this line would have looked like `message.message = to_outgoing_amqp_message(message.raw_amqp_message)`.
2) Next change made: Making the `EventData.message` ivar private and adding the `EventData.message` public property. Everywhere that we previously called `event_data.message`, we're now calling `event_data._message` instead. This is in preparation for adding LegacyMessage when we merge pyamqp. For right now, the EventData.message getter and setter are both = `self._message`. So technically, having this line be `message.message = ...` would be fine for now. However, b/c we likely won't be calling `message.message` when LegacyMessage is merged, I'm just mutating the internal message here.

Idk if I answered your original question and maybe I overexplained what you already knew, so please lmk if this doesn't make sense/you have concerns 😄

File: sdk/eventhub/azure-eventhub/azure/eventhub/_utils.py [hunk 23]
```diff
@@ -314,14 +288,16 @@ def transform_outbound_single_message(message, message_type):
...
+        return message  # type: ignore
     except AttributeError:
-        # AmqpAnnotatedMessage
         # pylint: disable=protected-access
+        # AmqpAnnotatedMessage
+        amqp_message = to_outgoing_amqp_message(message)
         return message_type._from_message(
-            message=message._to_outgoing_amqp_message(), raw_amqp_message=message  # type: ignore
+            message=amqp_message, raw_amqp_message=message  # type: ignore
```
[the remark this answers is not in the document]
swathipil: yeah sorry, naming everything message is def confusing and probably was not the best choice 😅. I can add more comments here to describe what's happening.
1) In the try block above (case: message=EventData), we want to create the outgoing uamqp/pyamqp internal Message object. All the changes that are made by the user to EventData properties are persisted on the EventData.raw_amqp_message object (AmqpAnnotatedMessage), and _not_ on the internal EventData._message object (uamqp/pyamqp Message). The `AmqpTransport.to_outgoing_amqp_message` takes the AmqpAnnotatedMessage, and returns the corresponding uamqp/pyamqp Message. The internal EventData._message is set to this outgoing uamqp/pyamqp Message object. The EventData._message objects will be sent.
2) In the except block here (case: message=AmqpAnnotatedMessage, message_type=EventData), we call EventData._from_message, which will create the EventData object from two args:
  1. `message: uamqp/pyamqp Message` param. In line 298, we're creating the pyamqp/uamqp Message from the AmqpAnnotatedMessage. Inside `EventData._from_message`, `event_data._message` will be set to this Message.
  2. `raw_amqp_message: AmqpAnnotatedMessage` optional param. This is optional b/c we call `EventData._from_message` in two places. 
    * One is here during sending.  This call would still work if we didn't pass in raw_amqp_message here. However, since we already have the raw_amqp_message, we pass it in so that it doesn't get re-constructed inside for efficiency.
    * The other is when we receive. In receive, we only pass in the message=uamqp/pyamqp Message to this `EventData._from_message` method.

Please lmk if anything was unclear or if you have any concerns. :)

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_connection_manager_async.py
```diff
@@ -28,6 +33,62 @@ async def reset_connection_if_broken(self) -> None:
...
+        self._channel_max = kwargs.get("channel_max")
+        self._idle_timeout = kwargs.get("idle_timeout")
+        self._remote_idle_timeout_empty_frame_send_ratio = kwargs.get(
+            "remote_idle_timeout_empty_frame_send_ratio"
+        )
+
+    async def get_connection(self, host: str, auth: "JWTTokenAsync") -> ConnectionAsync:
+        async with self._lock:
+            if self._conn is None:
+                self._conn = ConnectionAsync(
```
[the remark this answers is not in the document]
swathipil: Yep! For now, I haven't touched this code b/c I don't see anywhere in our code or tests where we're using the SharedConnectionManager. I might be missing something?

I can update this now, though.

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_producer_async.py
```diff
@@ -91,68 +94,47 @@ def __init__(self, client: "EventHubProducerClient", target: str, **kwargs) -> N
...
             keep_alive_interval=self._keep_alive,
             client_name=self._name,
             link_properties=self._link_properties,
             properties=create_properties(
-                self._client._config.user_agent  # pylint:disable=protected-access
+                self._client._config.user_agent,  # pylint: disable=protected-access
+                amqp_transport=self._amqp_transport,
             ),
-            **self._internal_kwargs
+            msg_timeout=self._timeout * 1000,
```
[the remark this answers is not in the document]
swathipil: ah, yep can update it! The pyamqp client doesn't take msg_timeout at all. Is that by design?

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_consumer_async.py
```diff
@@ -91,139 +88,70 @@ def __init__(self, client: "EventHubConsumerClient", source: str, **kwargs) -> N
...
+        link_properties: Dict[types.AMQPType, types.AMQPType] = {}
         partition = self._source.split("/")[-1]
         self._partition = partition
-        self._name = "EHReceiver-{}-partition{}".format(uuid.uuid4(), partition)
+        self._name = f"EHReceiver-{uuid.uuid4()}-partition{partition}"
         if owner_level is not None:
-            self._link_properties[types.AMQPSymbol(EPOCH_SYMBOL)] = types.AMQPLong(
-                int(owner_level)
-            )
+            link_properties[EPOCH_SYMBOL] = int(owner_level)
```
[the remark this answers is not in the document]
swathipil: yep, all the keys in link_properties here are getting wrapped by the uamqp.types.AMQPSymbol in the UamqpTransport.

sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
swathipil: Since this partition key is only added to the message.annotations/raw_amqp_message.annotations in the `send` method and is not settable on the EventData, I don't think this would ever return anything other than None before sending if the user calls it. So keeping only the except logic here, unless there are any concerns?

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_connection_manager_async.py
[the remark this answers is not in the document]
swathipil: updated with amqp transport.

sdk/eventhub/azure-eventhub/azure/eventhub/_client_base.py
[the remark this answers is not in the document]
annatisch: Nothing currently uses the SharedConnectionManager by default - it's only triggered on an (apparently undocumented) kwarg.
The full scope/behaviour of the ConnectionManager should remain in-tact.

sdk/eventhub/azure-eventhub/azure/eventhub/_client_base.py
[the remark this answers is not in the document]
annatisch: I would probably just replace this with a method:
`self._amqp_transport.build_message(...)`

sdk/eventhub/azure-eventhub/azure/eventhub/_client_base.py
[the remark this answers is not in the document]
annatisch: Yes - but that's intended.

sdk/eventhub/azure-eventhub/azure/eventhub/_client_base.py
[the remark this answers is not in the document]
annatisch: Yeah - it's probably not _too_ concerning - given that the batch max message size is determined from the service value any way - and I _think_ our old uamqp might have been incorrect/the same any way.
I would add two separate constants to pyamqp for this - even if their values are the same, so we can adapt them accordingly.

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_producer_async.py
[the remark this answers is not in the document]
annatisch: Yes and no - it's because the client transport in pyamqp is full of holes and inaccuracies and undocumented refactors.
There's a good chance most of the kwargs we pass into the pyamqp client are either being ignored or are not working the way we think they are - so whenever you see things like this please mark withm with a big scary TODO so we remember to go back and see what's missing in pyamqp.
I wont have to time to investigate the pyamqp client in time for this pr - so we probably shouldn't block on it.

sdk/eventhub/azure-eventhub/azure/eventhub/_client_base.py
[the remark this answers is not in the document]
annatisch: Ah yup - I get it now - for some reason I thought that the previous code `AuthenticationException` and `ConnectError` were servicebus errors - but I see we were raising native uamqp errors here.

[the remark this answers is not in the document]
annatisch: Maybe we could merge this into a single function that takes the status code as a parameter - that way we can move the 404/401 conditional logic into the transport so we don't need to use the AUTH_EXCEPTION/CONNECTION_ERROR constants to pass around class types which is more complicated to document/maintain?
```python
if status_code < 400:
    return response
else:
    raise raise self._amqp_transport.get_error(status_code)
```

sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
annatisch: Yes I would leave it out of the API for now and just default to UamqpTransport.
I think we can make the Batch Message itself transport agnostic (unless it's instantiated using `client.send_batch()`).
The only thing it needs the transport for is tracking the encoded size - and while, as you say, there's some difference in the exact byte count between the two implementations, I think we'd be pretty save using pyamqp encoder (which will encodes superfluous data and so is usually bigger).

sdk/eventhub/azure-eventhub/azure/eventhub/_consumer.py
```diff
@@ -97,110 +93,85 @@ def __init__(self, client, source, **kwargs):
...
+        self._handler = self._amqp_transport.create_receive_client(
+            config=self._client._config,    # pylint:disable=protected-access
+            source=source,
             auth=auth,
-            debug=self._client._config.network_tracing,  # pylint:disable=protected-access
-            prefetch=self._prefetch,
+            network_trace=self._client._config.network_tracing,  # pylint:disable=protected-access
+            link_credit=self._prefetch,
             link_properties=self._link_properties,
-            timeout=self._timeout,
```
annatisch, line 152: Why did this one get removed? I'm pretty sure this is not supported by pyamqp yet - but was probably doing something in uamqp?

sdk/eventhub/azure-eventhub/azure/eventhub/_producer.py
```diff
@@ -186,19 +167,33 @@ def _on_outcome(self, outcome, condition):
...
+        timeout_time: Optional[float] = None,
+        last_exception: Optional[Exception] = None,
+    ) -> None:
+        if self._unsent_events:
+            self._amqp_transport.send_messages(
+                self, timeout_time, last_exception, _LOGGER
+            )
+
+    def _send_event_data_with_retry(self, timeout: Optional[float] = None) -> None:
+        return self._do_retryable_operation(self._send_event_data, timeout=timeout)
```
annatisch, line 181: Later we might need to look at moving the "do_retryable_operation" into the transport as well - as this is built into pyamqp and so wont be needed later. (Don't need to change it now though).

sdk/eventhub/azure-eventhub/azure/eventhub/_producer.py
```diff
@@ -217,24 +212,26 @@ def _wrap_eventdata(
...
                 wrapper_event_data = event_data  # type:ignore
             else:
                 if partition_key:
-                    event_data = _set_partition_key(event_data, partition_key)
+                    event_data = _set_partition_key(
+                        event_data, partition_key, self._amqp_transport
+                    )
                 event_data = _set_trace_message(event_data, span)
-                wrapper_event_data = EventDataBatch._from_batch(event_data, partition_key)  # type: ignore  # pylint: disable=protected-access
-        wrapper_event_data.message.on_send_complete = self._on_outcome
```
annatisch, line 228: This line was no longer needed?

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_transport/_uamqp_transport_async.py
```diff
@@ -0,0 +1,371 @@
...
+            auth,
+            **kwargs
+        )
+
+    @staticmethod
+    async def close_connection_async(connection):
+        """
+        Closes existing connection.
+        :param connection: uamqp or pyamqp Connection.
+        """
```
annatisch, line 75: Isn't this missing the close call?

sdk/eventhub/azure-eventhub/azure/eventhub/_producer.py
annatisch: Nevermind - I see this moved to the transport.send_messages method.

sdk/eventhub/azure-eventhub/azure/eventhub/_client_base.py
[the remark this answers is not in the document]
swathipil: sounds good, added the SharedConnectionManager back in

sdk/eventhub/azure-eventhub/azure/eventhub/_client_base.py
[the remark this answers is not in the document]
swathipil: updated

sdk/eventhub/azure-eventhub/azure/eventhub/_client_base.py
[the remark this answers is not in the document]
swathipil: updated

sdk/eventhub/azure-eventhub/azure/eventhub/_client_base.py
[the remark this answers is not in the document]
swathipil: updated. I will also add a comment in the pyamqp switch PR that we should be able to pass in connection to pyamqp handler.open() as well.

sdk/eventhub/azure-eventhub/azure/eventhub/_consumer.py
swathipil: ah I meant to add this in the uamqp_transport.create_receive_client, but didn't add it back in. Since we plan on supporting this in pyamqp, even if we don't support it yet, will add this back in here.

sdk/eventhub/azure-eventhub/azure/eventhub/_producer.py
swathipil: sounds good, created an issue for this here: #25700

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_transport/_uamqp_transport_async.py
swathipil: yes it is 🤦 updated!

sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
swathipil: updated so that EventDataBatch doesn't take kwargs. Will make EventDataBatch transport agnostic in the pyamqp switch PR.

sdk/eventhub/azure-eventhub/azure/eventhub/_client_base.py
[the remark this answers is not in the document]
swathipil: updated to have to separate constants here + added a comment to the pyamqp PR to add both constants there too.

sdk/eventhub/azure-eventhub/azure/eventhub/_constants.py
annatisch: I don't mind much so long as if customers are using the existing enum that they imported directly from uamqp (for some reason), that it would still work.

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_producer_client_async.py
annatisch: Should we handle that AttributeError just in case?

annatisch (approved):

kashifkhan (approved):

sdk/eventhub/azure-eventhub/azure/eventhub/_constants.py
swathipil: will update one of the tests that takes TransportType to import the one from uamqp, to double check.

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_producer_client_async.py
swathipil: I updated the logic so that it does the following:
```
 if transport_type and TransportType.AmqpOverWebsocket.value == transport_type.value or http_proxy:
   # AmqpOverWebsocket logic
else:
   # Amqp logic
```. 
I believe this will account for the transport_type being None (aka. Amqp) and go into the else before the AttributeError is raised.

Issue #24777, opened 2022-06-09
[textanalytics] consider adding `kind` to *Result types


kristapratico:
Adding a `kind` == typing.Literal to the *Result types in TA will make static type checking less painful, especially with the analyze actions union return type. This will let users write code which checks `kind` instead of doing `isinstance` checks to make mypy happy.

PR #25498, opened 2022-08-01
[textanalytics] add kind property to result | error

kristapratico:
Resolves https://github.com/Azure/azure-sdk-for-python/issues/24777

This allows users to do simple string checks instead of using isinstance or casting to make type checkers happy.

- adds `kind` to result and error types
- updates samples to be type checker clean

### DIFF
--- a/scripts/breaking_changes_checker/breaking_changes_allowlist.py
+++ b/scripts/breaking_changes_checker/breaking_changes_allowlist.py
@@ -8,7 +8,6 @@
 
 RUN_BREAKING_CHANGES_PACKAGES = [
     "azure-ai-formrecognizer",
-    "azure-ai-textanalytics"
 ]
 
 
--- a/sdk/textanalytics/azure-ai-textanalytics/CHANGELOG.md
+++ b/sdk/textanalytics/azure-ai-textanalytics/CHANGELOG.md
@@ -10,6 +10,7 @@
 - Added property `details` on returned poller objects which contain long-running operation metadata.
 - Added `TextAnalysisLROPoller` and `AsyncTextAnalysisLROPoller` protocols to describe the return types from long-running operations.
 - Added `cancel` method on the poller objects. Call it to cancel a long-running operation that's in progress.
+- Added `kind` to *Result types and to `DocumentError`.
 
 ### Breaking Changes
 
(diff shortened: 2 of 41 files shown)


kristapratico:
azp run python - textanalytics - tests

scripts/breaking_changes_checker/breaking_changes_allowlist.py
```diff
@@ -8,7 +8,6 @@
 
 RUN_BREAKING_CHANGES_PACKAGES = [
     "azure-ai-formrecognizer",
-    "azure-ai-textanalytics"
```
kristapratico, line 11: This legacy breaking changes checker is raising a false positive for `is_error` so skipping it in CI.

kristapratico, line 11: This legacy breaking changes checker is raising a false positive for `is_error` so skipping it in CI.

heaths, line 11: Is this relevant? No problem being in here, but wondering if this was just something you noticed or what. Mainly confused because you're adding APIs but removing a breaking change...thing.

mssfang (approved):

sdk/textanalytics/azure-ai-textanalytics/CHANGELOG.md
```diff
@@ -10,12 +10,16 @@
 - Added property `details` on returned poller objects which contain long-running operation metadata.
 - Added `TextAnalysisLROPoller` and `AsyncTextAnalysisLROPoller` protocols to describe the return types from long-running operations.
 - Added `cancel` method on the poller objects. Call it to cancel a long-running operation that's in progress.
+- Added property `kind` to `RecognizeEntitiesResult`, `RecognizePiiEntitiesResult`, `AnalyzeHealthcareEntitiesResult`,
```
heaths, line 13: Do you think this is something we should do across languages for much the same reason? I would say .NET is pretty inconsistent with type checks vs. discriminator checks, but wouldn't be opposed to it for each way.

heaths (approved):
Just a couple questions, really, but otherwise LGTM.

scripts/breaking_changes_checker/breaking_changes_allowlist.py
kristapratico: I tried to comment on that line but it didn't show up for some reason. This is a legacy breaking changes checker (which I wrote 😆) that we don't really use anymore. It was raising a false positive on `is_error` saying that I removed it from the API so I'm disabling it.

sdk/textanalytics/azure-ai-textanalytics/CHANGELOG.md
kristapratico: Discussed offline, but basically having a `kind` property makes more sense for Python/JS designs since we return a heterogenous list of result | error.

PR #25514, opened 2022-08-02
[Cosmos] Update max_integrated_cache_staleness_in_ms to allow 0

JericHunter:
Fixing a small issue with the integrated cache so that it can allow 0 as a value.

### DIFF
--- a/sdk/cosmos/azure-cosmos/CHANGELOG.md
+++ b/sdk/cosmos/azure-cosmos/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 #### Bugs Fixed
 - Fixed parsing of args for overloaded `container.read()` method.
+- Fixed `validate_cache_staleness_value()` method to allow max_integrated_cache_staleness to be an integer greater than or equal to 0
 
 #### Other Changes
 
--- a/sdk/cosmos/azure-cosmos/azure/cosmos/_base.py
+++ b/sdk/cosmos/azure-cosmos/azure/cosmos/_base.py
@@ -672,4 +672,5 @@ def create_scope_from_url(url):
 def validate_cache_staleness_value(max_integrated_cache_staleness):
     int(max_integrated_cache_staleness)  # Will throw error if data type cant be converted to int
     if max_integrated_cache_staleness < 0:
-        raise ValueError("Parameter 'max_integrated_cache_staleness_in_ms' can only be a positive integer.")
+        raise ValueError("Parameter 'max_integrated_cache_staleness_in_ms' can only be an "
+                         "integer greater than or equal to zero")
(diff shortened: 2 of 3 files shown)


sdk/cosmos/azure-cosmos/azure/cosmos/_base.py
```diff
@@ -671,5 +671,5 @@ def create_scope_from_url(url):
 
 def validate_cache_staleness_value(max_integrated_cache_staleness):
     int(max_integrated_cache_staleness)  # Will throw error if data type cant be converted to int
-    if max_integrated_cache_staleness <= 0:
+    if max_integrated_cache_staleness < 0:
         raise ValueError("Parameter 'max_integrated_cache_staleness_in_ms' can only be a positive integer.")
```
simorenoh, line 675: this message needs to be changed as well, since 0 is also accepted now

simorenoh (changes requested):
please update any mentions of value needing to be positive in README (if any), update CHANGELOG with these changes under bugfixes section, and add a relevant tests for setting 0 for integrated cache

sdk/cosmos/azure-cosmos/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 
 #### Bugs Fixed
 - Fixed parsing of args for overloaded `container.read()` method.
+- Fixed `validate_cache_staleness_value()` method to allow max_integrated_cache_staleness to be an integer greater than or equal to 0
```
simorenoh, line 12: please add the PR link here, other than that looks great thanks Jeric!

simorenoh (approved):

simorenoh (changes requested):
please fix the relevant tests for negative and 0 value to verify, the way the other tests do

sdk/cosmos/azure-cosmos/azure/cosmos/_base.py
```diff
@@ -671,5 +671,6 @@ def create_scope_from_url(url):
 
 def validate_cache_staleness_value(max_integrated_cache_staleness):
     int(max_integrated_cache_staleness)  # Will throw error if data type cant be converted to int
-    if max_integrated_cache_staleness <= 0:
-        raise ValueError("Parameter 'max_integrated_cache_staleness_in_ms' can only be a positive integer.")
+    if max_integrated_cache_staleness < 0:
+        raise ValueError("Parameter 'max_integrated_cache_staleness_in_ms' can only be an "
+                         "integer greater than or equal to zero")
```
tjprescott, line 676: Why are we doing client-side validation here instead of letting the service throw an error?

simorenoh: I believe a negative value still gets accepted by the service but gets ignored altogether - this way the user knows the ranges of numbers they can provide
@JericHunter can you please remove this validation and double check that this is the case?

tjprescott: If the service provides a good error, we should just remove this CS validation. If indeed it does simply ignore it, then this is good.

JericHunter: The service does not provide a good error so we should keep this

simorenoh (approved):

PR #25522, opened 2022-08-02
[Storage] Fix ranged download for client-side encryption

jalauzon-msft:
This change addresses a couple of issues with a ranged blob download (specifying `offset` and `length`) with client-side encryption enabled that can occur in rare circumstances. There is one fix for V1 and a separate fix for V2.


vincenttran-msft (approved):
Thanks for getting on this so quickly! Your focus on quality engineering work is impeccable 😄

jalauzon-msft:
Running just blob Live tests here: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=1756885&view=results

PR #25531, opened 2022-08-03
Use Python 3.10 to generate API stub file

praveenkuttappan:
APIView server has been using Python 3.10.2 to generate APIView file from wheel. So all automatic reviews were created using Python version 3.10.2.

CI pipeline generates APIView stub file using Python 3.9.x and this stub file is used to generate automatic review revision instead of uploading wheel now. APIView stub file has a lot of diff when running this using 3.9.x v/s 3.10.

This PR is to update Python version to run apistub step to 3.10 so three won't be any false negative diff in APIview which blocks release.


scbedd (approved):

scbedd:
Confirmed that using py310 isn't accidentally supplanting any builds (because we weren't assuring they were using py39). Given it's literally the last actual build step, we're fine 👍

Issue #25225, opened 2022-07-14
[Test Proxy] Set default header exclusions more reliably


mccoyp:
Today, we add a sanitizer and matcher to exclude certain, unnecessary headers in our tests as part of test proxy startup: https://github.com/Azure/azure-sdk-for-python/blob/ff97aa8a61dc6accc72ee11441f4909c32d7695e/tools/azure-sdk-tools/devtools_testutils/proxy_startup.py#L182-L186

There are a few problems with this approach:
1. The linked code is only run if the test proxy tool isn't already running locally. If someone had already started the tool before running tests -- which is a valid scenario -- recordings will include the headers that we intend to ignore, or tests will fail in playback if these headers aren't in recordings.
2. If someone restores the default sanitizer/matcher settings via `devtools_testutils.set_default_settings`, the sanitizer and matcher we added during setup will be removed.

Ideally, we should exclude these headers in all valid test running scenarios. We should also try to make this header exclusion resilient to getting overridden by proxy setting adjustments unless a user actually intends to undo this exclusion.

> Note: maybe re-adding the sanitizer and matcher as part of `set_default_settings` would work?

PR #25532, opened 2022-08-03
adding a proxy setup to base conftest. just need to disable it when p…

scbedd:
Resolves https://github.com/Azure/azure-sdk-for-python/issues/25225, in addition to sanitizing common culprits of credential leaks:
- Shared secret environment variables (`AZURE_X`)
- Library-specific secret environment variables (`SERVICE_X`)
- Authentication requests


mccoyp:
Hmm, I'll have to think about it. If it's something we'd want to be done whenever the test proxy is invoked, it might make sense to add it to the test proxy startup since I'm not sure if it otherwise has significantly more value in `conftest.py`. The new `recorded_test` fixture is still dependent on the `test_proxy` fixture, so putting this code in `test_proxy` would still cover the second round of migrations.

The main issue with this, I would think, is that most tests aren't using `AZURE_CLIENT_X` environment variables directly, but rather `{DIRECTORY}_CLIENT_X`. The former variables get assigned to the latter's values when the EnvironmentVariableLoader is invoked, but at test startup this wouldn't be sanitizing the correct values. If we sanitize service principal information in the EnviornmentVariableLoader, though -- like we do for explicitly requested variables -- then I think this plus that would cover our bases.

scbedd:
> The main issue with this, I would think, is that most tests aren't using `AZURE_CLIENT_X` environment variables directly, but rather `{DIRECTORY}_CLIENT_X`/

Dead on. BUT we could absolutely add these by default. Try to set up a few base ones that always get removed if not using PowerShellPreparer, etc. 


Your suggestion to move this into the test proxy startup is well taken. I'll make that code change here.

tools/azure-sdk-tools/devtools_testutils/sanitizers.py
```diff
@@ -597,6 +598,15 @@ def _send_reset_request(headers: dict) -> None:
     )
     response.raise_for_status()
 
+    headers_to_ignore = "Authorization, x-ms-client-request-id, x-ms-request-id"
```
scbedd, line 601: Not certain it's intended. But when we start up the test-proxy, we set a couple default custom items.

Then after reset, we don't actually resend those. I've moved all of the common default to `_send_reset` and then just called that during proxy startup.

One place, all the defaults. I could be missing reasoning here though 👍

tools/azure-sdk-tools/devtools_testutils/sanitizers.py
```diff
@@ -597,6 +598,15 @@ def _send_reset_request(headers: dict) -> None:
...
+    add_remove_header_sanitizer(headers=headers_to_ignore)
+    set_custom_default_matcher(excluded_headers=headers_to_ignore)
+
+    client_id = os.environ.get("AZURE_CLIENT_ID", "00000000-0000-0000-0000-000000000000")
+    client_secret = os.environ.get("AZURE_CLIENT_SECRET", "00000000-0000-0000-0000-000000000000")
+    add_general_regex_sanitizer(regex=client_id, value="00000000-0000-0000-0000-000000000000")
+    add_general_regex_sanitizer(regex=client_secret, value="00000000-0000-0000-0000-000000000000")
```
msyyc, line 608: sub_id and tenant_id are also mostly used so I think it is better to add them here too:
image

tools/azure-sdk-tools/devtools_testutils/sanitizers.py
mccoyp: Only setting the header rules once was actually intentional, albeit for rare edge cases. There are some teams that need request ID headers for tests, so being able to reset everything to a clean slate seemed like a good solution at the time.

I'd like to do something similar to your suggestion, so that we have a persistent default baseline of things to ignore -- otherwise, "resetting to default" isn't really giving people the default they're expecting.

To solve both those problems, it seems like the solution might be to provide methods for removing headers from this exclusion list 🤔

tools/azure-sdk-tools/devtools_testutils/sanitizers.py
scbedd: I think this base set of 4 is fine. Thanks for the heads up @msyyc

Issue #25489, opened 2022-08-01
Error while uploading an empty file to blob using azure-storage-blob: Data should be bytes or a seekable file-like object.


harneetvirk:
- **azure-storage-blob**: 
- **12.13.0**: 
- **Windows**:
- **Python 3.8**: 

**Describe the bug**
While using azure-storage-blob 12.13.0, we are getting an exception on uploading an empty file to blob storage: ValueError: Data should be bytes or a seekable file-like object.

**To Reproduce**
1. conda create -n storageblobpy38 python=3.8 -y
2. conda activate storageblobpy38
3. pip install azure-storage-blob==12.13.0
4. python scratch2.py

Content of scratch2.py
```
from azure.storage.blob import BlobServiceClient
import os

def write_script(script_path, content):
    """
    Util for generating a python script, currently writes the file to disk.
    """
    with open(script_path, "w") as stream:
        stream.write(content)

write_script('train.py', "")

with open('train.py', 'rb') as data:
    account_name = 'SOME_ACCOUNT_NAME'
    account_key = 'SOMEACCOUNTKEY'
    account_url = "https://{account_name}.blob.{endpoint}".format(
        account_name=account_name, endpoint='core.windows.net'
    )
    client = BlobServiceClient(
        account_url=account_url,
        credential={
            "account_name": account_name,
            "account_key": account_key
        }
    )

    client.get_blob_client(
        container='SOMECONTAINER',
        blob='3.py'
    ).upload_blob(data=data, overwrite=True, validate_content=True)

```

**Expected behavior**
The empty file should be allowed to be uploaded on the blob.

**Screenshots**
Screenshot showing the error and stack trace:

![image](https://user-images.githubusercontent.com/20194110/182202633-3723d117-121f-4813-9e38-95ba0d4a71a0.png)

**Additional context**
An Empty file was allowed to be uploaded in the previous versions of azure-storage-blob

Stevenjin8:
I think the issue is in

https://github.com/Azure/azure-sdk-for-python/blob/93a12a966d09c86c73386de531a121b21dd83232/sdk/core/azure-core/azure/core/rest/_rest_py3.py#L151-L158

I was able to solve it by changing the method to

```python
return self._data if self._data is not None else self._files
```

or by switching the order to

```python
return self._files or self._data
```

though the former seems more robust.

Stevenjin8:
[...]

PR #25540, opened 2022-08-03
[Core] empty content fix

Stevenjin8:
# Description

Before, `HttpRequest.content` would return `None` if we passed in `content=b""` when initializing the request object. 

### DIFF
--- a/sdk/core/azure-core/CHANGELOG.md
+++ b/sdk/core/azure-core/CHANGELOG.md
@@ -14,6 +14,7 @@ Azure-core is supported on Python 3.7 or later. For more details, please read ou
 
 - Fixed the bug that azure-core could not be imported under Python 3.11.0b3  #24928
 - `ContentDecodePolicy` can now correctly deserialize more JSON bodies with different mime types #22410
+- Fixed the bug where `HttpRequest.content` would return None when initialized with `content=b""` #25489
 
 ## 1.24.1 (2022-06-01)
 
--- a/sdk/core/azure-core/azure/core/rest/_rest_py3.py
+++ b/sdk/core/azure-core/azure/core/rest/_rest_py3.py
@@ -155,7 +155,9 @@ class HttpRequest(HttpRequestBackcompatMixin):
         :return: The request's content
         :rtype: any
         """
-        return self._data or self._files
+        if self._data is not None:
+            return self._data
+        return self._files
 
     def __repr__(self) -> str:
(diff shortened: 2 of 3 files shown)


sdk/core/azure-core/CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ Azure-core is supported on Python 3.7 or later. For more details, please read ou
 
 - Fixed the bug that azure-core could not be imported under Python 3.11.0b3  #24928
 - `ContentDecodePolicy` can now correctly deserialize more JSON bodies with different mime types #22410
+- Fixed the bug where `HttpRequest.content` would return None when initialized with `content=b""` #25489
```
Stevenjin8, line 17:
```suggestion
- Fixed the bug where `HttpRequest.content` would return `None` when initialized with `content=b""`#25489
```

iscai-msft (approved):
thanks @Stevenjin8

lmazuel (changes requested):
I'm not sure we want to make a semantic difference between `None` and empty body, those are technically the same thing in the HTTP spec. If we go that path. more trouble may come.
I'm not against the PR, but we shouldn't let the fast that this is a short PR let us think this is an easy problem, this should be discussed further.

johanste:
status code and content-type play into this. Status code 204, no content. Status 200, content-type application/octet-stream, content-length 0 is an empty `bytes`. Many azure services get this wrong, though.

Stevenjin8:
I should have added more context to this PR. This PR is a fix for #25489 where a customer was having issues uploading empty blobs to container storage. The issue arises when trying to computer the MD5 hash of a request with an empty body:

https://github.com/Azure/azure-sdk-for-python/blob/50d8cab2e7e6a57592e864b14798a5e2aa7306d4/sdk/storage/azure-storage-blob/azure/storage/blob/_shared/policies.py#L352-L371

Since `None` is not a readable file nor a bytes object, this above method raises a value error. An alternative method to doing this would be to update the `get_content_md5` method to have something along the lines of

```python
data = data or b""
```

Stevenjin8:
@johanste 

> status code and content-type play into this. Status code 204, no content. Status 200, content-type application/octet-stream, content-length 0 is an empty `bytes`. Many azure services get this wrong, though.

@johanste Not sure how status code plays into this since this PR is about requests, not responses.

Stevenjin8:
#25489 was resolved by #25655, so this PR is less of a priority, but still worth considering

PR #25454, opened 2022-07-28
Refactor Build, Versioning into `azure-sdk-tools`

scbedd:
This PR...

- [x] Refactors scripts present under `eng/versioning` to namespace `ci_tools.versioning` within the azure-sdk-tools package. All Integration points in CI are flipped over to consume entrypoints defined in the setup.py.
- [x] Refactors build_packages.py and build_package.py into single module `ci_tools.build`. This function `create_package` within this module is used wherever packages are built. 
- [x] Moves anything related to `setup.py` parsing into `ci_tools.parsing`. More specific, allows users access to `ParsedSetup` which has a classmethod `from_path` that can be aimed at any setup.py. Instead of a crazy tuple that we can get lost in the sauce in, it's just strongly typed members of a class now.
- [x] Moves anything related to processing of requirements into `ci_tools.functions`. 
- [x] Type and doc comment all the moved common functionality

This will be the most painful of the refactoring effort, but we will be able to make iterative fixes quite quickly once this one is in.

Due to the relative risk. I am delaying this merge until post-August-release-week. This is a significant improvement, but nothing that can't wait for a week.

Contributes to #23146

PR #25542, opened 2022-08-04
Resolve Failing Storage Builds

### LINKED ISSUE #25490: Storage Nightly Alpha Build Failing on unretrievable dependency
[Build Here](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=1746022&view=logs&j=3dc8fd7e-4368-5a92-293e-d53cefc8c4b3&t=76d9b154-b861-59e4-d8e3-dfd06fc0fc10)

`azure-storage-file-datalake` depends on `azure-storage-blob`. It has bumped it's dependency on `azure-storage-blob` to an as-yet-unreleased version. Now the nightly alpha build is crashing because it can't retrieve the correct version.

Given that this is supposed to be a supported scenario, I need to run down why this is failing.

@jalauzon-msft FYI. You're not blocked for release but nightly build failing is not an acceptable state.

scbedd:
Resolves #25490 

Friday of last week I noticed that azure-storage-file-datalake was having some issues on scheduled builds. The error was during installation, throwing out the fact that it couldn't resolve a version of `azure-storage-blob` that meats the requirements for `azure-storage-file-datalake`.

There were two main problems at play here:

1. We need to auth to the dev feed during the nightly build, if we want to actually talk to it! We weren't prior to this PR. This was causing any packages that have an unreleased dependency to fail **building the artifact or docs**.
2. We had an actual bug causing us to not retrieve from the dev feed even if we WERE auth-ing to the feed properly.

Issue for 2) is now properly resolved. Will definitely need to mirror this patch in the [refactor PR](https://github.com/Azure/azure-sdk-for-python/pull/25454) 

The core of the issue. We need up update the specifier to pull in an alpha package when the version is NOT on pypi. To do that, we do a string replacement operation. 

`azure-storage-blob<13.0.0,>=12.14.0b1` -> `azure-storage-blob<13.0.0,>=12.14.0ab1`

Notice the `ab1` that we were inserting? That's straight up wrong. I have implemented the regex to match [PEP440](https://peps.python.org/pep-0440/#pre-releases)

The incorrect prerelease section`ab1` is what was breaking pip's resolution.

### DIFF
--- a/eng/tox/sanitize_setup.py
+++ b/eng/tox/sanitize_setup.py
@@ -9,6 +9,7 @@
 
 import argparse
 import sys
+import re
 import os
 import logging
 import glob
@@ -109,7 +110,9 @@ def process_requires(setup_py_path):
             version = get_version(pkg_name)
             base_version = get_base_version(pkg_name)
             logging.info("Updating version {0} in requirement {1} to dev build version".format(version, old_req))
-            new_req = old_req.replace(version, "{}{}".format(base_version, DEV_BUILD_IDENTIFIER))
+            # to properly replace the version, we must replace the entire version, not a partial piece of it
+            rx = r'{}(((a|b|rc)\d+)?(\.post\d+)?)?'.format(base_version)
+            new_req = re.sub(rx, "{}{}".format(base_version, DEV_BUILD_IDENTIFIER), str(req), flags=re.IGNORECASE)
             logging.info("New requirement for package {0}: {1}".format(pkg_name, new_req))
             requirement_to_update[old_req] = new_req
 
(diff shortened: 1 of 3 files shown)


mccoyp (approved):
The regex is a little hairy, but it looks correct to me. The dev feed auth bit is more straightforward and makes sense 🙂

eng/tox/sanitize_setup.py
```diff
@@ -109,7 +110,9 @@ def process_requires(setup_py_path):
             version = get_version(pkg_name)
             base_version = get_base_version(pkg_name)
             logging.info("Updating version {0} in requirement {1} to dev build version".format(version, old_req))
-            new_req = old_req.replace(version, "{}{}".format(base_version, DEV_BUILD_IDENTIFIER))
+            # to properly replace the version, we must replace the entire version, not a partial piece of it
+            rx = r'{}(((a|b|rc)\d+)?(\.post\d+)?)?'.format(base_version)
```
semick-dev, line 114:
```suggestion
            #       {} =             <must have a base version>
            #       (                <optionally, we have a prerelease version>
            #        (               <we must have 0 or 1 prerelease versions>
            #            (a|b|rc)    <we must have a prelease identifier>
            #            \d+         <followed by a number of digits N>
            #        )?
            #        (               <optionally, we have a postrelease version>
            #            \.post      <which is ".post">
            #            \d+         <followed by number of digits N>
            #        )?
            #       )?
            rx = r'{}(((a|b|rc)\d+)?(\.post\d+)?)?'.format(base_version)
```

PR #25566, opened 2022-08-04
[Tables] Drop Python 3.6

YalinLi0312:


xiangyan99 (commented):
Please bump the minor version

sdk/tables/azure-data-tables/README.md
```diff
@@ -313,7 +314,7 @@ Get started with our [Table samples][tables_samples].
 Several Azure Tables Python SDK samples are available to you in the SDK's GitHub repository. These samples provide example code for additional scenarios commonly encountered while working with Tables.
 
 ### Common Scenarios
-These code samples show common scenario operations with the Azure Tables client library. The async versions of the samples (the python sample files appended with _async) show asynchronous operations with Tables and require Python 3.6 or later.
+These code samples show common scenario operations with the Azure Tables client library. The async versions of the samples (the python sample files appended with _async) show asynchronous operations with Tables and require Python 3.7 or later.
```
mccoyp, line 317: To make future maintenance easier, we can remove the bit about requiring a certain version of Python for async samples. Since we dropped 2.7 support, users will always be on a version of Python that support async operations.
```suggestion
These code samples show common scenario operations with the Azure Tables client library. The async versions of the samples (the python sample files appended with _async) show asynchronous operations.
```

mccoyp (commented):
To fix the `Analyze` step failure, you'll just need to update `_version.py` to be 12.5.0. For the `mindependency` failures, the `azure-mgmt-cosmosdb` dev requirement is [causing issues](https://dev.azure.com/azure-sdk/public/_build/results?buildId=1778524&view=logs&j=4fa73ede-de59-50b0-851d-db96582ea108&t=813eb0cd-e8db-59f1-2e13-b4deadd36a36&l=1276) again... You can either update your minimum `msrest` version to 0.7.1, or pin your [`azure-mgmt-cosmosdb` requirement](https://github.com/Azure/azure-sdk-for-python/blob/8e9d731ddec4ba9dc1ac4f1b7bf5f94764574430/sdk/tables/azure-data-tables/dev_requirements.txt#L3) to an earlier version, like 7.0.0
``` diff
- -e ../../cosmos/azure-mgmt-cosmosdb
../../core/azure-core
../../nspkg/azure-data-nspkg
aiohttp>=3.0
azure-identity
+ azure-mgmt-cosmosdb==7.0.0
```

@xiangyan99, do you have an opinion on whether or not to bump the dependency?

YalinLi0312:
There's a similar [PR](https://github.com/Azure/azure-sdk-for-python/pull/25571) for ACR which is in the same situation, same question as McCoy. @xiangyan99

xiangyan99 (commented):
Updating _version.py is also needed. :)

mccoyp:
Since this is for a new minor version, I'd recommend going ahead and bumping your minimum `msrest` dependency to 0.7.1 🙂

mccoyp (approved):
LGTM!

PR #25574, opened 2022-08-05
[Storage] Migrate test_container.py

vincenttran-msft:
This PR migrates `test_container.py` and `test_container_async.py` to use test proxy. Still need to address #25623 once the batch sanitizer resolution has been reached.


sdk/storage/azure-storage-blob/tests/test_container.py
```diff
@@ -1772,7 +2090,10 @@ def test_standard_blob_tier_set_tiers_with_sas(self, storage_account_name, stora
     # once we have premium tests, still we don't want to test Py 2.7
     # @pytest.mark.skipif(sys.version_info < (3, 0), reason="Batch not supported on Python 2.7")
     @BlobPreparer()
-    def test_premium_tier_set_tier_api_batch(self, storage_account_name, storage_account_key):
+    def test_premium_tier_set_tier_api_batch(self, **kwargs):
```
jalauzon-msft, line 2093: Did you try anything with this test? Looks like we need a Premium Page Blob account that supports batch? That just may not be a thing :)

sdk/storage/azure-storage-blob/tests/test_container_async.py
```diff
@@ -5,64 +5,48 @@
 # Licensed under the MIT License. See License.txt in the project root for
```
jalauzon-msft, line 5: nit: (Can't leave comment on part of file not in diff) but can you remove the `# coding: utf-8` at the top? I don't think we need that anymore and I've been removing it.

sdk/storage/azure-storage-blob/tests/conftest.py
```diff
@@ -13,7 +13,8 @@
     add_header_regex_sanitizer,
     add_oauth_response_sanitizer,
     add_uri_regex_sanitizer,
-    test_proxy
+    test_proxy,
+    add_body_key_sanitizer
```
jalauzon-msft, line 17: I think you can undo this now?

jalauzon-msft (approved):
Again, great work on this tough file!

sdk/storage/azure-storage-blob/tests/test_container.py
vincenttran-msft: Yeah I have no idea what "premium" means in this context 😆. I remember trying a bunch of different things when initially triaging this in our 'triage all skipped tests' item from a couple weeks / month(?) ago, so just left it skipped for now (it's in the backlog!! 🤣)

sdk/storage/azure-storage-blob/tests/test_container_async.py
vincenttran-msft: That I can do!

sdk/storage/azure-storage-blob/tests/conftest.py
vincenttran-msft: Done!

Issue #25586, opened 2022-08-05
[Storage] Test Proxy Migrate DateTime (interim)


vincenttran-msft:
For the test proxy migration, there is a requirement of using `variables` when using `datetime` in order to allow recordings to work properly. This behavior will become a util, so after that change is made, we then need to go back and migrate the following files to remove the copy paste and use the util:

- [ ] `test_append_blob`
- [ ] `test_blob_access_conditions`
- [ ] `test_blob_api_version`
- [ ] `test_blob_client`
- [ ] `test_blob_service_properties`
- [ ] `test_blob_storage_account`
- [ ] `test_block_blob`
- [ ] `test_container`
- [ ] `test_page_blob`
**Note: Some files may not use `datetime` (this list was built by seeing what tests have `.json` vs .`yaml`, so peek the file and ctrl+f first)**

PR #25590, opened 2022-08-05
[Storage] Add datetime util

vincenttran-msft:
Moving out the helper to the utils so that we don't have to copy paste the helper everywhere 😄 

Issue tracking the list of tests that possibly use the copy pasted helper: https://github.com/Azure/azure-sdk-for-python/issues/25586


jalauzon-msft (approved):
Thanks for tackling this!

PR #25606, opened 2022-08-08
[formrecognizer] Doc updates

catalinaperalta:
Fix docs on `begin_build_model()` and run `black` on the code.

### DIFF
--- a/sdk/formrecognizer/azure-ai-formrecognizer/azure/ai/formrecognizer/_polling.py
+++ b/sdk/formrecognizer/azure-ai-formrecognizer/azure/ai/formrecognizer/_polling.py
@@ -71,22 +71,16 @@ class DocumentModelAdministrationLROPoller(Protocol[PollingReturnType]):
     ) -> PollingReturnType:
         ...
 
-    def wait(  # pylint: disable=no-self-use, unused-argument
-        self, timeout: Optional[float] = None
-    ) -> None:
+    def wait(self, timeout: Optional[float] = None) -> None:  # pylint: disable=no-self-use, unused-argument
         ...
 
     def done(self) -> bool:  # pylint: disable=no-self-use
         ...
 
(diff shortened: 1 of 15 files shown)


swathipil (approved):

sdk/formrecognizer/azure-ai-formrecognizer/azure/ai/formrecognizer/_polling.py
```diff
@@ -101,11 +113,13 @@ def details(self) -> Mapping[str, Any]:
             created_on = datetime.datetime.strptime(created_on, "%Y-%m-%dT%H:%M:%SZ")
         last_updated_on = self._current_body.get("lastUpdatedDateTime", None)
         if last_updated_on:
-            last_updated_on = datetime.datetime.strptime(last_updated_on, "%Y-%m-%dT%H:%M:%SZ")
+            last_updated_on = datetime.datetime.strptime(
```
mccoyp, line 116: I'm surprised that shorter lines like this got reformatted -- did you run `black` with `-l 120`?

catalinaperalta: Hmmm I did not let me check if that changes it.

mccoyp: For context, `black` defaults to enforcing a max line length of 90 characters, but we usually enforce a max of 120 in the SDK. Ultimately shorter is better than longer, but worth double-checking 🙂

catalinaperalta: Ok ran with -l 120 and it fixed spacing due to line length that was applied during several of my previous black runs on this code. I shall try to remember to always run it with `-l 120`. Thanks McCoy!

mccoyp (approved):
LGTM!

PR #25629, opened 2022-08-09
[Storage] Setup datalake test proxy and migrate first two files

jalauzon-msft:
This change sets up `conftest` in datalake for test proxy as well as migrates `test_cpk` and `test_datalake_service_client` to use test proxy.


vincenttran-msft (approved):
LGTM! No surprises on these files, lucky man 😄

PR #25630, opened 2022-08-09
[Pyamqp] Intial TODOS Clean Up

kashifkhan:
* `ws.close` on async does close the socket in the library -> [link](https://github.com/websocket-client/websocket-client/blob/61171591b08ee031e02cc6cb129952259062f502/websocket/_core.py#L4580)
* I kept the py2 comment in `_transport.py` about `from os import set_cloexec`. Looks like its not in py3 ( yet ) and that import will throw an `ImportError` across Windows, Mac & Linux.
* Rejecting `links` on a `ValueError` in `_incoming_attach`. Is that the proper way?

### DIFF
--- a/sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_transport.py
+++ b/sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_transport.py
@@ -389,7 +389,7 @@ class _AbstractTransport(object):
             self.sock = None
         self.connected = False
 
-    def read(self, verify_frame_type=0, **kwargs):  # TODO: verify frame type?
+    def read(self, verify_frame_type=0, **kwargs):
         read = self._read
         read_frame_buffer = BytesIO()
         try:
--- a/sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_transport.py
+++ b/sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_transport.py
@@ -446,7 +446,6 @@ class _AbstractTransport(object):
                 decoded = decode_empty_frame(header)
             else:
                 decoded = decode_frame(payload)
-            # TODO: Catch decode error and return amqp:decode-error
             return channel, decoded
         except (socket.timeout, TimeoutError):
             return None, None
--- a/sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_transport.py
+++ b/sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_transport.py
@@ -508,14 +508,7 @@ class SSLTransport(_AbstractTransport):
         # Setup the right SSL version; default to optimal versions across
         # ssl implementations
         if ssl_version is None:
-            # older versions of python 2.7 and python 2.6 do not have the
-            # ssl.PROTOCOL_TLS defined the equivalent is ssl.PROTOCOL_SSLv23
-            # we default to PROTOCOL_TLS and fallback to PROTOCOL_SSLv23
-            # TODO: Drop this once we drop Python 2.7 support
-            if hasattr(ssl, 'PROTOCOL_TLS'):
-                ssl_version = ssl.PROTOCOL_TLS
-            else:
-                ssl_version = ssl.PROTOCOL_SSLv23
+            ssl_version = ssl.PROTOCOL_TLS
 
         opts = {
             'sock': sock,
--- a/sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_transport.py
+++ b/sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_transport.py
@@ -384,7 +384,7 @@ class _AbstractTransport(object):
             except Exception as exc:
                 # TODO: shutdown could raise OSError, Transport endpoint is not connected if the endpoint is already
                 #  disconnected. can we safely ignore the errors since the close operation is initiated by us.
-                _LOGGER.info("An error occurred when shutting down the socket: %r", exc)
+                _LOGGER.info("Transport endpoint is already disconnected: %r", exc)
             self.sock.close()
             self.sock = None
         self.connected = False
--- a/sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_session_async.py
+++ b/sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_session_async.py
@@ -162,7 +162,8 @@ class Session(object):
             _LOGGER.info("<- %r", EndFrame(*frame), extra=self.network_trace_params)
         if self.state not in [SessionState.END_RCVD, SessionState.END_SENT, SessionState.DISCARDING]:
             await self._set_state(SessionState.END_RCVD)
-            # TODO: Clean up all links
+            for _, link in self.links.items():
+                await link.detach()
             await self._outgoing_end()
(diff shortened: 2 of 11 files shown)


l0lawrence (approved):
lgtm

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_transport.py
```diff
@@ -384,12 +384,12 @@ def close(self):
             except Exception as exc:
                 # TODO: shutdown could raise OSError, Transport endpoint is not connected if the endpoint is already
```
swathipil, line 385: checking that we're intentionally leaving the TODO here?

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_transport_async.py
```diff
@@ -92,13 +92,12 @@ async def receive_frame(self, *args, **kwargs):
                 decoded = decode_empty_frame(header)
             else:
                 decoded = decode_frame(payload)
-            # TODO: Catch decode error and return amqp:decode-error
             #_LOGGER.info("ICH%d <- %r", channel, decoded)
```
swathipil, line 95: not part of the TODO, but do we want to uncomment or delete this line?

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_transport_async.py
```diff
@@ -483,11 +482,8 @@ async def _read(self, n, buffer=None, **kwargs): # pylint: disable=unused-argume
 
     def close(self):
         """Do any preliminary work in shutting down the connection."""
-        # TODO: async close doesn't:
-        # 1) shutdown socket and close. --> self.sock.shutdown(socket.SHUT_RDWR) and self.sock.close()
```
swathipil, line 487: checking that we don't need to add `self.ws.shutdown(socket.SHUT_RDWR)` before the ws.close call?

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/session.py
```diff
@@ -181,7 +182,8 @@ def _incoming_attach(self, frame):
             self._output_handles[outgoing_handle] = new_link
             self._input_handles[frame[1]] = new_link
         except ValueError:
-            pass  # TODO: Reject link
+            # Reject Link
+            new_link.detach()
```
swathipil, line 186: it looks like this new link is being defined in the except KeyError block, but being detached here where it wouldn't exist unless I'm missing something. Do we want to indent this and do another try/except in the KeyError block? Or, more likely, this will be a ValueError from the try on line 172, so should this rejection be handled some other way?

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_transport_async.py
swathipil, line 487: checking that self.ws.shutdown(socket.SHUT_RDWR) doesn't need to be called before the ws.close() call below?

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_transport_async.py
swathipil, line 95: not part of the TODO, but do we want to uncomment or delete this line?

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_transport.py
swathipil, line 385: checking that we're intentionally leaving the TODO here?

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_session_async.py
```diff
@@ -184,7 +185,8 @@ async def _incoming_attach(self, frame):
             self._output_handles[outgoing_handle] = new_link
             self._input_handles[frame[1]] = new_link
         except ValueError:
-            pass  # TODO: Reject link
+            # Reject Link
+            await new_link.detach()
```
swathipil, line 189: it looks like this new link is being defined in the except KeyError block, but being detached here where it wouldn't exist unless I'm missing something.

swathipil (commented):
In general, I was thinking that we could keep a separate file for outlining negative tests/any tests that we'll want to add in for testing our TODOS. (maybe under something like `tests/todo_tests.py`?) And in each TODO PR, we can update this file with a comment of the test/pseudocode for the test/something similar.

Since the actual tests will be added to main, but all of our TODO changes will be in pyamqp, I think it would be helpful to be able to couple these TODO test outlines with their corresponding changes. This way, we can clearly track all the tests we need to add, rather than retrospectively figuring out which TODO changes need to be tested and how exactly to test them.

Thoughts?

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_transport.py
kashifkhan: Yes intentionally leaving this one as that particular OSError occurs very often and in code using that flow they just catch the exception and pass.

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_transport_async.py
kashifkhan: yep, the `websocket-client` library does exactly that when you can call wsl.close() ... [link](https://github.com/websocket-client/websocket-client/blob/61171591b08ee031e02cc6cb129952259062f502/websocket/_core.py#L497) to library code

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_transport_async.py
kashifkhan: Ill uncomment it for now and we can review during the 'logging review' portion of the work

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/session.py
kashifkhan: good catch !! now its rejecting the right link :)

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_session_async.py
kashifkhan: good catch !! now its rejecting the right link :)

kashifkhan:
> In general, I was thinking that we could keep a separate file for outlining negative tests/any tests that we'll want to add in for testing our TODOS. (maybe under something like `tests/todo_tests.py`?) And in each TODO PR, we can update this file with a comment of the test/pseudocode for the test/something similar.
> 
> Since the actual tests will be added to main, but all of our TODO changes will be in pyamqp, I think it would be helpful to be able to couple these TODO test outlines with their corresponding changes. This way, we can clearly track all the tests we need to add, rather than retrospectively figuring out which TODO changes need to be tested and how exactly to test them.
> 
> Thoughts?

I like the idea, we can include this in either the PR that fixes those groupings of TODOs or in the Onenote with a linkout to the PR

swathipil (approved):

PR #25633, opened 2022-08-09
Add storage configuration options

lzchen:


sdk/monitor/azure-monitor-opentelemetry-exporter/azure/monitor/opentelemetry/exporter/export/_base.py
```diff
@@ -119,10 +139,10 @@ def _transmit(self, envelopes: List[TelemetryItem]) -> ExportResult:
                             error.message,
                             envelopes[error.index] if error.index is not None else "",
                         )
-                if resend_envelopes:
+                if self.storage and resend_envelopes:
                     envelopes_to_store = [x.as_dict()
                                           for x in resend_envelopes]
-                    self.storage.put(envelopes_to_store)
+                    self.storage.put(envelopes_to_store, 0)
```
jeremydvoss, line 145: What's the difference from adding 0?

sdk/monitor/azure-monitor-opentelemetry-exporter/setup.py
```diff
@@ -82,8 +82,8 @@
     install_requires=[
         "azure-core<2.0.0,>=1.23.0",
         "msrest>=0.6.10",
-        "opentelemetry-api<2.0.0,>=1.12.0rc1,!=1.10a0",
-        "opentelemetry-sdk<2.0.0,>=1.12.0rc1,!=1.10a0",
```
jeremydvoss, line 86: Just curious, what did the <2.0.0,> mean?

sdk/monitor/azure-monitor-opentelemetry-exporter/azure/monitor/opentelemetry/exporter/export/_base.py
lzchen: Adding 0 just means we can attempt to export the failed telemetry immediately, as opposed to ones with the default value (60s). We want the ones here to be attempting retry immediately because usually the ones that result from a returned `206` are valid and something jus happened coincidentally (perhaps in the backend) that rejected only partial envelopes.

sdk/monitor/azure-monitor-opentelemetry-exporter/setup.py
lzchen: It means we don't want to install 2.* versions of the package (because there might be breaking changes). See https://semver.org/#why-use-semantic-versioning

sdk/monitor/azure-monitor-opentelemetry-exporter/azure/monitor/opentelemetry/exporter/export/_base.py
jeremydvoss: Thanks

jeremydvoss (approved):

PR #25638, opened 2022-08-10
[Test Proxy] Add `environment_variables` fixture

mccoyp:
# Description

This introduces an `environment_variables` fixture that emulates today's EnvironmentVariableLoader. This fixture provides an interface for easily fetching and sanitizing environment variables, and can either be used on a test-by-test request basis, or on a module/session basis to prepare test variables for all tests.

For example, a test file could do something like the following to sanitize test variables while making them available to all tests, without duplicating code:
```python
import pytest
from devtools_testutils import AzureRecordedTestCase

test_variables = {}

def prepare_test_variables(environment_variables):
    sanitizings = {
        "SERVICE_VARIABLE_A": "fake_value_a",
        "SERVICE_VARIABLE_B": "fake_value_b",
        "SERVICE_VARIABLE_C": "fake_value_c",
    }
    test_variables = environment_variables.sanitize_batch(sanitizings)

@pytest.mark.usefixtures("recorded_test")
class TestClass(AzureRecordedTestCase):
    def test_method(self):
        result = do_thing(test_variables["SERVICE_VARIABLE_A"])
```

  - Tested locally with already-migrated libraries.

### DIFF
--- a/tools/azure-sdk-tools/devtools_testutils/proxy_fixtures.py
+++ b/tools/azure-sdk-tools/devtools_testutils/proxy_fixtures.py
@@ -5,6 +5,7 @@
 # --------------------------------------------------------------------------
 from inspect import iscoroutinefunction
 import logging
+import os
 from typing import TYPE_CHECKING
 import urllib.parse as url_parse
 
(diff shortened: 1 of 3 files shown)


tools/azure-sdk-tools/devtools_testutils/proxy_fixtures.py
```diff
@@ -16,15 +17,102 @@
...
+        self._fake_values[variable] = value
+        real_value = os.getenv(variable)
+        if real_value:
+            add_general_regex_sanitizer(regex=real_value, value=value)
+        else:
+            _LOGGER.info(f"No value for {variable} was found, so a sanitizer could not be registered for the variable.")
+
+        return real_value if is_live() else value
+
+    def sanitize_batch(self, variables: "Dict[str, str]") -> "Dict[str, str]":
```
scbedd, line 54: I am not going to block the PR on this because I think it'll actually work, but I'm curious.

[Our typehint docs](https://github.com/Azure/azure-sdk-for-python/blob/main/doc/dev/docstring_typehint.md#type-hints-for-python-35) show `Dict[str, str]` instead of `"Dict[str, str]"`. That's also the style that' I've been following in my code. What makes you choose the `surrounded by quotes`? There's gotta be a reason and I'd like to understand it!

scbedd: Oh I see. Because you have the conditional import above. Due to that, you HAVE to leave it string so it can resolve during type checking. Is that correct?

tools/azure-sdk-tools/devtools_testutils/proxy_fixtures.py
```diff
@@ -16,15 +17,102 @@
...
+
+@pytest.fixture
+def environment_variables(test_proxy: None) -> EnvironmentVariableSanitizer:
+    """Fixture that returns an EnvironmentVariableSanitizer for convenient environment variable fetching and sanitizing.
+
+    :param test_proxy: The fixture responsible for starting up the test proxy server.
+    :type test_proxy: None
+
+    :returns: An EnvironmentVariableSanitizer object. Calling:
+        - `sanitize(a, b)` will sanitize the value of environnment variable `a` with value `b`
```
scbedd, line 106:
```suggestion
        - `sanitize(a, b)` will sanitize the value of environment variable `a` with value `b`
```

tools/azure-sdk-tools/devtools_testutils/proxy_fixtures.py
```diff
@@ -16,15 +17,102 @@
...
+        real_value = os.getenv(variable)
+        if real_value:
+            add_general_regex_sanitizer(regex=real_value, value=value)
+        else:
+            _LOGGER.info(f"No value for {variable} was found, so a sanitizer could not be registered for the variable.")
+
+        return real_value if is_live() else value
+
+    def sanitize_batch(self, variables: "Dict[str, str]") -> "Dict[str, str]":
+        """Registers sanitizers that replace the values of multiple envrionment variables with the provided values.
```
scbedd, line 55:
```suggestion
        """Registers sanitizers that replace the values of multiple environment variables with the provided values.
```

scbedd (approved):

tools/azure-sdk-tools/devtools_testutils/proxy_fixtures.py
mccoyp: Yep, that's right! To be honest, I think my conditional import is outdated -- I've seen libraries starting to import `typing` types regardless of `TYPE_CHECKING`. But if the import is conditional, you need the quotes to avoid runtime errors

PR #25641, opened 2022-08-10
[pipeline] some features and fixes

msyyc:
- Release Helper: add `HoldOn` policy
- Auto Release: use random number when inserting `share_requirements.txt` to avoid merge confict
- Changelog: 
(1) updated name of changelog item with guidance: https://azure.github.io/azure-sdk/policies_releases.html#change-logs
(2) Fix for fake breaking change when package structure changed to multiapi

test link:
- changelog: https://github.com/openapi-env-test/azure-rest-api-specs/pull/3566
- auto-release: https://github.com/Azure/azure-sdk-for-python/pull/25642, https://github.com/Azure/azure-sdk-for-python/pull/25643

### DIFF
--- a/tools/azure-sdk-tools/packaging_tools/code_report.py
+++ b/tools/azure-sdk-tools/packaging_tools/code_report.py
@@ -165,6 +165,7 @@ def merge_report(report_paths):
         merged_report["models"]["exceptions"].update(report_json["models"]["exceptions"])
         merged_report["models"]["models"].update(report_json["models"]["models"])
         merged_report["operations"].update(report_json["operations"])
+        merged_report["client"] = report_json["client"]
     return merged_report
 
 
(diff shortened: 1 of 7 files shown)


tools/azure-sdk-tools/packaging_tools/code_report.py
```diff
@@ -165,6 +165,7 @@ def merge_report(report_paths):
         merged_report["models"]["exceptions"].update(report_json["models"]["exceptions"])
         merged_report["models"]["models"].update(report_json["models"]["models"])
         merged_report["operations"].update(report_json["operations"])
+        merged_report["client"] = report_json["client"]
```
changlong-liu, line 168: Is it correct to assign 'report_json["client"]' directly to merged_report["client"], not something like below?
~~~
merged_report["client"].update(report_json["client"])
~~~

msyyc: merged_report["client"] is string and each sub-folder shall have the same client name, so we just need to get one

changlong-liu (approved):

Issue #25640, opened 2022-08-10
Result of `query_entities` of `TableClient` cannot be returned in `with` keyword successfully


matt-mi:
- **Package Name**: azure-data-tables
- **Package Version**: 12.4.0
- **Operating System**: macos
- **Python Version**: 3.9.13

**Describe the bug**
Result of `query_entities` of `TableClient` cannot be returned in `with` keyword

**To Reproduce**
1. create a record in table store. E.g., a row has value `cluster4` under `RowKey`.
2. Query the record from table store by following
```
def get_cluster():
    cluster_name_filter = "RowKey eq 'cluster4'"
    with TableClient(endpoint=sta_endpoint, table_name=table_name, credential=credential) as table_client:
        return table_client.query_entities(query_filter=cluster_name_filter)

results = get_cluster()
for i in results:
    print(i['RowKey'])
```
3. Execute code snippet above and got exception below
```
(.venv) ➜  /cronjob/.venv/bin/python scheduler.py
Traceback (most recent call last):
  File "/cronjob/scheduler/scheduler.py", line 135, in <module>
    for i in results:
  File "/cronjob/.venv/lib/python3.9/site-packages/azure/core/paging.py", line 128, in __next__
    return next(self._page_iterator)
  File "/cronjob/.venv/lib/python3.9/site-packages/azure/core/paging.py", line 76, in __next__
    self._response = self._get_next(self.continuation_token)
  File "/cronjob/.venv/lib/python3.9/site-packages/azure/data/tables/_models.py", line 363, in _get_next_cb
    return self._command(
  File "/cronjob/.venv/lib/python3.9/site-packages/azure/data/tables/_generated/operations/_table_operations.py", line 380, in query_entities
    pipeline_response = self._client._pipeline.run(request, stream=False, **kwargs)
  File "/cronjob/.venv/lib/python3.9/site-packages/azure/core/pipeline/_base.py", line 211, in run
    return first_node.send(pipeline_request)  # type: ignore
  File "/cronjob/.venv/lib/python3.9/site-packages/azure/core/pipeline/_base.py", line 71, in send
    response = self.next.send(request)
  File "/cronjob/.venv/lib/python3.9/site-packages/azure/core/pipeline/_base.py", line 71, in send
    response = self.next.send(request)
  File "/cronjob/.venv/lib/python3.9/site-packages/azure/core/pipeline/_base.py", line 71, in send
    response = self.next.send(request)
[...]

PR #25651, opened 2022-08-10
preventing opening transport if it is closed

xiangyan99:
https://github.com/Azure/azure-sdk-for-python/issues/25640#issuecomment-1210162459


Stevenjin8:
turns a that a few of these are needed for typing

Stevenjin8:
@xiangyan99 sorry i was looking at the wrong pr

sdk/core/azure-core/tests/async_tests/test_basic_transport_async.py
```diff
@@ -989,3 +989,9 @@ def test_aiohttp_loop():
     loop = asyncio.get_event_loop()
     with pytest.raises(ValueError):
         transport = AioHttpTransport(loop=loop)
+
+def test_open_closed_aiohttp_transport():
+    transport = AioHttpTransport()
+    transport.close()
+    transport.open()
+    assert not transport.session
```
annatisch, line 997: What would this change mean for a client that attempts to send?
From what I can see in the transport code, if I attempted `transport.send()` after calling `transport.close()` (which I think is what a Client built on top of this would likely do) - then I think it will just raise an AttributeError on line 198.

annatisch (commented):
As I understand it, none of our transports are "reopenable" right? If you attempt `client.get_something()` after calling close/exited context, it will error - is that currently correct?
If so - what's the error experience?
I think this particular 'gap' only happens when a user never actually used the client (i.e. never opened the connection) before calling close/exited context. Given that the general behaviour of using a client after it's been closed should raise an error, I would consider plugging this gap to be a bug fix, but I guess it could potentially be breaking?

sdk/core/azure-core/tests/async_tests/test_basic_transport_async.py
xiangyan99: This is consistent with current behavior if user calls

with client:
     client.do_sth() 


# now transport is closed
client.do_sth_else()

right?

PR #25655, opened 2022-08-10
None-check when computing md5

Stevenjin8:
# Description

Fix for #25489. Our md5 method was not considering the possibility that a request's content could be `None` instead of `b""`. Since from an HTTP perspective `None` and `b""` are the same, we just add a simple `None`-check,

### DIFF
--- a/sdk/storage/azure-storage-blob/CHANGELOG.md
+++ b/sdk/storage/azure-storage-blob/CHANGELOG.md
@@ -13,6 +13,7 @@ This version and all future versions will require Python 3.7+. Python 3.6 is no 
 - Removed dead retry meachism from async `azure.storage.blob.aio.StorageStreamDownloader`.
 - Updated exception catching of `azure.storage.blob.StorageStreamDownloader`'s retry mechanism.
 - Adjusted type hints for `upload_blob` and `StorageStreamDownloader.readall`.
+- Fixed bug where `BlobClient` couldn't upload empty blob with `validate_content=True`.
 
 ## 12.13.1 (2022-08-04)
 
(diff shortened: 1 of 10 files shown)


sdk/storage/azure-storage-blob/CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ This version and all future versions will require Python 3.7+. Python 3.6 is no
 - Removed dead retry meachism from async `azure.storage.blob.aio.StorageStreamDownloader`.
 - Updated exception catching of `azure.storage.blob.StorageStreamDownloader`'s retry mechanism.
 - Adjusted type hints for `upload_blob` and `StorageStreamDownloader.readall`.
+- Fixed bug where `BlobClient` couldn't upload empty blob with `validate_content=True`.
```
jalauzon-msft, line 16: How about: "Fixed a bug where uploading an empty blob via `upload_blob` would fail with `validate_content=True`"

sdk/storage/azure-storage-blob/tests/test_blob_client.py
```diff
@@ -711,4 +711,24 @@ def test_closing_pipeline_client_simple(self, **kwargs):
                 self.account_url(storage_account_name, "blob"), credential=storage_account_key, container_name='foo', blob_name='bar')
             service.close()
 
+    @BlobPreparer()
+    def test_validate_empty_blob(self, **kwargs):
```
jalauzon-msft, line 715: Three asks for the test:
- Would you mind moving this test to `test_common_blob.py`? This file is more for testing the BlobClient rather than upload. I think you'll have to change the test a little since that file has not yet migrated to test proxy.
- Can you record the test?
- Can you add the same test to `test_common_blob_async.py` too? We usually duplicate all of tests for sync and async. Obviously, you'll have to make a couple updates to support async. Please also record this one.

jalauzon-msft (commented):
Thanks for picking this up! Change looks good to me. Just some comments on the CHANGELOG and test. Let me know if you have any questions.

jalauzon-msft (approved):
Looks good, thanks again!

PR #25656, opened 2022-08-10
[Storage] Attempt to fix flaky read timeout retry test

jalauzon-msft:
The `test_retry_on_socket_timeout` test has been flaky in the Live test pipeline recently. I believe it was caused by attempting to have a read timeout on an operation that does not have a payload to read (Create Container). This changes the test to perform the retry test on a 5 KiB download which should be more consistent in getting a read timeout (hopefully).

Also cleaning up the recordings for these Live only tests.


vincenttran-msft (approved):
Be gone flaky tests!

PR #25658, opened 2022-08-10
[Storage] Migrate test_blob_tags.py to test proxy

vincenttran-msft:
This PR migrates `test_blob_tags.py` and `test_blob_tags_async.py` to test proxy 😄


sdk/storage/azure-storage-blob/tests/test_blob_tags.py
```diff
@@ -258,71 +302,87 @@ def test_start_copy_from_url_with_tags(self, storage_account_name, storage_accou
...
+        storage_account_key = kwargs.pop("storage_account_key")
+        variables = kwargs.pop('variables', {})
+
         self._setup(storage_account_name, storage_account_key)
         tags = {"tag1": "firsttag", "tag2": "secondtag", "tag3": "thirdtag"}
         source_blob = self.bsc.get_blob_client(self.container_name, self._get_blob_reference())
         source_blob.upload_blob(b'Hello World', overwrite=True, tags=tags)
 
-        source_sas = generate_blob_sas(
+        expiry_time = self.get_datetime_variable(variables, 'expiry_time', datetime.utcnow() + timedelta(hours=1))
```
jalauzon-msft, line 331: For SAS, you shouldn't actually need to create a variable for dates. The `self.generate_sas` takes care of that for you somehow (not with variables but somehow)

Same with other SAS below and async SAS usage.

sdk/storage/azure-storage-blob/tests/test_blob_tags.py
```diff
@@ -98,33 +103,40 @@ def test_set_blob_tags(self, storage_account_name, storage_account_key):
...
         resp = blob_client.get_blob_tags(lease=lease)
 
-        self.assertIsNotNone(resp)
-        self.assertEqual(len(resp), 3)
+        assert resp is not None
+        assert len(resp) == 3
 
         blob_client.delete_blob(lease=lease)
 
     @pytest.mark.live_test_only
```
jalauzon-msft, line 134: Any idea why this is live only? I don't think it needs to be so can you try recording it?

Same with async version.

sdk/storage/azure-storage-blob/tests/test_blob_tags_async.py
```diff
@@ -339,13 +402,17 @@ async def test_list_blobs_returns_tags(self, storage_account_name, storage_accou
 
         #Assert
         async for blob in blob_list:
-            self.assertEqual(blob.tag_count, len(tags))
+            assert blob.tag_count == len(tags)
             for key, value in blob.tags.items():
-                self.assertEqual(tags[key], value)
+                assert tags[key] == value
 
     @pytest.mark.playback_test_only
```
jalauzon-msft, line 409: This probably doesn't need to be playback only? It's not in sync :)

jalauzon-msft (commented):
Thanks, looks good! Just a couple of minor comments.

sdk/storage/azure-storage-blob/tests/test_blob_tags.py
vincenttran-msft: `E           azure.core.exceptions.ServiceRequestError: <urllib3.connection.HTTPSConnection object at 0x0000023F39C954E0>: Failed to establish a new connection: [Errno 11001] getaddrinfo failed`
`ServiceRequestError` 😨

vincenttran-msft: Good catch! I originally tried but was failing in recorded (as always, it was because I forgot to put `recorded_by_proxy` so no recording was being generated 🤦

sdk/storage/azure-storage-blob/tests/test_blob_tags_async.py
vincenttran-msft: Good catch!

jalauzon-msft (approved):

PR #25661, opened 2022-08-11
[Storage] Migrate test_quick_query.py

vincenttran-msft:
This file migrates `test_quick_query.py` to test proxy 😄


sdk/storage/azure-storage-blob/tests/test_quick_query.py
```diff
@@ -100,8 +98,6 @@ def _teardown(self, bsc):
             except:
                 pass
 
-        return super(StorageQuickQueryTest, self).tearDown()
```
jalauzon-msft, line 103: Can you see if this `_teardown` method is being automatically called? (via a breakpoint or just seeing if the container is being deleted in Live mode). If it is not, can you change the name to `_tearDown` and then see?

This is mostly curiosity but if this isn't being called at all we can delete it all together. Though I do like a test file that cleans up its own containers! So, it would be nice to know if either of these names get automatically called so we can use them in the future.

sdk/storage/azure-storage-blob/tests/test_quick_query.py
```diff
@@ -82,7 +80,7 @@
 # ------------------------------------------------------------------------------
 
 
-class StorageQuickQueryTest(StorageTestCase):
+class TestStorageQuickQuery(StorageRecordedTestCase):
```
jalauzon-msft, line 83: No `async` file? Weird.

jalauzon-msft (approved):

sdk/storage/azure-storage-blob/tests/test_quick_query.py
vincenttran-msft: Will discuss at 1-1; but I have not tested et if this is automatically called, but even if it is, I think we lose this functionality as `StorageRecordedTestCase` does not have a `tearDown`, but it seems that `StorageTestCase` did (although to be fair, I didn't dig through `StorageRecordedTestCase` to see if it had an equivalent, I just tried to blindly call it)

sdk/storage/azure-storage-blob/tests/test_quick_query.py
vincenttran-msft: Yeah.. I guess not! 😆

sdk/storage/azure-storage-blob/tests/test_quick_query.py
vincenttran-msft: Update: `_teardown` does get called 😄 (lowercase d)

Issue #22297, opened 2022-01-01
Exception: TypeError: '<' not supported between instances of 'int' and 'NoneType' hit in download_to_stream


AdmiralGT:
- **Package Name**: azure-storage-blob
- **Package Version**: 12.9.0
- **Operating System**: Linux (via Azure App Service)
- **Python Version**: 3.8.6

**Describe the bug**
Originally raised at https://github.com/jschneier/django-storages/issues/1097, when using django-storages to provide a FileResponse, the following exception is hit.

```
2021-12-12T13:36:51.739891275Z Traceback (most recent call last):
2021-12-12T13:36:51.739908176Z   File "/opt/python/3.8.6/lib/python3.8/site-packages/gunicorn/workers/sync.py", line 136, in handle
2021-12-12T13:36:51.739920476Z     self.handle_request(listener, req, client, addr)
2021-12-12T13:36:51.739931676Z   File "/opt/python/3.8.6/lib/python3.8/site-packages/gunicorn/workers/sync.py", line 182, in handle_request
2021-12-12T13:36:51.739943076Z     resp.write_file(respiter)
2021-12-12T13:36:51.739953676Z   File "/opt/python/3.8.6/lib/python3.8/site-packages/gunicorn/http/wsgi.py", line 385, in write_file
2021-12-12T13:36:51.739964877Z     if not self.sendfile(respiter):
2021-12-12T13:36:51.739975477Z   File "/opt/python/3.8.6/lib/python3.8/site-packages/gunicorn/http/wsgi.py", line 355, in sendfile
2021-12-12T13:36:51.739986577Z     if not util.has_fileno(respiter.filelike):
2021-12-12T13:36:51.739997177Z   File "/opt/python/3.8.6/lib/python3.8/site-packages/gunicorn/util.py", line 569, in has_fileno
2021-12-12T13:36:51.740008278Z     if not hasattr(obj, "fileno"):
2021-12-12T13:36:51.740019178Z   File "/tmp/8d9bd73cfc52ab6/antenv/lib/python3.8/site-packages/django/core/files/utils.py", line 38, in <lambda>
2021-12-12T13:36:51.740031578Z     fileno = property(lambda self: self.file.fileno)
2021-12-12T13:36:51.740042278Z   File "/tmp/8d9bd73cfc52ab6/antenv/lib/python3.8/site-packages/django/core/files/utils.py", line 38, in <lambda>
[...]

PR #25662, opened 2022-08-11
[Storage] Raise error if download response does not contain Content-Range

jalauzon-msft:
As discovered in #22297, downloading a blob through the Microsoft CDN can return a response that does not contain a `Content-Range` header. This causes an unhelpful `NoneType` exception to occur while we are processing the download in `StorageStreamDownloader`.

The missing `Content-Range` is an unexpected case but this change aims to handle it more gracefully by raising a `ValueError` if we do not get a valid `Content-Range` in the response.


vincenttran-msft (approved):
Pesky CDN 😞, thanks for getting to this to hopefully help point us in the right direction!

PR #25681, opened 2022-08-12
[Storage] Migrate `test_retry.py` to test proxy

vincenttran-msft:
This PR migrates `test_retry.py` and `test_retry_async.py` to test proxy.


jalauzon-msft (approved):
Another one down!

PR #25693, opened 2022-08-13
[Storage] Migrate `test_get_blob.py` to test proxy migration

vincenttran-msft:
This PR migrates `test_get_blob.py` and `test_get_blob_async.py` to test proxy migration 😄


sdk/storage/azure-storage-blob/tests/get_blob_to_stream_async.temp.b349881e-0f45-4eb2-aeb3-296aa3649006.dat
```diff
@@ -1,138 +1 @@
-interactions:
```
jalauzon-msft, line 1: Can you remove these `dat` files? These are temp files some of the tests and usually clean up but if a test fails, they don't always get cleaned up. They should not be checked in though.

sdk/storage/azure-storage-blob/tests/test_get_blob.py
```diff
@@ -142,11 +151,14 @@ def test_get_blob_to_bytes(self, storage_account_name, storage_account_key):
         content = blob.download_blob(max_concurrency=2).readall()
 
         # Assert
-        self.assertEqual(self.byte_data, content)
+        assert self.byte_data == content
 
     @pytest.mark.live_test_only
```
jalauzon-msft, line 156: Sorry, this may create a bit of work, but this test (and likely others) does not need to be Live Only. It has the comment `parallel tests introduce random order of requests, can only run live` but this test is not parallel since it does not use `max_concurrency` > 1 anywhere (like the test about it does). I saw this quite a bit in some other files too and I think it was just copy/pasted without consideration before.

I think this is something we should be addressing as part of this migration so would you mind recording this test and deleting that comment. Additionally, can you have a quick scan through other Live Only tests here to see if any other fall into this category?

sdk/storage/azure-storage-blob/tests/test_get_blob.py
```diff
@@ -800,7 +905,10 @@ def callback(response):
...
```
jalauzon-msft, line 906: This looks like another example of a test that can be recorded.

sdk/storage/azure-storage-blob/tests/test_get_blob.py
```diff
@@ -1,34 +1,28 @@
...
 
-from azure.storage.blob import (
-    BlobServiceClient,
-    StorageErrorCode,
-    BlobProperties
-)
-from devtools_testutils.storage import StorageTestCase
+import pytest
+from azure.core.exceptions import HttpResponseError
+from azure.storage.blob import BlobProperties, BlobServiceClient, StorageErrorCode
```
jalauzon-msft, line 15: nit: new line after this line to separate libraries and local imports.

sdk/storage/azure-storage-blob/tests/test_get_blob_async.py
```diff
@@ -1,51 +1,29 @@
...
-    BlobProperties
-)
-
-from azure.storage.blob.aio import (
-    BlobClient,
-    BlobServiceClient,
-)
-from devtools_testutils.storage.aio import AsyncStorageTestCase
+from azure.storage.blob import BlobProperties, StorageErrorCode
+from azure.storage.blob.aio import BlobClient, BlobServiceClient
```
jalauzon-msft, line 16: nit: same new line here

jalauzon-msft (commented):
Nice work! I thought this one would be harder than it seems like it was. Just a couple minor comments.

sdk/storage/azure-storage-blob/tests/get_blob_to_stream_async.temp.b349881e-0f45-4eb2-aeb3-296aa3649006.dat
vincenttran-msft: Will do, did not even see these!

sdk/storage/azure-storage-blob/tests/test_get_blob.py
vincenttran-msft: Good point. Will audit the Live Only tests and get rid of it accordingly 😄

jalauzon-msft (approved):

PR #25707, opened 2022-08-15
[Storage] Add `read` to datalake `StorageStreamDownloader`

jalauzon-msft:
As a follow-up to #24275, this adds the `read` method to the `StorageStreamDownloader` for datalake. This was fairly trivial since datalake's downloader just calls into blob. Added one basic test to ensure functionality but more advanced testing is handled in blob tests.


vincenttran-msft (approved):
Thanks for getting to this! Glad you were able to double dip the benefits for all the effort you put in for Blob 😄

PR #25714, opened 2022-08-16
add instance_discovery support

xiangyan99:


sdk/identity/azure-identity/CHANGELOG.md
```diff
@@ -6,6 +6,8 @@
 
 ### Breaking Changes
 
+- Replaced `validate_authority` with `known_authority_hosts`. Now instead of setting validate_authority=False to disable authority validation, you need to explicitly specify your custom authority host as known_authority_hosts.
```
rayluo, line 9: Nit.

```suggestion
- Replaced `validate_authority` with `known_authority_hosts`. Now instead of setting validate_authority=False to disable authority validation, you need to explicitly specify your custom authority host(s) as known_authority_hosts.
```

sdk/identity/azure-identity/tests/test_known_authority_hosts.py
```diff
@@ -0,0 +1,16 @@
...
+
+def test_set_known_authority_host():
+    known_authority_hosts = ["private.cloud"]
+    credential = MsalCredential(
+        client_id="CLIENT_ID",
+        known_authority_hosts=known_authority_hosts,
+    )
+    app = credential._get_app()
+    expected = frozenset(known_authority_hosts)
+    assert app._known_authority_hosts == expected
```
rayluo, line 16: I'm fine to keep this test case as-is. Just want to call out that it relies on an upstream internal attribute, so it may potentially break in the future. You can consider porting [this upstream test case](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/492/files#diff-1982837bc801e709175e25839e1046a445df689d7a2e6309ccb6856e350dee7eR187-R191) instead. That is the public behavior that MSAL will promise for backward compatibility.

xiangyan99: Looks like the test you linked was to call set_known_authority_hosts twice?

rayluo: Indeed. It mimics the calling pattern in Azure Identity that multiple MsalCredential(..., known_authority_hosts=[...]) could be used.

sdk/identity/azure-identity/tests/test_known_authority_hosts.py
```diff
@@ -14,3 +14,16 @@ def test_set_known_authority_host():
...
+    app = credential._get_app()
+    expected = frozenset(known_authority_hosts)
+    assert app._known_authority_hosts == expected
```
rayluo, line 29: Would you consider this approach? It tests the behavior of public API, without relying on an internal attribute.
```suggestion
    app = credential._get_app()
    app = credential._get_app()  # It is allowed to create a new instance with same known_authority_hosts
    with pytest.raises(ValueError):
        credential = MsalCredential(
        client_id="CLIENT_ID",
        known_authority_hosts=["hacked.com"],
        )._get_app()
```

sdk/identity/azure-identity/tests/test_known_authority_hosts.py
```diff
@@ -0,0 +1,30 @@
+# ------------------------------------
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+# ------------------------------------
+from azure.identity._internal.msal_credentials import MsalCredential
+
+
+def test_set_known_authority_host():
```
rayluo, line 8: With the new test case below, I think we can remove this old one which relies on MSAL's internal attribute. But it is up to you. I'm leaving an approval here, even before your fixing the build errors. :-)

rayluo (approved):

sdk/identity/azure-identity/azure/identity/_internal/msal_credentials.py
```diff
@@ -29,7 +29,7 @@ def __init__(
             **kwargs
     ) -> None:
         authority = kwargs.pop("authority", None)
-        # self._validate_authority = kwargs.pop("validate_authority", True)
```
mccoyp, line 32: Is this something we plan to revert at some point? Why comment this out instead of removing it?

xiangyan99: I think I deleted this line?

mccoyp: You're completely right, and I'm dumb 🙂

mccoyp (approved):

PR #25724, opened 2022-08-16
use renamed parameter in function

kristapratico:
@scbedd

L93 renamed test_variables to test_output so updating the parameter name as well. Works locally to generate recordings, but let me know if this isn't what we should do


kristapratico:
You guys are too fast :) 
Closing this since https://github.com/Azure/azure-sdk-for-python/pull/25725 merged.

Issue #22162, opened 2021-12-14
[Test Proxy] Attempt to auto-update Docker image


mccoyp:
Currently, if there's an existing `ambitious_azsdk_test_proxy` container, the `test_proxy` fixture will just [start that container](https://github.com/Azure/azure-sdk-for-python/blob/main/tools/azure-sdk-tools/devtools_testutils/proxy_docker_startup.py#L132). This means that the version of the test proxy tool could be outdated if the container was created with an old image tag.

A new container will be created with the latest version if the existing container is deleted before the fixture is called, but this requires manual intervention. It would be ideal to either check the existing image tag of the container, or fetch the correct image each time the fixture is invoked. The latter solution could be minimally time-consuming if Docker is smart about recognizing that the container already exists with the same image tag.

mccoyp:
After discussion with other language teams, it seems like the easiest solution would be to fetch an image tag from `/eng/common`  each time the test proxy is started up ([JS example](https://github.com/Azure/azure-sdk-for-js/blob/a47af26db471d30488112c1c2855c3ea6c9b6d07/common/tools/dev-tool/src/util/testProxyUtils.ts#L84)) instead of only doing so when an `ambitious-azsdk-test-proxy` container is absent. Docker won't re-fetch the image if we already have the tag cached, so it wouldn't inhibit performance unless there's a new tag present in the repo (which happens infrequently).

ponopono0322:
According to the `eng/common/testproxy/target_version.txt` file, i deleted the previous version of the container and completed the code to create the required version of the docker image and container. Could you please check if it is the same as the intended result?

mccoyp:
@ponopono0322 I'll take a look -- thanks so much for opening a PR!

PR #25730, opened 2022-08-17
Update proxy_startup.py

ponopono0322:
Auto check and change docker container

# Description
Fixed an issue(#22162) where docker containers were not automatically updated from `eng/common/testproxy/target_version.txt` file.

This is the process of checking the container named `ambitious_azsdk_test_proxy` and then force-deleting it if the versions do not match. This only deletes the container, not the image.


ponopono0322:
This is a scenario for the current change.

1. Check `eng/common/testproxy/target_version.txt` is old. In this scenario past version is `1.0.0-dev.20220805.3`
Screen Shot 2022-08-17 at 12 02 23 PM

2. Check docker container same as 1.
Screen Shot 2022-08-17 at 11 58 54 AM

3. Change 1 to latest version. In this scenario past version is `1.0.0-dev.20220810.2`
Screen Shot 2022-08-17 at 12 03 02 PM

4. Check the name of the docker container if it has been changed well.
Screen Shot 2022-08-17 at 12 03 38 PM

Also, `azsdkengsys.azurecr.io/engsys/test-proxy` image works better performance on Silicon Mac. Consider using it.
In AMD mode, it took 02:12.
Screen Shot 2022-08-17 at 12 12 57 PM

In ARM mode, it took 01:23.
Screen Shot 2022-08-17 at 12 16 17 PM
Screen Shot 2022-08-17 at 12 17 27 PM

tools/azure-sdk-tools/devtools_testutils/proxy_startup.py
```diff
@@ -72,8 +73,26 @@ def get_container_info() -> "Optional[dict]":
...
+        #     image_prefix = MAC_IMAGE_SOURCE_PREFIX
+        else:
+            image_prefix = LINUX_IMAGE_SOURCE_PREFIX
+
+        if container_spec['Image'] == image_prefix + ':' + get_image_tag():
+            # This will succeed if we found a container with CONTAINER_NAME
+            return container_spec
+        else:
+            remove_proc = subprocess.Popen(
+                shlex.split("docker rm -f " + CONTAINER_NAME),
```
mccoyp, line 89: This line got me thinking about whether it's actually necessary to keep our test proxy container at all. The bulk of the work when starting a container is fetching the image if we don't have it locally -- but creating a container from a cached image takes almost no time.

I looked at [JavaScript's test proxy startup/teardown](https://github.com/Azure/azure-sdk-for-js/blob/main/common/tools/dev-tool/src/util/testProxyUtils.ts) code for inspiration, and they take the temporary container approach by starting up the container with a [`docker run --rm` command](https://github.com/Azure/azure-sdk-for-js/blob/1d1b9c44ac0d32af9ac9a348205cc6461c654540/common/tools/dev-tool/src/util/testProxyUtils.ts#L62) so that the container is deleted after tests run. The _image_ is cached locally still, so running tests doesn't take noticeably longer with this approach compared to what we're doing today.

I think we should do the same in Python by removing the `get_container_info` method and using `create_container` to create a new container whenever tests are run with Docker. The changes necessary are fairly significant, so I created a gist [here](https://gist.github.com/mccoyp/48fc6b262c2cab532f47aaf1d7882dcb) with the code. What do you think?

mccoyp (commented):
Thank you again for opening a PR!

tools/azure-sdk-tools/devtools_testutils/proxy_startup.py
ponopono0322: This is exactly what I wanted. But I found one potential problem. The container is not deleted normally and remains due to a container creation error. The following is an example of an error caused by a port problem.
1. First, activate port 5000. I am using Mac, I use port 5000 by turning on the airplay function in the system. Maybe you can activate port 5000 in another container and get the same result.  

|Before|After|
|:-----:|:----:|
|Screen Shot 2022-08-18 at 11 25 36 AM|Screen Shot 2022-08-18 at 11 24 12 AM|

2. Run the test proxy. It does not proceed due to an error.

3. You can see that the container was not deleted because it was not shut down properly due to an error.
Screen Shot 2022-08-18 at 11 17 40 AM

4. In this state, stop the port 5000 and run the test proxy, but it does not work properly.
Screen Shot 2022-08-18 at 11 29 00 AM

Screen Shot 2022-08-18 at 11 20 22 AM

It may be manually controlled by docker commands or desktops, but for more convenient use, I think we need code to see if containers exist. Therefore, i propose to create the delete_container() function and add it to the beginning of the start_container() function.

Please check the [code](https://gist.github.com/ponopono0322/3b9f8189f25bb36bbd64ecba4bf809ce) I suggest.

mccoyp: Awesome, that makes sense to me. Handling the case when the container is hanging around is a great idea 🙂

I think the code you linked looks great! The only change I would suggest would be to simplify the `delete_container` method just a bit -- more specifically, to simplify the `subprocess.Popen()` call since we're just running a simple command:
```python
def delete_container() -> None:
    """Delete container if it remained"""
    proc = subprocess.Popen(shlex.split(f"docker rm -f {CONTAINER_NAME}"))
    output, stderr = proc.communicate()
    return None
```
The `proc.communicate()` call is still important here; when I was testing this locally, it seems that deleting the container may take a second to register and the `docker run` command can get stuck if the deletion wasn't waited on until completion (which `proc.communicate()` does for us).

mccoyp (approved):
This is great -- thanks again!!

ponopono0322:
@mccoyp Thank you. Have a nice day!

Issue #19755, opened 2021-07-12
Listing blobs names is very slow


jochen-ott-by:
- **Package Name**: azure-storage-blob
- **Package Version**: 12.8.1
- **Operating System**: linux (Debian 9)
- **Python Version**: 3.8

**Describe the bug**

For containers containing many blobs, listing blob names takes a lot of time and uses a lot of CPU, while it was fast for azure-storage-blob 2.1.0

**To Reproduce**
1. create a blob store container with at least 5000 blobs (i.e. [maxresults for a single page returned by the list blob API](https://docs.microsoft.com/en-us/rest/api/storageservices/list-blobs#uri-parameters)).
2. use azure-storage-blob 2.1.0, `list_blob_names` to list the blob names for this container and write down the CPU time it takes (for my machine, it's 376ms).
3. use azure-storage-blob 12.8.1. Unfortunately, it does not have a list_blob_names function, so I have to use `list_blobs` and access `blob.name` for the result. Again, write down the CPU time this takes (for my machine, it's 2760ms)
4. Compare the CPU times from 2. and 3. and note the large factor (more than 7 for my case)

**Expected behavior**

There is a way of listing blob names in azure-storage-blob 12.X that has similar performance as in azure-storage-blob 2.X. 

**Additional context**

This might not be relevant for contains with a few thousand blobs. However, we have containers with a few hundred thousand to a million blobs, and bookkeeping operations that rely on listing the blob content that used to consume a little more than 1 minute CPU time for azure-storage-blob now consume take almost 10 minutes, which is a significant contribution to the runtime of these tasks.

azure-storage-blob 2.X was affected by a similar problem which has been addressed via https://github.com/Azure/azure-storage-python/pull/545
See also additional context there, in particular the use cases listed in https://github.com/Azure/azure-storage-python/pull/545#issuecomment-457504607

I would be willing to contribute an according patch. However, it seems azure-storage-blobs 12.X uses a different way to deserialize the xml response that makes it harder to customize deserialization for azure-storage-blob 12.X, compared to 2.X.

xiangyan99:
Thanks for the feedback, we’ll investigate asap.

jochen-ott-by:
[...]

PR #25747, opened 2022-08-17
[Storage] Add `list_blob_names` API to Blob SDK

jalauzon-msft:
In response to #19755

Adding a new API to `ContainerClient` called `list_blob_names` which will call the List Blobs API but only deserialize and return the names of the blobs. This should be much faster than the existing `list_blobs`.

Initial perf test results comparing regular `list_blobs` to `list_blob_names`:
|            | Throughput - Ops/s |           |             |
|------------|--------------------|-----------|-------------|
| Blob Count | list_blobs | list_blob_names | Diff  |
| 5          | 255.7              | 367.1     | 1.435666797 |
| 500        | 8.556              | 75.88     | 8.868630201 |
| 50000      | 0.0708             | 0.9809    | 13.85451977 |


sdk/storage/azure-storage-blob/azure/storage/blob/_container_client.py
```diff
@@ -785,6 +796,39 @@ def list_blobs(self, name_starts_with=None, include=None, **kwargs):
             command, prefix=name_starts_with, results_per_page=results_per_page,
             page_iterator_class=BlobPropertiesPaged)
 
+    @distributed_trace
+    def list_blob_names(self, **kwargs: Any) -> ItemPaged[str]:
+        """Returns a generator to list the names of blobs under the specified container.
+        The generator will lazily follow the continuation tokens returned by
+        the service.
+
+        :keyword str name_starts_with:
```
annatisch, line 805: Do we need to support the `include` keyword param?

annatisch (approved):
This is really awesome - thanks!

annatisch:
Adding @mikeharder as an FYI - @jalauzon-msft has updated the Storage list_blobs perf test to optionally switch to the "list names only" API which has big perf improvements.
Not sure there a way we can add that to the pipeline...?

mikeharder:
> Adding @mikeharder as an FYI - @jalauzon-msft has updated the Storage list_blobs perf test to optionally switch to the "list names only" API which has big perf improvements. Not sure there a way we can add that to the pipeline...?

We can certainly add these tests to the pipeline eventually, so we can compare listing names only with full metadata.  There are few complications:

1. If we run this perf test in the pipeline, it will fail against the GA package since it doesn't contain this API.  I think we can live with this, the perf framework should gracefully handle failures.  We'll get results for sources but a failure for the GA package.

2. Currently, we run the same set of tests across all languages, as defined in https://github.com/Azure/azure-sdk-tools/blob/main/tools/perf-automation/Azure.Sdk.Tools.PerfAutomation/tests.yml.  There are hacky ways we could add a test for just python, but the long-term fix is for me to split tests.yml into one file per package per language: 
 https://github.com/Azure/azure-sdk-tools/issues/3530.  I will move this issue to the top of my backlog, and start by enabling pipelines to specify their own tests.yml, so we can slowly migrate pipelines to the new model.

sdk/storage/azure-storage-blob/azure/storage/blob/_container_client.py
jalauzon-msft: As discussed offline, will leave this out for now.

PR #25752, opened 2022-08-18
Implement network statsbeat - success count

lzchen:
Adds infrastructure for statsbeat metrics
Implements network statsbeat - success count
Following spec:
https://github.com/microsoft/Telemetry-Collection-Spec/blob/main/ApplicationInsights/statsbeat.md


sdk/monitor/azure-monitor-opentelemetry-exporter/azure/monitor/opentelemetry/exporter/export/_base.py
```diff
@@ -86,6 +91,11 @@ def __init__(self, **kwargs: Any) -> None:
                 name="{} Storage".format(self.__class__.__name__),
                 lease_period=self._storage_min_retry_interval,
             )
+        # statsbeat initialization
+        if self._check_stats_collection():
+            # Import here to avoid circular dependencies
+            from azure.monitor.opentelemetry.exporter.statsbeat._statsbeat import collect_statsbeat_metrics
```
jeremydvoss, line 97: Could you explain the circular dependencies?

sdk/monitor/azure-monitor-opentelemetry-exporter/azure/monitor/opentelemetry/exporter/export/_base.py
```diff
@@ -187,6 +201,12 @@ def _transmit(self, envelopes: List[TelemetryItem]) -> ExportResult:
         self._consecutive_redirects = 0
         return ExportResult.SUCCESS
 
+    # check to see whether its the case of stats collection
+    def _check_stats_collection(self):
```
jeremydvoss, line 205: I think this function name could be better. It sounds like it's doing some sort of test. Is there something else we could call it like should_collect_stats or is_statsbeat_on?

sdk/monitor/azure-monitor-opentelemetry-exporter/azure/monitor/opentelemetry/exporter/export/_base.py
```diff
@@ -121,13 +131,15 @@ def _transmit(self, envelopes: List[TelemetryItem]) -> ExportResult:
         if len(envelopes) > 0:
             try:
                 track_response = self.client.track(envelopes)
-                if not track_response.errors:
+                if not track_response.errors:  # 200
                     self._consecutive_redirects = 0
+                    if self._check_stats_collection():
+                        _update_requests_map('success')
```
jeremydvoss, line 137: Since it seems these are standardized, I think we should set up an enum or at least a constant.

sdk/monitor/azure-monitor-opentelemetry-exporter/azure/monitor/opentelemetry/exporter/export/metrics/_exporter.py
```diff
@@ -81,12 +81,12 @@ def export(
...
         self,
         timeout_millis: float = 10_000,
-    ) -> None:
+    ) -> bool:
         """
         Ensure that export of any metrics currently received by the exporter
         are completed as soon as possible.
         """
-        # TODO
+        return True
```
jeremydvoss, line 89: Is this just to confirm that the function completed correctly? There's no False scenario.

sdk/monitor/azure-monitor-opentelemetry-exporter/azure/monitor/opentelemetry/exporter/statsbeat/_exporter.py
```diff
@@ -0,0 +1,37 @@
...
+    # pylint: disable=protected-access
+    def _point_to_envelope(
+        self,
+        point: DataPointT,
+        name: str,
+        resource: Optional[Resource] = None,
+        scope: Optional[InstrumentationScope] = None
+    ) -> TelemetryItem:
+        # map statsbeat name from OpenTelemetry name
+        for ot_name, sb_name in _STATSBEAT_METRIC_NAMES:
```
jeremydvoss, line 28: So, every time we're sending a metric, we're iterating over this. It probably won't have a significant performance impact, but I think we should just set up a static map from ot_name to sb_name and remove this.

sdk/monitor/azure-monitor-opentelemetry-exporter/azure/monitor/opentelemetry/exporter/statsbeat/_statsbeat.py
```diff
@@ -0,0 +1,108 @@
...
+import threading
+
+from opentelemetry.sdk.metrics import MeterProvider
+from opentelemetry.sdk.metrics.export import PeriodicExportingMetricReader
+
+from azure.monitor.opentelemetry.exporter.statsbeat._exporter import _StatsBeatExporter
+from azure.monitor.opentelemetry.exporter.statsbeat._statsbeat_metrics import _StatsbeatMetrics
+
+# pylint: disable=line-too-long
+_DEFAULT_NON_EU_STATS_CONNECTION_STRING = "InstrumentationKey=c4a29126-a7cb-47e5-b348-11414998b11e;IngestionEndpoint=https://westus-0.in.applicationinsights.azure.com/"
```
jeremydvoss, line 13: We have so many constants here and in _state. What do you think of just moving all or most of them to a separate constants.py file. That should also make circular dependencies less of an issue.

sdk/monitor/azure-monitor-opentelemetry-exporter/azure/monitor/opentelemetry/exporter/statsbeat/_statsbeat.py
```diff
@@ -0,0 +1,108 @@
...
+#             # with _STATSBEAT_STATE_LOCK:
+#             #     _STATSBEAT_STATE["SHUTDOWN"] = True
+#             pass
+
+
+def _get_stats_connection_string(endpoint: str) -> str:
+    cs_env = os.environ.get("APPLICATION_INSIGHTS_STATS_CONNECTION_STRING")
+    if cs_env:
+        return cs_env
+    for ep in _EU_ENDPOINTS:
```
jeremydvoss, line 88: Is ep a kind of prefix rather than a whole endpoint? If so, let's change the name to be clearer

sdk/monitor/azure-monitor-opentelemetry-exporter/azure/monitor/opentelemetry/exporter/statsbeat/_statsbeat.py
jeremydvoss: Not a blocker. Just a thought.

sdk/monitor/azure-monitor-opentelemetry-exporter/azure/monitor/opentelemetry/exporter/statsbeat/_statsbeat_metrics.py
```diff
@@ -0,0 +1,90 @@
...
+
+# pylint: disable=unused-argument
+# pylint: disable=protected-access
+def _get_success_count(options: CallbackOptions) -> Iterable[Observation]:
+    observations = []
+    attributes = dict(_StatsbeatMetrics._COMMON_ATTRIBUTES)
+    attributes.update(_StatsbeatMetrics._NETWORK_ATTRIBUTES)
+    attributes["statusCode"] = 200
+    with _REQUESTS_LOCK:
+        interval_count = _REQUESTS_MAP.get("success", 0)
```
jeremydvoss, line 32: Another place to use a constant

sdk/monitor/azure-monitor-opentelemetry-exporter/azure/monitor/opentelemetry/exporter/statsbeat/_statsbeat.py
```diff
@@ -0,0 +1,108 @@
...
+                statsbeat_exporter,
+                export_interval_millis=_get_stats_short_export_interval() * 1000,  # 15m by default
+            )
+            _STATSBEAT_METER_PROVIDER = MeterProvider(metric_readers=[reader])
+            statsbeat_metrics = _StatsbeatMetrics(
+                _STATSBEAT_METER_PROVIDER,
+                exporter._instrumentation_key,
+                exporter._endpoint,
+            )
+            return statsbeat_metrics
```
jeremydvoss, line 53: This return value does not seem used and the method header is -> None. What is this needed for?

sdk/monitor/azure-monitor-opentelemetry-exporter/azure/monitor/opentelemetry/exporter/statsbeat/_statsbeat_metrics.py
```diff
@@ -0,0 +1,90 @@
...
+        )
+
+# cSpell:enable
+
+def _shorten_host(host):
+    if not host:
+        host = ""
+    match = _HOST_PATTERN.match(host)
+    if match:
+        host = match.group(1)
```
jeremydvoss, line 89: Since we're using a regex, it might be a good idea to add another host none check here.

sdk/monitor/azure-monitor-opentelemetry-exporter/azure/monitor/opentelemetry/exporter/export/_base.py
lzchen: `_base.py` depends on `_statsbeat.py` for `collect_statsbeat_metrics` and `statsbeat.py` depends on `_StatsBeatExporter` which inherits `BaseExporter` from `_base.py`.

sdk/monitor/azure-monitor-opentelemetry-exporter/azure/monitor/opentelemetry/exporter/export/_base.py
lzchen: Replaced with `_should_collect_stats`

sdk/monitor/azure-monitor-opentelemetry-exporter/azure/monitor/opentelemetry/exporter/export/_base.py
lzchen: Changed this to use `_REQ_SUCCESS_NAME` constant.

sdk/monitor/azure-monitor-opentelemetry-exporter/azure/monitor/opentelemetry/exporter/export/metrics/_exporter.py
lzchen: Right, `flush` in metricexporters don't really do anything since the flushing mechanism is handled by the `BatchExportingMetricReader`, we just include them to be compliant with the spec.

See [this](https://github.com/open-telemetry/opentelemetry-python/blob/main/opentelemetry-sdk/src/opentelemetry/sdk/metrics/_internal/export/__init__.py#L158) as an example.

sdk/monitor/azure-monitor-opentelemetry-exporter/azure/monitor/opentelemetry/exporter/statsbeat/_exporter.py
lzchen: Created mapping `_STATSBEAT_METRIC_NAME_MAPPINGS`

sdk/monitor/azure-monitor-opentelemetry-exporter/azure/monitor/opentelemetry/exporter/statsbeat/_statsbeat.py
lzchen: Circular dependencies would still be an issue regardless (see above explanation). I believe it would be more Pythonic and easier to read to have all related constants in the same file (similar to [OpenTelemetry](https://github.com/open-telemetry/opentelemetry-python/blob/main/opentelemetry-sdk/src/opentelemetry/sdk/trace/__init__.py#L71)). Also based off personal preference, I prefer not putting constants in a separate file that are not accessed by anywhere else in the codebase (things like `_DEFAULT_NON_EU_STATS_CONNECTION_STRING` will only ever be accessed by the file that it is in).

sdk/monitor/azure-monitor-opentelemetry-exporter/azure/monitor/opentelemetry/exporter/statsbeat/_statsbeat.py
lzchen: Changed to return nothing.

sdk/monitor/azure-monitor-opentelemetry-exporter/azure/monitor/opentelemetry/exporter/statsbeat/_statsbeat_metrics.py
lzchen: If `host` is `None` we just store it as `None`.

sdk/monitor/azure-monitor-opentelemetry-exporter/azure/monitor/opentelemetry/exporter/statsbeat/_statsbeat.py
jeremydvoss: Sounds good to me.

sdk/monitor/azure-monitor-opentelemetry-exporter/azure/monitor/opentelemetry/exporter/export/_base.py
jeremydvoss: Cool.

sdk/monitor/azure-monitor-opentelemetry-exporter/azure/monitor/opentelemetry/exporter/export/_base.py
jeremydvoss: Perfect

sdk/monitor/azure-monitor-opentelemetry-exporter/azure/monitor/opentelemetry/exporter/statsbeat/_statsbeat.py
jeremydvoss: New version is clearer

sdk/monitor/azure-monitor-opentelemetry-exporter/azure/monitor/opentelemetry/exporter/statsbeat/_statsbeat_metrics.py
jeremydvoss: changed

sdk/monitor/azure-monitor-opentelemetry-exporter/azure/monitor/opentelemetry/exporter/statsbeat/_statsbeat_metrics.py
jeremydvoss: Oh I see. I was thinking about the regex wrong. If there is a match, it by definition means the second group exists. Got it.

sdk/monitor/azure-monitor-opentelemetry-exporter/azure/monitor/opentelemetry/exporter/export/_base.py
jeremydvoss: Because the StatsBeatExporter not only extends the base exporter, but the metrics exporter as well, which needs to collect statsbeat, there is no clear fix for this right now.

jeremydvoss (approved):

PR #25775, opened 2022-08-19
[Storage] Migrate `test_file.py` test proxy migration

vincenttran-msft:
This PR migrates `test_file.py` and `test_file_async.py` 😄


sdk/storage/azure-storage-file-datalake/tests/test_file.py
```diff
@@ -98,10 +95,23 @@ def _create_file_and_return_client(self, directory="", file=None):
         file_client.create_file()
         return file_client
 
+    def _is_almost_equal(self, first, second, delta):
```
vincenttran-msft, line 98: Curious to see your feedback on this helper name. I avoided using `assert` since I don't want to mix up this class' helpers with any real `assert` functions from our test utils.

sdk/storage/azure-storage-file-datalake/tests/test_file_async.py
```diff
@@ -98,10 +94,23 @@ async def _create_file_and_return_client(self, directory="", file=None):
         await file_client.create_file()
         return file_client
 
+    def _is_almost_equal(self, first, second, delta):
```
vincenttran-msft, line 97: Same feedback here, just making a note so that I remember to change it in both places (if we decide on a name change)

sdk/storage/azure-storage-file-datalake/tests/test_file.py
```diff
@@ -1,18 +1,15 @@
-# coding: utf-8
-
 # -------------------------------------------------------------------------
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License. See License.txt in the project root for
 # license information.
 # --------------------------------------------------------------------------
+import pytest
```
jalauzon-msft, line 6: Technically this should go just above `azure.core` since its a library not a built-in

sdk/storage/azure-storage-file-datalake/tests/test_file.py
jalauzon-msft: Seems fine to me :)

sdk/storage/azure-storage-file-datalake/tests/test_file.py
```diff
@@ -639,10 +740,15 @@ def test_preauthorize_user_with_user_delegation_key(self, datalake_storage_accou
                                              credential=sas_token)
 
         acl = new_file_client.set_access_control(permissions='0777')
-        self.assertIsNotNone(acl)
+        assert acl is not None
 
+    # TODO: MAYBE LIVE ONLY BC CONCURRENCY 2
```
jalauzon-msft, line 745: Probably remove the TODO :)

But to answer the question, the data size here is only 1024 which would be smaller than the default chunk size (and I don't think it's changed here) so it will download in a single chunk which means `max_concurrency` is doing nothing here. I think you can remove the `max_concurrency` and leave this recorded.

Applies to test below as well. I wouldn't be surprised if this occurs more in the file too.

sdk/storage/azure-storage-file-datalake/tests/test_file.py
```diff
@@ -1002,12 +1183,15 @@ def test_rename_file_with_file_sas(self, datalake_storage_account_name, datalake
         new_client = file_client.rename_file(file_client.file_system_name+'/'+'newname'+'?'+new_token)
 
         data = new_client.download_file().readall()
-        self.assertEqual(data, data_bytes)
-        self.assertEqual(new_client.path_name, "newname")
+        assert data == data_bytes
+        assert new_client.path_name == "newname"
 
     @pytest.mark.skip(reason="Service bug, requires further investigation.")
```
jalauzon-msft, line 1189: Did you try running this? We should probably "investigate further" :)

sdk/storage/azure-storage-file-datalake/tests/test_file_async.py
```diff
@@ -1,20 +1,16 @@
-# coding: utf-8
-
 # -------------------------------------------------------------------------
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License. See License.txt in the project root for
 # license information.
 # --------------------------------------------------------------------------
+import asyncio
+import pytest
```
jalauzon-msft, line 7: Same comment. `pytest` should move down.

sdk/storage/azure-storage-file-datalake/tests/test_file_async.py
jalauzon-msft: Same response. Looks good.

sdk/storage/azure-storage-file-datalake/tests/test_file_async.py
```diff
@@ -309,21 +362,29 @@ async def test_append_data_async(self, datalake_storage_account_name, datalake_s
...
+
         await self._setUp(datalake_storage_account_name, datalake_storage_account_key)
         file_client = await self._create_file_and_return_client()
 
         # Act
         await file_client.flush_data(0)
         file_props = await file_client.get_file_properties()
 
-        self.assertIsNotNone(file_props['size'], 0)
+        assert file_props['size'] is not None
```
jalauzon-msft, line 380: Seems this assert was supposed to be equality check with `0`. Would you mind updating this. Likely applies to sync as well.

sdk/storage/azure-storage-file-datalake/tests/test_file_async.py
```diff
@@ -430,11 +505,14 @@ async def test_upload_data_async(self, datalake_storage_account_name, datalake_s
         await file_client.upload_data(data, overwrite=True, max_concurrency=5)
 
         downloaded_data = await (await file_client.download_file()).readall()
-        self.assertEqual(data, downloaded_data)
+        assert data == downloaded_data
 
     @pytest.mark.live_test_only
     @DataLakePreparer()
-    async def test_upload_data_to_existing_file_async(self, datalake_storage_account_name, datalake_storage_account_key):
+    async def test_upload_data_to_existing_file(self, **kwargs):
```
jalauzon-msft, line 512: This test and the next 2 are not live only in sync. This test does not use any concurrency so good to record. The others have concurrency but a blob size of 100 so again, it's doing nothing. Can you remove any concurrency here and record these?
Can you also update sync to remove concurrency from same tests?

sdk/storage/azure-storage-file-datalake/tests/test_file_async.py
```diff
@@ -562,16 +653,20 @@ async def test_read_file_with_user_delegation_key_async(self, datalake_storage_a
...
                                              file_client.path_name,
                                              credential=sas_token)
         downloaded_data = await (await new_file_client.download_file()).readall()
-        self.assertEqual(data, downloaded_data)
+        assert data == downloaded_data
 
     @DataLakePreparer()
-    async def test_read_file_into_file_async(self, datalake_storage_account_name, datalake_storage_account_key):
+    @recorded_by_proxy_async
+    async def test_read_file_into_file(self, **kwargs):
```
jalauzon-msft, line 666: Like sync, can remove concurrency from this test and the next one. They are already recorded which is good.

sdk/storage/azure-storage-file-datalake/tests/test_file_async.py
```diff
@@ -875,9 +1039,10 @@ async def test_rename_file_to_existing_file_async(self, datalake_storage_account
 
         data = await (await new_client.download_file()).readall()
         # the existing file was overridden
-        self.assertEqual(data, data_bytes)
+        assert data == data_bytes
 
     @DataLakePreparer()
+    @recorded_by_proxy_async
     async def test_file_encryption_scope_from_file_system_async(self, datalake_storage_account_name,
```
jalauzon-msft, line 1046: Script didn't catch these params since they are on multiple lines. Can you update this to the `kwargs` approach manually?

sdk/storage/azure-storage-file-datalake/tests/test_file_async.py
```diff
@@ -930,12 +1098,15 @@ async def test_rename_file_with_file_sas_async(self, datalake_storage_account_na
         new_client = await file_client.rename_file(file_client.file_system_name+'/'+'newname'+'?'+new_token)
 
         data = await (await new_client.download_file()).readall()
-        self.assertEqual(data, data_bytes)
-        self.assertEqual(new_client.path_name, "newname")
+        assert data == data_bytes
+        assert new_client.path_name == "newname"
 
     @pytest.mark.live_test_only
```
jalauzon-msft, line 1104: Not sure why this is live only, sync isn't. Would you mind updating?

jalauzon-msft (commented):
Thanks for taking this one on! Some minor comments as well as some auditing of live only. May be good to sweep both files for uses of `max_concurrency` and live only as I probably missed some in comments.

sdk/storage/azure-storage-file-datalake/tests/test_file.py
vincenttran-msft: Working now, sweet beans 😄

jalauzon-msft (approved):
Thanks for addressing all the feedback! Look great now and these files are in a much better state than before the migration :)

PR #25781, opened 2022-08-19
[Storage] Attempt to fix flaky retry test with larger blob

jalauzon-msft:
In a past change I made `test_retry_on_socket_timeout` use a blob download in attempt to make it less flaky by making it timeout more consistently. It turns out this made it flakier, so this change increases the blob download size by a lot in hopes to make the test less flaky.


amnguye (approved):

jaschrep-msft (approved):
Maybe leave a comment to point out the large blob size should anyone every try to make this test recordable. Good to avoid big recordings when possible.

jalauzon-msft:
This particular test can never actually be recorded since its testing socket timeouts over the network which would not happen in playback so we should be good with anyone trying to record this.

Issue #25764, opened 2022-08-18
Azure  Blob Storage delete_blobs() raises TypeError: quote_from_bytes() expected bytes


wallflower762:
- **Package Name**: azure-storage-blob
- **Package Version**: 12.13.1
- **Operating System**: Windows 10
- **Python Version**: 3.8.6

**Describe the bug**
Passing a list of `BlobProperties `or string blob names to `delete_blobs` raises `TypeError: quote_from_bytes() expected bytes`

**To Reproduce**
```
import os
from typing import List
from azure.storage.blob import BlobProperties, ContainerClient


def get_blob_connection() -> ContainerClient:
    return ContainerClient.from_connection_string(
        conn_str=os.environ["CONNECTION_STRING"],
        container_name="answer-products",
    )


def list_blobs(blob_connection: ContainerClient) -> List[BlobProperties]:
    prefix = 'a690b29f-88c5-4af2-9896-06c0752e9221/9_2100_19_1720_DP1/'
    blobs = [i for i in blob_connection.list_blobs(name_starts_with=prefix)]
    return blobs


def delete_blobs(
    blob_connection: ContainerClient, blob_list: List[BlobProperties]
):
    blob_connection.delete_blobs(blob_list)


def main():
    blob_connection = get_blob_connection()
    print('Connection created.')
    blob_list = list_blobs(blob_connection)
    print(f'Found {len(blob_list)} total blobs.')
    # blob_list = [str(b.name) for b in blob_list] # Also doesn't work, same exception
    delete_blobs(blob_connection, blob_list)
    print(f'Deleted {len(blob_list)} filtered blobs.')


if __name__ == "__main__":
    main()
```

The above code should give a stack trace like this:
```
File "c:/_DEV/prototyping/clean_up_qaqc_artifacts.py", line 40, in delete_blobs
    blob_connection.delete_blobs(blob_list)
  File "C:\_DEV\prototyping\lib\site-packages\azure\core\tracing\decorator.py", line 78, in wrapper_use_tracer
    return func(*args, **kwargs)
  File "C:\_DEV\prototyping\lib\site-packages\azure\storage\blob\_container_client.py", line 1358, in delete_blobs
    reqs, options = self._generate_delete_blobs_options(*blobs, **kwargs)
  File "C:\_DEV\prototyping\lib\site-packages\azure\storage\blob\_container_client.py", line 1262, in _generate_delete_blobs_options
[...]

PR #25785, opened 2022-08-19
[Storage] Update type hints for Blob batch operations

jalauzon-msft:
As brought up in #25764 the type hint for the `*blobs` parameter of `delete_blobs` (also our other blob batch operations) is a bit misleading saying it takes a `list` when in reality it is a variable length argument. The type hint should be the expected type of the first argument, not a `list`.

This change cleans up the type hints for all of our batch operations that accept `*blobs` in `ContainerClient`.


vincenttran-msft (approved):
LGTM! Thanks 👍

PR #25802, opened 2022-08-22
[PyAMQP] Connections TODO

kashifkhan:
This PR tries to address 2 TODOs for Connection

* `_incoming_open` :  if a connection flow during Open receives an illegal state, how do we handle. Today [`uamqp`](https://github.com/Azure/azure-uamqp-python/blob/cdcf44c06a5d066447f702ddc959e3193495c300/src/vendor/azure-uamqp-c/src/connection.c#L813) & `pyamqp` both ignore it.  Right now I've set it that if any other state appears while opening, close. Looking at some other libraries and the amqp specs, there doesnt seem to be any defined thing that should happen if the Open process goes off path, so it might be open to the implementer ?
* `_incoming_close`: How to handle a close frame on an invalid channel. I moved the todos around popping the channels on both incoming and outgoing since the spec mentions either party can close and then they mutually close. If the channel specified is not found, then is it an log and move on or close the entire thing with an error. Ive chosen the latter now

### DIFF
--- a/sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_connection.py
+++ b/sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_connection.py
@@ -415,7 +415,7 @@ class Connection(object):
             self.close(error=AMQPError(
                 condition=ErrorCondition.IllegalState, 
                 description=f"connection is an illegal state: {self.state}"))
-            _LOGGER.error(f"connection is an illegal state: {self.state}"))
+            _LOGGER.error(f"connection is an illegal state: {self.state}")
 
     def _outgoing_close(self, error=None):
         # type: (Optional[AMQPError]) -> None
--- a/sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_connection.py
+++ b/sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_connection.py
@@ -412,7 +412,10 @@ class Connection(object):
             self._outgoing_open()
             self._set_state(ConnectionState.OPENED)
         else:
-            pass # TODO what now...?
+            self.close(error=AMQPError(
+                condition=ErrorCondition.IllegalState, 
+                description=f"connection is an illegal state: {self.state}"))
+            _LOGGER.error(f"connection is an illegal state: {self.state}"))
 
     def _outgoing_close(self, error=None):
         # type: (Optional[AMQPError]) -> None
--- a/sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_connection.py
+++ b/sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_connection.py
@@ -513,7 +513,7 @@ class Connection(object):
             self._incoming_endpoints.pop(channel)
             self._outgoing_endpoints.pop(channel)
         except KeyError:
-            end_error = AMQPError(condition=ErrorCondition.ClientError, description=f"Invalid channel {channel}", info=None)
+            end_error = AMQPError(condition=ErrorCondition.InvalidField, description=f"Invalid channel {channel}", info=None)
             _LOGGER.error(f"Invalid channel {channel} ")
             self.close(error=end_error)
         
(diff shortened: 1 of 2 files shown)


sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_connection.py
```diff
@@ -507,10 +510,13 @@ def _incoming_end(self, channel, frame):
         """
         try:
             self._incoming_endpoints[channel]._incoming_end(frame)  # pylint:disable=protected-access
+            self._incoming_endpoints.pop(channel)
```
l0lawrence, line 513: Just curious: For an outgoing_close, could we receive a response from the service on that channel to confirm it is closing? (i.e from the incoming_endpoint channel). Or could we receive an error on outgoing_close that we would need the incoming channel to still be there for?

kashifkhan: I think if the entire `Connection` was being closed, then the state would be set to END after sending after sending a CLOSE to the other side. In this case while this channel has closed, there might be other channels still open. Unless I missed it here, doesnt seem to be the case in channels http://docs.oasis-open.org/amqp/core/v1.0/csprd01/amqp-core-transport-v1.0-csprd01.html#doc-idp191152

annatisch: It's actually a really good question - and why this was a TODO in the first place.
It's hard to know where to pop the channel/session/link because if it's initiated by us, we don't always wait around for the service response. 
So we need to test/be able to handle:
- If we get an incoming close/end/detach, then is this in response to one we already sent? If not, we can probably just send the response frame then pop it.
- If we are sending a close/end/detach, if we pop it then we need to be able to handle additional incoming frames that are sent to it. Otherwise, do we wait until we explicitly receive the response? That can be slow and messy if there's other things going on in the connection.

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_connection.py
```diff
@@ -507,10 +510,13 @@ def _incoming_end(self, channel, frame):
...
+            self._outgoing_endpoints.pop(channel)
         except KeyError:
-            pass  # TODO: channel error
-        #self._incoming_endpoints.pop(channel)  # TODO
-        #self._outgoing_endpoints.pop(channel)  # TODO
+            end_error = AMQPError(condition=ErrorCondition.InvalidField, description=f"Invalid channel {channel}", info=None)
```
annatisch, line 516: Is InvalidField the best error condition here?

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_connection.py
```diff
@@ -412,7 +412,10 @@ def _incoming_open(self, channel, frame):
             self._outgoing_open()
             self._set_state(ConnectionState.OPENED)
         else:
-            pass # TODO what now...?
+            self.close(error=AMQPError(
+                condition=ErrorCondition.IllegalState, 
+                description=f"connection is an illegal state: {self.state}"))
+            _LOGGER.error(f"connection is an illegal state: {self.state}")
```
annatisch, line 418: When it comes to logging, it's best not to pre-emptively format the string.
It's a small thing, but when using the f-string syntax, the string will be formatted regardless of whether there's a logger attached.
So by using this syntax, while uglier, the string will only be formatted if the logger is actually going to log the statement - which is more performant:
`_LOGGER.error("connection is in an illegal state: %r", self.state)`

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_connection.py
kashifkhan: Changed it to `ClientError` since if a channel is missing from the dictionary (either when the frame comes from the client itself or the service) its something the SDK didn't handle on close properly.

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_connection_async.py
```diff
@@ -305,7 +305,11 @@ async def _incoming_open(self, channel, frame):
             await self._outgoing_open()
             await self._set_state(ConnectionState.OPENED)
         else:
-            pass  # TODO what now...?
+            # TODO what now...?
```
annatisch, line 308: We could remove the todo?

PR #25807, opened 2022-08-22
Update identity build machine for msal 3.7 from ubunto 18.04 to 20.04

scbedd:
@mccoyp FYI

Attempting to lock down as to why this is happening in CI when I _cannot_ get a repro on this failure locally.

### DIFF
--- a/scripts/devops_tasks/tox_harness.py
+++ b/scripts/devops_tasks/tox_harness.py
@@ -5,7 +5,6 @@ import shutil
 import re
 import multiprocessing
 import glob
-import pdb
 
 if sys.version_info < (3, 0):
     from Queue import Queue
@@ -211,7 +210,8 @@ def inject_custom_reqs(file, injected_packages, package_dir):
                 logging.info("Attempting to parse {}".format(line))
                 try:
                     parsed_req = [req for req in parse_requirements(line)]
-                except RequirementParseError as e:
+                except Exception as e:
+                    logging.error(e)
                     parsed_req = [None]
                 req_lines.append((line, parsed_req))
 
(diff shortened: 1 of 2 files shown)


xiangyan99 (approved):

scripts/devops_tasks/tox_harness.py
```diff
@@ -208,9 +207,11 @@ def inject_custom_reqs(file, injected_packages, package_dir):
         )
         with open(file, "r") as f:
             for line in f:
+                logging.info("Attempting to parse {}".format(line))
                 try:
                     parsed_req = [req for req in parse_requirements(line)]
-                except RequirementParseError as e:
+                except Exception as e:
+                    logging.error(e)
```
scbedd, line 214: If we can't parse it, it's going to be a valid non-pypi requirement.

EG: `git vcs` or `raw path` or `external URI reference`

I'm pretty comfortable just letting the issues ride here due to the fact that this runs within CI only. (`TF_BUILD=true` only in CI unless devs set it manually)

The requirements we are getting out of this thing are correct.

sdk/identity/platform-matrix.json
```diff
@@ -8,9 +8,9 @@
   "include": [
     {
       "Config": {
-        "ubuntu_18.04_3.7_msal": {
-          "OSVmImage": "MMSUbuntu18.04",
-          "Pool": "azsdk-pool-mms-ubuntu-1804-general",
+        "ubuntu_20.04_3.7_msal": {
```
scbedd, line 11: This needed to be done regardless. Maintenance effort.

mccoyp (approved):

PR #25726, opened 2022-08-16
Target 3.11-rc.1 in `nightly` python build

scbedd:

PR #25813, opened 2022-08-23
Adding the ability to skip compatibility filtering

scbedd:
This feature will be used [on this line](https://github.com/Azure/azure-sdk-for-python/pull/25726/files#r952053598) to prevent 3.11 from _not_ running.

### DIFF
--- a/scripts/devops_tasks/setup_execute_tests.py
+++ b/scripts/devops_tasks/setup_execute_tests.py
@@ -26,7 +26,7 @@ from common_tasks import (
     create_code_coverage_params
 )
 from tox_harness import prep_and_run_tox
-from ci_tools.functions import discover_targeted_packages, filter_packages_by_compatibility_override
+from ci_tools.functions import discover_targeted_packages
 
 logging.getLogger().setLevel(logging.INFO)
 
--- a/scripts/devops_tasks/setup_execute_tests.py
+++ b/scripts/devops_tasks/setup_execute_tests.py
@@ -282,7 +282,6 @@ if __name__ == "__main__":
         help="Filter type to identify eligible packages. for e.g. packages filtered in Build can pass filter type as Build,",
         choices=['Build', "Docs", "Regression", "Omit_management", "None"]
     )
-    
     parser.add_argument(
         "-d",
         "--dest-dir",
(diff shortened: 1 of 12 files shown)


scripts/devops_tasks/setup_execute_tests.py
```diff
@@ -280,9 +280,9 @@ def execute_global_install_and_test(
         dest="filter_type",
         default='Build',
         help="Filter type to identify eligible packages. for e.g. packages filtered in Build can pass filter type as Build,",
-        choices=['Build', "Docs", "Regression", "Omit_management"]
+        choices=['Build', "Docs", "Regression", "Omit_management", "None"]
     )
-
+
```
mccoyp, line 285:
```suggestion

```

scripts/devops_tasks/setup_execute_tests.py
```diff
@@ -301,21 +301,23 @@ def execute_global_install_and_test(
     else:
         target_dir = root_dir
 
-    targeted_packages = discover_targeted_packages(args.glob_string, target_dir, "", args.filter_type)
-    compatible_targeted_packages = filter_packages_by_compatibility_override(targeted_packages)
```
mccoyp, line 305: This seems to be the only use of `filter_packages_by_compatibility_override` -- should we still hold onto that method?

scbedd: Check within `discover_targeted_packages` 

https://github.com/Azure/azure-sdk-for-python/blob/fc9e6f937141977b326eaffd5bd168096cee40cd/tools/azure-sdk-tools/ci_tools/functions.py#L112-L145

mccoyp: That's using `filter_for_compatibility` though; `filter_packages_by_compatibility_override` seems to be doing something slightly different, but at a glance I'm not sure if it's a duplicate or slightly different in functionality

scbedd: I will clean up filter_for_compatibility and filter_packages_by_compatibility_override into a single function. Good catch.

tools/azure-sdk-tools/ci_tools/functions.py
```diff
@@ -42,7 +42,7 @@
     "azure-mgmt-core",
 ]
 
-TEST_COMPATIBILITY_MAP = {"azure-core-tracing-opentelemetry": "<3.10"}
```
scbedd, line 45: Whatever the issue with azure-core-tracing-opentelemetry was, the specifier including py310 is there now, so I'm cleaning it out of this.

scripts/devops_tasks/setup_execute_tests.py
scbedd: Done! Thanks for the follow-up on this one McCoy!

mccoyp (approved):

PR #25828, opened 2022-08-23
Bump pinned version of wrapt to one that is compatible with py311

scbedd:
See title.

### DIFF
--- a/eng/ci_tools.txt
+++ b/eng/ci_tools.txt
@@ -17,7 +17,7 @@ codecov==2.1.0
 beautifulsoup4==4.9.1
 pkginfo==1.5.0.1
 pip==20.3.3
-wrapt<=1.12.1; python_version == '2.7'
+wrapt<=1.14.1;
 markupsafe==2.0.1; python_version > '2.7'
 markupsafe==1.1.1; python_version == '2.7'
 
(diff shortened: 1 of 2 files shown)


scbedd:
The update to `wrapt` is causing a pip resolver thrash during installation. I gotta figure out where in the graph the break is coming from.

scbedd:
I believe my latest commit should resolve this.

The issue is that we pin  `pylint==2.9.3`. `pylint==2.9.3` has a dependency on `wrapt==1.12.1`. When we update the pinned dependency of `wrapt` in `test_tools.txt` to `1.14.1` in all cases, this puts us in an unresolvable state. After that, pip just thrashes forever.

To address this, I've updated the pins to include two new ones for `py311` only that will prevent the issues from biting us.

eng/ci_tools.txt
```diff
@@ -17,7 +17,7 @@ codecov==2.1.0
 beautifulsoup4==4.9.1
 pkginfo==1.5.0.1
 pip==20.3.3
-wrapt<=1.12.1; python_version == '2.7'
+wrapt<=1.14.1;
```
scbedd, line 20:
```suggestion
wrapt==1.12.1
wrapt==1.14.1; python_version >= '3.11'
```

mccoyp (approved):

PR #25841, opened 2022-08-24
[formrecognizer] Add sub-classes for operation details

catalinaperalta:
WIP
implement sub-classes for operation details and rename resource details properties to be more specific to custom document models.


catalinaperalta:
Will not be implemented yet.

PR #25843, opened 2022-08-24
[formrecognizer] Update operation names to include document model

catalinaperalta:
See changelog for list of all renames. Updating to follow service naming for these operations.


sdk/formrecognizer/azure-ai-formrecognizer/azure/ai/formrecognizer/aio/_document_model_administration_client_async.py
```diff
@@ -461,11 +461,11 @@ def list_operations(self, **kwargs: Any) -> AsyncItemPaged[OperationSummary]:
 
     @distributed_trace_async
     async def get_operation(self, operation_id: str, **kwargs: Any) -> OperationDetails:
-        """Get a document model operation by its ID.
+        """Get a model operation by its ID.
```
kristapratico, line 464: just wondering, if everything is a document model, why is this one just a model operation?

sdk/formrecognizer/azure-ai-formrecognizer/samples/v3.2/async_samples/sample_manage_models_async.py
```diff
@@ -48,30 +48,30 @@ async def sample_manage_models_async():
 
         # Next, we get a paged list of all of our custom models
         # [START list_models_async]
```
kristapratico, line 50: I think I saw these START/END tags updated in client docstrings to include "document". We may need to update the corresponding sphinx tags in the samples.

sdk/formrecognizer/azure-ai-formrecognizer/azure/ai/formrecognizer/aio/_document_model_administration_client_async.py
catalinaperalta: I was actually thinking of removing the word model completely, since this endpoint is about operations. After checking the swagger description for this endpoint, I am going to remove model from this description as well since the service is leaving this open to other types of operations, not just limited to document models. Fixing!

sdk/formrecognizer/azure-ai-formrecognizer/samples/v3.2/async_samples/sample_manage_models_async.py
catalinaperalta: Thanks for catching that, I thought I'd got them all. Will double check and push updates

catalinaperalta: I see what happened, the async tags didn't get updated anywhere so they were all still working, but updated for symmetry with the sync tags. Fixed!

sdk/formrecognizer/azure-ai-formrecognizer/azure/ai/formrecognizer/aio/_document_model_administration_client_async.py
catalinaperalta: Done

PR #25844, opened 2022-08-24
[Storage] Migrate `test_cpk` in Blob to test proxy

jalauzon-msft:
Migrates `test_cpk` and `test_cpk_async` in the Blob package to test proxy.


vincenttran-msft (approved):
Thanks for getting to this one! I remember trying this one but I lacked critical information (namely, the sanitizer 🤡)

PR #25845, opened 2022-08-24
Implement all network statsbeat

lzchen:


sdk/monitor/azure-monitor-opentelemetry-exporter/azure/monitor/opentelemetry/exporter/export/_base.py
```diff
@@ -131,27 +145,32 @@ def _transmit(self, envelopes: List[TelemetryItem]) -> ExportResult:
         """
         if len(envelopes) > 0:
             try:
+                start_time = time.time()
                 track_response = self.client.track(envelopes)
                 if not track_response.errors:  # 200
                     self._consecutive_redirects = 0
+                    if not self._is_stats_exporter():
```
jeremydvoss, line 152: Do we want these logs to happen when statsbeat is disabled? Wondering if we should use self._should_collect_stats() instead

lzchen: We want to still log even if statsbeat is disabled.

jeremydvoss (approved):

PR #25881, opened 2022-08-26
[Storage] STG85/86 Cold Tier Preview

vincenttran-msft:
This PR aims to add `Cold` as a new Standard tier for blobs. This is for a private preview drop.

Additionally, this file fixes up some tests that were not using the `**kwargs` pop stuff from test proxy 😄


sdk/storage/azure-storage-blob/azure/storage/blob/_serialize.py
```diff
@@ -51,7 +51,8 @@
     '2021-02-12',
     '2021-04-10',
     '2021-06-08',
-    '2021-08-06'
+    '2021-08-06',
+    '2021-12-02'
```
vincenttran-msft, line 55: Slightly concerned with this change + the fact that I had to use preprod for these tests to work. I'm assuming the pipeline is not going to be very happy?

sdk/storage/azure-storage-blob/tests/test_common_blob_async.py
```diff
@@ -3098,4 +3196,21 @@ async def test_validate_empty_blob(self, **kwargs):
...
+        blob_name = await self._create_block_blob(standard_blob_tier=StandardBlobTier.Hot, overwrite=True)
+        blob = self.bsc.get_blob_client(self.container_name, blob_name)
+        await blob.set_standard_blob_tier(StandardBlobTier.Cold)
+
+        # Act
+        props = await blob.get_blob_properties()
+
+        # Assert
+        assert props.blob_tier == StandardBlobTier.Cold
+
```
jalauzon-msft, line 3215: Not sure where to put this comment but let's also a test for `list_blobs` probably in `test_container` since it's a new property there too. 

If there is an existing test that tests access tier, maybe just add a Cold blob into that test instead of making a whole new test. If there is not a test, maybe test Hot, Cool, and Cold in one test for completeness.

jalauzon-msft (commented):
Thanks for getting to this! Just a one-line change to actual code, that's nice :).

Just some comments on tests and I'd be curious what the new generated code actually looks like for this. It seems this works without it, probably because we have our own enum defined rather than relying on the generated Enum. May still be good to include new generated code for completeness.

sdk/storage/azure-storage-blob/tests/test_common_blob.py
```diff
@@ -483,6 +539,22 @@ def test_upload_blob_from_pipe(self, **kwargs):
         # Assert
         assert data == blob_data
 
+    @BlobPreparer()
+    @recorded_by_proxy
+    def test_upload_block_blob_with_tier_specified_cold(self, **kwargs):
```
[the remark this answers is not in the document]
vincenttran-msft: Sounds good, I can make that change + address the nit 😄

sdk/storage/azure-storage-blob/tests/test_common_blob.py
```diff
@@ -1575,6 +1647,48 @@ def test_copy_blob_with_blob_tier_specified(self, **kwargs):
...
+        # Assert
+        assert copy_blob_properties.blob_tier == blob_tier
+
+    @BlobPreparer()
+    @recorded_by_proxy
+    def test_copy_blob_with_cold_tier_sync(self, **kwargs):
+        storage_account_name = kwargs.pop("storage_account_name")
+        storage_account_key = kwargs.pop("storage_account_key")
+
+        self._sync_setup(storage_account_name, storage_account_key)
```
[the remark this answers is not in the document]
vincenttran-msft: Yep. I can move it over there 😄

sdk/storage/azure-storage-blob/tests/test_common_blob_async.py
vincenttran-msft: Sounds good! For now, I made them standalone tests as the only tests in `test_container` that assert against the tier are live-only, so I think for simplicity I will leave them standalone (for both `list_blobs` and `walk_blobs`)

sdk/storage/azure-storage-blob/azure/storage/blob/_serialize.py
jalauzon-msft: Yep, the Live test pipeline will not be happy against this PR, but ci should be fine due to recordings. That's fine for the private preview and the reason this can't be merged to `main`

sdk/storage/azure-storage-blob/tests/test_common_blob.py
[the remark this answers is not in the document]
jalauzon-msft: Thanks! I guess that means you'll have to manually edit out Preprod from the recordings since it's not test proxy yet :)

sdk/storage/azure-storage-blob/tests/test_common_blob.py
```diff
@@ -483,6 +483,22 @@ def test_upload_blob_from_pipe(self, **kwargs):
         # Assert
         assert data == blob_data
 
+    @BlobPreparer()
+    @recorded_by_proxy
+    def test_upload_blob_with_tier_specified_cold(self, **kwargs):
```
jalauzon-msft, line 488: Sorry, I didn't pick up on this earlier, but would you mind moving all the tests in `common_blob` to `block_blob`? Standard access tier is only supported on Block Blobs, so it doesn't really make sense in common (I just learned this a few minutes ago :))

jalauzon-msft (approved):
Looks great! Thanks

PR #25892, opened 2022-08-26
[Cosmos] commenting out analytical ttl tests

JericHunter:
this is to fix pipelines


tjprescott:
@JericHunter why is this breaking pipelines? It didn't when this change was merged?

JericHunter:
@tjprescott because the account we use to run our tests doesn't have synapse link enabled which is required to run tests with analytical ttl

kushagraThapar:
@JericHunter  / @simorenoh  - instead of commenting out the tests, why don't we fix the pipeline cosmos db accounts. 
If we comment out the tests, we will loose any future regressions for this feature.

PR #25893, opened 2022-08-26
[Pyamqp] Exception Todos

kashifkhan:
Exception todos


sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_management_operation_async.py
```diff
@@ -129,7 +129,7 @@ async def ready(self):
         # ManagementOpenResult.ERROR or CANCELLED
         # TODO: update below with correct status code + info
         raise AMQPLinkError(
-            condition=ErrorCondition.ClientError,
+            condition=ManagementOpenResult.ERROR,
```
annatisch, line 132: This is not an error condition - it's a state enum.

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_session_async.py
```diff
@@ -175,8 +177,14 @@ async def _incoming_attach(self, frame):
             self._input_handles[frame[1]] = self.links[frame[0].decode('utf-8')]  # name and handle
             await self._input_handles[frame[1]]._incoming_attach(frame)
         except KeyError:
-            outgoing_handle = self._get_next_output_handle()  # TODO: catch max-handles error
-            if frame[2] == Role.Sender:  # role
+            try:
+                outgoing_handle = self._get_next_output_handle()  # TODO: catch max-handles error
```
annatisch, line 181: We can remove the todo :)

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_session_async.py
```diff
@@ -175,8 +177,14 @@ async def _incoming_attach(self, frame):
...
+            except ValueError:
+                self._connection.close(error=AMQPSessionError(
```
annatisch, line 183: This is async - so this needs to be awaited.
Additionally I'm not sure this should be a connection close - it's currently only the Session that's invalid?
Lastly - AMQPSessionError is an exception - so you can't pass it into connection.close.

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_session_async.py
```diff
@@ -288,7 +296,11 @@ async def _incoming_transfer(self, frame):
         try:
             await self._input_handles[frame[0]]._incoming_transfer(frame)  # handle
         except KeyError:
-            pass  #TODO: "unattached handle"
+            await self._set_state(SessionState.DISCARDING)
+            await self._connection.close(error=AMQPSessionError(
```
annatisch, line 300: Same comments as above - we shouldn't be passing in an exception here. Additionally this is a link-level error, so I'm not sure it should result in the connection being closed.

annatisch (changes requested):

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_session_async.py
kashifkhan: I went with `close` after reading the `handle-max` description [here](http://docs.oasis-open.org/amqp/core/v1.0/csprd01/amqp-core-transport-v1.0-csprd01.html#type-begin). 

>  A peer that receives a handle outside the supported range MUST close the connection with the framing-error error-code


Would this only apply to the begin frame? Seeing a couple of other libraries its just an exception that is raised, so it could be I'm reading it incorrectly ( client generating an out of range id versus peer receiving one)

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_connection_async.py
```diff
@@ -349,12 +349,9 @@ async def _incoming_open(self, channel, frame):
             _LOGGER.info("<- %r", OpenFrame(*frame), extra=self._network_trace_params)
         if channel != 0:
             _LOGGER.error("OPEN frame received on a channel that is not 0.")
-            await self.close(
-                error=AMQPError(
-                    condition=ErrorCondition.NotAllowed,
-                    description="OPEN frame received on a channel that is not 0."
-                )
-            )
+            await self.close(error=AMQPConnectionError(
```
annatisch, line 352: This must be an AMQPError object - the ConnectionError is an Exception.... maybe we should consider renaming these...

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_link_async.py
```diff
@@ -182,11 +182,11 @@ async def _incoming_attach(self, frame):
         if self.network_trace:
             _LOGGER.info("<- %r", AttachFrame(*frame), extra=self.network_trace_params)
         if self._is_closed:
-            raise ValueError("Invalid link")
+            raise AMQPLinkError(condition=ErrorCondition.LinkDetachForced, description="Invalid link", info=None)
```
annatisch, line 185: I think this should probably be a ClientError.
LinkDetachForced only makes sense if it's coming from or going over the wire.

Though what sets `self._is_closed`? And where else is it checked?

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_link_async.py
```diff
@@ -182,11 +182,11 @@ async def _incoming_attach(self, frame):
...
         elif not frame[5] or not frame[6]: 
             _LOGGER.info("Cannot get source or target. Detaching link")
             await self._set_state(LinkState.DETACHED) 
-            raise ValueError("Invalid link")
+            raise AMQPLinkError(condition=ErrorCondition.LinkDetachForced, description="Invalid link", info=None)
```
annatisch, line 189: Should this be a call to `self.detach(error=AMQPError(...))`?
We can probably also make the description here more descriptive.

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_session_async.py
annatisch: Receiving a handle outside the supported range is not the same as running out of available handles.
If would mean that `frame[1]` was a value > `handle-max`.
That's not to say that closing the connection is invalid - however we have to assume that this may not be the only open session, and if we've run out of handles, there's probably additional open links. So calling `connection.close()` must gracefully shut down all other sessions and links on the connection.

However rather than close the session it might be more valid to detach the attempted link?

Lastly, this still needs an await, and you cannot pass in AMQPSessionError, it must be AMQPError.

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_session_async.py
```diff
@@ -316,7 +328,10 @@ async def _incoming_detach(self, frame):
             #     self._input_handles.pop(link.remote_handle, None)
             #     self._output_handles.pop(link.handle, None)
         except KeyError:
-            pass  # TODO: close session with unattached-handle
+            await self._set_state(SessionState.DISCARDING)
+            await self._connection.close(error=AMQPSessionError(
```
annatisch, line 332: Must be AMQPError

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_link_async.py
kashifkhan: `self._is_closed` is set by ` _outgoing_detach(self, close=False, error=None)` when close is set to True

It is then checked here:
*`_check_if_closed`
*`_on_session_state_change`
*`_incoming_attach`
*`_incoming_detach`
*`attach`

I changed it to ClientError because the Link was not forcibly closed but was already closed.

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_session_async.py
kashifkhan: I went over uamqp again and on an invalid handle_max the link is freed up, so detaching 
 the link would be the valid choice.

kashifkhan: Updated this to detach the link that would have been assigned and then return

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_link_async.py
```diff
@@ -182,11 +183,12 @@ async def _incoming_attach(self, frame):
         if self.network_trace:
             _LOGGER.info("<- %r", AttachFrame(*frame), extra=self.network_trace_params)
         if self._is_closed:
-            raise ValueError("Invalid link")
+            raise AMQPLinkError(condition=ErrorCondition.ClientError, description="Invalid link", info=None)
```
annatisch, line 186: Could we update the `description` to something like "Received Attach frame on a link that is already closed."

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/link.py
```diff
@@ -181,11 +182,12 @@ def _incoming_attach(self, frame):
...
         if self._is_closed:
-            raise ValueError("Invalid link")
+            raise AMQPLinkError(condition=ErrorCondition.ClientError, description="Invalid link", info=None)
         elif not frame[5] or not frame[6]:  
             _LOGGER.info("Cannot get source or target. Detaching link")
-            self._set_state(LinkState.DETACHED) 
-            raise ValueError("Invalid link")
+            self.detach(
+                error=AMQPError(condition=ErrorCondition.LinkDetachForced, 
+                                    description="Cannot get source or target from the frame. Detaching link", info=None))
```
annatisch, line 190: Seems to need some whitespace formatting.

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_link_async.py
annatisch: Same comment applies to the sync `link.py`

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_session_async.py
```diff
@@ -175,8 +177,13 @@ async def _incoming_attach(self, frame):
             self._input_handles[frame[1]] = self.links[frame[0].decode('utf-8')]  # name and handle
             await self._input_handles[frame[1]]._incoming_attach(frame)
         except KeyError:
-            outgoing_handle = self._get_next_output_handle()  # TODO: catch max-handles error
-            if frame[2] == Role.Sender:  # role
+            try:
+                outgoing_handle = self._get_next_output_handle()
+            except ValueError:
+                # detach the link that would have been set.
+                await self.links[frame[0].decode('utf-8')].detach()
```
annatisch, line 184: Shouldn't we be setting an AMQPError in the detach?

annatisch (approved):

PR #25896, opened 2022-08-26
[Storage] Migrate `test_cpk_n` to test proxy

jalauzon-msft:
This change migrates `test_cpk_n.py` and `test_cpk_n_async.py` to test proxy.


vincenttran-msft (approved):
LGTM! Thanks for knocking this one out, I can see it was pretty laborious with all the SAS and encryption scope stuff 🥇

PR #25897, opened 2022-08-26
[Test Proxy] Support AsyncioRequestsTransport

mccoyp:
# Description

We currently only patch `AioHttpTransport`, the most common async transport in the SDK. App Configuration uses `AsyncioRequestsTransport` as its async transport though -- whether or not it should hard code this is being discussed separately, but in any case our test proxy infrastructure should support any transport that clients choose to use.

This adds support for `AsyncioRequestsTransport` by patching its `send` method during recorded tests, just as we currently do for `AioHttpTransport`.

  - Tested using App Configuration tests.

### DIFF
--- a/tools/azure-sdk-tools/devtools_testutils/aio/proxy_testcase_async.py
+++ b/tools/azure-sdk-tools/devtools_testutils/aio/proxy_testcase_async.py
@@ -8,7 +8,7 @@ import urllib.parse as url_parse
 
 from azure.core.exceptions import ResourceNotFoundError
 from azure.core.pipeline.policies import ContentDecodePolicy
-from azure.core.pipeline.transport import AioHttpTransport
+from azure.core.pipeline.transport import AioHttpTransport, AsyncioRequestsTransport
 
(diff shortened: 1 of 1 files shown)


mccoyp:
Just talked to Laurent about this. His opinion is that the test proxy should be allowed to depend on a single async transport -- in this case, `AioHttpTransport`. The problem is that App Config currently has some tests fail if `AioHttpTransport` is used with clients.

The solution should be to fix App Config's dependence on `AsyncioRequestsTransport` so that tests can use `azure-core`'s default async transport (`AioHttpTransport`). That's going to take some investigation since there are some tests failing with `AioHttpTransport` today, so I'll leave this PR up in case we need test proxy support to get App Config migrated. (Or if we change our minds and decide to support more than just `AioHttpTransport`)

mccoyp:
Merging, per https://github.com/Azure/azure-sdk-for-python/pull/26140#issuecomment-1242579382. We can drop this support later on once App Config has proper support for `AioHttpTransport`, but for now the lack of `AsyncioRequestsTransport` in the test proxy is blocking their migration

PR #25898, opened 2022-08-27
[Test Proxy] Accept parameter-level SAS sanitization

mccoyp:
# Description

**Context:** the Storage team discovered that there are cases where SAS tokens generated by tests can fail to match in playback because they contain tenant IDs. The problem is that `AzureRecordedTestCase.generate_sas` currently sanitizes only the `sig`, `st`, and `se` parameters of a generated token. In Storage's case, their tokens also contain a `sktid` field that includes a tenant ID. This ID doesn't get sanitized, and in playback the local value of the tenant ID ("00000000-0000-0000-0000-000000000000") doesn't match when a new SAS is generated.

This PR expands the functionality of the `generate_sas` method to resolve this issue. A `fake_parameters` kwarg can now be provided, specifying SAS token keys to sanitize. Since the change is purely additive, it's backwards-compatible for tests that rely on the existing `generate_sas` behavior. The Storage team has reviewed this solution offline and given a thumbs-up.

  - Tested with the test that motivated the change, as well as tests that currently use the `generate_sas` method.

### DIFF
--- a/.vscode/cspell.json
+++ b/.vscode/cspell.json
@@ -292,6 +292,7 @@
     "signup",
     "simplifiable",
     "skipif",
+    "sktid",
     "skus",
     "southcentralus",
     "ssis",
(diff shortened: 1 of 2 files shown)


scbedd (approved):
I really like `_create_fake_token`'s implementation!

PR #25929, opened 2022-08-29
[Cosmos] remove wrongly placed ayncs

JericHunter:
# Description
removing some wrongly placed `async` in our files to fix a bug, documentation below.
https://docs.python.org/3.9/reference/datamodel.html#asynchronous-iterators


simorenoh (approved):
LGTM, thanks Jeric!

PR #25940, opened 2022-08-29
[Storage] Another fix to flaky retry test

jalauzon-msft:
For some reason the socket retry test was sometimes failing when trying to delete the container saying the resource did not exist. I think this was because the timeout on the client was not properly being reset and so the first request may time out reading the response, but the service would still delete the container. The retry would then get the resource does not exist error.

This change constructs a new client with normal timeouts.


vincenttran-msft (approved):
BE GONE FLAKY TESTS 🎩

Issue #25918, opened 2022-08-29
Storage blob samples issues


v-xuto:
1.
Section [link1](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/storage/azure-storage-blob/samples/blob_samples_common.py#L207), [link2](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/storage/azure-storage-blob/samples/blob_samples_common_async.py#L209):
![image](https://user-images.githubusercontent.com/57166602/187160031-ae0e3e52-419f-4185-905d-4ffbcc2e2f2c.png)

Reason:
azure.core.exceptions.ResourceExistsError: There is currently no pending copy operation.

Suggestion:
Add `if props.copy.status != "success":` before 
`copy_id = props.copy.id`.
Update to:
```
if props.copy.status != "success":
    copy_id = props.copy.id
    # [START abort_copy_blob_from_url]
    # Passing in copy id to abort copy operation
    copied_blob.abort_copy(copy_id)
    # check copy status
    props = copied_blob.get_blob_properties()
    print(props.copy.status)
    # [END abort_copy_blob_from_url]
```

2.
Section [link1](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/storage/azure-storage-blob/samples/blob_samples_containers.py#L228), [link2](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/storage/azure-storage-blob/samples/blob_samples_containers_async.py#L226):
![image](https://user-images.githubusercontent.com/57166602/187160142-87f157cf-9df4-4a8a-876a-1342caf78809.png)

Reason:
In function get_container_client_from_blob_client()，azure.core.exceptions.ResourceExistsError: The specified container is being deleted. Try operation later. 
There are two same container names in function
`get_blob_client_from_container` and `get_container_client_from_blob_client`.

Suggestion:
Update `blobcontainer` to `myblobcontainer` in function `get_container_client_from_blob_client`.

@lmazuel , @kyle-patterson , @kasobol-msft , @annatisch for notification.

vincenttran-msft:
Hi @v-xuto Tong, thanks for bringing these issues to our attention!

For your first feedback, I can see that having a check such as that could be useful, although the only reason why you should be running into that exception is a timing related issue (i.e. the operation completes before you have a chance to abort it). 

[...]

PR #25943, opened 2022-08-29
[Storage] Update sample files in blob

vincenttran-msft:
As raised in #25918, there are some instances where the sample can face an exception due to timing issues. This PR aims to alleviate those possible timing issues but making the names unique in the test file and also checking the status of a `copy` rather than assuming its state and operating on it directly!


jalauzon-msft (approved):
You beat me to it! Looks good, thanks.

PR #25945, opened 2022-08-29
[Storage] Fix recordings for encryption scope tests

jalauzon-msft:
There was previously a bug in the test proxy recording framework where fake SAS generation would improperly sanitize the `ses` SAS param as if it was `se`. This was not realized when recordings with `ses` were generated and the fix was recently checked into `main`, breaking these tests.

This change re-records all of the Encryption Scope tests with the fix for fake SAS generation.


vincenttran-msft (approved):
Be gone encryption scope SAS failures! Thanks for getting to this 😄

PR #25947, opened 2022-08-29
[Storage] Migrate `test_ors.py`

vincenttran-msft:
This PR migrates `test_ors.py` and `test_ors_async.py`

Side note: When using objection replication policy, things may take awhile to appear. The most consistent manner I've observed is to create the object replication policy, set it on 'Everything' (not just on new files), then reupload the files and then check your `dest` storage account 😄


jalauzon-msft (approved):
Awesome, thanks for getting to these pesky ORS tests!

PR #25950, opened 2022-08-29
[Storage] Adjust default `read_timeout` and README entry

jalauzon-msft:
As brought up after #24031, our previous documentation and default value for the socket-level `read_timout` was not correct. This change aims to improve the documentation and adjusts the default value to a fixed 60 seconds for all packages.


vincenttran-msft (approved):
Thanks for this :) your focus on quality and customer obsession does not go unnoticed!

PR #25956, opened 2022-08-30
Implement attach statsbeat metrics

lzchen:
Things to note:

1. Some statsbeats are exported after a short export interval and some are exported after long intervals. These will be referred to as short metrics and long metrics. The only short metrics are network related statsbeat and everything else is a long metric.
2. Long metrics are exported once on application start, hence the `_STATSBEAT_METER_PROVIDER.force_flush()` call once `collect_statsbeat_metrics` is first called.
3. It is not very easy given OpenTelemetry metrics SDK to configure different metrics to export at different intervals with only ONE `PeriodicExportingMetricReader`. So we achieve this by setting the export interval to "short interval" and maintaining a counter `_long_interval_count` that increments everytime a short interval collect is called. Then once the `_long_interval_count` hits the `long_interval_threshold` (which is how many short intervals are within a long interval), the long metrics are observed. There is no easy way to access when the short metrics "collect" is called, so we simply increment `long_interval_threshold` when `_get_success_metric`, one of the short metrics, is invoked.
4. `_get_attach_metric()` finds which kind of rp the application is running in (`appsvc`, `functions`, `vm`, `None`). This is done through environment variable for the first two, and calling `Azure Meta Data Service` for `vm` via `_get_azure_compute_metadata()`, which is simply a `request`s call to a specific endpoint with specific headers. If the response returns success, we are in a VM, if it raises a timeout exception, we are not in a VM. If it raises some other exception, we try again after.
5. For determining which rp the application is running in, we continuously check which rp it is, even though the usual use case is that it won't change. This is by design, as there are some extreme edge cases in which the rp specified could change (especially in a VM).


sdk/monitor/azure-monitor-opentelemetry-exporter/CHANGELOG.md
```diff
@@ -8,6 +8,8 @@
     ([#25752](https://github.com/Azure/azure-sdk-for-python/pull/25752))
 - Implement all network statsbeat
     ([#25845](https://github.com/Azure/azure-sdk-for-python/pull/25845))
+- Implement attach statsbeat
+    ([#25956](https://github.com/Azure/azure-sdk-for-python/pull/25956))
```
jeremydvoss, line 12: I usually add new changes at the top of the section not the bottom. Guess I'll switch that for this package.

sdk/monitor/azure-monitor-opentelemetry-exporter/azure/monitor/opentelemetry/exporter/statsbeat/_state.py
```diff
@@ -14,7 +14,8 @@
 _STATSBEAT_STATE_LOCK = threading.Lock()
 
 def is_statsbeat_enabled():
-    return not os.environ.get("APPLICATIONINSIGHTS_STATSBEAT_DISABLED_ALL")
+    disabled = os.environ.get("APPLICATIONINSIGHTS_STATSBEAT_DISABLED_ALL")
+    return disabled is None or disabled.lower() != "true"
```
jeremydvoss, line 18: Not the biggest deal since I doubt this keyword will change, but since it's used here and in tests, I think we should use a constant.

sdk/monitor/azure-monitor-opentelemetry-exporter/tests/logs/test_logs.py
```diff
@@ -43,6 +43,7 @@ def setUpClass(cls):
         os.environ[
             "APPINSIGHTS_INSTRUMENTATIONKEY"
         ] = "1234abcd-5678-4efa-8abc-1234567890ab"
+        os.environ["APPLICATIONINSIGHTS_STATSBEAT_DISABLED_ALL"] = "true"
```
jeremydvoss, line 46: constant for tests

sdk/monitor/azure-monitor-opentelemetry-exporter/tests/test_base_exporter.py
```diff
@@ -208,6 +209,12 @@ def test_transmission_200(self):
             result = self._base._transmit(self._envelopes_to_export)
         self.assertEqual(result, ExportResult.SUCCESS)
 
+    @mock.patch.dict(
+        os.environ,
+        {
+        "APPLICATIONINSIGHTS_STATSBEAT_DISABLED_ALL": "false",
+        }
+    )
```
jeremydvoss, line 217: Is there a difference between mocking the envvar like this and assigning it like you do in other places?

jeremydvoss:
"There is no easy way to access when the short metrics "collect" is called"
Curious why this is the case.

sdk/monitor/azure-monitor-opentelemetry-exporter/azure/monitor/opentelemetry/exporter/statsbeat/_statsbeat_metrics.py
```diff
@@ -16,42 +24,29 @@
...
 _ENDPOINT_TYPES = ["breeze"]
 _RP_NAMES = ["appsvc", "functions", "vm", "unknown"]
 
 _HOST_PATTERN = re.compile('^https?://(?:www\\.)?([^/.]+)')
 
 # cSpell:disable
 
-# pylint: disable=unused-argument
-# pylint: disable=protected-access
-def _get_success_count(options: CallbackOptions) -> Iterable[Observation]:
```
jeremydvoss, line 33: I understand that only success_count is moved because it needs to utilize the counter, but I'm wondering if for simplicity we should just move all of the methods into the class

sdk/monitor/azure-monitor-opentelemetry-exporter/tests/statsbeat/test_statsbeat.py
```diff
@@ -73,50 +82,62 @@ def test_collect_statsbeat_metrics_exists(self):
         _statsbeat.collect_statsbeat_metrics(exporter)
         self.assertEqual(_statsbeat._STATSBEAT_METER_PROVIDER, mock_mp)
 
-    def test_collect_statsbeat_metrics_non_eu(self):
+    @mock.patch.object(_StatsbeatMetrics, 'init_non_initial_metrics')
+    @mock.patch.object(_StatsbeatMetrics, '_get_attach_metric')
+    def test_collect_statsbeat_metrics_non_eu(self,  attach_mock, non_init_mock):
```
jeremydvoss, line 87: Bit confused about where these new arguments come from.

sdk/monitor/azure-monitor-opentelemetry-exporter/tests/statsbeat/test_statsbeat.py
```diff
@@ -73,50 +82,62 @@ def test_collect_statsbeat_metrics_exists(self):
...
         self.assertIsNone(_statsbeat._STATSBEAT_METER_PROVIDER)
         with mock.patch.dict(
-                os.environ, {
-                    "APPLICATION_INSIGHTS_STATS_CONNECTION_STRING": "",
-                }):
+            os.environ, {
+                "APPLICATION_INSIGHTS_STATS_CONNECTION_STRING": "",
+            }), \
+            mock.patch.object(_StatsbeatMetrics, 'init_non_initial_metrics'), \
+            mock.patch.object(_StatsbeatMetrics, '_get_attach_metric'):
```
jeremydvoss, line 97: Don't these 2 mocks already happen at the header?

sdk/monitor/azure-monitor-opentelemetry-exporter/tests/statsbeat/test_statsbeat.py
```diff
@@ -73,50 +82,62 @@ def test_collect_statsbeat_metrics_exists(self):
...
+            stats_exporter._endpoint,
+            _statsbeat._DEFAULT_EU_STATS_CONNECTION_STRING.split(";")[1].split("=")[1]   # noqa: E501
+        )
 
+_StatsbeatMetrics_COMMON_ATTRS = dict(
+    _StatsbeatMetrics._COMMON_ATTRIBUTES
+)
+_StatsbeatMetrics_NETWORK_ATTRS = dict(
+    _StatsbeatMetrics._NETWORK_ATTRIBUTES
+)
```
jeremydvoss, line 140: If this is only needed for TestStatsbeatMetrics but not TestStatsbeat, is there a reason we don't put it in the TestStatsbeatMetrics  setup?

sdk/monitor/azure-monitor-opentelemetry-exporter/azure/monitor/opentelemetry/exporter/statsbeat/_statsbeat_metrics.py
jeremydvoss: Not an approval blocker. Just a thought

sdk/monitor/azure-monitor-opentelemetry-exporter/azure/monitor/opentelemetry/exporter/statsbeat/_state.py
lzchen: I think tests do not count as "used in multiple places". If we ever use this keyword somewhere else in the source code itself then we can think about putting it in constants.

sdk/monitor/azure-monitor-opentelemetry-exporter/tests/test_base_exporter.py
lzchen: Mocking the envvar will only set the variable for that test. Assigning it like in `test_logs` will set the env var for the entire test suite.

lzchen:
> "There is no easy way to access when the short metrics "collect" is called"
Curious why this is the case.

I've raised this issue in the community. The only way to do it is to use some sort of global state (which is what we are doing) that callbacks have access to. I've tried exploring adding `kwargs` to `CallbackOptions` in OpenTelemetry metrics sdk but was met with pushback.

sdk/monitor/azure-monitor-opentelemetry-exporter/tests/statsbeat/test_statsbeat.py
lzchen: `setUp` is run for every single test. We created those two variables to hold the ORIGINAL values of `_COMMON_ATTRIBUTES` and `_NETWORK_ATTRIBUTES` so we can reset them every time a test is run.

sdk/monitor/azure-monitor-opentelemetry-exporter/azure/monitor/opentelemetry/exporter/statsbeat/_statsbeat_metrics.py
lzchen: I think you're right. Moved all of the methods into the class.

sdk/monitor/azure-monitor-opentelemetry-exporter/tests/statsbeat/test_statsbeat.py
jeremydvoss: Got it. I feel like it would make sense to put those values inside the TestStatsbeatMetrics class before setUp.

sdk/monitor/azure-monitor-opentelemetry-exporter/tests/test_base_exporter.py
jeremydvoss: Got it

sdk/monitor/azure-monitor-opentelemetry-exporter/tests/statsbeat/test_statsbeat.py
lzchen: Yeah we could. Just chose not too because accessing `TestStatsbeatMetrics` static method is kind of weird. Not a huge opinion though.

jeremydvoss (approved):

Issue #25936, opened 2022-08-29
[Python AMQP] fix pylint on feature/eventhub/pyproto branch


swathipil:
- fix pylint errors and:
- uncomment this section: https://github.com/Azure/azure-sdk-for-python/blob/dcacd0cb2372d3a47fd78a0e9fcd1e738d562070/eng/pipelines/templates/steps/build-artifacts.yml#L83
- remove eventhub and checkpointstoreblob-aio: https://github.com/Azure/azure-sdk-for-python/blob/dcacd0cb2372d3a47fd78a0e9fcd1e738d562070/eng/tox/allowed_pylint_failures.py#L63

swathipil:
addressed

Issue #25875, opened 2022-08-25
[Python AMQP] add timeout/msg_timeout/kwargs TODOs


swathipil:
For parity with uamqp:
- [ ] Add a `msg_timeout` kwarg to the [pyamqp.SendClient](https://github.com/Azure/azure-sdk-for-python/blob/dcacd0cb2372d3a47fd78a0e9fcd1e738d562070/sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/client.py#L398) like in the [uamqp.SendClient](https://github.com/Azure/azure-uamqp-python/blob/375b99b1c6aa890270bde6af2e2cc9e1630d8f0e/uamqp/client.py#L488)
- [ ] Add a `timeout` kwarg to the [pyamqp.ReceiveClient](https://github.com/Azure/azure-sdk-for-python/blob/dcacd0cb2372d3a47fd78a0e9fcd1e738d562070/sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/client.py#L618) like in the [uamqp.ReceiveClient](https://github.com/Azure/azure-uamqp-python/blob/375b99b1c6aa890270bde6af2e2cc9e1630d8f0e/uamqp/client.py#L889)
- [ ] Check that there are two separate constants in pyamqp + add as needed for: 
 > 1. MAX_FRAME SIZE_BYTES - the maximum frame size. This [already exists in pyamqp](https://github.com/Azure/azure-sdk-for-python/blob/dcacd0cb2372d3a47fd78a0e9fcd1e738d562070/sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/constants.py#L60). Same constant in uamqp [[here](https://github.com/Azure/azure-uamqp-python/blob/375b99b1c6aa890270bde6af2e2cc9e1630d8f0e/uamqp/constants.py#L16)].
 > 2. MAX_MESSAGE_LENGTH_BYTES - The maximum payload size _within_ the frame. This doesn't exist yet in pyamqp, so figure out what this should be. Same constant in uamqp [[here](https://github.com/Azure/azure-uamqp-python/blob/375b99b1c6aa890270bde6af2e2cc9e1630d8f0e/uamqp/constants.py#L17)].

[...]

PR #25193, opened 2022-07-13
[EventHub] merge uamqp and pyamqp

swathipil:
Merging uamqp (from main) and pyamqp (from feature/eventhub/pyproto):
- Made update to pyamqp ReceiveClient in `_client_run`
- moved all pyamqp and uamqp specific out to "Transport" object created in clients to be passed around.
TODO:
- [x] update tests to test both uamqp and pyamqp + get passing
- [x] merge with most current main branch with buffered producer changes
- [x] keep in sync with draft PR here for uamqp bits: #24259

ADDRESS ALL OF ANNA'S COMMENTS:
- [ ] Make BatchMessage transport agnostic: https://github.com/Azure/azure-sdk-for-python/pull/25494#discussion_r944096418
- [ ] use generate sas token method from pyamqp.utils instead of _utils.py: https://github.com/Azure/azure-sdk-for-python/pull/25494#discussion_r940498638
- [ ] add TODO in pyamqp handler.open() that they should take an existing connection: https://github.com/Azure/azure-sdk-for-python/pull/25494#discussion_r945936475
- [ ] check that there is an async SharedConnectionManager in pyamqp: https://github.com/Azure/azure-sdk-for-python/pull/25494#discussion_r940636843
- [ ] add TODO in pyamqp that SenderClient should take msg_timeout: https://github.com/Azure/azure-sdk-for-python/pull/25494#discussion_r944086240
- [ ] add TODO in pyamqp that ReceiveClient should take timeout: https://github.com/Azure/azure-sdk-for-python/pull/25494#discussion_r945941080
- [ ] add both MAX_MESSAGE_LENGTH_BYTES and MAX_FRAME_SIZE_BYTES to pyamqp: https://github.com/Azure/azure-sdk-for-python/pull/25494#discussion_r944084248

PR #25965, opened 2022-08-30
[EventHubs] add amqp switch support

### LINKED ISSUE #21246: [AMQP Python] Enabling switching between uamqp and pyamqp in EventHub
in the EH we want to preview with pure python amqp, we should provide a keyword argument to allow users toggle between uamqp and pyamqp as the underlying networking implementation.

### background:

provide a way for users to switch between uamqp and pyamqp implementation in the azure-eventhub sdk

### scope

- allow users to switch between uamqp and pyamqp as the underlying networking protocol implementation when creating the producer/consumer client, could be a keyword argument on the client constructor.
  - beware of the difference between client construction/method call/config variable creation
  - non-blocking-release goal: maximize code reusability in EH when making use of the two different libs

4/27
Tasks:
- [x] Create design gist for interface that uamqp/pyamqp can plug into that producers/consumers use. Discuss with Kashif.
- [x] Initially, if difficult to do otherwise, allow `uamqp_transport` keywords in both Credentials + clients + EventData. Can think of ways to remove this from the credential later.
- [x] If possible, move Amqp Message construction when sending from EventData/AmqpAnnotatedMessage to inside the producer. (This should be possible.)
- [x] If possible, move Credential internals create to inside the producer. (This might be difficult without completely re-creating the object. Though, might be okay to do this since the credential will only be passed in once).
- [ ] push interface changes to main so that we can plug in pyamqp easily when we merge the pyproto branch.

swathipil:
This PR is for adding switch support to the `feature/eventhub/pyproto` changes including the uamqp switch from current `main`.
fixes #21246
Addressing Anna's comments from uamqp switch PR (#25193) + main changes:
- [x] Moving pyamqp logic out to the PyamqpTransport
- [x] **Confirmed: The size of encoded pyamqp.Message is larger than uamqp.Message.** I thought otherwise b/c I was adding the header/property objects by default when building the outgoing uamqp message, even if all values inside those are None. I've fixed this.
- [ ] Make BatchMessage transport agnostic: https://github.com/Azure/azure-sdk-for-python/pull/25494#discussion_r944096418
  - Instead, updated EventDataBatch so that it takes an amqp_transport. If EventDataBatch is manually created and uses PyamqpTransport, inside `send_batch()`, if the producer client transport uses UamqpTransport, the BatchMessage corresponding to the client's amqp_transport will be built and sent.
- [X] Add `message` property to `EventData`/`EventDataBatch`, which return `LegacyMessage`/`LegacyMessageBatch` from `_pyamqp` for backcompat.
- [x] add `connection=None` parameter to `pyamqp.AMQPClient.open()` as per [this discussion](https://github.com/Azure/azure-sdk-for-python/pull/25494#discussion_r945936475)
- [x] Add an async SharedConnectionManager in pyamqp: https://github.com/Azure/azure-sdk-for-python/pull/25494#discussion_r940636843

Issue created [[here](https://github.com/Azure/azure-sdk-for-python/issues/25875)] to address the below TODOs in a separate PR:
- [x] add TODO in pyamqp that SenderClient should take msg_timeout: https://github.com/Azure/azure-sdk-for-python/pull/25494#discussion_r944086240
- [x] add TODO in pyamqp that ReceiveClient should take timeout: https://github.com/Azure/azure-sdk-for-python/pull/25494#discussion_r945941080
- [x] add both MAX_MESSAGE_LENGTH_BYTES and MAX_FRAME_SIZE_BYTES to pyamqp: https://github.com/Azure/azure-sdk-for-python/pull/25494#discussion_r944084248

TODO:
- [ ] fix mypy/pylint issues
  - made partial progress. fix rest in separate PR for issue: #25936
- [x] investigate how to remove `message` property from public API.
  - mark as deprecated and log a deprecation warning if accessed


sdk/eventhub/azure-eventhub/azure/eventhub/_eventprocessor/_eventprocessor_mixin.py
```diff
@@ -99,7 +99,9 @@ def _context(self, links=None):
         if span_impl_type is None:
             yield
         else:
-            child = span_impl_type(name="Azure.EventHubs.process", kind=SpanKind.CONSUMER, links=links)
```
swathipil, line 102: NOTE: ALL FILES IN _EVENTPROCESSOR FOLDER CAN BE IGNORED. No actual changes here.

sdk/eventhub/azure-eventhub/azure/eventhub/_connection_string_parser.py
```diff
@@ -66,7 +66,9 @@ def parse_connection_string(conn_str):
     :type conn_str: str
```
swathipil, line 66: NO ACTUAL CHANGES MADE TO THIS FILE

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_eventprocessor/_ownership_manager.py
```diff
@@ -52,8 +52,7 @@ def __init__(
         self.partition_id = partition_id
```
swathipil, line 52: NO ACTUAL CHANGES MADE TO THIS _eventprocessor FOLDER

sdk/eventhub/azure-eventhub/azure/eventhub/_buffered_producer/_buffered_producer.py
```diff
@@ -0,0 +1,218 @@
...
+    def __init__(
+        self,
+        producer: EventHubProducer,
+        partition_id: str,
+        on_success: Callable[["SendEventTypes", Optional[str]], None],
+        on_error: Callable[["SendEventTypes", Optional[str], Exception], None],
+        max_message_size_on_link: int,
+        executor: ThreadPoolExecutor,
+        *,
+        amqp_transport: AmqpTransport,
```
swathipil, line 35: Comparing against `main`: Added this param.

sdk/eventhub/azure-eventhub/azure/eventhub/_buffered_producer/_buffered_producer.py
```diff
@@ -0,0 +1,218 @@
...
+            # add single event into current batch
+            self._cur_batch.add(events)
+        except AttributeError:  # if the input events is a EventDataBatch, put the whole into the buffer
+            # if there are events in cur_batch, enqueue cur_batch to the buffer
+            with self._lock:
+                if self._cur_batch:
+                    self._buffered_queue.put(self._cur_batch)
+                self._buffered_queue.put(events)
+            # create a new batch for incoming events
+            self._cur_batch = EventDataBatch(self._max_message_size_on_link, amqp_transport=self._amqp_transport)
```
swathipil, line 118: Comparing against `main`: passing in `amqp_transport` here.

sdk/eventhub/azure-eventhub/azure/eventhub/_buffered_producer/_buffered_producer.py
```diff
@@ -0,0 +1,218 @@
...
+        self._running = False
+        self._cur_batch: Optional[EventDataBatch] = None
+        self._max_message_size_on_link = max_message_size_on_link
+        self._check_max_wait_time_future = None
+        self.partition_id = partition_id
+        self._amqp_transport = amqp_transport
+
+    def start(self):
+        with self._lock:
+            self._cur_batch = EventDataBatch(self._max_message_size_on_link, amqp_transport=self._amqp_transport)
```
swathipil, line 58: Comparing against `main`: passing in `amqp_transport` here.

sdk/eventhub/azure-eventhub/azure/eventhub/_buffered_producer/_buffered_producer.py
```diff
@@ -0,0 +1,218 @@
...
+                            "Failed to flush {!r} within {}".format(
+                                self.partition_id, timeout_time
+                            )
+                        )
+                    break
+            # after finishing flushing, reset cur batch and put it into the buffer
+            self._last_send_time = time.time()
+            #reset buffered count
+            self._cur_buffered_len = 0
+            self._cur_batch = EventDataBatch(self._max_message_size_on_link, amqp_transport=self._amqp_transport)
```
swathipil, line 196: Comparing against `main`: We're now passing in `amqp_transport` here.

sdk/eventhub/azure-eventhub/azure/eventhub/_buffered_producer/_buffered_producer_dispatcher.py
```diff
@@ -0,0 +1,187 @@
...
+    def __init__(
+        self,
+        partitions: List[str],
+        on_success: Callable[["SendEventTypes", Optional[str]], None],
+        on_error: Callable[["SendEventTypes", Optional[str], Exception], None],
+        create_producer: Callable[..., EventHubProducer],
+        eventhub_name: str,
+        max_message_size_on_link: int,
+        *,
+        amqp_transport: AmqpTransport,
```
swathipil, line 34: Comparing against `main`: Added an `amqp_transport` parameter here.

sdk/eventhub/azure-eventhub/azure/eventhub/_buffered_producer/_buffered_producer_dispatcher.py
```diff
@@ -0,0 +1,187 @@
...
+        self._on_success = on_success
+        self._on_error = on_error
+        self._create_producer = create_producer
+        self._eventhub_name = eventhub_name
+        self._max_message_size_on_link = max_message_size_on_link
+        self._partition_resolver = PartitionResolver(self._partition_ids)
+        self._max_wait_time = max_wait_time
+        self._max_buffer_length = max_buffer_length
+        self._existing_executor = False
+        self._amqp_transport = amqp_transport
```
swathipil, line 51: Comparing against `main`: Added this line.

sdk/eventhub/azure-eventhub/azure/eventhub/_buffered_producer/_buffered_producer_dispatcher.py
```diff
@@ -0,0 +1,187 @@
...
+                buffered_producer = BufferedProducer(
+                    self._create_producer(pid),
+                    pid,
+                    self._on_success,
+                    self._on_error,
+                    self._max_message_size_on_link,
+                    executor=self._executor,
+                    max_wait_time=self._max_wait_time,
+                    max_buffer_length=self._max_buffer_length,
+                    amqp_transport = self._amqp_transport,
```
swathipil, line 93: Comparing against `main`: We're now passing in `amqp_transport` here.

sdk/eventhub/azure-eventhub/azure/eventhub/_buffered_producer/_partition_resolver.py
```diff
@@ -0,0 +1,287 @@
+# --------------------------------------------------------------------------------------------
```
swathipil, line 1: No changes made to this file compared to main.

sdk/eventhub/azure-eventhub/azure/eventhub/_client_base.py
```diff
@@ -261,9 +264,28 @@ def get_token(self, *scopes, **kwargs):  # pylint:disable=unused-argument
         return AccessToken(signature, expiry)
 
 
+if TYPE_CHECKING:
```
swathipil, line 267: had to add another TYPE_CHECKING block here, b/c mypy was raising an error here if we included the "EventHubSharedKeyCredential", even with forward referencing.

sdk/eventhub/azure-eventhub/azure/eventhub/_client_base.py
```diff
@@ -534,13 +521,16 @@ def _open(self):
...
-            self._handler.open()
+            conn = self._client._conn_manager.get_connection(  # pylint: disable=protected-access
+                host=self._client._address.hostname, auth=auth
+            )
+            self._handler.open(connection=conn)
             while not self._handler.client_ready():
                 time.sleep(0.05)
             self._max_message_size_on_link = (
-                self._handler._link.remote_max_message_size
-                or constants.MAX_FRAME_SIZE_BYTES
```
swathipil, line 542: created an issue here to add a MAX_MESSAGE_LENGTH_BYTES constant to pyamqp: https://github.com/Azure/azure-sdk-for-python/issues/25875

For now, MAX_MESSAGE_LENGTH_BYTES = MAX_FRAME_SIZE_BYTES for pyamqp

sdk/eventhub/azure-eventhub/azure/eventhub/_transport/_base.py
```diff
@@ -0,0 +1,289 @@
...
+        :keyword bool debug: Required.
+        :keyword str encoding: Required.
+        """
+
+    @staticmethod
+    @abstractmethod
+    def close_connection(connection):
+        """
+        Closes existing connection.
+        :param connection: uamqp or pyamqp Connection.
```
l0lawrence, line 114: Does this need a type? and below too

sdk/eventhub/azure-eventhub/azure/eventhub/_transport/_base.py
```diff
@@ -0,0 +1,289 @@
...
+        :param last_exception: Exception to raise if message timed out. Only used by uamqp transport.
+        :param logger: Logger.
+        """
+
+    @staticmethod
+    @abstractmethod
+    def set_message_partition_key(message, partition_key, **kwargs):
+        """Set the partition key as an annotation on a uamqp message.
+
+        :param message: The message to update.
```
l0lawrence, line 159: type here?

l0lawrence: Oh are we not adding them b/c there is a diff between uamqp and pyamqp?

l0lawrence (commented):
Just some questions, this is super awesome Swathi!

sdk/eventhub/azure-eventhub/azure/eventhub/_configuration.py
```diff
@@ -30,14 +31,20 @@ def __init__(self, **kwargs):
         self.max_batch_size = kwargs.get("max_batch_size", self.prefetch)  # type: int
         self.receive_timeout = kwargs.get("receive_timeout", 0)  # type: int
         self.send_timeout = kwargs.get("send_timeout", 60)  # type: int
-        self.custom_endpoint_address = kwargs.get("custom_endpoint_address")  # type: Optional[str]
+        self.custom_endpoint_address = kwargs.pop(
```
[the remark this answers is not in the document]
swathipil: so for this one, I'm actually popping it b/c we modify the custom endpoint address in config, and then pass that new one in to different places. However, if we use get here, we'll get an error that says we're passing it in twice. I'm popping this to avoid that error.

sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
```diff
@@ -502,33 +526,59 @@ def __init__(self, max_size_in_bytes=None, partition_id=None, partition_key=None
...
+            max_size_in_bytes or self._amqp_transport.MAX_MESSAGE_LENGTH_BYTES
         )
-        return "EventDataBatch({})".format(batch_repr)
 
-    def __len__(self):
+        self._count = 0
+        self._internal_events: List[
+            Union[EventData, AmqpAnnotatedMessage]
+        ] = []
+        self._uamqp_message = None if PyamqpTransport.TIMEOUT_FACTOR == self._amqp_transport.TIMEOUT_FACTOR else self._message
```
swathipil, line 545: If PyamqpTransport, then we want to construct a LegacyBatchMessage. Otherwise, I set this to the uamqp BatchMessage.

sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
```diff
@@ -541,18 +591,27 @@ def _load_events(self, events):
                     "the Event Hub frame size limit. Please send a smaller collection of EventData "
                     "or use EventDataBatch, which is guaranteed to be under the frame size limit"
                 )
+    
+    @property
+    def message(self) -> Union["uamqp_BatchMessage", LegacyBatchMessage]:
```
swathipil, line 596: In EventData, I'm always returning LegacyMessage, b/c we don't store uamqp Messages in EventData. However, here, since we do store the uamqp.BatchMessage in self._message, I'm also returning that.

Will double check that message getters/setters are not showing up in docs/apiview.

sdk/eventhub/azure-eventhub/azure/eventhub/_configuration.py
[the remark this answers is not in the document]
swathipil: jk just double checked and it looks like this doesn't actually affect anything. good catch! changing this back.

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_message_backcompat.py
```diff
@@ -49,16 +49,17 @@ def __init__(self, message, **kwargs):
         self.delivery_no = kwargs.get('delivery_no')
         self.delivery_tag = kwargs.get('delivery_tag') or None
         self.on_send_complete = None
-        self.properties = LegacyMessageProperties(self._message.properties)
+        self.properties = LegacyMessageProperties(self._message.properties) if self._message.properties else None
```
swathipil, line 52: This resulted in an error if self._message.properties was None, so updated this.

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_transport.py
```diff
@@ -165,6 +165,7 @@ def connect(self):
             self._init_socket(
                 self.socket_settings, self.read_timeout, self.write_timeout,
             )
+            self.sock.settimeout(0.2)
```
swathipil, line 168: @annatisch - This seems to be a sweet spot for SSLTransport. Previously, the default timeout was 1, and this resulted in really slow receives when a partition was empty. I'm not sure whether this is the best solution to this, or a temporary one. When trying this out with Kashif, we realized that this "sweet spot" might change depending on the connection/latency. Is this reasonable for now? Do you know if there's a better way to resolve this?

sdk/eventhub/azure-eventhub/azure/eventhub/_transport/_base.py
swathipil: I didn't, just for the sake of time and b/c it's internal. But added an issue to the backlog to track: #26070.

Lmk if you think we should def fix these/get them in before the GA/release?

sdk/eventhub/azure-eventhub/azure/eventhub/_transport/_base.py
swathipil: Yeah it can be either uamqp or pyamqp Messages. So I didn't add the type. I did add the type in their respective transports though.

sdk/eventhub/azure-eventhub/azure/eventhub/_transport/_uamqp_transport.py
```diff
@@ -0,0 +1,650 @@
...
+
+        @staticmethod
+        def to_outgoing_amqp_message(annotated_message):
+            """
+            Converts an AmqpAnnotatedMessage into an Amqp Message.
+            :param AmqpAnnotatedMessage annotated_message: AmqpAnnotatedMessage to convert.
+            :rtype: uamqp.Message
+            """
+            message_header = None
+            if annotated_message.header and any(annotated_message.header.values()):
```
swathipil, line 127: @annatisch - previously, we had:
```
            if annotated_message.header:
```
and were adding a header, even if there were no actual values in it. Updated this and found that pyamqp encoded messages are bigger, as you mentioned.

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_buffered_producer/_buffered_producer_async.py
```diff
@@ -0,0 +1,220 @@
...
+        self,
+        producer: EventHubProducer,
+        partition_id: str,
+        on_success: Callable[["SendEventTypes", Optional[str]], Awaitable[None]],
+        on_error: Callable[
+            ["SendEventTypes", Optional[str], Exception], Awaitable[None]
+        ],
+        max_message_size_on_link: int,
+        *,
+        amqp_transport: AmqpTransportAsync,
```
swathipil, line 36: added this line.

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_buffered_producer/_buffered_producer_async.py
```diff
@@ -0,0 +1,220 @@
...
+        self._running = False
+        self._cur_batch: Optional[EventDataBatch] = None
+        self._max_message_size_on_link = max_message_size_on_link
+        self._check_max_wait_time_future = None
+        self.partition_id = partition_id
+        self._amqp_transport = amqp_transport
+
+    async def start(self):
+        async with self._lock:
+            self._cur_batch = EventDataBatch(self._max_message_size_on_link, amqp_transport=self._amqp_transport)
```
swathipil, line 58: passing in amqp_transport

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_buffered_producer/_buffered_producer_async.py
```diff
@@ -0,0 +1,220 @@
...
+            # add single event into current batch
+            self._cur_batch.add(events)
+        except AttributeError:  # if the input events is a EventDataBatch, put the whole into the buffer
+            # if there are events in cur_batch, enqueue cur_batch to the buffer
+            async with self._lock:
+                if self._cur_batch:
+                    self._buffered_queue.put(self._cur_batch)
+                self._buffered_queue.put(events)
+            # create a new batch for incoming events
+            self._cur_batch = EventDataBatch(self._max_message_size_on_link, amqp_transport=self._amqp_transport)
```
swathipil, line 118: passing in amqp_transport

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_buffered_producer/_buffered_producer_async.py
```diff
@@ -0,0 +1,220 @@
...
+        except ValueError:
+            # add single event exceeds the cur batch size, create new batch
+            async with self._lock:
+                self._buffered_queue.put(self._cur_batch)
+            self._cur_batch = EventDataBatch(self._max_message_size_on_link, amqp_transport=self._amqp_transport)
```
swathipil, line 123: passing in amqp_transport

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_buffered_producer/_buffered_producer_async.py
```diff
@@ -0,0 +1,220 @@
...
+        async with self._lock:
+            await self._flush(timeout_time, raise_error)
+
+    async def _flush(self, timeout_time=None, raise_error=True):
+        # pylint: disable=protected-access
+        # try flushing all the buffered batch within given time
+        _LOGGER.info("Partition: %r started flushing.", self.partition_id)
+        if self._cur_batch:  # if there is batch, enqueue it to the buffer first
+            self._buffered_queue.put(self._cur_batch)
+            self._cur_batch = EventDataBatch(self._max_message_size_on_link, amqp_transport=self._amqp_transport)
```
swathipil, line 152: passing in amqp_transport

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_buffered_producer/_buffered_producer_async.py
```diff
@@ -0,0 +1,220 @@
...
+                        "Failed to flush {!r} within {}".format(
+                            self.partition_id, timeout_time
+                        )
+                    )
+                break
+        # after finishing flushing, reset cur batch and put it into the buffer
+        self._last_send_time = time.time()
+        #reset curr_buffered
+        self._cur_buffered_len = 0
+        self._cur_batch = EventDataBatch(self._max_message_size_on_link, amqp_transport=self._amqp_transport)
```
swathipil, line 199: passing in amqp_transport

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_buffered_producer/_buffered_producer_dispatcher_async.py
```diff
@@ -0,0 +1,175 @@
...
+        partitions: List[str],
+        on_success: Callable[["SendEventTypes", Optional[str]], Awaitable[None]],
+        on_error: Callable[
+            ["SendEventTypes", Optional[str], Exception], Awaitable[None]
+        ],
+        create_producer: Callable[..., EventHubProducer],
+        eventhub_name: str,
+        max_message_size_on_link: int,
+        *,
+        amqp_transport: AmqpTransportAsync,
```
swathipil, line 36: added this line

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_buffered_producer/_buffered_producer_dispatcher_async.py
```diff
@@ -0,0 +1,175 @@
...
+            except KeyError:
+                buffered_producer = BufferedProducer(
+                    self._create_producer(partition_id=pid),
+                    pid,
+                    self._on_success,
+                    self._on_error,
+                    self._max_message_size_on_link,
+                    max_wait_time=self._max_wait_time,
+                    max_buffer_length=self._max_buffer_length,
+                    amqp_transport=self._amqp_transport,
```
swathipil, line 84: passing in amqp_transport

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_buffered_producer/_partition_resolver_async.py
```diff
@@ -0,0 +1,33 @@
+# --------------------------------------------------------------------------------------------
```
swathipil, line 1: file not touched

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_connection_manager_async.py
```diff
@@ -0,0 +1,115 @@
+# --------------------------------------------------------------------------------------------
```
swathipil, line 1: added this file from main + modified for pyamqp

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_producer_async.py
```diff
@@ -134,53 +131,93 @@ async def _send_event_data(
...
         last_exception: Optional[Exception] = None,
     ) -> None:
-        # TODO: Correct uAMQP type hints
         if self._unsent_events:
-            await self._open()
-            timeout = timeout_time - time.time() if timeout_time else 0
-            await self._handler.send_message_async(self._unsent_events[0], timeout=timeout)
-            self._unsent_events = None
+            await self._amqp_transport.send_messages_async(
+                self, timeout_time, last_exception, _LOGGER
```
kashifkhan, line 136: unless I missed it along the way, the timeout no longer needs default of 0 or delta between current time ?

sdk/eventhub/azure-eventhub/tests/unittest/test_event_data.py
```diff
@@ -118,11 +164,162 @@ def test_event_data_from_message():
...
+    assert amqp_message.properties.reply_to_group_id == message.properties.reply_to_group_id
+    assert amqp_message.header.time_to_live == message.header.ttl
+    assert amqp_message.header.delivery_count == message.header.delivery_count
+    assert amqp_message.header.first_acquirer == message.header.first_acquirer
+    assert amqp_message.header.durable == message.header.durable
+    assert amqp_message.header.priority == message.header.priority
+    assert amqp_message.annotations == message.message_annotations
+
+def test_legacy_message(uamqp_transport):
+    if uamqp_transport:
```
swathipil, line 248: added test for testing LegacyMessage and LegacyBatchMessage

sdk/eventhub/azure-eventhub/CHANGELOG.md
```diff
@@ -1,6 +1,6 @@
 # Release History
 
-## 5.8.0b6 (Unreleased)
+## 5.8.0b1 (Unreleased)
```
kashifkhan, line 3: Youll have to update `_version.py` as well I think

sdk/eventhub/azure-eventhub/azure/eventhub/_transport/_base.py
l0lawrence: Eventually yes, but doesn't need to be for this PR

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_producer_async.py
swathipil: 🤔 hmm, I must've missed this. Why do we no longer need this timeout?

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_client_base_async.py
```diff
@@ -156,6 +229,13 @@ def __init__(
             fully_qualified_namespace=fully_qualified_namespace,
             eventhub_name=eventhub_name,
             credential=self._credential,
+            uamqp_transport=uamqp_transport,
```
annatisch, line 232: This needs to be passed in?

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_client_base_async.py
```diff
@@ -323,12 +367,14 @@ async def _management_request_async(self, mgmt_msg: Message, op_type: bytes) ->
                 await mgmt_client.close_async()
 
     async def _get_eventhub_properties_async(self) -> Dict[str, Any]:
-        mgmt_msg = Message(application_properties={"name": self.eventhub_name})
+        mgmt_msg = mgmt_msg = self._amqp_transport.build_message(
```
annatisch, line 370: double assignment :)

sdk/eventhub/azure-eventhub/azure/eventhub/_transport/_pyamqp_transport.py
```diff
@@ -0,0 +1,561 @@
...
+        :keyword retry_policy: Required.
+        :keyword keep_alive_interval: Required.
+        :keyword str client_name: Required.
+        :keyword dict link_properties: Required.
+        :keyword properties: Required.
+        """
+
+        target = kwargs.pop("target")
+        # TODO: extra passed in to pyamqp, but not used. should be used?
+        msg_timeout = kwargs.pop("msg_timeout")  # pylint: disable=unused-variable  # TODO: not used by pyamqp?
```
annatisch, line 238: Is the plan to investigate this in this PR or separate? (I'd probably vote separate...)

sdk/eventhub/azure-eventhub/azure/eventhub/_transport/_pyamqp_transport.py
```diff
@@ -0,0 +1,561 @@
...
+        :param ~azure.eventhub._producer.EventHubProducer producer: The producer with handler to send messages.
+        :param int timeout_time: Timeout time.
+        :param last_exception: Exception to raise if message timed out. Only used by uamqp transport.
+        :param logger: Logger.
+        """
+        # pylint: disable=protected-access
+        producer._open()
+        timeout = timeout_time - time.time() if timeout_time else 0
+        producer._handler.send_message(producer._unsent_events[0], timeout=timeout)
+        producer._unsent_events = None
```
annatisch, line 263: This seems to be resetting `unsent_events` even if there might have been more than one event there?

sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py
```diff
@@ -107,17 +192,124 @@ def __init__(
...
+            event,
+            partition_id=kwargs.get("partition_id"),
+            partition_key=partition_key,
+            timeout_time=timeout_time,
+        )
+
     def _get_partitions(self):
         # type: () -> None
         if not self._partition_ids:
-            _LOGGER.debug("Populating partition IDs so producers can be started.")
```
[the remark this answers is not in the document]
kashifkhan: I decided to move them from info level to debug, so that if the issue arises in the future where we want to see which path send ended up taking, it would be helpful. That line would be handy in situations where the producer was open, then closed and then reopened and something went wonky after wards.

My choice is to keep them there :)

sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
swathipil: as per Johan + Anna: adding deprecation warnings to this instead.

sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
```diff
@@ -52,23 +53,36 @@
     AmqpMessageHeader,
     AmqpMessageProperties,
 )
+from ._pyamqp._message_backcompat import LegacyMessage, LegacyBatchMessage
+from ._pyamqp.message import Message
+from ._transport._pyamqp_transport import PyamqpTransport
 
-from ._pyamqp import constants, utils as pyutils
-from ._pyamqp.message import BatchMessage, Message
```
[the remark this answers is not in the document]
swathipil: We no longer need to import it here, since the BatchMessage is built inside the `amqp_transport.build_batch_message()` method called inside the EventDataBatch constructor.

sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
```diff
@@ -543,16 +596,38 @@ def _load_events(self, events):
...
+        """
+        DEPRECATED: Get the underlying uamqp.BatchMessage or LegacyBatchMessage.
+         This is deprecated and will be removed in a later release.
+        :rtype: uamqp.BatchMessage or LegacyBatchMessage
+        """
+        warnings.warn("The `message` property is deprecated and will be removed in future versions.", DeprecationWarning)
+        if not self._uamqp_message:
+            message = AmqpAnnotatedMessage(message=Message(*self._message))
+            self._uamqp_message = LegacyBatchMessage(
+                message, to_outgoing_amqp_message=PyamqpTransport().to_outgoing_amqp_message
```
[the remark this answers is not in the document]
swathipil: This method only takes a pyamqp Message, I believe? so, 2 cases:
1) The transport is UamqpTransport. In which case, self._uamqp_message will already be set in the constructor and this won't be built here. (And I think passing in the uamqp Message to this would result in an error anyway?)
2) The transport is PyamqpTransport. In which case, self._uamqp_message will be set to None in the constructor, and this will always need the PyamqpTransport.to_outgoing_amqp_message).

Does this sound reasonable?

sdk/eventhub/azure-eventhub/azure/eventhub/_connection_manager.py
```diff
@@ -105,7 +127,7 @@ def reset_connection_if_broken(self):
 
 def get_connection_manager(**kwargs):
     # type: (...) -> 'ConnectionManager'
-    connection_mode = kwargs.get("connection_mode", _ConnectionMode.SeparateConnection)
+    connection_mode = kwargs.get("connection_mode", _ConnectionMode.SeparateConnection)  # type: ignore
```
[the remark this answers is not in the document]
swathipil: We haven't. We actually don't have any tests for uamqp using SharedConnection either and we never have afaik. :( Created an issue here: #26274

sdk/eventhub/azure-eventhub/azure/eventhub/_consumer.py
```diff
@@ -220,22 +194,17 @@ def receive(self, batch=False, max_batch_size=300, max_wait_time=None):
...
         )
         self._receive_start_time = self._receive_start_time or time.time()
-        deadline = self._receive_start_time + (max_wait_time or 0)  # max_wait_time can be None
+        deadline = self._receive_start_time + (
+            max_wait_time or 0
+        )
         if len(self._message_buffer) < max_batch_size:
-            # TODO: the retry here is a bit tricky as we are using low-level api from the amqp client.
-            #  Currently we create a new client with the latest received event's offset per retry.
-            #  Ideally we should reuse the same client reestablishing the connection/link with the latest offset.
```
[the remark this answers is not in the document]
swathipil: I actually don't know what this comment was referencing. I don't see a new client being created on retry, but I'm not 100% sure so will leave this in here for now? (Since everything is working, can we leave this in and investigate later? or should we investigate this before we GA?)

sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py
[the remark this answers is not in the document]
swathipil: leaving in.

sdk/eventhub/azure-eventhub/azure/eventhub/_producer.py
```diff
@@ -103,115 +111,151 @@ def __init__(self, client, target, **kwargs):
...
         )
 
-    def _open_with_retry(self):
-        # type: () -> None
+    def _open_with_retry(self) -> None:
         return self._do_retryable_operation(self._open, operation_need_param=False)
 
-    def _send_event_data(self, timeout_time=None):
-        # type: (Optional[float]) -> None
+    def _on_outcome(
```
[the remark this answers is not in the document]
swathipil: yep, added this back in for uamqp use and double checked that this is only used by uamqp. I believe this is b/c the uamqp Message has an `on_send_complete` callback, while pyamqp Message doesn't. Is that an intentional design choice for pyamqp? If not, I can create an issue to investigate.

sdk/eventhub/azure-eventhub/azure/eventhub/_producer.py
```diff
@@ -233,22 +277,21 @@ def send(
...
+
+                if not wrapper_event_data:
+                    return
+
+                self._unsent_events = [wrapper_event_data._message]  # pylint: disable=protected-access
                 if child:
                     self._client._add_span_request_attributes(  # pylint: disable=protected-access
                         child
                     )
+                self._send_event_data_with_retry(timeout=timeout)
```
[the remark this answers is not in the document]
swathipil: No , not part of BP. So originally we were just calling `self._handler.send_messages` in the `send` method in pyamqp. The `send_event_data_with_retry` method wasn't being used at all (and it is in uamqp EH). I modified the `send_event_data_with_retry` method so it uses the logic that was in the `send` method. 

I think the only change in behavior is that it retries, but the underlying logic is the same.

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_client_base_async.py
swathipil: ah yeah, this is then passed into Configuration inside the `ClientBase` constructor. And is used to build the hostname/custom_endpoint_address, which are different for uamqp vs. pyamqp

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_client_base_async.py
swathipil: 🤦

sdk/eventhub/azure-eventhub/azure/eventhub/_transport/_pyamqp_transport.py
swathipil: yep! separate PR and issue created for this here: https://github.com/Azure/azure-sdk-for-python/issues/25875

sdk/eventhub/azure-eventhub/azure/eventhub/_transport/_pyamqp_transport.py
swathipil: ah will fix this, to set to None only if the list is >1

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_transport.py
swathipil: sounds good! just updated in the _transport.py file:
1) `TIMEOUT_INTERVAL = 0.2` - this will be used for SSLTransport now
2) `WS_TIMEOUT_INTERVAL = 1` - this has been added b/c this interval works for WebSocketTransport.

sdk/eventhub/azure-eventhub/azure/eventhub/_transport/_pyamqp_transport.py
annatisch: Gotcha - maybe we should add a comment to this code describing that.
It probably wouldn't hurt to refactor at some point - but I don't think this is the time :)

sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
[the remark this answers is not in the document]
annatisch: Yes I think it seems reasonable... though does this property work okay if I'm using the uamqp transport? As it seems to be creating an instance of the pyamqp Message on line 607

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_transport.py
annatisch: The new constant does not appear to be used?
And it looks like it should be added to the `init_socket` method, with some kind of kwarg override.

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_client_base_async.py
annatisch: I was figuring we'd just get the distinction from the transport that's passed in on line 233. Though it's internal so it doesn't matter too much.

sdk/eventhub/azure-eventhub/azure/eventhub/_producer.py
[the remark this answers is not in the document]
annatisch: Does this mean that `self._outcome` and `self._condition` aren't needed for pyamqp? Or they are set via the "on_complete" callback?

sdk/eventhub/azure-eventhub/azure/eventhub/_common.py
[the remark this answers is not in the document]
swathipil: TL;DR: yes, this property works for both uamqp transport and pyamqp transport. `test_legacy_message` in `unittest/test_event_data.py` checks for this. When it's using the UamqpTransport, the pyamqp Message will never be created - `self._uamqp_message` will be set in the constructor and that if block will not be entered.

More info:
upon further digging, the following properties never worked for uamqp.BatchMessage: `get_data(), settled, delivery_annotations, encode_message(), footer, get_message()`. (Tested with EH 5.9.0 and uamqp 1.5.2.) `print(event_data_batch.message)` also resulted in an AttributeError.

It looks BatchMessage was inheriting a few of these attributes from Message, but these were never actually tested. (For ex. [`get_data()` checks for `self._message`](https://github.com/Azure/azure-uamqp-python/blob/4b1cb844e1c62366600e3770c3dde4a7334cb5d9/uamqp/message.py#L491), but BatchMessage doesn't have a `_message` attribute.)

Given that these never worked, I'm assuming that customers never used/called these properties. So, I only added tests to check that the working properties still work. (Although, the properties that don't work in the uamqp BatchMessage actually do work in the LegacyBatchMessage.)

sdk/eventhub/azure-eventhub/azure/eventhub/_producer.py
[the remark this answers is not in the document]
swathipil: no they're not needed at all for pyamqp.

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_client_base_async.py
swathipil: ah, originally, I wasn't going to pass in the uamqp_transport kwarg at all and do what you said. When I tried it, it seemed a little messier inside the ClientBase constructor, so I decided to just pass it in. Before:
``python
class ClientBase:
def __init__(...):
  # check for async AmqpTransport first
  self._amqp_transport = kwargs.pop("amqp_transport", None)
  uamqp_transport = False
  
  # if PyamqpTransportAsync
  if self._amqp_transport and self._amqp_transport.TIMEOUT_FACTOR == PyamqpTransport.TIMEOUT_FACTOR:
    uamqp_transport = True
  elif not self._amqp_transport:  # if sync transports
    uamqp_transport = kwargs.pop("uamqp_transport", False)
    if uamqp_transport and not UamqpTransport:
      raise ValueError("...import uamqp...")
    self._amqp_transport = UamqpTransport if uamqp_transport else PyamqpTransport
```
Although I might've been making it more complicated than necessary?

annatisch: This seems complicated...
I just can't see why we're tracking the boolean, and why we're checking sync vs async here?

swathipil: TL;DR: I removed this line, and in line 218 of `_client_base_async.py`, I changed `uamqp_transport = kwargs.pop("uamqp_transport", False)` to `uamqp_transport = kwargs.get("uamqp_transport", False)`. This way, the boolean is not passed in explicitly. It's passed in implicitly in `kwargs`. Answering your questions below.

1) We're tracking the boolean (i.e. passing it into `Configuration`) so that we can set two properties in the configuration object: 1) hostname and 2) custom_endpoint_address. I'm not sure if this is what you were thinking, but we're not setting a new `uamqp_transport` property in Configuration (i.e. we use `uamqp_transport` locally in the Configuration constructor, and don't set self.uamqp_transport).
2) In the snippet above, I assumed that `uamqp_transport` was not being passed in inside the ClientBaseAsync constructor to super(). So, just in that snippet, I was checking if the amqp_transport was passed in, b/c that would tell us that it was being called by ClientBaseAsync. And if that was the case, then we would still need to set `uamqp_transport` based on the type of the `amqp_transport` (to later be passed into Configuration). The only reason we were checking sync vs. async is that we'd need to set uamqp_transport if async. If sync, we can call `kwargs.pop("uamqp_transport", False).

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_transport.py
swathipil: ah sorry, just passed `connect_timeout` into `init_socket` as a kwarg, and set that final `settimeout` in `init_socket` to connect_timeout.

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_client_base_async.py
annatisch: That sounds like it really needs a refactor! :D
But it's not public, so we don't need to block the PR on it.

annatisch (approved):

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_client_base_async.py
swathipil: added to issue #26226

PR #25968, opened 2022-08-30
[Storage] Migrate `test_block_blob_sync_copy.py` to test proxy

vincenttran-msft:
This PR migrates `test_block_blob_sync_copy.py` and `test_block_blob_sync_copy_async.py` to test proxy 😄


sdk/storage/azure-storage-blob/tests/test_block_blob_sync_copy.py
```diff
@@ -1,35 +1,33 @@
-# coding: utf-8
-
 # -------------------------------------------------------------------------
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License. See License.txt in the project root for
 # license information.
 # --------------------------------------------------------------------------
 import pytest
```
jalauzon-msft, line 6: nit: `pytest` goes down above `azure.core.exceptions` since it's a library and not a built-in. Simlarly, for async.

jalauzon-msft (approved):
Looks great! One minor comment about import ordering.

sdk/storage/azure-storage-blob/tests/test_block_blob_sync_copy.py
vincenttran-msft: Good catch! The `py` in pytest always gets me..

PR #25970, opened 2022-08-31
Update ci.yml

kristapratico:
The CI has been timing out over the past 2 weeks. Seeing if bumping it another 30 mins helps.


scbedd (approved):

scbedd:
@kristapratico it looks like it's still timing out :| But that timeout is in `analyze`. I'm wondering if `TestTimeOutInMinutes` only applies to the test run. I'm checking that now.

scbedd:
@kristapratico yep. That's why you're seeing the timeout. I just pushed a commit that should resolve this. Other than those transport errors, you should be green now.

Longer term, I think we should probably break cogservices up into multiple `ci.yml`.

kristapratico:
AHhh I see, thanks @scbedd!

PR #25978, opened 2022-08-31
[Storage] Migrate `test_blob_service_stats` and `test_large_block_blob` to test proxy

jalauzon-msft:
This change migrates `test_blob_service_stats`, `test_blob_service_stats_async`, `test_large_block_blob`, and `test_large_block_blob_async` to test proxy.

`test_blob_service_stats` was changed a bit to stop using overridden responses.
`test_large_block_blob` is all live only.


vincenttran-msft (approved):
Thanks for getting to these two! Blob is getting closer to completion 😄

PR #25981, opened 2022-08-31
[Storage] Migrate `test_logging.py` to test proxy

vincenttran-msft:
This PR migrates `test_logging.py` and `test_logging_async.py` to test proxy! Additionally, some tests were previously marked as live-only due to using SAS, but are now recorded and run in both live and recorded mode thanks to test proxy extending our testing capabilities 😄


jalauzon-msft (approved):
Look good. Thanks for recording the SAS tests!

PR #25999, opened 2022-09-01
[formrecognizer] Documentation and typing updates

catalinaperalta:
Did an in-depth pass over the models and clients in preparation for GA. Updating documentation and typing to be more accurate.


sdk/formrecognizer/azure-ai-formrecognizer/samples/v3.2/advanced_samples/async_advanced_samples/sample_get_elements_with_spans_async.py
```diff
@@ -86,7 +86,7 @@ async def get_elements_with_spans_async():
     # This example uses DocumentTable, but other elements that also have a `spans` or `span` field
```
kristapratico, line 86: just wondering - is it still necessary to have this sample under an advanced samples dir?

kristapratico (approved):
🚀

sdk/formrecognizer/azure-ai-formrecognizer/samples/v3.2/advanced_samples/async_advanced_samples/sample_get_elements_with_spans_async.py
catalinaperalta: I dont think so, we did that to try and track some interest but I think we'll rely more on the PM team for that going forward. I'll change this in the samples PR I have open.

PR #26005, opened 2022-09-01
[Storage] Fix flaky retry test once and for all (hopefully)

jalauzon-msft:
After many failed attempts to fix the flaky read timeout retry test, this change will hopefully be the final fix. This changes the test to essentially mock raising a timeout exception rather than relying on actually causing a read timeout as causing a real read timeout was difficult to achieve in the Live test environment (likely because inter-datacenter requests were just too fast).

I tried to make the raised timeout exceptions as close as possible to what they would be in a real timeout scenario to test that our retry mechanism is catching them. This meant wrapping the correct underlying transport's (`requests` or `aiohttp`) read timeout exception with the core `ServiceResponseError`.

Hopefully this should eliminate the flakiness in this test for good.

### DIFF
--- a/sdk/storage/azure-storage-blob/tests/test_retry.py
+++ b/sdk/storage/azure-storage-blob/tests/test_retry.py
@@ -11,35 +11,35 @@ from azure.core.exceptions import (
     AzureError,
     ClientAuthenticationError,
     HttpResponseError,
-    ResourceExistsError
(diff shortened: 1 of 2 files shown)


vincenttran-msft (approved):
LGTM! Hopefully this is the nail in the coffin!

PR #26007, opened 2022-09-01
Use SkipCheckoutNone parameter name for sparse checkout

benbp:


scbedd (approved):
I didn't see any other usages in the repo either when I looked, thanks Ben. 👍

PR #26009, opened 2022-09-01
Feature stats

lzchen:


sdk/monitor/azure-monitor-opentelemetry-exporter/tests/statsbeat/test_statsbeat.py
```diff
@@ -407,6 +432,82 @@ def test_get_azure_compute_metadata_vm_retry(self):
...
+            True,
+            2,
+        )
+        metric._long_interval_count_map[_FEATURE_METRIC_NAME[0]] = 2
+        observations = metric._get_feature_metric(options=None)
+        self.assertEqual(len(observations), 1)
+        self.assertEqual(metric._long_interval_count_map[_FEATURE_METRIC_NAME[0]], 0)
+
+    # pylint: disable=protected-access
+    def test_get_feature_metric_does_not_meet_threshold(self):
```
jeremydvoss, line 453: Does "does not meet threshold" just mean it's only short metrics?

sdk/monitor/azure-monitor-opentelemetry-exporter/azure/monitor/opentelemetry/exporter/statsbeat/_statsbeat_metrics.py
```diff
@@ -38,6 +39,18 @@
...
 _HOST_PATTERN = re.compile('^https?://(?:www\\.)?([^/.]+)')
 
+
+class _FEATURE_TYPES:
+    FEATURE = 0
+    INSTRUMENTATION = 1
+
+
+class _StatsbeatFeature:
+    NONE = 0
```
jeremydvoss, line 49: Why do we need a NONE feature?

sdk/monitor/azure-monitor-opentelemetry-exporter/tests/statsbeat/test_statsbeat.py
lzchen: Yup, since feature metric is a long metric we don't expect get_feature_metric() to observe anything in the case in which it does not meet the threshold.

sdk/monitor/azure-monitor-opentelemetry-exporter/azure/monitor/opentelemetry/exporter/statsbeat/_statsbeat_metrics.py
lzchen: Encoding the features enabled into a 64-bit long needs to have a baseline 0 value, which represents all the bits turned off, which represents no features are enabled (which is what NONE represents).

sdk/monitor/azure-monitor-opentelemetry-exporter/tests/statsbeat/test_statsbeat.py
jeremydvoss: I'm trying to figure out if there's a better term we should use in all of this instead of threshold. Threshold sounds like there's some tolerance we've breached, like for latency, rather than an interval ratio. But for now, it's fine.

jeremydvoss (approved):

PR #26023, opened 2022-09-03
Instrumentation statsbeat

lzchen:
We use a bit map for instrumentations just like for feature.
We update the instrumentation map only when we see a signal using the instrumentation (currently, there is only tracing signal so we only check for instrumentations generating span). In the future, we will also update the map when using logging (for logging instrumentation) and for metrics (for metric instrumentations like `system-metrics`.


sdk/monitor/azure-monitor-opentelemetry-exporter/azure/monitor/opentelemetry/exporter/_utils.py
```diff
@@ -40,6 +41,26 @@ def ns_to_duration(nanoseconds):
...
+_INSTRUMENTATIONS_BIT_MASK_LOCK = threading.Lock()
+
+def get_instrumentations():
+    return _INSTRUMENTATIONS_BIT_MASK
+
+
+def add_instrumentation(instrumentation_name):
+    with _INSTRUMENTATIONS_BIT_MASK_LOCK:
+        global _INSTRUMENTATIONS_BIT_MASK  # pylint: disable=global-statement
+        instrumentation = _INSTRUMENTATIONS_BIT_MAP.get(instrumentation_name, 0)
```
jeremydvoss, line 54: "instrumentation" would be the bit representing that instrumentation, correct? I think this variable name could be clearer.

sdk/monitor/azure-monitor-opentelemetry-exporter/azure/monitor/opentelemetry/exporter/export/trace/_exporter.py
```diff
@@ -48,7 +49,7 @@
 class AzureMonitorTraceExporter(BaseExporter, SpanExporter):
     """Azure Monitor Trace exporter for OpenTelemetry."""
 
-    def export(self, spans: Sequence[Span], **kwargs: Any) -> SpanExportResult: # pylint: disable=unused-argument
+    def export(self, spans: Sequence[ReadableSpan], **kwargs: Any) -> SpanExportResult: # pylint: disable=unused-argument
```
jeremydvoss, line 52: Could you explain the difference between these 2 Span types? What is a non-readable span?

sdk/monitor/azure-monitor-opentelemetry-exporter/azure/monitor/opentelemetry/exporter/statsbeat/_statsbeat_metrics.py
```diff
@@ -197,12 +204,22 @@ def _get_feature_metric(self, options: CallbackOptions) -> Iterable[Observation]
...
-        observations.append(Observation(1, dict(attributes)))
+        if self._feature is not _StatsbeatFeature.NONE:
+            attributes = dict(_StatsbeatMetrics._COMMON_ATTRIBUTES)
+            attributes.update(_StatsbeatMetrics._FEATURE_ATTRIBUTES)
+            observations.append(Observation(1, dict(attributes)))
+
+        # instrumentation metric
+        # Don't send observation if no instrumentations enabled
+        instrumentations = _utils.get_instrumentations()
+        if instrumentations != 0:
```
jeremydvoss, line 217: Similar here. If we could name this something like "instrumentation_bits" or "instrumentation_bit_map", I think it would be clearer.

sdk/monitor/azure-monitor-opentelemetry-exporter/tests/statsbeat/test_statsbeat.py
```diff
@@ -79,19 +86,23 @@ def test_collect_statsbeat_metrics_exists(self):
...
         exporter = mock.Mock()
         exporter._instrumentation_key = "1aa11111-bbbb-1ccc-8ddd-eeeeffff3333"
         exporter._endpoint = "https://westus-0.in.applicationinsights.azure.com/"
         self.assertIsNone(_statsbeat._STATSBEAT_METER_PROVIDER)
         with mock.patch.dict(
             os.environ, {
                 "APPLICATION_INSIGHTS_STATS_CONNECTION_STRING": "",
-            }), \
-            mock.patch.object(_StatsbeatMetrics, 'init_non_initial_metrics'), \
-            mock.patch.object(_StatsbeatMetrics, '_get_attach_metric'):
```
jeremydvoss, line 94: Could you explain these removals?

sdk/monitor/azure-monitor-opentelemetry-exporter/azure/monitor/opentelemetry/exporter/export/trace/_exporter.py
lzchen: We were simply using the wrong typing as defined in this [interface](https://github.com/open-telemetry/opentelemetry-python/blob/main/opentelemetry-sdk/src/opentelemetry/sdk/trace/export/__init__.py#L61). `Span` is the base class in the sdk, `ReadableSpan` is the non-editable span that we send to exporters.

jeremydvoss (approved):

PR #26062, opened 2022-09-07
[CognitiveLanguage] Strip trailing endpoint slashes

pvaneck:
# Description

When the user passes in an endpoint with a trailing slash to the various clients, the resulting URIs for the various operations begin with two leading '/'. For example:

"https://languagesdkresource.cognitiveservices.azure.com//language/..."

Although these URIs still resolve, it's better not to depend on this and ensure that the URIs are constructed correctly. Here, we ensure that when a user passes in an endpoint to any of the Cognitive Language clients, any trailing slashes are removed.


kristapratico (approved):
Looks good 🚀

PR #26073, opened 2022-09-07
[Test Proxy] Fix error with recorded_test fixture

pvaneck:
In the case of live testing with recording disabled, the `recorded_test` pytest fixture tries to unpack a NoneType object. Here, some small refactoring is done where the `is_live_and_not_recording` check is moved to the beginning of `recorded_test`.

Example error:
```bash
>       test_id, recording_id, variables = start_proxy_session()
E       TypeError: cannot unpack non-iterable NoneType object
```

### DIFF
--- a/tools/azure-sdk-tools/devtools_testutils/proxy_fixtures.py
+++ b/tools/azure-sdk-tools/devtools_testutils/proxy_fixtures.py
@@ -113,7 +113,7 @@ def environment_variables(test_proxy: None) -> EnvironmentVariableSanitizer:
 
 
 @pytest.fixture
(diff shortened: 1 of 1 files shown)


pvaneck:
@mccoyp Friendly ping :)
 Would be good to have this in so that live tests using the `recorded_test` pytest fixture can be verified in the CI where recording is disabled.

mccoyp (approved):
Thank you for catching this! This makes a lot more sense, honestly. (And thanks for pinging me, sorry about the delay here 🙂)

pvaneck:
@mccoyp Thanks for the review!

PR #26074, opened 2022-09-07
Adjust how we discover the "root" of the repository when booting the test-proxy

scbedd:
@mccoyp 

The previous method was dependent on `azure-sdk-tools` installed in `-e` mode. The fallback path ascension logic worked in _most_ cases, but on my machine it just spun forever instead of booting the docker instance.

This adjusts to use the `request` fixture -> retrieve collected tests -> ascend to repo root from one of the collected tests -> use that as the repo root.

It is much less hacky, and should work wherever we invoke tests from, even outside the repo.

### DIFF
--- a/tools/azure-sdk-tools/devtools_testutils/proxy_startup.py
+++ b/tools/azure-sdk-tools/devtools_testutils/proxy_startup.py
@@ -171,7 +171,7 @@ def start_test_proxy(request) -> None:
     set_custom_default_matcher(excluded_headers=headers_to_ignore)
 
 
-def stop_test_proxy(request) -> None:
+def stop_test_proxy() -> None:
     """Stops any running instance of the test proxy"""
 
     if not PROXY_MANUALLY_STARTED:
--- a/tools/azure-sdk-tools/devtools_testutils/proxy_startup.py
+++ b/tools/azure-sdk-tools/devtools_testutils/proxy_startup.py
@@ -28,25 +28,23 @@ WINDOWS_IMAGE_SOURCE_PREFIX = "azsdkengsys.azurecr.io/engsys/testproxy-win"
 CONTAINER_STARTUP_TIMEOUT = 6000
 PROXY_MANUALLY_STARTED = os.getenv("PROXY_MANUAL_START", False)
 
-REPO_ROOT = os.path.abspath(os.path.join(os.path.abspath(__file__), "..", "..", "..", ".."))
 PROXY_CHECK_URL = PROXY_URL.rstrip("/") + "/Info/Available"
 TOOL_ENV_VAR = "PROXY_PID"
 
 
-def get_image_tag() -> str:
+def get_image_tag(repo_root: str) -> str:
     """Gets the test proxy Docker image tag from the target_version.txt file in /eng/common/testproxy"""
     version_file_location = os.path.relpath("eng/common/testproxy/target_version.txt")
-    version_file_location_from_root = os.path.abspath(os.path.join(REPO_ROOT, version_file_location))
+    version_file_location_from_root = os.path.abspath(os.path.join(repo_root, version_file_location))
 
     try:
         with open(version_file_location_from_root, "r") as f:
             image_tag = f.read().strip()
-
     # In live pipeline tests the root of the repo is in a different location relative to this file
     except FileNotFoundError:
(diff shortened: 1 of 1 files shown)


tools/azure-sdk-tools/devtools_testutils/proxy_startup.py
```diff
@@ -57,6 +55,30 @@ def get_image_tag() -> str:
...
+    """
+    if os.path.isfile(start_dir_or_file):
+        current_dir = os.path.dirname(start_dir_or_file)
+    else:
+        current_dir = start_dir_or_file
+
+    while current_dir is not None and not (os.path.dirname(current_dir) == current_dir):
+        possible_root = os.path.join(current_dir, ".git")
+
+        # we need to check for assets.json first!
```
scbedd, line 72:
```suggestion
```

tools/azure-sdk-tools/devtools_testutils/proxy_startup.py
```diff
@@ -28,25 +28,23 @@
...
     version_file_location = os.path.relpath("eng/common/testproxy/target_version.txt")
-    version_file_location_from_root = os.path.abspath(os.path.join(REPO_ROOT, version_file_location))
+    version_file_location_from_root = os.path.abspath(os.path.join(repo_root, version_file_location))
 
     try:
         with open(version_file_location_from_root, "r") as f:
             image_tag = f.read().strip()
-
     # In live pipeline tests the root of the repo is in a different location relative to this file
     except FileNotFoundError:
```
scbedd, line 44: I should probably just rip this whole thing out now yeah?

tools/azure-sdk-tools/devtools_testutils/proxy_startup.py
```diff
@@ -57,6 +55,29 @@ def get_image_tag() -> str:
...
+    while current_dir is not None and not (os.path.dirname(current_dir) == current_dir):
+        possible_root = os.path.join(current_dir, ".git")
+
+        # we need the git check to prevent ascending out of the repo
+        if os.path.exists(possible_root):
+            return current_dir
+        else:
+            current_dir = os.path.dirname(current_dir)
+
+    raise Exception('Requested target "{}" does not exist within a git repo.'.format(start_dir_or_file))
```
mccoyp, line 78: nit: I love f-strings
```suggestion
    raise Exception(f'Requested target "{start_dir_or_file}" does not exist within a git repo.')
```

tools/azure-sdk-tools/devtools_testutils/proxy_startup.py
```diff
@@ -148,7 +171,7 @@ def start_test_proxy() -> None:
     set_custom_default_matcher(excluded_headers=headers_to_ignore)
 
 
-def stop_test_proxy() -> None:
+def stop_test_proxy(request) -> None:
```
mccoyp, line 174: Doesn't hurt to accept `request`, but it looks like we can also drop it for now

mccoyp (approved):

tools/azure-sdk-tools/devtools_testutils/proxy_startup.py
scbedd: Know what. you're absolutely right. I should modernize my style on this.

tools/azure-sdk-tools/devtools_testutils/proxy_startup.py
scbedd: Right! Cleaning it. thanks for the catch

scbedd, line 174:
```suggestion
def stop_test_proxy() -> None:
```

tools/azure-sdk-tools/devtools_testutils/proxy_startup.py
```diff
@@ -166,13 +189,13 @@ def stop_test_proxy() -> None:
...
     if is_live_and_not_recording():
         yield
     else:
-        start_test_proxy()
+        start_test_proxy(request)
         # Everything before this yield will be run before fixtures that invoke this one are run
         # Everything after it will be run after invoking fixtures are done executing
         yield
-        stop_test_proxy()
+        stop_test_proxy(request)
```
scbedd, line 201:
```suggestion
        stop_test_proxy()
```

PR #26076, opened 2022-09-07
[Test Proxy] Update environment_variables scope

pvaneck:
This updates the `environment_variables` pytest fixture scope to "session" (as opposed to the default "function" scope) in order to enable users to use it in other session scoped fixtures. Also, for most cases, users will only need one instance of `EnvironmentVariableSanitizer` per test session.


mccoyp (approved):
LGTM, thanks!

PR #26078, opened 2022-09-07
[Asset-Sync] extensions to the test framework to allow external assets

scbedd:
@mccoyp This is ready to rock.

I will remove the new assets.json prior to merging this PR. Just wanted to get green across the board before doing that.


tools/azure-sdk-tools/devtools_testutils/proxy_testcase.py
```diff
@@ -34,6 +36,27 @@
...
+    for root in discovered_roots:
+        current_dir = os.path.dirname(test_id)
+        while current_dir is not None and not (os.path.dirname(current_dir) == current_dir):
+            possible_assets = os.path.join(current_dir, "assets.json")
+            possible_root = os.path.join(current_dir, ".git")
+
+            # we need to check for assets.json first!
+            if os.path.exists(os.path.join(root, possible_assets)):
+                complete_path = os.path.abspath(os.path.join(root, possible_assets))
+                return os.path.relpath(complete_path, root)
```
mccoyp, line 52: Similarly to what we do in [`get_test_id`](https://github.com/Azure/azure-sdk-for-python/blob/ff97aa8a61dc6accc72ee11441f4909c32d7695e/tools/azure-sdk-tools/devtools_testutils/helpers.py#L52), I think we have to add `.replace("\\", "/")` to this path result to make it compatible on Windows. Otherwise, I get
```
The provided assets.json path of \u0022/srv/testproxy/sdk\\keyvault\\azure-keyvault-secrets\\assets.json\u0022 does not exist.
```
when running `azure-keyvault-secrets` tests
```suggestion
                return os.path.relpath(complete_path, root).replace("\\", "/")
```

mccoyp (approved):

PR #26084, opened 2022-09-08
add typing docs

kristapratico:
Part of https://github.com/Azure/azure-sdk-for-python/issues/23351

Adding typing guidance and typing cheat sheet for the repo.

### DIFF
--- a/doc/dev/static_type_checking.md
+++ b/doc/dev/static_type_checking.md
@@ -3,7 +3,7 @@
 This document contains some general typing tips and guidance as it relates to common types/patterns we use in the Python SDK.
 It also walks through the setup necessary to run mypy and pyright, static type checkers, on client library code.
 
-For the TL;DR version, please see the [Static Type Checking Cheat Sheet](TODO).
+For the TL;DR version, please see the [Static Type Checking Cheat Sheet](https://github.com/Azure/azure-sdk-for-python/blob/main/doc/dev/static_type_checking_cheat_sheet.md).
 
 ## Table of contents
   - [Intro to typing in Python](#intro-to-typing-in-python)
@@ -15,7 +15,6 @@ For the TL;DR version, please see the [Static Type Checking Cheat Sheet](TODO).
     - [How to ignore type checking errors](#how-to-ignore-type-checking-errors)
     - [How to opt out of type checking](#how-to-opt-out-of-type-checking)
   - [Typing tips and guidance for the Python SDK](#typing-tips-and-guidance-for-the-python-sdk)
-    - [Debug type checking with reveal_type and reveal_locals](#debug-type-checking-with-reveal_type-and-reveal_locals)
     - [Use typing.Any sparingly](#use-typingany-sparingly)
     - [Use typing.Union when accepting more than one type](#use-typingunion-when-accepting-more-than-one-type)
     - [Use typing.Optional when a parameter can explicitly be None](#use-typingoptional-when-a-parameter-can-explicitly-be-none)
@@ -23,6 +22,7 @@ For the TL;DR version, please see the [Static Type Checking Cheat Sheet](TODO).
       - [Mapping types](#mapping-types)
       - [List](#list)
       - [Tuples](#tuples)
+    - [Parameters not needing type annotations](#parameters-not-needing-type-annotations)
     - [Typing variadic arguments - *args and **kwargs](#typing-variadic-arguments---args-and-kwargs)
     - [Use TYPE_CHECKING to avoid circular imports](#use-type_checking-to-avoid-circular-imports)
     - [Passing a function or a class as a parameter or return type](#passing-a-function-or-a-class-as-a-parameter-or-return-type)
@@ -37,6 +37,7 @@ For the TL;DR version, please see the [Static Type Checking Cheat Sheet](TODO).
     - [Use typing.Literal to restrict based on exact values](#use-typingliteral-to-restrict-based-on-exact-values)
     - [Use typing.NewType to restrict a type to a specific context](#use-typingnewtype-to-restrict-a-type-to-a-specific-context)
     - [Use typing.Final to restrict a type from changing its value](#use-typingfinal-to-restrict-a-type-from-changing-its-value)
+    - [Debug type checking with reveal_type and reveal_locals](#debug-type-checking-with-reveal_type-and-reveal_locals)
   - [Additional Resources](#additional-resources)
 
 ## Intro to typing in Python
@@ -63,18 +64,18 @@ Since Python 3, type annotations were introduced and are the **preferred** metho
 Annotations provide a cleaner syntax and let you keep the type information closer, or inline with the code.
 
 ```python
-from typing import Any, Optional
+from typing import Any, Union
 
 def download_blob_from_url(
         blob_url: str,
         output: str,
-        credential: Optional[Any] = None,
+        credential: Union[AzureNamedKeyCredential, TokenCredential],
         **kwargs: Any
 ) -> None:
     ...
 ```
 
-> Note: Do not use comment style type hints (`# type: Any`). Some of our libraries use type comments due to legacy code supporting Python 2/3, but these will be updated to annotation style.
+> Note: Do not use comment style type hints (`# type: str`). Some of our libraries use type comments due to legacy code supporting Python 2, but these will be updated to annotation style.
 
 A fully annotated signature includes type annotations for all parameters and the return type. The type of a parameter
 should follow the `:` syntax and a default argument can be supplied like in the `credential` parameter above. A return
@@ -149,12 +150,14 @@ else:
 from typing_extensions import Protocol
 ```
 
+If using `typing-extensions`, you must add it to the install dependencies for your library (do not rely on install by `azure-core`). 
 When importing a backported type into code, `typing-extensions` does a try/except on your behalf (either importing
 from `typing` if supported, or `typing-extensions` if the Python version is too old) so there is no need to do this
 check yourself.
 
 See the [typing-extensions](https://github.com/python/typing_extensions) docs to check what has been
-backported.
+backported. This document calls out which types are needed through `typing-extensions` based on support of Python 3.7+.
+Some commonly used types imported from `typing-extensions` are Literal, TypedDict, Protocol, runtime_checkable, and Final.
 
 ## Install and run type checkers on your client library code
 
@@ -274,49 +277,13 @@ from running checks.
 1) Place the package name on this block list: TODO
 2) Open an issue tracking that "library-name" should be opted in to running type checking
 
+> Note: Blocking your library from type checking is a *temporary* state. It is expected that checks are re-enabled as soon as possible.
 
 ## Typing tips and guidance for the Python SDK
 
 This is not intended to be a complete guide to typing in Python. This section covers some common typing scenarios
 encountered when working on the SDK and provides some tips for you to get started. Typing is relatively new to Python,
 with over 20 PEPs and counting. Please check the [typing](https://docs.python.org/3/library/typing.html) documentation for the most up-to-date information.
-
-### Debug type checking with reveal_type and reveal_locals
-
-Sometimes the type checker might raise an error that is difficult to understand -- luckily it is possible to "debug" the type checker and
-see what it is thinking / what types it is inferring.
-`reveal_type(expr)` and `reveal_locals()` are debugging functions recognized by type checkers. `reveal_type(expr)` can be
-placed in code and will tell you the inferred static type of an expression.
-`reveal_locals()` can also be placed on any line and will tell you the inferred types of all local variables.
-
-```python
-a = [1]  #  `int` type is inferred
-reveal_type(a)
-
-c = [1, 'a']  # `int` and `str` are not "duck type" compatible, so type is narrowed to `object`
-reveal_type(c)
-
-
-def hello_world(message: str) -> None:
-    print(f'Hello world! {message}')
-    reveal_locals()
-
-
-reveal_type(hello_world)
-```
-
-Running mypy on the above reveals the types that mypy sees:
-
-```cmd
-main.py:2: note: Revealed type is "builtins.list[builtins.int*]"
-main.py:5: note: Revealed type is "builtins.list[builtins.object*]"
-main.py:9: note: Revealed local types are:
-main.py:9: note:     message: builtins.str
-main.py:11: note: Revealed type is "def (message: builtins.str)"
-```
-
-These debugging functions don't need to be imported from anywhere and are only recognized by the type checkers - therefore you will
-need to remove them from code before runtime.
 
 ### Use typing.Any sparingly
 
@@ -359,11 +326,14 @@ def begin_build_model(
     ...
 ```
 
-A `Union` requires at least two types and should be used with type that are not consistent with each other. For
+A `Union` requires at least two types and should be used with types that are not consistent with each other. For
 example, usage of `Union[int, float]` is not necessary since `int` is consistent with `float` (implements all the supported operations of `float`) -- you can just use `float`. It's
 also recommended avoiding having functions return `Union` types as it causes the user to need to understand or perform `isinstance` checks
 on the return value before acting on it to write type-safe code. Sometimes this can be resolved by using an [overload](#use-typingoverload-to-overload-a-function).
 
+Note that adding support for a `Union` on input is generally non-breaking. However, adding a `Union`
+(or additional options to an existing `Union`) for a return type could be considered breaking and such a change should
+be brought to the attention of the Python architects.
 
 ### Use typing.Optional when a parameter can explicitly be None
 
@@ -395,12 +365,11 @@ Per PEP 484, type checkers have moved towards requiring the Optional type to be 
 
 A few things to think about when typing collections in Python.
 
-1) "Be conservative in what you send, be liberal in what you accept." - The Robustness Principle
-
-   In other words, we should aim to be lenient in the types we accept as a parameter because this allows for more
+1) We should aim to be lenient in the types we accept as a parameter because this allows for more
    flexibility for the caller, e.g. accepting an `Iterable` (where appropriate) allows for more possible types to be
-   passed (list, tuple, dict, etc.) than just specifying a `List`. On the other hand, we should aim to be specific in
-   what we say we return - it's more helpful to specify concrete types which a user can reason about than abstract types.
+   passed (list, tuple, dict, etc.) than just specifying a `List`. For return types, since we support APIs which are
+   expected to evolve over time, it is also helpful to return more flexible types. So long as our return type
+   adheres to a given protocol, we are more likely to not break our user's code.
 
 2) It can be more useful to consider the set of supported operations as the defining characteristic of a type.
    [Structural subtyping](https://en.wikipedia.org/wiki/Structural_type_system) is what supports duck typing in
@@ -467,9 +436,10 @@ At runtime, `employee` is perfectly valid even though it does not conform to a t
 <class 'dict'>
 ```
 
-Note that this creates a plain `dict` at runtime. Usage of `TypedDict` is a great way to inform the type checkers and
-Intellisense how a specific dict should be constructed and help the type checkers warn users if they try to access
-keys which don't exist.
+Note that this creates a plain `dict` at runtime. Due to this, `TypedDict` cannot be used in `isinstance` checks,
+e.g. `isinstance(employee, Employee)` will throw a `TypeError`. Usage of `TypedDict` is a great way to inform the
+type checkers and Intellisense how a specific dict should be constructed and help the type checkers warn users if
+they try to access keys which don't exist.
 
 Use of generic TypedDict is also supported via typing_extensions (>=4.3.0). For example,
 
@@ -487,8 +457,11 @@ class Employee(TypedDict, Generic[T]):
     id: int
     current: bool
     additional_info: List[T]
+
+employee = Employee[str](name="krista", title="swe", id=123, current=True, additional_info=["Redmond"])
 ```
 
+See the [TypedDict](https://docs.python.org/3/library/typing.html#typing.TypedDict) docs for more options.
 
 > Note that TypedDict is backported to older versions of Python by using typing_extensions.
 
@@ -538,6 +511,22 @@ Note that `Point` is consistent with `Tuple[float, float]`, but not vice versa. 
 vanilla `Tuple[float, float]` into the `bounding_box` function above, but if it were annotated
 as `points: Tuple[float, float]`, a `Point` would be permissible. This is because NamedTuple is a class builder
 and may have extra methods or user-defined methods that `Tuple` does not.
+
+### Parameters not needing type annotations
+
+There are some parameters that we do not need to type annotate in the SDK.
+
+The `self` parameter in a class method does not need to be typed. Nor does the `cls` parameter in a `@classmethod`.
+However, it is customary to return `None` on a `__init__` like below since this will indicate that it should be
+type checked (since there are no other typed parameters found in signature).
+
+```python
+class KeyCredential:
+    
+    def __init__(self) -> None:
+        ...
+```
+
 
 ### Typing variadic arguments - *args and **kwargs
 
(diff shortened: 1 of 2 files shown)


doc/dev/static_type_checking.md
```diff
@@ -0,0 +1,1399 @@
...
+### Typing for collections
+
+A few things to think about when typing collections in Python.
+
+1) "Be conservative in what you send, be liberal in what you accept." - The Robustness Principle
+
+   In other words, we should aim to be lenient in the types we accept as a parameter because this allows for more
+   flexibility for the caller, e.g. accepting an `Iterable` (where appropriate) allows for more possible types to be
+   passed (list, tuple, dict, etc.) than just specifying a `List`. On the other hand, we should aim to be specific in
+   what we say we return - it's more helpful to specify concrete types which a user can reason about than abstract types.
```
annatisch, line 403: This is true, however for APIs that we expect to grow over time, a more abstract return type gives us more flexibility.
For example typing the return value as `Iterable` rather than `List` means we can develop the returned object so long as it adheres to the protocol.

doc/dev/static_type_checking.md
```diff
@@ -0,0 +1,1399 @@
...
+
+2) It can be more useful to consider the set of supported operations as the defining characteristic of a type.
+   [Structural subtyping](https://en.wikipedia.org/wiki/Structural_type_system) is what supports duck typing in
```
annatisch, line 406: In the discussion of structural vs nominal typing - would it be valuable to have a section on Protocols vs ABCs?

doc/dev/static_type_checking.md
```diff
@@ -0,0 +1,1399 @@
...
+>>>employee = Employee(name="krista", title="swe", ident="nah", current=True)
+>>>print(employee)
+{'name': 'krista', 'title': 'swe', 'ident': 'nah', 'current': True}
+>>>type(employee)
+<class 'dict'>
+```
+
+Note that this creates a plain `dict` at runtime. Usage of `TypedDict` is a great way to inform the type checkers and
+Intellisense how a specific dict should be constructed and help the type checkers warn users if they try to access
+keys which don't exist.
```
annatisch, line 472: It might be worth explicitly calling out here that because this is a plain `dict` at runtime, you cannot do an `isinstance` check, e.g. `isinstance(employee, Employee)`

doc/dev/static_type_checking.md
```diff
@@ -0,0 +1,1399 @@
...
+T = TypeVar("T")
+
+
+class Employee(TypedDict, Generic[T]):
+    name: str
+    title: str
+    id: int
+    current: bool
+    additional_info: List[T]
+```
```
annatisch, line 490: I'd add a small usage snippet here:
```python
employee: Employee[str] = Employee(name="krista", title="swe", id=123, current=True, additional_info=["Redmond"])
```

doc/dev/static_type_checking.md
```diff
@@ -0,0 +1,1399 @@
...
+
+
+> Note that TypedDict is backported to older versions of Python by using typing_extensions.
```
annatisch, line 493: I'd add a link to the TypedDict docs, as IIRC there's a few other options people could find useful.

doc/dev/static_type_checking.md
```diff
@@ -0,0 +1,1399 @@
...
+
+```python
+result = analyze_text("my text", LanguageDetection())
+reveal_type(result)
+```
+
+`main.py:28: note: Revealed type is "__main__.LanguageDetectionResult"`
+
+`typing.overload` can also be used when there are a different number of typed arguments. We can take the above example and
+imagine that `analyze_text` has an additional overload which does not take an `analysis_kind` (note that we switched to accepting `*args` and `**kwargs` in the actual implementation for this case):
```
annatisch, line 835: Maybe we could explicitly mention here that overloads can be used to present optional parameters as required, or vice versa, in the particular overload scenario.

doc/dev/static_type_checking.md
```diff
@@ -0,0 +1,1399 @@
...
+import sys
+
+if sys.version_info < (3, 9):
+    from typing import Sequence
+else:
+    from collections.abc import Sequence
+```
+
+- The `typing-extensions` library backports types that are introduced only in later versions of Python (e.g. `Protocol`
+  which was introduced in Python 3.8) so that they can be used in earlier Python versions.
```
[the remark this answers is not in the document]
annatisch: I see you've mentioned throughout where typing-extensions are needed - in that case you could either make a summary of that here, or maybe add a comment stating that it's called out below where this import will be needed.

doc/dev/static_type_checking.md
```diff
@@ -0,0 +1,1399 @@
...
+
+```python
+from typing import Optional, Any
+
+def begin_export_project(
+    project_name: str,
+    *,
+    string_index_type: str = "UnicodeCodePoint",  # Param with a default, doesn't allow None, doesn't require Optional
+    asset_kind: Optional[str] = None,  # None is allowed, type as Optional
+    **kwargs: Any
```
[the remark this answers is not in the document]
annatisch: I see you've already addressed some of these below :)

doc/dev/static_type_checking.md
annatisch: haha - I see you've already done this in the protocol section.

doc/dev/static_type_checking.md
kristapratico: Yeah I see what you mean. I think what's written is a good rule of thumb, but maybe not so great for long-lived APIs which cannot have breaking changes. Updating!

doc/dev/static_type_checking_cheat_sheet.md
```diff
@@ -0,0 +1,410 @@
+# Static Type Checking Cheat Sheet
+
+This cheat sheet details guidance for typing as it relates to the Python SDK. Use your own judgment to achieve the best balance of clarity and flexibility for your Python client library. 
+
+### General guidance
+
+- Do provide type annotations (per PEP 484) to public APIs in the client library.
+- You do not need to type annotate all functions in a client library, but you should provide type hints where unit tests are worth writing or where type annotations will assist in understanding of the code.
```
annatisch, line 8: Maybe further clarify that this is referring to internal implementation functions.

doc/dev/static_type_checking_cheat_sheet.md
```diff
@@ -0,0 +1,410 @@
...
+
+@validate_table
+def create_table(table_name: str, **kwargs: Any) -> None:
+    ...
+```
+
+
+### Type Aliases
+
+- Do use type aliases to help make lengthy, complicated type hints more readable or help convey meaning to the reader.
```
annatisch, line 329: I think it's worth making a distinction here between the type hint and the docstring parameter type.
Currently - if we're using type aliases, they should be fully expanded in the docstring. At least, last I checked, using the type alias in the docstring didn't render in a useful way in sphinx.

doc/dev/static_type_checking_cheat_sheet.md
```diff
@@ -0,0 +1,410 @@
...
+        ...
+
+# No: the function calls len(), Iterable doesn't support it
+def create_batch(entities: Iterable[Entity]) -> None:
+    if len(entities) > 1:
+        ...
+```
+
+- If you don't need to mutate the collection, prefer parameter types `Sequence` over `MutableSequence`, `Mapping` over `MutableMapping`, etc.
+- Use `from __future__ import annotations` to use built-in generic collection types (`list` instead of `typing.List`).
```
annatisch, line 183: I think this is repeated from a previous section.

doc/dev/static_type_checking_cheat_sheet.md
```diff
@@ -0,0 +1,410 @@
...
+
+```python
+from typing import TypeVar
+
+T_co = TypeVar("T_co", covariant=True)
+T_contra = TypeVar("T_contra", contravariant=True)
+```
+
+- Use the pre-defined TypeVar, `typing.AnyStr`, when your parameter or return type expects both `str` or `bytes`.
+- Do use `typing.ParamSpec` (a specialized TypeVar) to forward your function type hints from one callable to another callable (e.g. decorators). This preserves type hinting in the original function.
```
annatisch, line 305: Cool - I didn't know about this :)

doc/dev/static_type_checking_cheat_sheet.md
annatisch: Question: If using a type alias in a public API - does the alias itself need to be part of the public API?
I imagine not for use of a static type checker, but what about rendered docs?

annatisch (approved):

doc/dev/static_type_checking.md
```diff
@@ -0,0 +1,1427 @@
...
+
+
+def add_things(x: Any, y: Any) -> Any:
+    return x + y
+```
+
+`Any` should be used sparingly since it effectively turns the type checker off. If the type can be narrowed down to
+something more specific, that is preferred since over usage of `Any` can undermine the type checker's ability to find
+bugs in other parts of the code. Additionally, `Any` does not benefit our customers who may be using type checking with
+our libraries.
```
kashifkhan, line 311: Do we want to provide examples where use of `Any` is ok and in line with the patterns seen in the SDK overall ?

doc/dev/static_type_checking_cheat_sheet.md
kristapratico: Good point - making that distinction. Yeah type alias in the docstring is no good with sphinx.

> If using a type alias in a public API - does the alias itself need to be part of the public API? I imagine not for use of a static type checker, but what about rendered docs?

I don't believe it needs to be public API. edit: tested it and in the rendered docs it expands the type alias to the actual type so we're good here.

doc/dev/static_type_checking.md
kristapratico: I tried to cover this in the variadic arguments (**kwargs) section. Wondering what other examples you're thinking of? I don't think we tend to use `Any` in public APIs (aside from kwargs) from what I've seen.

kashifkhan: yeah I was thinking of other than `kwargs` myself, so was wondering if you had seen any other examples used in the SDK :)

doc/dev/static_type_checking_cheat_sheet.md
```diff
@@ -0,0 +1,410 @@
...
+    from typing import Union, TypeVar, Any
+
+
+# Yes:
+if TYPE_CHECKING:
+    from a import b  # avoiding a circular import
+    from expensive import c  # avoiding runtime costs
+```
+
+- Use `from __future__ import annotations` for forward declarations or when using built-in generic collection types, like `dict` or `list`.
```
gvanrossum, line 84: There are some caveats here. The Python community is torn about whether this will become the default at some point in the future or whether a different approach (PEP 649) will be used instead, at which point this form will be deprecated.

In general I would favor using this as a last resort -- if you have the occasional generic dict and you want to support Python 3.8 or before, it's probably better to use `typing.Dict[str, str]` instead of `dict[str, str]` + future import.

doc/dev/static_type_checking_cheat_sheet.md
```diff
@@ -0,0 +1,410 @@
...
+
+```python
+from __future__ import annotations  # must be first import in file
+
+class Triangle:
+
+    @classmethod
+    def from_shape(cls) -> Triangle: # Don't need to make Triangle a forward reference / string
```
gvanrossum, line 92: The simpler fix would be to use `-> "Triangle"`.

doc/dev/static_type_checking_cheat_sheet.md
```diff
@@ -0,0 +1,410 @@
...
+- Do not silence the type checker with `type: ignore` unless other options are exhausted. Consider first using `typing.cast` or refactoring the code.
+- If you must use a `type: ignore`, try to be specific in what error code you're ignoring.
+
+```python
+# type: ignore[misc]  # mypy ignores only the error code in brackets
+
+# pyright: ignore[reportPrivateUsage]  # pyright ignores only the error code in brackets
+
+# type: ignore  # all errors ignored by both mypy and pyright
+```
```
gvanrossum, line 110: This example might mislead people into thinking that these comments have a global effect. Typically they are placed on a specific line of code, e.g.
```py
class Foo(Any):  # type: ignore # Can't inherit from Any in some versions
    ...
```

doc/dev/static_type_checking_cheat_sheet.md
```diff
@@ -0,0 +1,410 @@
...
+from typing import Union
+
+
+def begin_build_model(
+    self, source: str, build_mode: Union[str, DocumentBuildMode], **kwargs: Any
+) -> DocumentModelAdministrationLROPoller[DocumentModelInfo]:
+    ...
+```
+
+- Try to using `Union` as a return type since this typically makes the user need to do `isinstance` checks. See if `@overload` can be used to improve the typing experience.
```
gvanrossum, line 131: Missing negation
```suggestion
- Try not using `Union` as a return type since this typically makes the user need to do `isinstance` checks. See if `@overload` can be used to improve the typing experience.
```

doc/dev/static_type_checking_cheat_sheet.md
```diff
@@ -0,0 +1,410 @@
...
+
+# type: ignore  # all errors ignored by both mypy and pyright
+```
+
+- Do leave a comment with a link or explanation for the ignore so that it may be fixed later.
+- If you need to ignore type checking all files under a directory for your library, use a `mypy.ini` and `pyrightconfig.json` at the package-level.
+- If you must opt-out of all type checking temporarily, open an issue to re-enable type checking for your library.
+- Try not to use `typing.Any` if it is possible to be more specific. `Any` essentially turns off type checking.
+
+### Unions
```
gvanrossum, line 117: Any word of advice regarding `X | Y` instead of `Union[X, Y]`, and `X | None` instead of `Optional[X]`? See PEP 604.

doc/dev/static_type_checking_cheat_sheet.md
```diff
@@ -0,0 +1,410 @@
...
+
+```python
+from typing import overload, Union
+
+
+@overload
+def analyze(text: str, task: LanguageDetection) -> LanguageDetectionResult:
+    ...
+
+
```
gvanrossum, line 199: No matter what PEP 8 and related tooling say, I would leave only one blank line between the overloads and between the last overload and the implementation.

doc/dev/static_type_checking_cheat_sheet.md
```diff
@@ -0,0 +1,410 @@
...
+@overload
+def analyze(text: str, task: SentimentAnalysis) -> SentimentResult:
+    ...
+
+# actual implementation
+def analyze(text: str, task: Union[SentimentAnalysis, EntityRecognition, LanguageDetection]) -> Union[
+    SentimentResult, EntityRecognitionResult, LanguageDetectionResult]:
```
gvanrossum, line 211: If you format this heading using the black style it'll come out more readable.

doc/dev/static_type_checking_cheat_sheet.md
```diff
@@ -0,0 +1,410 @@
...
+def choose(items: Sequence[T]) -> T:
+    ...
+```
+
+- Try to constrain your `TypeVar`'s by specifying types or using the `bound` keyword argument.
+
+```python
+from typing import TypeVar
+
+T = TypeVar("T", int, str)  # limited to int or str
```
gvanrossum, line 271: I would honestly not even mention this form here. It has some unexpected behaviors and you should be able to specify a union upper bound to get what most users expect from such a function. (If you decide to delete this, the bullet text on line 266 should also be updated.)

doc/dev/static_type_checking_cheat_sheet.md
```diff
@@ -0,0 +1,410 @@
...
+```python
+from typing import Union
+from typing_extensions import Literal
+
+
+# client library code
+class SelectionMark:
+
+    def __init__(self):
+        self.kind: Literal["selectionMark"] = "selectionMark"
```
gvanrossum, line 354: Could you instead show an example that uses the more commonly found form `Literal[lit1, lit2, ...]`? E.g.
```py
Color = Literal["red", "green", "blue"]
```

doc/dev/static_type_checking_cheat_sheet.md
```diff
@@ -0,0 +1,410 @@
...
+
+if ele.kind == "selectionMark":
+    print(ele.state)
+elif ele.kind == "word":
+    print(ele.word)
+elif ele.kind == "line":
+    print(ele.line)
+```
+
+- If you have a `Literal` type that should not change or get re-assigned in the code, consider typing it as `typing.Final`.
```
gvanrossum, line 387: Maybe also mention the `@final` decorator for classes and methods? And `Final` variables can be used even without `Literal`.

gvanrossum (commented):
Hi, I took your invitation and reviewed the cheat sheet. This is a great source of information about type checking!

If you think this is useful I can also try to find time to review the other file. You could also at-mention me if you have specific questions for me.

kristapratico (commented):
Thank you so much for taking the time to review!

doc/dev/static_type_checking_cheat_sheet.md
gvanrossum: That's cool. With `from __future__ import annotations` this could work, but as I wrote there are other concerns about that.

doc/dev/static_type_checking_cheat_sheet.md
```diff
@@ -0,0 +1,410 @@
...
+    # type: (str) -> Table
+    ...
+
+# Yes:
+def create_table(table_name: str) -> Table:
+    ...
+```
+
+- Do fully annotate function signatures - this includes type annotations for all parameters and the return type.
+- Do fully annotate function signatures in all samples - this ensures the type checks will check the sample code.
```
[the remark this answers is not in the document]
gvanrossum: Okay, fine.

doc/dev/static_type_checking.md
```diff
@@ -0,0 +1,1452 @@
...
+def analyze_text(text: str, analysis_kind: LanguageDetection) -> LanguageDetectionResult:
+    ...
+
+
+@overload
+def analyze_text(text: str, analysis_kind: EntityRecognition) -> EntityRecognitionResult:
+    ...
+
+
+@overload
```
swathipil, line 822: super extra tiny nit, but based on GVM's rec, should there be one blank line b/w these methods?

doc/dev/static_type_checking_cheat_sheet.md
```diff
@@ -0,0 +1,399 @@
+# Static Type Checking Cheat Sheet
+
+This cheat sheet details guidance for typing as it relates to the Python SDK. Use your own judgment to achieve the best balance of clarity and flexibility for your Python client library. 
+
+### General guidance
+
+- Do provide type annotations (per PEP 484) to public APIs in the client library.
```
swathipil, line 7:
```suggestion
- Do provide type annotations ([per PEP 484](https://peps.python.org/pep-0484/)) to public APIs in the client library.
```

swathipil (approved):
This is super helpful - I'll be looking at this at least once a day from now on! 👏

doc/dev/static_type_checking.md
```diff
@@ -0,0 +1,1452 @@
...
+and any code without type hints will just be ignored by the type checker. Typing is optional and this is a good
+thing! Pushing to achieve full coverage of annotated code can lead to low signal-to-noise and sometimes is not practical 
+given the expressiveness of Python as a language. So, in practice, what should you aim to type?
+
+1) Add type hints to publicly exposed APIs in the client library. Type hints get shipped with our client libraries (both whl and sdist)
+   and provide benefit to our users. To ensure that type hints do get shipped in your library and can be checked by a static
+   type checker, follow the steps per [PEP 561](https://mypy.readthedocs.io/en/stable/installed_packages.html#creating-pep-561-compatible-packages) below:
+
+    - add an empty `py.typed` file to your package directory. E.g. `.../sdk/azure-core/azure/core/py.typed`
+    - include `py.typed` under `package_data` in your setup.py (`package_data={"azure-core": ["py.typed"]}`) ([example](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/core/azure-core/setup.py))
```
swathipil, line 110: question @kristapratico: it looks like the setup.py file in azure-core has the following line:
`package_data={"pytyped": ["py.typed"]}`.
Here, you say, "azure-core" instead of "pytyped". Is there a difference?

doc/dev/static_type_checking.md
```diff
@@ -0,0 +1,1452 @@
...
+        credential: Union[AzureNamedKeyCredential, TokenCredential],
+        **kwargs: Any
+) -> None:
+    ...
+```
+
+> Note: Do not use comment style type hints (`# type: str`). Some of our libraries use type comments due to legacy code supporting Python 2, but these will be updated to annotation style.
+
+A fully annotated signature includes type annotations for all parameters and the return type. The type of a parameter
+should follow the `:` syntax and a default argument can be supplied like in the `credential` parameter above. A return
```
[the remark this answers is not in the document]
kristapratico: Thanks for catching that. The example changed from feedback, but I missed updating the text!

doc/dev/static_type_checking.md
```diff
@@ -0,0 +1,1452 @@
...
+```
+
+- The `typing-extensions` library backports types that are introduced only in later versions of Python (e.g. `Protocol`
+  which was introduced in Python 3.8) so that they can be used in earlier Python versions.
+
+```python
+from typing_extensions import Protocol
+```
+
+If using `typing-extensions`, you must add it to the install dependencies for your library (do not rely on install by `azure-core`).
```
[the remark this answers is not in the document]
kristapratico: Good idea, done!

doc/dev/static_type_checking.md
```diff
@@ -0,0 +1,1452 @@
...
+ignore_me: int = 5  # type: ignore  https://www.github.com/Azure/azure-sdk-for-python/issues/my-issue
+```
+
+Note that it is possible to be specific in the error to ignore, instead of globally turning off type checking on
+that line. Syntax for this involves putting the specific error code in brackets of the ignore comment. Error codes
+are found next to the type checking error. 
+
+*mypy:*
+```python
+ignore_me: int = 5  # type: ignore[misc]
```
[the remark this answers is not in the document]
kristapratico: Nice! We'll be configuring mypy to show error codes soon so it will be easier to find them.

doc/dev/static_type_checking.md
```diff
@@ -0,0 +1,1452 @@
...
+official [mypy docs](https://mypy.readthedocs.io/en/stable/type_inference_and_annotations.html?highlight=ignore#silencing-type-errors) and [pyright docs](https://github.com/microsoft/pyright/blob/main/docs/comments.md#line-level-diagnostic-suppression).
+
+
+### How to opt out of type checking
+
+All client libraries in the Python SDK repo are automatically opted in to running type checking. If there is a
+reason why a particular library should not run type checking, it is possible to add that library to a block list to prevent mypy/pyright
+from running checks.
+
+1) Place the package name on this block list: TODO
```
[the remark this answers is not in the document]
kristapratico: It will be a new file, e.g. type_checker_ignore_list or something. You will need to add your package to the file if you want to _opt-out_ of type checking in CI.

doc/dev/static_type_checking.md
```diff
@@ -0,0 +1,1452 @@
...
+def begin_build_model(
+    self, source: str, build_mode: Union[str, DocumentBuildMode], **kwargs: Any
+) -> DocumentModelAdministrationLROPoller[DocumentModelInfo]:
+    ...
+```
+
+A `Union` requires at least two types and should be used with types that are not consistent with each other. For
+example, usage of `Union[int, float]` is not necessary since `int` is consistent with `float` (implements all the supported operations of `float`) -- you can just use `float`. It's
+also recommended avoiding having functions return `Union` types as it causes the user to need to understand or perform `isinstance` checks
+on the return value before acting on it to write type-safe code. Sometimes this can be resolved by using an [overload](#use-typingoverload-to-overload-a-function).
```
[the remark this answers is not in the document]
kristapratico: Yeah that is supposed to be the purpose of the cheat sheet. Let me know if you see any gaps!

PR #26096, opened 2022-09-08
[Identity] Move RequestsTransport Import

kashifkhan:
In an environment without `requests` and only `aiohttp` the import `from azure.identity.aio import EnvironmentCredential` throws an `ImportError: requests package is not installed`. 

Moving the RequestsTransports import inside `build_pipeline` so its imported when needed.


mccoyp (commented):
This would fix the immediate problem, but should we even be constructing our own pipeline here? I know we shouldn't for typical SDKs -- @xiangyan99, do you know if Identity is an exception?

kashifkhan:
> This would fix the immediate problem, but should we even be constructing our own pipeline here? I know we shouldn't for typical SDKs -- @xiangyan99, do you know if Identity is an exception?

bumping this up to see how to move forward with this?

PR #26105, opened 2022-09-08
[Storage] Migrate blob encryption tests to test proxy

jalauzon-msft:
This change migrates `test_blob_encryption`, `test_blob_encryption_async`, `test_blob_encryption_v2`, `test_blob_encryption_v2_async` to test proxy.

Some notes:
- To make recording possible, `os.urandom` was mocked to not be random and instead return fixed values for IV, encryption key, and nonce.
- For V1, the mock was applied at a class level since most tests were recorded but for V2, the mock was only applied to the specific test cases that would be recorded. This means that the many live tests will still use random values for more test coverage. Also, for V2 async, mocks were applied using context manager rather than decorator as `async` mock decorators are not supported in Python 3.7.
- For V2, the production encryption code was updated slightly to change the use of `AESGCM.generate_key` to just `os.urandom` to make mocking easier. This code is functionally equivalent. 
- For V1, usage of `content_as_bytes` was removed as that API is deprecated now.


vincenttran-msft (approved):
Thanks for getting to this! I know we've been discussing getting to these for a while, so I'm glad you were able to figure out the mocking and knocked it out of the park with all of these tests 😄

PR #26133, opened 2022-09-09
add additionally allowed tenant ids support

xiangyan99:
Cherry picked the commits.

Removed beta changes:
1. `tenant_id` for `AzureCliCredential`
2. removed `VisualStudioCodeCredential` from `DefaultAzureCredential` token chain.
3. `AZURE_CLIENT_CERTIFICATE_PASSWORD` support for `EnvironmentCredential`
4. `validate_authority` support


sdk/identity/azure-identity/azure/identity/_credentials/authorization_code.py
```diff
@@ -30,6 +30,9 @@ class AuthorizationCodeCredential(GetTokenMixin):
         the authority for Azure Public Cloud (which is the default). :class:`~azure.identity.AzureAuthorityHosts`
         defines authorities for other clouds.
     :keyword str client_secret: One of the application's client secrets. Required only for web apps and web APIs.
+    :keyword list[str] additionally_allowed_tenant_ids: optional additional tenant ids for which the credential
+            may acquire tokens. Add the wildcard value "*" to allow the credential to acquire tokens for
+            any tenant the application is installed.
```
mccoyp, line 35: nit: the current language is a little confusing (at least, to me)
```suggestion
            may acquire tokens. Add the wildcard value "*" to allow the credential to attempt to acquire tokens
            for any tenant.
```

mccoyp (commented):
Did architects prefer `additionally_allowed_tenant_ids` over something like `additionally_allowed_tenants`, and/or is the naming conversation still ongoing?

xiangyan99:
> Did architects prefer `additionally_allowed_tenant_ids` over something like `additionally_allowed_tenants`, and/or is the naming conversation still ongoing?

We are working with the architects about the name. :)

sdk/identity/azure-identity/azure/identity/_credentials/azure_cli.py
```diff
@@ -73,8 +74,11 @@ def get_token(self, *scopes, **kwargs): # pylint: disable=no-self-use
 
         resource = _scopes_to_resource(*scopes)
         command = COMMAND_LINE.format(resource)
-        tenant = resolve_tenant(default_tenant=self.tenant_id, **kwargs)
-
+        tenant = resolve_tenant(
```
mccoyp, line 77: I think I'm a bit confused by this behavior. Because the default tenant for this credential isn't defined (`""`), any `tenant_id` passed to `get_token` will be returned here. That makes sense for the first time a tenant is provided -- we can consider that to be the tenant configured for this credential.

But if a new tenant ID is provided to `get_token` at a later time, without that new tenant being added to `additionally_allowed_tenants`, the tenant would still resolve to the new tenant. Isn't that an issue?

sdk/identity/azure-identity/tests/test_cli_credential.py
```diff
@@ -239,8 +209,5 @@ def fake_check_output(command_line, **_):
         token = credential.get_token("scope")
         assert token.token == expected_token
 
-        with mock.patch.dict(
-            "os.environ", {EnvironmentVariables.AZURE_IDENTITY_DISABLE_MULTITENANTAUTH: "true"}
```
mccoyp, line 243: I could be mistaken, but I don't think this test makes sense anymore without validating the behavior of setting this environment variable. Shouldn't we have a new test that checks for providing a tenant that's:
1. Not configured on the credential
2. Not in the list of allowed tenants

and verify that we raise a `ClientAuthenticationError` in that case?

sdk/identity/azure-identity/azure/identity/_credentials/azure_cli.py
xiangyan99: Took offline

pvaneck (commented):
With https://github.com/Azure/azure-sdk-for-net/pull/31037 merged after a lot of documentation review/revision, should we reflect the same documentation here for consistency? Or is that a follow-up as a part of https://github.com/Azure/azure-sdk-for-python/issues/26159?

sdk/identity/azure-identity/azure/identity/_credentials/authorization_code.py
```diff
@@ -30,6 +30,9 @@ class AuthorizationCodeCredential(GetTokenMixin):
         the authority for Azure Public Cloud (which is the default). :class:`~azure.identity.AzureAuthorityHosts`
         defines authorities for other clouds.
     :keyword str client_secret: One of the application's client secrets. Required only for web apps and web APIs.
+    :keyword list[str] additionally_allowed_tenants: optional additional tenant ids for which the credential
```
mccoyp, line 33: Have you double-checked the generated docs for these type hints? I just wonder since we usually would provide the type hint as `:paramtype additionally_allowed_tenants: List[str]`, using `List` from `typing`

xiangyan99:
> With [Azure/azure-sdk-for-net#31037](https://github.com/Azure/azure-sdk-for-net/pull/31037) merged after a lot of documentation review/revision, should we reflect the same documentation here for consistency? Or is that a follow-up as a part of #26159?

We will need more time to finalize the docs/troubleshooting guideline/etc.

My plan is to merge the code and start the API review with our architects and we will have another PR to update the docs.

sdk/identity/azure-identity/azure/identity/_internal/__init__.py
```diff
@@ -79,10 +82,20 @@ def resolve_tenant(default_tenant, tenant_id=None, **_):
...
-    return tenant_id
+    if not default_tenant:
+        return tenant_id
+    if additionally_allowed_tenants == ['*'] or tenant_id in additionally_allowed_tenants:
+        _LOGGER.info("A token was requested for a different tenant than was configured on the credential, "
+                     "and the requested tenant ID was used to authenticate. Configured tenant ID: %s, "
+                     "Requested tenant ID %s", default_tenant, tenant_id)
+        return tenant_id
+    raise ClientAuthenticationError(
+                message='The current credential is not configured to acquire tokens for tenant {}. '
```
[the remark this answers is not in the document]
xiangyan99: Good catch!

mccoyp (commented):
Aside from my nitpicking, I think this looks good! I don't think I see a negative test though, where we verify that sending an invalid tenant raises the expected error; I think we should have that for at least one credential to validate the behavior of `resolve_tenant`. Feel free to correct me if there is such a test and I just missed it 🙂

sdk/identity/azure-identity/tests/test_aad_client.py
```diff
@@ -304,7 +305,13 @@ def test_multitenant_cache():
...
-    client_c = AadClient(tenant_id=tenant_c, **common_args)
+    client_c = AadClient(tenant_id=tenant_c, additionally_allowed_tenants=['*'], **common_args)
     assert client_c.get_cached_access_token([scope]) is None
     token = client_c.get_cached_access_token([scope], tenant_id=tenant_a)
     assert token.token == expected_token
+
+    # but d does not add target tenant into allowed list therefore fail
+    client_d = AadClient(tenant_id=tenant_d, **common_args)
+    assert client_d.get_cached_access_token([scope]) is None
+    with pytest.raises(ClientAuthenticationError):
```
mccoyp, line 316: Can we also add an `assert`ion that the raised exception mentions "additionally_allowed_tenants"? Just to be sure that we preserve an informative exception message

sdk/identity/azure-identity/azure/identity/_credentials/azure_cli.py
```diff
@@ -36,12 +36,13 @@ class AzureCliCredential(object):
 
     This requires previously logging in to Azure via "az login", and will use the CLI's currently logged in identity.
 
-    :keyword str tenant_id: optional tenant to include in the token request.
+    :keyword List[str] additionally_allowed_tenants: optional additional tenant ids for which the credential
```
mccoyp, line 39: Same (nitpicking) comment from earlier about capitalizing the O in "optional" -- this applies to all of the docstrings with `additionally_allowed_tenants` 🙂

sdk/identity/azure-identity/azure/identity/_credentials/azure_powershell.py
```diff
@@ -51,7 +51,14 @@ class AzurePowerShellCredential(object):
     """Authenticates by requesting a token from Azure PowerShell.
 
     This requires previously logging in to Azure via "Connect-AzAccount", and will use the currently logged in identity.
+
+    :keyword List[str] additionally_allowed_tenants: optional additional tenant ids for which the credential
+        may acquire tokens. Add the wildcard value "*" to allow the credential to acquire tokens for
+        any tenant the application is installed.
```
mccoyp, line 57: For credentials like this, where not setting a default tenant is equivalent to saying "this credential is good for all tenants", I wonder if we should specify that in the docstring

xiangyan99: We don't want to add the words now because currently we don't allow user to specify tenant_id in the ctor (it is a beta feature which will ship in Oct.)

sdk/identity/azure-identity/azure/identity/_credentials/azure_cli.py
xiangyan99: Checked other libraries and I believe we should captalize it.

sdk/identity/azure-identity/azure/identity/_credentials/azure_powershell.py
mccoyp: In that case, does `additionally_allowed_tenants` do anything right now? If not, should it be included in the docstring?

mccoyp (approved):
I'm still unsure about the documentation of `additionally_allowed_tenants` in the CLI and PowerShell credentials since it doesn't seem like the kwarg actually does anything right now, but I'm not going to block the PR on it since the feature will at least apply in the near future.

xiangyan99:
> I'm still unsure about the documentation of `additionally_allowed_tenants` in the CLI and PowerShell credentials since it doesn't seem like the kwarg actually does anything right now, but I'm not going to block the PR on it since the feature will at least apply in the near future.

Yes. Before we add the support of user-specified tenant-id, `additionally_allowed_tenants` will not take effect.

sdk/identity/azure-identity/azure/identity/_credentials/azure_powershell.py
xiangyan99: Given we will GA tenant_id support in Oct, I am afraid we don't want to say additionally_allowed_tenants will not take effect in our docstring now. :)

PR #26134, opened 2022-09-09
[Cosmos] Add AutoScale Header Support 2/2

JericHunter:
Adding the ability to update the autoscale settings

### DIFF
--- a/sdk/cosmos/azure-cosmos/azure/cosmos/_base.py
+++ b/sdk/cosmos/azure-cosmos/azure/cosmos/_base.py
@@ -713,26 +713,21 @@ def _set_throughput_options(offer: Union[int, ThroughputProperties], request_opt
 
 def _deserialize_throughput(throughput: list) -> Any:
     throughput_properties = throughput
-    try:
-        max_throughput = throughput_properties[0]['content']['offerAutopilotSettings']['maxThroughput']
-    except (KeyError, TypeError):  # Adding TypeError just in case one of these dicts is None
-        max_throughput = None
-    try:
-        increment_percent = \
-            throughput_properties[0]['content']['offerAutopilotSettings']['autoUpgradePolicy']['throughputPolicy'][
-                'incrementPercent']
-    except (KeyError, TypeError):
-        increment_percent = None
-    try:
-        throughput = throughput_properties[0]["content"]["offerThroughput"]
-    except (KeyError, TypeError):
-        throughput = None
-    return ThroughputProperties(
-        auto_scale_max_throughput=max_throughput,
-        auto_scale_increment_percent=increment_percent,
-        offer_throughput=throughput,
-        properties=throughput_properties[0]
-    )
+    if 'offerAutopilotSettings' in throughput_properties[0]['content'] and 'autoUpgradePolicy' in \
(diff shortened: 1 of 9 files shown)


sdk/cosmos/azure-cosmos/azure/cosmos/_base.py
```diff
@@ -733,3 +733,19 @@ def _deserialize_throughput(throughput: list) -> Any:
         offer_throughput=throughput,
         properties=throughput_properties[0]
     )
+
+
+def _replace_throughput(throughput: Union[int, ThroughputProperties], new_throughput_properties: list):
+    max_throughput = throughput.auto_scale_max_throughput
+    increment_percent = throughput.auto_scale_increment_percent
+
+    if isinstance(throughput, object):
```
annatisch, line 742: Don't use this - you're already access the attributes on line 739.
This should be using the same kind of logic as your `set_throughput_options` above
Instead try something like:
```python
try:
    if throughput.auto_scale_max_throughput is not None:
        ...
except AttributeError:
    if isinstance(throughput, int)
        ...
    else:
        raise TypeError(...)
```

sdk/cosmos/azure-cosmos/azure/cosmos/_base.py
```diff
@@ -733,3 +733,19 @@ def _deserialize_throughput(throughput: list) -> Any:
...
+        if max_throughput is not None:
+            new_throughput_properties['content']['offerAutopilotSettings'][
+                'maxThroughput'] = max_throughput
+            if increment_percent:
+                new_throughput_properties['content']['offerAutopilotSettings']['autoUpgradePolicy']['throughputPolicy'][
+                    'incrementPercent'] = increment_percent
```
annatisch, line 748: Also need to handle:
```python
            if throughput.offer_throughput:
                new_throughput_properties["content"]["offerThroughput"] = throughput.offer_throughput
```

sdk/cosmos/azure-cosmos/test/test_auto_scale.py
```diff
@@ -59,7 +59,6 @@ def test_auto_scale(self):
         self.assertEqual(
             created_container_properties.auto_scale_max_throughput, 7000)
         self.assertEqual(created_container_properties.auto_scale_increment_percent, 0)
-        self.assertEqual(created_container_properties.offer_throughput, None)
```
annatisch, line 62: Why was this removed?

JericHunter: I messaged you about this previously this was removed because it will not work due to change of logic in deserialize throughput

annatisch: This seems kind of a bug... why does the service return both the autoscale values _and_ a default throughput?

annatisch: As I mentioned in chat - let's perhaps restore the deserialization logic so that we don't have to modify this test in order to unblock this PR/beta release. I think we need to open an issue to explore this more fully before GA.

annatisch (approved):

simorenoh (approved):

PR #26140, opened 2022-09-09
[AppConfig] Migrate tests to test proxy

YalinLi0312:
Resolves https://github.com/Azure/azure-sdk-for-python/issues/21146


xiangyan99:
How about splitting this PR into two.
1. Replace AsyncioRequestsTransport with AioHttpTransport and use existing tests to verify the change.
2. Migrate to use test proxy

YalinLi0312:
I don't think it's necessary.
1. Transport updates are part of migration. We met the transport issue while migrating.
2. There's no motivation to replace the transport type in the old test infrastructure.

xiangyan99:
I think the point is we want to make sure changing transport does not change the behavior at all.

How did you confirm that?

YalinLi0312:
Right, there are 5 async tests  failed with new transport, I have marked them `live_only` for now.
The rest tests are passing in live and playback, they could confirm the behavior.
@mccoyp What do you think?

mccoyp:
Hm. I just realized that I misspoke when I said that we could just make the 5 failing async tests live-only for the time being.

The issue is that async paging operations don't work in App Config when using `AioHttpTransport` -- that means that we shouldn't remove the `AsyncioRequestsTransport` default until that issue's fixed.

But, if we don't remove the `AsyncioRequestsTransport` dependency when using the test proxy, things won't work since the test proxy only patches the `send` method of `AioHttpTransport`.

So the options we have are:
1. Fix the async paging issue when using `AioHttpTransport` and remove the `AsyncioRequestsTransport` dependency.
  a. We should do this eventually, but after a _lot_ of debugging between Yalin and myself, it's not going to be easy. Infeasible in the short-term.
2. Keep the `AsyncioRequestsTransport` dependency and make all tests live-only.
  a. Not a good idea (CI would become useless).
3. Keep the `AsyncioRequestsTransport` dependency and support `AsyncioRequestsTransport` in the test proxy.
  a. Given the state of options **1.** and **2.**... this is really the only option.
  
All the above is to say: I don't see a harm in briefly supporting `AsyncioRequestsTransport`, especially given the circumstances. I'll merge https://github.com/Azure/azure-sdk-for-python/pull/25897, and in this PR we should revert the change to App Config's transport usage. We'll revisit that issue at a later time, after everything is migrated to the test proxy.

xiangyan99 (approved):

sdk/appconfiguration/ci.yml
```diff
@@ -28,6 +28,7 @@ pr:
 extends:
   template: ../../eng/pipelines/templates/stages/archetype-sdk-client.yml
   parameters:
+    TestProxy: true
```
mccoyp, line 31: This will get things working in playback test pipelines, but we'll also need to add `AZURE_TEST_RUN_LIVE` and `AZURE_SKIP_LIVE_RECORDING` per the [pipeline configuration instructions](https://github.com/Azure/azure-sdk-for-python/blob/main/doc/dev/test_proxy_migration_guide.md#enable-the-test-proxy-in-pipelines) to cover live tests

mccoyp (approved):

PR #26141, opened 2022-09-10
Verify challenge resource matches request domain

mccoyp:


sdk/keyvault/azure-keyvault-administration/azure/keyvault/administration/_access_control_client.py
```diff
@@ -23,8 +23,16 @@ class KeyVaultAccessControlClient(KeyVaultClientBase):
...
+        You should validate that this URL references a valid Key Vault or Managed HSM resource.
+        See https://aka.ms/azsdk/blog/vault-uri for details.
+    :param credential: An object which can provide an access token for the vault, such as a credential from
         :mod:`azure.identity`
+    :type credential: :class:`~azure.core.credentials.TokenCredential`
+
+    :keyword api_version: Version of the service API to use. Defaults to the most recent.
+    :paramtype api_version: ~azure.keyvault.administration.ApiVersion
+    :keyword bool verify_challenge_resource: Whether to verify the authentication challenge resource matches the Key
+        Vault or Managed HSM domain. Defaults to True.
```
annatisch, line 35: I thought in the initial proposal there was talk of supplying a list of resources here to verify against - was that determined to be unnecessary?

sdk/keyvault/azure-keyvault-administration/azure/keyvault/administration/_internal/async_challenge_auth_policy.py
```diff
@@ -69,6 +71,15 @@ async def on_challenge(self, request: "PipelineRequest", response: "PipelineResp
         except ValueError:
             return False
 
+        if self._verify_challenge_resource:
```
annatisch, line 74: Most of our pipeline configuration parameters can be supplied either via Client constructor, or per-operation kwargs. Looks like this one is only the former - is that intentional?
It might make sense in this context, I don't think there's a particularly strong case for enabling/disabling this on a per operation basis.

sdk/keyvault/azure-keyvault-administration/azure/keyvault/administration/_internal/async_challenge_auth_policy.py
```diff
@@ -69,6 +71,15 @@ async def on_challenge(self, request: "PipelineRequest", response: "PipelineResp
...
+            resource_domain = urlparse(scope).netloc
+            if not resource_domain:
+                raise ValueError(f"The challenge contains invalid scope '{scope}'.")
+
+            request_domain = urlparse(request.http_request.url).netloc
+            if not request_domain.lower().endswith(f".{resource_domain.lower()}"):
```
annatisch, line 80: We don't need to do any further stipping of "/" characters or anything?
I'm pretty sure `netloc` shouldn't include these.... what about things like port numbers? Will that ever interfere? I can't recall exactly what formatting `netloc` returns.

sdk/keyvault/azure-keyvault-certificates/azure/keyvault/certificates/_shared/async_challenge_auth_policy.py
```diff
@@ -69,6 +71,15 @@ async def on_challenge(self, request: "PipelineRequest", response: "PipelineResp
...
             return False
 
+        if self._verify_challenge_resource:
+            resource_domain = urlparse(scope).netloc
+            if not resource_domain:
+                raise ValueError(f"The challenge contains invalid scope '{scope}'.")
+
+            request_domain = urlparse(request.http_request.url).netloc
+            if not request_domain.lower().endswith(f".{resource_domain.lower()}"):
+                raise ValueError(f"The challenge resource '{resource_domain}' does not match the requested domain.")
```
annatisch, line 81: We want this to be a ValueError? As opposed to some kind of AuthenticationError?
I guess that's technically correct, though I'm wondering if there would be any benefit to users to be able to see the response that caused the failure like they could if this was some flavour of HttpResponseError.

sdk/keyvault/azure-keyvault-keys/CHANGELOG.md
```diff
@@ -1,5 +1,26 @@
...
+## 4.7.0 (2022-09-13)
+
+### Breaking Changes
+- Clients verify the challenge resource matches the vault domain. This should affect few customers,
+  who can provide `verify_challenge_resource=False` to client constructors to disable.
+
+### Other Changes
+- Changes from version 4.7.0b1 have been reverted and will be included in version 4.8.0b1
+
+## 4.7.0b1 (2022-08-12)
```
annatisch, line 12: Was this released from another branch or something?

mccoyp: 4.7.0b1 was released from `main` after 4.6.1 was released, and these changes are being applied on top of 4.6.1's release tag. So I just had to pull over the changelog entry from the beta release to make the timeline consistent -- after this minor version is released, I'll release 4.8.0b1 that merges the changes here with the updates that were made in 4.7.0b1

sdk/keyvault/azure-keyvault-administration/azure/keyvault/administration/_internal/async_challenge_auth_policy.py
mccoyp: I honestly don't know if per-operation verification was considered -- @heaths might know if it was. I would think that it's something we could add later if it's requested

sdk/keyvault/azure-keyvault-administration/azure/keyvault/administration/_access_control_client.py
mccoyp: @heaths might know more about the initial proposal -- I've only known about this binary option. I know that Identity is adding a list of supported tenants for cross-tenant auth, but that may be unrelated

sdk/keyvault/azure-keyvault-administration/azure/keyvault/administration/_internal/async_challenge_auth_policy.py
mccoyp: That's a good point. I'll see if I can add a test to specifically check for port numbers; our auth policy is currently port-agnostic when considering the endpoint of a request, but I need to make sure this particular check wouldn't be disrupted

sdk/keyvault/azure-keyvault-certificates/azure/keyvault/certificates/_shared/async_challenge_auth_policy.py
mccoyp: That's a good point, seeing the offending response would probably be helpful here. The response code at this point would be a 401, so raising a `ClientAuthenticationError` wouldn't be inconsistent. Do you think that would be an appropriate exception to raise in this case?

sdk/keyvault/azure-keyvault-administration/azure/keyvault/administration/_access_control_client.py
heaths: @annatisch, are you thinking about the Identity SDK change? The Key Vault change is only to make sure the resource (which gets used for the scope) matches the host domain. Separate features.

sdk/keyvault/azure-keyvault-administration/azure/keyvault/administration/_internal/async_challenge_auth_policy.py
heaths: It has to be per-client. Challenge-based auth policies should be caching per-scope bearer tokens and sending them per-call until they need to be refreshed.

sdk/keyvault/azure-keyvault-keys/CHANGELOG.md
heaths: Is the target hotfix/keyvault-verify branch branched from your last GA then?

sdk/keyvault/azure-keyvault-certificates/azure/keyvault/certificates/_shared/async_challenge_auth_policy.py
heaths: Not sure about your pipeline, but I would hope authentication headers are sanitized so I'm not sure this would help. Besides, the response itself wouldn't be an error, so I not sure any sort of response-based exception makes sense.

sdk/keyvault/azure-keyvault-administration/tests/recordings/test_access_control.pyTestAccessControltest_role_assignment[7.2].json
```diff
@@ -1,7 +1,7 @@
 {
   "Entries": [
     {
-      "RequestUri": "https://managedhsmvaultname.vault.azure.net/providers/Microsoft.Authorization/roleDefinitions?api-version=7.2",
+      "RequestUri": "https://managedhsmvaultname.managedhsm.azure.net/providers/Microsoft.Authorization/roleDefinitions?api-version=7.2",
```
Comment (heaths) [inline sdk/keyvault/azure-keyvault-administration/tests/recordings/test_access_control.pyTestAccessControltest_role_assignment[7.2].json:4]: Given how much size recordings contribute to the repos, are these necessary? No difference in recordings would be significant here.

sdk/keyvault/azure-keyvault-keys/tests/_async_test_case.py
```diff
@@ -59,7 +59,7 @@ def is_public_cloud():
 class AsyncKeysClientPreparer(AzureRecordedTestCase):
     def __init__(self, *args, **kwargs):
         vault_playback_url = "https://vaultname.vault.azure.net"
-        hsm_playback_url = "https://managedhsmvaultname.vault.azure.net"
+        hsm_playback_url = "https://managedhsmvaultname.managedhsm.azure.net"
```
heaths, line 62: Same comment here about recordings. Applies throughout the PR.

sdk/keyvault/azure-keyvault-keys/tests/test_challenge_auth_async.py
```diff
@@ -349,3 +351,75 @@ def verify(request):
...
+
+@pytest.mark.asyncio
+@empty_challenge_cache
+@pytest.mark.parametrize("verify_challenge_resource", [True, False])
+async def test_verify_challenge_resource_matches(verify_challenge_resource):
+    """The auth policy should raise if the challenge resource doesn't match the request URL unless check is disabled"""
+
+    url = get_random_url()
+    token = "**"
+    resource = "https://bad-resource.azure.net"
```
heaths, line 364: Would be a good idea to add a test - for a future regression test - to make sure your check against ".{domain}" e.g., ".vault.azure.net" continues working (note the preceeding ".").

heaths (approved):

sdk/keyvault/azure-keyvault-administration/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Release History
 
+## 4.2.0 (2022-09-13)
```
heaths, line 3: Ah, here and elsewhere, please update to 9/19 or 9/20 (we plan to slip to 9/20 for the announcement).

sdk/keyvault/azure-keyvault-administration/tests/recordings/test_access_control.pyTestAccessControltest_role_assignment[7.2].json
Comment (mccoyp) [reply sdk/keyvault/azure-keyvault-administration/tests/recordings/test_access_control.pyTestAccessControltest_role_assignment[7.2].json:4]: This is unfortunately necessary because of the validation check. I suppose we could alternatively initialize our client with `verify_challenge_resource=False`, but otherwise our MHSM tests fail in playback because our playback URL was using the wrong URL suffix. Great for validating that the check works, but less than ideal for backwards compatibility

Comment (heaths) [reply sdk/keyvault/azure-keyvault-administration/tests/recordings/test_access_control.pyTestAccessControltest_role_assignment[7.2].json:4]: I wouldn't stress about it, then. I just thought maybe it was happenstance.

sdk/keyvault/azure-keyvault-keys/CHANGELOG.md
mccoyp: > Is the target hotfix/keyvault-verify branch branched from your last GA then?

Yep!

sdk/keyvault/azure-keyvault-administration/azure/keyvault/administration/_access_control_client.py
annatisch: Thanks @heaths - you're totally right, I was confusing this with the Identity change. :)

sdk/keyvault/azure-keyvault-certificates/azure/keyvault/certificates/_shared/async_challenge_auth_policy.py
annatisch: We don't sanitize anything from the response that's included in the HttpResponseError - but it's true that the response in that case is not actually an error status, so probably doesn't make much sense...

sdk/keyvault/azure-keyvault-keys/tests/test_challenge_auth_async.py
mccoyp: I just checked .NET's tests to understand what you mean by this, but I don't see any tests using a resource including a preceding ".", like "https://.vault.azure.net".

Do you mean that such a resource should be considered valid, or that it should be invalid even if "https://vault.azure.net" (without the preceding ".") would be correct? Or do you mean that we should still correctly identify a challenge resource as invalid if it's preceded by a "."?

mccoyp: Oh, I think I understand what you mean. Are you saying that we should have a check that explicitly ensures a KV endpoint (`https://vault-name.vault.azure.net`) succeeds when the resource is `https://vault.azure.net`?

If so, I'm happy to add another test if you think it would be helpful. Otherwise, my thinking was that this regression is implicitly checked by the other tests in this file using URLs from `get_random_url` -- which all have a domain ending with `.vault.azure.net` -- and requiring a matching challenge resource. `get_random_url` was updated to make this work, since it didn't actually have the correct domain suffix before

sdk/keyvault/azure-keyvault-certificates/azure/keyvault/certificates/_shared/async_challenge_auth_policy.py
mccoyp: Hm. I think I'll stick with `ValueError` for this release if there are no complaints, and we can revisit this when the issue is less pressing. I'm about to update the error message to include remediation steps (e.g. setting `verify_challenge_resource`), which more clearly indicate that a `ValueError` is appropriate, if not perfect

sdk/keyvault/azure-keyvault-keys/tests/test_challenge_auth_async.py
heaths: No, I'm saying a URI like "https://myvault.azure.net" should not succeed. This is why that preceding "." e.g., ".vault.azure.net", is important.

heaths (approved):

sdk/keyvault/azure-keyvault-keys/tests/test_challenge_auth_async.py
mccoyp: Ahhh I see, that makes total sense. I'll get that added 🙂

PR #26167, opened 2022-09-13
[RemoteRendering] Migrate tests to test proxy

pvaneck:
Tests for "azure-mixedreality-remoterendering" were refactored to use Test proxy and pytest fixtures.

- Recordings were generated using test proxy.
- Polling interval for live tests was increased to decrease the recording file sizes for some tests.


sdk/remoterendering/azure-mixedreality-remoterendering/tests/conftest.py
```diff
@@ -0,0 +1,105 @@
...
+        "id_placeholder": TEST_ID_PLACEHOLDER
+    }
+
+
+@pytest.fixture
+def arr_client(account_info):
+    # Give a small interval for playback tests to avoid
+    # a race condition where a poller consumes more recorded requests
+    # than expected.
+    polling_interval = 10 if is_live() else 0.2
```
mccoyp, line 82: Do you know why we need a non-zero interval for the synchronous client, but not for the async client?

mccoyp (approved):

sdk/remoterendering/azure-mixedreality-remoterendering/tests/conftest.py
pvaneck: I am not quite certain, but I've never had any flakiness with the async tests with a `0` polling interval. This [PR](https://github.com/Azure/azure-sdk-for-python/pull/25820) goes into some detail about how the race condition is hit. Might have something to do with the async/await event loop logic which somehow ensures `get_asset_conversion` is called before the poller uses all the recorded requests.

PR #26181, opened 2022-09-13
diagnostic pr update

JericHunter:
making changes to update our cosmos diagnostics release


sdk/cosmos/azure-cosmos/azure/cosmos/_cosmos_http_logging_policy.py
```diff
@@ -40,7 +40,7 @@ def __init__(self, logger=None, **kwargs): # pylint: disable=unused-argument
         self._enable_diagnostics_logging = kwargs.pop("enable_diagnostics_logging", None)
         super().__init__(logger, **kwargs)
         if self._enable_diagnostics_logging:
-            self.logger = logger or logging.getLogger(__name__)
+            self.logger = logger
```
annatisch, line 43: I _think_ these two lines (42-43) can be removed altogether and we just let `self.logger` be assigned in the Core policy.
It's not perfect, because the default logger will be in the azure.core namespace. But it will be okay for now and we can review that later.

annatisch (approved):

simorenoh (approved):

PR #26254, opened 2022-09-15
Fix python live test template variables

benbp:


sdk/template/tests.yml
```diff
@@ -1,16 +1,15 @@
...
       ServiceDirectory: template
       MatrixReplace:
         - TestSamples=.*/true
       EnvVars:
-        TEMPLATE_CONFIG_CONNECTION: $(python-template-connection-string) # Add variables/mappings as appropriate for your service.
-        AZURE_CLIENT_ID: $(aad-azure-sdk-test-client-id)
-        AZURE_CLIENT_SECRET: $(aad-azure-sdk-test-client-secret)
-        AZURE_SUBSCRIPTION_ID: $(azure-subscription-id)
-        AZURE_TENANT_ID: $(aad-azure-sdk-test-tenant-id)
+        AZURE_CLIENT_ID: $(TEMPLATE_CLIENT_ID)
```
benbp, line 10: I also removed the `TEMPLATE_CONFIG_CONNECTION` because we don't want people setting up keyvault variable references directly, but rather using the bicep templates for stuff (otherwise local testing setup is harder).

scbedd (approved):

PR #26273, opened 2022-09-16
[Storage] Add lease action support to datalake `append_data` and `flush_data`

jalauzon-msft:
This change adds support for the STG85/86 feature of performing lease actions when appending or flushing data to a datalake file. Both append and flush now take a new set of keyword args that allow for specific lease actions to be taken in the same operation as the append or flush.

In both APIs, the existing `lease` argument was re-used as the lease id for creating a new lease with the "acquire" or "acquire-release" actions.


vincenttran-msft (approved):
LGTM! Thanks for getting to this 👍

PR #26276, opened 2022-09-16
[Storage] Support `walk_blobs` with prefix, delimiter, and versions specified

jalauzon-msft:
An STG85/86 feature is to support specifying `prefix`, `delimiter`, and `include=versions` with `list_blobs` (`walk_blobs` in this case since `delimiter` is used) at the same time.

This was already supported, just needed to add a test for it. Additionally, fixing up the documentation for the `include` keyword on `walk_blobs`.


vincenttran-msft (approved):
LGTM! Thanks 😄

PR #26278, opened 2022-09-16
Swap ML to only linux tests

scbedd:
@annatisch @lmazuel I'd prefer sign off from at least one of you before I merge this.

@needuv here is the matrix as requested.

### DIFF
--- a/sdk/ml/platform-matrix.json
+++ b/sdk/ml/platform-matrix.json
@@ -1,0 +1,40 @@
+{
+   "displayNames": {
+     "--disablecov": "",
+     "false": "",
+     "true": ""
+   },
+   "matrix": {
+     "Agent": {
+       "ubuntu-20.04": { "OSVmImage": "MMSUbuntu20.04", "Pool": "azsdk-pool-mms-ubuntu-2004-general" },
+     },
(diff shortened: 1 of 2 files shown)


sdk/ml/platform-matrix.json
```diff
@@ -0,0 +1,40 @@
...
+           "TestSamples": "false"
+         }
+       }
+     },
+     {
+       "Config": {
+         "Ubuntu2004_310": {
+           "OSVmImage": "MMSUbuntu20.04",
+           "Pool": "azsdk-pool-mms-ubuntu-2004-general",
+           "PythonVersion": "3.10.4",
```
benbp, line 32: I think this could just be added to the base matrix `PythonVersion` array, unless you planned on adding other entries here?

scbedd: It's specific for the coverage entry. that's all

benbp: Not sure I understand? This entry has the same `CoverageArg` as the base matrix.

scbedd:
Closing this as @needuv and I ran down the hang on windows.

Issue #26109, opened 2022-09-08
[CLU] update polling_interval


kristapratico:
This currently defaults to 30s, per the azure-core default. It should be overridden in the library to 5s per the service team.

PR #26297, opened 2022-09-19
[CLU+QA] update polling_interval per service recommendation

kristapratico:
Resolves https://github.com/Azure/azure-sdk-for-python/issues/26108 https://github.com/Azure/azure-sdk-for-python/issues/26109

Request from the service team to update the default polling_interval to 5 seconds for both QA and CLU. This was previously defaulting to the azure-core default of 30 seconds.


heaths:
.NET defaults to 1s unless the `Retry-After` header is sent. @peytonfraser @kayousef why isn't the service sending this header if that's what it recommends?

kristapratico:
> .NET defaults to 1s unless the `Retry-After` header is sent. @peytonfraser @kayousef why isn't the service sending this header if that's what it recommends?

Just to give background, in Text Analytics we've always used 5 secs from what the service team recommended a long time ago. This is updating CLU and QA to use the same per my conversation with Mikael.

The 5 seconds is used as the initial polling interval between the POST and the first GET (before we have a chance to see a Retry-After header). After that, if the service returns a Retry-After header, then that's what is used in subsequent calls (like .NET).

heaths (approved):

PR #26303, opened 2022-09-19
[SchemaRegistry] change avro mindep

swathipil:
- changing avro min dependency version from 1.11.1 back to 1.11.0, b/c bug fix in avro didn't require any changes in the SDK.


sdk/schemaregistry/azure-schemaregistry-avroencoder/CHANGELOG.md
```diff
@@ -6,8 +6,6 @@ This version and all future versions will require Python 3.7+, Python 3.6 is no
 
 ### Bugs Fixed
 
-- Updated `avro` minimum dependency to 1.11.1, which fixed a bug that previously restricted complex types from being used as names.
```
kashifkhan, line 9: do we have a link to the bug that was reported ?

kashifkhan (approved):

sdk/schemaregistry/azure-schemaregistry-avroencoder/CHANGELOG.md
swathipil: So this was actually a [bug in avro](https://issues.apache.org/jira/browse/AVRO-3370), and it resulted in one of our tests failing. But, wasn't a bug that any of our customers reported. Originally, I thought we needed to bump the min dependency to account for this bug. Laurent mentioned that we don't need to since it wasn't a bug in our code + we don't want to force users to update their avro version.

PR #26344, opened 2022-09-21
[Storage] Migrate remaining blob tests to test proxy

jalauzon-msft:
This change migrates `test_blob_retry`, `test_blob_retry_async`, `test_largest_block_blob`, and `test_largest_block_blob_async` to test proxy. Also migrate `test_upload_chunking` to just a standard `unittest.TestCase` as neither of the tests make network calls and are standard unit tests.


vincenttran-msft (approved):
LGTM! Blob ✔️

Issue #26374, opened 2022-09-21
Type conflict of "credential" in AppConfig


YalinLi0312:
The "credential" parameter in `AzureAppConfigurationClient` is in type `Union[AppConfigConnectionStringCredential, "AsyncTokenCredential"]` , this value will pass to initialize a `AzureAppConfiguration` object(a generated class) which only accept `AsyncTokenCredential` as credential type.

The swagger settings file has been removed, this is the swagger README I found from history:  https://github.com/seankane-msft/azure-sdk-for-python/tree/08a03304fe0f2e9c0e2de178c5fef86b2bb12880/sdk/appconfiguration/azure-appconfiguration/swagger

PR #26375, opened 2022-09-21
[AppConfig] Regenerate code without credentials

YalinLi0312:
Resolves https://github.com/Azure/azure-sdk-for-python/issues/26374
The package was generated with "add-credential: true" flag, which introduces a type conflict of parameter "credential".

To fix it, we regenerate without the flag, and add the swagger setting file back for better tracking the autorest settings while generating.


xiangyan99 (commented):
core >= 1.20.0 to have rest support.

iscai-msft (approved):
looks good besides core mindep, should be at least 1.24.0 https://github.com/Azure/autorest.python/blob/autorestv3/packages/autorest.python/ChangeLog.md#2022-09-21---617

PR #26379, opened 2022-09-21
[STG86] File Share Support for Invalid XML Characters

vincenttran-msft:
This PR adds in support for invalid XML characters `\uFFFF` and `\uFFFE`.


sdk/storage/azure-storage-file-share/azure/storage/fileshare/_models.py
```diff
@@ -5,7 +5,7 @@
 # --------------------------------------------------------------------------
 # pylint: disable=too-few-public-methods, too-many-instance-attributes
 # pylint: disable=super-init-not-called, too-many-lines
-
+import urllib.parse
```
jalauzon-msft, line 8: nit: Can you change this to `from urllib.parse import unquote` and change the usages below to just `unquote(`? Just looks cleaner that way.

sdk/storage/azure-storage-file-share/tests/test_directory.py
```diff
@@ -427,6 +427,42 @@ def test_list_subdirectories_and_files(self, storage_account_name, storage_accou
         self.assertEqual(list_dir[5]['name'], 'file3')
         self.assertEqual(list_dir[5]['is_directory'], False)
 
+    @FileSharePreparer()
+    def test_list_subdirectories_and_files_encoded(self, storage_account_name, storage_account_key):
```
jalauzon-msft, line 431: I think you just migrated this file to test proxy? If so, you'll need to merge main in here and update these tests :).

Also, okay to do in another PR after this merges.

sdk/storage/azure-storage-file-share/tests/test_directory.py
```diff
@@ -427,6 +427,42 @@ def test_list_subdirectories_and_files(self, storage_account_name, storage_accou
...
+        self._setup(storage_account_name, storage_account_key)
+        share_client = self.fsc.get_share_client(self.share_name)
+        directory = share_client.create_directory('directory\uFFFE')
```
jalauzon-msft, line 434: So just to confirm, like .NET, we don't expose this top-level directory anywhere? The one that is the top-level `DirectoryPath` in the XML response?

sdk/storage/azure-storage-file-share/tests/test_directory_async.py
```diff
@@ -494,6 +494,52 @@ async def test_list_subdirectories_and_files_async(self, storage_account_name, s
         self.assertEqual(list_dir[5]['is_directory'], False)
         self.assertEqual(list_dir[5]['size'], 5)
 
+    @FileSharePreparer()
+    @AsyncStorageTestCase.await_prepared_test
```
jalauzon-msft, line 498: nit: You don't really need this line even for non test-proxy files but since this will migrate to test proxy soon, fine to leave it :)

sdk/storage/azure-storage-file-share/tests/test_directory.py
vincenttran-msft: If I am understanding your question correctly, yes it would seem that we don't expose this top-level directory (in the return from `get_directory_properties()` API call and in the return from `list_directories_and_files()` API call)

sdk/storage/azure-storage-file-share/tests/test_directory.py
vincenttran-msft: Will do! I somehow have managed to merge main into the feature branch and then the feature branch into this branch without alerting all of Microsoft 😄 so will do it all in this PR!

jalauzon-msft (approved):

sdk/storage/azure-storage-file-share/tests/test_directory.py
jalauzon-msft: Cool, just checking in case we missed it.

PR #26395, opened 2022-09-22
[Storage] Fix encryption tests after test proxy migration

jalauzon-msft:
When migrating the blob encryption tests to test proxy, it was missed that the recordings contained the package version inside the encryption metadata header. This package version caused mismatched recording errors when these tests run in the nightly pipeline with the alpha package version. This would also become a problem once we bumped the blob package version.

This change adds an additional regex sanitizer to mask the package version used into just Python x.x.x. All encryption tests were then re-recorded.


vincenttran-msft (approved):
One sanitizer later.. 😷 
Thanks for getting to this pesky issue!

PR #26407, opened 2022-09-22
Ensure bandit always installs its requirements

scbedd:
Given that we can enable disable any of these pieces, we really need each step to install the tools it requires.


luigiw (approved):

luigiw:
FYI, the same issue happens with "Run Breaking Changes" step, please consider applying the same thing to "Run Breaking Changes" step.

PR #26430, opened 2022-09-24
CI Logging Updates

scbedd:
@mccoyp I am working through the `optional` requirements testing right. As I did that, I realized that I wanted to ensure that any new code I added did a better job of avoiding spam.

There is a ton of usage refactoring that is still necessary, but this PR does the following.

- [x] Provides an easy access point that we can add to all of our tox scripts. `logger = initialize_logger('<script name>')` at the top of each script file. 
- [x] The provided logger will log in one or two places. 
  - A default console logger. By default, we limit to ERROR. For further information, set `LOGLEVEL` environment variable to `INFO` or `DEBUG`.
  - A file logger that writes to `<repo root>/.logs` as necessary.

We still need to change the default log level to ERROR for CI, optionally overridable via queue time variable. Currently, this PR currently redirects the wheel and sdist assembly output to files.


mccoyp (approved):
LGTM!

scbedd:
Removing `build` output reduces test log output by an overall 26%.

PR #26055, opened 2022-09-06
[SchemaRegistry] add support for getting schema by version

swathipil:
Service already supports getting schema by version and this operation has been added to the swagger. This PR regenerates from that swagger and adds support for this feature. Specifically:
- Adding the `get_schema_by_version` methods on the sync and async clients.
- Adding `version` ivar to the `SchemaProperties` model.

TODO:
- [x] add tests
- [x] update README, samples README, and changelog

addressing: #25072
fixes: #26180

Issue #25072, opened 2022-07-05
Provide a function for fetching schema from Schema Registry using group name, schema name and schema version.


harshdeep78singh:
Currently, I am fetching the schema from the Schema Registry using the following endpoint

/$schemaGroups/{groupName}/schemas/{schemaName}/versions/{version}

There is no function provided in the **azure.schemaregistry.SchemaRegistryClient** class for pulling the schema from Event Hub Schema Registry using group name, schema name, and version.

Can we have a function for the same?

Also, let me know if you are open to a PR?

mccoyp:
Hi @harshdeep78singh, thank you for opening an issue! I'll tag the appropriate folks so we can answer your question about the potential availability of this API.

As for opening a pull request, we're always happy to accept PRs 🙂 For something like adding a client method there will need to be some design discussion, but opening a PR could kickstart that conversation! I'd recommend waiting just a bit before doing any work so that someone who works with the service can provide some input in this case.

swathipil:
Hi @harshdeep78singh - Thanks for bringing this to our attention! I have discussed this with the team and we will be adding this functionality asap. I saw that you created a PR and took a look.

Unfortunately, we will not be able to merge this, because the changes you made were to our "generated" code. This code is automatically generated from our internal "swagger" file that describes operations that we allow. So, ultimately, the changes that you made will actually be added upon re-generation after we update our swagger file to match. Really appreciate your efforts here, however.

We are currently making changes to the swagger file and will be having API design discussions after that. I'll keep you updated as we continue to make progress. Thanks again!

swathipil:
Hi @harshdeep78singh - Appreciate your patience here! I just merged the feature. (sample [[here](https://github.com/Azure/azure-sdk-for-python/blob/376224e541576c351e6b5cec27fcbc030ca5685f/sdk/schemaregistry/azure-schemaregistry/samples/sync_samples/schema_registry.py#L100)]), and it'll be out in our next release. I'll update this issue as soon as it's out.

In the meantime, if you're interested, you can manually build and install the wheel:
1) `azure-sdk-for-python/sdk/schemaregistry/azure-schemaregistry> python setup.py bdist_wheel -d [dest_folder]`
2) `[dest_folder]> pip install azure_schemaregistry-1.2.0-py3-none-any.whl`

Thanks!

harshdeep78singh:
Thank you for the update

harshdeep78singh:
@swathipil thank you for the update

[...]

PR #26478, opened 2022-09-27
[SchemaRegistry] overload get schema to take version

swathipil:
OOB discussion with Johan:
* Remove `get_schema_by_version` method from sync and async SchemaRegistryClient
* Add `get_schema` overload, which takes `group_name`, `name`, and `version` as keyword args.

Addressing: #25072
Follow up from: #26055


sdk/schemaregistry/azure-schemaregistry/azure/schemaregistry/aio/_schema_registry_client_async.py
```diff
@@ -136,19 +138,36 @@ async def register_schema(
...
+    @overload
+    async def get_schema(
+        self, *, group_name: str, name: str, version: int, **kwargs
+    ) -> Schema:
+        ...
+
+    @distributed_trace_async
+    async def get_schema(self, *args: Union[str, int], **kwargs: Any) -> Schema:
+        """Gets a registered schema. There are two ways to call this method:
+        1) To get a registered schema by its unique ID, pass the `schema_id` parameter and any optional
```
JoshLove-msft, line 161: What optional keyword args are there when passing in schema ID?

sdk/schemaregistry/azure-schemaregistry/azure/schemaregistry/aio/_schema_registry_client_async.py
```diff
@@ -136,19 +138,36 @@ async def register_schema(
...
+        ...
+
+    @overload
+    async def get_schema(
+        self, *, group_name: str, name: str, version: int, **kwargs
+    ) -> Schema:
+        ...
+
+    @distributed_trace_async
+    async def get_schema(self, *args: Union[str, int], **kwargs: Any) -> Schema:
```
JoshLove-msft, line 159: why are the args a Union of str and int? Can you pass schema Id or version as args?

JoshLove-msft (approved):
Just some questions that are probably a result of me not understanding Python 😆 
But based on the samples, it all looks good.

sdk/schemaregistry/azure-schemaregistry/azure/schemaregistry/aio/_schema_registry_client_async.py
swathipil: so optional keywords are ones that are reserved by azure-core that users can [pass in for the request itself](https://github.com/Azure/azure-sdk-for-python/blob/b37efe5deb6aaa096f039a565e7222201257f0f8/sdk/core/azure-core/azure/core/rest/_rest_py3.py#L68). for ex, params/json/content/data.

Yep, both overloads would appear in the docs!

sdk/schemaregistry/azure-schemaregistry/azure/schemaregistry/aio/_schema_registry_client_async.py
swathipil: ah good catch! I think args should just be str - will fix this. I added int here b/c version is an int. However, version will be passed in as a kwarg(`version=`) and not a positional argument, so we don't need the type of version to be included in args.

So schema_id can be passed in as a positional argument, (i.e. we don't require the `schema_id=` to be passed in) or as a keyword arg (i.e. schema_id=schema_id). If it's passed in as a positional arg, it'll be in `args`. If it's passed in as a kwarg, then the type will be included in `Any`.

sdk/schemaregistry/azure-schemaregistry/azure/schemaregistry/_schema_registry_client.py
```diff
@@ -162,50 +181,51 @@ def get_schema(self, schema_id, **kwargs):
...
         """
         http_request_kwargs = get_http_request_kwargs(kwargs)
-        request = schema_rest.build_get_schema_version_request(
-            group_name=group_name, schema_name=name, schema_version=version, **http_request_kwargs
-        )
+        try:
+            try:
+                schema_id = args[0]
+            except IndexError:
+                schema_id = kwargs.pop("schema_id")
```
deyaaeldeen, line 196: nitpick, is there a nicer way to do this access? like checking args' length first?

deyaaeldeen: also could readability be improved by factoring the schema_id logic into a helper function?

deyaaeldeen (approved):
🚀 happy to see all languages aligned

sdk/schemaregistry/azure-schemaregistry/azure/schemaregistry/_schema_registry_client.py
swathipil: In Python, we usually use the try/except pattern over the if/else. I can add comments for each block for more clarity?

sdk/schemaregistry/azure-schemaregistry/azure/schemaregistry/_schema_registry_client.py
```diff
@@ -162,50 +181,54 @@ def get_schema(self, schema_id, **kwargs):
...
+            )
+        except KeyError:
+            # If group_name, name, and version aren't passed in as kwargs, raise error.
+            try:
+                group_name = kwargs.pop("group_name")
+                name = kwargs.pop("name")
+                version = kwargs.pop("version")
+            except KeyError as exc:
+                raise TypeError(
+                    f"""If getting schema by version, '{exc.args[0]}' is a required keyword."""
```
johanste, line 211: We should make the error message look more like a "vanilla" TypeError you get if calling with mismatched parameters. And we should assume that it was the original method (by id) that was intended unless we believe that the "by version" method is more commonly used.

sdk/schemaregistry/azure-schemaregistry/azure/schemaregistry/_generated/aio/operations/_operations.py
```diff
@@ -415,6 +431,7 @@ async def register(  # pylint: disable=inconsistent-return-statements
             401: ClientAuthenticationError,
             404: ResourceNotFoundError,
             409: ResourceExistsError,
+            304: ResourceNotModifiedError,
```
johanste, line 434: Is this a GET call? If not, why are we expecting a 304 answer? Is this just "let's apply the same map everywhere" thing?

swathipil: No, this is a [PUT call](https://github.com/Azure/azure-rest-api-specs/blob/4b2f5a45a34e351e38e8b075cf1da5a775fd414a/specification/schemaregistry/data-plane/Microsoft.EventHub/stable/2021-10/schemaregistry.json#L385). According to Izzy and [[this issue](https://github.com/Azure/autorest.python/issues/1415)], this is added by default everywhere and doesn't take input from the swagger.

PR #26493, opened 2022-09-27
[Test Proxy] Always return dict from recorded_test fixture

mccoyp:
# Description

The `recorded_test` fixture was recently optimized to yield immediately when live tests are run with recording skipped, since we don't need to load recordings in that scenario. This did introduce a bug though, since the `variable_recorder` fixture always attempts to fetch variables from `recorded_test`'s return value.

One option would be to check if a `variables` key is present in `recorded_test` inside of `variable_recorder`. But, I think that things are simpler and more consistent if we enforce always returning a certain set of information about the test via `recorded_test` -- for now, that just includes `variables`.


pvaneck (approved):
Sorry for not validating the previous fix with `variable_recorder`! I think this is probably the cleanest approach to handle the regression. Thanks!

PR #26519, opened 2022-09-28
[Storage] Migrate `test_directory.py` to test proxy

vincenttran-msft:
This PR is the first test proxy migration effort for the `file-share` package. This PR covers the following:

- Migrate `test_directory.py`
- Migrate `test_directory_async.py`
- Setup `conftest.py`


sdk/storage/azure-storage-file-share/tests/conftest.py
```diff
@@ -1,33 +1,37 @@
...
+    add_header_regex_sanitizer,
+    add_oauth_response_sanitizer,
+    add_uri_string_sanitizer,
+    test_proxy
+)
 
-# Ignore async tests for Python < 3.5
+# Ignore async tests for PyPy
 collect_ignore_glob = []
-if sys.version_info < (3, 5):
```
vincenttran-msft, line 32: When setting up `conftest.py` I just "borrowed" 😛 the `conftest.py` from datalake. Should I pop back in this explicit version check? Tests seemed to run fine in _this_ file without it, but want to point it out 😄

jalauzon-msft (approved):
Looks good! Thanks for jumping into fils-share. Just one comment on `conftest`.

PR #26522, opened 2022-09-28
[QA] release cleanup

kristapratico:


shared_requirements.txt
```diff
@@ -250,6 +250,7 @@ opentelemetry-sdk<2.0.0,>=1.5.0,!=1.10a0
 #override azure-ai-language-conversations azure-core<2.0.0,>=1.24.0
 #override azure-mgmt-eventgrid msrest>=0.7.1
 #override azure-ai-language-conversations isodate<1.0.0,>=0.6.1
+#override azure-ai-language-questionanswering isodate<1.0.0,>=0.6.1
```
heaths, line 253: I don't fully understand what this does, but just to make sure: should this be "1.1.0" that you're releasing, or is indeed meant to be the previous GA?

heaths (approved):

shared_requirements.txt
kristapratico: This is describing the version of `isodate` that the QA package takes as dependency. We describe the shared requirements and dependencies across the repo in this file. `isodate` is currently set as [0.6.0](https://github.com/Azure/azure-sdk-for-python/blob/3c3e40228ea48800bc0341e1f967d1f8b620abcb/shared_requirements.txt#L128) for the repo, but since we used the latest generator we require 0.6.1 and so we put this override line in this file so the CI doesn't scream.

heaths: Thanks. Now that you described it I can see how to parse that. :smile:

PR #26549, opened 2022-09-29
[AppConfig] Drop msrest

YalinLi0312:
AppConfig package was regenerated with latest version of `autorest`(https://github.com/Azure/azure-sdk-for-python/pull/26375), which doesn't use `msrest`. So we can drop this requirement.


mccoyp (commented):
LGTM! I would just wait for Xiang's approval since I'm not a package owner 🙂

xiangyan99 (approved):

PR #26607, opened 2022-10-03
Regression Runs Proxy prior to invoking tests

scbedd:
We dealt with a [breaking change](https://github.com/Azure/azure-sdk-for-python/pull/26494) for running the test-proxy in CI. The problem is, the regression tests are now hitting this as well because they use the engsys at the time of writing. That would mean that the `sdk-tools` startup of the test-proxy will hit issues from these older branches.

An alternative is to go in and apply the `proxy_startup.py` patch to each old tag/branch that doesn't have this adjustment. I'd rather save myself that effort. This PR is an attempt to avoid a couple dozen manual patches.

### DIFF
--- a/eng/pipelines/templates/jobs/regression.yml
+++ b/eng/pipelines/templates/jobs/regression.yml
@@ -86,7 +86,7 @@ jobs:
 
       - template: /eng/common/testproxy/test-proxy-tool.yml
         parameters:
-          runProxy: false
+          runProxy: true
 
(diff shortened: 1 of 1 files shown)


swathipil (approved):
thanks Scott!

benbp (approved):

scbedd:
There are a couple failing regression tests, but at least we're to the real failures, versus failures of the system. Going to override and merge this.

PR #26611, opened 2022-10-03
[Storage] Fix type hints for classmethods in Blob and Datalake

jalauzon-msft:
The type hint for the return type of various `@classmethod` methods in blob and datalake clients (such as `from_connection_string`) were set to a `TypeVar` that was not configured to correctly point at the class type. This meant that Intellisense did not know the return type of these methods, making development with them a bit harder.

This change adjusts the type hints to use `typing_extensions.Self` so Intellisense can properly pick up the type. Along the way, this change also moves these methods and client constructors to the new type hint style and includes some missing imports for credential types.


annatisch (approved):

alherm7:
Thanks @jalauzon-msft; I'm impressed at how quick you guys were to respond to this request. :)

How long does it take before the changes will be available on pypi?

jalauzon-msft:
@alherm7, it may take a little while for this change to be released due to the timing of the fix. Our public API surface was just locked for our next upcoming release this month so that means this will have to wait until after that. Tentatively, our next release would be in November but that will likely be a beta release.

vincenttran-msft (approved):
LGTM! Thanks for getting to this 😄 should make for a much better user experience!

PR #26623, opened 2022-10-04
fix eventgrid pipeline, failing dep from mgmt

kristapratico:


kristapratico:
@rakshith91 these version updates are due to the test dependency on azure-mgmt-eventgrid. Let me know if there is a different way you want to go about this

kristapratico:
Looks like this change got merged in another PR, closing

PR #26626, opened 2022-10-04
Fix py.typed verification script

pvaneck:
In order to ensure that a py.typed file is packaged with both source and binary distributions of a package, we have a script that verifies various configurations. This script had a bug that would validate a package as passing as long as the MANIFEST.in file was correctly configured, disregarding issues in the setup.py file.

This fixes the issue and updates pertinent impacted setup.py files.

Note that the following packages also have issues with their configurations, and their pipelines have been failing for a while due to the verify_sdist check failing. Presumably we'd want to disable the CI for these entirely since they are no longer maintained.

- applicationinsights/azure-applicationinsights
- loganalytics/azure-loganalytics

### DIFF
--- a/eng/tox/verify_sdist.py
+++ b/eng/tox/verify_sdist.py
@@ -97,10 +97,9 @@ def verify_sdist_pytyped(
     if os.path.exists(manifest_location):
         with open(manifest_location, "r") as f:
             lines = f.readlines()
-            result = any([include for include in lines if "py.typed" in include])
-
-            if not result:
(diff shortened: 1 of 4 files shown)


scbedd (approved):
Hey @pvaneck , thanks for the contribution.

~~Honestly the update to `verify_sdist` is just a style change, but I have no reason to say no. The updates to these other packages are absolutely necessary and I'm grateful you updated em.~~

I see what you're calling out here now.

Thanks!

pvaneck:
Thanks for the review, @scbedd!

PR #26649, opened 2022-10-05
[formrecognizer] Skip errors in live tests

catalinaperalta:
Skip more error types that affect live tests that are not related to the SDK.

### DIFF
--- a/sdk/formrecognizer/azure-ai-formrecognizer/tests/conftest.py
+++ b/sdk/formrecognizer/azure-ai-formrecognizer/tests/conftest.py
@@ -79,7 +79,7 @@ def skip_flaky_test(f):
             elif "Timeout" in error.message.casefold():
                 pytest.mark.skip("flaky service response: {}".format(error))
                 logger.debug("flaky service response: {}".format(error))
-            elif "InvalidRequest" in error.message.casefold():
+            elif "InvalidImage" in error.message.casefold():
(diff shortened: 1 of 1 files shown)


catalinaperalta:
I want to run one more pipeline run on my latest changes, but I wanted to get a review in for this.

sdk/formrecognizer/azure-ai-formrecognizer/tests/conftest.py
```diff
@@ -69,4 +69,11 @@ def wrapper(*args, **kwargs):
         except HttpResponseError as error:
             if "Invalid request".casefold() in error.message.casefold():
                 pytest.mark.skip("flaky service response")
+            elif "Generic error".casefold() in error.message.casefold():
```
kristapratico, line 72: Do you think we should include the flaky response in the skip message or log the response? Just so we don't miss somewhere where we skip a test that is failing due to something on our end.

kristapratico (approved):

sdk/formrecognizer/azure-ai-formrecognizer/tests/conftest.py
catalinaperalta: Done, added the message to the skip and logged since I'm still experimenting with what will be easier to see and analyze in devops

PR #26658, opened 2022-10-05
[Storage] Migrate `test_file.py` to test proxy (file-share)

vincenttran-msft:
This PR migrates `test_file.py` and `test_file_async.py` to use test proxy 😄


sdk/storage/azure-storage-file-share/tests/test_file.py
```diff
@@ -1,38 +1,38 @@
-# coding: utf-8
-
 # -------------------------------------------------------------------------
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License. See License.txt in the project root for
 # license information.
 # --------------------------------------------------------------------------
 import base64
 import os
+import requests
```
jalauzon-msft, line 8: nit: `requests` is actually an external lib and therefore should down above `pytest`.

sdk/storage/azure-storage-file-share/tests/test_file.py
```diff
@@ -859,12 +1012,13 @@ def test_update_range_from_file_url(self, storage_account_name, storage_account_
...
-        self.assertEqual(data, file_content)
+        assert 1 == len(file_ranges)
+        assert 0 == file_ranges[0].get('start')
+        assert 511 == file_ranges[0].get('end')
+        assert data == file_content
 
     @FileSharePreparer()
+    @recorded_by_proxy
     def test_update_range_from_file_url_with_oauth(
             self, storage_account_name, storage_account_key):
```
jalauzon-msft, line 1023: Looks like the multi-line caused the script to miss this one. Should convert this to `kwargs`.

sdk/storage/azure-storage-file-share/tests/test_file.py
```diff
@@ -1415,19 +1652,26 @@ def test_copy_file_async_private_file_with_sas(self, storage_account_name, stora
...
-    def test_abort_copy_file(self, storage_account_name, storage_account_key, secondary_storage_account_name, secondary_storage_account_key):
+    @recorded_by_proxy
+    def test_abort_copy_file(self, **kwargs):
+        storage_account_name = kwargs.pop("storage_account_name")
+        storage_account_key = kwargs.pop("storage_account_key")
+        secondary_storage_account_name = kwargs.pop("secondary_storage_account_name")
+        secondary_storage_account_key = kwargs.pop("secondary_storage_account_key")
+
         self._setup(storage_account_name, storage_account_key, secondary_storage_account_name, secondary_storage_account_key)
-        data = b'12345678' * 1024 * 1024
```
jalauzon-msft, line 1427: Hmm my guess is that this will be flaky in the pipeline with a smaller file size since it requires the copy to take enough time so it can be aborted. Maybe we should put this back to `* 1024` but mark it as Live only and delete the recording since we really shouldn't have the 8 MiB recording?

sdk/storage/azure-storage-file-share/tests/test_file.py
```diff
@@ -1963,11 +2243,11 @@ def test_create_file_with_md5_small(self, storage_account_name, storage_account_
 
         # Assert
 
+    @pytest.mark.live_test_only
```
jalauzon-msft, line 2246: Hmm I don't really love how so many of these tests are Live only just because they are parallel. I think we are losing some testing of other useful things just because all these tests are parallel. I wonder if some of the tests should be made non-parallel so they can be recorded...

sdk/storage/azure-storage-file-share/tests/test_file.py
```diff
@@ -2089,17 +2381,20 @@ def test_sas_signed_identifier(self, storage_account_name, storage_account_key):
         content = file_client.download_file().readall()
 
         # Assert
-        self.assertEqual(self.short_byte_data, content)
+        assert self.short_byte_data == content
+
+        return variables
 
+    @pytest.mark.live_test_only
```
jalauzon-msft, line 2388: Since you are recording most other SAS tests here, any reason this one is live only? Same with some of the "shared access" ones below. It's fine if the reason is you didn't want to make all the SAS tests recorded since we're not really requiring it as part of this migration.

sdk/storage/azure-storage-file-share/tests/test_file_async.py
```diff
@@ -925,13 +1015,13 @@ async def test_update_range_from_file_url(self, storage_account_name, storage_ac
...
-        self.assertEqual(data, file_content)
+        assert 1 == len(file_ranges)
+        assert 0 == file_ranges[0].get('start')
+        assert 511 == file_ranges[0].get('end')
+        assert data == file_content
 
     @FileSharePreparer()
-    @AsyncStorageTestCase.await_prepared_test
+    @recorded_by_proxy_async
     async def test_update_range_from_file_url_with_oauth(
```
jalauzon-msft, line 1025: Convert to `kwargs`.  I think this is the same one as sync.

sdk/storage/azure-storage-file-share/tests/test_file_async.py
```diff
@@ -2345,27 +2520,31 @@ async def test_azure_named_key_credential_access(self, storage_account_name, sto
         properties = await file_client.get_file_properties()
 
         # Assert
-        self.assertIsNotNone(properties)
+        assert properties is not None
 
     @FileSharePreparer()
-    def test_account_sas_raises_if_sas_already_in_uri(self, storage_account_name, storage_account_key):
-        with self.assertRaises(ValueError):
+    @recorded_by_proxy_async
```
jalauzon-msft, line 2526: I assume the recording for this is empty? If so, you can remove the record_by_proxy since this doesn't make any network calls. This applies to sync as well.

jalauzon-msft (commented):
Thanks for tackling this one! It may be the biggest yet. Some minor comments/questions mostly left in sync but assume everything applies to async as well.

sdk/storage/azure-storage-file-share/tests/test_file.py
vincenttran-msft: Yeah I can see that. From my local testing, I found that 256 (so roughly 2MB recording) hasn't flaked on me and is a 75% reduction in recording size, so maybe we can leave it and see how it plays out? Then if it goes wrong, I'll go back and make it Live only (should be easy resolution since we know we are running this little "experiment")

jalauzon-msft (approved):

PR #26704, opened 2022-10-07
Type checking updates for CI

kristapratico:
Resolves https://github.com/Azure/azure-sdk-for-python/issues/23351

More and more, we see that our customers expect our client libraries to 1) include type hints and, 2) that our type hints are correct. We have decided to put more emphasis on typing and type checking for the Python SDK and want to begin to enforce this through CI. 

To summarize the changes in this PR:
- Changes running type checking from an opt-in check to an opt-out check in CI. The expectation is that any new library includes type hints and is type checked regularly. This matches our workflow with pylint which is also opt-out.
- Adds pyright check to CI (another type checker, like mypy. Used by pylance/VSCode)
- In addition to running on the src code, updates type checkers to also run on a library's samples dir so we can fix typing issues our customers see
- Adds verifytypes check (a feature of pyright) which reports type completeness for a library. We fail on this check only if the type completeness score reported in the PR is less than that reported from the latest release (i.e. type completeness worsens).
- Adds granular "opt-out" lists for running type checkers. In the same spirit of gradual typing, this will allow developers to incrementally improve typing in the libraries and opt-in to checks when clean.
- Updates the docs and removes outdated information about type hints.

Any libraries that previously opted-in to type checking via mypy continue to do so. Merging this PR should create no new failures, as we currently ignore packages for any of the new checks. During MQ, there is a team-wide work item to work on improving typing/type checking of libraries and this PR will enable developers to pull their libraries off the opt-out lists and work on improving our type completeness.


eng/pipelines/templates/steps/run_pyright.yml
```diff
@@ -0,0 +1,49 @@
...
+#     eng/pipelines/templates/jobs/ci.yml
+#     eng/pipelines/templates/jobs/live.test.yml
+
+# Please use `$(TargetingString)` to refer to the python packages glob string. This was previously `${{ parameters.BuildTargetingString }}`.
+steps:
+  - task: UsePythonVersion@0
+    displayName: 'Use Python 3.10'
+    inputs:
+     versionSpec: '3.10'
+    condition: and(succeededOrFailed(), ne(variables['Skip.Pyright'],'true'))
```
benbp, line 18: Is the intent that one would disable pyright for one-off pipeline runs, or that a service could exclude it for all pipelines? If the latter, I think a parameter that gets passed down to this template from `ci.yml` would be a better approach, otherwise we'd have to register variables in the pipelines which aren't as easily inspectable or changeable en masse.

kristapratico: Intent is that pyright could be excluded for all pipelines for a service. I don't imagine people will disable for one-off runs. This matches what we currently do with mypy, our other type checker, and pylint. Are you thinking that this should be updated across all static / linting checks?

eng/tox/allowed_type_checking_failures.py
```diff
@@ -0,0 +1,323 @@
...
+  "azure-synapse-artifacts",
+  "azure-synapse-managedprivateendpoints",
+  "azure-synapse-monitoring",
+  "azure-synapse-spark",
+  "azure-ai-translation-document",
+  "azure-messaging-webpubsubservice",
+]
+
+
+# --------------------------------------------------------------------------------------------------------------------
```
scbedd, line 294: If we're going to have a single file, we should add `allowed_pylint_failures` to this PR and update `run_pylint` for the same.

eng/pipelines/templates/steps/run_pyright.yml
```diff
@@ -0,0 +1,49 @@
+parameters:
+  BuildTargetingString: 'azure-*'
+  ServiceDirectory: ''
+  TestMarkArgument: ''
+  EnvVars: {}
```
scbedd, line 5: We don't run `analyze` steps during livetesting. Think about removing?

eng/tox/tox.ini
```diff
@@ -118,6 +119,40 @@ commands =
...
+    {envbindir}/python {toxinidir}/../../../eng/tox/run_pyright.py -t {toxinidir}
+
+[testenv:verifytypes]
+skipsdist = true
+skip_install = true
+usedevelop = true
+changedir = {toxinidir}
+setenv =
+  {[testenv]setenv}
+  PROXY_URL=http://localhost:5003
```
scbedd, line 148:
```suggestion
  PROXY_URL=http://localhost:5019
```

eng/tox/tox.ini
```diff
@@ -118,6 +119,40 @@ commands =
...
 
 
+[testenv:pyright]
+skipsdist = true
+skip_install = true
+usedevelop = true
+changedir = {toxinidir}
+setenv =
+  {[testenv]setenv}
+  PROXY_URL=http://localhost:5003
```
scbedd, line 129:
```suggestion
  PROXY_URL=http://localhost:5018
```

scbedd: No common PROXY_URLs across the tox environments. Don't want them to spin up proxy instances and stomp on each other 👍

scbedd (approved):
Before merging, please:

- pull pylint_allowed_failures into your allowed_type_checking and rename to like, `environment_exclusion_list` or something. (you don't need to rename immediately, I can also do it when I update the usage in tox_harness)
- Update the PROXY_URL usage even if it's nitpicky as hell.

Other than that excellent PR and thanks for driving this forward!

scbedd (approved):
🚀 🚀 🚀 🚀 

Thanks Krista! I will follow-up on this ASAP.

eng/pipelines/templates/steps/run_pyright.yml
kristapratico: Discussed offline. The tox harness will be extended to honor the common exclusion list in this PR and won't invoke a tox environment if it's unsupported. This will help save build time.

PR #26718, opened 2022-10-09
fix: combo fix for internal components I

elliotzh:
# Description

+ fix: support data-binding expression on internal node runsettings
+ fix: raise error if unknown field errors are met in load_funcs
+ fix: support is_control output for internal components
+ refactor: slim internal component spec directories
+ feat: ignore unknown fields on internal output port definition instead of raising error
+ feat: enable hdinsight.compute_name
+ feat: support dependency files in additional_includes


sdk/ml/azure-ai-ml/azure/ai/ml/entities/_validation.py
```diff
@@ -483,35 +483,58 @@ def from_single_message(
...
             obj.resolve_location_for_diagnostics(source_path, resolve_value=True)
         return obj
 
     @classmethod
-    def from_validation_messages(cls, errors: typing.Dict, data: typing.Dict):
+    def from_validation_messages(
+        cls,
+        errors: typing.Dict,
+        data: typing.Dict,
+        error_on_unknown_field: bool = False
```
brynn-code, line 510: better to have a '*' before 'error_on_unknown_field' I think, since it's a bool with default.

sdk/ml/azure-ai-ml/azure/ai/ml/_internal/_schema/input_output.py
```diff
@@ -39,19 +50,17 @@ class InternalOutputPortSchema(OutputPortSchema):
     datastore_mode = fields.Str()
 
 
+class InternalPrimitiveOutputSchema(OutputPortSchema):
+    type = DumpableEnumField(
+        allowed_values=SUPPORTED_INTERNAL_PARAM_TYPES,
+        required=True,
+    )
+    is_control = fields.Bool()
```
brynn-code, line 58: What's the behavior if there is a 'mode' specified for primitive output? 'mode' is different from 1.5 to v2, like 'mount' vs 'rw_mount' for output.

sdk/ml/azure-ai-ml/azure/ai/ml/_internal/entities/node.py
```diff
@@ -186,6 +187,15 @@ def __init__(self, **kwargs):
...
         self._init = False
 
+    @property
+    def compute_name(self) -> str:
+        """Name of the compute to be used."""
+        return self._compute_name
+
+    @compute_name.setter
+    def compute_name(self, value: str):
+        self._compute_name = value
```
brynn-code, line 185: will binding supported on this?

sdk/ml/azure-ai-ml/azure/ai/ml/operations/_component_operations.py
```diff
@@ -495,7 +496,7 @@ def preprocess_job(node):
                     # Put binding string to field
                     setattr(node, field_name, val._data_binding())
```
brynn-code, line 497: Add to supported fields above if we are going to support binding on hdi_node.compute_name

sdk/ml/azure-ai-ml/azure/ai/ml/_internal/entities/component.py
```diff
@@ -176,6 +177,21 @@ def _additional_includes(self):
...
         return InternalBaseComponentSchema(context=context)
 
+    def _validate(self, raise_error=False) -> MutableValidationResult:
+        if self._additional_includes is not None and self._additional_includes._validate().passed:
+            # update source path in case dependency file is in additional_includes
+            with self._resolve_local_code() as tmp_base_path:
+                origin_base_path, origin_source_path = self._base_path, self._source_path
+
+                try:
+                    self._base_path, self._source_path = \
```
brynn-code, line 187: This is not safe right? Nodes created from one component func will share that component, and if we are going to make the create/validate parallel in the future for pipeline perf, here is something we need to aware.

sdk/ml/azure-ai-ml/azure/ai/ml/entities/_validation.py
elliotzh: good point!

sdk/ml/azure-ai-ml/azure/ai/ml/_internal/entities/node.py
elliotzh: Should be supported by default. Let me add a test for this

sdk/ml/azure-ai-ml/azure/ai/ml/_internal/entities/component.py
elliotzh: Agree that there can be perf issue. I keep the implementation to reuse the logic to resolve local code & we may cache validation result later to resolve perf issue.

But why do you think it's not safe?

sdk/ml/azure-ai-ml/azure/ai/ml/operations/_component_operations.py
elliotzh: ... Thank you for pointing out this. Maybe we can think about if there is a better way to organize such logic later

sdk/ml/azure-ai-ml/azure/ai/ml/_internal/_schema/input_output.py
elliotzh: Interesting... I haven't found any mode specified in component definition, but it does exist on OutputPortSchema

Mode in pipeline node will follow v2 style.

wangchao1230 (approved):

brynn-code (approved):

PR #26733, opened 2022-10-10
RC

xiangyan99:


sdk/identity/azure-identity/CHANGELOG.md
```diff
@@ -1,14 +1,8 @@
...
-## 1.12.0b2 (Unreleased)
+## 1.12.0b2 (2022-10-11)
 
-### Features Added
-
-### Breaking Changes
-
-### Bugs Fixed
-
-### Other Changes
```
mccoyp, line 11: I would consider including an "Other Changes" note about the fact that the VS Code credential can again be added to the DefaultAzureCredential chain, but is now opt-in instead of opt-out.

mccoyp (approved):

sdk/identity/azure-identity/CHANGELOG.md
xiangyan99: Enable vs code cred can do nothing but fail. I don't think we want to mention it. We will add it back when we have the long-term solution implemented.

PR #26739, opened 2022-10-10
[Storage] Remove `msrest` dependency

jalauzon-msft:
Now that the generated code does not use `msrest` the dependency can be removed from all Storage packages.


vincenttran-msft (approved):
Be gone `msrest` 😄 👋

PR #26742, opened 2022-10-10
Core auth policy improvement

xiangyan99:


sdk/core/azure-core/tests/async_tests/test_authentication_async.py
```diff
@@ -226,3 +228,125 @@ def get_completed_future(result=None):
...
+        def __init__(self):
+            self._first = True
+        async def __aexit__(self, exc_type, exc_val, exc_tb):
+            pass
+        async def close(self):
+            pass
+        async def open(self):
+            pass
+
+        async def send(self, request, **kwargs):  # type: (PipelineRequest, Any) -> PipelineResponse
```
iscai-msft, line 244: can we use the testserver to mock tests?

sdk/core/azure-core/azure/core/pipeline/policies/_authentication_async.py
```diff
@@ -128,3 +132,13 @@ def on_exception(self, request: "PipelineRequest") -> None:
 
     def _need_new_token(self) -> bool:
         return not self._token or self._token.expires_on - time.time() < 300
+
+    def _need_adding_header(self, url):
```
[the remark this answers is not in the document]
xiangyan99: We do. But I don't know why we modified the code to maintain two copies. We may need to refactor the code. But not in this PR.

sdk/core/azure-core/tests/async_tests/test_authentication_async.py
xiangyan99: I am afraid testserver cannot mock the scenario that new target is in different domain.

sdk/core/azure-core/azure/core/pipeline/policies/_authentication.py
```diff
@@ -65,6 +68,18 @@ def _need_new_token(self):
...
 
 
+    def _need_adding_header(self, url):
+        if self._always_adding_header:
+            return True
+        domain = urlparse(url).netloc
+        if not self._original_domain:
+            self._original_domain = domain
+            return True
+        if self._original_domain == domain:
```
johanste, line 78: Case sensitivity?

sdk/core/azure-core/azure/core/pipeline/policies/_authentication.py
```diff
@@ -65,6 +68,18 @@ def _need_new_token(self):
         return not self._token or self._token.expires_on - time.time() < 300
 
 
+    def _need_adding_header(self, url):
```
johanste, line 71: This name doesn't resonate with me. You are trying to guard against the domain changing, right? Maybe something like "domain_changed"?

xiangyan99: It is more than domain change check. If _always_adding_header is set to True, we skip the domain parsing step and return True directly.

Are you suggesting moving the if (_always_adding_header) out of the method?

johanste: > Are you suggesting moving the if (_always_adding_header) out of the method?

Yes.

PR #26743, opened 2022-10-11
[Storage] Adjust type hints for __init__ and classmethods in file-share, queue, and changefeed

jalauzon-msft:
This is a follow-up change to #26611 which adjusts type hints for the constructors and classmethods of all clients in file-share, queue, and even changefeed. Adjustments include:

- Using the modern Python 3 type hint style
- Using `typing_extensions.Self` for return type of `@classmethod`
- Adding some missing imports for type hint of `credential`
- Adding `typing_extensions` as a dependency of file-share and queue


vincenttran-msft (approved):
Thanks for getting to this! Great customer "ask" to complete 😄

Issue #16411, opened 2021-01-28
Add support for pre-aggregated standard metrics for Azure monitor


lzchen:
Currently the Azure backend takes in request/dependency telemetry and automatically generates metrics for them (duration, count, etc) in the form of standard metrics. We want to do this generation of metrics in the SDK side.

lzchen:
Fixed by https://github.com/Azure/azure-sdk-for-python/pull/26753

PR #26753, opened 2022-10-11
Implement standard metrics - requests/duration, dependencies/duration

lzchen:
From [specs](https://github.com/microsoft/Telemetry-Collection-Spec/blob/main/OpenTelemetry/metrics/ApplicationInsightsStandardMetrics.md).

Part of https://github.com/Azure/azure-sdk-for-python/issues/16411


sdk/monitor/azure-monitor-opentelemetry-exporter/azure/monitor/opentelemetry/exporter/export/trace/_exporter.py
```diff
@@ -433,6 +436,13 @@ def _convert_span_to_envelope(span: ReadableSpan) -> TelemetryItem:
         span.attributes,
         lambda key, val: not _is_standard_attribute(key)
     )
+
+    # Standard metrics special properties
+    # Only add the property if span was generated from instrumentation that supports metrics collection
+    if span.instrumentation_scope is not None and \
+        span.instrumentation_scope.name in _INSTRUMENTATION_SUPPORTING_METRICS_LIST:
+        data.properties["_MS.ProcessedByMetricExtractors"] = "True"
```
jeremydvoss, line 444: Could you explain why we add this to ALL the telemetry from the metrics-supporting instrumentation and not just that which corresponds to the requests/duration and dependencies/duration

sdk/monitor/azure-monitor-opentelemetry-exporter/azure/monitor/opentelemetry/exporter/export/metrics/_exporter.py
```diff
@@ -174,6 +185,64 @@ def _convert_point_to_envelope(
...
+        properties["_MS.IsAutocollected"] = "True"
+        properties["request/resultCode"] = str(status_code)
+        # TODO: operation/synthetic
+        properties["cloud/roleInstance"] = tags["ai.cloud.roleInstance"]
+        properties["cloud/roleName"] = tags["ai.cloud.role"]
+        properties["Request.Success"] = str(_is_status_code_success(status_code, 500))
+    else:
+        # Any other autocollected metrics are not supported yet for standard metrics
+        # We ignore these envelopes in these cases
+        return None
```
jeremydvoss, line 233: Related to my comment in the trace exporter. Since there are other autocollected metrics we are not supporting yet, shouldn't we leave those associated telemetry w/o ProcessedByMetricExtractors.

sdk/monitor/azure-monitor-opentelemetry-exporter/azure/monitor/opentelemetry/exporter/export/metrics/_exporter.py
```diff
@@ -62,14 +65,14 @@ def export(
...
-                                    scope_metric.scope
-                                )
+                            envelope = self._point_to_envelope(
+                                point,
+                                metric.name,
+                                resource_metric.resource,
+                                scope_metric.scope
                             )
+                            if envelope is not None:
+                                envelopes.append(envelope)
```
jeremydvoss, line 75: isn't this missing an "else: envelopes = [envelope]"? Seems we're throwing away the envelope if it's the only one.

sdk/monitor/azure-monitor-opentelemetry-exporter/azure/monitor/opentelemetry/exporter/export/trace/_exporter.py
lzchen: `exceptions/count` and `traces/count` are not included in the scope of this PR. I have a PR out in defining how those will be populated [here](https://github.com/aep-health-and-standards/Telemetry-Collection-Spec/pull/136)

sdk/monitor/azure-monitor-opentelemetry-exporter/azure/monitor/opentelemetry/exporter/export/metrics/_exporter.py
lzchen: See my previous comment.

sdk/monitor/azure-monitor-opentelemetry-exporter/azure/monitor/opentelemetry/exporter/export/metrics/_exporter.py
lzchen: I'm not sure what you mean? The logic here is that it only appends to envelopes if envelope is not `None`.

sdk/monitor/azure-monitor-opentelemetry-exporter/azure/monitor/opentelemetry/exporter/export/trace/_exporter.py
jeremydvoss: got it.

sdk/monitor/azure-monitor-opentelemetry-exporter/azure/monitor/opentelemetry/exporter/export/metrics/_exporter.py
lzchen: The only case in which `envelope` can be `None` is if a metric comes in from an instrumentation that is not supported (for example ""http.server.active_requests","). An argument can be made that users might want these to show up in "customMetrics", but this would generally be confusing (since we don't advertise that these are automatically collected). So it makes more sense to just ignore them.

jeremydvoss (approved):

PR #26769, opened 2022-10-11
[Storage] Migrate `test_file_client.py` to test proxy (file-share)

vincenttran-msft:
This PR migrates `test_file_client.py` and `test_file_client_async.py` to test proxy.


sdk/storage/azure-storage-file-share/tests/recordings/test_file_client.pyTestStorageFileClienttest_closing_pipeline_client.json
```diff
@@ -0,0 +1,4 @@
+{
+  "Entries": [],
```
jalauzon-msft, line 2: Got quite a few empty recordings here. Haven't looked at the tests yet but these should all likely not need to be recorded.

sdk/storage/azure-storage-file-share/tests/test_file_client.py
```diff
@@ -44,18 +43,20 @@ def _teardown(self, FILE_PATH):
...
+        assert service.credential.account_name == self.account_name
+        assert service.credential.account_key == self.account_key
+        assert service.primary_endpoint.startswith('{}://{}.{}.core.windows.net/'.format(protocol, self.account_name, service_type)) is True
+        assert service.secondary_endpoint.startswith('{}://{}-secondary.{}.core.windows.net/'.format(protocol, self.account_name, service_type)) is True
 
     # --Direct Parameters Test Cases --------------------------------------------
     @FileSharePreparer()
-    def test_create_service_with_key(self, storage_account_name, storage_account_key):
+    @recorded_by_proxy
+    def test_create_service_with_key(self, **kwargs):
```
jalauzon-msft, line 56: Yeah, tests like this that don't make any service requests shouldn't need to be recorded. It would seem most of the tests in this file fall into this category.

jalauzon-msft (commented):
Seems a lot of tests in here don't make requests and therefore don't need recordings. If you could clean all the empty recordings up, that'd be great.

sdk/storage/azure-storage-file-share/tests/test_file_client.py
vincenttran-msft: Gotcha. I think I had asked you about this during the archboard meeting, but I realize describing it as "no service requests" rather than "empty recordings" would have more accurately described the recordings. Will fix 😄

jalauzon-msft (approved):

PR #26771, opened 2022-10-12
Fix sampleRate in AppInsightsSampler

lzchen:
Also some small doc changes.

### DIFF
--- a/sdk/monitor/azure-monitor-opentelemetry-exporter/azure/monitor/opentelemetry/exporter/export/trace/_exporter.py
+++ b/sdk/monitor/azure-monitor-opentelemetry-exporter/azure/monitor/opentelemetry/exporter/export/trace/_exporter.py
@@ -11,6 +11,7 @@ from opentelemetry.sdk.trace import ReadableSpan
 from opentelemetry.sdk.trace.export import SpanExporter, SpanExportResult
 from opentelemetry.trace import SpanKind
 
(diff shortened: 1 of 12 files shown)


sdk/monitor/azure-monitor-opentelemetry-exporter/azure/monitor/opentelemetry/exporter/export/trace/_exporter.py
```diff
@@ -45,6 +45,10 @@
     "code.",
 ]
 
+_STANDARD_AZURE_MONITOR_ATTRIBUTES = [
```
hectorhdzg, line 48: I was thinking to maybe add something at the beginning of the name to ensure those are not added in the envelope properties, we may need different ones to sampleRate and with the name you don't need to mantain the list here. Something like "azureMonitor.sampleRate" or similar, just a suggestion.

hectorhdzg (approved):
LGTM

sdk/monitor/azure-monitor-opentelemetry-exporter/azure/monitor/opentelemetry/exporter/export/trace/_exporter.py
lzchen: Using "_MS.sampleRate"

PR #26775, opened 2022-10-12
[Communication] - Phone Numbers - Use static connection string in azure-communication-phonenumber tests

danielortega-msft:
Fixes an issue in which the live tests for the azure-communication-phonenumbers package were using the connection string for the dynamically created resource instead of the static pre-configured one. This caused problems since those tests require a resource that already has acquired phone numbers in order to fully test all of the features.


sofiar-msft:
LGTM!

sofiar-msft (approved):

whisper6284 (approved):

PR #26787, opened 2022-10-12
[Pyamqp] Revert back to websocket-client

kashifkhan:
aiohttp websocket has 2 bugs when a connection is abruptly cut:

* websocket send does not throw an exception -> [issue](https://github.com/aio-libs/aiohttp/issues/2309) -> known since 2017

* websocket receive gets stuck if network is suddenly disconnected -> [PR](https://github.com/aio-libs/aiohttp/pull/5860) -> unmerged since July 2021

Reverting back to websocket-client in async as that is what most are using. There is another async websocket client but it doesnt support proxies.

### DIFF
--- a/sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_transport_async.py
+++ b/sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_transport_async.py
@@ -378,7 +378,8 @@ class AsyncTransport(
 
     async def _write(self, s):
         """Write a string out to the SSL socket fully."""
-        self.writer.write(s)
+        self.loop.run_in_executor(None, self.writer.write,s)
+        await self.writer.drain()
 
     async def close(self):
         if self.writer is not None:
@@ -386,6 +387,7 @@ class AsyncTransport(
                 # see issue: https://github.com/encode/httpx/issues/914
                 self.writer.transport.abort()
             self.writer.close()
+            await self.writer.wait_closed()
             self.writer, self.reader = None, None
         self.sock = None
         self.connected = False
--- a/sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_transport_async.py
+++ b/sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_transport_async.py
@@ -425,16 +425,25 @@ class AsyncTransport(
 
 
 class WebSocketTransportAsync(AsyncTransportMixin): # pylint: disable=too-many-instance-attributes
-    def __init__(self, host, port=WEBSOCKET_PORT, connect_timeout=None, ssl=None, **kwargs):
+    def __init__(
+        self,
+        host,
+        *,
+        port=WEBSOCKET_PORT,
+        connect_timeout=None,
+        ssl_opts=None,
+        **kwargs
+    ): # pylint: disable=unused-argument
         self._read_buffer = BytesIO()
         self.loop = asyncio.get_running_loop()
         self.socket_lock = asyncio.Lock()
-        self.sslopts = ssl if isinstance(ssl, dict) else {}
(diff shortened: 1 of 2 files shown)


sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_transport_async.py
```diff
@@ -149,6 +149,36 @@ async def send_frame(self, channel, frame, **kwargs):
         await self.write(data)
         # _LOGGER.info("OCH%d -> %r", channel, frame)
 
+class AsyncTransport(
```
swathipil, line 152: can we copy changes to SB pyamqp as well?

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_transport_async.py
```diff
@@ -425,108 +424,77 @@ async def negotiate(self):
...
-            self.connected = True
-
         except ImportError:
-            raise ValueError(
-                "Please install aiohttp library to use websocket transport."
-            )
+            raise ValueError("Please install websocket-client library to use websocket transport.")
 
-    async def _read(self, n, buffer=None, **kwargs):  # pylint: disable=unused-argument
+    async def _read(self, n, buffer=None, **kwargs):  # pylint: disable=unused-arguments
```
swathipil, line 464: might be a typo and need to be 
```suggestion
    async def _read(self, n, buffer=None, **kwargs):  # pylint: disable=unused-argument
```

swathipil: not sure if this will be a lot of work, but would you be able to compare with the sync transport/websocket to see if there were changes that need to be copied over here? I think I may have added some mypy/pylint things?

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_transport_async.py
kashifkhan: yes these changes will have to go back there too

swathipil: sorry, I meant can you copy them over in this PR? 🙂
or do we want to do that later?

kashifkhan: lol my reply wasnt clear, Ill do it in this PR itself :)

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_transport_async.py
kashifkhan: not a lot of work, I just brought in the old `WebSocketTransportAsync` and the current portions untouched. Any re-shuffling will keep all the pylint/mypy work. Will take a look at the `WebSocketTransportAsync` though

kashifkhan:
Due to issues with running websocket-client in async mode, we decided to stay with aiohttp for async. We were able to use heartbeat from aiohttp and timeouts to mitigate the bugs.

gil-cohen:
@kashifkhan im having the same issue, can you elaborate on what you did with aiohttp, which timeouts did you use?

kashifkhan:
> @kashifkhan im having the same issue, can you elaborate on what you did with aiohttp, which timeouts did you use?

@gil-cohen I passed in heartbeat parameter in ws_connect with a value of 10 seconds. That was enough time for the library to determine that the underlying connection has been lost. 

Additionally you can take a look at these 2 PRs where you can see the changes and some of the exceptions that I am now addressing

https://github.com/Azure/azure-sdk-for-python/pull/26856
https://github.com/Azure/azure-sdk-for-python/pull/27006/files

PR #26791, opened 2022-10-12
[Storage] Add `aio` extra_requires to Storage packages

jalauzon-msft:
This change adds an `extra_requires` entry for `aio` to all Storage packages that can be used to install async dependencies (`aiohttp`) when installing the Storage packages. This can be used by doing `pip install azure-storage-*[aio]`.

The extra dep is actually on the newly added `azure-core[aio]` which contains the `aiohttp` dependency. This also meant we needed to bump the `azure-core` dependency to the latest 1.26.0 to have support for the `aio` install.


vincenttran-msft (approved):
Thanks for getting this customer ask 😄

dimaqq (approved):
Nice :)

PR #26793, opened 2022-10-13
[Storage] Fix failing pipeline tests (`test_share.py`) + `name_starts_with` async bug fix

vincenttran-msft:
This PR aims to fix the pipeline failures from a recent test proxy migration

EDIT: This PR also addresses a bug that exists in the `async` client where `name_starts_with` was not being properly propagated as the `prefix`. This includes a changelog entry to document this bugfix 🐛 👍

### DIFF
--- a/sdk/storage/azure-storage-file-share/CHANGELOG.md
+++ b/sdk/storage/azure-storage-file-share/CHANGELOG.md
@@ -7,6 +7,7 @@
 ### Breaking Changes
 
 ### Bugs Fixed
+- Fixed an issue where keyword `name_starts_with` was not being passed to the service properly for the `list_shares` async API
 
 ### Other Changes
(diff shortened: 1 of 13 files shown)


jalauzon-msft (approved):
Thanks for hunting down and fixing this bug!

Looks like the Live tests are passing now which is great! I think there are probably other tests that _could_ be changed to use a unique prefix as well but they probably didn't fail in this case since there were no tests failures to cascade extra shares. Probably fine to leave it like this if everything is green but if we see flakiness in the future, we may need to update more tests in a similar fashion. 

Looks like there are a couple of recorded tests failing in this PR since they were recorded before the bug fix and used `name_starts_with` so `prefix` is not part of the recording. Should be an easy fix to re-record those.

PR #26801, opened 2022-10-13
[SchemaRegistry] add json format

swathipil:
Adding JSON format to list of supported Schema Formats.
fixes: #26002

TODO:
- [x] update tests to run for both Avro + JSON (with parametrize?)
- [x] check if we want to add Custom to this release --> yes
- [x] confirm enum value: "Json" vs. "JSON"?
- [x] confirm: JSON request content-type == JSON response content-type?
- [x] re-generate from updated swagger


sdk/schemaregistry/azure-schemaregistry/azure/schemaregistry/_common/_constants.py
```diff
@@ -34,6 +34,8 @@ class SchemaFormat(str, Enum, metaclass=CaseInsensitiveEnumMeta):
     AVRO = "Avro"
     """Represents the Apache Avro schema format."""
 
+    JSON = "JSON"
```
swathipil, line 37: TODO:
do we want to add Custom to this release?

TODO:
check with Anna/Johan, do we want "JSON" or "Json"? .NET is going with "JSON" as per the official capitalization.

deyaaeldeen (commented):
The receiving logic seems to be in line with what the service team shared so far but I am not sure which content type is used in PUT requests. I am looking at the recordings and it seems like they're all sending avro schemas?

swathipil:
> The receiving logic seems to be in line with what the service team shared so far but I am not sure which content type is used in PUT requests. I am looking at the recordings and it seems like they're all sending avro schemas?

@deyaaeldeen - yeah, so far, I've only re-recorded Avro tests. I'll be recording JSON tests with the resources created in the test-resources.json file, whenever it's valid to create a JSON SR group that way. Haven't had any luck creating a JSON SR group that way yet :(. I've just tested the changes locally with the sample files using the connection info that Eric provided.

sdk/schemaregistry/azure-schemaregistry/azure/schemaregistry/_generated/aio/operations/_operations.py
```diff
@@ -172,27 +201,43 @@ async def get_by_id(self, id: str, **kwargs: Any) -> AsyncIterator[bytes]:
 
         response = pipeline_response.http_response
 
-        if response.status_code not in [200]:
+        if response.status_code not in [200, 200]:
```
swathipil, line 204: creating an issue in autorest.python for duplication

sdk/schemaregistry/azure-schemaregistry/azure/schemaregistry/_common/_constants.py
swathipil: as per Anna: we can do "Json"

sdk/schemaregistry/azure-schemaregistry/CHANGELOG.md
```diff
@@ -1,5 +1,16 @@
 # Release History
 
+## 1.3.0b1 (Unreleased)
+
+### Features Added
+
+- `V2022_10` has been added to `ApiVersion` and set as the default api version.
+  - `Json` and `Custom` have been added supported formats in `SchemaFormat`.
```
kashifkhan, line 8: nit: have been added to supported ?

kashifkhan (approved):

sdk/schemaregistry/azure-schemaregistry/README.md
```diff
@@ -85,7 +85,7 @@ from azure.identity import DefaultAzureCredential
 from azure.schemaregistry import SchemaRegistryClient
 
 token_credential = DefaultAzureCredential()
-fully_qualified_namespace = os.environ['SCHEMA_REGISTRY_FULLY_QUALIFIED_NAMESPACE']
+fully_qualified_namespace = os.environ['SCHEMAREGISTRY_AVRO_FULLY_QUALIFIED_NAMESPACE']
```
deyaaeldeen, line 88: is this change necessary? is there automation that actually runs these code snippets?

sdk/schemaregistry/azure-schemaregistry/azure/schemaregistry/_common/_response_handlers.py
```diff
@@ -42,12 +42,22 @@ def _parse_response_schema_properties(response, format):
...
+def _get_format(content_type: str) -> SchemaFormat:
+    # pylint:disable=redefined-builtin
+    try:
+        format = content_type.split("serialization=")[1]
+        try:
+            format = SchemaFormat(format)
+        except ValueError:
+            format = SchemaFormat(format.capitalize())
+    except IndexError:
+        format = SchemaFormat.CUSTOM
```
deyaaeldeen, line 54: I am not sure how to feel about this detection logic for the custom format. On one hand, it is not precise, i.e. it doesn't check for `text/plain; charset=utf-8`. On the other hand, I don't think the service could return anything else in this case but who knows! Just bringing this up for discussion.

sdk/schemaregistry/azure-schemaregistry/README.md
swathipil: nothing runs the snippets in the README, but I wanted these to work when copy pasted + to reflect what we have in the samples.

sdk/schemaregistry/azure-schemaregistry/azure/schemaregistry/_common/_response_handlers.py
swathipil: I felt comfortable doing this since we generate from the swagger that specifies that these are the only content types that the service can return. + in Python, the convention is to use try/except rather than if/else.

I'll change this check if we find out that the service will return other content types/the swagger is updated.

deyaaeldeen: I agree that we should tailor our code to the swagger but my main concern is about the customer experience in the case the service didn't match the swagger because of a bug or something. JS autorest generates code that guards against unknown content types, see [[here]](https://github.com/Azure/azure-sdk-for-js/blob/e0b2d74424251abcdf7a27f894d18a10833aae77/sdk/automation/arm-automation/src/operations/dscConfigurationOperations.ts#L218-L244) for an example. I wonder if Python has a similar pattern and whether it is applicable here. /cc @lmazuel

sdk/schemaregistry/azure-schemaregistry/azure/schemaregistry/_schema_registry_client.py
```diff
@@ -211,6 +213,7 @@ def get_schema_properties(
                 :caption: Get schema id.
 
         """
+        format = get_case_insensitive_format(format)
```
deyaaeldeen, line 216: Is it necessary to change the case of format when building the request? or should it be enough to standardize the casing when performing the equality checking?

deyaaeldeen (approved):
I left a couple of minor remarks but overall, it looks great!

sdk/schemaregistry/azure-schemaregistry/azure/schemaregistry/_common/_response_handlers.py
swathipil: It looks like in JS, you're throwing an error. Is this error surfaced to the user? Or is it caught and handled some other way somewhere else in the code?

B/c I don't think we'd want to throw an error that's surfaced to the user. Since this is a service response, if they are returning an unknown content type, or a content type with a bug in it, then there's nothing the user can do to fix it anyway. They'd have to wait for the service fix.

Am I interpreting this correctly?

deyaaeldeen: The error is surfaced to the customer.

> Since this is a service response, if they are returning an unknown content type, or a content type with a bug in it, then there's nothing the user can do to fix it anyway. They'd have to wait for the service fix.

Yeah, but we should make them aware there is an issue on the service side instead of silently passing unexpected stuff to them.

swathipil: In the error message that you have [[here](https://github.com/Azure/azure-sdk-for-js/blob/e0b2d74424251abcdf7a27f894d18a10833aae77/sdk/automation/arm-automation/src/operations/dscConfigurationOperations.ts#L242)], it sounds like this is an error that is caused by the user/one that they can fix. It doesn't really seem to indicate that this is a service issue.

+ I don't know how likely it is that the service responses will not match the swagger. And if they don't, seems like either it would either be a bug or something that they'd notify us about + update the swagger. Thoughts @lmazuel?

sdk/schemaregistry/azure-schemaregistry/azure/schemaregistry/_schema_registry_client.py
swathipil: This format is actually used in both the request and response checking, so I thought that I'd just standardize here so that I wouldn't have to do the check twice.

sdk/schemaregistry/azure-schemaregistry/azure/schemaregistry/_common/_response_handlers.py
deyaaeldeen: Actually, this is a poor example, the content type in that case is indeed chosen by the customer/convenience-layer before sending a request. Let me check what is the logic on the receiving side.

deyaaeldeen: I was mistaken about the autorest part, the generated JS code doesn't guard against unknown content types, the `produces` list in the swagger doesn't seem to be taken into consideration in the generated code. I wonder if that is intentional and why?

In my experience in the realm of LROs, the client needs to guard against unexpected situations and error accordingly to provide a good customer experience in the case of bugs on the service side. I wonder why the same approach is not followed in the case of unexpected content types.

swathipil: Ah, I didn't know that about LROs. I'm not sure why this is the case for generated code, but might be something to ask autorest folks?

PR #26813, opened 2022-10-13
[PyAMQP] EH Bug Fix Proposal

l0lawrence:
This would be replicated on the async side as well -- now if _client_run raises/handles a ValueError and sets self._shutdown to be True, our retries will continue and we will gracefully fail after 3 retries.

 This differs from the current behavior where after a ValueError we remain in an infinite loop not receiving and not checking the boolean return of do_work.

### DIFF
--- a/sdk/eventhub/azure-eventhub/azure/eventhub/_consumer.py
+++ b/sdk/eventhub/azure-eventhub/azure/eventhub/_consumer.py
@@ -219,7 +219,18 @@ class EventHubConsumer(
             while retried_times <= max_retries:
                 try:
                     if self._open():
-                        self._handler.do_work(batch=self._prefetch)  # type: ignore
+                        running = self._handler.do_work(batch=self._prefetch)  # type: ignore
+                        if not running:
+                            last_exception = self._handle_exception(exception)
+                            retried_times += 1
+                            if retried_times > max_retries:
+                                _LOGGER.info(
+                                    "%r operation has exhausted retry. Last exception: %r.",
+                                    self._name,
+                                    last_exception,
+                                )
+                                raise last_exception
+                        # If self._shutdown is false because of a ValueError would this get propogated through
                     break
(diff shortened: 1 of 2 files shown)


sdk/eventhub/azure-eventhub/azure/eventhub/_consumer.py
```diff
@@ -212,14 +212,24 @@ def receive(self, batch=False, max_batch_size=300, max_wait_time=None):
...
                     if self._open():
-                        self._handler.do_work(batch=self._prefetch)  # type: ignore
+                        running = self._handler.do_work(batch=self._prefetch)  # type: ignore
+                        if not running:
+                            retried_times += 1
+                            if retried_times > max_retries:
+                                _LOGGER.info(
+                                    "%r operation has exhausted retry. Last exception: %r.",
+                                    self._name,
+                                    ValueError,
```
l0lawrence, line 229: this will have to change probably

sdk/eventhub/azure-eventhub/azure/eventhub/_consumer.py
```diff
@@ -212,14 +212,24 @@ def receive(self, batch=False, max_batch_size=300, max_wait_time=None):
...
-        if len(self._message_buffer) < max_batch_size:
+        if len(self._message_buffer) < max_batch_size:  #pylint: disable=too-many-nested-blocks
             # TODO: the retry here is a bit tricky as we are using low-level api from the amqp client.
             #  Currently we create a new client with the latest received event's offset per retry.
             #  Ideally we should reuse the same client reestablishing the connection/link with the latest offset.
             while retried_times <= max_retries:
                 try:
                     if self._open():
-                        self._handler.do_work(batch=self._prefetch)  # type: ignore
+                        running = self._handler.do_work(batch=self._prefetch)  # type: ignore
```
l0lawrence, line 222: now checking the return value of do_work

PR #26836, opened 2022-10-14
[Storage] Fix docs for `list_blobs` and `walk_blobs`

jalauzon-msft:
This change includes some small changes to the docstrings for `list_blobs` and `walk_blobs` around the `include` parameter.

- For `list_blobs` change the type hint docstring line from `paramtype` to `type` and update the possible include list in async.
- For `walk_blobs` update the possible include list and add additonal type `str`


vincenttran-msft (approved):
LGTM! 👍

PR #26838, opened 2022-10-14
[TA] skip tests which are failing due to service bugs

kristapratico:


sdk/textanalytics/azure-ai-textanalytics/tests/test_analyze.py
```diff
@@ -1762,7 +1762,6 @@ def test_cancel_fail_terminal_state(self, client):
             polling_interval=self._interval(),
         )
         poller.result()
-        assert poller.status() == "succeeded"
```
kristapratico, line 1765: this is returning "partiallyCompleted" sometimes, so just going to remove this assertion since it's creating a flaky test. As long as it's reached a terminal state, that should work for the purposes of this test.

deyaaeldeen (approved):

Issue #18539, opened 2021-05-06
Revisit aiohttp transport stream downloading logic for the case "content-encoding" header is 'br'


xiangyan99:
Currently we set auto_decompress to false when creating aiohttp client session and do decoding in azure-core.

We mimic the logic of aiohttp http_parser for the decoding.

One problem is if the 'content-encoding' header is 'br'. 

We don't want to have a hidden dependency on brotlipy or Brotli so for this release, we will not decode the stream if it is 'br'.

We need to revisit it.

xiangyan99:
Closing.

Will re-open if customers ask for the feature.

PR #26859, opened 2022-10-17
[core]: Activate aiothttp auto decompress feature to handle brotli decompression

nffdiogosilva:
# Description

The `BlobClient` (`from azure.storage.blob`) knows how to decompress brotli data (data with `br` encoding), because behind the scenes the `requests` lib knows how to make the actual decompression.

But when using the asynchronous `BlobClient` (`from azure.storage.blob.aio`), since it's using the `aiohttp` client, it doesn't know, by default, how to decompress brotli encoding.

I then came across this ticket #18539 which talks about the `auto_decompress` flag that you can inject into an `aiohttp` session. It was decided back then to be always `False`, but since the default value of the `aiohttp` session is `True` ([see docs](https://docs.aiohttp.org/en/stable/client_reference.html?highlight=session#client-session)) shouldn't it have that behaviour as well?

I understand the motive is to avoid having a hidden dependency installed, but I'd reckon the behaviour should be the same being applied within the synchronous client: The responsible of installing the dependency should fall over to the service/program that actually needs to compress/decompress with brotli.

I took the liberty of hardcoding to `True`. But eventually, an `ENV_VAR` controlling this might make sense...?

Here's a script to reproduce the issue:
```python

import os
import asyncio

from azure.storage.blob import ContentSettings
from azure.storage.blob import ContainerClient
from azure.storage.blob import BlobClient as SyncBlobClient
from azure.storage.blob.aio import BlobClient
import brotli

CONN_STR = os.environ["AZURE_STORAGE_CONN_STR"]
CONTAINER_NAME = "test-async-brotli"
BLOB_NAME = "brotli.json"
DATA_CONTENT = "[]".encode()


def upload_compressed_data():
    data = brotli.compress(DATA_CONTENT, quality=1)

    with ContainerClient.from_connection_string(
        CONN_STR,
        container_name=CONTAINER_NAME,
        max_single_get_size=64 * 1024 * 1024,
    ) as container_client:
        # comment if container already created.
        container_client.create_container()
        print("Uploading blob...")

        container_client.upload_blob(
            BLOB_NAME,
            data,
            content_settings=ContentSettings(
                content_type="application/json", content_encoding="br"
            ),
            overwrite=True,
        )
        print(f"Blob {BLOB_NAME} uploaded")


def read_sync_data():
    with SyncBlobClient.from_connection_string(
        CONN_STR,
        container_name=CONTAINER_NAME,
        blob_name=BLOB_NAME,
    ) as blob_data:
        sync_data = blob_data.download_blob().readall()

    assert sync_data == DATA_CONTENT, f"unexpected data: {sync_data}"


async def read_async_data():
    async with BlobClient.from_connection_string(
        CONN_STR,
        container_name=CONTAINER_NAME,
        blob_name=BLOB_NAME,
    ) as blob_data:
        async_data = await (await blob_data.download_blob()).readall()

    assert async_data == DATA_CONTENT, f"unexpected data: {async_data}"


async def main():
    upload_compressed_data()

    read_sync_data()

    await read_async_data()


if __name__ == "__main__":
    asyncio.run(main())

```

### DIFF
--- a/sdk/core/azure-core/azure/core/pipeline/transport/_aiohttp.py
+++ b/sdk/core/azure-core/azure/core/pipeline/transport/_aiohttp.py
@@ -100,7 +100,7 @@ class AioHttpTransport(AsyncHttpTransport):
             clientsession_kwargs = {
                 "trust_env": self._use_env_settings,
                 "cookie_jar": jar,
-                "auto_decompress": False,
+                "auto_decompress": True,
             }
             if self._loop is not None:
                 clientsession_kwargs["loop"] = self._loop


nffdiogosilva:
Then, since that responsibility falls upon the azure-core lib itself, wouldn't it make sense to at least address the case were a split isn't necessary, like it's currently doing with other compressions algorithms like `gzip`?

in this file: https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/core/azure-core/azure/core/utils/_pipeline_transport_rest_shared.py

```python
def _aiohttp_body_helper(response):
    # type: (PipelineTransportAioHttpTransportResponse) -> bytes
    """Helper for body method of Aiohttp responses.

    Since aiohttp body methods need decompression work synchronously,
    need to share this code across old and new aiohttp transport responses
    for backcompat.

    :rtype: bytes
    """
    if response._content is None:
        raise ValueError("Body is not available. Call async method load_body, or do your call with stream=False.")
    if not response._decompress:
        return response._content
    if response._decompressed_content:
        return response._content
    enc = response.headers.get('Content-Encoding')
    if not enc:
        return response._content
    enc = enc.lower()
        import zlib
        response._content = decompressor.decompress(response._content)
        response._decompressed_content = True
        return response._content
    return response._content
```

xiangyan99:
aiohttp implementation is different from requests and httpx.

In aiohttp, we need to decide the auto-decompression setting when creating sessions.

Given we don't know if we are in the scenario of splitting compressed file or not at that moment, we have to assume yes.

xiangyan99:
And if we want to implement the br decompression in azure-core, we need to add an extra dependency or br library.

That's the reason we choose to hold on adding the support and wait to see how many real-world scenarios that need br compression.

PR #26875, opened 2022-10-17
Update setup.py

kristapratico:
Fixing the key for package_data when including py.typed. Also removing python 3.6 since we dropped support awhile ago.


swathipil (approved):
cool, thanks!

PR #26897, opened 2022-10-18
[Storage] Hotfix for empty file download through Azurite and `list_shares` prefix issue

jalauzon-msft:
This change cherry-picks two bug fixes from `main` into the hotfix branch for a patch release.

- Fixes `ValueError` on empty blob download from Azurite for blob, datalake (calls into blob), and file-share.
- Fixes a bug where `name_starts_with` was not being sent to the service for async `list_shares` in file-share


vincenttran-msft (approved):
LGTM! That pesky `0`, `None`, `NULL`, `False` mumbo jumbo that is usually handy in Python proved to not be so helpful here 😛

PR #26899, opened 2022-10-18
add snippet update step

xiangyan99:


eng/pipelines/templates/steps/analyze.yml
```diff
@@ -115,6 +115,11 @@ steps:
       TestMarkArgument: ${{ parameters.TestMarkArgument }}
       AdditionalTestArgs: ${{parameters.AdditionalTestArgs}}
 
+  - template: ../steps/update_snippet.yml
```
weshaggard, line 118:
```suggestion
  - template: /eng/pipeines/templates/steps/update_snippet.yml
```

More portable for copy/paste if you use absolute paths from root.

eng/pipelines/templates/steps/update_snippet.yml
```diff
@@ -0,0 +1,19 @@
+parameters:
+  ServiceDirectory: ''
+  ValidateFormatting: false
+  EnvVars: {}
+
+steps:
+  - task: UsePythonVersion@0
```
weshaggard, line 7: Do you think this is going to conflict with or override any other python version usage in the job that uses this?

xiangyan99: This script should work with any py3.

Can we specify something like use any python that is available?

weshaggard: I don't think so with this task. This step mutates the global path so I just wanted to make sure it doesn't conflict with other steps in this job.

xiangyan99: We have same settings for tasks

run_pylint & run_black.

I guess this should be fine. :)

eng/pipelines/templates/steps/update_snippet.yml
```diff
@@ -0,0 +1,19 @@
...
+    displayName: 'Use Python 3.7'
+    inputs:
+     versionSpec: '3.7'
```
scbedd, line 10:
```suggestion
     versionSpec: '3.9'
```

scripts/devops_tasks/python_snippet_updater.py
```diff
@@ -0,0 +1,94 @@
...
+            print(f'Warning: snippet "{name}" is not up to date.')
+            global not_up_to_date
+            not_up_to_date = True
+            content = content.replace(s, target)
+    with open(file, 'w') as f:
+        f.write(content)
+
+
+if __name__ == "__main__":
+    arg_len = len(sys.argv)
```
scbedd, line 71: Use `argparse`. I won't block on this but definitely use `argparse`.

[Example in our own](https://github.com/Azure/azure-sdk-for-python/blob/b0982ba999c3fd1188ba44c3ca5e2e4963d34863/scripts/devops_tasks/use_pypy_version.py#L66)

eng/pipelines/templates/steps/update_snippet.yml
xiangyan99: As Scott suggested, changed it to use py 3.9

scripts/devops_tasks/python_snippet_updater.py
```diff
@@ -0,0 +1,94 @@
...
+        file_name = str(file.name)[:-3]
+        identifier = ".".join([file_name, name])
+        if identifier in snippets.keys():
+            print(f'Warning: found duplicated snippet name "{identifier}".')
+            print(file)
+        # print(f"Found: {file.name}.{name}")
+        snippets[identifier] = snippet
+
+
+def update_snippet(file):
```
scbedd, line 46: Any new scripts or updates, please add `type annotations`.

scbedd: After the big refactor that dropped py2.7 I don't add EngSys code that isn't at least type annotated and doc-commented. 👍

scripts/devops_tasks/python_snippet_updater.py
```diff
@@ -0,0 +1,94 @@
...
+        exit(1)
+    path = sys.argv[1]
+    print(f"Path: {path}")
+    for source in target_snippet_sources:
+        for py_file in Path(path).rglob(source):
+            try:
+                get_snippet(py_file)
+            except UnicodeDecodeError:
+                pass
+    # for key in snippets.keys():
```
scbedd, line 83: nit: clean code comments (even though it's clearly debugging code 👍 )

scripts/devops_tasks/python_snippet_updater.py
```diff
@@ -0,0 +1,94 @@
...
+        f.write(content)
+
+
+if __name__ == "__main__":
+    arg_len = len(sys.argv)
+    if arg_len < 2:
+        print(f"Usage: PythonSnippetUpdater <path>")
+        exit(1)
+    path = sys.argv[1]
+    print(f"Path: {path}")
```
scbedd, line 76: nit: use a generic logger instance and `logger.info` and `logger.error`.

scripts/devops_tasks/python_snippet_updater.py
```diff
@@ -0,0 +1,94 @@
...
+snippets = {}
+not_up_to_date = False
+
+target_snippet_sources = ["samples/*.py", "samples/*/*.py"]
+target_md_files = ["README.md"]
+
+def get_snippet(file):
+    with open(file, 'r') as f:
+        content = f.read()
+    pattern = "# \\[START[A-Z a-z0-9_]+\\][\\s\\S]+?# \\[END[A-Z a-z0-9_]+\\]"
```
scbedd, line 14: Just checking, given that this is a regex, nested snippets aren't supported with this. Totally fine with that yes?

xiangyan99: Right. We intentionally don't want to support nested snippets.

scripts/devops_tasks/python_snippet_updater.py
```diff
@@ -0,0 +1,94 @@
...
+    matches = re.findall(pattern, content)
+    for match in matches:
+        s = match
+        pos1 = s.index("[START")
```
scbedd, line 18: [You can use regex groups](https://python-tutorials.in/regex-capturing-groups/#:~:text=Summary%201%20Use%20a%20regex%20capturing%20group%20to,group%2C%20where%20n%20is%20the%20capturing%20group%20number.)

To simplify this code.

scripts/devops_tasks/python_snippet_updater.py
```diff
@@ -0,0 +1,94 @@
...
+        snippet = snippet[1:].rstrip()
+        if snippet[-1] == "\n":
+            snippet = snippet[:-1]
+
+        file_name = str(file.name)[:-3]
+        identifier = ".".join([file_name, name])
+        if identifier in snippets.keys():
+            print(f'Warning: found duplicated snippet name "{identifier}".')
+            print(file)
+        # print(f"Found: {file.name}.{name}")
```
scbedd, line 42: `logging.debug` instead of commenting. it won't show up by default in CI.

scripts/devops_tasks/python_snippet_updater.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+def update_snippet(file):
+    with open(file, 'r') as f:
+        content = f.read()
+    pattern = "<!-- SNIPPET:[A-Z a-z0-9_.]+-->[\\s\\S]*?<!-- END SNIPPET -->"
+    matches = re.findall(pattern, content)
+    for match in matches:
+        s = match
+        pos1 = s.index("-->")
```
scbedd, line 53: Named or just a capturing group useful here too.

eng/pipelines/templates/steps/update_snippet.yml
```diff
@@ -0,0 +1,19 @@
...
+    condition: succeededOrFailed()
+
+  - task: PythonScript@0
+    displayName: 'Update Snippets'
+    inputs:
+      scriptPath: 'scripts/devops_tasks/python_snippet_updater.py'
```
scbedd, line 16: If you want to make this part of `azure-sdk-tools` under `ci_tools` as a tool, maybe you do that and add some unit tests so that you can be certain of this new code?

scbedd (changes requested):
Lotsa feedback!

scripts/devops_tasks/python_snippet_updater.py
xiangyan99: Are you saying using re.match or re.search?

Unfortunately, it does not work as we expected. (after I re.match and get the results, match.groups is always empty)

eng/pipelines/templates/steps/update_snippet.yml
xiangyan99: Sorry, what's the difference between putting the script in this repo vs making it part of azure-sdk-tools?

scripts/devops_tasks/python_snippet_updater.py
scbedd: I'm trying to avoid having you do string index of when you already should have the data available to you.

eng/pipelines/templates/steps/update_snippet.yml
scbedd: I'm really telling you

> I want tests for this feature

xiangyan99: Also created https://github.com/Azure/azure-sdk-tools/pull/4559 to add into sdk-tools repo.

xiangyan99: moved into azure-sdk-tools folder.

tools/azure-sdk-tools/ci_tools/snippet_update/python_snippet_updater.py
```diff
@@ -0,0 +1,125 @@
...
+                get_snippet(py_file)
+            except UnicodeDecodeError:
+                pass
+    for key in snippets.keys():
+        _LOGGER.debug(f"Found snippet: {key}")
+    for target in target_md_files:
+        for md_file in Path(path).rglob(target):
+            try:
+                update_snippet(md_file)
+            except UnicodeDecodeError:
```
scbedd, line 120: Does an existing markdown file cause this currently? Which one?

scbedd (approved):
You can optionally add an entrypoint so you can use the snippet updater directly after installing azure-sdk-tools, but I'm definitely not blocking you on that.

tools/azure-sdk-tools/ci_tools/snippet_update/python_snippet_updater.py
xiangyan99: One possibly way to reproduces it is to create a new README on a Windows machine.

xiangyan99: It does not consistently repro-able.

PR #26923, opened 2022-10-19
fix ci for core

xiangyan99:


sdk/metricsadvisor/azure-ai-metricsadvisor/setup.py
```diff
@@ -67,5 +67,6 @@
         "azure-core<2.0.0,>=1.23.0",
         "msrest>=0.6.21",
         'six>=1.11.0',
+        "python-dateutil>=2.8.0"
```
kristapratico, line 70: I think this should only be a test dependency

sdk/core/azure-core/tests/async_tests/test_polling_async.py
```diff
@@ -145,9 +145,9 @@ def deserialization_callback(response):
     assert raw_poller.polling_method() is method
     done_cb.assert_called_once_with(poller)
 
-    # Test with a basic Model
-    poller = AsyncLROPoller(client, initial_response, Model, method)
-    assert poller._polling_method._deserialization_callback == Model.deserialize
+    # # Test with a basic Model
```
scbedd, line 148: Did you want to clean this?

scbedd (approved):
Nothing here looks insane to me.

sdk/core/azure-core/tests/async_tests/test_polling_async.py
xiangyan99: That's kind of todo. :)

PR #26932, opened 2022-10-19
Add registry replication count value

MilesHolland:
# Description

Adds a replication_count value to the schema, which is used when converting from local to API object representations to copy managed storage accounts, resulting in duplicated created server-side. As part of these changes, the to/from rest for storage accounts were moved from the SystemCreatedStorageAccount class to the 'parent' RegistryRegionDetails class. This change was made due to the polymorphic nature of the conversion (since the storage_config value of the RegistryRegionDetails class can be either a single SystemCreatedStorageAccount or a list of strings for user storage account ids.)

This is still a rough draft, and I have two major questions for reviewers:

 - Do you agree with the location of replication_count in the schema?
 - Do you agree with the refactoring of to/from_rest functions.


sdk/ml/azure-ai-ml/azure/ai/ml/entities/_registry/registry_support_classes.py
```diff
@@ -210,10 +178,89 @@ def _to_rest_object(self) -> RestRegistryRegionArmDetails:
...
+            ))
+            # duplicate this value based on the replication_count
+            count = storage.replication_count
+            return [deepcopy(account) for _ in range(0, count)]
+        elif storage is not None and len(storage) > 0:
+            return [make_rest_user_storage_from_id(user_id=user_id) for user_id in storage]
+        else:
+            return []
+
+    @classmethod
```
banibrata, line 215: Why does it need to be classmethod?

sdk/ml/azure-ai-ml/azure/ai/ml/entities/_registry/registry_support_classes.py
```diff
@@ -210,10 +178,89 @@ def _to_rest_object(self) -> RestRegistryRegionArmDetails:
...
-                storage) for storage in self.storage_config]
+            storages = self._storage_config_to_rest_object()
         return RestRegistryRegionArmDetails(
             acr_details=converted_acr_details,
             location=self.location,
             storage_account_details=storages,
         )
+
+
+    def _storage_config_to_rest_object(self) -> List[RestStorageAccountDetails]:
```
banibrata, line 189: Sorry may be I am missing something here, on why these two methods cannot be part of SystemCreatedStorageAccount

sdk/ml/azure-ai-ml/azure/ai/ml/entities/_registry/util.py
```diff
@@ -0,0 +1,16 @@
...
+# ---------------------------------------------------------
+# pylint:disable=protected-access,no-else-return
+
+from azure.ai.ml._restclient.v2022_10_01_preview.models import (
+    ArmResourceId as RestArmResourceId,
+    StorageAccountDetails as RestStorageAccountDetails,
+    UserCreatedStorageAccount as RestUserCreatedStorageAccount)
+
+
+def make_rest_user_storage_from_id(*, user_id: str) -> RestStorageAccountDetails:
```
banibrata, line 12: Will this be exposed, if not reccomendation would be to start with "_", to make it private in pythonic way

sdk/ml/azure-ai-ml/azure/ai/ml/entities/_registry/registry_support_classes.py
MilesHolland: Two reasons:
 - These functions need the adaptability to deal with user-created storage accounts as well. 
 - Due to the replication count feature, the parity between SystemCreatedStorageAccount and the rest equivalent is not 1-1, so I moved this logic one layer up to deal with that more easily.

sdk/ml/azure-ai-ml/azure/ai/ml/entities/_registry/registry_support_classes.py
MilesHolland: The from rest object methods need to be class methods because there's no entity-registry when this is called. The use case for this function is "We have an Autorest registry class, and we need to convert it into a local entity class registry, so we'll call Registry._from_rest_object(RestRegistry) to do the conversion." That function then recursively calls child class conversion functions, including this.

banibrata (approved):

sdk/ml/azure-ai-ml/azure/ai/ml/_schema/registry/system_created_storage_account.py
```diff
@@ -15,10 +15,13 @@
 @experimental
 class SystemCreatedStorageAccountSchema(metaclass=PatchedSchemaMeta):
     arm_resource_id = fields.Str(dump_only=True)
-    storage_account_hns = fields.Bool()
+    storage_account_hns = fields.Bool(load_default=False)
     storage_account_type = StringTransformedEnum(
+        load_default=StorageAccountType.STANDARD_LRS,
         allowed_values=[accountType.value for accountType in StorageAccountType], casing_transform=lambda x: x.lower()
     )
+    replication_count = fields.Int(load_default=1, validate=lambda count : count > 0)
```
luigiw, line 23: Minor: Usually entity is a better place to add validation, as SDK users can also get the validation.

sdk/ml/azure-ai-ml/azure/ai/ml/entities/_registry/registry.py
```diff
@@ -34,7 +34,6 @@ def __init__(
         name: str,
         location: str,
         identity: IdentityConfiguration = None,
-        description: str = None,
```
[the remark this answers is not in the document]
MilesHolland: The description value is actually an unused artifact caused by swagger inheritance - it's not actually used by the registry. Considering the fact that this is experimental, I don't think it's a major problem.

sdk/ml/azure-ai-ml/azure/ai/ml/_schema/registry/system_created_storage_account.py
MilesHolland: Added another check in the entity-API object conversion function, and checked for this in tests.

luigiw (approved):

PR #26936, opened 2022-10-20
[Maps] Add 'include_package_data' field to setup

pvaneck:
A recent [fix ](https://github.com/Azure/azure-sdk-for-python/pull/26626/files) to the py.typed verification script that is called in the CI now fails if the field `include_package_data` does not exist or does not equal `True` in a package's setup.py.

This adds this file to recently onboarded packages that did not have this.

### DIFF
--- a/sdk/maps/azure-maps-render/setup.py
+++ b/sdk/maps/azure-maps-render/setup.py
@@ -79,6 +79,7 @@ setup(
         'azure',
         'azure.maps',
     ]),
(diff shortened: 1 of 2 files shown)


pvaneck:
@scbedd Realizing now that two additional packages were added after I created my other py.typed PR, so these also needed updating in order for `maps -  ci` to pass.

scbedd (approved):
👍 🚀 🚀 🚀

PR #26937, opened 2022-10-20
[formrecognizer] Fix logging test to take into account new header

catalinaperalta:
The service is sending a new header which is getting redacted and breaking logging tests. Fixing the tests meanwhile I figure out if this header should be redacted at the INFO level.


sdk/formrecognizer/azure-ai-formrecognizer/tests/test_logging.py
```diff
@@ -48,7 +48,7 @@ def test_logging_info_dac_client(self, **kwargs):
         for message in mock_handler.messages:
             if message.levelname == "INFO":
                 # not able to use json.loads here. At INFO level only API key should be REDACTED
-                if message.message.find("Ocp-Apim-Subscription-Key") != -1:
```
catalinaperalta, line 51: Tracking issue: #26938

kristapratico (approved):

PR #26952, opened 2022-10-20
Fix phone number serialization

maximrytych-ms:
Fixed the logic of PhoneNumberIdentifier to always maintain the original phone number string whether it included the leading + sign or not.


petrsvihlik (dismissed):
please add changelog entries and version bumps for all touched packages except those that haven't been released since the introduction of the rawId (de)serialization.

sdk/communication/azure-communication-chat/azure/communication/chat/_shared/models.py
```diff
@@ -201,7 +198,7 @@ def identifier_from_raw_id(raw_id: str) -> CommunicationIdentifier:
     """
     if raw_id.startswith('4:'):
         return PhoneNumberIdentifier(
-            value='+{}'.format(raw_id[len('4:'):])
+            value='{}'.format(raw_id[len('4:'):])
```
AikoBB, line 201: do we really need formatting here? we can directly pass `raw_id[len('4:'):]` as a value

AikoBB (commented):
why the jobrouter modality was excluded from the change?

sdk/communication/azure-communication-chat/azure/communication/chat/_shared/models.py
maximrytych-ms: Fixed.

sdk/communication/azure-communication-identity/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Release History
 
+## 1.3.1 (Unreleased)
+
+### Bug Fixes
+
+- Fixed the logic of `PhoneNumberIdentifier` to always maintain the original phone number string whether it included the leading + sign or not.
```
ostoliarova-msft, line 7: As  `_shared` folder is copied across modalities, this change should be reflected in the changelog of other modalities (or at least those modalities that use `PhoneNumberIdentifier`)

AikoBB (approved):

sdk/communication/azure-communication-identity/CHANGELOG.md
maximrytych-ms: We should only mention for modalities that expose this functionality to public. Job router does not expose this to public, so we don't have to mention it in their changelog. In other modalities we didn't change anything in the shared folder, so no need in mentioning it.

AikoBB (approved):

ostoliarova-msft (approved):

JimSuplizio:
I'm going to override, the person who requested changes has been out and two others have approved it since then.

PR #26953, opened 2022-10-20
Fix #26857: separate stdout from stderr in AzureCliCredential

### LINKED ISSUE #26857: AzureCliCredential fails due to Python deprecation message getting mixed with token output
- **Package Name**: azure-identity
- **Package Version**: 1.10.0
- **Operating System**: WSL
- **Python Version**: 3.6.0

**Describe the bug**
In https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/identity/azure-identity/azure/identity/_credentials/azure_cli.py#L137, the option `"stderr": subprocess.STDOUT` makes `check_output` also include stderr in the returned string. I was trying to use this package in a container with a somewhat out-of-date python setup, which caused `az account get-access-token` to output some warnings to stderr:
RequestsDependencyWarning from python screenshot

Since `_run_command`'s return value is parsed as JSON, this caused an exception and resulted in a somewhat confusing error message:

```
azure.core.exceptions.ClientAuthenticationError: DefaultAzureCredential failed to retrieve a token from the included credentials.
...
```

**To Reproduce**
(Environment-specific, but reasoning for this issue described above)

**Expected behavior**
`AzureCliCredential` should work

micromaomao:
This PR fixes https://github.com/Azure/azure-sdk-for-python/issues/26857 in the way proposed in https://github.com/Azure/azure-sdk-for-python/issues/26857#issuecomment-1281134407. Essentially, AzureCliCredentials now treats stderr separately from stdout, and therefore if for some reason the call to `az` was successful but with some warnings, AzureCliCredentials will still work.

I have manually tested that the various error code paths still works like it does currently, by using a mocked `az` binary.


pvaneck (commented):
Thanks for the PR!

Looks like users will also hit a similar issue with the [async version](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/identity/azure-identity/azure/identity/aio/_credentials/azure_cli.py#L103) of `AzureCLICredential`.

Do you mind verifying with something like:

```python
import asyncio
from azure.identity.aio import AzureCliCredential

async def test():
    credential = AzureCliCredential()
    await credential.get_token("https://management.core.windows.net")

asyncio.run(test())
```
and applying similar changes to the async version if it fails?

micromaomao:
Hi @pvaneck - this has been addressed, please review.

sdk/identity/azure-identity/azure/identity/aio/_credentials/azure_cli.py
```diff
@@ -100,12 +100,13 @@ async def _run_command(command: str) -> str:
...
             stdout=asyncio.subprocess.PIPE,
-            stderr=asyncio.subprocess.STDOUT,
+            stderr=asyncio.subprocess.PIPE,
             cwd=working_directory,
             env=dict(os.environ, AZURE_CORE_NO_COLOR="true")
         )
-        stdout, _ = await asyncio.wait_for(proc.communicate(), 10)
-        output = stdout.decode()
+        stdout_b, stderr_b = await asyncio.wait_for(proc.communicate(), 10)
+        output = stdout_b.decode()
```
xiangyan99, line 108: Why we need this?

sdk/identity/azure-identity/CHANGELOG.md
```diff
@@ -8,6 +8,8 @@
 
 ### Bugs Fixed
 
+* `AzureCliCredential` now works even when `az` prints warnings to stderr.
```
pvaneck, line 11:
```suggestion
* `AzureCliCredential` now works even when `az` prints warnings to stderr. ([#26857](https://github.com/Azure/azure-sdk-for-python/issues/26857))
```

pvaneck (approved):
Thanks for adding the fix for the async credential!

Initially, was thinking it may be useful to log the stderr in cases where the `az` returncode is `0` but stderr is not `None`, though I am not sure how beneficial/necessary that would be. So, I think this looks good. :)

I'll wait for @xiangyan99's approval before merging.

sdk/identity/azure-identity/azure/identity/aio/_credentials/azure_cli.py
xiangyan99: I see you have

output = stdout_b.decode()
stderr = stderr_b.decode()

and seems like now we only use stderr.

Is there any reason to keep output = stdout_b.decode()?

pvaneck: output is the actual return value on line 119 :)

xiangyan99: Why can't we return stderr? :)

They are same, no?

pvaneck: From what I understand, stdout contains the actual access token json output returned by the az cli, while stderr will naturally have the error messages and sometimes warnings as in the tagged issue's case.

xiangyan99 (approved):

PR #26954, opened 2022-10-20
Remove requests dependency from PyPIClient

scbedd:
This should fix `latestdependency` from failing. @kristapratico @xiangyan99

### DIFF
--- a/eng/tox/tox.ini
+++ b/eng/tox/tox.ini
@@ -222,10 +222,11 @@ deps =
   {[packaging]pkgs}
   requests
   cryptography<4
+  urllib3
 commands =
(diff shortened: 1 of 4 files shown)


eng/tox/tox.ini
```diff
@@ -222,6 +222,7 @@ deps =
   {[packaging]pkgs}
   requests
```
kristapratico, line 223: should this be removed?

scbedd, line 223:
```suggestion
```

Removing on suggestion from @kristapratico . Thanks for keeping me honest 👍

kristapratico (approved):

scbedd:
Ok I think this is solid. Merging. A follow-up PR is coming later today to clean up the remaining `requests` usage in the proxy side.

PR #26961, opened 2022-10-20
[chore] Remove CI for unmaintained packages

pvaneck:
ApplicationInsights and LogAnalytics packages are no longer being maintained and the nightly CI continually fail for these. Here, the CI configuration is removed to disable the pipelines from running.

### DIFF
--- a/sdk/applicationinsights/ci.yml
+++ b/sdk/applicationinsights/ci.yml
@@ -27,6 +27,7 @@ extends:
   template: ../../eng/pipelines/templates/stages/archetype-sdk-client.yml
   parameters:
     ServiceDirectory: applicationinsights
+    BuildTargetingString: azure-mgmt-applicationinsights
     Artifacts:
     - name: azure-mgmt-applicationinsights
       safeName: azuremgmtapplicationinsights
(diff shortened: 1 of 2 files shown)


scbedd (approved):
Sounds great. Is there anything we want to emplace into the README that indicates it is deprecated? (Really asking, I haven't looked)

pvaneck:
@scbedd The data plane package READMEs for loganalytics and applicationinsights already have the "no longer maintained" disclaimers. However, looking at the mgmt packages, it seems these are still being maintained (e.g. https://github.com/Azure/azure-sdk-for-python/pull/24640), albeit with CI failures bypassed with check-enforcer override. 

Seems, we need to keep ci.yml after all. Is there an easy way to only run the pipeline on the `azure-mgmt-*` directories? Seems like with the case of applicationinsights, removing `azure-applicationinsights` from the `Artifacts` list of `ci.yml` stops it running tests on that package, but the `Analyze` step is still run on it (which fails due to the verify sdist step failing because `azure-applicationinsights` isn't configured correctly.

scbedd:
@pvaneck definitely is a way to do this.

image

Make this change to the `applicationinsights/ci.yml`.

pvaneck:
@scbedd Thanks! That did the trick :)

PR #26973, opened 2022-10-21
[cadl pipeline] fix for async problem of `os.makedirs` in docker 

msyyc:
https://github.com/Azure/autorest.python/issues/1514
https://github.com/Azure/azure-sdk-tools/issues/4433

### DIFF
--- a/tools/azure-sdk-tools/packaging_tools/generate_utils.py
+++ b/tools/azure-sdk-tools/packaging_tools/generate_utils.py
@@ -3,6 +3,7 @@ import json
 import logging
 import os
 import re
+import time
 
 from azure_devtools.ci_tools.git_tools import get_add_diff_file_list
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


tools/azure-sdk-tools/packaging_tools/generate_utils.py
```diff
@@ -354,6 +355,8 @@ def gen_cadl(cadl_relative_path: str, spec_folder: str) -> Dict[str, Any]:
     output_path = Path(os.getcwd()) / project_yaml["emitters"][cadl_python]["sdk-folder"]
     if not output_path.exists():
         os.makedirs(output_path)
+        while(not output_path.exists()):
+            time.sleep(1)
```
msyyc, line 359: `os.makedirs` is not sync operation so we have to make sure the path exists before running the codegen

msyyc:
The issue is fixed

PR #26989, opened 2022-10-21
[BOTRULES] ML changes

ronniegeraghty:
Matching changes from #26981

@jsquire please double check this one for me. Think I got it right, but it is Friday afternoon 😁

### DIFF
--- a/.github/fabricbot.json
+++ b/.github/fabricbot.json
@@ -1080,7 +1080,7 @@
             ]
           },
           {
-            "label": "ML-RegistryManagements",
+            "label": "ML-Registry Management",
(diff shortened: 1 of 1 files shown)


.github/fabricbot.json
```diff
@@ -1079,6 +1079,49 @@
               ""
             ]
           },
+          {
+            "label": "ML-RegistryManagements",
```
jsquire, line 1083:
```suggestion
            "label": "ML-Registry Management",
```

I've adjusted the spacing of the label to match.

jsquire (approved):
Thanks very much for putting this together!

Issue #10277, opened 2020-03-12
[azure-storage-blob] AuthenticationFailed when metadata keys contain underscores


bno1:
- **Package Name**: azure-storage-blob
- **Package Version**: 12.2.0
- **Operating System**: Ubuntu 18.04
- **Python Version**: Python 2.7.17

**Describe the bug**
Azure seems to sort metadata keys containing underscores differently, causing a `AuthenticationFailed` error.

In the example below the metadata keys are: `i0` and `i_`. Lexicographically, `i0` < `i_`, but in the Azure response (in the canonicalized request)  `i_` is placed before `i0`.

**To Reproduce**
```python
blob_service_client = BlobServiceClient(
    account_url="http://<...>.blob.core.windows.net",
    credential="<...>"
)

blob_client = blob_service_client.get_blob_client('mybucket', 'blob')

blob_client.upload_blob(
    '',
    metadata={
        'i0': 'a',
        'i_': 'a',
    })
```

Creates the following HTTP request:
```
PUT /mybucket/blob HTTP/1.1
Host: <...>.blob.core.windows.net
x-ms-version: 2019-07-07
Accept-Encoding: gzip, deflate
x-ms-blob-type: BlockBlob
Content-Length: 0
Accept: */*
User-Agent: azsdk-python-storage-blob/12.2.0 Python/2.7.17 (Linux-5.5.8-arch1-1-x86_64-with-Ubuntu-18.04-bionic)
Connection: keep-alive
x-ms-client-request-id: 9386d62c-6474-11ea-a2cd-0242ac130002
x-ms-meta-i_: a
x-ms-meta-i0: a
If-None-Match: *
Content-Type: application/octet-stream
Authorization: SharedKey <...>:<...>
x-ms-date: Thu, 12 Mar 2020 15:17:24 GMT
```

And Azure responds with the following error:
```
HTTP/1.1 403 Server failed to authenticate the request. Make sure the value of Authorization header is formed correctly including the signature.
Content-Length: 809
Content-Type: application/xml
Server: Microsoft-HTTPAPI/2.0
x-ms-request-id: d5cab1f8-801e-002e-6f81-f81f83000000
x-ms-error-code: AuthenticationFailed
Date: Thu, 12 Mar 2020 15:17:24 GMT

[...]

PR #26991, opened 2022-10-21
[Storage] Change algorithm used to sort headers

vincenttran-msft:
This PR aims to solve the differences between how Python's built-in sort algorithm sorts the headers compared to the service. Although this is usually not problematic as headers are quite contrived and restricted to what values they can be, the issue arises with `metadata` where it allows user input to be part of the headers we are going to sort. #10277

With that in mind, we discovered that only ASCII values are acceptable on the service side, making our acceptable character-space much smaller.

Then we can eliminate ASCII values 0-32 as they are nonsensical (`NULL`, `LF`, `Space`, etc.) and also 127 (`DEL`).

This leaves us with the following characters (following ASCII value sorting scheme): 
"!\"#$%&\'()*+,./-^_`0123456789:;<=>? @ ABCDEFGHIJKLMNOPQRSTUVWXYZ[]abcdefghijklmnopqrstuvwxyz{|}~"

Now, we need to check what values are accepted once we send the headers off for authentication. And so, the following characters either fail in the `_is_legal_header_name` check or simply throws a `Bad Request` (comma separated):
:, ", `, ^, (, ), comma, /, \, ;, <, =, >, ?, @, [, ], {, }

Now at this point, if we use our pared-down list of the ASCII values and remove all the illegal characters as stated above, the sort will still fail if we use these weights (ASCII-value as the weight), as the service **does not** follow ASCII value weight convention.

And so, after much testing, we finally land on this as the proper weighting to match how service is sorting the characters:
"**-!#$%&*.^_|~+\"\'(),/`~0123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[]abcdefghijklmnopqrstuvwxyz{}**"


sdk/storage/azure-storage-blob/azure/storage/blob/_shared/authentication.py
```diff
@@ -48,6 +48,33 @@ def _wrap_exception(ex, desired_type):
     # TODO: In the future we will log the trace
     return desired_type('{}: {}'.format(ex.__class__.__name__, msg))
 
+# This method attempts to emulate the sorting done by the service
+@staticmethod
+def _service_sort(input_header):
```
vincenttran-msft, line 53: Open to naming suggestions 😄

sdk/storage/azure-storage-blob/azure/storage/blob/_shared/authentication.py
```diff
@@ -48,6 +48,33 @@ def _wrap_exception(ex, desired_type):
...
+
+    for k, v in input_header:
+        header_dict[k] = v
+        header_keys.append(k)
+
+    # Sort according to custom defined weights
+    try:
+        header_keys = sorted(header_keys, key=lambda word: [custom_weights.index(c) for c in word])
+    except ValueError:
+        raise ValueError("Illegal character encountered when sorting headers.")
```
vincenttran-msft, line 69: Open to suggestions for what error to raise. Although it already was raising a `ValueError` even without putting this logic here, the error was rather useless to users as it could be considered sort of cryptic, but what it really meant was that a character it attempted to sort is not in our custom alphabet.

Sample error without this logic:
```
>   header_keys = sorted(header_keys, key=lambda word: [custom_weights.index(c) for c in word])
E   ValueError: substring not found
```

sdk/storage/azure-storage-blob/azure/storage/blob/_shared/authentication.py
jalauzon-msft, line 53: Four comments here:
- How about something like `_storage_header_sort`?
- Why is this a `@staticmethod`? It's not in a class so does it need that?
- nit: Can we change the param name to be plural `input_headers` just so its more obvious it contains all the headers.
- Can we add type hints for this? I know it's just a helper method but I'm trying to get into the habit of adding Python 3 style type hints for all new functions.

To sum it all up, something like this (assuming these are the right types)
```py
def _storage_header_sort(input_headers: List[Tuple(str, str)]) -> List[Tuple(str, str)]):
```

vincenttran-msft: 1. Sounds good to me!
2. Nice catch! I originally had this as a class method then copied and pasted it out, removed 😄 
3. 👍 
4. 👍  will have to import `from Typing import List, Tuple` but I also think this is very helpful for future maintainers so worth it!

sdk/storage/azure-storage-blob/azure/storage/blob/_shared/authentication.py
```diff
@@ -48,6 +48,33 @@ def _wrap_exception(ex, desired_type):
     # TODO: In the future we will log the trace
     return desired_type('{}: {}'.format(ex.__class__.__name__, msg))
 
+# This method attempts to emulate the sorting done by the service
+@staticmethod
+def _service_sort(input_header):
+    # Define the custom alphabet for weights
+    custom_weights = "-!#$%&*.^_|~+\"\'(),/`~0123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[]abcdefghijklmnopqrstuvwxyz{}"
```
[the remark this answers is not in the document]
vincenttran-msft: Good point. Done!

jalauzon-msft (approved):

PR #27034, opened 2022-10-25
[pyAMQP] Stress fixes for aiohttp and valueError

l0lawrence:
To be merged after Kashif's PR

### DIFF
--- a/sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_transport.py
+++ b/sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_transport.py
@@ -744,4 +744,4 @@ class WebSocketTransport(_AbstractTransport):
             raise ConnectionError('send disconnected by SSL (%s)' % e)
         except WebSocketConnectionClosedException as e:
             raise ConnectionError('send disconnected (%s)' % e)
-
+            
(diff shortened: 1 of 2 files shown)


l0lawrence:
Right now, the error I am catching is OSerror to get the aiohttp.client session closed error

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_transport.py
```diff
@@ -416,7 +416,7 @@ def read(self, verify_frame_type=0):
             offset = frame_header[4]
             frame_type = frame_header[5]
             if verify_frame_type is not None and frame_type != verify_frame_type:
-                raise ValueError(
+                _LOGGER.debug(
                     f"Received invalid frame type: {frame_type}, expected: {verify_frame_type}"
```
kashifkhan, line 420: We want to use printf style for logging for perf reasons :) Basically we want to hold back formatting this string till the very last minute (or when there is actually a problem). f-string or any other will evaluate it even if there is no exception

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_transport_async.py
```diff
@@ -101,7 +102,7 @@ async def read(self, verify_frame_type=0):
                 offset = frame_header[4]
                 frame_type = frame_header[5]
                 if verify_frame_type is not None and frame_type != verify_frame_type:
-                    raise ValueError(
+                    _LOGGER.debug(
```
kashifkhan, line 105: same as previous comment :)

kashifkhan (approved):

Issue #26229, opened 2022-09-14
[EventHubs] SDK layer exception parity for uamqp/pyamqp


swathipil:
- EH error story: add a bunch of tests to for all test scenarios. and check that uamqp and pyamqp raise the same errors. check that we have tests for optional params, like timeout flags etc.

Test cases:
- [x] 1) Sending EventHubProducerClient(...connection_verify="cacert.pem").
    - Behavior: `uamqp`- pytest.raises `EventHubError`. pyamqp - pytest.raises `ConnectError`. --> Issue here : #27128
    - Expected: Both result in `EventHubError`.
  - Probably b/c uamqp takes connection_verify in to[ JWTTokenAuth](https://github.com/Azure/azure-sdk-for-python/blob/cb2c5b7745f211783450477885be353642dfb1c3/sdk/eventhub/azure-eventhub/azure/eventhub/_transport/_uamqp_transport.py#L485). In pyamqp, verify is taken by [the Send/ReceiveClient](https://github.com/Azure/azure-sdk-for-python/blob/35c5a7aaf90516fc9554b1311a22d161cf0677b5/sdk/eventhub/azure-eventhub/azure/eventhub/_transport/_pyamqp_transport.py#L253) and [the values being sent in to JWTTokenAuth are disregarded](https://github.com/Azure/azure-sdk-for-python/blob/35c5a7aaf90516fc9554b1311a22d161cf0677b5/sdk/eventhub/azure-eventhub/azure/eventhub/_transport/_pyamqp_transport.py#L438). Fix this.
- [x] 2) Buffered Producer with send_batch(), (potentially on both sync/async unsure) doesn't send messages on  AttributeError("'list' object has no attribute '_internal_events'")} 
- [x] 3) Setting retry_total>0 and 0 and have consumer client error in 2 ways:
    - retry_total>0; pass in a SASTokenCredential that expires quickly. See how client + eventprocessor behaves. **should call error callback** --> Issue created for this (#27137)
    - retry_total=0, Force link detach by updating properties of EH while receiving. If retry set to 0, should call error callback right away and continue running.
    - should these close and re-open consumer and partition?
    - should only non-retryable errors close the partition?
```python
import os
import sys
from azure.eventhub import EventHubConsumerClient

CONNECTION_STR = os.environ["EVENT_HUB_CONN_STR"]
EVENTHUB_NAME = os.environ['EVENT_HUB_NAME']

import logging
[...]

Issue #24846, opened 2022-06-14
[Messaging] Add In Negative Testing


kashifkhan:
To maintain parity between exceptions between pyamqp and uamqp add in negative tests to expose behavior/exceptions during a failure.

cases to test:
- [x] when creating client with DefaultAzureCredential, pass in `fully_qualified_namespace=None`. 
 > - In pyamqp: gives us `'can only concatenate str (not "NoneType") to str\ncan only concatenate str (not "NoneType") to str')`. This is b/c the AMQPClient._hostname = None, so in `def _open`, creating Connection object with `"amqps://" + self._hostname` gives us error.
- [x] passing None in as sas_policy and key in `EventHubSharedKeyCredential`
 > - in pyamqp, error:
 > - in uamqp, error:
- [x] token auth failure/bad signature:
> - pyamqp:
azure.eventhub._pyamqp.error.TokenAuthFailure: CBS Token authentication failed.
Status code: 401
Description: InvalidSignature: The token has an invalid signature. TrackingId:14e015a9-f58f-44ec-8030-0faf2ec77923, SystemTracker:NoSystemTracker, Timestamp:2022-05-12T16:19:46
> - for uamqp error is: Authentication Put-Token failed. Retries exhausted.
- [x]  for credentials: note that a bad/expired token will be checked during client creation. For pyamqp, it looks like this token failure is not being retried. Need to double check

l0lawrence:
testing timeout in client.py

Issue #24680, opened 2022-06-02
[AMQP Python] Exception Parity


l0lawrence:
Currently between uamqp and Pyamqp we have different exceptions being handled (i.e Websocket TimeoutError on Pyamqp), we want to ensure that both uamqp and pyamqp have similar exceptions so that customers get similar exceptions across the board.

kashifkhan:
@swathipil this will come in to play when you are working on the uamqp/pyamqp switch

PR #27065, opened 2022-10-26
[EventHubs] kwargs/error testing

swathipil:
fixes: #26229
fixes: #24846
fixes: #24680

Table of kwargs tested are in the issue above.
* Most tests added were for EventHubConsumerClient/ProducerClient constructor kwargs.
* Added tests for invalid key/policy name in the EventHubSharedKeyCredential.
* A lot of the params for send_event/send_batch were already being tested.
* receive/receive_batch kwargs seemed to pretty thoroughly tested as well.
* A few errors were caught by changing permissions/settings on the EH. (ex. Disabling the EH in the middle of receiving, Not adding the EH Data Owner Role + using AAD)
* sync consumer is not raising/handling errors correctly for retry_total > 0. This was happening in previously GA'ed versions as well, so can be addressed later. Issue created: #27137

### DIFF
--- a/sdk/eventhub/azure-eventhub/azure/eventhub/_client_base.py
+++ b/sdk/eventhub/azure-eventhub/azure/eventhub/_client_base.py
@@ -411,7 +411,7 @@ class ClientBase(object):  # pylint:disable=too-many-instance-attributes
                 conn = self._conn_manager.get_connection(  # pylint:disable=assignment-from-none
                     host=self._address.hostname, auth=mgmt_auth
                 )
-                mgmt_client.open(connection=conn)
+                self._amqp_transport.open_mgmt_client(mgmt_client, conn)
                 while not mgmt_client.client_ready():
                     time.sleep(0.05)
                 mgmt_msg.application_properties[
--- a/sdk/eventhub/azure-eventhub/azure/eventhub/_client_base.py
+++ b/sdk/eventhub/azure-eventhub/azure/eventhub/_client_base.py
@@ -559,10 +559,10 @@ class ConsumerProducerMixin(object):
         self._close_handler()
         self._client._conn_manager.reset_connection_if_broken()  # pylint: disable=protected-access
 
-    def _handle_exception(self, exception):
+    def _handle_exception(self, exception, *, is_consumer=False):
         exception = self._amqp_transport.check_timeout_exception(self, exception)
         return self._amqp_transport._handle_exception(  # pylint: disable=protected-access
-            exception, self
+            exception, self, is_consumer=is_consumer
         )
 
     def _do_retryable_operation(self, operation, timeout=None, **kwargs):
--- a/sdk/eventhub/azure-eventhub/azure/eventhub/_client_base.py
+++ b/sdk/eventhub/azure-eventhub/azure/eventhub/_client_base.py
@@ -411,7 +411,7 @@ class ClientBase(object):  # pylint:disable=too-many-instance-attributes
                 conn = self._conn_manager.get_connection(  # pylint:disable=assignment-from-none
                     host=self._address.hostname, auth=mgmt_auth
                 )
-                self._amqp_transport.open_mgmt_client(mgmt_client, conn)
+                mgmt_client.open(connection=conn)
                 while not mgmt_client.client_ready():
                     time.sleep(0.05)
                 mgmt_msg.application_properties[
@@ -432,8 +432,14 @@ class ClientBase(object):  # pylint:disable=too-many-instance-attributes
                     return response
                 raise self._amqp_transport.get_error(status_code, description)
             except Exception as exception:  # pylint: disable=broad-except
+                # is_consumer=True passed in here, ALTHOUGH this method is shared by the producer and consumer.
+                # is_consumer will only be checked if FileNotFoundError is raised by self.mgmt_client.open() due to
+                # invalid/non-existent connection_verify filepath. The producer will encounter the FileNotFoundError
+                # when opening the SendClient, so is_consumer=True will not be passed to amqp_transport.handle_exception
+                # there. This is for uamqp exception parity, which raises FileNotFoundError in the consumer and
+                # EventHubError in the producer. TODO: Remove `is_consumer` kwarg when resolving issue #27128.
                 last_exception = self._amqp_transport._handle_exception(  # pylint: disable=protected-access
-                    exception, self
+                    exception, self, is_consumer=True
                 )
                 self._backoff(
                     retried_times=retried_times, last_exception=last_exception
--- a/sdk/eventhub/azure-eventhub/azure/eventhub/_consumer.py
+++ b/sdk/eventhub/azure-eventhub/azure/eventhub/_consumer.py
@@ -223,8 +223,6 @@ class EventHubConsumer(
                     break
                 except Exception as exception:  # pylint: disable=broad-except
                     self._amqp_transport.check_link_stolen(self, exception)
-                    if not self.running:  # exit by close
-                        return
                     if self._last_received_event:
                         self._offset = self._last_received_event.offset
                     last_exception = self._handle_exception(exception)
--- a/sdk/eventhub/azure-eventhub/azure/eventhub/_consumer.py
+++ b/sdk/eventhub/azure-eventhub/azure/eventhub/_consumer.py
@@ -229,7 +229,7 @@ class EventHubConsumer(
                         return
                     if self._last_received_event:
                         self._offset = self._last_received_event.offset
-                    last_exception = self._handle_exception(exception)
+                    last_exception = self._handle_exception(exception, is_consumer=True)
                     retried_times += 1
                     if retried_times > max_retries:
                         _LOGGER.info(
--- a/sdk/eventhub/azure-eventhub/azure/eventhub/_consumer.py
+++ b/sdk/eventhub/azure-eventhub/azure/eventhub/_consumer.py
@@ -223,6 +223,10 @@ class EventHubConsumer(
                     break
                 except Exception as exception:  # pylint: disable=broad-except
                     self._amqp_transport.check_link_stolen(self, exception)
+                    # TODO: below block hangs when retry_total > 0
+                    # need to remove/refactor, issue #27137
+                    if not self.running:  # exit by close
+                        return
                     if self._last_received_event:
                         self._offset = self._last_received_event.offset
                     last_exception = self._handle_exception(exception)
--- a/sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py
+++ b/sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py
@@ -672,7 +672,7 @@ class EventHubProducerClient(
                 )
                 if self._on_success:
                     self._on_success(batch._internal_events, pid)
-            except (KeyError, AttributeError, EventHubError) as e:
+            except (KeyError, AttributeError, EventHubError) as e:  # TODO: do we want all EventHubErrors to be caught?
                 _LOGGER.debug(
                     "Producer for partition ID %s not available: %s. Rebuilding new producer.",
                     partition_id,
--- a/sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py
+++ b/sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py
@@ -672,7 +672,7 @@ class EventHubProducerClient(
                 )
                 if self._on_success:
                     self._on_success(batch._internal_events, pid)
-            except (KeyError, AttributeError, EventHubError) as e:  # TODO: do we want all EventHubErrors to be caught?
+            except (KeyError, AttributeError, EventHubError) as e:
                 _LOGGER.debug(
                     "Producer for partition ID %s not available: %s. Rebuilding new producer.",
                     partition_id,
--- a/sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_connection.py
+++ b/sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_connection.py
@@ -188,13 +188,13 @@ class Connection(object):  # pylint:disable=too-many-instance-attributes
             else:
                 self._set_state(ConnectionState.HDR_SENT)
         except (OSError, IOError, SSLError, socket.error) as exc:
+            if isinstance(exc, FileNotFoundError) and exc.filename and "ca_certs" in exc.filename:
+                raise
             raise AMQPConnectionError(
                 ErrorCondition.SocketError,
                 description="Failed to initiate the connection due to exception: " + str(exc),
                 error=exc,
             )
-        except Exception:  # pylint:disable=try-except-raise
-            raise
 
     def _disconnect(self):
         # type: () -> None
--- a/sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_connection.py
+++ b/sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_connection.py
@@ -188,6 +188,8 @@ class Connection(object):  # pylint:disable=too-many-instance-attributes
             else:
                 self._set_state(ConnectionState.HDR_SENT)
         except (OSError, IOError, SSLError, socket.error) as exc:
+            # FileNotFoundError is being raised for exception parity with uamqp when invalid
+            # `connection_verify` file path is passed in. Remove later when resolving issue #27128.
             if isinstance(exc, FileNotFoundError) and exc.filename and "ca_certs" in exc.filename:
                 raise
             raise AMQPConnectionError(
--- a/sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_transport.py
+++ b/sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_transport.py
@@ -36,6 +36,7 @@
 from __future__ import absolute_import, unicode_literals
 
 import errno
+from multiprocessing import AuthenticationError
 import re
 import socket
 import ssl
@@ -57,6 +58,7 @@ from .constants import (
     TransportType,
     AMQP_WS_SUBPROTOCOL,
 )
+from .error import AuthenticationException, ErrorCondition
 
 
 try:
@@ -495,7 +497,11 @@ class SSLTransport(_AbstractTransport):
 
     def _setup_transport(self):
         """Wrap the socket in an SSL object."""
-        self.sock = self._wrap_socket(self.sock, **self.sslopts)
+        try:
+            self.sock = self._wrap_socket(self.sock, **self.sslopts)
+        except FileNotFoundError as exc:
+            # TODO: invalid connection_verify, should we raise some other error?
+            raise 
         self.sock.do_handshake()
         self._quick_recv = self.sock.recv
 
@@ -684,7 +690,7 @@ class WebSocketTransport(_AbstractTransport):
             if username or password:
                 http_proxy_auth = (username, password)
         try:
-            from websocket import create_connection
+            from websocket import create_connection, WebSocketAddressException
 
             self.ws = create_connection(
                 url="wss://{}".format(self._custom_endpoint or self._host),
@@ -695,6 +701,12 @@ class WebSocketTransport(_AbstractTransport):
                 http_proxy_host=http_proxy_host,
                 http_proxy_port=http_proxy_port,
                 http_proxy_auth=http_proxy_auth,
+            )
+        except WebSocketAddressException as exc:
+            raise AuthenticationException(
+                ErrorCondition.SocketError, # TODO: ClientError?
+                description="Failed to authenticate the connection due to exception: " + str(exc),
+                error=exc,
             )
         except ImportError:
             raise ValueError(
--- a/sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_transport.py
+++ b/sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_transport.py
@@ -417,8 +417,8 @@ class _AbstractTransport(object):  # pylint: disable=too-many-instance-attribute
             offset = frame_header[4]
             frame_type = frame_header[5]
             if verify_frame_type is not None and frame_type != verify_frame_type:
-                raise ValueError(
-                    f"Received invalid frame type: {frame_type}, expected: {verify_frame_type}"
+                _LOGGER.debug(
+                    "Received invalid frame type: %r, expected: %r", frame_type, verify_frame_type
                 )
 
             # >I is an unsigned int, but the argument to sock.recv is signed,
--- a/sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_transport.py
+++ b/sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_transport.py
@@ -555,6 +555,10 @@ class SSLTransport(_AbstractTransport):
         try:
             sock = ssl.wrap_socket(**opts)  # pylint: disable=deprecated-method
         except FileNotFoundError as exc:
+            # FileNotFoundError does not have missing filename info, so adding it below.
+            # Assuming that this must be ca_certs, since this is the only file path that
+            # users can pass in (`connection_verify` in the EH/SB clients) through opts above.
+            # For uamqp exception parity. Remove later when resolving issue #27128.
             exc.filename = {"ca_certs": ca_certs}
             raise exc
         # Set SNI headers if supported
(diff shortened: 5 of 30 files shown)


sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_transport.py
```diff
@@ -36,6 +36,7 @@
 from __future__ import absolute_import, unicode_literals
 
 import errno
+from multiprocessing import AuthenticationError
```
kashifkhan, line 39: this is the wrong import :) 

should be from eventhub exceptions

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_transport.py
```diff
@@ -696,6 +702,12 @@ def connect(self):
                 http_proxy_port=http_proxy_port,
                 http_proxy_auth=http_proxy_auth,
             )
+        except WebSocketAddressException as exc:
+            raise AuthenticationException(
+                ErrorCondition.SocketError, # TODO: ClientError?
```
kashifkhan, line 707: SocketError is fine here since the address not found is a SocketException.

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_transport.py
swathipil: it must've auto-imported, good catch!

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_transport_async.py
```diff
@@ -480,15 +481,25 @@ async def connect(self):
...
+                    protocols=[AMQP_WS_SUBPROTOCOL],
+                    autoclose=False,
+                    proxy=http_proxy_host,
+                    proxy_auth=http_proxy_auth,
+                    ssl=self.sslopts,
+                )
+            except ClientConnectorError as exc:
+                if self._custom_endpoint:
+                    raise AuthenticationException(
+                        ErrorCondition.SocketError, # TODO: ClientError?
```
kashifkhan, line 497: same comment as above, does ClientConnectError pop up when there is an auth issue ?

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_transport.py
swathipil: yeah I think socket error makes sense to me too

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_transport_async.py
swathipil: this happens when the custom_endpoint_address that is passed in is invalid.

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_transport.py
kashifkhan, line 39: This should from azure.eventhub :)

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_transport.py
kashifkhan: Then I would be more inclined towards raising a ConnectionError versus an Auth error

sdk/eventhub/azure-eventhub/azure/eventhub/_consumer.py
```diff
@@ -223,8 +223,6 @@ def receive(self, batch=False, max_batch_size=300, max_wait_time=None):
                     break
                 except Exception as exception:  # pylint: disable=broad-except
                     self._amqp_transport.check_link_stolen(self, exception)
-                    if not self.running:  # exit by close
-                        return
```
swathipil, line 227: @annatisch - removing this results in raising the exception once the retries are exhausted. The event processor catches this except, processes the error, and closes the consumer + partition with "OwnershipLost" as the reason. It'll then re-initialize the partition and consumer and continue running.

If this is left in here, when retry_total is > 0, the error callback is never called b/c the error is never raised and caught by the event processor.

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_transport.py
```diff
@@ -495,7 +497,11 @@ def __init__(
 
     def _setup_transport(self):
         """Wrap the socket in an SSL object."""
-        self.sock = self._wrap_socket(self.sock, **self.sslopts)
+        try:
+            self.sock = self._wrap_socket(self.sock, **self.sslopts)
+        except FileNotFoundError as exc:
+            # TODO: invalid connection_verify, should we raise some other error?
+            raise
```
swathipil, line 504: just a placeholder until we figure out if current behavior is correct.

@annatisch - Should invalid connection_verify result in a ConnectError?
Currently for uamqp, this scenario results in `FileNotFoundError` for the ConsumerClient and `EventHubError` for ProducerClient. Given that uamqp is raising different errors, can we say that this was a bug in our previous code? And that our current error handling for this scenario is correct?

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_transport.py
swathipil: In that case, this would be a breaking change since for uamqp, we currently raise an Authentication error for an invalid custom endpoint address.

This one seems like AuthenticationException _could_ make sense, so I'm thinking that a breaking change here wouldn't be worth it. Thoughts? @annatisch

kashifkhan: Agree, for keeping the transition to pyamqp smooth its better to be consistent now and then be right later.

annatisch: ConnectionError probably _would_ have made more sense - but I'm inclined to not change the behaviour.
Though I do question the use of Condition.SocketError - is that also what we were using before?
Seems like that should be ClientError to me...

sdk/eventhub/azure-eventhub/azure/eventhub/_consumer.py
annatisch: That seems reasonable - though a couple of things to maybe investigate before we simply remove this:
- This is in a broad-except clause, could we better resolve this by using a finer-grained except clause? For example, perhaps if `close()` is called, it raises a particular error that we can `return` on?
- Why is `self.running` set to `False` in the case of the error scenario you describe? Perhaps we should only be explicitly settings this to `False` in the `close()` scenario?

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_transport.py
annatisch: To clarify - what does "invalid" mean in this context?
Does it mean that value supplied was unusable, and failed client-side?
Or does it mean we attempted to use the value and it failed to connect?

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_transport.py
swathipil: It was previously a SocketError as well, and the previous error was caught and an AMQPConnectionError raised [[here](https://github.com/Azure/azure-sdk-for-python/blob/5013c2d8464a518ff4f2a6ca691f7913e1620c6e/sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_connection.py#L190)].

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_transport_async.py
swathipil: keeping same as old error

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_transport.py
swathipil: keeping old behavior including SocketError ErrorCondition.

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_transport.py
swathipil: OOB disussion: keeping EventHubError and FileNotFoundError for now for parity. Creating an issue to potentially fix later here:

sdk/eventhub/azure-eventhub/azure/eventhub/_consumer.py
swathipil: TL;DR - I've reverted this and created an issue (#27137) to resolve this after GA, since it seems to be a bug from before in previously GA'ed versions.

More details:
So running is being set to False b/c:
 1) The first time there's an exception here, consumer._handle_exception is called on line 230.
 2) In handle_exception, if the error is an EventHubError, we call [consumer.close_handler()](https://github.com/Azure/azure-sdk-for-python/blob/84f05915c7a7d66c0f11b03d8b290e337ee970ac/sdk/eventhub/azure-eventhub/azure/eventhub/_transport/_uamqp_transport.py#L609)
3) Inside close_handler(), we close the consumer and set `self.running = False`.

sdk/eventhub/azure-eventhub/azure/eventhub/_client_base.py
```diff
@@ -411,7 +411,7 @@ def _management_request(
                 conn = self._conn_manager.get_connection(  # pylint:disable=assignment-from-none
                     host=self._address.hostname, auth=mgmt_auth
                 )
-                mgmt_client.open(connection=conn)
+                self._amqp_transport.open_mgmt_client(mgmt_client, conn)
```
swathipil, line 414: added to handle different errors

sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py
```diff
@@ -672,7 +672,7 @@ def send_batch(self, event_data_batch, **kwargs):
                 )
                 if self._on_success:
                     self._on_success(batch._internal_events, pid)
-            except (KeyError, AttributeError, EventHubError) as e:
+            except (KeyError, AttributeError, EventHubError) as e:  # TODO: do we want all EventHubErrors to be caught?
```
swathipil, line 675: @kashifkhan - do we want all EventHubErrors to be caught? b/c, for ex, OperationSendTimeout gets caught and re-run here, even though it can be raised. (It'll be raised anyway when the except code is run again, but it adds an extra run with the same outcome so just checking)

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_connection.py
```diff
@@ -188,13 +188,13 @@ def _connect(self):
             else:
                 self._set_state(ConnectionState.HDR_SENT)
         except (OSError, IOError, SSLError, socket.error) as exc:
+            if isinstance(exc, FileNotFoundError) and exc.filename and "ca_certs" in exc.filename:
```
swathipil, line 191: added for parity with uamqp. Checking specifically for "ca_certs", b/c we're not sure if other cert/key files would raise the same error but be handled differently + not sure how to test that.

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_transport.py
```diff
@@ -551,7 +552,11 @@ def _wrap_socket_sni(  # pylint: disable=no-self-use
         }
 
         # TODO: We need to refactor this.
-        sock = ssl.wrap_socket(**opts)  # pylint: disable=deprecated-method
+        try:
+            sock = ssl.wrap_socket(**opts)  # pylint: disable=deprecated-method
+        except FileNotFoundError as exc:
+            exc.filename = {"ca_certs": ca_certs}
```
swathipil, line 558: adding this line so we can raise FileNotFoundError for the connection_verify cert file, specifically. Don't want to necessarily surface this type of error for other files that may also raise a FileNotFoundError.

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_transport.py
```diff
@@ -696,6 +706,24 @@ def connect(self):
                 http_proxy_port=http_proxy_port,
                 http_proxy_auth=http_proxy_auth,
             )
+        except WebSocketAddressException as exc:
+            raise AuthenticationException(
+                ErrorCondition.SocketError, # TODO: ClientError?
+                description="Failed to authenticate the connection due to exception: " + str(exc),
+                error=exc,
+            )
+        except (WebSocketTimeoutException, SSLError, WebSocketConnectionClosedException) as exc:    # type: ignore
```
swathipil, line 715: ignoring the type, b/c this line was causing a pylance error for some reason. (Something like `Tuple(Unbound | WebSocketTimeoutExcept ...)`). Needs more investigation later.

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_transport_async.py
```diff
@@ -391,7 +401,18 @@ async def close(self):
...
+            # reach the AMQP broker if the program exits after
+            # calling this method.
+            try:
+                self.sock.shutdown(socket.SHUT_RDWR)
+            except Exception as exc:  # pylint: disable=broad-except
+                # TODO: shutdown could raise OSError, Transport endpoint is not connected if the endpoint is already
+                #  disconnected. can we safely ignore the errors since the close operation is initiated by us.
+                _LOGGER.info("Transport endpoint is already disconnected: %r", exc)
+            self.sock.close()
+            self.sock = None
```
swathipil, line 415: copied from sync

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_transport_async.py
```diff
@@ -442,7 +463,7 @@ def __init__(
     ):
         self._read_buffer = BytesIO()
         self.socket_lock = asyncio.Lock()
-        self.sslopts = self._build_ssl_opts(ssl_opts) if isinstance(ssl_opts, dict) else None
+        self.sslopts = ssl_opts if isinstance(ssl_opts, dict) else None
```
swathipil, line 466: building here causes the error to surface from client.open() rather than client.connect(), which causes error handling discrepancy b/w pyamqp and uamqp. updated for parity.

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_transport_async.py
swathipil: need to remove TODO comment

sdk/eventhub/azure-eventhub/azure/eventhub/_transport/_pyamqp_transport.py
```diff
@@ -266,22 +269,19 @@ def send_messages(producer, timeout_time, last_exception, logger):
...
+            producer._handler.send_message(
+                producer._unsent_events[0], timeout=timeout
+            )
+            # TODO: The unsent_events list will always be <= 1. Even for a batch, it gets the underlying singular BatchMessage.
+            # May want to refactor in the future so that this isn't a list.
+            producer._unsent_events = None
+        except TimeoutError as exc:
+            raise OperationTimeoutError(message=str(exc), details=exc)
+        #except Exception as exc:
+        #   raise producer._handle_exception(exc)
```
swathipil, line 284: pretty sure this should be removed, since the handle_exception is called in the producer.send method when retries are exhausted.

sdk/eventhub/azure-eventhub/azure/eventhub/_producer_client.py
kashifkhan: it would be raised again if the send operation times out again, but for example if whatever cause the timeout is now gone and you can have a successful send and we decide to raise we  could lose out on the opportunity for a send?

will it hit the retry if its raised instead of going through the except block?

sdk/eventhub/azure-eventhub/azure/eventhub/_transport/_pyamqp_transport.py
```diff
@@ -461,6 +461,24 @@ def create_mgmt_client(
...
+        """
+        try:
+            mgmt_client.open(connection=conn)
+        except FileNotFoundError as exc:
+            # TODO: added for uamqp error parity with invalid connection_verify path
+            # consumer raises FileNotFoundError, producer raises EventHubError
+            # consumer opens connection through mgmt client first, producer through SendClient
+            # need to remove in the future
+            if exc.filename:
+                exc.filename2 = "consumer"
```
swathipil, line 479: only consumer raises FileNotFoundError for invalid connection_verify scenario, and encounters this error when opening mgmt client. producer raises EventHubError, and encounters the error when opening the send client. adding to distinguish and raise the correct error.

sdk/eventhub/azure-eventhub/azure/eventhub/_transport/_pyamqp_transport.py
```diff
@@ -532,8 +550,11 @@ def _create_eventhub_exception(exception):
         if isinstance(exception, errors.AuthenticationException):
             error = AuthenticationError(str(exception), exception)
         elif isinstance(exception, errors.AMQPLinkError):
-            error = ConnectError(str(exception), exception)
-        # TODO: do we need MessageHanlderError in amqp any more
+            if exception.condition in PyamqpTransport.ERROR_CONDITIONS:
+                error = ConnectionLostError(str(exception), exception)
+            else:
+                error = ConnectError(str(exception), exception)
```
swathipil, line 556: exception parity with uampq

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_connection.py
annatisch: Could you add that as a comment into the code?

annatisch: Essentially any time you want to add a comment to a PR to explain something - that should probably be documented in the code itself.

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_transport.py
annatisch: But it doesn't look like we're check that this _was_ raised because of the connection_verify cert file?

swathipil: so this FileNotFoundError doesn't have the name of the file that was not found stored in any of the attributes in the error, except the error message. So, unless I parse the error message, I'm not able to check against the cert file.

However, from the client side, this is the only file that the user can specify/pass through the kwargs. So, I assumed that the cert file would be the issue here. Maybe a bad assumption though?

annatisch: It's probably a reasonable assumption... Do we allow kwargs passthrough to the transport? I don't think we do, so I don't think we'd ever have a situation was that `ssl_opts` could include some filepath we don't know about...

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_transport.py
annatisch: I was looking at uamqp and we never had this socketerror condition.... so I'm not really sure why it was introduced.
I'm confused as the link above is also pyamqp, not uamqp - or was that not the intended comparison?

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_transport.py
```diff
@@ -696,6 +706,24 @@ def connect(self):
...
+            self.close()
+            if isinstance(exc, WebSocketTimeoutException):
+                message = f'send timed out ({str(exc)})' 
+            elif isinstance(exc, SSLError):
+                message = f'send disconnected by SSL ({str(exc)})'
```
annatisch, line 720: Minor nitpick: I hate to be that person... but can we start sentences with capital letters.... >_<

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_transport.py
swathipil: ah sorry, I meant before as in a previous version of pyamqp. Previously, it was caught in the _connection.py pyamqp file I believe. In uamqp, we didn't have an error condition at all as you mentioned.

sdk/eventhub/azure-eventhub/azure/eventhub/_transport/_pyamqp_transport.py
annatisch: It looks like we're using ConnectionLostError for effectively all the protocol error conditions - what are the conditions we would expect to raise ConnectError?

sdk/eventhub/azure-eventhub/azure/eventhub/_transport/_pyamqp_transport.py
```diff
@@ -542,7 +563,11 @@ def _create_eventhub_exception(exception):
         elif isinstance(exception, TimeoutError):
             error = ConnectionLostError(str(exception), exception)
         else:
-            error = EventHubError(str(exception), exception)
+            if isinstance(exception, FileNotFoundError) and exception.filename2 == "consumer":
```
annatisch, line 566: Rather than put this attribute on the error - can we not simply pass an optional `is_consumer` flag into the method signature?

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_transport.py
swathipil: No, we don't pass kwargs through.

connection_verify path is [passed into the SendClient](https://github.com/Azure/azure-sdk-for-python/blob/5013c2d8464a518ff4f2a6ca691f7913e1620c6e/sdk/eventhub/azure-eventhub/azure/eventhub/_transport/_pyamqp_transport.py#L253), and we set it to an attribute on the client. We then pass that through to the transport constructor [[here](https://github.com/Azure/azure-sdk-for-python/blob/5013c2d8464a518ff4f2a6ca691f7913e1620c6e/sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/client.py#L291)].

annatisch: Okay - though once again could we make sure things like this are _thoroughly_ documented with inline comments?

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_transport_async.py
annatisch: I don't know if this is necessary? We pass the socket into `open_connection` - doesn't that close it on our behalf when we close the reader and writer?

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_transport.py
swathipil: yes, adding this now!

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_transport_async.py
swathipil: comment added to code.

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/_transport.py
swathipil: TODO comment added here to address later.

sdk/eventhub/azure-eventhub/azure/eventhub/_transport/_pyamqp_transport.py
swathipil: We raise the ConnectError when the error condition starts with "com." (aka. the "VendorLinkDetach" from uamqp). 

https://github.com/Azure/azure-sdk-for-python/blob/b2d5a9a2d0d70d400da562ecef5e5dafa14cc729/sdk/eventhub/azure-eventhub/azure/eventhub/_transport/_uamqp_transport.py#L575

Adding this comment to the code here.

sdk/eventhub/azure-eventhub/azure/eventhub/_transport/_pyamqp_transport.py
swathipil: removing this and passing in a `is_consumer` kwarg from mgmt_request method in client bases. Adding a comment in the code there to clarify.

sdk/eventhub/azure-eventhub/azure/eventhub/_transport/_pyamqp_transport.py
swathipil: updated + adding a comment for posterity

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_transport_async.py
swathipil: ah, I see. That's prob why this wasn't here before 🙂. Will add a note here.

annatisch (approved):

PR #27093, opened 2022-10-26
[Storage] Remove `HandleItem` from export and fix handle related docs

jalauzon-msft:
The generated model `HandleItem` is currently being exported despite us having our own hand-written model, `Handle`, to replace it. This change removes `HandleItem` from the export list. There were also some docs for `list_handles` and `close_handle` that referenced `HandleItem` despite actually using `Handle` so this change updates those docs as well.

As part of this change, I verified that `HandleItem` was not being used as an input or output to any API (desipte the incorrect documentation) and therefore should be safe to remove as an export.


vincenttran-msft (approved):
Thanks for catching this! Hopefully should clean up those API Views 😄

PR #27097, opened 2022-10-27
[formrecognizer] Skip bad model ID copy test

catalinaperalta:
Currently this test case is returning an InternalServerError, will follow up with the service team.

Tracking https://github.com/Azure/azure-sdk-for-python/issues/27099


sdk/formrecognizer/azure-ai-formrecognizer/tests/test_dmac_copy_model.py
```diff
@@ -91,7 +91,7 @@ def test_copy_model_with_model_id_and_desc(self, client, formrecognizer_storage_
                 assert field["type"]
                 assert doc_type.field_confidence[key] is not None
 
-    @skip_flaky_test
+    @pytest.mark.skip()
```
catalinaperalta, line 94: This should have been caught with the raises statement, but something changed, so will have to investigate. Skipping to unblock pipelines from this error.

kristapratico (approved):

PR #27123, opened 2022-10-27
App configuration provider geo replication

mrm9084:
# Description
Adds Geo Replication support, currently in Beta in the App Configuration Service. Only Supported for endpoint.

Added new supported property `endpoints`, which takes in a list of endpoints and will attempt them in order if one fails to load configurations.


sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/_azureappconfigurationprovider.py
```diff
@@ -27,9 +31,11 @@ def __init__(self):
         self._dict = {}
         self._trim_prefixes = []
         self._client = None
+        self._clients = []
 
     @classmethod
-    def load(cls, connection_string=None, endpoint=None, credential=None, **kwargs):
+    def load(cls, connection_string=None, endpoint=None, endpoints=None, credential=None,
```
xiangyan99, line 37: This is breaking change. By design?

mrm9084:
@avanigupta

sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/_azureappconfigurationprovider.py
mrm9084: Is it breaking as it has a default value, and is a new configuration? Also, yes. Also at some point connection-strings will be added as an option. This could be done through kwargs, but as a primary connection method it seemed correct to be here.

xiangyan99: Code:
AzureAppConfigProvider.load("conn_str", "end_point","credential")

old behavior:
connection_string = "conn_str"
endpoint = "end_point"
credential = "credential"

New behavior:
connection_string = "conn_str"
endpoint = "end_point"
endpoints = "credential"
credential = None

xiangyan99: Btw, given you are still in beta, it is totally fine to have breaking change. 

Just wanted to make sure you are aware.

mrm9084: I always do:

`AzureAppConfigProvider.load(connection_str="")`

Didn't think of that. As endpoint and connection string are never set at the same time.

mrm9084: Also, endpoint and endpoints are never set at the same time.

xiangyan99: Maybe we can make them key word only arguments?

mrm9084: I guess the question would be what wouldn't be a keyword argument at that point?

xiangyan99: Given all of them are optional, maybe we can use 

def load(cls, *, connection_string=None, endpoint=None, endpoints=None, credential=None, **kwargs)

to make all of them keyword only.

xiangyan99: And if you want, you can even use function overloading to make it clearer to use. e.g. https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/core/azure-core/azure/core/pipeline/transport/_requests_basic.py#L270

mrm9084: The tricky thing with overloading is `connection_string` and `endpoint` are both Strings, and `connection_strings` and `endpoints` are both lists of Strings. And only one is ever set. We would have to do some weird validation of endpoint vs connection_string which I don't like.

xiangyan99: If we make them keyword only, we can get rid of the ambiguity, right?

mrm9084: Well then it would just be:

`def load(cls, **kwargs):`

so no different than:

`def load(cls, *, connection_string=None, endpoint=None, endpoints=None, credential=None, **kwargs):`

xiangyan99: Sorry, not quite clear about "no different than"?

mrm9084: What I mean is that from a users perspective there is no difference between the two interfaces for a user.

xiangyan99: If they use IDE (pycharm, vscode, etc.), def load(cls, *, connection_string=None, endpoint=None, endpoints=None, credential=None, **kwargs): shows intellisense.

mrm9084: I have changed it to be the second one, so it will have the intellisense.

sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/_azureappconfigurationprovider.py
```diff
@@ -50,44 +56,98 @@ def load(cls, connection_string=None, endpoint=None, credential=None, **kwargs):
...
+        if endpoints:
+            for endpoint_geo in endpoints:
+                provider._clients.append([endpoint_geo, provider.__build_provider(None, endpoint_geo, credential,
+                    key_vault_options)])
+        else:
+            store_endpoint = endpoint if endpoint is not None else provider.__parse_connection_string(connection_string)
+            provider._clients.append([store_endpoint, provider.__build_provider(connection_string, endpoint,
+                credential, key_vault_options)])
 
         selects = kwargs.pop("selects", {SettingSelector("*", "\0")})
```
albertofori, line 74: Not related to this pr but it seems the keyword for "selects" has been specified as "selectors" on line 48. Correct me if I'm wrong about this.

sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/_azureappconfigurationprovider.py
```diff
@@ -50,44 +56,98 @@ def load(cls, connection_string=None, endpoint=None, credential=None, **kwargs):
...
+            try:
+                provider.__load_selected_configurations(client, secret_clients, selects, key_vault_options)
+                # If we get here, we have successfully loaded the configuration settings from the client.
+                # We don't need to check the other clients.
+                break
+
+            except ServiceRequestError as e:
+                print("Failed to load configuration settings from " + client[0] + " from Service Request Failure {}."
+                    .format(e))
+                if len(provider._clients) == 1 or len(provider._clients) == client_count:
```
albertofori, line 94: Could there be a case where client_count will not be equal to the number of clients when the number of clients is 1?

sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/_azureappconfigurationprovider.py
```diff
@@ -50,44 +56,98 @@ def load(cls, connection_string=None, endpoint=None, credential=None, **kwargs):
...
+                    len(provider._clients) > 1) or len(provider._clients) == client_count:
+                    raise e
+
+                print("Trying next client.")
+
+
+        return provider
+
+    def __load_selected_configurations(self, client, secret_clients, selects, key_vault_options):
+        # type: (list, dict, list[SettingSelector], KeyVaultOptions) -> None
```
albertofori, line 110: Is `client` a list?

sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/_azureappconfigurationprovider.py
mrm9084: Good catch. I have a different PR to update few small things. It will go into that.

sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/_azureappconfigurationprovider.py
mrm9084: The answer to your first question is No. I am trying to think of why I wrote it like this, but it might have been hedging a bet about an "optional" implementation being added. Though now I am thinking they should be different errors. One is you gave us one client we tried and failed. The other is you gave us a geo replicated store and all failed. @avanigupta thoughts?

PR #27127, opened 2022-10-27
[formrecognizer] Adjust API version errors

catalinaperalta:
Some test updates due to changes in Python 3.11

### DIFF
--- a/sdk/formrecognizer/azure-ai-formrecognizer/azure/ai/formrecognizer/_api_versions.py
+++ b/sdk/formrecognizer/azure-ai-formrecognizer/azure/ai/formrecognizer/_api_versions.py
@@ -27,13 +27,13 @@ def validate_api_version(api_version: str, client_kind: str) -> None:
 
     if client_kind == "form":
         try:
-            api_version = FormRecognizerApiVersion(api_version)
+            _ = FormRecognizerApiVersion(api_version)
         except ValueError:
             err_message = "Unsupported API version '{}'. Please select from: {}".format(
                 api_version, ", ".join(v.value for v in FormRecognizerApiVersion)
             )
             try:
-                api_version = DocumentAnalysisApiVersion(api_version)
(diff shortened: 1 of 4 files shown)


sdk/formrecognizer/azure-ai-formrecognizer/azure/ai/formrecognizer/_api_versions.py
```diff
@@ -27,13 +27,13 @@ def validate_api_version(api_version: str, client_kind: str) -> None:
 
     if client_kind == "form":
         try:
-            api_version = FormRecognizerApiVersion(api_version)
```
catalinaperalta, line 30: Doing this to not modify the original API version string that is appended to the error message. This will keep the string untouched in the error message.

sdk/formrecognizer/azure-ai-formrecognizer/azure/ai/formrecognizer/_api_versions.py
```diff
@@ -27,13 +27,13 @@ def validate_api_version(api_version: str, client_kind: str) -> None:
...
+            _ = FormRecognizerApiVersion(api_version)
```
kristapratico, line 30: Do we even need a throwaway variable?

kristapratico (approved):
I'm still curious about this change. Did you see anything in the Python 3.11 notes that would explain it?

sdk/formrecognizer/azure-ai-formrecognizer/azure/ai/formrecognizer/_api_versions.py
catalinaperalta: You're right, just added the update :)

catalinaperalta:
> I'm still curious about this change. Did you see anything in the Python 3.11 notes that would explain it?

Still looking for the specific change that caused this. There were several updates to enum in the 3.11 release.

Issue #26509, opened 2022-09-28
Add precommit hooks to repo


needuv:
**Is your feature request related to a problem? Please describe.**

We would like to have pre-commit hooks in azure-sdk-for-python that can automate fixing common linting issues. These hooks can be things such as removing trailing whitespaces for lines, making sure files have an extra line at the end of the file, running `black` to format code, running `pycln` to clean up unused imports, running `isort` to sort imports, and many other things.

**Describe the solution you'd like**

The ML SDK team was using a precommit configuration that looked like this: 

```yaml

# See https://pre-commit.com for more information
# See https://pre-commit.com/hooks.html for more hooks

exclude: (^docs/)|(^tests/test_configs/)|(^autorest/)|(^src/cli/src/machinelearningservices/(README.md|setup.py|azext_mlv2/manual/__init__.py))|(^src/azure-ai-ml/azure/ai/ml/_restclient/)|(^src/azure-ai-ml/azure/ai/ml/_vendor/)

repos:
    - repo: https://github.com/pre-commit/pre-commit-hooks
      rev: v3.2.0
      hooks:
          - id: trailing-whitespace
          - id: end-of-file-fixer
          - id: check-yaml
            exclude: src/azure-ai-ml/tests/test_configs/components/invalid
          # -   id: check-added-large-files
          - id: check-json
          - id: check-merge-conflict
    - repo: https://github.com/psf/black
      rev: 22.3.0 # Replace by any tag/version: https://github.com/psf/black/tags
      hooks:
          - id: black
            language_version: python3 # Should be a command that runs python3.6+
            args:
                - "--line-length=120"
    - repo: https://github.com/pre-commit/pre-commit-hooks
      rev: v2.3.0
      hooks:
          - id: flake8
    - repo: https://github.com/hadialqattan/pycln
      rev: v0.0.2 # Possible releases: https://github.com/hadialqattan/pycln/tags
      hooks:
          - id: pycln
            name: "Clean unused python imports"
            args: [--config=setup.cfg]
            exclude: src/azure-ai-ml/tests/test_configs
    - repo: https://github.com/pycqa/isort
      rev: 5.8.0
      hooks:
        - id: isort
          # stages: [commit]
          name: isort-python
          # Use black profile for isort to avoid conflicts
          # see https://github.com/PyCQA/isort/issues/1518
          args: ["--profile", "black", --line-length=120]

```

[...]

PR #27162, opened 2022-10-28
git: add pre-commit configuration to .gitignore

elliotzh:
# Description

Add pre-commit configuration file to .gitignore.

We don't want to enable git hooks in root level as package specific hooks may bring delay and in some cases error to all developers.

By adding the configuration file to .gitignore, package developers may enable pre-commit locally if needed without impact the repo.


jsquire:
//cc: @lmazuel, @weshaggard:   Please offer your thoughts, as this would impact pipelines for all of Python.

jsquire:
@elliotzh :  Please do not merge this without explicit sign-off from Laurent or Wes.

weshaggard (changes requested):
For our mono repositories we avoid having any commit hooks because they often cause reliability issues, and they cannot be depended on because it requires that everyone that clones the repo do a setup step to enable them. Instead, we suggest adding checks in the CI pipeline for whatever you are trying to enforce.

luigiw:
@weshaggard Thx for looking into this. We already have the pylint check in ML's CI pipeline. The problem is that the pylint error in CI pipeline show up too late, our engineers often find themselves waiting for CI pipeline for ~20 minutes just to find out a simple pylint error. This impacts our engineering productivity.  Is there a way to add pre commit at sub folder level?

weshaggard:
>CI pipeline show up too late, our engineers often find themselves waiting for CI pipeline for ~20 minutes just to find out a simple pylint error. This impacts our engineering productivity. Is there a way to add pre commit at sub folder level?

There is not a way to configure a git commit to a given directory only as it is a global repo hook and will intercept all git commit calls if configured locally. There is also no way to enforce that the git commit hook is enabled on a local clone. 

CI should only be the enforcer so folks shouldn't ever hit it if they are correctly running the pylint commands locally before they submit a PR. It should simply be part of the workflow folks use to do development locally and CI is only there to enforce they did that work.

kristapratico:
@luigiw take a look at using `tox` for running pylint locally: https://github.com/Azure/azure-sdk-for-python/blob/main/doc/dev/tests.md#tox

The `lint` environment command is identical to what's run in CI.

elliotzh:
@weshaggard @kristapratico Thank you for your explanation! Then what if we add the configuration file to .gitignore? Then we may enable pre-commit locally without impacting other developers. And we may maintain related files & docs inside azure-ai-ml.

cc @luigiw @wangchao1230 

PS: using `tox` to run pylint locally is great, I just usually forget to do it before committing and it will take longer than just running pylint on changed files. Actually, current pre-commit required the developer to run pylint with `tox` first to set up the environment.

needuv:
@elliotzh @luigiw @jsquire @weshaggard @kristapratico 

During the ML SDK migration, we identified not having pre-commit hooks as one of our process gaps-- we had it in our internal repo to do automated code quality checks. @scbedd suggested I create an issue for this so that the Azure SDK team could look into this, which I did back in September: https://github.com/Azure/azure-sdk-for-python/issues/26509. Something that we discussed was enabling it **just** for the ML SDK, and other teams could opt in if they wanted to (but they are not forced to). In addition to the pylint check, we would like to have things like black, new line end of file checker, flake8, isort, etc run on our commits, and I've included the pre-commit file we had in our internal repo in the issue. Let me know your thoughts?

weshaggard:
I will say from our past experiences that doing this as part of a git hook is likely going to cause more issues than you might want (i.e. it causes performance and reliability issues during git operations that can leave the repo in odd states). There is also still the problem that folks have to remember to do something post-cloning of the repo in order to get these hooks setup otherwise they will also not be ran. 

If you are worried about folks forgetting to run locally before submitting a PR I'd recommend hooking it up to a command folks already run all the time like running tests or perhaps a wrapper script that does tests and these other things you want to be sure folks don't forget. Also depending on things like .editorconfig files will help folks do a number of these linting and formatting things directly while editing which is a more natural way to get these integrated.

weshaggard:
After talking I'm OK with adding this to the gitignore so folks can opt into this if wish. As we talked we should ensure any commands ran as part of the git hook should be consistent with what we run in CI.

wangchao1230 (approved):

PR #27173, opened 2022-10-28
Resolve 3.11 Failures for `storage`

scbedd:
I'm expecting to have to assemble a `yarl` and `multidict` wheel.

Confirming [storage build](https://dev.azure.com/azure-sdk/public/_build/results?buildId=1952847&view=logs&j=84b20eb0-323b-5722-f041-399ea9857d18&t=1024ff5b-59a4-5548-1c1b-bbb298ca1664).

### DIFF
--- a/eng/test_tools.txt
+++ b/eng/test_tools.txt
@@ -1,8 +1,5 @@
 cryptography<4
 
-# python 3.10 pinned packages
-cffi==1.15.0rc2; python_version >= '3.10'
-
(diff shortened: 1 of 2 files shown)


scbedd:
@xiangyan99 both of those core failures are because we're expecting bing to redirect us and get a 301, but bing isn't redirecting anymore, so it's 200.

My change has nothing to do with that 👍

xiangyan99 (approved):

PR #27229, opened 2022-11-01
Graphrbac is deprecated

lmazuel:
Make a clear message on PyPI


lmazuel:
Force merge, as minDep is failing for weird reasons and there is no code change here

akx:
> Make a clear message on PyPI

It looks like this has never actually been updated on PyPI though? 😄 https://pypi.org/project/azure-graphrbac/

PR #27260, opened 2022-11-02
[Pyamqp] Changes for blocking exceptions

kashifkhan:
This PR is to potentially fix hangs when a user defined callback raises an error.

* Changes to how we close a socket connection. There is a bug where `streamwrite.await_closed` can intermittently hang.  I added in `asyncio.sleep(0)` so that the buffer is written out, then destroy the ssl object and then await_closed. It matches what trio and http core have done
* Includes the hanging fix for receive_task and callback_task which would address the missing CancelledError and hanging unit test
* I also removed some of the ensure_future calls that were being made. Them randomly firing off in the back and not being handled was causing more stuff running post close. We can add these back once we design and think about how we want to handle them failing, exceptions etc to properly utilize their function

### DIFF
--- a/sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_connection_async.py
+++ b/sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_connection_async.py
@@ -751,12 +751,17 @@ class Connection(object):  # pylint:disable=too-many-instance-attributes
                 ):
                     # TODO: compare the perf difference between ensure_future and direct await
                     break
+                
         except (OSError, IOError, SSLError, socket.error) as exc:
             self._error = AMQPConnectionError(
                 ErrorCondition.SocketError,
                 description="Can not send frame out due to exception: " + str(exc),
                 error=exc,
             )
+        except asyncio.CancelledError:
+            raise
+        except Exception:
+            raise
 
     def create_session(self, **kwargs):
         # type: (Any) -> Session
--- a/sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_connection_async.py
+++ b/sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_connection_async.py
@@ -751,7 +751,6 @@ class Connection(object):  # pylint:disable=too-many-instance-attributes
                 ):
                     # TODO: compare the perf difference between ensure_future and direct await
                     break
-                
         except (OSError, IOError, SSLError, socket.error) as exc:
             self._error = AMQPConnectionError(
                 ErrorCondition.SocketError,
@@ -760,7 +759,7 @@ class Connection(object):  # pylint:disable=too-many-instance-attributes
             )
         except asyncio.CancelledError:
             raise
-        except Exception:
+        except Exception: # pylint: disable=try-except-raise
             raise
 
     def create_session(self, **kwargs):
--- a/sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_connection_async.py
+++ b/sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_connection_async.py
@@ -757,10 +757,6 @@ class Connection(object):  # pylint:disable=too-many-instance-attributes
                 description="Can not send frame out due to exception: " + str(exc),
                 error=exc,
             )
-        except asyncio.CancelledError:
-            raise
-        except Exception: # pylint: disable=try-except-raise
-            raise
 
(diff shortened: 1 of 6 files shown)


sdk/eventhub/azure-eventhub/azure/eventhub/aio/_transport/_pyamqp_transport_async.py
```diff
@@ -214,16 +214,20 @@ async def receive_messages_async(consumer, batch, max_batch_size, max_wait_time)
...
+        callback_task = asyncio.create_task(
             PyamqpTransportAsync._callback_task(consumer, batch, max_batch_size, max_wait_time)
         )
-        receive_task = asyncio.ensure_future(PyamqpTransportAsync._receive_task(consumer))
+        receive_task = asyncio.create_task(PyamqpTransportAsync._receive_task(consumer))
 
-        try:
-            await receive_task
-        finally:
-            consumer._callback_task_run = False
```
annatisch, line 225: What was this check for?

kashifkhan: I think its used to break the callback [task](https://github.com/Azure/azure-sdk-for-python/blob/5013c2d8464a518ff4f2a6ca691f7913e1620c6e/sdk/eventhub/azure-eventhub/azure/eventhub/aio/_transport/_pyamqp_transport_async.py#L151) out from the infinite while loop so it could be awaited.

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_session_async.py
```diff
@@ -246,37 +212,22 @@ async def _outgoing_flow(self, frame=None):
...
-                if (
-                    self.remote_incoming_window > 0
-                    and not link._is_closed  # pylint: disable=protected-access
-                ):
-                    futures.append(
-                        link._incoming_flow(frame)  # pylint: disable=protected-access
-                    )
+                if self.remote_incoming_window > 0 and not link._is_closed:  # pylint: disable=protected-access
+                    futures.append(link._incoming_flow(frame))  # pylint: disable=protected-access
             await asyncio.gather(*futures)
```
annatisch, line 231: We're keeping the `asyncio.gather` pattern here?

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_transport/_pyamqp_transport_async.py
```diff
@@ -214,16 +214,26 @@ async def receive_messages_async(consumer, batch, max_batch_size, max_wait_time)
...
+                except Exception: # pylint: disable=broad-except
+                    consumer._callback_task_run = False
+            for task in tasks:
+                if task.done() and task.exception():
+                    raise task.exception()
+        except asyncio.CancelledError:
             consumer._callback_task_run = False
-            await callback_task
+            await asyncio.sleep(0)
+            raise
```
annatisch, line 236: I think this can still cause asyncio warnings about cancelled pending tasks or task exceptions not retrieved.
I will use this patch in my testing this afternoon to see if I still can repro that in combination with your other fixes.

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_session_async.py
kashifkhan: No , this needs to go as well

sdk/eventhub/azure-eventhub/azure/eventhub/aio/_transport/_pyamqp_transport_async.py
kashifkhan: yeah this was mostly too see if there were some other exceptions floating around, then would the main loop catch them.

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_connection_async.py
```diff
@@ -757,6 +757,10 @@ async def listen(self, wait=False, batch=1, **kwargs):
                 description="Can not send frame out due to exception: " + str(exc),
                 error=exc,
             )
+        except asyncio.CancelledError:
+            raise
+        except Exception: # pylint: disable=try-except-raise
+            raise
```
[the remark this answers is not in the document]
kashifkhan: yes will get rid of these. Left over from the initial search

annatisch (approved):

PR #27282, opened 2022-11-03
[TA,DT] api_version enum update

kristapratico:


mccoyp (approved):
LGTM! Seems like CI auto-triggering might be stuck again...

PR #27293, opened 2022-11-04
[Storage] Fix flaky abort copy blob test

jalauzon-msft:
This change aims to eliminate the flakiness of `test_abort_copy_blob` in the live test pipeline. This test initiates a copy blob and tries to abort it but occasionally the copy finishes before the abort request is made. This change adds a try/except around the end of the test for this case and asserts on the `NoPendingCopyOperation` error code is raised. This should prevent the test from failing in the case that the copy finishes early.

While this sort of defeats the purpose of the test in the case where the copy finishes early, I think it's better than marking this as playback only since the failure case is rare so most of the time the abort is being tested.


vincenttran-msft (approved):
LGTM! Be gone flaky tests! 👍

PR #27304, opened 2022-11-04
[DT] update python version for INT tests

kristapratico:
These tests just validate service behavior. We don't need to run with more than one version of Python (latest is fine).

### DIFF
--- a/sdk/translation/tests.yml
+++ b/sdk/translation/tests.yml
@@ -16,14 +16,8 @@ stages:
           # We run tests in Dogfood for the service team. This cloud config is just used to validate the service deployment
           # so we don't need to run on every version of Python.
           SubscriptionConfiguration: $(sub-config-translation-int-test-resources)
-          # TODO fix this to specify which version to run rather than filtering out versions.
           MatrixFilters:
-            - PythonVersion=^(?!pypy3).*
-            - PythonVersion=^(?!3.6).*
-            - PythonVersion=^(?!3.7).*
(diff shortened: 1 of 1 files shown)


sdk/translation/tests.yml
```diff
@@ -23,6 +23,7 @@ stages:
             - PythonVersion=^(?!3.7).*
             - PythonVersion=^(?!3.8).*
             - PythonVersion=^(?!3.9).*
+            - PythonVersion=^(?!3.10).*
```
benbp, line 26: Is this a good opportunity to address the above TODO and switch to specifying the version TO run?

kristapratico: Sure! Do you know if there are docs that explain how to do that?

benbp: The filters are just a regex ([docs](https://github.com/Azure/azure-sdk-tools/blob/main/eng/common/scripts/job-matrix/README.md#filters)). In this case, I think you could delete all the above negative regex items in favor of only including `3.11`. Either way this filter will need to be updated when the version is bumped.

kristapratico: Thanks, those docs are great, bookmarked! I'll update to your suggestion.

kristapratico: @benbp is this good to merge?

benbp (approved):

benbp:
@kristapratico LGTM

PR #27310, opened 2022-11-04
[PyAMQP] Fix buff prod async not closing down on stress tests

l0lawrence:


sdk/eventhub/azure-eventhub/azure/eventhub/aio/_buffered_producer/_buffered_producer_dispatcher_async.py
```diff
@@ -3,7 +3,7 @@
 # Licensed under the MIT License. See License.txt in the project root for license information.
 # --------------------------------------------------------------------------------------------
 from __future__ import annotations
-import asyncio
+import asyncio # pylint: disable=unused-import
```
swathipil, line 6: can we remove this if it's not being used?

annatisch (commented):
Changes look good. We probably should be doing the flush concurrently (maybe something like a shielded gather), however if this has been tested on 32 partitions and the flush delay isn't a problem then we can table that for a future improvement.

PR #27337, opened 2022-11-07
[ML][Endpoints] Ensure defaults property is instantiated

qusayalsaifi:
Our document referenced below, states that "defaults" dictionary defaults to {}, however, it is current defaults to None.

Reference: https://learn.microsoft.com/en-us/python/api/azure-ai-ml/azure.ai.ml.entities.batchendpoint?view=azure-python

# Description

### DIFF
--- a/sdk/ml/azure-ai-ml/azure/ai/ml/entities/_endpoint/batch_endpoint.py
+++ b/sdk/ml/azure-ai-ml/azure/ai/ml/entities/_endpoint/batch_endpoint.py
@@ -54,7 +54,7 @@ class BatchEndpoint(Endpoint):
         auth_mode: str = AAD_TOKEN_YAML,
         description: str = None,
         location: str = None,
-        defaults: Dict[str, str] = None,
+        defaults: Dict[str, str] = {},
         default_deployment_name: str = None,
         scoring_uri: str = None,
         openapi_uri: str = None,
--- a/sdk/ml/azure-ai-ml/CHANGELOG.md
+++ b/sdk/ml/azure-ai-ml/CHANGELOG.md
@@ -18,6 +18,8 @@
 - Make registry delete operation return an LROPoller, and change name to begin_delete.
 - Prevent registering an already existing environment that references conda file.
 - Fix ARM id logic for registry environments (ex: Creating a registry component that references a registry environment).
+- Update "defaults" dictionary to match our public documentation.
+
 
 ### Other Changes
 


sdk/ml/azure-ai-ml/azure/ai/ml/entities/_endpoint/batch_endpoint.py
```diff
@@ -54,7 +54,7 @@ def __init__(
         auth_mode: str = AAD_TOKEN_YAML,
         description: str = None,
         location: str = None,
-        defaults: Dict[str, str] = None,
+        defaults: Dict[str, str] = {},
```
Man-MSFT, line 57: Default values for mutable types should always be `None`, as their value is instantiated only once.

Man-MSFT (changes requested):

sdk/ml/azure-ai-ml/azure/ai/ml/entities/_endpoint/batch_endpoint.py
Man-MSFT: I'm not fully understanding the problem tho, our public doc will look at our type hints, here is `Dict[str,str]`, which is what we expect the caller to pass. 

Is the problem that, user can not pass `defaults` and also not `default_deployment` which makes `self.defaults == None`? Then we can break the logic s.t. `self.defaults` is always at least `{}`.

qusayalsaifi:
Hello Man,

Our main concern is that Doc<https://learn.microsoft.com/en-us/python/api/azure-ai-ml/azure.ai.ml.entities.batchendpoint?view=azure-python> saying that "defaults" should default to {}, however, it's defaulting to None.

How this is impacting?

In the following azureml-example:
azureml-examples/mlflow-for-batch-tabular.ipynb at main * Azure/azureml-examples (github.com)<https://github.com/Azure/azureml-examples/blob/main/sdk/python/endpoints/batch/mlflow-for-batch-tabular.ipynb>

In 3.1 Configure the endpoint, code is not setting "deployment_name" attribute, nor "defaults" dictionary.
Therefore, code is calling BatchEndpoint and setting None value to "defaults".

Later on, in 4.5 Create the deployment:

-    endpoint.defaults.deployment_name = deployment.name

This line of code is causing failure because defaults = None.

If we used the attribute deployment_name and set it in 3.1, this will force the class to create defaults dictionary, following to the code in batch_endpoint.py:
if not self.defaults and default_deployment_name:
            self.defaults = {}
            self.defaults["deployment_name"] = default_deployment_name

However, we cannot use "deployment_name" attribute at this stage of code, because we don't have the deployment created yet.

I found a stable workaround, by adding the following line in 3.1 of notebook:
endpoint = BatchEndpoint(
    name="heart-classifier-batch",
    description="A heart condition classifier for batch inference",
    defaults = {}
)

Here I'm creating the dictotionary of defaults manually, but the question is, should this happen automatically by calling BatchEndpoint?

Thee reason I opened this pull request, is because I saw the document saying that "defaults" should be {}.

Qusay Alsaifi
Technical Support Engineer
AI and Intelligent Devices
Office: +962 (6) 556 4115
***@***.******@***.***>

Mon - Fri  | 9 AM - 6 PM UTC+3

[Microsoft Logo]

qusayalsaifi:
Hello Man,

I have changed the name as requested:
[ML][Endpoints] Ensure defaults property is instantiated by qusayalsaifi * Pull Request #27337 * Azure/azure-sdk-for-python (github.com)<https://github.com/Azure/azure-sdk-for-python/pull/27337>


And yes, this is exactly what we need, batch endpoint should initiate defaults by default, so it can be usable later on.

Qusay Alsaifi
Technical Support Engineer
AI and Intelligent Devices
Office: +962 (6) 556 4115
***@***.******@***.***>

Mon - Fri  | 9 AM - 6 PM UTC+3

[Microsoft Logo]

sdk/ml/azure-ai-ml/azure/ai/ml/entities/_endpoint/batch_endpoint.py
Man-MSFT: @qusayalsaifi Still looking for a change here. We cannot make this change. 
Here's a simple example:
```
class Foo:
	def __init__(self, defaults = {}):
		self.defaults = defaults

a = Foo()
print(a.defaults) # prints {}
a.defaults['key1'] = 'value1'
print(a.defaults) # prints {'key1': 'value1'}

b = Foo()
print(b.defaults) # prints {'key1': 'value1'}
```

If you want `self.defaults` to be always instantiated, you must do it in the ctor itself and not as a default value.

hugoaponte: @qusayalsaifi : Could you share more about this issue? Do you have the user code so we can repro?

I am hesitant to change the default since it could be a breaking change, and also @Man-MSFT has a point.

But I would like to understand better the root cause to find the right solution.

qusayalsaifi:
@***@***.***> agree company="Microsoft"


Qusay Alsaifi
Support Engineer
Azure Machine Learning Developer Support

+962 (6) 556 4115
Working Hours: Sun - Thu 6AM - 3PM (UTC)
Manager: Daniel ***@***.***>

***@***.***

PR #27370, opened 2022-11-08
[Storage] Refactor tests in `test_avro_async.py`

vincenttran-msft:
This is the start of the work to help clear up some failures in the pipeline: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=1972571&view=ms.vss-test-web.build-test-results-tab

This fixes the tests in the tests/avro path

Additionally, this fixes a typo (`test_with_hearder_reader` --> `test_with_header_reader`)

### DIFF
--- a/sdk/storage/azure-storage-blob/tests/avro/test_avro_async.py
+++ b/sdk/storage/azure-storage-blob/tests/avro/test_avro_async.py
@@ -84,9 +84,8 @@ class AvroReaderTestsAsync(unittest.TestCase):
             correct,
             len(CODECS_TO_VALIDATE) * len(SCHEMAS_TO_VALIDATE))
 
-    def test_reader(self):
-        loop = asyncio.get_event_loop()
-        loop.run_until_complete(self._test_reader())
+    async def test_reader(self):
+        await self._test_reader()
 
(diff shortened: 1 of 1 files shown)


sdk/storage/azure-storage-blob/tests/avro/test_avro_async.py
```diff
@@ -84,9 +84,8 @@ async def _test_reader(self):
             correct,
             len(CODECS_TO_VALIDATE) * len(SCHEMAS_TO_VALIDATE))
 
-    def test_reader(self):
-        loop = asyncio.get_event_loop()
-        loop.run_until_complete(self._test_reader())
+    async def test_reader(self):
+        await self._test_reader()
```
jalauzon-msft, line 88: I would say get rid of these methods all together and just get rid of `_` in `_test_reader` to make that the test that is automatically run. No need for this helper method now :)

vincenttran-msft: Ah gotcha. I remember us chatting about this offline and was slightly confused but this clears it up 😄

jalauzon-msft (approved):
Thanks for jumping on this!

Issue #25213, opened 2022-07-14
Change test-resource.json to bicep


rakshith91:
I would be very happy if you decided to change this to a bicep template (`az bicep decompile -f test-resources.json  # outputs test-resources.bicep`) ;) In general it's a much simpler format that's easier to work with over time.

_Originally posted by @benbp in https://github.com/Azure/azure-sdk-for-python/pull/25195#discussion_r920692464_

PR #27400, opened 2022-11-09
[Monitor] Refactor tests

pvaneck:
This applies to both query and ingestion tests:

* Better integration with test proxy using fixtures.
* Updated `test-resources.json` file to be a bicep file. Aso cleaned out resources that were deployed but never used (i.e. App Config, Event Hubs)
* Removed dependency on preallocated resources for live tests. Now live tests will only use resources deployed from the `test-resource.bicep` file.
* Integrated tests with assets repo. Added playback tests and recordings into assets repo tag.
* Increased msrest version in setup.py as the mindependency test job was failing because the msrest version used was too old.

Closes: #25213


benbp:
Love the test resource changes!

pvaneck (commented):
Note: Pending update to use the assets repo for recordings before merging. This should substantially reduce the size of this PR.

mccoyp (approved):
Amazing! Thanks so much for your work on this 🙂

PR #27408, opened 2022-11-09
[Storage] Fix live test `test_create_service_with_token`

jalauzon-msft:
This test started failing in Live (along with some others) for async loop related reasons. My theory is that something changed in `pytest` or another library where if a test was not defined with `async def` it did not get an event loop on the main thread and likely just ran synchronously.

This particular test was not defined with `async def` as we had no local `await` calls but the call to `generate_oauth_token` did depend on the event loop because it made a real `ClientSecretCredential` which created an `asyncio.Lock` object. Changing this test to use `generate_fake_token` should fix the issue by removing the dependency on the event loop.


vincenttran-msft (approved):
LGTM 👍 makes sense to me 😄

jalauzon-msft:
Only tests failing are conda tests for unrelated reason. Will override

PR #27411, opened 2022-11-09
[EventHubs] update sync receive client ready flow

swathipil:
When reading from 32 partitions, the sync consumers were taking a long time to connect, b/c the client ready check was happening in a while loop. Each consumer was blocking while checking that the client was ready. Overall, this resulted in very slow connection times for the consumer client.

This is updating the consumers so that the client ready check is no longer blocking.

### DIFF
--- a/sdk/eventhub/azure-eventhub/azure/eventhub/_consumer.py
+++ b/sdk/eventhub/azure-eventhub/azure/eventhub/_consumer.py
@@ -198,10 +198,12 @@ class EventHubConsumer(
             )
             self._handler = cast("ReceiveClient", self._handler)
             self._handler.open(connection=conn)
-            while not self._handler.client_ready():
-                time.sleep(0.05)
-            self.handler_ready = True
+            self.handler_ready = False
             self.running = True
+        
+        if not self.handler_ready:
+            if self._handler.client_ready():
+                self.handler_ready = True
 
         return self.handler_ready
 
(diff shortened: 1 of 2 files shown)


sdk/eventhub/azure-eventhub/azure/eventhub/_consumer.py
```diff
@@ -198,10 +198,12 @@ def _open(self) -> bool:
             )
             self._handler = cast("ReceiveClient", self._handler)
             self._handler.open(connection=conn)
-            while not self._handler.client_ready():
-                time.sleep(0.05)
-            self.handler_ready = True
+            self.handler_ready = False
             self.running = True
+        
+        if not self.handler_ready:
```
kashifkhan, line 204: is `self.handler_ready` being used/set anywhere outside this? Since we are doing `self.handler_ready = False` a few lines above we might not need this check and only have the second if statement

swathipil: nope not used anywhere else!

kashifkhan (approved):

l0lawrence (commented):
lgtm!

annatisch (approved):
Awesome thanks for this!
Is there any investigation needed for async? Or the producers?
(If so, we could probably open issues to address as needed).
The live tests were all good?

swathipil:
> Awesome thanks for this! Is there any investigation needed for async? Or the producers? (If so, we could probably open issues to address as needed). The live tests were all good?

Going to run the live tests on this to double check that this doesn't result in any issues.

Producers and async consumer don't need to be updated. + The perf connect issue goes away with just the sync consumer fix.

More details:
Async consumer open was always calling client ready in a loop (for uamqp as well). As you mentioned, I think this is because it calls client_ready() and asyncio.sleep with await, so it wouldn't be blocking.

Sync and async producers were also always calling client_ready() in a loop.

PR #27413, opened 2022-11-10
[ACR] Support pull/push OCI blob and manifest in async client

YalinLi0312:
Resolves: https://github.com/Azure/azure-sdk-for-python/issues/23643

The support in sync client has been published in last release. The async implementation _was blocked_ by aiohttp client which not allowed to rewind stream data's position as **aiohttp client closes the file after sending**. We find a workaround, in this PR, we change the behavior of aiohttp client to do nothing after request sending out. 

Since the implementation is pretty hacky, want to hear concerns/thoughts from reviewers.

### DIFF
--- a/sdk/containerregistry/azure-containerregistry/azure/containerregistry/aio/_async_container_registry_client.py
+++ b/sdk/containerregistry/azure-containerregistry/azure/containerregistry/aio/_async_container_registry_client.py
@@ -3,7 +3,8 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
 # ------------------------------------
-from typing import TYPE_CHECKING, Any, Optional, overload, Union
+from io import BytesIO
+from typing import TYPE_CHECKING, Any, IO, Optional, overload, Union
 
 from azure.core.async_paging import AsyncItemPaged, AsyncList
 from azure.core.exceptions import (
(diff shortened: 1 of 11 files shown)


sdk/containerregistry/azure-containerregistry/azure/containerregistry/aio/_async_container_registry_client.py
```diff
@@ -741,6 +758,144 @@ async def update_tag_properties(
...
+            _, _, start_upload_response_headers = await self._client.container_registry_blob.start_upload(
+                repository, cls=_return_response, **kwargs
+            )
+            original = data.close
+            data.close = lambda: None
+            _, _, upload_chunk_response_headers = await self._client.container_registry_blob.upload_chunk(
+                start_upload_response_headers['Location'], data, cls=_return_response, **kwargs
+            )
+            data.close = original
+            digest = _compute_digest(data)
```
annelo-msft, line 828: Do we have a strategy for computing the digest incrementally if the input stream represents a 10GB file stream?

sdk/containerregistry/azure-containerregistry/azure/containerregistry/aio/_async_container_registry_client.py
```diff
@@ -741,6 +758,144 @@ async def update_tag_properties(
...
+
+        return digest
+
+    @distributed_trace_async
+    async def upload_blob(self, repository, data, **kwargs):
+        # type: (str, IO, **Any) -> str
+        """Upload an artifact blob.
+
+        :param str repository: Name of the repository
+        :param data: The blob to upload. Note: This must be a seekable stream.
```
annelo-msft, line 812: > seekable stream

I was able to rework the .NET implementation to upload a blob from a non-seekable stream, by uploading in chunks and incrementally computing the digest.  Let me know if you're interested in looking at that implementation.

sdk/containerregistry/azure-containerregistry/tests/test_container_registry_client_async.py
```diff
@@ -612,6 +614,130 @@ async def test_get_misspell_property(self, containerregistry_endpoint):
             last_udpated_on = properties.last_udpated_on
         last_updated_on = properties.last_updated_on
         assert last_udpated_on == last_updated_on
+    
+    @acr_preparer()
+    @recorded_by_proxy_async
+    async def test_upload_oci_manifest(self, containerregistry_endpoint):
```
annelo-msft, line 620: Do you have tests for downloading manifests and blobs as well?

sdk/containerregistry/azure-containerregistry/azure/containerregistry/aio/_async_container_registry_client.py
YalinLi0312: Large stream is not supported in this PR, haven't think about it yet. But I know it is in the new requirements.

sdk/containerregistry/azure-containerregistry/azure/containerregistry/aio/_async_container_registry_client.py
YalinLi0312: Sure! We can sync offline.

sdk/containerregistry/azure-containerregistry/tests/test_container_registry_client_async.py
YalinLi0312: Yes. I'm not testing them in separately like test_download_XXX(), each upload test has included `download` test, like in line631.

YalinLi0312:
Close it as there are much gap with current main, the updates will get reviewed in other PRs.

PR #27444, opened 2022-11-11
[EventHubs] check for any non-None values in amqp header/properties

swathipil:
In [[this line](https://github.com/Azure/azure-sdk-for-python/blob/3711eeb4269214caf70c4f8b034042a9b4b13385/sdk/eventhub/azure-eventhub/azure/eventhub/_transport/_uamqp_transport.py#L127)], we use `any()` to check if the AmqpAnnotatedMessage.header has any values set. However, 0/False/"" are falsy values and therefore not detected as values by any().

Adding check of the `len(header/properties.values()` against `values.count(None)` to check that non-None values are detected as values.

### DIFF
--- a/sdk/eventhub/azure-eventhub/azure/eventhub/amqp/_amqp_message.py
+++ b/sdk/eventhub/azure-eventhub/azure/eventhub/amqp/_amqp_message.py
@@ -346,6 +346,12 @@ class AmqpMessageHeader(DictMixin):
         self.durable = kwargs.get("durable")
         self.priority = kwargs.get("priority")
 
+    def _any(self):
+        # Returns True for values 0, False, "", etc. unlike any()
+        for val in self.values():
+            if val is not None:
+                return True
+        return False
 
 class AmqpMessageProperties(DictMixin):
     # pylint: disable=too-many-instance-attributes
@@ -437,3 +443,10 @@ class AmqpMessageProperties(DictMixin):
         self.group_id = kwargs.get("group_id")
         self.group_sequence = kwargs.get("group_sequence")
         self.reply_to_group_id = kwargs.get("reply_to_group_id")
+
+    def _any(self):
+        # Returns True for values 0, False, "", etc. unlike any()
+        for val in self.values():
+            if val is not None:
+                return True
+        return False
--- a/sdk/eventhub/azure-eventhub/azure/eventhub/amqp/_amqp_message.py
+++ b/sdk/eventhub/azure-eventhub/azure/eventhub/amqp/_amqp_message.py
@@ -346,12 +346,6 @@ class AmqpMessageHeader(DictMixin):
         self.durable = kwargs.get("durable")
         self.priority = kwargs.get("priority")
 
-    def _any(self):
-        # Returns True for values 0, False, "", etc. unlike any()
-        for val in self.values():
-            if val is not None:
-                return True
-        return False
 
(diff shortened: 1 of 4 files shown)


sdk/eventhub/azure-eventhub/azure/eventhub/amqp/_amqp_message.py
```diff
@@ -346,6 +346,12 @@ def __init__(self, **kwargs):
         self.durable = kwargs.get("durable")
         self.priority = kwargs.get("priority")
 
+    def _any(self):
```
swathipil, line 349: @annatisch -

I tried different things and adding an `_any` method seemed to be the fastest (and the most readable I think). All options:
```
1. any(a)
2. any(v is not None for v in a)
3. a.count(None) != len(a)
4.
def _any():
    for val in a:
        if val is not None:
            return True
    return False
```

I timed the above with different `a` values and generally got results like the following:
```
a = [None]*13
a[1] = False

0.9561780149946572
0.966271400002006
1.0146710249930038
0.9387445699903765

a = [None]*13
a[6] = False

0.9399554749878007
1.0586297749890945
0.9952111150036217
0.9624503749917495

a = [None]*13

1.0357584150042385
1.2306415550032397
0.9641022199939471
0.950713879999239
```

For the second `a` list above, `_any()` is the second fastest. However, in other runs, it was the fastest.

So, some randomness, but in general, `_any()` seemed fastest. Thoughts?

annatisch (approved):

sdk/eventhub/azure-eventhub/azure/eventhub/amqp/_amqp_message.py
annatisch: Awesome stuff! Thanks so much for doing this analysis :)
It's probably worth noting that from a perf perspective - calling a standalone function is not _exactly_ the same as calling an object method - however I expect the difference to be pretty minimal - and even if it's not - we could always just use a standalone function for this as well.

swathipil: so I tested actually calling AmqpMessageProperties, just to double check the speed of method vs. function vs. other calls. In general, it actually looks like 3. is consistently the fastest.
```
1. any(a.values())  # baseline
2. any(v is not None for v in a.values())
3.
v = a.values() 
v.count(None) != len(v)
5.
def _any():
    for val in a.values():
        if val is not None:
            return True
    return False
```

This is the result of grabbing the min of timeit.repeat.
```

from azure.eventhub.amqp import AmqpMessageProperties
a = AmqpMessageProperties()
a.message_id = False    # [False] + [None]*12

1.9633787999628112
2.168429500015918
1.89844889997039
2.05894580000313
1.9218437999952585

from azure.eventhub.amqp import AmqpMessageProperties
a = AmqpMessageProperties()
a.content_type = False  # [None]*6 + [False] + [None]*6

1.8671079999767244
2.518583500001114
1.971821799990721
2.1595329999690875
2.083914199960418

from azure.eventhub.amqp import AmqpMessageProperties
a = AmqpMessageProperties()     # [None]*13

1.9913484000135213
2.7721921000047587
1.979206099989824
2.3778754000086337
2.2316732999752276
```

Esp in the case of all values being None, 3. is the only one that is faster than the baseline. Based on this switching to using solution 3. Any concerns?

PR #27469, opened 2022-11-14
Communication/identity/migrate tests proxy

ostoliarova-msft:
# Description

### DIFF
--- a/sdk/communication/azure-communication-identity/tests.yml
+++ b/sdk/communication/azure-communication-identity/tests.yml
@@ -19,3 +19,6 @@ stages:
               - $(sub-config-communication-int-test-resources-common)
               - $(sub-config-communication-int-test-resources-python)
       Clouds: Public,Int
+      EnvVars:
+        AZURE_SKIP_LIVE_RECORDING: 'True'
+        AZURE_TEST_RUN_LIVE: 'true'
--- a/sdk/communication/azure-communication-identity/tests/conftest.py
+++ b/sdk/communication/azure-communication-identity/tests/conftest.py
@@ -1,0 +1,95 @@
+# --------------------------------------------------------------------------
+#
+# Copyright (c) Microsoft Corporation. All rights reserved.
+#
+# The MIT License (MIT)
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the ""Software""), to
+# deal in the Software without restriction, including without limitation the
+# rights to use, copy, modify, merge, publish, distribute, sublicense, and/or
+# sell copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED *AS IS*, WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+# FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+# IN THE SOFTWARE.
+#
+# --------------------------------------------------------------------------
+import pytest
+import os
+from devtools_testutils import test_proxy, add_general_regex_sanitizer, add_header_regex_sanitizer, \
+    set_default_session_settings, add_uri_regex_sanitizer, add_body_key_sanitizer, add_oauth_response_sanitizer
+from azure.communication.identity._shared.utils import parse_connection_str
+
+@pytest.fixture(scope="session", autouse=True)
+def add_sanitizers(test_proxy):
+    set_default_session_settings()
+
+    connection_str = os.getenv('COMMUNICATION_LIVETEST_DYNAMIC_CONNECTION_STRING') or \
+                          os.getenv('COMMUNICATION_LIVETEST_STATIC_CONNECTION_STRING')
+    endpoint, _ = parse_connection_str(connection_str)
+    resource_name = endpoint.split(".")[0]
+    add_general_regex_sanitizer(regex=resource_name, value="sanitized")
+
+    subscription_id = os.environ.get("AZURE_SUBSCRIPTION_ID", "00000000-0000-0000-0000-000000000000")
+    add_general_regex_sanitizer(regex=subscription_id, value="00000000-0000-0000-0000-000000000000")
+
+    m365_client_id = os.environ.get("COMMUNICATION_M365_APP_ID", "sanitized")
+    add_general_regex_sanitizer(regex=m365_client_id, value="sanitized")
+    #
+    m365_aad_authority = os.environ.get("COMMUNICATION_M365_AAD_AUTHORITY", "sanitized")
+    add_general_regex_sanitizer(regex=m365_aad_authority, value="sanitized")
+
+    m365_aad_tenant = os.environ.get("COMMUNICATION_M365_AAD_TENANT", "sanitized")
+    add_general_regex_sanitizer(regex=m365_aad_tenant, value="sanitized")
+    #
+    msal_username = os.environ.get("COMMUNICATION_MSAL_USERNAME", "sanitized")
+    add_general_regex_sanitizer(regex=msal_username, value="sanitized")
+
+    msal_password = os.environ.get("COMMUNICATION_MSAL_PASSWORD", "sanitized")
+    add_general_regex_sanitizer(regex=msal_password, value="sanitized")
+
+    expired_teams_token = os.environ.get("COMMUNICATION_EXPIRED_TEAMS_TOKEN", "sanitized")
+    add_general_regex_sanitizer(regex=expired_teams_token, value="sanitized")
+
+    azure_tenant_id = os.environ.get("AZURE_TENANT_ID", "sanitized")
+    add_general_regex_sanitizer(regex=azure_tenant_id, value="sanitized")
+
+    azure_client_secret = os.environ.get("AZURE_CLIENT_SECRET", "sanitized")
+    add_general_regex_sanitizer(regex=azure_client_secret, value="sanitized")
+
+    azure_client_id = os.environ.get("AZURE_CLIENT_ID", "sanitized")
+    add_general_regex_sanitizer(regex=azure_client_id, value="sanitized")
+
+    add_body_key_sanitizer(json_path="*.id", value="sanitized")
+    add_body_key_sanitizer(json_path="*.token", value="sanitized")
+    add_body_key_sanitizer(json_path="token", value="sanitized")
+    add_body_key_sanitizer(json_path="*.userId", value="sanitized")
+    add_body_key_sanitizer(json_path="userId", value="sanitized")
+    add_body_key_sanitizer(json_path="*.domain_name", value="sanitized")
+
+    add_general_regex_sanitizer(regex='/identities/([^/?]+)', value='/identities/sanitized')
+    add_general_regex_sanitizer(regex='common/userrealm/([^/.]+)', value='common/userrealm/sanitized@test')
+
+    add_header_regex_sanitizer(key="P3P", value="sanitized")
+    add_header_regex_sanitizer(key="Set-Cookie", value="sanitized")
+    add_header_regex_sanitizer(key="Date", value="sanitized")
+    add_header_regex_sanitizer(key="Cookie", value="sanitized")
+    add_header_regex_sanitizer(key="client-request-id", value="sanitized")
+    add_header_regex_sanitizer(key="MS-CV", value="sanitized")
+    add_header_regex_sanitizer(key="X-Azure-Ref", value="sanitized")
+    add_header_regex_sanitizer(key="x-ms-content-sha256", value="sanitized")
+    add_header_regex_sanitizer(key="x-ms-client-request-id", value="sanitized")
+    add_header_regex_sanitizer(key="x-ms-date", value="sanitized")
+    add_header_regex_sanitizer(key="x-ms-ests-server", value="sanitized")
+    add_header_regex_sanitizer(key="x-ms-request-id", value="sanitized")
+    add_header_regex_sanitizer(key="Content-Security-Policy-Report-Only", value="sanitized")
+
--- a/sdk/communication/azure-communication-identity/tests/conftest.py
+++ b/sdk/communication/azure-communication-identity/tests/conftest.py
@@ -44,13 +44,7 @@ def add_sanitizers(test_proxy):
 
     m365_client_id = os.environ.get("COMMUNICATION_M365_APP_ID", "sanitized")
     add_general_regex_sanitizer(regex=m365_client_id, value="sanitized")
-    #
-    m365_aad_authority = os.environ.get("COMMUNICATION_M365_AAD_AUTHORITY", "sanitized")
-    add_general_regex_sanitizer(regex=m365_aad_authority, value="sanitized")
 
-    m365_aad_tenant = os.environ.get("COMMUNICATION_M365_AAD_TENANT", "sanitized")
-    add_general_regex_sanitizer(regex=m365_aad_tenant, value="sanitized")
-    #
     msal_username = os.environ.get("COMMUNICATION_MSAL_USERNAME", "sanitized")
     add_general_regex_sanitizer(regex=msal_username, value="sanitized")
 
--- a/sdk/communication/azure-communication-identity/tests/test_communication_identity_client.py
+++ b/sdk/communication/azure-communication-identity/tests/test_communication_identity_client.py
@@ -19,13 +19,6 @@ from azure.communication.identity._shared.utils import parse_connection_str
 from msal import PublicClientApplication
 
 
-class ArgumentPasserTime:
-    def __call__(self, fn):
-        def _wrapper(test_class, _, time, **kwargs):
-            fn(test_class, _, time, **kwargs)
-        return _wrapper
-
-
 class ArgumentPasser:
     def __call__(self, fn):
         def _wrapper(test_class, _, value, **kwargs):
@@ -87,8 +80,8 @@ class TestCommunicationIdentityClient(AzureRecordedTestCase):
         assert token_response.token is not None
 
     @CommunicationPreparer()
-    @pytest.mark.parametrize("_, time", [("min_valid_hours", 1), ("max_valid_hours", 24)])
-    @ArgumentPasserTime()
+    @pytest.mark.parametrize("_, value", [("min_valid_hours", 1), ("max_valid_hours", 24)])
+    @ArgumentPasser()
     @recorded_by_proxy
     def test_create_user_and_token_with_valid_custom_expirations_new(self, _, valid_hours):
         self.communication_environment()
@@ -108,8 +101,8 @@ class TestCommunicationIdentityClient(AzureRecordedTestCase):
             assert is_token_expiration_within_allowed_deviation(token_expires_in, token_response.expires_on)
 
     @CommunicationPreparer()
-    @pytest.mark.parametrize("_, time", [("min_invalid_mins", 59), ("max_invalid_mins", 1441)])
-    @ArgumentPasserTime()
+    @pytest.mark.parametrize("_, value", [("min_invalid_mins", 59), ("max_invalid_mins", 1441)])
+    @ArgumentPasser()
     @recorded_by_proxy
     def test_create_user_and_token_with_invalid_custom_expirations(self, _, time):
         self.communication_environment()
@@ -162,10 +155,10 @@ class TestCommunicationIdentityClient(AzureRecordedTestCase):
         assert token_response.token is not None
 
     @CommunicationPreparer()
-    @pytest.mark.parametrize("_, time", [("min_valid_hours", 1), ("max_valid_hours", 24)])
-    @ArgumentPasserTime()
+    @pytest.mark.parametrize("_, value", [("min_valid_hours", 1), ("max_valid_hours", 24)])
+    @ArgumentPasser()
     @recorded_by_proxy
-    def test_get_token_with_valid_custom_expirations(self, _, time):
+    def test_get_token_with_valid_custom_expirations(self, _, value):
         self.communication_environment()
 
         identity_client = CommunicationIdentityClient.from_connection_string(
@@ -174,7 +167,7 @@ class TestCommunicationIdentityClient(AzureRecordedTestCase):
         )
         user = identity_client.create_user()
 
-        token_expires_in = timedelta(hours=time)
+        token_expires_in = timedelta(hours=value)
         token_response = identity_client.get_token(user, scopes=[CommunicationTokenScope.CHAT],
                                                    token_expires_in=token_expires_in)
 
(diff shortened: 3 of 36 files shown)


sdk/communication/azure-communication-identity/tests.yml
```diff
@@ -19,3 +19,6 @@ stages:
               - $(sub-config-communication-int-test-resources-common)
               - $(sub-config-communication-int-test-resources-python)
       Clouds: Public,Int
+      EnvVars:
+        AZURE_SKIP_LIVE_RECORDING: 'True'
+        AZURE_TEST_RUN_LIVE: 'true'
```
petrsvihlik, line 24: What do these settings do exactly? (link to some docs would help)

sdk/communication/azure-communication-identity/tests/asynctestcase.py
```diff
@@ -1,28 +0,0 @@
...
-# Licensed under the MIT License. See License.txt in the project root for
-# license information.
-# --------------------------------------------------------------------------
-import functools
-import asyncio
-from azure_devtools.scenario_tests.utilities import trim_kwargs_from_test_function
-from testcase import CommunicationIdentityTestCase
-
-
-class AsyncCommunicationIdentityTestCase(CommunicationIdentityTestCase):
```
petrsvihlik, line 14: Why is this class no longer required?

sdk/communication/azure-communication-identity/tests/conftest.py
```diff
@@ -0,0 +1,92 @@
...
+
+# cSpell:ignore ests
+import pytest
+import os
+from devtools_testutils import test_proxy, add_general_regex_sanitizer, add_header_regex_sanitizer, \
+    set_default_session_settings, add_body_key_sanitizer, add_oauth_response_sanitizer
+from azure.communication.identity._shared.utils import parse_connection_str
+
+@pytest.fixture(scope="session", autouse=True)
+def add_sanitizers(test_proxy):
```
petrsvihlik, line 35: How was the sanitization configured before? Where did the sanitizers live before?

petrsvihlik (commented):
@ostoliarova-msft just a couple of questions since I'm not familiar with the migration.
it would be helpful to always provide a summary of changes in the description of a PR...at least for bigger PRs.

sdk/communication/azure-communication-identity/tests.yml
ostoliarova-msft: @petrsvihlik these lines enable proxy on live pipelines
You can check the migration guide for more information: https://github.com/Azure/azure-sdk-for-python/blob/main/doc/dev/test_proxy_migration_guide.md#live-test-pipelines

sdk/communication/azure-communication-identity/tests/asynctestcase.py
ostoliarova-msft: This class inherits from `CommunicationIdentityTestCase` which inherited from `CommunicationTestCase`, which in turn inherits from `AzureTestCase`. As now our tests should inherit from `AzureRecordedTestCase`, we cannot use this chain of inherited classes anymore (and don't need too because they were mostly (re)setting environment variables). Using proxy allows to make a shortcut and only use a decorator to setup these environment variables where they are needed.
As `CommunicationIdentityTestCase` and `CommunicationTestCase` are not used anymore, I removed these files.

sdk/communication/azure-communication-identity/tests/conftest.py
ostoliarova-msft: @petrsvihlik they were configured with scrubber.
"Since the test proxy doesn't use [vcrpy](https://vcrpy.readthedocs.io/), tests don't use a scrubber to sanitize values in recordings". You can see more on this here: https://github.com/Azure/azure-sdk-for-python/blob/main/doc/dev/test_proxy_migration_guide.md#register-sanitizers

sdk/communication/azure-communication-identity/tests/conftest.py
```diff
@@ -0,0 +1,92 @@
...
+    set_default_session_settings()
+    add_oauth_response_sanitizer()
+
+    connection_str = os.getenv('COMMUNICATION_LIVETEST_DYNAMIC_CONNECTION_STRING')
+    if connection_str is not None:
+        endpoint, _ = parse_connection_str(connection_str)
+        resource_name = endpoint.split(".")[0]
+        add_general_regex_sanitizer(regex=resource_name, value="sanitized")
```
mccoyp, line 43: There are some cases where complex strings (like connection strings) have caused issues when being used with a regex sanitizer. It looks like you're sanitizing a simplified value instead of the full connection string, but I would recommend using `add_general_string_sanitizer` in place of the regex sanitizer for environment variables since we know that these variables will have plain string values

sdk/communication/azure-communication-identity/tests/test_communication_identity_client.py
```diff
@@ -8,42 +8,35 @@
...
-        super(CommunicationIdentityClientTest, self).setUp()
-        self.recording_processors.extend([
-            BodyReplacerProcessor(keys=["id", "token", "appId", "userId", "domain_name"]),
-            URIIdentityReplacer(),
-            URIMsalUsernameReplacer()])
-    
+from acs_identity_decorator import acs_identity_decorator
+
+
+class ArgumentPasser:
```
mccoyp, line 22: Adding context for other reviewers: this class is necessary for using parametrized tests with the test proxy. Documentation explaining why is here: https://github.com/Azure/azure-sdk-for-python/blob/main/doc/dev/test_proxy_migration_guide.md#use-pytestmarkparametrize-with-migrated-tests

sdk/communication/azure-communication-identity/tests/test_communication_identity_client.py
```diff
@@ -69,58 +66,57 @@ def test_create_user_and_token(self, communication_livetest_dynamic_connection_s
...
-    @parameterized.expand(
-        [
-            ("min_valid_hours", 1),
-            ("max_valid_hours", 24),
-        ]
-    ) 
-    def test_create_user_and_token_with_valid_custom_expirations(self, _, valid_hours):
+
+    @CommunicationPreparer()
+    @pytest.mark.parametrize("_, value", [("min_valid_hours", 1), ("max_valid_hours", 24)])
```
mccoyp, line 71: Does this `pytest.mark.parametrize` use satisfy your test requirements, or do you still need a workaround for using `parameterized.expand`? I assumed your tests were sorted out once I saw your PR, but I just want to be sure.

In general I would recommend using pytest's `parametrize` decorator because it avoids a third-party dependency on `parameterized`, but it depends on the situation.

sdk/communication/azure-communication-identity/tests/test_communication_identity_client.py
```diff
@@ -250,16 +250,20 @@ def test_delete_user(self, communication_livetest_dynamic_connection_string):
...
         identity_client = CommunicationIdentityClient.from_connection_string(
             communication_livetest_dynamic_connection_string,
             http_logging_policy=get_http_logging_policy()
         )
 
         with pytest.raises(Exception) as ex:
-            user, token_response = identity_client.create_user_and_token(scopes=None)
+            identity_client.create_user_and_token(scopes=None)
+
+        assert ex is not None
```
mccoyp, line 263: I'm not sure if this assertion adds much since we know we caught an exception by this point, but it also doesn't hurt. Is there any consistent content in the exception message -- or a more specific type than `Exception` -- that we can assert for instead?

mccoyp (approved):
This is awesome work -- thank you for all of your efforts on this! Just left a couple of comments and suggestions, but things look great 🙂

sdk/communication/azure-communication-identity/tests/test_communication_identity_client.py
ostoliarova-msft: Yes, the approach I'm using in this PR satisfies our requirements. I also think that reducing the number of external dependencies is good.

sdk/communication/azure-communication-identity/tests/conftest.py
ostoliarova-msft: Thanks! I'll update.

sdk/communication/azure-communication-identity/tests/test_communication_identity_client.py
ostoliarova-msft: Thanks! That's a good point. I updated it.

sdk/communication/azure-communication-identity/tests.yml
petrsvihlik: thank you!

petrsvihlik (approved):
thanks @mccoyp for your help!

I have no more comments.

sdk/communication/azure-communication-identity/tests/test_communication_identity_client.py
```diff
@@ -8,42 +8,35 @@
...
+class ArgumentPasser:
+    def __call__(self, fn):
+        def _wrapper(test_class, _, value, **kwargs):
+            fn(test_class, _, value, **kwargs)
+        return _wrapper
+
+class TestClient(AzureRecordedTestCase):
     @CommunicationPreparer()
+    @recorded_by_proxy
     def test_create_user_from_managed_identity(self, communication_livetest_dynamic_connection_string):
```
AikoBB, line 31: I was wondering whether any test parameters(here I mean `communication_livetest_dynamic_connection_string`) it should be in the kwargs as explained in here [new test structure](https://github.com/Azure/azure-sdk-for-python/blob/main/doc/dev/test_proxy_migration_guide.md#new-test-structure) :
`So, arguments that are passed to a test by a preparer -- for example, example_variable in test_example_with_preparer above -- should be accepted via **kwargs and popped off at the start of a test.`

sdk/communication/azure-communication-identity/tests/conftest.py
```diff
@@ -0,0 +1,92 @@
...
+from devtools_testutils import test_proxy, add_general_regex_sanitizer, add_header_regex_sanitizer, \
+    set_default_session_settings, add_body_key_sanitizer, add_oauth_response_sanitizer, add_general_string_sanitizer
+from azure.communication.identity._shared.utils import parse_connection_str
+
+@pytest.fixture(scope="session", autouse=True)
+def add_sanitizers(test_proxy):
+    set_default_session_settings()
+    add_oauth_response_sanitizer()
+
+    connection_str = os.getenv('COMMUNICATION_LIVETEST_DYNAMIC_CONNECTION_STRING')
```
AikoBB, line 39: is there any reason behind usings `os.getenv` here and other places `os.environ.get`? :)

sdk/communication/azure-communication-identity/tests/conftest.py
```diff
@@ -0,0 +1,92 @@
...
+    if connection_str is not None:
+        endpoint, _ = parse_connection_str(connection_str)
+        resource_name = endpoint.split(".")[0]
+        add_general_string_sanitizer(target=resource_name, value="sanitized")
+
+    subscription_id = os.environ.get("AZURE_SUBSCRIPTION_ID", "00000000-0000-0000-0000-000000000000")
```
AikoBB, line 45: are we using `AZURE_SUBSCRIPTION_ID` env.var for our tests and is there any API returning or requiring a subscription id in the request? same for the lines 60, 63, 66

sdk/communication/azure-communication-identity/tests/conftest.py
```diff
@@ -0,0 +1,92 @@
...
+    azure_tenant_id = os.environ.get("AZURE_TENANT_ID", "sanitized")
+    add_general_regex_sanitizer(regex=azure_tenant_id, value="sanitized")
+
+    azure_client_secret = os.environ.get("AZURE_CLIENT_SECRET", "sanitized")
+    add_general_regex_sanitizer(regex=azure_client_secret, value="sanitized")
+
+    azure_client_id = os.environ.get("AZURE_CLIENT_ID", "sanitized")
+    add_general_regex_sanitizer(regex=azure_client_id, value="sanitized")
+
+    add_body_key_sanitizer(json_path="*.id", value="sanitized")
```
AikoBB, line 69: in the old sanitizer we were also sanitizing [appID](https://github.com/Azure/azure-sdk-for-python/blob/bacc704b36a8ac9fed0539801d54a6c6e7d8e956/sdk/communication/azure-communication-identity/tests/test_communication_identity_client.py#L33)

sdk/communication/azure-communication-identity/tests/conftest.py
```diff
@@ -0,0 +1,92 @@
...
+    add_body_key_sanitizer(json_path="*.token", value="sanitized")
+    add_body_key_sanitizer(json_path="token", value="sanitized")
+    add_body_key_sanitizer(json_path="*.userId", value="sanitized")
+    add_body_key_sanitizer(json_path="userId", value="sanitized")
+    add_body_key_sanitizer(json_path="*.domain_name", value="sanitized")
+
+    add_general_regex_sanitizer(regex='/identities/([^/?]+)', value='/identities/sanitized')
+    add_general_regex_sanitizer(regex='common/userrealm/([^/.]+)', value='common/userrealm/sanitized@test')
+
+    add_header_regex_sanitizer(key="P3P", value="sanitized")
```
AikoBB, line 79: I could not find the [url](https://github.com/Azure/azure-sdk-for-python/blob/bacc704b36a8ac9fed0539801d54a6c6e7d8e956/sdk/communication/azure-communication-identity/tests/_shared/helper.py#L61) sanitizer, is it missing?

sdk/communication/azure-communication-identity/tests/test_communication_identity_client.py
AikoBB: do we really need to assert the exeption for nonnullability and the message value? what if the auth team will change the exception message, so it means we have to update our tests each time the auth team update the message from service side? :)

sdk/communication/azure-communication-identity/tests/test_communication_identity_client.py
```diff
@@ -298,132 +315,143 @@ def test_get_token_with_no_scopes(self, communication_livetest_dynamic_connectio
...
-        ]
-    )
-    def test_get_token_for_teams_user_with_invalid_token(self, _, invalid_token):
-        if(self.skip_get_token_for_teams_user_test()):
+    @CommunicationPreparer()
+    @pytest.mark.parametrize("_, value", [("empty_token", ""), ("invalid_token", "invalid")])
+    @ArgumentPasser()
+    @acs_identity_decorator
+    @recorded_by_proxy
+    def test_get_token_for_teams_user_with_invalid_token(self, _, value):
```
AikoBB, line 362: I suggest reverting back to the old name of the param, same comment for everywhere where it's changed to `value`

sdk/communication/azure-communication-identity/tests/utils.py
```diff
@@ -18,4 +19,24 @@ def is_token_expiration_within_allowed_deviation(
...
+        scopes = [
+            "https://auth.msft.communication.azure.com/Teams.ManageCalls",
+            "https://auth.msft.communication.azure.com/Teams.ManageChats"
+        ]
+        result = msal_app.acquire_token_by_username_password(username=self.msal_username, password=self.msal_password, scopes=scopes)
+        teams_user_aad_token = result["access_token"]
+        teams_user_oid = result["id_token_claims"]["oid"]
+    return teams_user_aad_token, teams_user_oid
+
+def skip_get_token_for_teams_user_test(self):
```
AikoBB, line 41: maybe it's better to pass `self.skip_get_token_for_teams_user_tests` instead of `self`

sdk/communication/azure-communication-identity/tests/test_communication_identity_client_async.py
```diff
@@ -6,42 +6,31 @@
...
-        return self.token
-class CommunicationIdentityClientTestAsync(AsyncCommunicationIdentityTestCase):
-    def setUp(self):
-        super(CommunicationIdentityClientTestAsync, self).setUp()
-        self.recording_processors.extend([
-            BodyReplacerProcessor(keys=["id", "token", "appId", "userId", "domain_name"]),
-            URIIdentityReplacer(),
-            URIMsalUsernameReplacer()])
-    
+class TestClientAsync(AzureRecordedTestCase):
```
AikoBB, line 22: same comments that I left for sync tests

sdk/communication/azure-communication-identity/tests/test_communication_identity_client.py
mccoyp: That is a good point -- using the `**kwargs` structure is more future-proof. Using positional parameters for test parameters isn't compatible with pytest fixtures, which are ultimately what we'd like to use in tests in place of decorators like `@recorded_by_proxy`.

sdk/communication/azure-communication-identity/tests/utils.py
ostoliarova-msft: Thank you! I updated.

sdk/communication/azure-communication-identity/tests/test_communication_identity_client_async.py
ostoliarova-msft: Thanks! Updated this, too.

sdk/communication/azure-communication-identity/tests/test_communication_identity_client.py
```diff
@@ -69,58 +66,57 @@ def test_create_user_and_token(self, communication_livetest_dynamic_connection_s
...
-            ("min_valid_hours", 1),
-            ("max_valid_hours", 24),
-        ]
-    ) 
-    def test_create_user_and_token_with_valid_custom_expirations(self, _, valid_hours):
+
+    @CommunicationPreparer()
+    @pytest.mark.parametrize("_, value", [("min_valid_hours", 1), ("max_valid_hours", 24)])
+    @ArgumentPasser()
+    @acs_identity_decorator
```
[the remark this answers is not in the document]
ostoliarova-msft: I refactored and moved this logic to a super class.

sdk/communication/azure-communication-identity/tests/test_communication_identity_client.py
```diff
@@ -69,58 +66,57 @@ def test_create_user_and_token(self, communication_livetest_dynamic_connection_s
...
-        ]
-    )
-    def test_create_user_and_token_with_invalid_custom_expirations(self, _, invalid_mins):
+
+    @CommunicationPreparer()
+    @pytest.mark.parametrize("_, value", [("min_invalid_mins", 59), ("max_invalid_mins", 1441)])
+    @ArgumentPasser()
+    @acs_identity_decorator
+    @recorded_by_proxy
+    def test_create_user_and_token_with_invalid_custom_expirations(self, _, time):
```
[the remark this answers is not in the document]
ostoliarova-msft: Done

sdk/communication/azure-communication-identity/tests/test_communication_identity_client.py
```diff
@@ -142,59 +139,59 @@ def test_get_token(self, communication_livetest_dynamic_connection_string):
...
-        ]
-    )
-    def test_get_token_with_valid_custom_expirations(self, _, valid_hours):
+
+    @CommunicationPreparer()
+    @pytest.mark.parametrize("_, value", [("min_valid_hours", 1), ("max_valid_hours", 24)])
+    @ArgumentPasser()
+    @acs_identity_decorator
+    @recorded_by_proxy
+    def test_get_token_with_valid_custom_expirations(self, _, value):
```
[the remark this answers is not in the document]
ostoliarova-msft: Done

sdk/communication/azure-communication-identity/tests/test_communication_identity_client.py
ostoliarova-msft: Done

sdk/communication/azure-communication-identity/tests/conftest.py
ostoliarova-msft: We could use the same approach. Updated

sdk/communication/azure-communication-identity/tests/conftest.py
ostoliarova-msft: You're right. I removed these.

sdk/communication/azure-communication-identity/tests/conftest.py
ostoliarova-msft: I used  `add_general_regex_sanitizer(regex=m365_client_id, value="sanitized")` to sanitize it but I'll use `add_body_key_sanitizer` instead.

sdk/communication/azure-communication-identity/tests/test_communication_identity_client.py
ostoliarova-msft: Thank you for pointing this out. I refactored not to use test parameters like this.

sdk/communication/azure-communication-identity/tests/test_communication_identity_client.py
```diff
@@ -69,58 +66,57 @@ def test_create_user_and_token(self, communication_livetest_dynamic_connection_s
...
-        
-    @parameterized.expand(
-        [
-            ("min_valid_hours", 1),
-            ("max_valid_hours", 24),
-        ]
-    ) 
-    def test_create_user_and_token_with_valid_custom_expirations(self, _, valid_hours):
+
+    @CommunicationPreparer()
```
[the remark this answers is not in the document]
ostoliarova-msft: Removed. However, it needed for async tests not to be skipped. If we remove it, we need to use another decorator that will await http calls. So I left it for async tests to reuse existing code.

sdk/communication/azure-communication-identity/tests/conftest.py
ostoliarova-msft: `add_general_string_sanitizer(target=resource_name, value="sanitized")` will sanitize urls

sdk/communication/azure-communication-identity/tests/conftest.py
AikoBB: for other body keys you used any match regex + dot and then the key name(f.e "*.token"), why we are introducing these kind of santitizers and it should be applied for appId as well?

sdk/communication/azure-communication-identity/tests/utils.py
```diff
@@ -18,4 +19,24 @@ def is_token_expiration_within_allowed_deviation(
     expected_seconds = expected_token_expiration.total_seconds();
     time_difference = abs(expected_seconds - token_expiration_in_seconds)
     allowed_time_difference = expected_seconds * allowed_deviation
-    return time_difference < allowed_time_difference
\ No newline at end of file
+    return time_difference < allowed_time_difference
+
+def generate_teams_user_aad_token(self):
```
AikoBB, line 24: as we returned back the parent class, maybe better place for this method and skip_get_token_for_teams_user_test will be the ACSIdentityTestCase, instead of expecting any params they can reach to the class properties, wdyt?

sdk/communication/azure-communication-identity/tests/test_communication_identity_client.py
ostoliarova-msft: As there are several possible exceptions we are checking for and we need a way to distinguish them: I had an issue of getting the "wrong" exception in these tests and spending a lot of time to find a problem when the test was green. Checking the error message is a trade-off here as there seems to be no constant property to distinguish between exceptions raised for different reasons. 
Updating the tests if Auth team changes the wording shouldn't take much time.
Please share if you have an idea how else we could check the kind of exception we're getting here.

sdk/communication/azure-communication-identity/tests/conftest.py
ostoliarova-msft: The dot-syntax is for addressing nested fields in json. We don't have such cases for 'appId' and all values are sanitized

sdk/communication/azure-communication-identity/tests/test_communication_identity_client.py
AikoBB: I would prefer to check only for the response status code and non-emptiness of the error message, as we did for the [async tests](https://github.com/Azure/azure-sdk-for-python/blob/e778c87f176bd1a44b6cc6adbbf68ab0caa108c5/sdk/communication/azure-communication-identity/tests/test_communication_identity_client_async.py#L141), but if others are fine with your approach then we can ignore my comment here

sdk/communication/azure-communication-identity/tests/utils.py
ostoliarova-msft: Good idea! I'll update

sdk/communication/azure-communication-identity/tests/test_communication_identity_client.py
ostoliarova-msft: Let's stick to the approach from async tests than. I've updated

PR #27474, opened 2022-11-14
update samples check for mypy

kristapratico:
A few updates:

- mypy is unique in that **functions that do not contain annotations do not get checked**. Most of our samples are contained in functions without any type annotations like [this](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/translation/azure-ai-translation-document/samples/sample_begin_translation.py#L29). This gives the false impression that samples pass by mypy when in reality they are not being checked _at all_. We can fix this by running mypy with `--check-untyped-defs`, but this isn't something we want to enable on all of our library code (we don't enforce that all code needs type hints, unannotated functions are okay if non-public). To avoid developers missing type checking errors in samples, this PR enables `--check-untyped-defs` for samples _only_. We additionally must tell mypy not to follow imports from the samples or else it reports new errors in non-public code.
- We log the aka.ms link to the typing guide before raising exceptions now

### DIFF
--- a/eng/tox/run_mypy.py
+++ b/eng/tox/run_mypy.py
@@ -7,7 +7,7 @@
 
 # This script is used to execute mypy within a tox environment.
 
-from subprocess import check_call
+from subprocess import check_call, CalledProcessError
 import argparse
 import os
 import logging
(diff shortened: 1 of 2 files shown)


scbedd (approved):
👍

PR #27482, opened 2022-11-14
[Storage] Fix more async test bugs

jalauzon-msft:
Fixing some more async test bugs in datalake `test_cpk_async`. Some of the tests in this file were failing for the, now infamous, `RuntimeError: There is no current event loop in thread 'MainThread'` in the Live test pipeline. This was because the `teardown` method was not defined with `async def` and therefore wasn't running in an async context. 

This changes `teardown` to use `async def` and also changes the method to just use `await` instead of using the event loop. Additionally fixing this test file to use `AsyncStorageRecordedTestCase`.


vincenttran-msft (approved):
LGTM! Definitely infamous for sure 🤣

PR #27485, opened 2022-11-14
[Storage] Adjust test case for `test_rename_from_a_shorter_directory_to_longer_directory`

vincenttran-msft:
This PR adjusts this testcase which has been marked as skipped since it's migration to track2. 
The issue here is that the original test case had a `directory_name` variable that contained the source destination's directory name, but then didn't use the variable when creating the source destination's directory. It would then later pass the variable to the rename API, but this directory didn't exist due to using a different name when creating the source destination's directory ("old")


jalauzon-msft (approved):
Thanks for fixing this up!

PR #27518, opened 2022-11-16
[Storage] Add support for `AsyncIterable` as data type for blob/file upload

### LINKED ISSUE #17478: Blob Storage: Support AsyncIterable blob uploads
Currently, `azure.storage.blob.aio.BlobClient`'s `upload_blob` (as well as `stage_blob` etc), don't support data that is `AsyncIterable`. That is the following will not work:

```py
async def iterator() -> AsyncGenerator[bytes, None]:
    for i in range(10):
        yield str(i).encode("latin1")
        await asyncio.sleep(0.1)

await blob_client.upload_blob(iterator())
```

I'd like the `aio` versions of the upload functions to accept `AsyncIterable[AnyStr]` in addition to `Iterable[AnyStr]` and `IO[AnyStr]`.

**Describe alternatives you've considered**
Using intermediate pipes, creating a stream that we would feed manually. All alternatives do blocking and that doesn't fit in the asyncio model.

**Additional context**
Trying to have Starlette forward some data to blob storage, by using `starlette.request.Request`'s `stream()` async generator, so as to not load the entire document into memory.

jalauzon-msft:
Adding support for `AsyncIterable` (async generators) as an input data type for blob, datalake, and file-share uploads.

Matched the approach of `Iterable` support by creating `AsyncIterStreamer` that mirrors the existing `IterStreamer` except for async generators.

Resolves #17478.


vincenttran-msft (approved):
Thanks for getting to this! Was interesting to review to see what work goes into adding support for an async equivalent of a sync model we already have ✔️

sdk/storage/azure-storage-blob/azure/storage/blob/_shared/uploads_async.py
```diff
@@ -421,3 +423,51 @@ async def _upload_chunk(self, chunk_offset, chunk_data):
...
+    def __init__(self, generator: AsyncGenerator[Union[bytes, str], None], encoding: str = "UTF-8"):
+        self.generator = generator
+        self.iterator = generator.__aiter__()
+        self.leftover = b""
+        self.encoding = encoding
+
+    def __len__(self):
+        return self.generator.__len__()
+
+    def __aiter__(self):
```
jalauzon-msft, line 441: @annatisch, curious if you think we actually need to include `__aiter__` and `__anext__` here. I did this because the existing sync `IterStreamer` does but the idea behind this class is to wrap a generator in a stream so I think we only use stream methods from this class. I don't see why we would also need this class to be an Iterable.

sdk/storage/azure-storage-blob/azure/storage/blob/_shared/uploads_async.py
```diff
@@ -421,3 +423,51 @@ async def _upload_chunk(self, chunk_offset, chunk_data):
...
+    File-like streaming object for AsyncGenerators.
+    """
+    def __init__(self, generator: AsyncGenerator[Union[bytes, str], None], encoding: str = "UTF-8"):
+        self.generator = generator
+        self.iterator = generator.__aiter__()
+        self.leftover = b""
+        self.encoding = encoding
+
+    def __len__(self):
+        return self.generator.__len__()
```
annatisch, line 439: I don't believe generators define `__len__`?

sdk/storage/azure-storage-blob/azure/storage/blob/_shared/uploads_async.py
```diff
@@ -421,3 +423,51 @@ async def _upload_chunk(self, chunk_offset, chunk_data):
...
+                if isinstance(chunk, str):
+                    chunk = chunk.encode(self.encoding)
+                data += chunk
+                count += len(chunk)
+        # This means count < size and what's leftover will be returned in this call.
+        except StopAsyncIteration:
+            self.leftover = b""
+
+        if count >= size:
+            self.leftover = data[size:]
```
annatisch, line 471: If I do an upload with `concurrency > 1` am I going to hit race conditions around this?

sdk/storage/azure-storage-blob/azure/storage/blob/_shared/uploads_async.py
```diff
@@ -421,3 +423,51 @@ async def _upload_chunk(self, chunk_offset, chunk_data):
...
         pass
+
+
+class AsyncIterStreamer():
+    """
+    File-like streaming object for AsyncGenerators.
+    """
+    def __init__(self, generator: AsyncGenerator[Union[bytes, str], None], encoding: str = "UTF-8"):
+        self.generator = generator
+        self.iterator = generator.__aiter__()
```
annatisch, line 434: Do we need to store both of these on `self`?

sdk/storage/azure-storage-blob/azure/storage/blob/_blob_client.py
```diff
@@ -357,13 +358,12 @@ def get_account_information(self, **kwargs):
             process_storage_error(error)
 
     def _upload_blob_options(  # pylint:disable=too-many-statements
-            self, data,  # type: Union[bytes, str, Iterable[AnyStr], IO[AnyStr]]
-            blob_type=BlobType.BlockBlob,  # type: Union[str, BlobType]
-            length=None,  # type: Optional[int]
-            metadata=None,  # type: Optional[Dict[str, str]]
+            self, data: Union[bytes, str, Iterable[AnyStr], AsyncIterable[AnyStr], IO[AnyStr]],
```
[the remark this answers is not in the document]
jalauzon-msft: Yep, this is the shared method. Technically, that manes that `AsyncIterable` is supported on the sync `upload_blob` API but I only documented it on the async one.

sdk/storage/azure-storage-blob/azure/storage/blob/_container_client.py
```diff
@@ -910,14 +910,13 @@ def find_blobs_by_tags(
 
     @distributed_trace
     def upload_blob(
-            self, name,  # type: Union[str, BlobProperties]
-            data,  # type: Union[Iterable[AnyStr], IO[AnyStr]]
-            blob_type=BlobType.BlockBlob,  # type: Union[str, BlobType]
-            length=None,  # type: Optional[int]
-            metadata=None,  # type: Optional[Dict[str, str]]
+            self, name: Union[str, BlobProperties],
```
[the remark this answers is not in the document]
jalauzon-msft: Slowing trying to update these as I make changes to the individual APIs. Still a long way to go :)

sdk/storage/azure-storage-blob/azure/storage/blob/_shared/uploads_async.py
jalauzon-msft: Yeah, not sure why this was here. Removed it

sdk/storage/azure-storage-blob/azure/storage/blob/_shared/uploads_async.py
jalauzon-msft: After removing `__len__`, I was able to just save `self.iterator`

sdk/storage/azure-storage-blob/azure/storage/blob/_shared/uploads_async.py
jalauzon-msft: That's a good question. This is the same exact logic as the sync version of this object which we have never had a report of issues but I'm not sure how popular generators are for uploads. Looking at the code for parallel upload that actually reads from this generator (where `get_chunk_streams()` is an async generator itself):
https://github.com/Azure/azure-sdk-for-python/blob/ec68720d53779dc97e16c39474f53fe41afe47bf/sdk/storage/azure-storage-blob/azure/storage/blob/_shared/uploads_async.py#L86-L96

https://github.com/Azure/azure-sdk-for-python/blob/ec68720d53779dc97e16c39474f53fe41afe47bf/sdk/storage/azure-storage-blob/azure/storage/blob/_shared/uploads_async.py#L22-L33

It seems we form the initial set of tasks synchronously but then will grab a new chunk after each task finishes. I'm not sure how async iterables work in terms of thread safety so I can't say for sure if this is thread safe or not. Thoughts?

jalauzon-msft: I added a test for this and ran it a number of times without issue but, I don't think that's enough to guarantee it works all the time.

jalauzon-msft: Discussed further offline and we think this should be okay.

PR #27556, opened 2022-11-17
[Monitor][Ingestion] Adjust upload error handling

pvaneck:
This uses a callback approach where a user can optionally pass in an `on_error` callback function for custom error handing. 

This callback function should expect arbitrary keyword arguments (`**kwargs`) or expect an `error` and `logs` argument.

If no `on_error` callback function is passed in, then the default behavior will be to raise the first error encountered.

A sync and async sample were added demonstrating usage.

### DIFF
--- a/sdk/monitor/azure-monitor-ingestion/azure/monitor/ingestion/_helpers.py
+++ b/sdk/monitor/azure-monitor-ingestion/azure/monitor/ingestion/_helpers.py
@@ -4,22 +4,30 @@
 # ------------------------------------
 import json
 import logging
-from typing import Generator, List
+import sys
+from typing import Any, Generator, List, Tuple
 import zlib
 
+if sys.version_info >= (3, 9):
+    from collections.abc import MutableMapping
+else:
+    from typing import MutableMapping  # type: ignore  # pylint: disable=ungrouped-imports
+
+
 _LOGGER = logging.getLogger(__name__)
+JSON = MutableMapping[str, Any]  # pylint: disable=unsubscriptable-object
 
 MAX_CHUNK_SIZE_BYTES = 1024 * 1024 # 1 MiB
 CHAR_SIZE_BYTES = 4
 
 
-def _split_chunks(logs: List) -> Generator:
+def _split_chunks(logs: List[JSON]) -> Generator[List[JSON], None, None]:
     chunk_size = 0
     curr_chunk = []
     for log in logs:
         # each char is 4 bytes
         size = len(json.dumps(log)) * CHAR_SIZE_BYTES
-        if chunk_size + size < MAX_CHUNK_SIZE_BYTES:
+        if chunk_size + size <= MAX_CHUNK_SIZE_BYTES:
             curr_chunk.append(log)
             chunk_size += size
         else:
@@ -32,7 +40,7 @@ def _split_chunks(logs: List) -> Generator:
         yield curr_chunk
 
 
-def _create_gzip_requests(logs: List) -> Generator:
+def _create_gzip_requests(logs: List[JSON]) -> Generator[Tuple[bytes, List[JSON]], None, None]:
     for chunk in _split_chunks(logs):
         zlib_mode = 16 + zlib.MAX_WBITS  # for gzip encoding
         _compress = zlib.compressobj(wbits=zlib_mode)
--- a/sdk/monitor/azure-monitor-ingestion/azure/monitor/ingestion/_helpers.py
+++ b/sdk/monitor/azure-monitor-ingestion/azure/monitor/ingestion/_helpers.py
@@ -2,40 +2,41 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
 # ------------------------------------
-
 import json
+import logging
+from typing import Generator, List
 import zlib
 
-MAX_CHUNK_SIZE = 1000000
-CHAR_SIZE = 4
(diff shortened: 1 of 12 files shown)


sdk/monitor/azure-monitor-ingestion/azure/monitor/ingestion/_helpers.py
```diff
@@ -2,40 +2,41 @@
...
-    chunks = []
+def _split_chunks(logs: List) -> Generator:
     chunk_size = 0
     curr_chunk = []
     for log in logs:
         # each char is 4 bytes
-        size = len(json.dumps(log)) * CHAR_SIZE
-        if chunk_size + size < MAX_CHUNK_SIZE:
+        size = len(json.dumps(log)) * CHAR_SIZE_BYTES
+        if chunk_size + size < MAX_CHUNK_SIZE_BYTES:
```
annatisch, line 22: Should this be `<=`?

sdk/monitor/azure-monitor-ingestion/azure/monitor/ingestion/_helpers.py
```diff
@@ -2,40 +2,41 @@
...
-CHAR_SIZE = 4
+_LOGGER = logging.getLogger(__name__)
+
+MAX_CHUNK_SIZE_BYTES = 1024 * 1024 # 1 MiB
+CHAR_SIZE_BYTES = 4
 
 
-def _split_chunks(logs):
-    chunks = []
+def _split_chunks(logs: List) -> Generator:
```
annatisch, line 16: Do we have a more comprehensive type hint for these `logs`? Presumably the public facing API uses something more concrete?

annatisch: Looks like we're using the defined `JSON` type - in which case these should also be:
`_split_chunks(logs: List[JSON]) -> Generator[JSON]`

sdk/monitor/azure-monitor-ingestion/azure/monitor/ingestion/_operations/_patch.py
```diff
@@ -25,32 +29,40 @@ def upload(  # pylint: disable=arguments-renamed, arguments-differ
         rule_id: str,
         stream_name: str,
         logs: Union[List[JSON], IO],
+        on_error: Optional[Callable[[Exception, List[JSON]], None]] = None,
```
annatisch, line 32: If this is optional it should be keyword-only.

sdk/monitor/azure-monitor-ingestion/azure/monitor/ingestion/_helpers.py
```diff
@@ -2,40 +2,41 @@
...
-    return chunks
+        _LOGGER.debug('Yielding chunk with size: %d', chunk_size)
+        yield curr_chunk
 
 
-def _create_gzip_requests(logs):
-    requests = []
-    chunks = _split_chunks(logs)
-    for chunk in chunks:
+def _create_gzip_requests(logs: List) -> Generator:
```
annatisch, line 35: The type hint in the public `upload` method states that we accept `IO` as an input type for logs?
This would probably work if the IO data is structured with one log entry per line because I think a file IO will iterate lines by default...
Do we have tests for this? Does it work with any unstructured IO data passed in?

sdk/monitor/azure-monitor-ingestion/samples/sample_send_small_logs.py
```diff
@@ -25,9 +25,4 @@
       }
     ]
 
-response = client.upload(rule_id=rule_id, stream_name=os.environ['LOGS_DCR_STREAM_NAME'], logs=body)
-
-## iterates directly over Tuple[HttpResponseError, JSON]:
-for error, failed_logs in response:
-    # prints nothing if there are no failed_logs (i/e the status is success)
-    print(failed_logs, error)
+client.upload(rule_id=rule_id, stream_name=os.environ['LOGS_DCR_STREAM_NAME'], logs=body)
```
annatisch, line 28: If this is a sample - we should demonstrate correct error handling. Now that errors are no longer handled by iterating the result, and no callback is passed in, we should use something like:
```python
try:
    client.upload(...)
except HttpResponseError as e:
    print(f"Upload failed: {e})
```

sdk/monitor/azure-monitor-ingestion/azure/monitor/ingestion/_helpers.py
pvaneck: I think the `IO` type was prematurely added during the transition just before I took over, and I don't think it was tested. Next on my list is ensure file streams can be handled, so I will remove the `IO` typing for now and then re-add it once I've verified and added io tests for upload in a subsequent PR. Hopefully that works.

sdk/monitor/azure-monitor-ingestion/samples/sample_send_small_logs.py
pvaneck: Good call. Updated all examples where an error could be raised.

sdk/monitor/azure-monitor-ingestion/azure/monitor/ingestion/_helpers.py
pvaneck: I believe it indeed should be!

annatisch (approved):

PR #27559, opened 2022-11-17
p0 component deployment creation

TajiHarrisMicrosoft:
# Description

== PRIVATE PREVIEW FEATURE ==
Creating a batch deployment of component type with a component attribute that references a created component or a component yaml

Examples of execution commands:
`az ml batch-deployment create -f sdk/ml/azure-ai-ml/tests/test_configs/deployments/batch/batch_component_deployment.yml `

Component property under job_definition could have two different formats:
- component YAML 
   In case we get a component YAML we need to create an anonymous component associated to it.
- azureml:<name_component>:<version>
   In this case, we do a GET call to get the component id. If the provided value does not have a valid format it will throw a `validationSchemaException`

These changes also cover `GET` scenario: 
`az ml batch-deployment show -e <endpointname> -n <deploymentName>`
If users have private preview enabled and response has component_deployment properties, we will populate `job_definition` via `_from_rest_object` function


sdk/ml/azure-ai-ml/azure/ai/ml/constants/_common.py
```diff
@@ -577,3 +577,8 @@ class IdentityType:
     AML_TOKEN = "aml_token"
     USER_IDENTITY = "user_identity"
     MANAGED_IDENTITY = "managed_identity"
+
+class JobTypes:
```
banibrata, line 581: Consider reusing from here https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/ml/azure-ai-ml/azure/ai/ml/constants/_job/job.py#:~:text=class%20JobType(,%3D%20%22spark%22

sdk/ml/azure-ai-ml/azure/ai/ml/operations/_batch_deployment_operations.py
```diff
@@ -237,3 +252,46 @@ def _get_workspace_location(self) -> str:
...
+        elif isinstance(deployment.job_definition.component, str):
+            split_name = deployment.job_definition.component.split(':')
+            if len(split_name) == 2:
+                component = self._component_operations.get(
+                    name = split_name[0],
+                    resource_group_name=self._resource_group_name,
+                    workspace_name=self._workspace_name,
+                    version= split_name[1],
+                    **self._init_kwargs
+                )
```
banibrata, line 286: Not sure why are we making a rest call to resolve this, we should comply with expanding the short form and not making Get() call. have you considered using this https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/ml/azure-ai-ml/azure/ai/ml/operations/_operation_orchestrator.py#:~:text=def%20resolve_azureml_id(self%2C%20arm_id%3A%20str%20%3D%20None%2C%20**kwargs)%20%2D%3E%20str%3A?

sdk/ml/azure-ai-ml/tests/test_configs/deployments/batch/batch_component_deployment.yml
```diff
@@ -0,0 +1,12 @@
+$schema: http://azureml/sdk-2-0/BatchDeployment.json
+name: component-deployment
+description: A sample deployment with pre and post processing done before and after inference.
+endpoint_name: component-endpoint
+compute: azureml:cpu-cluster
+type: component
+job_definition:
+    type: pipeline
+    component: ../../components/component_e2e.yml
+    # component: azureml:atestcomponent:1
```
banibrata, line 10: It is worth adding another test yaml with this short form!

sdk/ml/azure-ai-ml/azure/ai/ml/_schema/_deployment/batch/run_settings_schema.py
```diff
@@ -0,0 +1,29 @@
...
+    name = fields.Str()
+    display_name = fields.Str()
+    experiment_name = fields.Str()
+    description = fields.Str()
+    tags = fields.Dict()
+    settings= fields.Dict()
+
+    @post_load
+    def make(self, data: Any, **kwargs: Any) -> Any:  # pylint: disable=unused-argument
+        from azure.ai.ml.entities._deployment.run_settings import RunSettings
```
Man-MSFT, line 27: non-blocking, I thought that we did a Python pep8/black/some kind of lint pass before we moved to GitHub, where one of the rules was that imports had to be top of file?

sdk/ml/azure-ai-ml/azure/ai/ml/constants/_common.py
Man-MSFT: We should 100% reuse that class. If we are trying to say "component deployment creation only supports these X job types" then we can have a separate array or Set of those specific job types instead

sdk/ml/azure-ai-ml/azure/ai/ml/entities/_deployment/batch_deployment.py
```diff
@@ -110,6 +112,9 @@ def __init__(
         **kwargs,
     ) -> None:
 
+        self.deployment_type = kwargs.pop("type", "Model")
```
Man-MSFT, line 115: We spec'd it s.t. we default to Model?

sdk/ml/azure-ai-ml/azure/ai/ml/entities/_deployment/batch_deployment.py
```diff
@@ -212,6 +217,14 @@ def _to_rest_object(self, location: str) -> BatchDeploymentData:  # pylint: disa
             properties=self.properties,
         )
 
+        if is_private_preview_enabled() and self.job_definition:
+            non_flat_data = {}
+            non_flat_data["component_deployment"] = self.job_definition._to_dict()
```
Man-MSFT, line 222: Should this be "job_definition" over component_deployment? Or are these synonymous terms?

sdk/ml/azure-ai-ml/azure/ai/ml/entities/_deployment/batch_deployment.py
```diff
@@ -212,6 +217,14 @@ def _to_rest_object(self, location: str) -> BatchDeploymentData:  # pylint: disa
...
+            flat_data = flatten(non_flat_data, ".")
+            flat_data_keys = flat_data.keys()
+            for k in flat_data_keys:
```
Man-MSFT, line 225: So we append and overwrite `self.properties` but if there are other things inside `self.properties`, we're ok to keep those things?
Or, should we just do `self.properties = flat_data` ?

sdk/ml/azure-ai-ml/azure/ai/ml/entities/_deployment/batch_deployment.py
```diff
@@ -248,6 +261,21 @@ def _from_rest_object(cls, deployment: BatchDeploymentData):  # pylint: disable=
             properties=deployment.properties.properties,
         )
 
+        # Job definition is in private preview. If private preview environment is
+        # not enable we need to remove job definition from properties.
+        if not is_private_preview_enabled():
+            del_key = []
+            for k in deployment.properties:
+                if k.startswith("component_deployment"):
```
Man-MSFT, line 269: Regardless of the key name (one of my other comments) please make this a constant

sdk/ml/azure-ai-ml/azure/ai/ml/entities/_deployment/batch_deployment.py
Man-MSFT: I think it must be append+overwrite but not complete overwrite based on the `is_private_preview_enabled` flag if I'm not mistaken?

sdk/ml/azure-ai-ml/azure/ai/ml/entities/_deployment/batch_deployment.py
```diff
@@ -248,6 +261,21 @@ def _from_rest_object(cls, deployment: BatchDeploymentData):  # pylint: disable=
             properties=deployment.properties.properties,
         )
 
+        # Job definition is in private preview. If private preview environment is
+        # not enable we need to remove job definition from properties.
+        if not is_private_preview_enabled():
+            del_key = []
```
Man-MSFT, line 267:
```suggestion
            deployment.properties = {key : val for key, val in deployment.properties.items() if not key.startswith("component_deployment")}
```

sdk/ml/azure-ai-ml/azure/ai/ml/entities/_deployment/batch_deployment.py
```diff
@@ -248,6 +261,21 @@ def _from_rest_object(cls, deployment: BatchDeploymentData):  # pylint: disable=
...
+            for k in deployment.properties:
+                if k.startswith("component_deployment"):
+                    del_key.append(k)
+            if len(del_key) > 0:
+                for k in del_key:
+                    del deployment.properties[k]
+        else:
+            unflat_data = unflatten(deployment.properties, ".")
```
Man-MSFT, line 275: Will always return non-None value?

sdk/ml/azure-ai-ml/azure/ai/ml/entities/_deployment/job_definition.py
```diff
@@ -0,0 +1,52 @@
...
+from azure.ai.ml._schema._deployment.batch.job_definition_schema import JobDefinitionSchema
+from azure.ai.ml.constants._common import BASE_PATH_CONTEXT_KEY
+from azure.ai.ml.entities._component.component import Component
+
+
+class JobDefinition:
+    """Job Definition entity
+
+    :param job: Job definition
+    :type job: Union[Component, str]
```
Man-MSFT, line 16: It's a bit weird to me that the class is called `JobDefinition` and we have a param called "job" described as also "Job definition" but it's of a different class `Component`.
How can we simplify these concepts (arg name or description changes? I assume the type is correct?)

sdk/ml/azure-ai-ml/azure/ai/ml/entities/_deployment/job_definition.py
```diff
@@ -0,0 +1,52 @@
...
+    :param component: Component definition
+    :type component: Union[Component, str]
+    :param type: Job definition type
+    :type type: str
```
Man-MSFT, line 20: Job def type is an arbitrary str? We don't have an enum to use?

sdk/ml/azure-ai-ml/azure/ai/ml/entities/_deployment/job_definition.py
```diff
@@ -0,0 +1,52 @@
...
+    :type tags: Dict[str, Any]
+    """
+
+    def __init__(
+        self,
+        job: Union[Component, str] = None,
+        component : Union[Component, str] = None,
+        type: str = None, # pylint: disable=redefined-builtin
+        settings: Dict[str, Any] = None,
+        name: str = None,
```
Man-MSFT, line 37: Why are all these optional values? I would have expected at least `name` and `type` to be required

Man-MSFT, line 37: nit: I think most of the other entities have name first?

sdk/ml/azure-ai-ml/azure/ai/ml/operations/_batch_deployment_operations.py
```diff
@@ -106,7 +112,16 @@ def begin_create_or_update(
             operation_config=self._operation_config,
         )
         upload_dependencies(deployment, orchestrators)
-
+        if is_private_preview_enabled() and deployment.job_definition:
+            component = self._validate_component(deployment)
```
Man-MSFT, line 116: Is this right? We validate and don't assign the `component` we validated?

sdk/ml/azure-ai-ml/azure/ai/ml/operations/_batch_deployment_operations.py
Man-MSFT: Yes, any reason to go against the existing flow? Else let's keep with paradigm that client does light validation (URI formats are reasonable) but let service handle existence checks

sdk/ml/azure-ai-ml/azure/ai/ml/operations/_batch_deployment_operations.py
```diff
@@ -237,3 +252,46 @@ def _get_workspace_location(self) -> str:
...
+        for batch deployment
+        :param deployment: Batch deployment
+        :type deployment: ~azure.ai.ml.entities._deployment.deployment.Deployment
+        :raises ~azure.ai.ml.exceptions.ValidationException: Raised if BatchDeployment cannot be
+            successfully validated. Details will be provided in the error message.
+        :return: Component entity associated to the values provided in yaml file
+        :rtype: ~azure.ai.ml.entities.Component
+        """
+        component = None
+        if isinstance(deployment.job_definition.component, PipelineComponent):
```
Man-MSFT, line 268: If object is component -> make a POST/PUT api call and create
else if str -> only fetch or throw

I'm not sure how the rest of our SDK behaves but it's non-obvious to me that the type of the object I'm passing will dictate what's going to happen. I would have thought that regardless, by the end of this function call my expectations are same. That this thing is guaranteed to exist (assuming I passed valid data). 
If we want this behavior split, should we have a "create_if_not_exist" flag?

sdk/ml/azure-ai-ml/azure/ai/ml/_schema/_deployment/batch/run_settings_schema.py
nancy-mejia: We don't have specific rules. But during the initial migration during reviews the "rule" was to add it to the specific line.

sdk/ml/azure-ai-ml/azure/ai/ml/entities/_deployment/batch_deployment.py
nancy-mejia: I'm not sure about this. Taji worked on these changes according to the spec. So, it most likely is documented but if not we can document it too.

sdk/ml/azure-ai-ml/azure/ai/ml/operations/_batch_deployment_operations.py
nancy-mejia: we just need to assign the component_id value

sdk/ml/azure-ai-ml/azure/ai/ml/entities/_deployment/job_definition.py
nancy-mejia: Reordered them, to have name at the beginning

sdk/ml/azure-ai-ml/azure/ai/ml/entities/_deployment/job_definition.py
nancy-mejia: Yes, we validate that this str is valid in job_definition schema. For now, the only valid value is `pipeline`

sdk/ml/azure-ai-ml/azure/ai/ml/entities/_deployment/job_definition.py
nancy-mejia: Yes, that was wrong. The right type was `Job`. Thanks for catching that

sdk/ml/azure-ai-ml/azure/ai/ml/entities/_deployment/job_definition.py
nancy-mejia: Because users can specify existing jobs or components, name is not required. But I updated `type` to be required

sdk/ml/azure-ai-ml/azure/ai/ml/operations/_batch_deployment_operations.py
nancy-mejia: This behavior is the one specified in the spec. 
Because this is coming from a CLI request we will load the properties from the YAML file provided.
So, depending on those values we will create a "PipelineComponent" entity, or we will keep the `str` if it starts with `azureml:` prefix. 

I would suggest not adding a flag because we do have similar behavior for other scenarios using:
`upload_dependencies` function

https://github.com/Azure/azure-sdk-for-python/blob/97e1e7b0f5e52af09ea83721aecfe98816e9f4b0/sdk/ml/azure-ai-ml/azure/ai/ml/_utils/_endpoint_utils.py#L130

sdk/ml/azure-ai-ml/tests/test_configs/deployments/batch/batch_component_deployment.yml
nancy-mejia: Will add once backend changes are ready

sdk/ml/azure-ai-ml/azure/ai/ml/entities/_deployment/batch_deployment.py
TajiHarrisMicrosoft: That is correct, if private preview is enabled we will append the flattened component deployment values that are in flat_data

banibrata (approved):

sdk/ml/azure-ai-ml/azure/ai/ml/operations/_batch_deployment_operations.py
Man-MSFT: Mmm, I still think that that is a bit non-obvious, but if there are other instances of this pattern then at least we are consistent. And this is a `_` prefix function so at least it's not customer facing.

sdk/ml/azure-ai-ml/azure/ai/ml/entities/_deployment/batch_deployment.py
```diff
@@ -223,7 +236,7 @@ def _from_rest_object(cls, deployment: BatchDeploymentData):  # pylint: disable=
             if deployment.properties.code_configuration
             else None
         )
-        return BatchDeployment(
+        deployment =  BatchDeployment(
```
Man-MSFT, line 239:
```suggestion
        deployment = BatchDeployment(
```
nit: extra space. Which I hope if we have better repo or package level lint rules that would enforce this.

Man-MSFT (approved):

PR #27587, opened 2022-11-18
[ML][Pipelines] Fix NodeOutput bool test in download

zhengfeiwang:
# Description

Use `is not None` to check if a `NodeOutput` object is `None` to avoid banned bool test for expression.

### DIFF
--- a/sdk/ml/azure-ai-ml/azure/ai/ml/operations/_job_operations.py
+++ b/sdk/ml/azure-ai-ml/azure/ai/ml/operations/_job_operations.py
@@ -769,7 +769,7 @@ class JobOperations(_ScopeDependentOperations):
             try:
                 job = self.get(job_name)
                 artifact_store_uri = job.outputs[DEFAULT_ARTIFACT_STORE_OUTPUT_NAME]
-                if artifact_store_uri and artifact_store_uri.path:
+                if artifact_store_uri is not None and artifact_store_uri.path:
                     outputs[DEFAULT_ARTIFACT_STORE_OUTPUT_NAME] = artifact_store_uri.path
             except (AttributeError, KeyError):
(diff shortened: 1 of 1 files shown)


sdk/ml/azure-ai-ml/azure/ai/ml/operations/_job_operations.py
```diff
@@ -769,7 +769,7 @@ def _get_named_output_uri(
             try:
                 job = self.get(job_name)
                 artifact_store_uri = job.outputs[DEFAULT_ARTIFACT_STORE_OUTPUT_NAME]
-                if artifact_store_uri and artifact_store_uri.path:
+                if artifact_store_uri is not None and artifact_store_uri.path:
```
wangchao1230, line 772: As download test is only running in live mode, so it's not early detect such error?

D-W-, line 772: `artifact_store_uri` got recognized as expression op because standalone job's outputs becomes `NodeOutput`, right?

D-W-: Maybe we can also investigate if we can just use `Output` instead of `NodeOutput` in standalone job.

zhengfeiwang: No, existing download test will not cover this logic. Need some investigation on how to cover this from test.

zhengfeiwang: Yes, and I think it's worthy to do so as it can avoid some unexpected behavior and error

wangchao1230 (approved):

PR #27602, opened 2022-11-18
os error to aiohttp oserror for perf fix

l0lawrence:
Excepting on just OSError caused issues with perf testing

### DIFF
--- a/sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_transport_async.py
+++ b/sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_transport_async.py
@@ -485,22 +485,14 @@ class WebSocketTransportAsync(
             password = self._http_proxy.get("password", None)
 
         try:
-            from aiohttp import ClientSession, ClientConnectorError
+            from aiohttp import ClientSession, ClientConnectorError, ClientOSError
             from urllib.parse import urlsplit
 
             if username or password:
                 from aiohttp import BasicAuth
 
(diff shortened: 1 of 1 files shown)


sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_transport_async.py
```diff
@@ -520,25 +520,24 @@ async def connect(self):
                     ssl=self.sslopts,
                     heartbeat=DEFAULT_WEBSOCKET_HEARTBEAT_SECONDS,
                 )
-            except ClientConnectorError as exc:
+            except ClientConnectionError as exc:
```
annatisch, line 523: What's the difference between the Connection and the Connector errors? Do we want to handle them both?

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_transport_async.py
```diff
@@ -558,7 +557,7 @@ async def _read(self, n, buffer=None, **kwargs):  # pylint: disable=unused-argum
             return view
         except asyncio.TimeoutError as te:
             raise ConnectionError('Receive timed out (%s)' % te)
-        except OSError as e:
```
annatisch, line 561: So we are happy for OSErrors to be propagated from here rather than caught? I'm guessing this will be caught higher up in the `read()` or `receive_frame()` methods?

PR #27630, opened 2022-11-21
[Storage] Migrate `test_handle` to test proxy

jalauzon-msft:
This change migrates `test_handle` and `test_handle_async` to test proxy.

These tests require manual setup of SMB handles and therefore are all playback only. A sanitizer for client IP for the list handles responses was also added.


vincenttran-msft (approved):
LGTM! Thanks for getting to this 😄

PR #27631, opened 2022-11-22
[Storage] Pin `azure-mgmt-storage` dev requirement to fix ci

jalauzon-msft:
Storage ci broke because the recent release of `azure-mgmt-storage` added a dep on a `typing-extensions` version that is higher than the required req for Storage packages. Since we only use this package for test, pinning it to a previous version to fix this break and avoid future breaks.


vincenttran-msft (approved):
Makes sense  😄

PR #27660, opened 2022-11-23
feat: validate component for node

elliotzh:
# Description

Always serialize LocalPathField into an absolute path to make dumped object valid for LocalPathField validation.
Remove component from skip_field_in_schema_validation for BaseNode.

Background:
We use LocalPathField to validate if a value is an existed file or directory, like:
```yaml
code: ./src
```
We should raise validation error if and only if "./src" is not an existed file or directory based on the correct base path.

In current implementation of validate and _to_dict, we use schema to serialize the path aware resources into a dict first, like component and environment, which makes it possible to reuse existed function of schema to do serialization and validation.
 
Then we will meet a problem that the base path information will be dropped after serialization. So in this PR, we maintain base_path in context during dump and make LocalPathField always serialize value into an absolute path in string.

After resolving this problem, we can remove component from skip_field_in_schema_validation for BaseNode


wangchao1230:
Add more description on what the PR does

sdk/ml/azure-ai-ml/azure/ai/ml/_schema/core/fields.py
```diff
@@ -109,17 +109,27 @@ def _serialize(self, value, attr, obj, **kwargs) -> typing.Optional[str]:
...
-        return super(LocalPathField, self)._serialize(value, attr, obj, **kwargs)
+        # always dump path as absolute path as base_path will be dropped after serialization
+        return super(LocalPathField, self)._serialize(
+            self.__resolve_path(value).as_posix(),
+            attr,
+            obj,
+            **kwargs
+        )
+
+    def __resolve_path(self, value):
```
wangchao1230, line 120: why  __ not _ here?

wangchao1230: Add some docstrings

sdk/ml/azure-ai-ml/azure/ai/ml/_schema/core/fields.py
```diff
@@ -109,17 +109,27 @@ def _serialize(self, value, attr, obj, **kwargs) -> typing.Optional[str]:
         if value is None:
             return None
         self._validate(value)
-        return super(LocalPathField, self)._serialize(value, attr, obj, **kwargs)
+        # always dump path as absolute path as base_path will be dropped after serialization
+        return super(LocalPathField, self)._serialize(
+            self.__resolve_path(value).as_posix(),
```
wangchao1230, line 114: why additional as_posix() here?

sdk/ml/azure-ai-ml/azure/ai/ml/_schema/pipeline/component_job.py
```diff
@@ -124,6 +123,8 @@ class CommandSchema(BaseNodeSchema, ParameterizedCommandSchema):
         ],
         required=True,
     )
+    # code is directly linked to component.code, so no need to validate it
+    code = fields.Str(allow_none=True)
```
wangchao1230, line 127: This should be load only?

sdk/ml/azure-ai-ml/azure/ai/ml/_schema/core/fields.py
elliotzh: value filled in LocalPathField is a string instead of a Path object, so return string in serialization

sdk/ml/azure-ai-ml/azure/ai/ml/_schema/pipeline/component_job.py
elliotzh: Load only make sense. Here I just follow the original logic. Let me make a change.

sdk/ml/azure-ai-ml/azure/ai/ml/_schema/core/fields.py
elliotzh: Also serialization result should be a string instead of a Path in dumped dict

sdk/ml/azure-ai-ml/azure/ai/ml/_schema/core/fields.py
elliotzh: updated. Want to make it private in the beginning, but seems not necessary

wangchao1230 (approved):

PR #27667, opened 2022-11-23
[Communication] - azure-communication-phonenumbers - Update changelog with polling interval changes

sorrego-msft:
Update the changelog to reflect the previously made changes to the polling_interval

### DIFF
--- a/sdk/communication/azure-communication-phonenumbers/CHANGELOG.md
+++ b/sdk/communication/azure-communication-phonenumbers/CHANGELOG.md
@@ -5,10 +5,10 @@
 ### Features Added
 - Added new SIP routing client for handling Direct routing numbers.
 - Added the ability specify the API version by an optional `api_version` keyword parameter.
-- Updated default polling interval, it can still be overwritten by passing it as an argument.
 
 ### Other Changes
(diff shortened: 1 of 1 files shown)


sdk/communication/azure-communication-phonenumbers/CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### Features Added
 - Added new SIP routing client for handling Direct routing numbers.
 - Added the ability specify the API version by an optional `api_version` keyword parameter.
+- Updated default polling interval, it can still be overwritten by passing it as an argument.
```
danielortega-msft, line 8: A few suggestions:
- This is not necessarily a feature, so I would suggest adding it as a bullet point under Other Changes (also, it would be great if you change the current paragraph regarding Python 3.6 deprecation to be a bullet point as well)
- Specify which is the new default interval, so customers are able to see it in the release notes without the need to see the implementation
- Specify the name of the argument that can be used to overwrite it and make sure to clarify that it is a keyword argument

danielortega-msft (changes requested):

danielortega-msft (approved):

PR #27675, opened 2022-11-23
[ACR] Change audience to be optional

YalinLi0312:
Resolves https://github.com/Azure/azure-sdk-for-python/issues/22229


sdk/containerregistry/azure-containerregistry/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 ### Bugs Fixed
 
 ### Other Changes
+* Change `audience` to an optional parameter in `ContainerRegistryClient`, which the default value is `"https://management.azure.com"`. ([#27675](https://github.com/Azure/azure-sdk-for-python/pull/27675))
```
annatisch, line 12: I wouldn't put the pull request in the changelog - though you could put the issue number.

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_registry_client.py
```diff
@@ -43,24 +44,23 @@ def _return_response(pipeline_response, deserialized, response_headers):
 
 
 class ContainerRegistryClient(ContainerRegistryBaseClient):
-    def __init__(self, endpoint, credential=None, **kwargs):
-        # type: (str, Optional[TokenCredential], **Any) -> None
+    def __init__(self, endpoint, credential=None, audience=AZURE_RESOURCE_MANAGER_PUBLIC_CLOUD, **kwargs):
+        # type: (str, Optional[TokenCredential], Optional[str], **Any) -> None
```
annatisch, line 48: This parameter should remain keyword-only.

sdk/containerregistry/azure-containerregistry/azure/containerregistry/aio/_async_container_registry_client.py
```diff
@@ -28,24 +28,27 @@
 
 class ContainerRegistryClient(ContainerRegistryBaseClient):
     def __init__(
-        self, endpoint: str, credential: "Optional['AsyncTokenCredential']" = None, *, audience: str, **kwargs: "Any"
+        self,
+        endpoint: str,
+        credential: "Optional['AsyncTokenCredential']" = None,
+        audience: Optional[str] = AZURE_RESOURCE_MANAGER_PUBLIC_CLOUD,
```
annatisch, line 34: This should also remain keyword-only, and type is `str`.

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_registry_client.py
annatisch: Also, the type is not `Optional[str]` because `None` is not a valid value. Just `str`.

annatisch: While we're here - could we also update the constructor method signature to use Python3 syntax?

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_helpers.py
```diff
@@ -30,6 +30,11 @@
 ]
 OCI_MANIFEST_MEDIA_TYPE = "application/vnd.oci.image.manifest.v1+json"
 
+# Supported audiences
+AZURE_RESOURCE_MANAGER_CHINA = "https://management.chinacloudapi.cn"
+AZURE_RESOURCE_MANAGER_GOVERNMENT = "https://management.usgovcloudapi.net"
```
annatisch, line 35: We do we have these other two constants? We don't need them right?

YalinLi0312: They are used in tests to verify when audience doesn't match the cloud that the resources are in will fail.

sdk/containerregistry/azure-containerregistry/CHANGELOG.md
YalinLi0312: Sounds good, I'll make the change.

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_helpers.py
YalinLi0312: Has removed the constants, change to retrieve them directly from authorities.

sdk/containerregistry/azure-containerregistry/azure/containerregistry/aio/_async_container_registry_client.py
YalinLi0312: Since it's already in `kwargs`, it's not necessary to move to `params`. Other than that, I don't see the benefit of using audience in keyword here. We can discuss it offline.

sdk/containerregistry/azure-containerregistry/azure/containerregistry/aio/_async_container_registry_client.py
```diff
@@ -15,6 +15,7 @@
 )
 from azure.core.tracing.decorator import distributed_trace
 from azure.core.tracing.decorator_async import distributed_trace_async
+from msrestazure.azure_cloud import AZURE_PUBLIC_CLOUD
```
annatisch, line 18: We shouldn't have a dependency on msrestazure?

sdk/containerregistry/azure-containerregistry/azure/containerregistry/aio/_async_container_registry_client.py
```diff
@@ -28,7 +29,12 @@
 
 class ContainerRegistryClient(ContainerRegistryBaseClient):
     def __init__(
-        self, endpoint: str, credential: "Optional['AsyncTokenCredential']" = None, *, audience: str, **kwargs: "Any"
+        self,
+        endpoint: str,
+        credential: "Optional['AsyncTokenCredential']"=None,
```
annatisch, line 34:
```suggestion
        credential: Optional["AsyncTokenCredential"] = None,
```
I think we only need quotes around the credential type.

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_registry_client.py
```diff
@@ -43,8 +44,12 @@ def _return_response(pipeline_response, deserialized, response_headers):
 
 
 class ContainerRegistryClient(ContainerRegistryBaseClient):
-    def __init__(self, endpoint, credential=None, **kwargs):
-        # type: (str, Optional[TokenCredential], **Any) -> None
+    def __init__(
+        self,
+        endpoint: str,
+        credential: "Optional[TokenCredential]"=None,
```
annatisch, line 50:
```suggestion
        credential: Optional["TokenCredential"] = None,
```
I think we only need quotes around the credential type.

sdk/containerregistry/azure-containerregistry/azure/containerregistry/aio/_async_container_registry_client.py
annatisch: We should just define a constant for this.

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_registry_client.py
YalinLi0312: Updated.

sdk/containerregistry/azure-containerregistry/azure/containerregistry/aio/_async_container_registry_client.py
YalinLi0312: Updated.

sdk/containerregistry/azure-containerregistry/azure/containerregistry/aio/_async_container_registry_client.py
YalinLi0312: Have removed the usage of this package in testcase.py, so it's no longer a required dependency in dev.

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_helpers.py
YalinLi0312: Keep these other two constants - 1. convenient for customers to use, 2. used in testcase.py

sdk/containerregistry/azure-containerregistry/azure/containerregistry/aio/_async_container_registry_client.py
YalinLi0312: It's good to learn the reason behind the general rule to use positional parameters and keyword arguments, thanks Anna! :)

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_helpers.py
annatisch: None of these are currently exposed publicly, so customers can't use them anyway.
I'm also not sure we want to expose them publicly - are the other language SDKs doing so? And if so, what are they called?

YalinLi0312: The audience that customer provided should match the cloud, so these values will be used when customer is using China cloud or Us Gov cloud. We are using the same name as in Java: https://github.com/Azure/azure-sdk-for-java/blob/main/sdk/containerregistry/azure-containers-containerregistry/src/main/java/com/azure/containers/containerregistry/models/ContainerRegistryAudience.java#L14 cc: @lmolkova

annatisch: Have those names been through archboard review? I believe we have terminology for these in other SDKs and I'd prefer we have consistent naming. Alternatively, we can just not make these public for now - the strings themselves are already documented in the docstring.

YalinLi0312: Got response from ACR v-team that we are figuring out the audience story with ACR folks. I removed these public constants for now.

annatisch (approved):

PR #27706, opened 2022-11-28
[Storage] Fix flaky upload tests

jalauzon-msft:
Attempting to fix some tests that occasionally fail in live test pipeline when trying to talk with external site. Changed the tests to upload and use a sperate blob rather than downloading from external site.


vincenttran-msft (approved):
Thanks for getting to this!

PR #27707, opened 2022-11-28
[Storage] Cleanup old test classes after test proxy migration

jalauzon-msft:
Now that all Storage tests have been migrated to test proxy, we can remove the old VCR test base classes that were replaced with new test proxy versions. Additionally cleaning up some other functions that are no longer used.


vincenttran-msft (approved):
Be gone test proxy migration!

mccoyp (approved):
I think there's another pesky import of `AsyncStorageTestCase` in `/storage/aio/__init__.py` that we'll also need to remove, but otherwise should be good to go! It's so satisfying to see this removed -- thank you so much for all the work on the migration!!

Issue #27640, opened 2022-11-22
[Storage] Fix changefeed sharding tests


vincenttran-msft:
This issue tracks that more work is remaining to be done for `test_changefeed.py`. Since there was some changes to the default behavior, for now #27546 is merged, but pending work is required for the remaining 5 tests that involve sharding.

PR #27709, opened 2022-11-28
[Storage] Fix shard tests for changefeed

vincenttran-msft:
This PR adds in recordings for the 5 tests in `test_changefeed.py` that require sharding to be enabled. Below is some general tips and the process used to get the manual setup right for the last 5 tests:

**My exact procedure:**
•	Upload 20 files
•	Run 'test_read_change_feed_tail_where_only_1_shard_has_data'✔️
•	Run 'test_read_change_feed_with_3_shards_in_a_time_range'✔️

•	Upload 75 more files
•	Run 'test_list_3_shards_events_works_with_1_shard_cursor'✔️

•	Upload 300 more files
•	Run 'test_read_3_shards_change_feed_during_a_time_range_in_multiple_times_gives_same_result_as_reading_all'✔️

•	Upload 800 more files
•	Upload 750 more files
•	Run 'test_read_change_feed_tail_where_3_shards_have_data' with debugger (to know when we are at the sleep statements)
o	----->At first sleep, upload 120 more files
o	----->At second sleep, upload 200 more files✔️
**NOTE: Although unclear if it is necessary, these uploads may/may not have been spread across different 15 minute blocks / hours (due to general testing + waiting for changes to propagate)**

**General Tips:**

- This is what the path structure is: `// log/<shardIndex>/<YYYY>/<MM>/<DD>/<HH>/<MM>/<PartSequenceIndex>.avro`
- Check `start_time` and `end_time` cover your operations (note that due to timezone differences, the folders may be segmented at a different hour than you may expect)
- Read over the test and look at the asserts to get an idea of how many events need to be generated
- Check the contents of the `log` folder to make sure your events have a changefeed generated (may take some time for the events to be logged)

In general, the key thing to note is that some tests will require more events than others (i.e. those with continuation tokens will likely need more events so that a continuation token is necessary), and especially for something like `test_read_change_feed_tail_where_3_shards_have_data`, this setup is highly manual and requires more events after some events are already processed etc.

Addresses #27640


jalauzon-msft (approved):
Thanks for getting right back on this and getting it to work! Great work :)

PR #27723, opened 2022-11-29
[Storage] Use `tempfile` for datalake package

vincenttran-msft:
This PR starts the work for using `tempfile` across all packages for file IO operations in an effort to reduce the failures in CI related to file IO operations.

### DIFF
--- a/sdk/storage/azure-storage-file-datalake/tests/test_file.py
+++ b/sdk/storage/azure-storage-file-datalake/tests/test_file.py
@@ -3,7 +3,6 @@
 # Licensed under the MIT License. See License.txt in the project root for
 # license information.
 # --------------------------------------------------------------------------
-import tempfile
 import unittest
 from datetime import datetime, timedelta
 from math import ceil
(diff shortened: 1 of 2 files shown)


sdk/storage/azure-storage-file-datalake/tests/test_file.py
```diff
@@ -749,15 +749,17 @@ def test_read_file_into_file(self, **kwargs):
         file_client.append_data(data, 0, len(data))
         file_client.flush_data(len(data))
 
+        temp_file = tempfile.NamedTemporaryFile(delete=False)
```
vincenttran-msft, line 752: Looking for some feedback here. After testing, I think that `NamedTemporaryFile` matches our use case most closely here.
We can't just use a `TemporaryFile` (not Named) because that type of temporary file doesn't guarantee a visible name in the file system (can cause issues on different OSes) and so that means it cannot be accessed outside of the initial `with` context.
Furthermore, we need `delete=False` so that when we do exit the first `with`, it is not automatically garbage collected. Therefore, we need to delete it manually (close & unlink).

The only other alternative approach I see that `tempfile` facilitates is:

Shove everything into a single context manager (i.e. do the write and read in one context manager, therefore we can just use a vanilla `TemporaryFile`)

Let me know your thoughts 😄

vincenttran-msft: Resolved offline -- lets move to one context manager and vanilla `TemporaryFile`

jalauzon-msft (approved):

Issue #27645, opened 2022-11-22
azure.identity.DefaultAzureCredential.get_token docs do not explain scopes arg.


mhangaard:
What are `scopes`?

`azure.identity.DefaultAzureCredential` and other Credential classes have the method `get_token` with the arguments `scopes` and `tenant_id`, both of type str. It is not precisely explained what these arguments are or how they may be obtained.

It would be valuable to have a precise description here in the documentation or an example.
---
#### Document Details

⚠ *Do not edit this section. It is required for learn.microsoft.com ➟ GitHub issue linking.*

* ID: ca1bf974-4052-598a-7197-f052ee3985ed
* Version Independent ID: 948fc19f-f68d-258d-53b1-390cb34969f8
* Content: [azure.identity.DefaultAzureCredential class](https://learn.microsoft.com/en-us/python/api/azure-identity/azure.identity.defaultazurecredential?view=azure-python)
* Content Source: [docs-ref-autogen/azure-identity/azure.identity.DefaultAzureCredential.yml](https://github.com/MicrosoftDocs/azure-docs-sdk-python/blob/main/docs-ref-autogen/azure-identity/azure.identity.DefaultAzureCredential.yml)
* GitHub Login: @lmazuel
* Microsoft Alias: **lmazuel**

xiangyan99:
@mhangaard Thanks for your feedback.

You can get the introduction of scopes at https://learn.microsoft.com/azure/active-directory/develop/scopes-oidc.

As you can see, Microsoft Identity platform is a big topic and unfortunately, we cannot include all the information in our code doc.

In terms of how to get the scope string, it varies based on different azure services you use.

If you use our Azure SDK, you don't need to be aware of the scope stuff. All are handled by our Azure SDK automatically.

stack-mriso:
Can you include the link you just sent to the scopes section in the documentation at least?

xiangyan99:
Thanks for the suggestions.

It will be available in next release.

PR #27768, opened 2022-11-30
add doc link for scopes

xiangyan99:
#27645


pvaneck (commented):
Nit: perhaps remove the "please" in the text based on the [MS writing style guide](https://styleguides.azurewebsites.net/StyleGuide/Read?id=2700&topicid=28832).

PR #27806, opened 2022-12-02
Resolve Failing Storage Build

scbedd:
azure-mgmt-storagesync -> azure-mgmt-core -> azure-core.

Required version of azure-mgmt-core needs newest version of azure-core, but at this point of the build process we don't want to force access to the dev feed. Add relative dep on version of azure-core that is actually necessary.

@jalauzon-msft


jalauzon-msft (approved):
Thanks for jumping on this!

PR #27811, opened 2022-12-05
[ML][Pipelines] Not use tmp path in flaky test

zhengfeiwang:
# Description

Component end-to-end test `test_component_create_twice_same_code_arm_id` consistently breaks in live test with error "azure.ai.ml.exceptions.AssetException: Error creating codes asset : Directory is empty. path or local_path must be a non-empty directory.", which may be caused by tmp path. This test previously use tmp path as we have weekly-built workspace and need to avoid reusing; as we have daily new workspace in live test now, there is no need to use tmp path any more, change it to local file.

### DIFF
--- a/sdk/ml/azure-ai-ml/tests/component/e2etests/test_component.py
+++ b/sdk/ml/azure-ai-ml/tests/component/e2etests/test_component.py
@@ -367,24 +367,9 @@ class TestComponent(AzureRecordedTestCase):
 
     @pytest.mark.disable_mock_code_hash
     @pytest.mark.skipif(condition=not is_live(), reason="reuse test, target to verify service-side behavior")
-    def test_component_create_twice_same_code_arm_id(
-        self, client: MLClient, randstr: Callable[[str], str], tmp_path: Path
-    ) -> None:
+    def test_component_create_twice_same_code_arm_id(self, client: MLClient, randstr: Callable[[str], str]) -> None:
+        component_path = "./tests/test_configs/components/component_for_reuse_test/component.yml"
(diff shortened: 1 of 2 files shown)


sdk/ml/azure-ai-ml/tests/component/e2etests/test_component.py
```diff
@@ -367,24 +367,9 @@ def test_component_update(self, client: MLClient, randstr: Callable[[str], str])
 
     @pytest.mark.disable_mock_code_hash
     @pytest.mark.skipif(condition=not is_live(), reason="reuse test, target to verify service-side behavior")
-    def test_component_create_twice_same_code_arm_id(
-        self, client: MLClient, randstr: Callable[[str], str], tmp_path: Path
-    ) -> None:
+    def test_component_create_twice_same_code_arm_id(self, client: MLClient, randstr: Callable[[str], str]) -> None:
+        component_path = "./tests/test_configs/components/component_for_reuse_test/component.yml"
         component_name = randstr("component_name")
```
wangchao1230, line 372: why is it using tmp path before?

zhengfeiwang: `# create new component to prevent the issue when same component code got created at the same time`
the above line is the comment for this test previously

I understand based on previous weekly workspace, the code asset will only create once for the first time; now we have brand new workspace every time for this test, so I think it's reasonable to remove the use of tmp path.

wangchao1230 (approved):

PR #27815, opened 2022-12-05
[ML] [Pipelines] fix: try to enable internal components in pipeline_node_factory

elliotzh:
# Description

+ refactor to try to enable internal component in pipeline node factory (previously only do this in component factory and may skip registration if we directly get a pipeline job with internal nodes)
+ use `==` to judge whether we should update previous base path in schema.pre_dump to avoid resolving the path (then a non-existent path will lead to exception)

### DIFF
--- a/sdk/ml/azure-ai-ml/azure/ai/ml/_schema/core/resource.py
+++ b/sdk/ml/azure-ai-ml/azure/ai/ml/_schema/core/resource.py
@@ -40,10 +40,11 @@ class ResourceSchema(YamlFileSchema):
     def update_base_path_pre_dump(self, data, **kwargs):
         # inherit from parent if base_path is not set
         if data.base_path:
-            cur_base_path: Path = Path(data.base_path).resolve()
-            if not cur_base_path.samefile(self.context[BASE_PATH_CONTEXT_KEY]):
-                self._previous_base_path = Path(self.context[BASE_PATH_CONTEXT_KEY])
-                self.context[BASE_PATH_CONTEXT_KEY] = cur_base_path
+            cur_base_path: Path = Path(data.base_path)
+            base_path_in_context = Path(self.context[BASE_PATH_CONTEXT_KEY])
+            if cur_base_path == base_path_in_context:
+                return data
+            self.context[BASE_PATH_CONTEXT_KEY] = cur_base_path
         return data
 
     @post_dump
--- a/sdk/ml/azure-ai-ml/azure/ai/ml/_schema/core/resource.py
+++ b/sdk/ml/azure-ai-ml/azure/ai/ml/_schema/core/resource.py
@@ -5,7 +5,6 @@
 # pylint: disable=unused-argument,no-self-use,protected-access
 
 import logging
-from pathlib import Path
 
 from marshmallow import fields, post_load, pre_dump, post_dump
 
@@ -40,11 +39,8 @@ class ResourceSchema(YamlFileSchema):
     def update_base_path_pre_dump(self, data, **kwargs):
         # inherit from parent if base_path is not set
         if data.base_path:
-            cur_base_path: Path = Path(data.base_path)
-            base_path_in_context = Path(self.context[BASE_PATH_CONTEXT_KEY])
-            if cur_base_path == base_path_in_context:
-                return data
-            self.context[BASE_PATH_CONTEXT_KEY] = cur_base_path
+            self._previous_base_path = self.context[BASE_PATH_CONTEXT_KEY]
+            self.context[BASE_PATH_CONTEXT_KEY] = data.base_path
         return data
 
     @post_dump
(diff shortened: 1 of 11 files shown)


sdk/ml/azure-ai-ml/azure/ai/ml/_schema/core/resource.py
```diff
@@ -40,10 +40,11 @@ def pass_source_path(self, data, original, **kwargs):
     def update_base_path_pre_dump(self, data, **kwargs):
         # inherit from parent if base_path is not set
         if data.base_path:
-            cur_base_path: Path = Path(data.base_path).resolve()
-            if not cur_base_path.samefile(self.context[BASE_PATH_CONTEXT_KEY]):
-                self._previous_base_path = Path(self.context[BASE_PATH_CONTEXT_KEY])
-                self.context[BASE_PATH_CONTEXT_KEY] = cur_base_path
+            cur_base_path: Path = Path(data.base_path)
```
wangchao1230, line 43: Did not get why we change this? Let's sync offline

sdk/ml/azure-ai-ml/azure/ai/ml/entities/_util.py
```diff
@@ -411,3 +414,37 @@ def normalize_job_input_output_type(input_output_value):
...
+def get_type_from_spec(data: dict, *, valid_keys: Iterable[str]) -> str:
+    """Get the type of the node or component from the yaml spec.
+    Yaml spec must have a key named "type" and exception will be raised if it's not once of valid_keys.
+
+    If internal components are enabled, related factory and schema will be updated.
+    """
+    _type, _ = extract_label(data.get(CommonYamlFields.TYPE, None))
+    schema = data.get(CommonYamlFields.SCHEMA, None)
+
+    from azure.ai.ml._utils.utils import try_enable_internal_components
```
wangchao1230, line 428: why is this function not under _internal namepsace?

sdk/ml/azure-ai-ml/azure/ai/ml/_schema/core/resource.py
elliotzh: it's about the implementation of Path.samefile. It will try to resolve the 2 paths and then determine whether they point to the same file/directory. The good part is that, if `self` is a link to `other`, it will return `True`; the bad part is, if one of the base path is not an existent path, it will raise an Exception.

So here I made this change for 2 reasons:
1. actually we can always push base path here, so a soft condition to skip push bring no harm;
2. it will avoid resolving the base path in Schema. So even if the base path is a temp folder and is deleted before dumping, ValidationError will be raised and only raised when trying to dump a relative LocalPathField.

sdk/ml/azure-ai-ml/azure/ai/ml/entities/_util.py
elliotzh: internal components will automatically be enabled once we implicitly import _internal namespace, so this `try_enable` must be put outside _internal namespace.

sdk/ml/azure-ai-ml/tests/internal/unittests/test_pipeline_job.py
```diff
@@ -602,3 +606,14 @@ def test_job_properties(self):
             assert "AZURE_ML_PathOnCompute_" in list(node_dict["properties"].keys())[0]
             assert node_dict["properties"] == rest_node_dict["properties"]
 
+    def test_load_pipeline_job_with_internal_nodes_from_rest(self):
+        # this is a simplified test case which avoid constructing a complete pipeline job rest object
+        from azure.ai.ml.entities._job.pipeline._load_component import pipeline_node_factory
+
+        unregister_internal_components()
```
D-W-, line 613: why unregister here? Did we opened internal component flag for CI?

sdk/ml/azure-ai-ml/tests/internal/unittests/test_pipeline_job.py
```diff
@@ -602,3 +606,14 @@ def test_job_properties(self):
...
+        internal_node_type = "CommandComponent"
+        with environment_variable_overwrite(AZUREML_INTERNAL_COMPONENTS_ENV_VAR, "False"):
+            with pytest.raises(ValidationException, match=f"Unsupported component type: {internal_node_type}."):
```
D-W-, line 616: When will "Internal components is a private feature in v2, please set environment variable" error thrown?

sdk/ml/azure-ai-ml/tests/internal/unittests/test_pipeline_job.py
elliotzh: enabled for this test. Here is to check whether it will **try** register internal components on `PipelineJob._from_rest_object`

sdk/ml/azure-ai-ml/tests/internal/unittests/test_pipeline_job.py
elliotzh: environment variable not set but schema field of yaml spec is of v1.5

PR #27827, opened 2022-12-05
Repair CI Code Coverage

scbedd:
At some point in the past year, the code coverage upload to the devops build definition was disabled from CI. This PR re-enables it.

We've simplified the process by quite a bit, as the existing job publishing was just stomping over itself anyway.

DevOps doesn't have comprehension of `multiple reports`, so there is only one winner anyway.

@harneetvirk for FYI


eng/pipelines/templates/steps/build-test.yml
```diff
@@ -105,13 +105,19 @@ steps:
...
-    displayName: 'Publish .coverage files'
+    displayName: 'Publish raw .coverage files'
     continueOnError: true
     condition: and(succeeded(), ${{ parameters.RunCoverage }})
     inputs:
       targetPath: '$(Build.SourcesDirectory)/_coverage'
       artifactType: 'pipeline'
 
+  - pwsh: |
+      python -m pip install coverage==4.5.4
```
scbedd, line 116: @mccoyp this will stay for now, but we should bump `pytest-cov`, `coverage`, etc across the board (which will remove the need for this).

I don't want to break into jail on that work item here, as I want this to be enabled for our PRs sooner rather than later.

mccoyp (approved):

eng/pipelines/templates/steps/build-test.yml
scbedd: Here is the follow-up issue. https://github.com/Azure/azure-sdk-for-python/issues/27828

Won't just get forgotten.

scbedd:
Intermittent failure on the livetest run. Single failure of a precondition, not related to my systematic change 👍

PR #27101, opened 2022-10-27
Eliminate Remaining `requests` usage within `azure-sdk-tools`

scbedd:
Resolves #26925

Remaining to complete:

- [x] Raise for status when sanitizer, matcher, or transform add fails. We weren't seeing the sanitizers raise before because we weren't using the `sanitizers.py` http client. 
- [x] Ensure the new `urllib3` honors the overridden environment variables that we use to override requests cert bundles.

PR #27842, opened 2022-12-06
[Monitor] Remove monitor nspkg

pvaneck:
This is not used and is no longer needed.


pvaneck:
Needs https://github.com/Azure/azure-sdk-for-python/pull/27101 in first.

Issue #27848, opened 2022-12-07
Symlink to file is not resolved correctly


helgehatt:
- **Package Name**: azure-ai-ml
- **Package Version**: 1.2.0
- **Operating System**: Ubuntu 20.04.4
- **Python Version**: 3.10.6

**Describe the bug**
Symlinks to files are not resolved correctly (symlink to folder works).

**To Reproduce**
Create the following folder structure:
./config.json
./hello_world/component.py
./hello_world/config.json (symlinked using `ln -s ../config.json config.json`)

In ./hello_world/component.py:
```
import mldesigner


@mldesigner.command_component(name="hello_world")
def hello_world():
    pass


if __name__ == "__main__":
    from azure.ai.ml import MLClient
    from azure.identity import InteractiveBrowserCredential

    client = MLClient.from_config(credential=InteractiveBrowserCredential())
    client.components.create_or_update(hello_world)

```

Run the file `python hello_world/component.py`.

**Expected behavior**
The symlink is resolved correctly and the component can be uploaded without a problem.

**Actual behavior**
The following error is raised:
`FileNotFoundError: [Errno 2] No such file or directory: '/home/azureuser/localfiles/my-workspace/../config.json'`

Setting a breakpoint at line 358 in azure/ai/ml/_utils/_asset_utils.py shows that `target_absolute_path` is set incorrectly.
```
>> path
'/home/azureuser/localfiles/my-workspace/hello_world/config.json'
>> os.readlink(path)
'../config.json'
>> target_absolute_path
'/home/azureuser/localfiles/my-workspace/../config.json'
```

`target_absolute_path` is set using:
```
target_absolute_path = os.path.join(working_dir, os.readlink(path))
```
where `working_dir` is `/home/azureuser/localfiles/my-workspace`

My suggestion is to set it using:
```
target_absolute_path = os.path.join(os.path.dirname(path), os.readlink(path))
```
since the symlink is relative to its parent, given by `os.path.dirname(path)`, and not `working_dir`.

kashifkhan:
Thank you for the feedback @helgehatt . We will investigate this asap

cc @azureml-github

diondrapeck:
Thank you for finding this bug and providing a potential solution, @helgehatt. We have created a work item for this and will update when the fix is in.

helgehatt:
My workaround is to run it from the component directory `(cd hello_world; python component.py)`.
That works, but it should not be necessary to do it that way.

PR #27849, opened 2022-12-07
Fix symlink to file is not resolved correctly #27848

helgehatt:
Implement suggestion described in issue #27848.


helgehatt:
Fixes the first exception, but now another appears at a later stage:

```
Exception has occurred: AssetException
Error with code: [Errno 2] No such file or directory: '../../config.json'
  File "/anaconda/envs/my-workspace/lib/python3.10/site-packages/azure/ai/ml/operations/_operation_orchestrator.py", line 256, in _get_code_asset_arm_id
    code_asset = self._code_assets.create_or_update(code_asset)
```

however this is beyond my grasp of the code.

helgehatt:
Fixed in #27960

PR #27890, opened 2022-12-09
Update workspace create to use new version of app insights

joshharrin:
Make updates to the workspace create operation to make use of the new version of application insights (Log Analytics-based Application Insights). I have added the _app_insights_utils.py for most of the helper operations related to this. This mainly involves adding some information to templates being used and adds some deployments necessary to support this new version of application insights. The previously flow remains the same when a user brings their own app insights but now when creating app insights on the user's behalf we use the new version which requires log analytics workspace attached to the app insights resource. 

The naming convention used for default resources to be created with the AzureML Workspace related to Application Insights will be as follows. This naming convention will be consistent between the following paths for creating AzureML workspaces, meaning that they will be reused as intended within subscription/region scope: Portal/Studio/SDK/CLI

**"DefaultWorkspace-{location}"**
**"DefaultResourceGroup-{location}"**

Reasons for this change: 
1.	Reduce manual changes necessary when new regions are introduced (by removing the regionMapping logic, which is only used for creating the name) 
2.	Use simple, consistent naming pattern for default resources
3.	Remove dependency on external logic for regionMapping which is hard to track and not kept up to date. 

Changes have also been made to update the workspace_keys _from_rest_object as it can cause null reference exceptions when no ACR existed for the workspace. Added None check and will return None for those credentials in that case instead of throwing an error. 

Feature tracking this work in ADO
https://dev.azure.com/msdata/Vienna/_workitems/edit/1912636/

### DIFF
--- a/sdk/ml/azure-ai-ml/tests/workspace/unittests/test_workspace_operations.py
+++ b/sdk/ml/azure-ai-ml/tests/workspace/unittests/test_workspace_operations.py
@@ -212,16 +212,6 @@ class TestWorkspaceOperation:
         mock_workspace_operation._operation.begin_diagnose.assert_called_once()
         mocker.patch("azure.ai.ml._restclient.v2022_10_01.models.DiagnoseRequestProperties", return_value=None)
 
-<<<<<<< HEAD
-    def test_populate_arm_paramaters(
-        self, mock_workspace_operation: WorkspaceOperations, mocker: MockFixture
-    ) -> None:
-        mocker.patch("azure.ai.ml.operations._workspace_operations.get_resource_group_location", return_value="random_name")
-<<<<<<< HEAD
-=======
-        mocker.patch("azure.ai.ml.operations._workspace_operations.get_default_log_analytics_arm_id", return_value=("random_id", True))
->>>>>>> a42f60d482 (comment fix)
-=======
     def test_populate_arm_paramaters(self, mock_workspace_operation: WorkspaceOperations, mocker: MockFixture) -> None:
         mocker.patch(
             "azure.ai.ml.operations._workspace_operations.get_resource_group_location", return_value="random_name"
@@ -230,7 +220,6 @@ class TestWorkspaceOperation:
             "azure.ai.ml.operations._workspace_operations.get_default_log_analytics_arm_id",
             return_value=("random_id", True),
         )
->>>>>>> 969fdb7b0b (run black, remove % interpolation)
         mock_workspace_operation._populate_arm_paramaters(workspace=Workspace(name="name"))
 
     def test_populate_arm_paramaters_other_branches(


sdk/ml/azure-ai-ml/azure/ai/ml/_utils/_appinsights_utils.py
```diff
@@ -0,0 +1,535 @@
...
+        base_url=_get_base_url_from_metadata(),
+        api_version=ArmConstants.AZURE_MGMT_RESOURCE_API_VERSION,
+    )
+    default_rgName = get_default_resource_group_name(location)
+    # rgs = client.resource_groups.get("""try to get the default resource group""")
+    try:
+        rgs = client.resource_groups.get(resource_group_name=default_rgName)
+        return rgs.name
+    except:
+        created_rg = client.resource_groups.create_or_update(default_rgName, {"location":location})
```
luigiw, line 31: Shouldn't RG creation be part of the ARM template?

sdk/ml/azure-ai-ml/azure/ai/ml/_utils/_appinsights_utils.py
```diff
@@ -0,0 +1,535 @@
...
+    if location in ["westcentralus", "eastus2euap", "centraluseuap"]:
+        return region_mapping["regions"]["southcentralus"]["laRegionCode"]
+    try: 
+        region = region_mapping["regions"][location.lower()]
+        return region["laRegionCode"] if region["laRegionCode"]!="" else location
+    except: 
+        return location
+
+
+region_mapping = {
```
luigiw, line 103: It's scary to see this in the SDK code. Will this mapping ever change? Also does the UX have the same mapping so that the workplace creation experience is consistent between the CLI/SDK and the UX? I feel it may be worth to create a Rest API for this.

joshharrin: Yes, I suppose that brings a good point. I was following the practice of the Portal UI and Studio where these values are manually included within the code. This mapping should not change for existing regions, but additions would happen for new regions. [Portal](https://dev.azure.com/msdata/Vienna/_git/vienna?path=/IbizaExtension.MLServices/MLTeamAccounts/MLTeamAccounts/Client/Resource/Create/Utilities/AppInsightsUtilities.ts&version=GBmaster&line=86&lineEnd=86&lineStartColumn=1&lineEndColumn=45&lineStyle=plain&_a=contents) and [ML Studio](https://dev.azure.com/msdata/Vienna/_git/workspace-portal?path=/common/core/src/context/workspace/AppInsights.utils.ts&version=GCac30c6c3c3ba1494ecfaa8dcbc21900ab8c2eaa5&line=55&lineEnd=55&lineStartColumn=1&lineEndColumn=45&lineStyle=plain&_a=contents). These values are used for the naming of resources. I think this is not within scope of creating a REST API to fetch this information since these values are not defined by AzureML. I guess an alternative approach would be to change the naming convention for these resources to something like "DefaultWorkspace-{subscriptionId}-{location}" instead of "DefaultWorkspace-{subscriptionId}-{laRegionCode}" but this is not the approach taken by Portal/Studio. I could try to see if updates on that end would be possible, what are your thoughts?

sdk/ml/azure-ai-ml/azure/ai/ml/_utils/_appinsights_utils.py
joshharrin: I am creating it here since this is to create the default resource group that is needed to create the log analytics resource. It would need to exist before creating the log analytics resource inside of that resource group with the ARM template.

sdk/ml/azure-ai-ml/azure/ai/ml/_utils/_appinsights_utils.py
joshharrin: A concern is that we should continue to use the same naming convention used for previously created resources since the goal in this is to reuse the same Log Analytics resource across many AzureML deployments. If we change the naming convention, that would not be the case. I see both sides here but am not sure how to proceed.

joshharrin: Actually, this naming convention will not work "DefaultWorkspace-{subscriptionId}-{location}" due to the limitation that each Log Analytics Workspace name must be between 4 and 63 characters in length. With this, the current DefaultWorkspace-{subscriptionId}- uses 54 of those characters. Thus, locations over 9 characters will violate that rule, causing failure in creating resource (ie. northcentralus)

sdk/ml/azure-ai-ml/azure/ai/ml/_utils/_appinsights_utils.py
joshharrin: Updated to include resource group creation in ARM template as a deployment, added default resource group deployment to depends on for log analytics resource

sdk/ml/azure-ai-ml/azure/ai/ml/_utils/_appinsights_utils.py
joshharrin: Updated naming convention to "DefaultWorkspace-{location}" and removed region mapping, simplifying names for resources and removing external logic/dependencies

sdk/ml/azure-ai-ml/azure/ai/ml/_utils/_appinsights_utils.py
```diff
@@ -0,0 +1,128 @@
...
+    client = ResourceManagementClient(
+        credential=credentials,
+        subscription_id=subscription_id,
+        base_url=_get_base_url_from_metadata(),
+        api_version=ArmConstants.AZURE_MGMT_RESOURCE_API_VERSION,
+    )
+    default_resource_group = get_default_resource_group_name(location)
+    default_workspace = client.resources.list_by_resource_group(
+        default_resource_group,
+        filter="substringof('%s',name)"%get_default_log_analytics_name(location)
```
needuv, line 44:
```suggestion
        filter=f"substringof('{get_default_log_analytics_name(location)}',name)"
```

prefer f-string over % interpolation

sdk/ml/azure-ai-ml/azure/ai/ml/_utils/_appinsights_utils.py
```diff
@@ -0,0 +1,128 @@
...
+    return False
+
+
+def get_default_log_analytics_arm_id(
+    subscription_id: str,
+    location: str) -> str:
+    return '/subscriptions/%s/resourceGroups/%s/providers/Microsoft.OperationalInsights/workspaces/%s'%(
+        subscription_id,
+        get_default_resource_group_name(location),
+        get_default_log_analytics_name(location))
```
needuv, line 59:
```suggestion
    return (
        f"/subscriptions/{subscription_id}/"
        "resourceGroups/{ get_default_resource_group_name(location)}/"
        "providers/Microsoft.OperationalInsights/workspaces/{get_default_log_analytics_name(location))}"
    )
```

sdk/ml/azure-ai-ml/azure/ai/ml/_utils/_appinsights_utils.py
```diff
@@ -0,0 +1,128 @@
...
+                        "properties": {}
+                    }
+                ]
+            }
+        }
+    }
+
+
+def get_default_log_analytics_name(location: str) -> str:
+    return "DefaultWorkspace-%s"%(location)
```
needuv, line 124:
```suggestion
    return f"DefaultWorkspace-{location}"
```

sdk/ml/azure-ai-ml/azure/ai/ml/_utils/_appinsights_utils.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+def get_default_resource_group_name(location: str) -> str:
+    return "DefaultResourceGroup-%s"%(location)
```
needuv, line 128:
```suggestion
    return f"DefaultResourceGroup-{location}"
```

sdk/ml/azure-ai-ml/azure/ai/ml/operations/_workspace_operations.py
```diff
@@ -469,6 +476,61 @@ def _populate_arm_paramaters(self, workspace: Workspace) -> Tuple[dict, dict, di
...
+                app_insights_location = workspace.location
+            # check if default resource group already exists
+            rg_is_existing = default_resource_group_for_app_insights_exists(
+                self._credentials,
+                self._subscription_id,
+                app_insights_location)
+            # if default resource group does not exist yet, create rg and log analytics
+            if not rg_is_existing:
+                # add resource group and log analytics deployments to resources
+                app_insights_resource_group_deployment_name = "DeployResourceGroup%s"%get_deployment_name("")
```
needuv, line 492: couple more `%s` formatting things to clean up in this file

sdk/ml/azure-ai-ml/azure/ai/ml/operations/_workspace_operations.py
```diff
@@ -469,6 +476,61 @@ def _populate_arm_paramaters(self, workspace: Workspace) -> Tuple[dict, dict, di
...
+            # if workspace is located in a region where app insights is not supported, we do this swap
+            if workspace.location in ["westcentralus", "eastus2euap", "centraluseuap"]:
+                app_insights_location = "southcentralus"
+            else:
+                app_insights_location = workspace.location
+            # check if default resource group already exists
+            rg_is_existing = default_resource_group_for_app_insights_exists(
+                self._credentials,
+                self._subscription_id,
+                app_insights_location)
```
needuv, line 488: could you run `black` on each of the files in your PR? `black` is a python code formatter, and it will keep the style consistent with the rest of the SDK. It's on pypi: https://pypi.org/project/black/, and after installing it, you can just run:

```
black <path to file> --line-length=120
```

joshharrin: I have done this for all files except for the arm template file. It does not make sense for the ARM template for readability.

sdk/ml/azure-ai-ml/azure/ai/ml/operations/_workspace_operations.py
```diff
@@ -469,6 +476,63 @@ def _populate_arm_paramaters(self, workspace: Workspace) -> Tuple[dict, dict, di
                 group_name,
             )
         else:
+            # if workspace is located in a region where app insights is not supported, we do this swap
+            if workspace.location in ["westcentralus", "eastus2euap", "centraluseuap"]:
```
luigiw, line 480: Is this needed given the ARM template checks for the app insights location too?

joshharrin: I do this check to get the location to use for default Resource Group and default Log Analytics needed for app insights as well for the alternate region if necessary. It is needed for getting the deployments for those.

luigiw (approved):

luigiw:
There's one more thing before I merge the PR. Can you add a ChangeLog item? https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/ml/azure-ai-ml/CHANGELOG.md

joshharrin:
updated changelog

needuv (approved):

PR #27892, opened 2022-12-09
[AppConfig] Fix cspell errors

YalinLi0312:
- Resolves https://github.com/Azure/azure-sdk-for-python/issues/22589

- And also changed the test file names - that's why also has recordings updates for all test files in this PR.
I am thinking about removing file name updates to a separate PR - since the cspell fix don't have much, so just leave them there.
_Let me know if you think it's easier to review when extracting them to a separate PR._


xiangyan99:
Why we need to re-record?

YalinLi0312:
> Why we need to re-record?

As I explained in PR description, that's because I changed file names.

xiangyan99 (approved):

PR #27895, opened 2022-12-09
[Storage] Use `tempfile` for `blob` package

vincenttran-msft:
This PR migrates all FileIO operations (that can be migrated) to use the `tempfile` built-in library.


sdk/storage/azure-storage-blob/tests/test_common_blob.py
```diff
@@ -2775,19 +2770,14 @@ def test_download_to_stream_with_credential(self, **kwargs):
...
+        with tempfile.TemporaryFile() as temp_file:
+            download_blob_from_url(source_blob.url, temp_file, credential=storage_account_key)
+            temp_file.seek(0)
+            # Assert
+            actual = temp_file.read()
             assert data == actual
-        self._teardown(file_path)
 
     @BlobPreparer()
     @recorded_by_proxy
```
vincenttran-msft, line 2783: I don't know how to make comments that show a block of code but basically testcases:
`test_download_to_file_with_existing_file` and
`test_download_to_file_with_existing_file_overwrite`

I cannot figure out a way to use `tempfile` for these. Judging by the tests, the default behavior for using the `download_blob_from_url` API with a path to a real file will throw a `ValueError` if you don't pass in an overwrite flag, but the `tempfile` library does not have this default behavior (seems like overwrite is on my default with no way to turn it off). Furthermore, I noticed that when attempting to migrate `test_download_to_file_with_existing_file` specifically, I will get an auth error on the second `download_blob_from_url` call because there is no `credential=` being passed in, yet it doesn't fail in testcase as it currently is written (maybe the credentials are cached? no idea..)

tl;dr

1. `tempfile` is not rejecting overwrites as real path to files would
2. running into an auth error that doesn't exist when **not** using `tempfile`

jalauzon-msft: Hmm yeah, that's a tricky one. I think you'd need to use a `NamedTemporaryFile` and then pass in the `.name` property? That way it interprets as a file and not a stream:
https://github.com/Azure/azure-sdk-for-python/blob/81d788796b7db16d2a05465b6cacfd7a44958eef/sdk/storage/azure-storage-blob/azure/storage/blob/__init__.py#L178

But then you may still run into problems with this `os.path.isfile` check? Not sure how that works with `tempfile`. For the test seems like you need the file to exist so perhaps writing to it will do that?
https://github.com/Azure/azure-sdk-for-python/blob/81d788796b7db16d2a05465b6cacfd7a44958eef/sdk/storage/azure-storage-blob/azure/storage/blob/__init__.py#L181-L182

Auth error is probably because it made it further and actually tried to do the upload without a cred. Made it past this check ^

vincenttran-msft: Alrighty, after much testing, I was finally able to get it to work properly with `NamedTemporaryFile`. I will say that this is one of the first times I think the official Python documentation is **wrong**??

Judging from [this](https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile) documentation, the default mode is supposed to be `mode='w+b'`, but in my testing I found this to be untrue:
image
image
(bug #1)

However, then I will keep running into a:
`PermissionError: [Errno 13] Permission denied: 'C:\\Users\\VINCEN~1\\AppData\\Local\\Temp\\tmp573r9cgx'` when the `download_blob_from_url` API call would try to open the path.

I eventually got it to succeed after looking at SO examples of how other people are using `NamedTemporaryFile` and I ended up needing to pass a `delete=False` parameter.

I don't believe that is intended behavior and is another bug (unless I am misunderstanding something) because the documentation only talks about delete as the following:
`If delete is true (the default), the file is deleted as soon as it is closed.`

But my `download_blob_from_url` API call is still within the `with` that opens the `NamedTemporaryFile`, so not sure why adding a `delete=False` "fixed" anything 😱 (bug #2?)

image

jalauzon-msft: Firstly, I believe `w+b` and `rb+` are actually the same since the `+` means read and write so I don't think it matters which letter you put it with, `w` or `r`. Still a bit strange it doesn't match but should be functionally equivalent.

As far as `delete=False`, I bet it has something to do with allowing the OS/other things to use the file. Technically we are using the file outside of the tempfile context (not outside the `with` but outside what tempfile knows about) when we do `os.path.isfile(output)`. `delete=False` probably does something with the permissions to make this a real more accessible file (at least on Windows, this may have just worked on Linux).

sdk/storage/azure-storage-blob/tests/test_common_blob.py
```diff
@@ -2788,21 +2788,20 @@ def test_download_to_file_with_existing_file(self, **kwargs):
...
-        # Assert
-        with open(file_path, 'rb') as stream:
-            actual = stream.read()
+            # Assert
+            temp_file.seek(0)
+            actual = temp_file.read()
             assert data == actual
-        self._teardown(file_path)
+
+            temp_file.close()
```
jalauzon-msft, line 2804: I think to delete the file we need to do something additional? I'm seeing `os.unlink(temp_file)`? Maybe also throw that in a try/finally block so well always unlink the file even in the case of an error?

jalauzon-msft (approved):

Issue #26901, opened 2022-10-18
Duplicate code snippet: tables


xiangyan99:
tables\azure-data-tables\samples\sample_query_table.py:query_entities
tables\azure-data-tables\samples\async_samples\sample_query_table_async.py:query_entities

PR #27896, opened 2022-12-10
[Tables] Merge samples to one code snippet

YalinLi0312:
- Resolves https://github.com/Azure/azure-sdk-for-python/issues/26901

### DIFF
--- a/sdk/tables/azure-data-tables/samples/sample_query_table.py
+++ b/sdk/tables/azure-data-tables/samples/sample_query_table.py
@@ -106,7 +106,7 @@ class SampleTablesQuery(object):
 
                 print("Sample for querying entities' values:")
                 print("Entities with 25 < Value < 50")
-                parameters = {"lower": 25, "upper": 50}
+                parameters = {"lower": 25, "upper": 50} # type: ignore
                 name_filter = "Value gt @lower and Value lt @upper"
                 queried_entities = table_client.query_entities(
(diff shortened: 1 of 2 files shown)


sdk/tables/azure-data-tables/samples/sample_query_table.py
```diff
@@ -83,77 +83,42 @@ def sample_query_entities(self):
...
                 queried_entities = table_client.query_entities(
                     query_filter=name_filter, select=["Value"], parameters=parameters
                 )
-
                 for entity_chosen in queried_entities:
                     print(entity_chosen)
-
             except HttpResponseError as e:
-                print(e.message)
+                pass
```
annatisch, line 117: Using `pass` on an HttpResponseError is probably not a good idea for a sample - we don't want people to copy and paste this by accident. I would probably print some kind of message then raise it.

YalinLi0312: The exception raised here will be caught and print out in main(line131).

annatisch: No it wont - because this `pass` is suppressing it. It would need to be `raise`

YalinLi0312: `raise` is what I meant, updated!

annatisch (approved):

PR #27899, opened 2022-12-10
[ACR] Close client

YalinLi0312:
Checked `samples` and `tests` folder, close client in each of the test.


mccoyp (approved):
Out of curiosity, did you also try calling `.close()` on the client? I think that should also work, but I could be mistaken

YalinLi0312:
> Out of curiosity, did you also try calling `.close()` on the client? I think that should also work, but I could be mistaken

I didn't call `.close()`, but I think that also works. I was trying to call close after each test, but was recommended to write in `with` style.

PR #27917, opened 2022-12-13
[formrecognizer] Typing updates

### LINKED ISSUE #26834: [formrecognizer] Improve model typing

catalinaperalta:
Fixes #26834
Improve typing, enable more typing checks


tools/azure-sdk-tools/ci_tools/environment_exclusions.py
```diff
@@ -228,6 +228,7 @@
     "azure-eventhub-checkpointstoreblob",
     "azure-eventhub-checkpointstoreblob-aio",
     "azure-eventhub-checkpointstoretable",
+    "azure-identity",
```
catalinaperalta, line 231: Something weird happened when merging with upstream here. Will double check that this is currently included on the list.

catalinaperalta: Fixed

catalinaperalta:
NOTE: I reorganized some models so that they're declared before they're referenced from other models, the git diff may be a bit confusing due to that

sdk/formrecognizer/azure-ai-formrecognizer/azure/ai/formrecognizer/_models.py
```diff
@@ -231,70 +232,114 @@ class FormContentType(str, Enum, metaclass=CaseInsensitiveEnumMeta):
...
 
+class TextAppearance:
+    """An object representing the appearance of the text line.
+
+    .. versionadded:: v2.1
+        Support for *to_dict* and *from_dict* methods
+    """
+    style_name: str
+    """The text line style name.
+        Possible values include: "other", "handwriting"."""
```
kristapratico, line 272: I noticed when I generated documentation that the `:` did a weird thing and created a `Type` in the docstring. Not sure if you'll see the same thing when you run sphinx but I just removed the `:` and replaced with `possible values are...`

kristapratico (approved):

sdk/formrecognizer/azure-ai-formrecognizer/azure/ai/formrecognizer/_models.py
catalinaperalta: Gave the docs a skim and didnt see this causing a problem, however I'll do a more in-depth check, tracking issue: https://github.com/Azure/azure-sdk-for-python/issues/28326

PR #27959, opened 2022-12-15
[formrecognizer] Update samples

catalinaperalta:
Enable type checking on samples, fix typing bug on begin_recognizer_invoices()

### DIFF
--- a/sdk/formrecognizer/azure-ai-formrecognizer/samples/v3.1/async_samples/sample_authentication_v3_1_async.py
+++ b/sdk/formrecognizer/azure-ai-formrecognizer/samples/v3.1/async_samples/sample_authentication_v3_1_async.py
@@ -7,7 +7,7 @@
 # --------------------------------------------------------------------------
 
 """
-FILE: sample_authentication_async.py
+FILE: sample_authentication_v3_1_async.py
 
 DESCRIPTION:
     This sample demonstrates how to authenticate to the Form Recognizer service.
@@ -20,7 +20,7 @@ DESCRIPTION:
     https://docs.microsoft.com/azure/cognitive-services/authentication
 
 USAGE:
-    python sample_authentication_async.py
+    python sample_authentication_v3_1_async.py
 
     Set the environment variables with your own values before running the sample:
     1) AZURE_FORM_RECOGNIZER_ENDPOINT - the endpoint to your Form Recognizer resource.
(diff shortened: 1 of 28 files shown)


sdk/formrecognizer/azure-ai-formrecognizer/samples/v3.1/async_samples/sample_authentication_v3_1_async.py
```diff
@@ -7,7 +7,7 @@
 # --------------------------------------------------------------------------
 
 """
-FILE: sample_authentication_async.py
+FILE: sample_authentication_v3_1_async.py
```
kristapratico, line 10: Might want to double-check that these renames won't break any links in the MS docs

sdk/formrecognizer/azure-ai-formrecognizer/samples/v3.1/async_samples/sample_convert_to_and_from_dict_v3_1_async.py
```diff
@@ -60,8 +60,8 @@ async def convert_to_and_from_dict_async():
     # save the dictionary as JSON content in a JSON file, use the AzureJSONEncoder
     # to help make types, such as dates, JSON serializable
     # NOTE: AzureJSONEncoder is only available with azure.core>=1.18.0.
-    with open('data.json', 'w') as f:
-        json.dump(recognized_form_dict, f, cls=AzureJSONEncoder)
+    with open('data.json', 'w') as f:  # type: ignore
+        json.dump(recognized_form_dict, f, cls=AzureJSONEncoder)  # type: ignore
```
kristapratico, line 64: We should not have `type: ignore` in samples since we don't want to force our customers to have to use it. I think mypy is getting confused here since we use `f` in a difference context on L52-53. If you rename it to something else on these lines I think that will resolve your typing error and we won't need the ignore.

sdk/formrecognizer/azure-ai-formrecognizer/samples/v3.1/async_samples/sample_strongly_typing_recognized_form_async.py
```diff
@@ -39,20 +39,20 @@ class Receipt(object):
...
-        self.transaction_date = form.fields.get("TransactionDate", FormField())
-        self.transaction_time = form.fields.get("TransactionTime", FormField())
-
-    def convert_to_receipt_item(self, items):
+    def __init__(self, form: RecognizedForm) -> None:
+        self.receipt_type: FormField  = form.fields.get("ReceiptType", FormField())
+        self.merchant_name: FormField  = form.fields.get("MerchantName", FormField())
+        self.merchant_address: FormField  = form.fields.get("MerchantAddress", FormField())
+        self.merchant_phone_number: FormField  = form.fields.get("MerchantPhoneNumber", FormField())
+        self.receipt_items: list[ReceiptItem]  = self.convert_to_receipt_item(form.fields.get("Items", FormField()))
```
kristapratico, line 47: generic `list` as a type hint is not supported <3.9. I would just use typing.List

lmazuel (approved):

Issue #26256, opened 2022-09-15
[EventHubs&ServiceBus] add amqp transport info into user-agent


swathipil:
Anna: "I think there might be some value in us adding transport information into the user-agent?

That way (with service team cooperation) we can get some insight into uamqp usage over time and how long support for it should be maintained."

swathipil:
EH user agent changes copied over to SB

PR #27966, opened 2022-12-15
[Eventhub] Add Transport Information into user-agent

kashifkhan:
#26256 

Adding in pyamqp / uamqp into user agent so we can usage between the two libraries.

### DIFF
--- a/sdk/eventhub/azure-eventhub/azure/eventhub/_configuration.py
+++ b/sdk/eventhub/azure-eventhub/azure/eventhub/_configuration.py
@@ -6,7 +6,7 @@ from typing import Optional, Dict, Any
 from urllib.parse import urlparse
 
 from azure.core.pipeline.policies import RetryMode
-from ._constants import TransportType, DEFAULT_AMQPS_PORT, DEFAULT_AMQP_WSS_PORT, UAMQP_LIBRARY, PYAMQP_LIBRARY
+from ._constants import TransportType, DEFAULT_AMQPS_PORT, DEFAULT_AMQP_WSS_PORT
 
 
 
@@ -36,8 +36,6 @@ class Configuration(object):  # pylint:disable=too-many-instance-attributes
         self.custom_endpoint_hostname = None
         self.hostname = kwargs.pop("hostname")
         uamqp_transport = kwargs.pop("uamqp_transport")
-        amqp_library = PYAMQP_LIBRARY if not uamqp_transport else UAMQP_LIBRARY
-        self.user_agent = f"{amqp_library} {self.user_agent}"
 
         if self.http_proxy or self.transport_type.value == TransportType.AmqpOverWebsocket.value:
             self.transport_type = TransportType.AmqpOverWebsocket
--- a/sdk/eventhub/azure-eventhub/azure/eventhub/_configuration.py
+++ b/sdk/eventhub/azure-eventhub/azure/eventhub/_configuration.py
@@ -6,7 +6,7 @@ from typing import Optional, Dict, Any
 from urllib.parse import urlparse
 
 from azure.core.pipeline.policies import RetryMode
-from ._constants import TransportType, DEFAULT_AMQPS_PORT, DEFAULT_AMQP_WSS_PORT, UAMQP_TRANSPORT, PYAMQP_TRANSPORT
+from ._constants import TransportType, DEFAULT_AMQPS_PORT, DEFAULT_AMQP_WSS_PORT, UAMQP_LIBRARY, PYAMQP_LIBRARY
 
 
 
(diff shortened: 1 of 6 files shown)


sdk/eventhub/azure-eventhub/azure/eventhub/_configuration.py
```diff
@@ -36,6 +36,8 @@ def __init__(self, **kwargs):
         self.custom_endpoint_hostname = None
         self.hostname = kwargs.pop("hostname")
         uamqp_transport = kwargs.pop("uamqp_transport")
+        self.amqp_library = PYAMQP_TRANSPORT if not uamqp_transport else UAMQP_TRANSPORT
```
swathipil, line 39: is the `self.amqp_library` used anywhere else? (i.e. reason for making this an instance var?)

if not, maybe we can make this `amqp_user_agent_prefix`/`amqp_library`/etc.

sdk/eventhub/azure-eventhub/azure/eventhub/_constants.py
```diff
@@ -44,6 +44,8 @@
 MGMT_STATUS_CODE = b"status-code"
 MGMT_STATUS_DESC = b"status-description"
 USER_AGENT_PREFIX = "azsdk-python-eventhubs"
+UAMQP_TRANSPORT = "uamqp"
+PYAMQP_TRANSPORT = "pyamqp"
```
swathipil, line 48: since we use `uamqp_transport` as a bool in lots of other places, to avoid confusion, maybe we can call these `UAMQP_USER_AGENT_PREFIX`/`PYAMQP_USER_AGENT_PREFIX` or something similar?

sdk/eventhub/azure-eventhub/azure/eventhub/_configuration.py
kashifkhan: Ill make them _LIBRARY given thats the variable name I use to store em

sdk/eventhub/azure-eventhub/azure/eventhub/_configuration.py
```diff
@@ -36,6 +36,8 @@ def __init__(self, **kwargs):
         self.custom_endpoint_hostname = None
         self.hostname = kwargs.pop("hostname")
         uamqp_transport = kwargs.pop("uamqp_transport")
+        amqp_library = PYAMQP_LIBRARY if not uamqp_transport else UAMQP_LIBRARY
+        self.user_agent = f"{amqp_library} {self.user_agent}"
```
swathipil, line 40: one other thing, not sure if this matters or not: do we want to make the space a dash here?
```suggestion
        self.user_agent = f"{amqp_library}-{self.user_agent}"
```

b/c I see in the previous user agent, there are no spaces so we may want to follow the same convention

swathipil (approved):
lgtm! (just have that one comment about how to format the new user agent string, which I believe you mentioned you'll be resolving)

sdk/eventhub/azure-eventhub/azure/eventhub/_configuration.py
kashifkhan: yeah this is one I was unsure about since the final string looks like this `f"{USER_AGENT_PREFIX}/{VERSION} {framework} ({platform_str})"`

so maybe it should be `f"{self.user_agent}/{amqp_library}"` making the final string `f"{self.user_agent}/{amqp_library}/{VERSION} {framework} ({platform_str})"`?

annatisch: I assume we're following the [HTTP user agent formatting convention](https://azure.github.io/azure-sdk/general_azurecore.html#telemetry-policy)?
If so, it would be a list of packages paired with their versions, ordered from outer-most to inner-most - so something like:
`{eventhub-package}/{eventhub-version} {amqp-package}/{amqp-version} ...`

sdk/eventhub/azure-eventhub/azure/eventhub/_constants.py
```diff
@@ -44,6 +44,8 @@
 MGMT_STATUS_CODE = b"status-code"
 MGMT_STATUS_DESC = b"status-description"
 USER_AGENT_PREFIX = "azsdk-python-eventhubs"
+UAMQP_LIBRARY = "uamqp"
+PYAMQP_LIBRARY = "pyamqp"
```
annatisch, line 48: For a user-agent, both these should be paired with their version:
`uamqp_moniker = f"uamqp/{uamqp.__version__}"`

sdk/eventhub/azure-eventhub/azure/eventhub/_configuration.py
kashifkhan: I made some changes to bring about the transport library name & version from the transport classes itself, basically restoring  the Config class. 

The current user agent string is `'azsdk-python-eventhubs/5.11.0 Python/3.9.9 (Windows-10-10.0.22621-SP0)'`

After this change it will be `'azsdk-python-eventhubs/5.11.0 pyamqp/2.0.0a1 Python/3.9.9 (Windows-10-10.0.22621-SP0)'` or `'azsdk-python-eventhubs/5.11.0 uamqp/1.6.3 Python/3.9.9 (Windows-10-10.0.22621-SP0)'`

annatisch (approved):

PR #27989, opened 2022-12-16
pylint

l0lawrence:
Once the new pylint checker plugin is released, we will be disabling the new custom checkers so the only thing effecting the CI is the pylint version bump. Then we will enable the custom checkers one by one to reduce impact on CI.


pylintrc
```diff
@@ -10,14 +10,8 @@ load-plugins=pylint_guidelines_checker
...
-disable=locally-disabled,fixme,too-few-public-methods,import-outside-toplevel,deprecated-module,assigning-non-slot,deprecated-class,deprecated-argument,deprecated-decorator,consider-using-dict-items,consider-iterating-dictionary,signature-differs,duplicate-code,cyclic-import,consider-using-f-string,no-name-in-module,missing-module-docstring,inconsistent-return-statements,missing-class-docstring,missing-function-docstring,too-many-arguments,invalid-name,check-docstrings,networking-import-outside-azure-core-transport,non-abstract-transport-import,name-too-long,delete-operation-wrong-return-type,no-raise-with-traceback
 
+disable=useless-object-inheritance,missing-docstring,locally-disabled,fixme,cyclic-import,too-many-arguments,invalid-name,duplicate-code,too-few-public-methods,consider-using-f-string,super-with-arguments,redefined-builtin,import-outside-toplevel,client-suffix-needed,unnecessary-dunder-call,inconsistent-return-statements,check-docstrings,networking-import-outside-azure-core-transport,non-abstract-transport-import,client-method-has-tracing-decorator,client-method-missing-tracing-decorator-async,name-too-long,delete-operation-wrong-return-type,no-raise-with-traceback
```
tjprescott, line 14: To make these more diffable in the future, can these each be expanded onto separate lines?

pylintrc
```diff
@@ -10,13 +10,8 @@ load-plugins=pylint_guidelines_checker
 
 [MESSAGES CONTROL]
 # For all codes, run 'pylint --list-msgs' or go to 'https://pylint.pycqa.org/en/latest/technical_reference/features.html'
-# locally-disabled: Warning locally suppressed using disable-msg
-# cyclic-import: because of https://github.com/PyCQA/pylint/issues/850
-# too-many-arguments: Due to the nature of the CLI many commands have large arguments set which reflect in large arguments set in corresponding methods.
-# Let's black deal with bad-continuation
```
kristapratico, line 16: nit: looks like we still disable these rules, should we leave the comments that explain why?

pylintrc
```diff
@@ -45,16 +40,17 @@ min-similarity-lines=10
...
-const-name-hint=UPPER_CASE_WITH_UNDERSCORES
-class-name-hint=CapitalizedWords
-class-attribute-name-hint=lower_case_with_underscores
-attr-name-hint=lower_case_with_underscores
-method-name-hint=lower_case_with_underscores
-function-name-hint=lower_case_with_underscores
-argument-name-hint=lower_case_with_underscores
-variable-name-hint=lower_case_with_underscores
-inlinevar-name-hint=lower_case_with_underscores (short is OK)
+# keep short; underscores are discouraged
```
kristapratico, line 43: nit: I don't think this comment makes sense after removing "module-name-hint"

tjprescott (approved):

PR #27990, opened 2022-12-17
Update manage_recordings.py

kristapratico:
It looks like we're passing git commit "owner" when we should be passing git commit "email"


mccoyp (approved):
Thanks for the catch!

PR #28013, opened 2022-12-20
[Storage] Remove `six` library from Python SDK

vincenttran-msft:
This PR is a general improvement to the Python SDK to remove the use of `six` (which was mostly for Python 2 types compatibility) library.


jalauzon-msft (approved):
Thanks for getting to this!

PR #28015, opened 2022-12-21
[Monitor][Ingestion] Add additional perf options

pvaneck:
This adds two additional arguments to the UploadLogsTest:

`--log-content-length` - An integer allowing a user to control the length of each log entry to be uploaded.
`--random-log-content` - Whether the log content should be a random alphanumeric string instead of a repeating character.


kristapratico (approved):
LGTM

pvaneck:
@kristapratico Thanks for reviewing! 😄

PR #28021, opened 2022-12-21
Symlink experiment for _shared folder

petrsvihlik:
This is an experimental PR that is supposed to test whether symbolic links can be used to replace multiple copies of the `_shared` folder in Azure Communication Services.

Changes:
- removed all copies of the `_shared` folder with the exception of one in `communication-identity`
- added symlinks pointing to the `_shared` folder in `communication-identity`
- updated to the latest version of the utlis.py file (from `communication-rooms`) - this file wasn't in sync with other modalities

TODOs:
- replicate the changes for the `\tests\_shared` folders (can be a follow-up PR)


elavarasidc:
@petrsvihlik appreciate your effort in removing multiple copies of the shared folder / same code.

The changes look good to me as we are marching towards 'Don't repeat yourself' (DRY) principle.
I will tag experienced Python SDK engineers (who has context on _share folder original decision] to comment on this PR.

Questions:
Where you able to run all communication test (including live tests), to validate the changes @petrsvihlik ?

petrsvihlik:
@elavarasidc 

> Where you able to run all communication test (including live tests), to validate the changes @petrsvihlik ?

Yeah, the changes seems to work fine with `- tests` pipelines as well. I ran only those that were not failing previously - identity, chat, networktraversal, email.

mikeharder (changes requested):
I don't think we should use symlinks, since they are not well supported on Windows, and we want our repos to "just work" for developers on any platform with default git and OS settings.
 
https://github.com/git-for-windows/git/wiki/Symbolic-Links
 
On my Windows dev machine, `core.symlinks` is `false` by default.

sdk/communication/azure-communication-identity/azure/communication/identity/_shared/utils.py
```diff
@@ -67,6 +68,20 @@ def get_current_utc_as_int():
     return _convert_datetime_to_utc_int(current_utc_datetime)
 
 
+def verify_datetime_format(input_datetime):
```
ostoliarova-msft, line 71: This change seems unrelated to the stated purpose of the PR to reduce repeated code. I'd move it to a separate PR to have one feature per PR and keep focus on it

petrsvihlik: we're tracking a separate issue for this https://github.com/Azure/azure-sdk-for-python/issues/28034

the change here is required in order for the pipelines to pass. we might exclude the change if it gets fixed in the other issue first.

mikeharder:
> @mikeharder it seems that no new symlink objects are added to this PR and the change is that the import has a relative path, which I guess was always supported

In the diff, I believe the files which only contain a path to another file **are** symlinks.

AikoBB:
@petrsvihlik as we won't continue with this approach, can you pls close the PR?

PR #28043, opened 2022-12-23
[Perf] Perfstress framework profiling enhancements

pvaneck:
When a user uses the `--profile` argument with the PerfStress framework, sorted pstats output of the cprofile data is now output to stdout.

Also a `--profile-path` argument is added to allow customization of the output file/path.

Also added a check in  _perf_stress_runner.py to not run the warmup stage if profile is True (this avoids a deadlock).

Related perf automation framework PR: https://github.com/Azure/azure-sdk-tools/pull/5533

### DIFF
--- a/doc/dev/perfstress_tests.md
+++ b/doc/dev/perfstress_tests.md
@@ -82,11 +82,11 @@ class _PerfTestBase:
         # Can be optionally defined. Run once per test instance, after cleanup and global_cleanup.
         # The baseclasses will also define logic here, so if you override this method, make sure you include a call to super().
 
-    def run_all_sync(self, duration: int) -> None:
+    def run_all_sync(self, duration: int, *, run_profiler: bool = False, **kwargs) -> None:
         # Run all sync tests, including both warmup and duration. This method is implemented by the provided base
         # classes, there should be no need to overwrite this function.
 
-    async def run_all_async(self, duration: int) -> None:
+    async def run_all_async(self, duration: int, *, run_profiler: bool = False, **kwargs) -> None:
         # Run all async tests, including both warmup and duration. This method is implemented by the provided base
         # classes, there should be no need to overwrite this function.
 
--- a/doc/dev/perfstress_tests.md
+++ b/doc/dev/perfstress_tests.md
@@ -184,7 +184,8 @@ The framework has a series of common command line options built in:
 - `--no-cleanup` Whether to keep newly created resources after test run. Default is False (resources will be deleted).
 - `--insecure` Whether to run without SSL validation. Default is False.
 - `-x --test-proxies` Whether to run the tests against the test proxy server. Specify the URL(s) for the proxy endpoint(s) (e.g. "https://localhost:5001"). Multiple values should be semi-colon-separated.
-- `--profile` Whether to run the perftest with cProfile. If enabled (default is False), the output file of a single iteration will be written to the current working directory in the format `"cProfile-<TestClassName>-<TestID>-<sync|async>.pstats"`. **Note:** The profiler is not currently supported for the `EventPerfTest` baseclass.
+- `--profile` Whether to run the perftest with cProfile. **Note:** The profiler is not currently supported for the `EventPerfTest` baseclass.
+- `--profile-path` The file path to output profiling data to if profiling is enabled with `--profile`. If not specified, the output file of a single iteration will be written to the current working directory in the format `"cProfile-<TestClassName>-<TestID>-<sync|async>.pstats"`.
 
 ## Running with the test proxy
 Follow the instructions here to install and run the test proxy server:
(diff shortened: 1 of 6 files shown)


pvaneck:
Current behavior for when `--parallel` and `--profile` are both passed in is to run profiling in each parallel thread. Thus, the user will end up with as many pstat files as there are threads. Do you think this would be useful? Or should we cap it to 1 thread and 1 process if `--profile` is passed in? Seems like witha high number of parallel threads, the profile times start to become skewed and you see times for `{method 'enable' of '_lsprof.Profiler' objects}` which normally aren't part of the profiling stats.

Here is a sample run: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=2210387&view=logs&j=9c733e49-b732-5293-9178-f45bf467a578&t=0067490e-d4ca-5514-6944-373ba03217f7 with the perf automation tool where `--size 10240 --parallel 64 --profile` is passed in by default.

doc/dev/perfstress_tests.md
```diff
@@ -184,7 +184,8 @@ The framework has a series of common command line options built in:
 - `--no-cleanup` Whether to keep newly created resources after test run. Default is False (resources will be deleted).
 - `--insecure` Whether to run without SSL validation. Default is False.
 - `-x --test-proxies` Whether to run the tests against the test proxy server. Specify the URL(s) for the proxy endpoint(s) (e.g. "https://localhost:5001"). Multiple values should be semi-colon-separated.
-- `--profile` Whether to run the perftest with cProfile. If enabled (default is False), the output file of a single iteration will be written to the current working directory in the format `"cProfile-<TestClassName>-<TestID>-<sync|async>.pstats"`. **Note:** The profiler is not currently supported for the `EventPerfTest` baseclass.
+- `--profile` Whether to run the perftest with cProfile. **Note:** The profiler is not currently supported for the `EventPerfTest` baseclass.
+- `--profile-path` The file path to output profiling data to if profiling is enabled with `--profile`. If not specified, the output file of a single iteration will be written to the current working directory in the format `"cProfile-<TestClassName>-<TestID>-<sync|async>.pstats"`.
```
pvaneck, line 188: RIght now this is a file path, but potentially could be a directory instead depending on if we want to keep individual profile output files for each parallel thread.

annatisch: I don't think there's really any advantage to creating a profile for every thread. I suspect that's more data than people will realistically consume.

tools/azure-devtools/src/azure_devtools/perfstress_tests/_batch_perf_test.py
```diff
@@ -15,6 +16,10 @@
 from ._policies import PerfTestProxyPolicy
 
 
+PSTATS_PRINT_DEFAULT_SORT_KEY = pstats.SortKey.TIME
+PSTATS_PRINT_DEFAULT_LINE_COUNT = 36
```
annatisch, line 20: Is this chosen at random? Or was it in line with profile support in other languages?

tools/azure-devtools/src/azure_devtools/perfstress_tests/_perf_stress_runner.py
```diff
@@ -215,7 +218,7 @@ async def start(self):
 
             # If a warm up is configured, wait till all tests have finished all setup
             # stages before beginning "Warmup".
-            if self.per_test_args.warmup:
+            if self.per_test_args.warmup and not self.per_test_args.profile:
```
annatisch, line 221: What was causing the deadlock here?

pvaneck: In the `_start_tests` method, we skip the synchronization step for "Warmup" if [profiling is enabled](https://github.com/Azure/azure-sdk-for-python/blob/main/tools/azure-devtools/src/azure_devtools/perfstress_tests/_perf_stress_proc.py#L63). If we run `self._next_stage("Warmup"`, then don't run ` _synchronize(test_stages["Warmup"])`, the following `_synchronize(test_stages["Tests"])` gets stuck. So, here just making sure the conditional checks align.

tools/azure-devtools/src/azure_devtools/perfstress_tests/_batch_perf_test.py
pvaneck: This was chosen pretty arbitarily. In my testing, typical output was 400+ lines in from cProfile/pstats, so I thought 36 was enough to just give a glimpse into the performance heavy hitters for users just viewing a pipeline run without having to download/extract/run through a parser. 

From what I can tell, no other languages have stdout output for profiling and simply output profiling data to a file path.

doc/dev/perfstress_tests.md
pvaneck: I agree. In that case, I think if `--profile` is specified, we can pin `parallel` to 1. Looks like JS has similar restrictions when  profiling.

annatisch (approved):

PR #28047, opened 2022-12-24
[ML][Pipelines] feat: apply in-memory cache for pipeline component dependency resolution

elliotzh:
# Description

In this PR, we have:
1. add an exception handling for validation diagnostics location resolution
2. unify the interface to resolve dependencies for jobs in a pipeline component
3. do the dependencies resolution with iteration instead of recursion
4. fix a bug that the default display name doesn't take effect for sweep.trial in a pipeline job (**recording change involved**)
5. leave interfaces to apply cache and multiprocessing for component registration
6. update hash function for pipeline component (**recording change involved**)
7. apply in-memory cache for component registration (**recording change involved only if cache for anonymous pipeline component is used**)

For now, server-side will regenerate version for anonymous pipeline components on their creation. So, in original recording, 2 anonymous pipeline components of the same structure will have different arm id; after we apply in-memory cache, they will share the same arm id.

In other cases, new requests will be a subset of original requests, so previous recording still works. This is a guarantee on that hash function works well in existing e2e tests.

### DIFF
--- a/sdk/ml/azure-ai-ml/azure/ai/ml/_utils/_cache_utils.py
+++ b/sdk/ml/azure-ai-ml/azure/ai/ml/_utils/_cache_utils.py
@@ -1,36 +1,53 @@
 # ---------------------------------------------------------
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # ---------------------------------------------------------
-
 import uuid
 from collections import defaultdict
-from typing import Union, Callable, Dict
+from typing import Union, Tuple, List
 
+from azure.ai.ml.constants._common import AzureMLResourceType
 from azure.ai.ml.entities import Component
 from azure.ai.ml.entities._builders import BaseNode
 
 
 class CachedNodeResolver(object):
     """Class to resolve component in nodes with cached component resolution results.
+
+    This class is thread-safe if:
+    1) self._resolve_nodes and self._register_node_to_resolve are not called concurrently in the same thread
+    2) self._resolve_component is only called concurrently on independent components
+        a) we have used a component hash to deduplicate components to resolve;
+        b) nodes are registered & resolved layer by layer, so all child components are already resolved
+          when we register a pipeline node;
+        c) potential shared dependencies like compute and data have been resolved before calling self._resolve_nodes.
     """
-    def __init__(self, resolver: Callable[[Union[str, Component]], str]):
+    _ANONYMOUS_HASH_PREFIX = "anonymous_component:"
+    _YAML_SOURCE_PREFIX = "yaml_source:"
+    _ORIGIN_PREFIX = "origin:"
+
+    def __init__(self, resolver):
         self._resolver = resolver
         self._component_resolution_cache = {}
-        self._new_components: Dict[str, Component] = {}
+        self._components_to_resolve: List[Tuple[str, Component]] = []
         self._nodes_to_apply = defaultdict(list)
+
+    def _resolve_component(self, component: Union[str, Component]) -> str:
+        """Resolve a component with self._resolver."""
+        return self._resolver(component, azureml_type=AzureMLResourceType.COMPONENT)
 
     @classmethod
     def _get_component_hash(cls, component: Union[Component, str]):
         """Get a hash for a component."""
         if isinstance(component, str):
             # component can be arm string like "train_model:1"
-            return component
+            return cls._ORIGIN_PREFIX + component
+
         # TODO: add hash function here
         # for pipeline component, all dependencies of pipeline component has already been resolved
         # so we can safely use its anonymous hash
         # for non-pipeline component, we can't directly use anonymous hash as code hasn't been uploaded yet
         # return a random hash by default
-        return str(uuid.uuid4())
+        return cls._ANONYMOUS_HASH_PREFIX + str(uuid.uuid4())
 
     def register_node_to_resolve(self, node: BaseNode):
         """Register a node with its component to resolve.
@@ -42,24 +59,33 @@ class CachedNodeResolver(object):
         # valid after modification.
         component_hash = self._get_component_hash(component)
         if component_hash not in self._component_resolution_cache:
-            self._new_components[component_hash] = component
+            self._components_to_resolve.append((component_hash, component))
+            # set corresponding resolution cache to None to avoid duplicate resolution
             self._component_resolution_cache[component_hash] = None
 
         self._nodes_to_apply[component_hash].append(node)
+
+    def _resolve_components_without_in_memory_cache(self):
+        """Resolve all components to resolve and save the results in cache.
+        """
+        # TODO: Shouldn't reach this function when trying to resolve a subgraph
+        # TODO: apply local cache controlled by an environment variable here
+        # TODO: do concurrent resolution controlled by an environment variable here
+        # given deduplication has already been done, we can safely assume that there is no
+        # conflict in concurrent local cache access
+        # multiprocessing need to dump input objects before starting a new process, which will fail
+        # on _AttrDict for now, so put off the concurrent resolution
+        for component_hash, component in self._components_to_resolve:
+            self._component_resolution_cache[component_hash] = self._resolve_component(component)
+
+        self._components_to_resolve.clear()
 
     def resolve_nodes(self):
         """Resolve all dependent components with resolver and set resolved component arm id back to newly
         registered nodes. Registered nodes will be cleared after resolution.
         """
-        for component_hash, component in self._new_components.items():
-            # TODO: apply local cache controlled by an environment variable here
-            # TODO: do concurrent resolution controlled by an environment variable here
-            # given deduplication has already been done, we can safely assume that there is no
-            # conflict in concurrent local cache access
-            component_id = self._resolver(component)
-            self._component_resolution_cache[component_hash] = component_id
-
-        self._new_components.clear()
+        if self._components_to_resolve:
+            self._resolve_components_without_in_memory_cache()
 
         for component_hash, nodes in self._nodes_to_apply.items():
             for node in nodes:
--- a/sdk/ml/azure-ai-ml/azure/ai/ml/_utils/_cache_utils.py
+++ b/sdk/ml/azure-ai-ml/azure/ai/ml/_utils/_cache_utils.py
@@ -1,13 +1,17 @@
 # ---------------------------------------------------------
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # ---------------------------------------------------------
-import uuid
+import hashlib
+import logging
 from collections import defaultdict
-from typing import Union, Tuple, List
+from typing import Union, List, Dict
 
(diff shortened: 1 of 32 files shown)


zhengfeiwang (approved):

sdk/ml/azure-ai-ml/azure/ai/ml/_utils/_cache_utils.py
```diff
@@ -0,0 +1,76 @@
...
+            return cls._ORIGIN_PREFIX + component
+
+        # TODO: add hash function here
+        # for pipeline component, all dependencies of pipeline component has already been resolved
+        # so we can safely use its anonymous hash
+        # for non-pipeline component, we can't directly use anonymous hash as code hasn't been uploaded yet
+        # return a random hash by default
+        return cls._ANONYMOUS_HASH_PREFIX + str(uuid.uuid4())
+
+    def register_node_to_resolve(self, node: BaseNode):
```
wangchao1230, line 43: Is this one thread safe?

sdk/ml/azure-ai-ml/azure/ai/ml/operations/_component_operations.py
```diff
@@ -489,75 +507,153 @@ def _resolve_inputs_for_pipeline_component_jobs(self, jobs, base_path):
...
+                node.compute_name = resolver(node.compute_name, azureml_type=AzureMLResourceType.COMPUTE)
+
+    @classmethod
+    def _divide_nodes_to_resolve_into_layers(
+        cls,
+        component: PipelineComponent,
+        extra_operations: List[Callable]
+    ):
+        """Divide nodes to resolve into layers. Deeper layer nodes will be resolved
+        first so that their resolution result can be used by upper layer nodes.
```
wangchao1230, line 566: docstring on the return values

sdk/ml/azure-ai-ml/azure/ai/ml/operations/_component_operations.py
```diff
@@ -489,75 +507,153 @@ def _resolve_inputs_for_pipeline_component_jobs(self, jobs, base_path):
...
+        """
+        # add an empty layer to mark the end of the first layer
+        layers, cur_head_in_layer, cur_layer_i = [list(component.jobs.items()), []], 0, 0
+
+        while cur_layer_i < len(layers) and cur_head_in_layer < len(layers[cur_layer_i]):
```
wangchao1230, line 571: cur_layer_i -> cur_layer ?

sdk/ml/azure-ai-ml/azure/ai/ml/operations/_component_operations.py
```diff
@@ -489,75 +507,153 @@ def _resolve_inputs_for_pipeline_component_jobs(self, jobs, base_path):
...
+            ]
+        )
+
+        # cache anonymous component only for now
+        # request level in-memory cache can be a better solution for other type of assets as they are
+        # relatively simple and of small number of distinct instances
+        component_cache = CachedNodeResolver(resolver=resolver)
+        # start from the last layer and resolve layer by layer to ensure all dependencies are resolved
+        for layer in reversed(layers):
+            for _, job_instance in layer:
```
wangchao1230, line 635: can we start resolving those already leaf nodes in upper layer earlier?

sdk/ml/azure-ai-ml/azure/ai/ml/_utils/_cache_utils.py
elliotzh: Operations on attributes of `CachedNodeResolver` should be thread safe as they're not shared among threads.

Risk is on hash calculation part. The key is to make sure that the input `node` won't change on & after this function call. Here we not only resolve components layer by layer, but also register nodes layer by layer, which means that all dependencies of the input `node` has already been resolved.

elliotzh: on the other hand, if 2 nodes in the same layer share the same component, node update (`node._component = resolved_component`) will be put off till all registrations done, so still thread safe.

sdk/ml/azure-ai-ml/azure/ai/ml/operations/_component_operations.py
elliotzh: Name it in this way as it's an index instead of a "layer" (list of nodes)

sdk/ml/azure-ai-ml/azure/ai/ml/operations/_component_operations.py
elliotzh: I have considered so. The problem is that the benefit seems minor unless we can dynamically add nodes to the processing queue. If not, take below tree as an example: 
A
|  \  \
B C D
 / | \  
E F G 
by layer: 3 => 3
leaf first: 5 => 1
The round of concurrent component registration is the same.

For dynamically adding parent to the process queue once all its children have been resolved, the implementation can be more complicated and maybe we can consider it in further optimization.

sdk/ml/azure-ai-ml/azure/ai/ml/_utils/_cache_utils.py
```diff
@@ -0,0 +1,76 @@
...
+    """Class to resolve component in nodes with cached component resolution results.
+    """
+    _ANONYMOUS_HASH_PREFIX = "anonymous_component:"
+    _YAML_SOURCE_PREFIX = "yaml_source:"
+    _ORIGIN_PREFIX = "origin:"
+
+    def __init__(self, resolver):
+        self._resolver = resolver
+        self._component_resolution_cache = {}
+        self._new_components: List[Tuple[str, Component]] = []
```
brynn-code, line 23: Maybe a more meaningful name or comments? Not sure if I could get the usage by its name now.

elliotzh: Good catch! Updated.

sdk/ml/azure-ai-ml/azure/ai/ml/_utils/_cache_utils.py
wangchao1230: Yes. In the way we use this class, there is no thread-safe problem. 
- resolve_nodes & register_node_to_resolve will not be called concurrently
- nodes registered will be independent

May document the usage assumption.

sdk/ml/azure-ai-ml/azure/ai/ml/_utils/_cache_utils.py
```diff
@@ -0,0 +1,76 @@
...
+        registered nodes. Registered nodes will be cleared after resolution.
+        """
+        # TODO: apply local cache controlled by an environment variable here
+        # TODO: do concurrent resolution controlled by an environment variable here
+        # given deduplication has already been done, we can safely assume that there is no
+        # conflict in concurrent local cache access
+        # multiprocessing need to dump input objects before starting a new process, which will fail
+        # on _AttrDict for now, so put off the concurrent resolution
+        for component_hash, component in self._new_components:
+            self._component_resolution_cache[component_hash] = self._resolve_component(component)
```
wangchao1230, line 69: skip if we have already resolved this component?

sdk/ml/azure-ai-ml/azure/ai/ml/operations/_component_operations.py
```diff
@@ -489,75 +507,172 @@ def _resolve_inputs_for_pipeline_component_jobs(self, jobs, base_path):
...
-                )
 
-    def _resolve_arm_id_and_inputs(self, component):
+            for extra_operation in extra_operations:
+                extra_operation(job_instance, key)
+
+            if isinstance(job_instance, BaseNode) and isinstance(job_instance._component, PipelineComponent):
+                if cur_layer + 1 == len(layers):
+                    layers.append([])
+                layers[cur_layer+1].extend(job_instance.component.jobs.items())
```
brynn-code, line 602: Do we need to skip ConditionNode here?

elliotzh: ConditionNode is not a BaseNode so won't be processed here

sdk/ml/azure-ai-ml/azure/ai/ml/_utils/_cache_utils.py
elliotzh: if the component has already been resolved, `self._resolve_component` will directly return its `id` so there won't be duplicate remote call.

sdk/ml/azure-ai-ml/azure/ai/ml/_utils/_cache_utils.py
elliotzh: Updated docstring for the class

sdk/ml/azure-ai-ml/azure/ai/ml/_utils/_cache_utils.py
```diff
@@ -0,0 +1,126 @@
...
+    def _get_component_hash(cls, component: Component) -> str:
+        """Get a hash for a component."""
+        # For components with code, its code will be an absolute path before uploaded to blob,
+        # so we can use a mixture of its anonymous hash and its source path as its hash, in case
+        # there are 2 components with same code but different ignore files
+        # Here we can check if the component has a source path instead of check if it has code, as
+        # there is no harm to add a source path to the hash even if the component doesn't have code
+        # Note that here we assume that the content of code folder won't change during the submission
+        if component._source_path:  # pylint: disable=protected-access
+            object_hash = hashlib.sha224()
```
wangchao1230, line 53: why use sha224 not sha256?

sdk/ml/azure-ai-ml/azure/ai/ml/_utils/_cache_utils.py
```diff
@@ -0,0 +1,126 @@
...
+    1) self._resolve_nodes and self._register_node_to_resolve are not called concurrently in the same thread
+    2) self._resolve_component is only called concurrently on independent components
+        a) we have used a component hash to deduplicate components to resolve;
+        b) nodes are registered & resolved layer by layer, so all child components are already resolved
+          when we register a pipeline node;
+        c) potential shared dependencies like compute and data have been resolved before calling self._resolve_nodes.
+    """
+    _ANONYMOUS_HASH_PREFIX = "anonymous_component:"
+    _YAML_SOURCE_PREFIX = "yaml_source:"
+    _ORIGIN_PREFIX = "origin:"
```
wangchao1230, line 30: when is this origin prefix used?

wangchao1230 (approved):

sdk/ml/azure-ai-ml/azure/ai/ml/_utils/_cache_utils.py
elliotzh: no need now as str component has been resolved directly and skipped in node registration.

PR #28072, opened 2022-12-27
[ML][Pipelines]feat: apply on-disk cache for component registration in pipeline component resolution

elliotzh:
# Description

Apply on-disk cache for component registration in pipeline component resolution.


sdk/ml/azure-ai-ml/azure/ai/ml/entities/_component/component.py
```diff
@@ -23,7 +23,7 @@
     BASE_PATH_CONTEXT_KEY,
     PARAMS_OVERRIDE_KEY,
     REGISTRY_URI_FORMAT,
-    CommonYamlFields,
+    CommonYamlFields, AzureMLResourceType,
```
elliotzh, line 26: mark

sdk/ml/azure-ai-ml/azure/ai/ml/_utils/_cache_utils.py
```diff
@@ -52,76 +67,191 @@ def _get_component_hash(cls, component: Component) -> str:
...
+                path = code.path if os.path.isabs(code.path) else os.path.join(code.base_path, code.path)
+                content_hash = get_object_hash(path, code._ignore_file)  # pylint: disable=protected-access
+            object_hash.update(content_hash.encode("utf-8"))
+            return object_hash.hexdigest()
+
+    @staticmethod
+    def _get_on_disk_cache_base_dir() -> Path:
+        """Get the base path for on disk cache."""
+        from azure.ai.ml._version import VERSION
+        return Path.home().joinpath(".azureml", "azure-ai-ml", VERSION, "cache", "components")
```
wangchao1230, line 103: cache write to tmp folder?

sdk/ml/azure-ai-ml/azure/ai/ml/_utils/_cache_utils.py
```diff
@@ -52,76 +67,191 @@ def _get_component_hash(cls, component: Component) -> str:
...
+        # need to dedup disk hash first if concurrent resolution is enabled
+        for cache_content in cache_contents_to_resolve:
+            cache_content.arm_id = cls._load_from_on_disk_cache(cache_content.on_disk_hash)
+            if not cache_content.arm_id:
+                left_cache_contents_to_resolve.append(cache_content)
+
+        return left_cache_contents_to_resolve
 
+    @classmethod
+    def _save_cache_contents_to_disk(cls, cache_contents_to_resolve: List[_CacheContent]):
```
wangchao1230, line 188: why not doing save while resolved one component?

elliotzh: Hmm, I write it in this way to decouple logic of on-disk cache and in-memory cache, but seems no harm to save while resolved one component

sdk/ml/azure-ai-ml/azure/ai/ml/_utils/_cache_utils.py
```diff
@@ -3,45 +3,65 @@
...
+
+@dataclass
+class _CacheContent:
+    component_ref: Component
+    # in-memory hash assume that the code folders are not changed during the run and
+    # use the hash of code path instead of code content to simplify the calculation
+    in_memory_hash: str
+    # on-disk hash will be calculated base on code content if applicable,
+    # so it will work even if the code folders are changed among runs
+    on_disk_hash: Optional[str] = None
```
wangchao1230, line 35: hash key should include workspace/registry?

sdk/ml/azure-ai-ml/azure/ai/ml/_utils/_cache_utils.py
```diff
@@ -52,76 +91,183 @@ def _get_component_hash(cls, component: Component) -> str:
...
-                resolved_component = self._component_cache[component_hash]
-                if isinstance(node, Parallel) and isinstance(resolved_component, ParallelComponent):
-                    origin_component: ParallelComponent = node._component  # pylint: disable=protected-access
-                    if id(node.task) == id(origin_component.task):
-                        node.task.code = resolved_component.task.code
-
-                node._component = resolution_result  # pylint: disable=protected-access
-        self._nodes_to_apply.clear()
+        # all nodes will be skipped on register_node_to_resolve when resolving subgraph
+        self._lock.acquire()
```
D-W-, line 271: Since cache reading is thread safe, maybe we can add this lock before and after cache writing part?

sdk/ml/azure-ai-ml/azure/ai/ml/_utils/_cache_utils.py
```diff
@@ -52,76 +91,183 @@ def _get_component_hash(cls, component: Component) -> str:
...
         # been resolved before calling this function, so we can use their anonymous hash directly
-        return cls._ANONYMOUS_HASH_PREFIX + component._get_anonymous_hash()  # pylint: disable=protected-access
+        return _ANONYMOUS_HASH_PREFIX + component._get_anonymous_hash()  # pylint: disable=protected-access
 
-    def register_node_to_resolve(self, node: BaseNode):
-        """Register a node with its component to resolve.
+    @staticmethod
+    def _get_on_disk_hash_for_component(component: Component, in_memory_hash: str) -> str:
+        """Get a hash for a component. Will calculate the hash based on the component's code folder if
+        the component has code, otherwise will use the in-memory hash.
```
D-W-, line 102: Why the hashing logic is different for in memory and on disc hash?

sdk/ml/azure-ai-ml/azure/ai/ml/_utils/_cache_utils.py
elliotzh: Node resolution involves filling back and will change the state of nodes, e.g. hash of its inner component.

This is for concurrent pipeline job creation. Lock won't be acquired in subgraph resolution as all nodes will be skipped on calling `register_node_to_resolve`.
So, for example, if `node_a` is used in both `pipeline1` and `pipeline2` and we create the 2 pipeline jobs concurrently, `node_a._component.code` will be a local path before `_resolve_nodes` and an arm id after `_resolve_nodes`.

sdk/ml/azure-ai-ml/azure/ai/ml/operations/_component_operations.py
```diff
@@ -640,13 +679,20 @@ def _resolve_dependencies_for_pipeline_component_jobs(
...
                 self._set_default_display_name_for_anonymous_component_in_node,
+                partial(self._try_resolve_node_level_task_for_parallel_node, resolver=resolver),
             ]
         )
 
         # cache anonymous component only for now
         # request level in-memory cache can be a better solution for other type of assets as they are
         # relatively simple and of small number of distinct instances
-        component_cache = CachedNodeResolver(resolver=resolver)
+        component_cache = CachedNodeResolver(
```
D-W-, line 689: Each time when `_resolve_dependencies_for_pipeline_component_jobs` called, a `CachedNodeResolver` object will be created. Will there be issue since the lock object created multiple instances? e.g. when 2 different subgraph using same component

sdk/ml/azure-ai-ml/azure/ai/ml/_utils/_cache_utils.py
elliotzh: We can't assume that the code folder won't change among dependency resolution.

In in-memory hash calculate, we use path to calculate hash instead of folder content to avoid io operations, which is not safe across 2 `jobs.create_or_update` calls.

elliotzh: Will update docstring accordingly.

sdk/ml/azure-ai-ml/azure/ai/ml/_utils/_cache_utils.py
```diff
@@ -3,45 +3,84 @@
...
     This class is thread-safe if:
-    1) self._resolve_nodes and self._register_node_to_resolve are not called concurrently in the same thread
+    1) self._resolve_nodes is not called concurrently. We guarantee this with a lock in self.resolve_nodes.
     2) self._resolve_component is only called concurrently on independent components
-        a) we have used a component hash to deduplicate components to resolve;
+        a) we have used an in-memory component hash to deduplicate components to resolve first;
         b) nodes are registered & resolved layer by layer, so all child components are already resolved
           when we register a pipeline node;
-        c) potential shared dependencies like compute and data have been resolved before calling self._resolve_nodes.
+        c) potential shared dependencies like compute and data have been resolved before
```
D-W-, line 49: All dependencies except code will be resolved before calling self._resolve_component?

elliotzh: With current implementation, all dependencies except code and `environment`.

But the statement is not correct. resolution of shared dependencies is thread-safe given they don't involve hash calculation.

sdk/ml/azure-ai-ml/tests/dsl/e2etests/test_dsl_pipeline.py
```diff
@@ -1471,14 +1475,14 @@ def pipeline_distribution_components(job_in_number, job_in_path):
             hello_world_component_mpi.distribution.process_count_per_instance = 2
 
         pipeline: PipelineJob = pipeline_distribution_components(10, job_input)
-        assert mpi_func._is_anonymous is False
```
D-W-, line 1474: Why removed this?

sdk/ml/azure-ai-ml/azure/ai/ml/operations/_component_operations.py
elliotzh: Good catch! Let me reconsider where to claim the lock. Maybe just make the lock global

sdk/ml/azure-ai-ml/azure/ai/ml/_internal/entities/component.py
```diff
@@ -223,11 +224,22 @@ def _get_snapshot_id(
...
+        # otherwise, if there is no local code, yield super()._resolve_local_code() and return early
+        if self.code is not None:
+            with super()._resolve_local_code() as code:
+                if not isinstance(code, Code) or code._is_remote:
+                    yield code
+                    return
+
+        # This is forbidden by schema CodeFields for now so won't happen.
+        if isinstance(self.code, Code):
+            raise ValueError("Code is a private class and can't be set to component.code directly.")
```
D-W-, line 242: Why do we need this exception here? Will there be cases when `_resolve_local_code` called on component with code already resolved to `Code` object?

sdk/ml/azure-ai-ml/tests/dsl/e2etests/test_dsl_pipeline.py
elliotzh: 1. Given on-disk cache is enabled within a test, arm if of `mpi_func` will be from cache directly, while `mpi_func._is_anonymous` is set to `True` during `components.create_or_update`
2. `_is_anonymous` is a private implementation, so I change it to check public interface instead (whether it's a created anonymous component in the created job)

sdk/ml/azure-ai-ml/tests/conftest.py
```diff
@@ -538,6 +539,29 @@ def generate_compononent_hash(*args, **kwargs):
             side_effect=generate_compononent_hash
         )
 
+    if is_live():
+        # Disable on-disk cache when doing live recording as recorded requests may be impacted by
+        # the order to run tests. If we have enabled on-disk cache, tests run later will use the
+        # cached result from tests run earlier, so we won't found enough recordings when running
+        # tests in a different order.
+        if not is_live_and_not_recording():
```
D-W-, line 547: Seems this change only takes affect when someone run tests without recording locally then playback. Do we need some tests to guard it's behavior?

sdk/ml/azure-ai-ml/azure/ai/ml/_internal/entities/component.py
elliotzh: 1. Hmm... You're right, we don't need this exception here. Let me yield `self.code` and return directly.
2. No. With current logic, component.code will be resolved into an arm id directly. `Code` is just a temporarily used class in
```python
with component._resolve_local_code() as code:
    component.code: str = resolver(code)
```

sdk/ml/azure-ai-ml/azure/ai/ml/operations/_component_operations.py
```diff
@@ -640,13 +679,20 @@ def _resolve_dependencies_for_pipeline_component_jobs(
             component,
```
D-W-, line 679: What's the benefit for divide nodes into layers compared to in-order traversal?

elliotzh: It's a preparation for concurrent node resolution. Nodes in the same layer can be resolved concurrently if lower layers have already been resolved.

sdk/ml/azure-ai-ml/tests/conftest.py
elliotzh: 1. It will still take effect within a test if multiple `jobs.create_or_update` calls are made.
2. Will add some specific tests for this.

sdk/ml/azure-ai-ml/azure/ai/ml/operations/_component_operations.py
```diff
@@ -640,13 +679,20 @@ def _resolve_dependencies_for_pipeline_component_jobs(
...
         # request level in-memory cache can be a better solution for other type of assets as they are
         # relatively simple and of small number of distinct instances
-        component_cache = CachedNodeResolver(resolver=resolver)
+        component_cache = CachedNodeResolver(
+            resolver=resolver,
+            subscription_id=self._subscription_id,
+            resource_group_name=self._resource_group_name,
+            workspace_name=self._workspace_name,
+            registry_name=self._registry_name,
+        )
```
D-W-, line 695: Have a question for L709, will environment be resolved to arm id before checking cache?

sdk/ml/azure-ai-ml/tests/conftest.py
elliotzh: BTW, our daily live test for `pipeline_test` is live and not recording, right?

sdk/ml/azure-ai-ml/azure/ai/ml/operations/_component_operations.py
elliotzh: No. As explained [here](https://github.com/Azure/azure-sdk-for-python/pull/28072/files/1434e8ff747b6e6d89cccf7d0d2d9f677e5da8fb#r1058709107), it should be thread-safe.

elliotzh: But such dependencies resolution will impact the efficiency of cache hit across layers, so not sure whether we should guarantee that all other dependencies are resolved in extra operations

sdk/ml/azure-ai-ml/azure/ai/ml/operations/_component_operations.py
elliotzh: To further explain, we'd better not to resolve a pipeline component and one of its jobs concurrently, as hash calculation of the pipeline component will be impacted by the resolution of its children.

sdk/ml/azure-ai-ml/tests/conftest.py
elliotzh: Updated the fixture to avoid disable on-disk cache. Clear cache conditionally instead.

sdk/ml/azure-ai-ml/azure/ai/ml/operations/_component_operations.py
elliotzh: Interesting... Resolution of environment will be skipped in some cases in [early recording](https://github.com/Azure/azure-sdk-for-python/blob/ca26a259193d985b65e9e181a554e1529ccf5751/sdk/ml/azure-ai-ml/tests/recordings/dsl/e2etests/test_dsl_pipeline.pyTestDSLPipelinetest_distribution_components.json#L343) (as early as 11.14)

sdk/ml/azure-ai-ml/azure/ai/ml/_utils/_cache_utils.py
```diff
@@ -52,76 +122,187 @@ def _get_component_hash(cls, component: Component) -> str:
...
-        # avoid duplicate hash calculation. The risk is that we haven't implicitly forbidden component
-        # modification after it's been used in a node, and we can't guarantee that the hash is still
-        # valid after modification.
-        component_hash = self._get_component_hash(component)
-        if component_hash not in self._component_cache:
-            self._hash_of_components_to_resolve.append(component_hash)
-            self._component_cache[component_hash] = component
+    def _load_from_on_disk_cache(self, on_disk_hash: str) -> Optional[str]:
+        """Load component arm id from on disk cache."""
+        on_disk_cache_path = self._get_on_disk_cache_path(on_disk_hash)
```
elliotzh, line 175: mark: should disk cache expire? Or should we do some specific error handling to let user disable on-disk cache?

sdk/ml/azure-ai-ml/azure/ai/ml/operations/_component_operations.py
elliotzh: Got the reason. updated in latest commit

sdk/ml/azure-ai-ml/tests/dsl/e2etests/test_dsl_pipeline.py
elliotzh: Mixed the commented line. This assertion can be kept until we change related implementation. Rolled back this change.

sdk/ml/azure-ai-ml/tests/conftest.py
elliotzh: Misunderstood your question. I think the mentioned change has been merged with below and is well tested in ci now

D-W- (approved):

sdk/ml/azure-ai-ml/azure/ai/ml/_utils/_cache_utils.py
```diff
@@ -3,45 +3,115 @@
...
-        """Resolve a component with self._resolver."""
-        return self._resolver(component, azureml_type=AzureMLResourceType.COMPONENT)
-
-    @classmethod
-    def _get_component_hash(cls, component: Component) -> str:
-        """Get a hash for a component."""
+        self._cache: Dict[str, _CacheContent] = {}
+        self._nodes_to_resolve: List[BaseNode] = []
+
+        self._subscription_id = subscription_id
```
wangchao1230, line 96: seems this is not used in this class?

elliotzh: yes, just for recording actually

elliotzh: removed.

wangchao1230 (approved):

Issue #28084, opened 2022-12-27
[ACR] Pytest cannot collect all tests


YalinLi0312:
```
PytestCollectionWarning: cannot collect test class 'TestContainerRegistryClientAsync' because it has a __init__ constructor (from: sdk/containerregistry/azure-containerregistry/tests/test_container_registry_client_async.py)
```
Pytest only collect test prefixed test functions or methods inside Test prefixed test classes, the test class should not include an `__init___ `method. We added an `__init__` method in ContainerRegistryTestClass to make sure all test resources are ready before running all tests, it should move to another function and delete the `__init__`.

PR #28085, opened 2022-12-27
[ACR] Fix pytest collect issue

YalinLi0312:
Resolves: https://github.com/Azure/azure-sdk-for-python/issues/28084


sdk/containerregistry/azure-containerregistry/tests/testcase.py
```diff
@@ -186,6 +179,9 @@ def load_registry():
     if not is_live():
         return
     authority = get_authority(os.environ.get("CONTAINERREGISTRY_ENDPOINT"))
+    authority_anon = get_authority(os.environ.get("CONTAINERREGISTRY_ANONREGISTRY_ENDPOINT"))
+    registry_name = os.environ["CONTAINERREGISTRY_REGISTRY_NAME"]
```
mccoyp, line 183: nit: consider using `os.environ.get()` instead of `os.environ[]` here (and on the following line) to align with the lines above. That would also make it possible to run non-`anon` tests without setting the anonymous registry variables

mccoyp (approved):

sdk/containerregistry/azure-containerregistry/tests/testcase.py
YalinLi0312: Thanks, have updated!

Issue #27044, opened 2022-10-25
[Tables] Update to Python 3 inline type hints


annatisch:
We're still using the old backwards-compatible type hints, e.g.:
https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/tables/azure-data-tables/azure/data/tables/_table_client.py#L375-L380

Issue #27729, opened 2022-11-30
Update method signature to use Python3 syntax


YalinLi0312:
- [x] ACR
- [x] AppConfig
- [x] Tables

PR #28095, opened 2022-12-28
[Tables] Update method type hints to Python3 syntax

YalinLi0312:
Resolves: https://github.com/Azure/azure-sdk-for-python/issues/27044
Also for part of issue: https://github.com/Azure/azure-sdk-for-python/issues/27729


sdk/tables/azure-data-tables/azure/data/tables/_table_batch.py
```diff
@@ -15,18 +14,15 @@
...
 from ._models import UpdateMode, TransactionOperation
 from ._serialize import _get_match_headers, _add_entity_properties, _prepare_key
 from ._entity import TableEntity
-
-if TYPE_CHECKING:
-    from azure.core.pipeline.transport import HttpRequest
-    from ._generated import models, AzureTable
-    from ._generated._configuration import AzureTableConfiguration
-    from ._generated._serialization import Serializer
-    from ._generated._serialization import Deserializer
```
annatisch, line 29: Is there a reason you moved all these out of `TYPE_CHECKING`?

sdk/tables/azure-data-tables/azure/data/tables/_table_client.py
```diff
@@ -19,42 +20,30 @@
...
-    TableItem
-)
+from ._models import TableEntityPropertiesPaged, UpdateMode, TableAccessPolicy, TableItem
 
 if TYPE_CHECKING:
     from azure.core.credentials import AzureNamedKeyCredential, AzureSasCredential
 
 
-class TableClient(TablesBaseClient): # pylint: disable=client-accepts-api-version-keyword
+class TableClient(TablesBaseClient):  # pylint: disable=client-accepts-api-version-keyword
```
annatisch, line 36: Could we open an issue to fix this pylint disable statement? Our client constructors should have api_version.

sdk/tables/azure-data-tables/azure/data/tables/_table_client.py
```diff
@@ -19,42 +20,30 @@
...
 
     :ivar str account_name: The name of the Tables account.
     :ivar str table_name: The name of the table.
     :ivar str url: The full URL to the Tables account.
     """
 
-    def __init__(self, endpoint: str, table_name: str, **kwargs: Any) -> None: # pylint: disable=missing-client-constructor-parameter-credential
+    def __init__(
+        self, endpoint: str, table_name: str, **kwargs: Any
+    ) -> None:  # pylint: disable=missing-client-constructor-parameter-credential
```
annatisch: Can we please update the constructor to fix this pylint diable statement?
`def __init__(self, endpoint: str, table_name: str, * credential: Union[...] = None, **kwargs: Any) -> None`

sdk/tables/azure-data-tables/azure/data/tables/_deserialize.py
```diff
@@ -46,18 +46,15 @@ def get_enum_value(value):
         return value
 
 
-def _from_entity_binary(value):
-    # type: (str) -> EntityProperty
+def _from_entity_binary(value: str) -> EntityProperty:
```
annatisch, line 49: I'm not sure if this is actually returning an EntityProperty?

sdk/tables/azure-data-tables/azure/data/tables/_table_batch.py
YalinLi0312: * Python3 has improved the effiency of import, not worth to have `TYPE_CHECKING`
* Mypy cannot detect the types import under `TYPE_CHECKING`, have to quoted these types in hints

Based on these two reasons, I decided to remove all `TYPE_CHECKING` checks in this package.

sdk/tables/azure-data-tables/azure/data/tables/_table_client.py
YalinLi0312: I'll update but without default value `None`, because Pylance complains `parameter with "*" or "**" cannot have default value`.

YalinLi0312: I think it should have a comma after `*`, like in async: https://github.com/Azure/azure-sdk-for-python/blob/5eef0a2f55d7fa12894b7ace041d11087d72a268/sdk/tables/azure-data-tables/azure/data/tables/aio/_table_client_async.py#L52-L59

sdk/tables/azure-data-tables/azure/data/tables/_deserialize.py
YalinLi0312: It should be `bytes`

sdk/tables/azure-data-tables/azure/data/tables/_table_client.py
annatisch: Yup - sorry that was my typo :D

sdk/tables/azure-data-tables/azure/data/tables/_table_client.py
YalinLi0312: Logged an issue: https://github.com/Azure/azure-sdk-for-python/issues/28115

sdk/tables/azure-data-tables/azure/data/tables/_table_batch.py
annatisch: Sounds good.

sdk/tables/azure-data-tables/azure/data/tables/_table_client.py
YalinLi0312: Pylint still failing with the same error `missing-client-constructor-parameter-credential` with the fix. Should we move `credential` parameters instead of keywords? @annatisch

YalinLi0312: I have moved `credential` to params, please take a look. @annatisch

kristapratico: Yeah I don't think that pylint rule is smart enough to check for `credential` as a keyword-only parameter so you'll need to continue to ignore it for now. @annatisch - do we want to update the rule to let this pass?

sdk/tables/azure-data-tables/azure/data/tables/_base_client.py
```diff
@@ -55,17 +57,12 @@
...
-    from typing import Literal
-
 _SUPPORTED_API_VERSIONS = ["2019-02-02", "2019-07-07", "2020-12-06"]
 # cspell:disable-next-line
 _DEV_CONN_STRING = "DefaultEndpointsProtocol=http;AccountName=devstoreaccount1;AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;TableEndpoint=http://127.0.0.1:10002/devstoreaccount1" # pylint: disable=line-too-long
 
 
-def get_api_version(kwargs, default):
-    # type: (Dict[str, Any], Literal["2019-02-02"]) -> Literal["2019-02-02"]
+def get_api_version(kwargs: Dict[str, Any], default: Literal["2019-02-02"]) -> Literal["2019-02-02"]:
```
annatisch, line 65:
```suggestion
def get_api_version(kwargs: Dict[str, Any], default: Literal["2019-02-02"]) -> str:
```
If the user has overridden the api_version, we don't know what the value will be.

sdk/tables/azure-data-tables/azure/data/tables/_models.py
```diff
@@ -448,8 +417,8 @@ def from_string(
         :param str permission: Specify permissions in
             the string with the first letter of the word.
         :keyword callable cls: A custom type or function that will be passed the direct response
```
annatisch, line 419: Could we remove this line? It's causing some weirdness in APIview

sdk/tables/azure-data-tables/azure/data/tables/_table_client.py
```diff
@@ -54,13 +39,12 @@ class TableClient(TablesBaseClient):
...
+    def __init__( # pylint: disable=missing-client-constructor-parameter-credential
         self,
-        endpoint,  # type: str
-        table_name,  # type: str
-        **kwargs  # type: Any
-    ):
-        # type: (...) -> None
+        endpoint: str,
+        table_name: str,
+        **kwargs
```
annatisch, line 46: Could we add `credential` here using the keyword-only syntax, like we do for the async TableClient? That should improve the APIview.

sdk/tables/azure-data-tables/azure/data/tables/_table_client.py
```diff
@@ -114,18 +92,20 @@ def from_connection_string(
...
+        endpoint, credential = parse_connection_str(conn_str=conn_str, credential=None, keyword_args=kwargs)
         return cls(endpoint, table_name=table_name, credential=credential, **kwargs)
 
     @classmethod
-    def from_table_url(cls, table_url, **kwargs):
-        # type: (str, Any) -> TableClient
+    def from_table_url(
+        cls,
+        table_url: str,
+        credential: Optional[Union[AzureNamedKeyCredential, AzureSasCredential]] = None,
```
annatisch, line 102: Looks like we lost the keyword syntax here:
```suggestion
        *,
        credential: Optional[Union[AzureNamedKeyCredential, AzureSasCredential]] = None,
```

sdk/tables/azure-data-tables/azure/data/tables/_models.py
YalinLi0312: I searched cls docstring across services, don't see any service add line for cls. I think we can remove this line, will update in next commit.

sdk/tables/azure-data-tables/azure/data/tables/_base_client.py
```diff
@@ -211,14 +214,13 @@ def api_version(self):
...
+    def __init__(
         self,
-        endpoint,  # type: str
-        **kwargs  # type: Any
-    ):
-        # type: (...) -> None
-        credential = kwargs.pop('credential', None)
-        super(TablesBaseClient, self).__init__(endpoint, credential=credential, **kwargs)
+        endpoint: str,
+        credential: Optional[Union[AzureSasCredential, AzureNamedKeyCredential, TokenCredential]] = None,
```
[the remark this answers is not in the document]
YalinLi0312: Updated in the last commit.

sdk/tables/azure-data-tables/azure/data/tables/_base_client.py
YalinLi0312: Updated!

sdk/tables/azure-data-tables/azure/data/tables/_table_client.py
YalinLi0312: Have changed to be the same in async client.

sdk/tables/azure-data-tables/azure/data/tables/_table_client.py
YalinLi0312: Updated!

sdk/tables/azure-data-tables/azure/data/tables/_base_client.py
YalinLi0312: Well, mypy complains incompatible types in assignment with the change.

sdk/tables/azure-data-tables/azure/data/tables/_base_client.py
```diff
@@ -80,11 +76,10 @@ def get_api_version(kwargs, default):
 class AccountHostsMixin(object):  # pylint: disable=too-many-instance-attributes
     def __init__(
         self,
-        account_url,  # type: Any
-        credential=None,  # type: Optional[Union[AzureNamedKeyCredential, AzureSasCredential, "TokenCredential"]]
-        **kwargs  # type: Any
-    ):
-        # type: (...) -> None
+        account_url: Any,
```
annatisch, line 79:
```suggestion
        account_url: str,
```

sdk/tables/azure-data-tables/azure/data/tables/_base_client.py
```diff
@@ -265,23 +260,23 @@ def _configure_policies(self, **kwargs):
...
+            self._credential_policy = AzureSasCredentialPolicy(credential) # type: ignore
         elif isinstance(credential, AzureNamedKeyCredential):
-            self._credential_policy = SharedKeyCredentialPolicy(credential)  # type: ignore
+            self._credential_policy = SharedKeyCredentialPolicy(credential) # type: ignore
         elif credential is not None:
             raise TypeError("Unsupported credential: {}".format(credential))
 
-    def _batch_send(self, table_name, *reqs, **kwargs):
-        # type: (str, List[HttpRequest], Any) -> List[Mapping[str, Any]]
+    def _batch_send(self, table_name: str, *reqs: List[HttpRequest], **kwargs) -> List[Mapping[str, Any]]:
```
annatisch, line 279: I _think_ this should be:
```suggestion
    def _batch_send(self, table_name: str, *reqs: HttpRequest, **kwargs) -> List[Mapping[str, Any]]:
```

annatisch (approved):
Looks good - just a few small things :)

sdk/tables/azure-data-tables/azure/data/tables/_base_client.py
YalinLi0312: Yes, it's not a list of request based on the type hint in core: https://github.com/Azure/azure-sdk-for-python/blob/b83938fe8a238081c0a1fc41cc6f3b07aaed0595/sdk/core/azure-core/azure/core/pipeline/transport/_base.py#L315

PR #28098, opened 2022-12-28
[Tables] Remove unnecessary python version checks

YalinLi0312:
Since we only support python 3.7 and above, some version checks are not necessary now.

### DIFF
--- a/sdk/tables/azure-data-tables/azure/data/tables/_error.py
+++ b/sdk/tables/azure-data-tables/azure/data/tables/_error.py
@@ -26,7 +26,6 @@ _ERROR_TYPE_NOT_SUPPORTED = "Type not supported when sending data to the service
 _ERROR_VALUE_TOO_LARGE = "{0} is too large to be cast to type {1}."
 _ERROR_UNKNOWN = "Unknown error ({0})"
 _ERROR_VALUE_NONE = "{0} should not be None."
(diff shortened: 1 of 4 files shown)


sdk/tables/azure-data-tables/azure/data/tables/_error.py
```diff
@@ -51,15 +50,8 @@ def _wrap_exception(ex, desired_type):
...
     if len(ex.args) > 0:
         msg = ex.args[0]
-    if sys.version_info >= (3,):
-        # Automatic chaining in Python 3 means we keep the trace
-        return desired_type(msg)
-
-    # There isn't a good solution in 2 for keeping the stack trace
-    # in general, or that will not result in an error in 3
-    # However, we can keep the previous error type and message
+    return desired_type(msg)
```
annatisch, line 53: Where is the `_wrap_exception` function used?

annatisch (approved):

sdk/tables/azure-data-tables/azure/data/tables/_error.py
YalinLi0312: In _authentication.py: https://github.com/Azure/azure-sdk-for-python/blob/a1c902e000ddc0a20e64ed9bf51bcdf9a6bb255c/sdk/tables/azure-data-tables/azure/data/tables/_authentication.py#L151

Issue #26251, opened 2022-09-15
[Monitor Query] Add samples/docs for running big queries and overcoming API limits for size and record count


scottaddie:
For details of the ask, see https://dev.azure.com/devdiv/DevDiv/_workitems/edit/1597918.

PR #28148, opened 2023-01-03
[Monitor][Query] Add sample notebook for large queries

pvaneck:
A sample notebook was added to demonstrate how a user can overcome Log Analytics service limits when trying to query large amounts of data.

Closes: https://github.com/Azure/azure-sdk-for-python/issues/26251


sdk/monitor/azure-monitor-query/samples/sample_large_query.ipynb
```diff
@@ -0,0 +1,630 @@
...
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Query with large result sets\n",
+    "\n",
+    "This sample notebook demonstrates how to query large amounts of data using the Azure Monitor Query client library.\n",
+    "\n",
+    "Due to Log Analytics [service limits](https://learn.microsoft.com/en-us/azure/azure-monitor/service-limits#la-query-api), sometimes it may not be possible to retrieve all the expected data in a single query. For example, the number of rows returned or the maximum size of the data returned may exceed the stated limits. One approach for overcoming these limits is to split the queries into multiple smaller queries using different time ranges.\n",
```
scottaddie, line 12:
```suggestion
    "Due to Log Analytics [service limits](https://learn.microsoft.com/azure/azure-monitor/service-limits#la-query-api), sometimes it may not be possible to retrieve all the expected data in a single query. For example, the number of rows returned or the maximum size of the data returned may exceed the stated limits. One approach for overcoming these limits is to split the queries into multiple smaller queries using different time ranges.\n",
```

sdk/monitor/azure-monitor-query/samples/notebooks/sample_large_query.ipynb
```diff
@@ -0,0 +1,606 @@
...
+    "                workspace_id=LOGS_WORKSPACE_ID,\n",
+    "                query=query,\n",
+    "                timespan=(start_time, end_time),\n",
+    "                server_timeout=360,                 \n",
+    "                include_statistics=False, # Can be used for debugging.\n",
+    "                headers=headers\n",
+    "            )\n",
+    "        except HttpResponseError as e:\n",
+    "            print(f\"Error when attempting query {query_id} (query time span: {start_time} to {end_time}):\\n\\t\", e)\n",
+    "            if i < retries:\n",
```
johanste, line 239: Why are we retrying outside of the library method? In general, we want to pass in retry information to the method itself (e.g. max_retries to the client.query_workspace method). Double retry loops tend to lead to very long and (for the user) unpredictable waits...

sdk/monitor/azure-monitor-query/samples/notebooks/sample_large_query.ipynb
```diff
@@ -0,0 +1,592 @@
...
+    "    batch_endpoints += [row[0] for row in response.tables[0].rows]\n",
+    "    return batch_endpoints"
+   ]
+  },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Next, define a function that will asynchronously execute a given query over a time range specified by a given start time and end time. This function will call the `query_workspace` method of the `LogsQueryClient`. Internally, Azure Core will handle retries in case of connection-related errors or server errors (i.e. 500, 503, and 504 status codes). Check [here](https://github.com/Azure/azure-sdk-for-python/tree/main/sdk/core/azure-core#configurations) for more information on configuring retries."
```
kashifkhan, line 194: We can just say that the monitor query client library itself will handle retries in case of connection-related errors etc. and link out to core as you have :)

pvaneck: Good idea, that sounds better.

kashifkhan (approved):

PR #28156, opened 2023-01-04
Webpubsub client 

msyyc:
Champion scenario with sync client
 - [hello_world](https://github.com/Azure/azure-sdk-for-python/blob/15f2e754307f2457db19fb1b1b01e92391f7fee0/sdk/webpubsub/azure-webpubsub-client/samples/hello_world.py)

Pending
- more tests (more for mock on error scenario)
- embedded sample in doc
- distributed tracing
- async client

Arch review
- https://github.com/Azure/azure-sdk/issues/5646

WebSocket
- Use open-source package [websocket-client](https://github.com/websocket-client/websocket-client)

auto-apiview: https://github.com/Azure/azure-sdk-for-python/pull/28156#issuecomment-1411525090
history apiview: https://apiview.dev/Assemblies/Review/e23e634cecdd4793811106492ab03942


sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_models.py
```diff
@@ -0,0 +1,567 @@
...
+
+class WebPubSubClientProtocol:
+    def __init__(self) -> None:
+        self.is_reliable_sub_protocol = False
+        self.name = ""
+
+    @staticmethod
+    def parse_messages(
+        raw_message: str,
+    ) -> Union[ConnectedMessage, DisconnectedMessage, GroupDataMessage, ServerDataMessage, AckMessage]:
```
changlong-liu, line 341: Will it be beneficial to define a base class for these xxMessage? so that the return type here can just be the base class

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_client.py
```diff
@@ -0,0 +1,760 @@
...
+            i = 0
+            while i < 30 or self._is_stopping:
+                try:
+                    self._connect(recovery_url)
+                    return
+                except:
+                    delay(1000)
+                i = i + 1
+
+            _LOGGER.warning("Recovery attempts failed more then 30 seconds or the client is stopping")
```
changlong-liu, line 514: nit: 30 seconds => 30 times

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_client.py
```diff
@@ -0,0 +1,760 @@
...
+            )
+        try:
+            self._send_message(message)
+        except Exception as e:
+            self._ack_map.pop(ack_id)
+            raise e
+
+        # wait for ack from service
+        with self._ack_map[ack_id].cv:
+            self._ack_map[ack_id].cv.wait(60.0)
```
changlong-liu, line 187: I'm wondering whether we need to make magic numbers like 60 here to be configurable for the end user.
One solution is just define them as client variable and be configurable through a `**kwargs` in client.__init__(...)

msyyc: Make sense. I think client variable solution is better.

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_models.py
msyyc: I ever had a try like you said but it will cause mypy error.

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_models.py
```diff
@@ -0,0 +1,569 @@
...
+        self.message = message
+
+
+class GroupDataMessage:
+    def __init__(
+        self,
+        data_type: WebPubSubDataType,
+        data: Any,
+        group: str,
+        from_user_id: str,
```
zackliu, line 248: `from_user_id` should be Optional

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_models.py
```diff
@@ -0,0 +1,569 @@
...
+    ) -> None:
+        self.kind: Literal["sendToGroup"] = "sendToGroup"
+        self.data_type = data_type
+        self.data = data
+        self.group = group
+        self.no_echo = no_echo
+        self.ack_id = ack_id
+
+
+class OnRestoreGroupFailedArgs:
```
zackliu, line 284: In js and C# we use term `rejoinGroup` instead

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_models.py
```diff
@@ -0,0 +1,569 @@
...
+    if data_type == "json":
+        return data
+    if data_type in ("binary", "protobuf"):
+        if isinstance(data, (bytes, bytearray)):
+            return data
+        return bytes(base64.b64decode(data))
+    raise TypeError(f"Unsupported dataType: {data_type}")
+
+
+class WebPubSubClientProtocol:
```
zackliu, line 333: I'm not sure whether the signature here is extendible for protobuf subprotocol. C# protobuf part: https://github.com/zackliu/azure-sdk-for-net/blob/wps-client-protobuf-1/sdk/webpubsub/Azure.Messaging.WebPubSub.Client.Protobuf/Azure.Messaging.WebPubSub.Client.Protobuf/WebPubSubReliableProtobufProtocol.cs

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_models.py
```diff
@@ -0,0 +1,569 @@
...
+                    data=data, data_type=message["dataType"], sequence_id=message.get("sequenceId")
+                )
+            raise Exception(f"wrong message from: {message['from']}")
+        if message["type"] == "ack":
+            return AckMessage(
+                ack_id=message["ackId"],
+                success=message["success"],
+                error=message.get("error"),
+            )
+        raise Exception(f"wrong message type: {message['type']}")
```
zackliu, line 380: Consider returning null and log instead of throw Exception for unknown type or event for forward compatibility. Only throw when parsing has error

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_models.py
```diff
@@ -0,0 +1,569 @@
...
+            raise Exception("No input")
+        if not isinstance(raw_message, str):
+            raise Exception("Invalid input for JSON hub protocol. Expected a string.")
+
+        message = json.loads(raw_message)
+        if message["type"] == "system":
+            if message["event"] == "connected":
+                return ConnectedMessage(
+                    connection_id=message["connectionId"],
+                    user_id=message["userId"],
```
zackliu, line 352: `userId` is optional

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_models.py
```diff
@@ -0,0 +1,569 @@
...
+                return DisconnectedMessage(message=message["message"])
+            raise Exception(f"wrong message event type: {message['event']}")
+        if message["type"] == "message":
+            if message["from"] == "group":
+                data = parse_payload(message["data"], message["dataType"])
+                return GroupDataMessage(
+                    data_type=message["dataType"],
+                    data=data,
+                    group=message["group"],
+                    from_user_id=message["fromUserId"],
```
zackliu, line 365: `fromUserId` is optional

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_models.py
```diff
@@ -0,0 +1,569 @@
...
+        message = json.loads(raw_message)
+        if message["type"] == "system":
+            if message["event"] == "connected":
+                return ConnectedMessage(
+                    connection_id=message["connectionId"],
+                    user_id=message["userId"],
+                    reconnection_token=message.get("reconnectionToken"),
+                )
+            if message["event"] == "disconnected":
+                return DisconnectedMessage(message=message["message"])
```
zackliu, line 356: `message` is optional

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_models.py
```diff
@@ -0,0 +1,569 @@
...
+        self.max_retry_delay_in_ms = max_retry_delay_in_ms
+        self.mode = mode
+
+
+class WebPubSubClientOptions:
+    def __init__(
+        self,
+        protocol: Optional[WebPubSubClientProtocol] = None,
+        auto_reconnect: Optional[bool] = None,
+        auto_restore_groups: Optional[bool] = None,
```
zackliu, line 444: `auto_rejoin_groups`

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_model_base.py
```diff
@@ -0,0 +1,714 @@
...
+    """Converts a datetime.(datetime|date|time|timedelta) object into an ISO 8601 formatted string
+
+    :param timedelta dt: The date object to convert
+    :rtype: str
+    :return: ISO8601 version of this datetime
+    """
+    # First try datetime.datetime
+    if hasattr(dt, "year") and hasattr(dt, "hour"):
+        dt = typing.cast(datetime, dt)
+        # astimezone() fails for naive times in Python 2.7, so make make sure dt is aware (tzinfo is set)
```
kashifkhan, line 117: Since the library supports py 3.7+ is there a need for py 2 compatibility?

msyyc: It shall be residual code for py2, and I will remove it.

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_models.py
msyyc: we could override `parse_message` and `write_message` in protobuf subprotocol like:
```python
class X:
    @staticmethod
    def test():
        print("X")

class Y:
    @staticmethod
    def test():
        print("Y")

Y().test()
```

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_client.py
```diff
@@ -0,0 +1,779 @@
...
+            self._state = WebPubSubClientState.STOPPED
+            self._is_stopping = False
+            raise e
+
+    def stop(self) -> None:
+        """stop the client"""
+
+        if self._state == WebPubSubClientState.STOPPED or self._is_stopping:
+            return
+        self._is_stopping = True
```
zackliu, line 605: Is putting `self._is_stopping = False` in `_handle_connection_stopped` a right position? If call `stop()` first, then `start()`  can't invoke `_handle_connection_stopped` and there's no chance to change back the field.

msyyc: I follow similar logic with js SDK: https://github.com/Azure/azure-sdk-for-js/blob/6a76e6ac3f67fdf7b2d22c7b3a242538b0c60f54/sdk/web-pubsub/web-pubsub-client/src/webPubSubClient.ts#L759-L763

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_client.py
```diff
@@ -0,0 +1,776 @@
...
+            elif parsed_message.kind == "disconnected":
+                handle_disconnected_message(parsed_message)
+            elif parsed_message.kind == "ack":
+                handle_ack_message(parsed_message)
+            elif parsed_message.kind == "groupData":
+                handle_group_data_message(parsed_message)
+            elif parsed_message.kind == "serverData":
+                handle_server_data_message(parsed_message)
+            else:
+                _LOGGER.warning("unknown message type: {}", parsed_message.kind)
```
kashifkhan, line 505: Youll want to change the this to the %s style of formatting for loggers, so that its evaluation can be [deferred ](https://docs.python.org/3/howto/logging.html#optimization)

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_client.py
```diff
@@ -0,0 +1,776 @@
...
+
+                recovery_url = self._build_recovery_url()
+                if not recovery_url:
+                    _LOGGER.warning("Connection id or reconnection token is not available")
+                    self._handle_connection_close_and_no_recovery()
+                    return
+
+                self._state = WebPubSubClientState.RECOVERING
+                i = 0
+                while i < 30 or self._is_stopping:
```
kashifkhan, line 538: is the value for retries to connect something that can be brought in from a retry policy ?

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_model_base.py
```diff
@@ -0,0 +1,706 @@
...
+__all__ = ["NULL", "AzureJSONEncoder", "Model", "rest_field", "rest_discriminator"]
+
+
+class _Null(object):
+    """To create a Falsy object"""
+
+    def __bool__(self):
+        return False
+
+    __nonzero__ = __bool__  # Python2 compatibility
```
kashifkhan, line 41: shouldnt need this anymore as this is py3+

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_model_base.py
```diff
@@ -0,0 +1,706 @@
...
+            seconds_string = seconds_string.rstrip("0")
+    except AttributeError:  # int.is_integer() raises
+        seconds_string = "{:02}".format(seconds)
+
+    time_str += "{}S".format(seconds_string)
+
+    return "P" + date_str + time_str
+
+
+def _datetime_as_isostr(dt: typing.Union[datetime, date, time, timedelta]) -> str:
```
kashifkhan, line 107: Is this equivalent to [`_convert_to_isoformat`](https://github.com/Azure/azure-sdk-for-python/blob/faa1d1a32912e1c508948bc613508fab283e367a/sdk/core/azure-core/azure/core/utils/_utils.py#L48) in azure.core.utils ? There might be other functions in there that can be used too

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_client.py
msyyc: updated

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_model_base.py
msyyc: The .py is same with https://github.com/Azure/autorest.python/blob/main/packages/autorest.python/autorest/codegen/templates/model_base.py.jinja2 which will be moved into `azure.core` finally. So I think we can just stay same and optimize it gradually in the future.

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_model_base.py
msyyc: same with https://github.com/Azure/azure-sdk-for-python/pull/28156#discussion_r1118233645

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_client.py
```diff
@@ -0,0 +1,776 @@
...
+    @distributed_trace
+    def stop(self) -> None:
+        """stop the client"""
+
+        if self._state == WebPubSubClientState.STOPPED or self._is_stopping:
+            return
+        self._is_stopping = True
+
+        # we can't use self._ws.close otherwise on_close may not be triggered
+        if self._ws and self._ws.sock:
```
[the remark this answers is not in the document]
msyyc: I make https://github.com/websocket-client/websocket-client/issues/899 to track

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_client.py
msyyc: It is related with [connection-recovery](https://github.com/Azure/azure-webpubsub/blob/main/protocols/client/client-spec.md#14-connection-recovery) which is different with message_retry_policy and reconnect_retry_policy. In one word, service will keep data for 30 seconds if the client reconnects after drop accidently. To make it easier to maintain, I update with internal global variables instead of number.

msyyc:
main updates about problems left in last arcboard:
(1) use context manager instead of start/stop @annatisch 
code is [here](https://github.com/Azure/azure-sdk-for-python/blob/15f2e754307f2457db19fb1b1b01e92391f7fee0/sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_client.py#L808-L812)
(2) avoid thread deadlock when users callback client method to send message from @johanste 
solution: [create individual thread](https://github.com/Azure/azure-sdk-for-python/blob/15f2e754307f2457db19fb1b1b01e92391f7fee0/sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_client.py#L360-L364) to call handlers to avoid listener thread block so that we can avoid thread deadlock

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_client.py
```diff
@@ -0,0 +1,835 @@
...
+        credential: Union[WebPubSubClientCredential, str],
+        **kwargs: Any,
+    ) -> None:
+        if isinstance(credential, WebPubSubClientCredential):
+            self._credential = credential
+        elif isinstance(credential, str):
+            self._credential = WebPubSubClientCredential(credential)
+        else:
+            raise TypeError("type of credential must be str or WebPubSubClientCredential")
+        reconnect_retry_total = kwargs.pop("reconnect_retry_total", _RETRY_TOTAL)
```
iscai-msft, line 126: are these defaults synced to the defaults in azure core? we should ideally rely on the defaults in azure core instead of duplicating that in case anything changes

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_client.py
```diff
@@ -0,0 +1,835 @@
...
+            WebPubSubProtocolType.JSON_RELIABLE: WebPubSubJsonReliableProtocol,
+        }
+        if protocol_type in protocol_map:
+            self._protocol = protocol_map[protocol_type]()
+        else:
+            self._protocol = WebPubSubJsonReliableProtocol()
+
+        self._reconnect_retry_policy = RetryPolicy(
+            retry_total=reconnect_retry_total,
+            retry_backoff_factor=kwargs.pop("reconnect_retry_backoff_factor", _RETRY_BACKOFF_FACTOR),
```
iscai-msft, line 141: @annatisch just to make sure we're ok with the names for these public kwargs

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_client.py
```diff
@@ -0,0 +1,835 @@
...
+        :type content: Any.
+        :param data_type: The data type. Required.
+        :type data_type: Union[WebPubSubDataType, str].
+        :keyword int ack_id: The optional ackId. If not specified, client will generate one.
+        :keyword bool fire_and_forget: If true, the message won't contains ackId and no AckMessage
+         will be returned from the service.
+        """
+
+        def send_event_attempt():
+            fire_and_forget = kwargs.pop("fire_and_forget", False)
```
iscai-msft, line 291: right name for a kwarg? I know it's not documented but it is technically public

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_client.py
```diff
@@ -0,0 +1,835 @@
+# coding=utf-8
+# --------------------------------------------------------------------------
```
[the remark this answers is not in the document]
johanste: Yeah, we should not edit the generated code...

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_client.py
```diff
@@ -0,0 +1,835 @@
...
+    :keyword str user_agent: The user agent to be used for the request. If specified, this will be added in front of
+     the default user agent string.
+    :keyword bool logging_enable: Whether to output network trace logs to the logger. Default is `False`.
+    :keyword float ack_timeout: Time limit to wait for ack message from server. The default value is 30.0 seconds.
+    :keyword float start_timeout: Time limit to wait for successful client start. The default value is 30.0 seconds.
+    """
+
+    def __init__(
+        self,
+        credential: Union[WebPubSubClientCredential, str],
```
johanste, line 117: We do not allow `str` as a credential since a) it cannot roll/be updated dynamically and b) is generally very ambiguous w. regards to what it actually is. Is it a connection string? A token? An API key? Something else? And new auth methods get added over time, so even if it isn't unclear now, it likely will become an issue in the future.

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_client.py
```diff
@@ -0,0 +1,835 @@
...
+        if isinstance(credential, WebPubSubClientCredential):
+            self._credential = credential
+        elif isinstance(credential, str):
+            self._credential = WebPubSubClientCredential(credential)
+        else:
+            raise TypeError("type of credential must be str or WebPubSubClientCredential")
+        reconnect_retry_total = kwargs.pop("reconnect_retry_total", _RETRY_TOTAL)
+        self._auto_reconnect = reconnect_retry_total > 0
+        self._auto_rejoin_groups = kwargs.pop("auto_rejoin_groups", True)
+        protocol_type = kwargs.pop("protocol_type", WebPubSubProtocolType.JSON_RELIABLE)
```
johanste, line 129: Avoid using `kwargs`. Use keyword-only args instead.

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_client.py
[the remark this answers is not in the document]
msyyc: There is no swagger/typespec for this package and call the code is handwritten.

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_client.py
msyyc: I have discussed with anna about the name and we made an agreement with pattern `"reconnect/recovery_" + retry_variables of azure-core`

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_client.py
msyyc: retry policy in azure-core is for http and this package is for websocket so we can't use default retry policy directly. But I use similar name to try to keep consistency with [azure-core](https://github.com/Azure/azure-sdk-for-python/blob/f368f19b52229bbc29d50929b39bcb12e45c4b5c/sdk/core/azure-core/azure/core/pipeline/policies/_retry.py#L57-L72)

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_client.py
msyyc: I declar it in doc string: https://github.com/Azure/azure-sdk-for-python/blob/f392fb5871d95546c16ea6ac399e93d786036582/sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_client.py#L286-L287

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_client.py
msyyc: It is connection string which is declared in [readme.md](https://github.com/Azure/azure-sdk-for-python/tree/f392fb5871d95546c16ea6ac399e93d786036582/sdk/webpubsub/azure-webpubsub-client#2-connect-with-your-web-pubsub-resource). If users just want to open connection in short time, it is convenient to use the string directly since they can get it in Azure Portal. As for the expired problem, I declare it in docstring: 
https://github.com/Azure/azure-sdk-for-python/blob/webpubsub-client/sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_client.py#L67-L69

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_client.py
annatisch: erm I don't recall that naming pattern... I don't think we should have _both_ "reconnect/recovery" and "retry" - these mean the same thing. I think we should use _only_ one of these verbs where it makes sense to do so.

sdk/webpubsub/azure-webpubsub-client/README.md
```diff
@@ -0,0 +1,217 @@
...
+### 3. Join groups
+
+Note that a client can only receive messages from groups that it has joined and you need to add a callback to specify the logic when receiving messages.
+
+```python
+# ...continues the code snippet from above
+
+# Registers a listener for the event 'group-message' early before joining a group to not miss messages
+group_name = "group1";
+client.on("group-message", lambda e: print(f"Received message: {e.data}"));
```
annatisch, line 63: While it's not blocking for first preview - this still looks very JS. We need to do some Python ecosystem research on appropriate patterns here.

sdk/webpubsub/azure-webpubsub-client/README.md
```diff
@@ -0,0 +1,217 @@
...
+### Operation and retry
+
+By default, the operation such as `client.join_group()`, `client.leave_group()`, `client.send_to_group()`, `client.send_event()` has three retries. You can configure through the key-word arguments. If all retries have failed, an error will be thrown. You can keep retrying by passing in the same `ack_id` as previous retries so that the Web PubSub service can deduplicate the operation.
+
+```python
+try:
+  client.join_group(group_name)
+except Exception as e:
+  ack_id = None
+  if isinstance(e, SendMessageError):
```
annatisch, line 141: This should look something like:
```python
try:
    client.join_group(group_name)
except SendMessageError as e:
    client.join_group(group_name, ack_id=e.ack_id)

File: sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_model_base.py [hunk 27]
```diff
@@ -0,0 +1,705 @@
+# coding=utf-8
+# --------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+# --------------------------------------------------------------------------
```
annatisch, line 6: It looks like this has been copied from Core? Any particular reason we're not just using Core?

annatisch: And if there is a particular reason - can we make sure it's documented in comments so that other developers are aware of the context?

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_client.py
```diff
@@ -0,0 +1,864 @@
...
+    JoinGroupMessage,
+    LeaveGroupMessage,
+    AckMessageError,
+)
+from .models._enums import WebPubSubDataType, WebPubSubClientState, CallBackType, WebPubSubProtocolType
+from ._util import delay, format_user_agent
+
+_LOGGER = logging.getLogger(__name__)
+
+if sys.version_info >= (3, 8):
```
annatisch, line 50: You should be able to remove this check - just import from `typing_extensions` and it will do the check for you.

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_util.py
```diff
@@ -0,0 +1,19 @@
...
+# Code generated by Microsoft (R) AutoRest Code Generator.
+# Changes may cause incorrect behavior and will be lost if the code is regenerated.
+# --------------------------------------------------------------------------
+from typing import Optional
+import time
+from azure.core.pipeline.policies import UserAgentPolicy
+from ._version import VERSION
+
+
+def delay(delay_seconds: float):
```
annatisch, line 14: Not sure why we need this?

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/models/_enums.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+class WebPubSubDataType(str, Enum, metaclass=CaseInsensitiveEnumMeta):
+    BINARY = "binary"
+    JSON = "json"
+    TEXT = "text"
+    PROTOBUF = "protobuf"
+
+
+class CallBackType(str, Enum, metaclass=CaseInsensitiveEnumMeta):
```
annatisch, line 36: Can we just use a Literal union rather than have an enum for this? If not `Callback` is one word, so no capital "B".

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_serialization.py
```diff
@@ -0,0 +1,1998 @@
+# --------------------------------------------------------------------------
```
annatisch, line 1: We don't seem to be using this old msrest serialization - we can remove this?

sdk/webpubsub/ci.yml
```diff
@@ -28,8 +28,11 @@ extends:
...
     ServiceDirectory: webpubsub
+    TestProxy: true
     Artifacts:
     - name: azure-mgmt-webpubsub
       safeName: azuremgmtwebpubsub
     - name: azure-messaging-webpubsubservice
-      safeName: azuremessagingwebpubsubservice
\ No newline at end of file
+      safeName: azuremessagingwebpubsubservice
+    - name: azure-webpubsub-client
```
annatisch, line 37: Is there a reason this isn't called `azure-messaging-webpubsubclient`? Or was there a decision to drop the `messaging` namespace going forward?

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_client.py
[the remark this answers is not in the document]
annatisch: So we should update the comment? :)

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_client.py
annatisch: hmm actually I've refreshed myself on the context - disregard my previous comment!
These names are fine for now - Will have a think if there's a better way to represent this.

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_client.py
annatisch: I feel like there might be a cleaner name for this... like just `ack=True` - `fire_and_forget` is very long for a pretty simply concept. I wonder what other libraries are doing here.

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_client.py
```diff
@@ -0,0 +1,864 @@
...
+        event_name: str,
+        content: Any,
+        data_type: Union[WebPubSubDataType, str],
+        **kwargs: Any,
+    ) -> None:
+        """Send custom event to server
+        :param event_name: The event name. Required.
+        :type event_name: str.
+        :param content: The data content. Required.
+        :type content: Any.
```
annatisch, line 306: Are we able to give more guidance (either in the description or the type hint) as to what I should pass in here?

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_client.py
```diff
@@ -0,0 +1,864 @@
...
+            old_thread_seq_ack.join()
+        if old_thread and old_thread.is_alive():
+            _LOGGER.debug("wait for listener thread stop")
+            old_thread.join()
+
+        _LOGGER.info("stop client successfully")
+
+    @overload
+    def on(
+        self, event: Union[Literal[CallBackType.CONNECTED], str], listener: Callable[[OnConnectedArgs], None]
```
annatisch, line 705: I don't think you need to have a `Union` type here - the `Literal` is already `str`, so it's implied - and I think it will confuse the intellisense experience.

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_client.py
```diff
@@ -0,0 +1,864 @@
...
+        self, event: Union[CallBackType, str], listener: Callable, **kwargs: Any  # pylint: disable=unused-argument
+    ) -> None:
+        if event in self._handler:
+            self._handler[event].append(listener)
+        else:
+            _LOGGER.error("wrong event type: %s", event)
+
+    @overload
+    def off(
+        self, event: Union[Literal[CallBackType.CONNECTED], str], listener: Callable[[OnConnectedArgs], None]
```
annatisch, line 782: These `off` methods are going to be hard to use - we should investigate a better pattern here.
For example, unless I have the exact instance of the `Callable` that I passed into the `on` method - this wont work.

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_client.py
msyyc: updated

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_client.py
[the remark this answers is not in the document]
msyyc: Got it and I have updated all the comment.

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_client.py
msyyc: I think `ack` is fine.

sdk/webpubsub/azure-webpubsub-client/README.md
msyyc: Yeah. The only reason to use lambda here is to make the README cleaner. In [sample](https://github.com/Azure/azure-sdk-for-python/blob/webpubsub-client/sdk/webpubsub/azure-webpubsub-client/samples/hello_world.py) code, we use callback function.

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_client.py
msyyc: https://github.com/Azure/azure-sdk-for-python/blob/205c0e59e1cc4b8092c57a5d92fee63fdb056448/sdk/webpubsub/azure-webpubsub-client/setup.py#L68
For higher version, we don't hope install typing-extensions additionally.

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/models/_enums.py
msyyc: callbacktype is very like client state so I make an enum to make it clear. I think it is fine to rename it to `CallbackType`

sdk/webpubsub/ci.yml
msyyc: I remember @johanste agreed with the name: https://github.com/Azure/azure-sdk/issues/5623 although I can't find the specific words now...

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_client.py
msyyc: updated

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_client.py
msyyc: updated

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_client.py
msyyc: If the target listener doesn't exist, we will log info; If it exists, remove it; If event type is not right, we log error but don't raise exception. No matter the listener exists or not, we don't raise exception. It keeps consistency across language. Is there any ambiguity?

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_model_base.py
msyyc: This `model_base`  keeps consistency with [DPG model](https://github.com/Azure/autorest.python/blob/main/packages/typespec-python/test/generated/core/_specs_/azure/core/_model_base.py). It has not been moved into azure-core so we use it as vendor code

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_client.py
```diff
@@ -0,0 +1,850 @@
...
+        **kwargs: Any,
+    ) -> None:
+        """Send message to group.
+        :param group_name: The group name. Required.
+        :type group_name: str.
+        :param content: The data content. Required.
+        :type content: Any.
+        :param data_type: The data type. Required.
+        :type data_type: Any.
+        :keyword bool ack: If true, the message won't contains ackId and no AckMessage
```
zackliu, line 343: Why when `ack` is true, it won't contains `ackId`?

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_client.py
```diff
@@ -0,0 +1,850 @@
...
+            url_dict = dict(urllib.parse.parse_qsl(url_parse.query))
+            url_dict.update(params)
+            new_query = urllib.parse.urlencode(url_dict)
+            url_parse = url_parse._replace(query=new_query)
+            new_url = urllib.parse.urlunparse(url_parse)
+            return str(new_url)
+        return None
+
+    def _is_connected(self) -> bool:
+        """check whether the client is still coneected to server after start"""
```
zackliu, line 453: `coneected` typo

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_client.py
```diff
@@ -0,0 +1,850 @@
...
+            else:
+                _LOGGER.warning("unknown message type: %s", parsed_message.kind)
+
+        def on_close(_: Any, close_status_code: int, close_msg: str):
+            if self._state == WebPubSubClientState.CONNECTED:
+                _LOGGER.info("WebSocket connection closed. Code: %s, Reason: %s", close_status_code, close_msg)
+
+                self._last_close_event = CloseEvent(close_status_code=close_status_code, close_reason=close_msg)
+                # clean ack cache
+                self._ack_map.clear()
```
zackliu, line 566: Do we need to notify before clear?

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_client.py
msyyc: python rename `fire_and_forget` to `ack` and `fire_and_forget=true` shall equal to `ack=false`. I forget to update the description.
Updated.

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_client.py
msyyc: updated

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_client.py
msyyc: good point! Although we have ack_timeout, it is better to notify before clear to unblock the process.
Updated

sdk/webpubsub/azure-webpubsub-client/README.md
annatisch: My comment was not in regards to the use of `lambda`, but rather having a method called `on` used to register handlers. However the use of `lamda` is also a bit problematic because there is no way to remove the handler (i.e. we have nothing to pass into the `off` method.

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/models/_enums.py
annatisch: It's fine for preview. We can re-evaluate later.

sdk/webpubsub/ci.yml
annatisch: If that's what Johan agreed to then that's fine. Not sure when you plan to release the first beta package - he gets back from vacation next week so we can confirm then.

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_client.py
annatisch: The error behaviour sounds fine, but what I meant is that if I have code like in the readme:
```python
client.on("group-message", lambda e: print(f"Received message: {e.data}"))
```
There is no way to turn this handler "off". I have to actually hold onto an instance of the lamda or callable in order to remove it.
It's fine for first preview - but we have to revisit this.

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_client.py
```diff
@@ -0,0 +1,858 @@
...
+        self._cv: threading.Condition = threading.Condition()
+        self._thread_seq_ack: Optional[threading.Thread] = None
+        self._thread: Optional[threading.Thread] = None
+        self._ack_timeout: float = ack_timeout
+        self._start_timeout: float = start_timeout
+        self._user_agent: Optional[str] = user_agent
+        self._logging_enable: bool = logging_enable
+
+    def _next_ack_id(self) -> int:
+        self._ack_id = self._ack_id + 1
```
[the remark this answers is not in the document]
annatisch: This comment also applies to the maps stored on the client (e.g. `self._ack_map` etc) - which seem to get updated in a non-threadsafe way.
Is the expectation that this entire client is not threadsafe?

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_client.py
```diff
@@ -0,0 +1,858 @@
...
+                while (time.time() - recovery_start < _RECOVERY_TIMEOUT) and not self._is_stopping:
+                    try:
+                        self._connect(recovery_url)
+                        _LOGGER.debug("Recovery succeeded")
+                        return
+                    except:  # pylint: disable=bare-except
+                        _LOGGER.debug("Try to recover after %d seconds", _RECOVERY_RETRY_INTERVAL)
+                        time.sleep(_RECOVERY_RETRY_INTERVAL)
+
+                _LOGGER.warning("Recovery attempts failed after 30 seconds or the client is stopping")
```
annatisch, line 608: We have hard-coded recovery timeout to 30 seconds? Isn't that configurable?

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_client.py
```diff
@@ -0,0 +1,858 @@
...
+                self._handle_connection_close_and_no_recovery()
+            else:
+                _LOGGER.debug("WebSocket closed before open")
+                raise Exception(f"Fail to start Websocket: {close_status_code}")
+
+        if self._is_stopping:
+            raise Exception("Can't start a client during stopping")
```
annatisch, line 615: Just a generic exception? Is this something we handle internally or a customers expected to handle it? If so, it's quite difficult to write logic to act on an `Exception`. Maybe `RuntimeError`? Probably best to look at the Python std lib threading/futures libraries to see what errors they raise in an invalid state and align with that.

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_client.py
```diff
@@ -0,0 +1,858 @@
...
+                        return
+                    except:  # pylint: disable=bare-except
+                        _LOGGER.debug("Try to recover after %d seconds", _RECOVERY_RETRY_INTERVAL)
+                        time.sleep(_RECOVERY_RETRY_INTERVAL)
+
+                _LOGGER.warning("Recovery attempts failed after 30 seconds or the client is stopping")
+                self._handle_connection_close_and_no_recovery()
+            else:
+                _LOGGER.debug("WebSocket closed before open")
+                raise Exception(f"Fail to start Websocket: {close_status_code}")
```
annatisch, line 612: Same generic exception here - are customers going to see these errors? Looks like this should be some kind of IOError... Customers should never have to handle errors of type `Exception`

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_client.py
```diff
@@ -0,0 +1,858 @@
...
+            try:
+                self._start_from_restarting()
+                success = True
+                break
+            except Exception as e:  # pylint: disable=broad-except
+                _LOGGER.warning("An attempt to reconnect connection failed %s", e)
+                attempt = attempt + 1
+                delay_seconds = self._reconnect_retry_policy.next_retry_delay(attempt)
+                if not delay_seconds:
+                    break
```
annatisch, line 413: Should this be `break` or `continue`?

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_client.py
```diff
@@ -0,0 +1,858 @@
...
+            _LOGGER.debug("reconnect successfully")
+
+    def _handle_connection_stopped(self):
+        _LOGGER.debug("Connection stopped")
+        self._is_stopping = False
+        self._state = WebPubSubClientState.STOPPED
+        self._call_back(CallbackType.STOPPED)
+
+    def _handle_connection_close_and_no_recovery(self):
+        _LOGGER.debug("Connection closed and no recovery")
```
annatisch, line 428: What's the difference between "Connection stopped" (used above) and "Connection closed" (used here)?

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_client.py
```diff
@@ -0,0 +1,858 @@
...
+
+        def on_message(_: Any, data: str):
+            def handle_ack_message(message: AckMessage):
+                if message.ack_id in self._ack_map:
+                    if not (message.success or (message.error and message.error.name == "Duplicate")):
+                        self._ack_map[message.ack_id].error_detail = message.error
+                    else:
+                        self._ack_map[message.ack_id].error_detail = None
+                    self._ack_map[message.ack_id].ack_id = message.ack_id
+                    with self._ack_map[message.ack_id].cv:
```
annatisch, line 492: Shouldn't this whole section be inside the `with` statement?

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_client.py
```diff
@@ -0,0 +1,858 @@
...
+    from typing_extensions import Literal  # type: ignore  # pylint: disable=ungrouped-imports
+
+
+class WebPubSubClientCredential:
+    @overload
+    def __init__(self, client_access_url_provider: str) -> None:
+        ...
+
+    @overload
+    def __init__(self, client_access_url_provider: Callable[[], str]) -> None:
```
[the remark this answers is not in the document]
msyyc: updated

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_client.py
msyyc: It shall be "break". `delay_seconds` is None which means client shall not retry anymore.

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_client.py
msyyc: When "Connection stopped", client will be stopped.
When "Connection closed", client will try to reconnect or recover (configured when init). If success, it can be connected again; if fails, client will be stopped.

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_client.py
```diff
@@ -0,0 +1,858 @@
...
+        return self._ack_id
+
+    def _send_message(self, message: WebPubSubMessage, **kwargs: Any) -> None:
+        pay_load = self._protocol.write_message(message)
+        if not self._ws or not self._ws.sock:
+            raise Exception("The connection is not connected.")
+
+        self._ws.send(pay_load)
+        if kwargs.pop("logging_enable", False) or self._logging_enable:
+            _LOGGER.debug("\nconnection_id: %s\npay_load: %s", self._connection_id, pay_load)
```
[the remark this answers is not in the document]
msyyc: Yeah, we consider to add it in the future.

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_client.py
msyyc: `30` is rule of service: https://github.com/Azure/azure-webpubsub/blob/main/protocols/client/client-spec.md#14-connection-recovery

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_client.py
msyyc: I added some exception type for users to catch: https://github.com/Azure/azure-sdk-for-python/blob/6cd59e90caab73b69a4ab9aaafdf617ce836ea11/sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/models/_models.py#L566-L583

sdk/webpubsub/azure-webpubsub-client/azure/webpubsub/client/_client.py
msyyc: https://github.com/Azure/azure-sdk-for-python/pull/28156#discussion_r1169780844

sdk/webpubsub/ci.yml
johanste: If the other webpubsub package(s) use the messaging group name, we should do that here as well...

msyyc: Got it. Let us keep consistent with other language to add `messaging` in package name

zackliu (approved):

PR #28163, opened 2023-01-04
[Communication] - phone-numbers - Add tests to update capabilities

sorrego-msft:
Update capabilities was only being tested with an unauthorized number, but not with an empty string/ invalid number.

This will add an empty phone number check within the update capabilities function and tests for empty/invalid numbers with their
corresponding recordings


automagically (approved):

danielortega-msft:
I know this has been merged, but in the future please take into account that this kind of changes need to be reflected in the changelog file.

sorrego-msft:
> I know this has been merged, but in the future please take into account that this kind of changes need to be reflected in the changelog file.

You are completely right, my bad. 
At the beginning I was just planning on adding tests, should I still update the changelog if that was the case?
Would a "Trying to update capabilities with empty phone number will throw ValueError" have been good enough for the changelog?

PR #28175, opened 2023-01-05
[Monitor][Query] Improve typing

pvaneck:
This enables the mypy, pyright, and verifytypes checks in the CI, and also adds some typing improvements in order to pass the checks.


sdk/monitor/azure-monitor-query/azure/monitor/query/_models.py
```diff
@@ -31,11 +41,11 @@ class LogsTable(object):
     """
 
     def __init__(self, **kwargs: Any) -> None:
-        self.name = kwargs.pop("name", None)  # type: str
-        self.columns = kwargs.pop("columns", None)  # type: Optional[str]
-        self.columns_types = kwargs.pop("column_types", None)  # type: Optional[Any]
+        self.name: str = kwargs.pop("name", "")
+        self.columns: Optional[List[str]] = kwargs.pop("columns", None)
+        self.columns_types: Optional[List[Any]] = kwargs.pop("column_types", None)
```
kristapratico, line 46: The docs should be updated to show which instance attrs are optional, too. Better yet, if you update the type hints to the class-level (per my previous comment) and include their descriptions then you can just remove the redundant ivar docstrings, e.g.:

```py
class LogsTable:
    """Contains the columns and rows for one table in a query response.

    """

    name: str
    """Required. The name of the table."""
    columns: Optional[List[str]] = None
    """The labels of columns in this table."""
    column_types: Optional[List[Any]] = None
    """The types of columns in this table."""

    def __init__(self, **kwargs: Any) -> None:
        ...
```

sdk/monitor/azure-monitor-query/azure/monitor/query/_models.py
```diff
@@ -114,16 +124,16 @@ class MetricsQueryResult(object):
...
-        self.namespace = kwargs.get("namespace", None)
-        self.resource_region = kwargs.get("resource_region", None)
-        self.metrics = kwargs["metrics"]
+    def __init__(self, **kwargs: Any) -> None:
+        self.cost: Optional[int] = kwargs.get("cost", None)
+        self.timespan: str = kwargs["timespan"]
+        self.granularity: Optional[timedelta] = kwargs.get("granularity", None)
+        self.namespace: Optional[str] = kwargs.get("namespace", None)
+        self.resource_region: Optional[str] = kwargs.get("resource_region", None)
+        self.metrics: Metric = kwargs["metrics"]
```
kristapratico, line 133: docstring shows list[Metric], which is correct?

sdk/monitor/azure-monitor-query/azure/monitor/query/_models.py
```diff
@@ -143,7 +153,7 @@ def _from_generated(cls, generated):
 class MetricsList(list):
     """Custom list for metrics
     """
-    def __init__(self, **kwargs): # pylint: disable=super-init-not-called
+    def __init__(self, **kwargs: Any): # pylint: disable=super-init-not-called
```
kristapratico, line 156: nit: for type-completeness sake

```suggestion
    def __init__(self, **kwargs: Any) -> None: # pylint: disable=super-init-not-called
```

sdk/monitor/azure-monitor-query/azure/monitor/query/_models.py
```diff
@@ -369,28 +380,22 @@ class MetricDefinition(object):  # pylint: disable=too-many-instance-attributes
...
+        self.name: Optional[str] = kwargs.get("name", None)
+        self.unit: Optional[Union[str, MetricUnit]] = kwargs.get("unit", None)
+        self.primary_aggregation_type: Optional[Union[str, MetricAggregationType]] = \
+            kwargs.get("primary_aggregation_type", None)
+        self.supported_aggregation_types: Optional[List[Union[str, MetricAggregationType]]] = \
+            kwargs.get("supported_aggregation_types", None)
+        self.metric_availabilities: List[MetricAvailability] = kwargs.get("metric_availabilities", None)
+        self.id: Optional[str] = kwargs.get("id", None)
+        self.dimensions: Optional[List[str]] = kwargs.get("dimensions", None)
+        self.metric_class: Optional[Union[str, MetricClass]] = kwargs.get("metric_class", None)
```
kristapratico, line 395: This service seems to have a lot of Optional properties - are they all really not sent in the response? This is going to make type checking the samples / how our users write code with the library pretty difficult with a lot of `None` safety checks before accessing the property... if that's how it is, then fine, more feedback for the service than for your PR.

pvaneck: Yeeea, all the optionals are kinda a byproduct from how everything is passed in through kwargs. I believe a lot of these are sent in the response when retreiving metrics. I can double check and see If i can make some of these required keyword arguments. and non-optional types.

sdk/monitor/azure-monitor-query/azure/monitor/query/_models.py
pvaneck: Thanks for providing an example of this new format. I will update the classes to follow this convention.

sdk/monitor/azure-monitor-query/azure/monitor/query/_exceptions.py
```diff
@@ -25,12 +33,12 @@ class LogsQueryError(object):
     """
 
     def __init__(self, **kwargs: Any) -> None:
-        self.code = kwargs.get("code", None)
-        self.message = kwargs.get("message", None)
-        self.details = kwargs.get("details", None)
+        self.code: Optional[str] = kwargs.get("code", None)
+        self.message: Optional[str] = kwargs.get("message", None)
+        self.details: Optional[List[JSON]] = kwargs.get("details", None)
```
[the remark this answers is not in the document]
pvaneck: Thanks for providing a link to the discussion. I will update to follow suit.

sdk/monitor/azure-monitor-query/azure/monitor/query/_models.py
pvaneck: After investigating this, based on the api spec, nothing is explicitly listed as required. And if generating models, every field type is generated as Optional. From testing, seems like all these fields are sent in the response, but I'm not sure if this is guaranteed. I think okay for now to keep these optional.

I do want to eventually use explcit keyword args instead of just kwargs, but not as important since users aren't ever expected to instantiate the classes that have the `**kwargs` init function.

sdk/monitor/azure-monitor-query/azure/monitor/query/_models.py
```diff
@@ -564,58 +530,104 @@ def _from_generated(cls, generated) -> "MetricAvailability":
...
+    """The namespace the metric belongs to."""
+    name: Optional[str] = None
+    """The name and the display name of the metric, i.e. it is a localizable string."""
+    unit: Optional[Union[str, MetricUnit]] = None
+    """The unit of the metric. Possible values include "Count", "Bytes", "Seconds", "CountPerSecond",
+    "BytesPerSecond", "Percent", "MilliSeconds", "ByteSeconds", "Unspecified", "Cores", "MilliCores",
+    "NanoCores", "BitsPerSecond"."""
+    primary_aggregation_type: Optional[Union[str, MetricAggregationType]] = None
+    """The primary aggregation type value defining how to use the values for display. Possible values
+    include: "None", "Average", "Count", "Minimum", "Maximum", "Total"."""
```
kristapratico, line 551: When I generated the documentation with sphinx, I saw the docs get rendered kind of strange with any sentences including a `:`. Might be worth checking that it looks as expected.

sdk/monitor/azure-monitor-query/azure/monitor/query/_models.py
kristapratico: Ah got it. Thanks for looking into it.

kristapratico (approved):

sdk/monitor/azure-monitor-query/azure/monitor/query/_models.py
pvaneck: Great catch! I noticed the same when I generated the docs as it seems to interpret everything preceding the colon as a type. I thought I removed all the colons, but looks like I missed one. Fixed!

PR #28196, opened 2023-01-05
[ACR] Regenerate package with the latest autorest

YalinLi0312:
Resolves: https://github.com/Azure/azure-sdk-for-python/issues/27730


sdk/containerregistry/azure-containerregistry/azure/containerregistry/_exchange_client.py
```diff
@@ -76,7 +76,7 @@ def get_refresh_token(self, service, **kwargs):
 
     def exchange_aad_token_for_refresh_token(self, service, **kwargs):
         # type: (str, Any) -> str
-        refresh_token = self._client.authentication.exchange_aad_access_token_for_acr_refresh_token(
+        refresh_token = self._client.authentication.exchange_aad_access_token_for_acr_refresh_token( # type: ignore
```
kristapratico, line 79: This ignore looks like a real bug to me -- it doesn't look like these methods generate on the AuthenticationOperations anymore:

https://github.com/Azure/azure-sdk-for-python/blob/733a151cd13549070d64e2e3846de467fca7766b/sdk/containerregistry/azure-containerregistry/azure/containerregistry/_generated/operations/_operations.py#L2521-L2545

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_anonymous_exchange_client.py
```diff
@@ -28,7 +28,7 @@ def __init__(self, endpoint, **kwargs):  # pylint: disable=missing-client-constr
             endpoint = "https://" + endpoint
         self._endpoint = endpoint
         self._client = ContainerRegistry(
-            credential=None,
+            credential="", # type: ignore
```
kristapratico, line 31: Does "" mean something different than None to the service?

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_generated/models/_models.py
```diff
@@ -312,257 +331,126 @@ class ArtifactManifestProperties(msrest.serialization.Model):
...
-        'size': {'readonly': True},
-        'created_on': {'required': True, 'readonly': True},
-        'last_updated_on': {'required': True, 'readonly': True},
-        'architecture': {'readonly': True},
-        'operating_system': {'readonly': True},
-        'related_artifacts': {'readonly': True},
-        'tags': {'readonly': True},
+        "registry_login_server": {"readonly": True},
+        "repository_name": {"readonly": True},
+        "manifest": {"required": True, "readonly": True},
```
kristapratico, line 341: I'm confused, should manifest be readonly? It says required in the docstring. This is why we're seeing mypy error in container_registry_client.py L421 and 654

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_generated/models/_models.py
```diff
@@ -312,257 +331,126 @@ class ArtifactManifestProperties(msrest.serialization.Model):
...
     _validation = {
-        'registry_login_server': {'required': True, 'readonly': True},
-        'repository_name': {'required': True, 'readonly': True},
-        'name': {'required': True, 'readonly': True},
-        'digest': {'required': True, 'readonly': True},
-        'created_on': {'required': True, 'readonly': True},
-        'last_updated_on': {'required': True, 'readonly': True},
+        "registry_login_server": {"required": True, "readonly": True},
+        "repository_name": {"required": True, "readonly": True},
+        "tag": {"required": True, "readonly": True},
```
kristapratico, line 377: Same problem here.

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_exchange_client.py
YalinLi0312: Yes, there are two operations not generated, I pasted the code from previous generated to _patch.py, which is the recommended way to fix the implementation error.

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_anonymous_exchange_client.py
YalinLi0312: "" here is equivalent to None in previous generation.
The parameter "credential" becomes `required` in this generation, and the expected type is `TokenCredential`, I have to pass empty string to avoid the None check when passing down in anonymous access.

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_exchange_client.py
kristapratico: Got it, in this case you might have to run the autorest [postprocessing](https://github.com/Azure/autorest.python/blob/main/docs/customizations.md#postprocessing) script to make mypy happy since it seems like it's seeing AuthenticationOperations as imported from the generated code. cc/ @iscai-msft to confirm.

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_generated/models/_models.py
YalinLi0312: Comparing with the code changes, manifest is `readonly and required`. @iscai-msft Can you help to confirm?

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_registry_client.py
```diff
@@ -414,8 +414,11 @@ def get_manifest_properties(self, repository, tag_or_digest, **kwargs):
         if _is_tag(tag_or_digest):
             tag_or_digest = self._get_digest_from_tag(repository, tag_or_digest)
 
+        manifest_properties = self._client.container_registry.get_manifest_properties(
+            repository, tag_or_digest, **kwargs
+        )
         return ArtifactManifestProperties._from_generated(  # pylint: disable=protected-access
-            self._client.container_registry.get_manifest_properties(repository, tag_or_digest, **kwargs),
+            manifest_properties.manifest, # type: ignore
```
mccoyp, line 421: I'm surprised that this `# type: ignore` is necessary -- it looks like the type of `.manifest` should align with the type hint for `_from_generated`. Do you know if a Python 3-style type hint on `_from_generated` might help?
```python
def _from_generated(cls, generated: "ManifestAttributesBase", **kwargs) -> "ArtifactManifestProperties":
```

sdk/containerregistry/azure-containerregistry/azure/containerregistry/aio/_async_base_client.py
```diff
@@ -27,8 +27,8 @@ class ContainerRegistryBaseClient(object):
 
     :param endpoint: Azure Container Registry endpoint
     :type endpoint: str
-    :param credential: AAD Token for authenticating requests with Azure
-    :type credential: ~azure.identity.DefaultTokenCredential
+    :param credential: AAD Token for authenticating requests with Azure, or None in anonymous access
+    :type credential: ~azure.identity.DefaultTokenCredential or None
```
xiangyan99, line 31: AsyncTokenCredential?

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_base_client.py
```diff
@@ -28,8 +28,8 @@ class ContainerRegistryBaseClient(object):
     """Base class for ContainerRegistryClient
 
     :param str endpoint: Azure Container Registry endpoint
-    :param credential: AAD Token for authenticating requests with Azure
-    :type credential: ~azure.identity.DefaultTokenCredential
+    :param credential: AAD Token for authenticating requests with Azure, or None in anonymous access
+    :type credential: ~azure.identity.DefaultTokenCredential or None
```
xiangyan99, line 32: TokenCredential?

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_generated/models/_models.py
iscai-msft: We're documenting the ivar as required, but not offering it as input. Not sure why mypy would be mad here, it's not accepting it as input. What is the mypy error?

YalinLi0312: `error: Argument 1 to "_from_generated" of "ArtifactManifestProperties" has incompatible type "None"; expected "ManifestAttributesBase"`

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_exchange_client.py
YalinLi0312: Update: ignore the mypy error for now, the script failed to get `azure.containerregistry._patch.py`

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_anonymous_exchange_client.py
```diff
@@ -3,15 +3,22 @@
...
-from ._generated.models._container_registry_enums import TokenGrantType
+from ._generated.models import TokenGrantType
 from ._helpers import _parse_challenge
 from ._user_agent import USER_AGENT
 
 
+class AnonymousAccessCredential(TokenCredential):
+    def get_token(
+        self, *scopes: str, claims: Optional[str] = None, tenant_id: Optional[str] = None, **kwargs
+    ) -> AccessToken:
```
mccoyp, line 18: I'm tempted to say we should type this as returning `None` instead of `AccessToken`, since we won't actually return a token. That would violate the protocol and require a typing suppression though... I think that's fine, and that it would be better to suppress a warning for an edge case than to give an incorrect signature. @xiangyan99, do you have an opinion?

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_anonymous_exchange_client.py
```diff
@@ -48,7 +55,7 @@ def get_acr_access_token(self, challenge, **kwargs):
 
     def exchange_refresh_token_for_access_token(self, refresh_token, service, scope, grant_type, **kwargs):
```
mccoyp, line 56: This method doesn't seem to be used by anything but this class's `get_acr_access_token` method, which itself doesn't seem to be used (which makes sense, given that this is an anonymous client). Do you know if there's any reason to keep these methods?

YalinLi0312: https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/containerregistry/azure-containerregistry/azure/containerregistry/aio/_async_authentication_policy.py#L67

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_anonymous_exchange_client.py
xiangyan99: What's the scenario of using AnonymousAccessCredential and what's the expected behavior?

mccoyp: The anonymous client that uses the credential doesn't actually authenticate any requests, so the credential is intentionally unable to authenticate. The idea of using an actual TokenCredential here instead of using something like an empty string as the credential (which is now a required parameter to client construction) is to raise a more useful error in the off chance that token requests are made

YalinLi0312: Here the one in .Net: https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/containerregistry/Azure.Containers.ContainerRegistry/src/Authentication/ContainerRegistryAnonymousAccessCredential.cs#L14 I'm creating a similar one in Python to fix the mypy error about credential parameter type mismatching.
The return type should be same or a sub class of parent return type - `AccessToken`, using `None` will have override error in mypy.

iscai-msft (commented):
looks good, but have questions about type hints

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_registry_client.py
```diff
@@ -430,8 +433,9 @@ def get_tag_properties(self, repository: str, tag: str, **kwargs) -> ArtifactTag
             for tag in client.list_tag_properties("my_repository"):
                 tag_properties = client.get_tag_properties("my_repository", tag.name)
         """
+        tag_properties = self._client.container_registry.get_tag_properties(repository, tag, **kwargs)
         return ArtifactTagProperties._from_generated(  # pylint: disable=protected-access
-            self._client.container_registry.get_tag_properties(repository, tag, **kwargs),
+            tag_properties.tag, # type: ignore
```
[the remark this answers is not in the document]
YalinLi0312: mypy think `tag` could be `None` though it says required in docstring: https://github.com/Azure/azure-sdk-for-python/blob/9f7a94a46062ff80009ce24ccc2836984c82692d/sdk/containerregistry/azure-containerregistry/azure/containerregistry/_generated/models/_models.py#L391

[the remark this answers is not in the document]
YalinLi0312: There's a related comment above: https://github.com/Azure/azure-sdk-for-python/pull/28196#discussion_r1140720516

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_anonymous_exchange_client.py
mccoyp: Got it -- it shouldn't be used anyway, so aligning with what .NET already has makes sense to me 🙂 Thanks for looking into that!

sdk/containerregistry/azure-containerregistry/azure/containerregistry/_container_registry_client.py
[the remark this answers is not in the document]
iscai-msft: mypy doesn't look at docstrings, but basically it's a required readonly property so generated code sets it to none in the init. i think i'm ok with the tag properties. when the sdk switches over to customizations instead of wrapping i think this issue will go away

iscai-msft (approved):

PR #28214, opened 2023-01-06
Override broken old versions of azure-ai-ml when running storage regression tests

scbedd:
@jalauzon-msft


jalauzon-msft (approved):
Thanks for handling this!

scbedd:
> Thanks for handling this!

Sorry for the delay 👎

PR #28216, opened 2023-01-06
Improve typing

xiangyan99:


sdk/search/azure-search-documents/azure/search/documents/_headers_mixin.py
```diff
@@ -4,20 +4,15 @@
...
-if TYPE_CHECKING:
-    from typing import Optional
 
-
-class HeadersMixin(object):
+class HeadersMixin:
     @property
-    def _headers(self):
-        # type() -> dict
+    def _headers(self) -> Dict:
```
kashifkhan, line 12: if its known that the value of `self._credential.key` and `self._ODATA_ACCEPT` will be `str` the return can be Dict[str, str] otherwise `Dict[str, Any]`

sdk/search/azure-search-documents/azure/search/documents/_headers_mixin.py
```diff
@@ -4,20 +4,15 @@
...
         return {"api-key": self._credential.key, "Accept": self._ODATA_ACCEPT}
 
-    def _merge_client_headers(self, headers):
-        # type(Optional[dict]) -> dict
+    def _merge_client_headers(self, headers: Optional[Dict]) -> Dict:
```
kashifkhan, line 15: This one is interesting, since headers are defined as `Dict[str, str]` in azure.core the method can be `def _merge_client_headers(self, headers: Optional[Dict[str, str]) -> Dict:`.

Does the return also change to Dict[str, str] given that an empty dict can also return ?

sdk/search/azure-search-documents/azure/search/documents/_index_documents_batch.py
```diff
@@ -3,24 +3,19 @@
...
 from ._generated.models import IndexAction
 
-if TYPE_CHECKING:
-    # pylint:disable=unused-import,ungrouped-imports
-    from typing import List
 
-
-def _flatten_args(args):
-    # type (Union[List[dict], List[List[dict]]]) -> List[dict]
+def _flatten_args(args: Union[List[Dict], List[List[Dict]]]) -> List[Dict]:
```
kashifkhan, line 12: Same for the dicts here. Are the keys at least well defined to be strs ?

sdk/search/azure-search-documents/azure/search/documents/_index_documents_batch.py
```diff
@@ -111,23 +103,21 @@ def add_merge_or_upload_actions(
...
+    def actions(self) -> List[IndexAction]:
         """The list of currently index actions to index.
 
         :rtype: List[IndexAction]
         """
         return list(self._actions)
 
-    def dequeue_actions(self, **kwargs):  # pylint: disable=unused-argument
-        # type: () -> List[IndexAction]
+    def dequeue_actions(self, **kwargs) -> List[IndexAction]:  # pylint: disable=unused-argument
```
kashifkhan, line 120: I think for kwargs we have settled on `Any` but if moving to py3 style kwargs those can be typed as well

sdk/search/azure-search-documents/azure/search/documents/_utils.py
```diff
@@ -24,3 +22,31 @@ def get_authentication_policy(credential, **kwargs):
         credential, scope
     )
     return authentication_policy
+
+
+def odata(statement, **kwargs):
```
kashifkhan, line 27:
```suggestion
def odata(statement: str, **kwargs: Any) -> str:
```

sdk/search/azure-search-documents/azure/search/documents/_index_documents_batch.py
xiangyan99: I am not 100% sure. There are documents and it may be possible that some properties are not strings?

kashifkhan: @kristapratico is there any meaningful difference between `Dict` and `Dict[Any, Any]` in this case?

kristapratico: No difference for the type checkers - https://mypy-play.net/?mypy=latest&python=3.11&gist=c5651f063e241f5da69333782331fd0e

sdk/search/azure-search-documents/azure/search/documents/aio/_paging.py
```diff
@@ -44,28 +40,25 @@ def _first_iterator_instance(self):
...
+    async def get_coverage(self) -> float:
         """Return the coverage percentage, if `minimum_coverage` was
         specificied for the query.
 
         """
         return await self._first_iterator_instance().get_coverage()
 
-    async def get_count(self):
-        # type: () -> float
+    async def get_count(self) -> float:
```
kashifkhan, line 54: I think this one is a typo, count returning float would be odd

sdk/search/azure-search-documents/azure/search/documents/aio/_index_documents_batch_async.py
```diff
@@ -3,24 +3,19 @@
...
-if TYPE_CHECKING:
-    # pylint:disable=unused-import,ungrouped-imports
-    from typing import List
 
-
-def _flatten_args(args):
-    # type (Union[List[dict], List[List[dict]]]) -> List[dict]
-    if len(args) == 1 and isinstance(args[0], (list, tuple)):
+def _flatten_args(args: Union[List[Dict], List[List[Dict]]]) -> List[Dict]:
+    if len(args) == 1 and isinstance(args[0], (List, Tuple)):
```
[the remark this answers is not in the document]
xiangyan99: Tested and it still worked. :)

kristapratico (approved):
Can we remove search from the CI skip [list](https://github.com/Azure/azure-sdk-for-python/blob/main/tools/azure-sdk-tools/ci_tools/environment_exclusions.py)?

PR #28237, opened 2023-01-09
[EventHubs] add back api version + update test sr version

swathipil:
as per Laurent: re-record tests to include api-version in recordings, since that's crucial info to test against.

Instead, in dev_reqs.txt, restrict azure-schemaregistry version to <=1.2.0. In release pipeline, recordings are failing on windows due to potential pip bug b/c 1.3.0b1 is being installed. Restricting in the dev requirements should allow us to bypass this issue.

Will need to remove the restriction once we release the next stable SR.


kashifkhan (approved):
lgtm

PR #28241, opened 2023-01-09
App Config Provider Module Level Load

mrm9084:
# Description

Moves the load method of AppConfigurationProvider from a class method to a module method.


sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/_azureappconfigurationprovider.py
```diff
@@ -7,184 +7,168 @@
...
 
-        (main_type, sub_type) = type_parts
-        if main_type != "application":
-            return False
+    return False
 
-        sub_types = sub_type.split("+")
-        if "json" in sub_types:
-            return True
+class AzureAppConfigurationProvider:
```
annatisch, line 146: Given that this object holds an instance of a client, we will need to expose this as a context manger and allow it to be closeable (such that the connection can be cleaned up).
This means that we will probably need to still keep the object public so that it can be properly documented....

sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/_azureappconfigurationprovider.py
```diff
@@ -209,3 +202,25 @@ def values(self):
         type: () -> list
         """
         return self._dict.values()
+
+    def get(self, key, default=None):
+        """
+        Returns the value of the specified key. If the key does not exist, returns the default value.
+
+        type: (str, object) -> object
```
annatisch, line 210: These hints are in neither the type hint format nor the docstring format.
Using the inline syntax in the method signature is preferred.

sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/_azureappconfigurationkeyvaultoptions.py
```diff
@@ -16,11 +17,10 @@ class AzureAppConfigurationKeyVaultOptions:
...
     :type secret_clients: list[~azure.keyvault.secrets.SecretClient]
     :param secret_resolver: A function that takes a URI and returns a value.
-    :type secret_resolver: callable
+    :type secret_resolver: Callable[[str], str]
     """
-
-    def __init__(self, credential=None, secret_clients=None, secret_resolver=None):
-        # type: (TokenCredential, List[SecretClient], Callable) -> None
+    def __init__(self, *, credential=None, secret_clients=None, secret_resolver=None):
+        # type: (TokenCredential, List[SecretClient], Callable[[str], str]) -> None
```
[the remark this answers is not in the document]
mrm9084: If I know what you are talking about, the "vault endpoint" comes from App Configuration inside of the key vault reference.

sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/_azureappconfigurationprovider.py
```diff
@@ -7,184 +7,168 @@
...
-from azure.core.exceptions import ResourceNotFoundError
 from ._settingselector import SettingSelector
-from ._azure_appconfiguration_provider_error import KeyVaultReferenceError
 from ._constants import KEY_VAULT_REFERENCE_CONTENT_TYPE
 
 from ._user_agent import USER_AGENT
 
-
-class AzureAppConfigurationProvider:
+#pylint:disable=protected-access
```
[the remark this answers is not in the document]
mrm9084: I think I have moved this to the correct spot now.

sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/_azureappconfigurationprovider.py
```diff
@@ -7,184 +7,168 @@
...
+def loadProvider(*, connection_string=None, endpoint=None, credential=None, **kwargs):
```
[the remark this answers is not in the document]
mrm9084: I have made some of this modification, but something seems wrong. endpoint and the other method variables are put into kwargs, not args.

Also, it seems like `Mapping` isn't a `typing` type. I tried making a new one but it didn't work.

sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/_azureappconfigurationkeyvaultoptions.py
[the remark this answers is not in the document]
mrm9084: I also ran into an issue with `TokenCredential` it works fine locally, but for some reason it breaks in the build pipeline, I got `ImportError: cannot import name 'TokenCredential'`. I was using `from azure.core.credential import TokenCredential`. https://dev.azure.com/azure-sdk/public/_build/results?buildId=2105499&view=ms.vss-test-web.build-test-results-tab&runId=38663166&resultId=100000&paneView=debug

sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/_azureappconfigurationprovider.py
mrm9084: This should be all corrected.

sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/_azureappconfigurationprovider.py
mrm9084: Well, this would get a bit awkward when it comes to refresh. Typically it's best to reuse the same clients as it is a costly operation to create new ones.

Also, how would this work given we have one to many clients, for App Configuration (will be with geo-replication) and Key Vault.

sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/_azureappconfigurationprovider.py
```diff
@@ -3,209 +3,253 @@
...
+        return self._dict.get(key, default)
+
+    def __eq__(self, other):
+        if not isinstance(other, AzureAppConfigurationProvider):
+            return False
+        if self._dict != other._dict:
+            return False
+        if self._trim_prefixes != other._trim_prefixes:
+            return False
+        if self._client != other._client:
```
albertofori, line 250: Is there custom equality logic for the `AppConfigurationClient`?

mrm9084: No, this is the best we can do. The SDK isn't really open to letting you know it's internals.

sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/_azureappconfigurationkeyvaultoptions.py
```diff
@@ -3,27 +3,51 @@
...
+        :param credential: A credential for authenticating with the key vault. This is optional if secret_clients is
+        provided.
+        :type credential: ~azure.core.credentials.TokenCredential
+        :param secret_clients: A list of SecretClient from azure-keyvault-secrets. This is optional if credential is
+        provided.
+        :type secret_clients: list[~azure.keyvault.secrets.SecretClient]
+        :param secret_resolver: A function that takes a URI and returns a value.
+        :type secret_resolver: Callable[[str], str]
+        """
+        self.credential = kwargs.get("credential", None)
```
avanigupta, line 51: Should we validate to make sure credential and secret_resolver both cannot be provided together?

sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/_azureappconfigurationprovider.py
```diff
@@ -3,209 +3,253 @@
...
 from ._settingselector import SettingSelector
-from ._azure_appconfiguration_provider_error import KeyVaultReferenceError
 from ._constants import KEY_VAULT_REFERENCE_CONTENT_TYPE
 
 from ._user_agent import USER_AGENT
 
-
-class AzureAppConfigurationProvider:
+@overload
+def load_provider(endpoint: str, credential: str, **kwargs):
```
avanigupta, line 17: What does the usage look like after this change?

sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/_azureappconfigurationprovider.py
```diff
@@ -3,209 +3,253 @@
...
+def __buildprovider(connection_string:str, endpoint:str, credential,
+        key_vault_options:AzureAppConfigurationKeyVaultOptions):
+    provider = AzureAppConfigurationProvider()
+    headers = {}
+    correlation_context = "RequestType=Startup"
+
+    if key_vault_options and (
+        key_vault_options.credential or key_vault_options.secret_clients or key_vault_options.secret_resolver
+    ):
+        correlation_context += ",UsesKeyVault"
```
avanigupta, line 118: So we are adding RequestType and UsesKeyVault tracing right now. Is there a way to get host type and dev environment tracing for python apps?

sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/_azureappconfigurationprovider.py
```diff
@@ -3,209 +3,253 @@
...
-                else:
+    for select in selects:
+        configurations = provider._client.list_configuration_settings(
+            key_filter=select.key_filter, label_filter=select.label_filter
+        )
+        for config in configurations:
+
+            trimmed_key = config.key
+            # Trim the key if it starts with one of the prefixes provided
+            for trim in provider._trim_prefixes:
```
avanigupta, line 90: In .NET, we process KeyVault/Json types first, then trim prefixes. Can we do the same here to maintain the same behavior? Changing the order might produce different results, especially when we start supporting feature flags.

sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/_azureappconfigurationprovider.py
mrm9084: You can see the usage in the samples, it looks not different than now really, minus being part of the class.

sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/_azureappconfigurationkeyvaultoptions.py
mrm9084: Added.

sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/_azureappconfigurationprovider.py
mrm9084: It was already there. This was just moved. I can work on adding those, but not in this branch.

sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/_azureappconfigurationprovider.py
mrm9084: If you look down a bit, the trimmed_key is only used in the dictionary, `_resolve_keyvault_reference` uses the `config` which contains the untrimmed key.

avanigupta: Just thinking about how feature flags would look in future - we'd want to maintain the same schema that we use in .net and spring providers, right? So for a feature "Beta", dictionary key would be something like "FeatureManagement:Beta" (or whatever separator makes sense for python).  
Since feature flag flattening changes the key being added to dictionary, we shouldn't use trimmed_key as dictionary key.

mrm9084: I see two options:

```python
elif config.content_type == FeatureFlagConfigurationSetting._feature_flag_content_type: #pylint: disable=protected-access
    feature_management = provider._dict.get("FeatureManagement", {})
    feature_management[trimmed_key] = config.value
```

or

```python
elif config.content_type == FeatureFlagConfigurationSetting._feature_flag_content_type: #pylint: disable=protected-access
    provider._dict["FeatureManagement." + trimmed_key] = config.value
```

avanigupta: Discussed with Matt offline. Since the config dictionary structure is not limited to <string,string> in python, what you have right now should be good.

avanigupta (approved):

mrm9084:
Closing, I think I have gotten all of the comments from here and moved them too https://github.com/Azure/azure-sdk-for-python/pull/28497.

PR #28261, opened 2023-01-10
Include source branch name in API review label for automatic review

praveenkuttappan:
API review automatic revision doesn't indicate source branch from which each revision was created from. This PR is to include source branch name in API review revision.

### DIFF
--- a/eng/scripts/Create-ApiReview.ps1
+++ b/eng/scripts/Create-ApiReview.ps1
@@ -99,16 +99,18 @@ function ProcessPackage($PackageName)
                 return 1
             }
 
+            $revisionLabel = "$($pkgInfo.Version) - Source Branch:$($SourceBranch)"
             Write-Host "Version: $($version)"
             Write-Host "SDK Type: $($pkgInfo.SdkType)"
             Write-Host "Release Status: $($pkgInfo.ReleaseStatus)"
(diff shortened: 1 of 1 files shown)


praveenkuttappan:
Fixes https://github.com/Azure/azure-sdk-tools/issues/4847 for Python

eng/scripts/Create-ApiReview.ps1
```diff
@@ -99,16 +99,18 @@ function ProcessPackage($PackageName)
                 return 1
             }
 
+            $revisionLabel = "$($pkgInfo.Version) - Source Branch:$($SourceBranch)"
```
weshaggard, line 102: Does this require changes on the APIView side? Do we every make an assumption that this is only a version anywhere?

weshaggard:
```suggestion
            $revisionLabel = "$($pkgInfo.Version) - Source Branch:${SourceBranch}"
```

weshaggard (approved):

eng/scripts/Create-ApiReview.ps1
praveenkuttappan: Package version is part of token Code file itself. Label param in REST API to create a review revision is only used to create the label and it's not parsed or treated as package version. This param can have any text as value.

PR #28276, opened 2023-01-11
[Communication] - Phone Numbers - Update default API version

danielortega-msft:
Adds the missing API version `2022-12-01` to the supported API versions list for phone numbers and makes it the default.


whisper6284 (changes requested):
Fix CI build.

Issue #28288, opened 2023-01-11
[TA] consider renames


kristapratico:
- autodetect_default_language
 - AbstractSummaryAction.sentence_count to max_sentence_count to match Ext Summ

PR #28291, opened 2023-01-11
[TA] beta updates

kristapratico:
Related: https://github.com/Azure/azure-sdk-for-python/issues/28288 https://github.com/Azure/azure-sdk-for-python/issues/28274


sdk/textanalytics/azure-ai-textanalytics/CHANGELOG.md
```diff
@@ -2,10 +2,22 @@
 
 ## 5.3.0b2 (Unreleased)
 
+This version of the client library defaults to the service API version `2022-10-01-preview`.
```
mssfang, line 5: We already mentioned the version updates in the previous beta. Should we keep mentioning it until it changes?

kristapratico: I think it's useful information. I feel like we usually bump the API version when we do a release and this time we're not so it's worth calling out.

iscai-msft (approved):

sdk/textanalytics/azure-ai-textanalytics/CHANGELOG.md
```diff
@@ -2,10 +2,22 @@
...
+
 ### Features Added
 
+- Added `begin_extract_summary` client method to perform extractive summarization on documents.
+- Added `begin_abstract_summary` client method to perform abstractive summarization on documents.
+
 ### Breaking Changes
 
+- Removed models `BaseResolution` and `BooleanResolution`.
```
mssfang, line 14: Is "BaseResolution" a base class that all derived resolutions extends from?

kristapratico: Yes, but we don't really expose it and won't ever return it to the user and so architect feedback was to remove it.

mssfang (approved):

PR #23322, opened 2022-03-03
Raise error when specifying 'tenant_id' in ctor of dac

xiangyan99:
# Description

Please add an informative description that covers that changes made by the pull request and link all relevant issues.

If an SDK is being regenerated based on a new swagger spec, a link to the pull request containing these swagger spec changes has been included above.

# All SDK Contribution checklist:
- [ ] **The pull request does not introduce [breaking changes]**
- [ ] **CHANGELOG is updated for new features, bug fixes or other significant changes.**
- [ ] **I have read the [contribution guidelines](https://github.com/Azure/azure-sdk-for-python/blob/main/CONTRIBUTING.md).**

## General Guidelines and Best Practices
- [ ] Title of the pull request is clear and informative.
- [ ] There are a small number of commits, each of which have an informative message. This means that previously merged commits do not appear in the history of the PR. For more information on cleaning up the commits in your PR, [see this page](https://github.com/Azure/azure-powershell/blob/master/documentation/development-docs/cleaning-up-commits.md).

### [Testing Guidelines](https://github.com/Azure/azure-sdk-for-python/blob/main/CONTRIBUTING.md##building-and-testing)
- [ ] Pull request includes test coverage for the included changes.

PR #28299, opened 2023-01-12
[Identity] Fix args not being propagated in DAC constructor

pvaneck:
AzureCliCredential and AzurePowerShellCredential both accept some arguments inside their constructors. This commit makes sure these are passed in from the DefaultAzureCredential constructor.

Currently, this only propagates `additionally_allowed_tenants`. TBD if new keyword arguments for CLI tenant ID and PowerShell tenant ID need to be introduced.

### DIFF
--- a/sdk/identity/azure-identity/CHANGELOG.md
+++ b/sdk/identity/azure-identity/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### Bugs Fixed
 
+- Fixed an issue where certain arguments to `DefaultAzureCredential` were not being passed to the underlying credentials (i.e. `AzureCliCredential` and `AzurePowerShellCredential`). ([#28299](https://github.com/Azure/azure-sdk-for-python/pull/28299))
+
 ### Other Changes
 
 ## 1.13.0b1 (2023-01-10)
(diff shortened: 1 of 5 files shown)


pvaneck:
@xiangyan99 While PowerShellCredential and AzureCliCredential also accept a tenant_id, seems like I'd need to add additional keyword args that correspond to these credential tenant IDs. Is this something I should be doing or should we bring this up with the rest of the team?

This https://github.com/Azure/azure-sdk-for-python/pull/23322 was added a while ago seemingly to prevent users from using the `tenant_id` keyword instead of the dedicated tenant id kwargs (i.e. `interactive_browser_tenant_id`, `shared_cache_tenant_id`, `visual_studio_code_tenant_id`)

xiangyan99:
> @xiangyan99 While PowerShellCredential and AzureCliCredential also accept a tenant_id, seems like I'd need to add additional keyword args that correspond to these credential tenant IDs. Is this something I should be doing or should we bring this up with the rest of the team?
> 
> This #23322 was added a while ago seemingly to prevent users from using the `tenant_id` keyword instead of the dedicated tenant id kwargs (i.e. `interactive_browser_tenant_id`, `shared_cache_tenant_id`, `visual_studio_code_tenant_id`)

Seems like .net uses "TenantId" for cli https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/identity/Azure.Identity/src/DefaultAzureCredentialFactory.cs#L140 and "VSCodeTenantId" for powershell https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/identity/Azure.Identity/src/DefaultAzureCredentialFactory.cs#L187. I feel we'd better sync with them first. :)

PR #28345, opened 2023-01-13
Enable Pyproject Inclusion

scbedd:
A documentation update is coming for all of these changes.

- [x] Add `pyproject.toml` with custom tool section `[tool.azure-sdk-build]` to all `core` packages.
  - This new section allows users to enable/disable checks at the **package level**.
  - I will script out an update for the entire repository that pulls the remaining package definitions from `environment_exclusions.py` into individual `pyproject.toml` where necessary.
  - This is where the `optional` definitions will live.
- [x] Enables `black` on the t2 packages under azure-core with the new ability to ignore checks on packages.
- [x] Disables all packages with `Inactive` classifier from building/checks. If you want to re-enable, set queue time variable `ENABLE_PACKAGE_NAME=true`. Replace `-` with `_`.
- [x] Allows custom override of enable/disable at the individual check level. `PACKAGE_NAME_CHECK=true`.
  - `AZURE_CORE_PYLINT=true`
  - `AZURE_EVENTHUB_PYLINT=false` <- to disable when it's enabled by default
- [x] Marks a couple packages as `inactive` as they have been placed under the `exclude` due to long periods of disable/brokenness.

Well TIL! The management automation accidentally picks up the `pyproject.toml` instead of the `sdk_packaging.toml` file. Will repair that.

PENDING:

- [x] Repair `update_pr`
- [x] Update spellcheck so that `Build Analyze` passes.


tools/azure-sdk-tools/ci_tools/functions.py
```diff
@@ -89,6 +89,9 @@ def filter_for_compatibility(package_set: List[str]) -> List[str]:
         if running_major_version in spec_set:
             collected_packages.append(pkg)
 
+    logging.debug("Target packages after applying compatibility filter: {}".format(collected_packages))
+    logging.debug("Package(s) omitted by business filter: {}".format(generate_difference(package_set, collected_packages)))
```
mccoyp, line 93: I'm assuming these should be consistent -- correct me if I'm wrong
```suggestion
    logging.debug("Package(s) omitted by compatibility filter: {}".format(generate_difference(package_set, collected_packages)))
```

tools/azure-sdk-tools/ci_tools/functions.py
```diff
@@ -130,48 +161,59 @@ def discover_targeted_packages(
...
 
+def apply_inactive_filter(collected_packages: List[str]) -> List[str]:
+    packages = [pkg for pkg in collected_packages if is_package_active(pkg)]
 
-def remove_omitted_packages(collected_directories):
-    packages = [
-        package_dir for package_dir in collected_directories if os.path.basename(package_dir) not in OMITTED_CI_PACKAGES
-    ]
+    logging.debug("Target packages after applying compatibility filter: {}".format(collected_packages))
+    logging.debug("Package(s) omitted by business filter: {}".format(generate_difference(collected_packages, packages)))
```
mccoyp, line 216:
```suggestion
    logging.debug("Target packages after applying inactive filter: {}".format(collected_packages))
    logging.debug("Package(s) omitted by inactive filter: {}".format(generate_difference(collected_packages, packages)))
```

tools/azure-sdk-tools/tests/test_parse_functionality.py
```diff
@@ -5,7 +5,7 @@
 from unittest.mock import patch
 
 package_root = os.path.abspath(os.path.join(os.path.dirname(__file__), ".."))
-
+repo_root = os.path.abspath(os.path.join(os.path.dirname(__file__), "..", "..", ".."))
```
mccoyp, line 8: This variable doesn't seem to be used

tools/azure-sdk-tools/ci_tools/parsing/parse_functions.py
```diff
@@ -72,8 +84,29 @@ def from_path(cls, parse_directory_or_file: str):
             name_space,
             package_data,
             include_package_data,
+            classifiers,
         )
 
+    def get_build_config(self) -> Dict[str, Any]:
+        return get_build_config(self.folder)
+
+def get_build_config(package: str) -> Dict[str, Any]:
```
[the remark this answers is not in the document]
scbedd: Absolutely. Also I love nits like this.

tools/azure-sdk-tools/ci_tools/functions.py
scbedd: These should absolutely be consistent thanks for the catch 👍

mccoyp (approved):

Issue #26137, opened 2022-09-09
Support uploading a gzip file as input


rakshith91:
Currently, a list of logs is supported which is later gzipped.

It's not ideal that if the user has a gzipped file, they'd have to unzip it , upload it as a list and the we re-zip it again.

Issue #25245, opened 2022-07-15
[Ingestion] Support Stream for uploading logs


rakshith91:
https://docs.python.org/3/library/io.html

PR #28373, opened 2023-01-17
[Monitor][Ingestion] Allow IO type in upload

pvaneck:
Typing for logs argument is changed to indicate an IO type can be passed in, thus matching the [typing of the generated upload method](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/monitor/azure-monitor-ingestion/azure/monitor/ingestion/_operations/_operations.py#L168). An IO type is checked for and passed in as-is without chunking. If the stream is gzip-encoded, we set the content-encoding accordingly.

Closes: #25245
Closes: #26137

### DIFF
--- a/sdk/monitor/azure-monitor-ingestion/azure/monitor/ingestion/_operations/_patch.py
+++ b/sdk/monitor/azure-monitor-ingestion/azure/monitor/ingestion/_operations/_patch.py
@@ -55,11 +55,14 @@ class LogsIngestionClientOperationsMixin(GeneratedOps):
         :raises: ~azure.core.exceptions.HttpResponseError
         """
         if isinstance(logs, IOBase):
-            # Check if the stream is gzip-compressed..
+            if not logs.readable():
+                raise ValueError("The 'logs' stream must be readable.")
             content_encoding = None
-            if logs.read(2) == GZIP_MAGIC_NUMBER:
-                content_encoding = "gzip"
-            logs.seek(0)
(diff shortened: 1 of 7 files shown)


sdk/monitor/azure-monitor-ingestion/azure/monitor/ingestion/_operations/_patch.py
```diff
@@ -52,6 +54,16 @@ def upload(  # pylint: disable=arguments-renamed, arguments-differ
         :rtype: None
         :raises: ~azure.core.exceptions.HttpResponseError
         """
+        if isinstance(logs, IOBase):
+            # Check if the stream is gzip-compressed..
+            content_encoding = None
+            if logs.read(2) == GZIP_MAGIC_NUMBER:
+                content_encoding = "gzip"
+            logs.seek(0)
```
xiangyan99, line 62: IOBase is not guaranteed to be seekable, right?

pvaneck: Ah yea, that's right after looking into it. Good catch. I will refactor with some checks.

pvaneck: Changed the logic to check for readable streams, and only do the gzip check if the stream is seekable. In most cases where an IO type is passed in, I believe the stream will be seekable.

xiangyan99 (approved):

PR #28376, opened 2023-01-17
[formrecognizer] Add error handling for samples

catalinaperalta:
Showcase error handling in samples, plus this will help CI pipelines

### DIFF
--- a/sdk/formrecognizer/azure-ai-formrecognizer/samples/v3.2/async_samples/sample_analyze_business_cards_async.py
+++ b/sdk/formrecognizer/azure-ai-formrecognizer/samples/v3.2/async_samples/sample_analyze_business_cards_async.py
@@ -165,8 +165,19 @@ if __name__ == "__main__":
     try:
         asyncio.run(main())
     except HttpResponseError as error:
-        filter_errors = ["Generic error", "Timeout", "Invalid request", "InvalidImage"]
-        if any(example_error.casefold() in error.message.casefold() for example_error in filter_errors):
+        # Examples of how to check an HttpResponseError
+        # Check by error code:
+        if error.error is not None:
+            if error.error.code == "InvalidRequest":
+                print(f"Received an invalid request error: {error.error}")
+                sys.exit(1)
+            if error.error.code == "InvalidImage":
+                print(f"Received an invalid image error: {error.error}")
+                sys.exit(1)
+        # If the inner error is None and then it is possible to check the message to get more information:
+        filter_msg = ["Generic error", "Timeout", "Invalid request", "InvalidImage"]
+        if any(example_error.casefold() in error.message.casefold() for example_error in filter_msg):
             print(f"Uh-oh! Something unexpected happened: {error}")
             sys.exit(1)
-        print(error)
+        # Print the full error content:
+        print(f"Full HttpResponseError: {error}")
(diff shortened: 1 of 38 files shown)


sdk/formrecognizer/azure-ai-formrecognizer/samples/v3.2/async_samples/sample_analyze_business_cards_async.py
```diff
@@ -160,4 +160,13 @@ async def main():
 
 
 if __name__ == "__main__":
-    asyncio.run(main())
+    import sys
```
catalinaperalta, line 163: Thinking about having a comment here that says something like: 
"Example for how to catch specific errors"...or something like that.

sdk/formrecognizer/azure-ai-formrecognizer/samples/v3.2/async_samples/sample_analyze_business_cards_async.py
```diff
@@ -160,4 +160,13 @@ async def main():
...
+    from azure.core.exceptions import HttpResponseError
+    try:
+        asyncio.run(main())
+    except HttpResponseError as error:
+        filter_errors = ["Generic error", "Timeout", "Invalid request", "InvalidImage"]
```
kristapratico, line 168: Are we able to filter by error code? since that's less likely to change than the message

sdk/formrecognizer/azure-ai-formrecognizer/samples/v3.2/async_samples/sample_analyze_business_cards_async.py
```diff
@@ -160,4 +160,13 @@ async def main():
...
+        if any(example_error.casefold() in error.message.casefold() for example_error in filter_errors):
+            print(f"Uh-oh! Something unexpected happened: {error}")
```
kristapratico, line 170: brainstorming here, but wondering if there is more we can do to illustrate error handling here... would it make sense to specifically catch the different types of errors and provide more information on how to proceed with each one? IIRC you created a troubleshooting guide awhile ago -- is there something we can use from that?

catalinaperalta: Just getting back to this. This is a great idea. I looked through the troubleshooting guide, we specifically mention the generic error case, plus some cases that are specific to building custom models. I think the updates I added will be more comprehensive and showcase how to handle different error formats. In some cases we get the the error property populated on HttpResponseError and for other errors, such as the generic response error we do not. So showcasing both cases, plus the catch all.

catalinaperalta: I could go more in depth about some of the error types I'm checking like invalid request and invalid image, although there may be many reasons why someone could get those.

sdk/formrecognizer/azure-ai-formrecognizer/samples/v3.2/async_samples/sample_analyze_business_cards_async.py
catalinaperalta: Added some examples using error code, but that is not always returned in the format that we handle it in HttpResponseError

kristapratico (approved):

PR #28382, opened 2023-01-17
[Monitor][Query] Extract enums to separate file

pvaneck:
Enums are moved from the models file to a separate file to better organize and avoid a circular dependency when trying to import LogsQueryError from _exceptions.py for typing purposes.


kristapratico (approved):

pvaneck:
Thanks for the review @kristapratico! :)

Issue #23906, opened 2022-04-08
Azure-search-documents(PRE-RELEASE 11.3.0b8) semantic query does not return @search.rerankerScore in result


dinoelT:
Returned results when using the python search client do not contain the @search.rerankerScore field.
```
    results_generator =  search_client.search(search_text= user_query.lower(),
                                            query_type="semantic",
                                            query_language="de-DE",
                                            semantic_configuration_name = "simple-config",
                                            top = 10)

```
However, when making the request using the API/Portal it returns the @search.rerankerScore.
I'm not sure if this is an issue, or I just overlooked something.

l0lawrence:
Thanks for the feedback, we'll investigate this asap.

kashifkhan:
Hi @dinoelT, thanks for the feedback. We'll take a look at this and get back to you asap.

kashifkhan:
@dinoelT I am able to reproduce the bug on my side. I will let you know when I can get an update out

JustinCharbonneau:
Hi, I would like to mention that I am using the Python SDK version 11.3.0b8 and I'm having this issue also. The reranking seems to be applied on the ordering of the returned results, but I do not have access to the @search.rerankerScore. (Which is available in the Search Explorer). 

Any updates on this issue? Thanks!

kashifkhan:
@JustinCharbonneau I'll be meeting with the Search team and will have an update on when we will do our next beta release for this.

LucasVascovici:
#https://github.com/Azure/azure-sdk-for-python/pull/27502

LucasVascovici:
Created a PR #27502

PR #28398, opened 2023-01-18
Added the semantic reranker score and captions

xiangyan99:
#23906

### DIFF
--- a/sdk/search/azure-search-documents/tests/test_search_client_search_live.py
+++ b/sdk/search/azure-search-documents/tests/test_search_client_search_live.py
@@ -58,7 +58,9 @@ class TestSearchClient(AzureRecordedTestCase):
             "hotelName",
             "description",
             "@search.score",
+            "@search.reranker_score",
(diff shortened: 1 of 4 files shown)


sdk/search/azure-search-documents/tests/test_search_client_search_live.py
```diff
@@ -79,7 +81,9 @@ def _test_get_search_filter_array(self, client):
             "hotelName",
             "description",
             "@search.score",
+            "@search.reranker_score",
```
kashifkhan, line 84: There were no new recordings needed for this right ? The response had them already, we were just not exposing them

kashifkhan (approved):

sdk/search/azure-search-documents/tests/test_search_client_search_live.py
xiangyan99: Yes. You are right. These properties are already in the response. Just we did not expose them.

PR #28410, opened 2023-01-19
fixes https://github.com/Azure/azure-sdk-for-python/issues/28307

### LINKED ISSUE #28307: Upload async file data
Hi there! I'm trying to upload a block blob using an async source:
```python
async with aiofiles.open(file=path, mode="rb") as data:
    await blob_client.upload_blob(data, metadata=tags, overwrite=True)
```
This doens't work, the data is supposed to be bytes.

I'd like to be able to use an async stream as source for the async Blob Client

**Describe alternatives you've considered**
My current workaround is to load the whole file into RAM:

```python
async with aiofiles.open(file=path, mode="rb") as data:
    dt = await data.read()
    await blob_client.upload_blob(dt, metadata=tags, overwrite=True)

```
As long as my files don't get too large, this will work

**Additional context**

aersam:
trivially fixes #28307

### DIFF
--- a/sdk/storage/azure-storage-blob/azure/storage/blob/aio/_upload_helpers.py
+++ b/sdk/storage/azure-storage-blob/azure/storage/blob/aio/_upload_helpers.py
@@ -6,6 +6,7 @@
 # pylint: disable=no-self-use
 
 import asyncio
+import inspect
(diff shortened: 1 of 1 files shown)


jalauzon-msft:
Closing this in favor of #28472 but thanks again for your contribution!

Issue #26750, opened 2022-10-11
[Cosmos DB] Cross-region retry alignment across SDKs


ealsur:
The Retry Utility (https://github.com/Azure/azure-sdk-for-python/blob/1b778a9da634b33bdbd0a06f50bd5ba2a6ab20af/sdk/cosmos/azure-cosmos/azure/cosmos/aio/_retry_utility_async.py) should align with the other language SDKs in terms of failover mechanics.

* HTTP 503 (Service Unavailable) - This is documented on https://docs.microsoft.com/en-us/azure/cosmos-db/sql/troubleshoot-sdk-availability#transient-connectivity-issues-on-tcp-protocol, any 503 should be retried on the next available region. Reference https://github.com/Azure/azure-cosmos-dotnet-v3/blob/master/Microsoft.Azure.Cosmos/src/ClientRetryPolicy.cs#L250-L263
* HTTP timeouts - Reference: https://github.com/Azure/azure-cosmos-dotnet-v3/issues/3467, as a repair item from a recent outage, HTTP timeouts to a certain subset of HTTP operations should be locally retried and then attempted on the next available region (not marking the region unavailable, similar to HTTP 503 mechanic)

PR #28420, opened 2023-01-19
[Cosmos] Adds cross regional retries for Service Unavailable/Request Timeouts 

NaluTripician:
# Description

Fix for [#26750](https://github.com/Azure/azure-sdk-for-python/issues/26750).  Currently request timeouts (Status code 408) or service unavailable (Status code 503) do not failover and retry on other regions for multi-master accounts. 

The Python SDK needs to add retries for the following scenarios: 

| Type | Read/Write | 
|----------|------------|
| Data plane |  Read |
| Metadata |  Read | |
|  Query Plan | Read (it is always read) | 

### DIFF
--- a/sdk/cosmos/azure-cosmos/azure/cosmos/_timeout_failover_retry_policy.py
+++ b/sdk/cosmos/azure-cosmos/azure/cosmos/_timeout_failover_retry_policy.py
@@ -49,7 +49,6 @@ class TimeoutFailoverRetryPolicy(object):
     def needsRetry(self):
         if self.args:
             if (self.args[3].method == "GET") or (http_constants.HttpHeaders.IsQueryPlanRequest in self.args[3].headers):
-                print("NEEDS RETRY")
                 return True
         return False
 
--- a/sdk/cosmos/azure-cosmos/azure/cosmos/_timeout_failover_retry_policy.py
+++ b/sdk/cosmos/azure-cosmos/azure/cosmos/_timeout_failover_retry_policy.py
@@ -28,6 +28,9 @@ from ..cosmos.documents import _OperationType
 
 class _TimeoutFailoverRetryPolicy(object):
 
+    Max_retry_attempt_count = 1
+    Retry_after_in_milliseconds = 0
+
     def __init__(self, connection_policy, global_endpoint_manager, *args):
         self._max_retry_attempt_count = 120
         self.current_retry_attempt_count = 0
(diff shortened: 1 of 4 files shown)


sdk/cosmos/azure-cosmos/azure/cosmos/_timeout_failover_retry_policy.py
```diff
@@ -0,0 +1,102 @@
...
+        self.connection_policy = connection_policy
+        self.request = args[0] if args else None
+        # clear previous location-based routing directive
+        if self.request:
+            self.request.clear_route_to_location()
+
+            # Resolve the endpoint for the request and pin the resolution to the resolved endpoint
+            # This enables marking the endpoint unavailability on endpoint failover/unreachability
+            self.location_endpoint = self.global_endpoint_manager.resolve_service_endpoint(self.request)
+            self.request.route_to_location(self.location_endpoint)
```
simorenoh, line 51: not sure if we should be doing this in the init since we create this policy before running the request and would effectively be setting these flags for all requests passing through the retry utility.

sdk/cosmos/azure-cosmos/azure/cosmos/_timeout_failover_retry_policy.py
```diff
@@ -0,0 +1,102 @@
...
+
+class _TimeoutFailoverRetryPolicy(object):
+
+    Max_retry_attempt_count = 1
+    Retry_after_in_milliseconds = 0
+
+    def __init__(self, connection_policy, global_endpoint_manager, *args):
+        self._max_retry_attempt_count = 120
+        self.current_retry_attempt_count = 0
+        self.retry_after_in_milliseconds = 1000
```
simorenoh, line 37: what is the logic in having these values be different here?

sdk/cosmos/azure-cosmos/azure/cosmos/_timeout_failover_retry_policy.py
NaluTripician: It looks like we do the same thing in other retry policies. Since the same code is being repeated in `ShouldRetry` I will remove that from the `init` method.

sdk/cosmos/azure-cosmos/azure/cosmos/_timeout_failover_retry_policy.py
NaluTripician: will clean this up

sdk/cosmos/azure-cosmos/azure/cosmos/_timeout_failover_retry_policy.py
NaluTripician: It looks like this part is necessary.  If we do not include this, then we cannot mark the endpoints as unavailable for read later once a request fails.

simorenoh (approved):
LGTM, thanks for the great work Nalu!

PR #28425, opened 2023-01-20
[Monitor][Ingestion] Encapsulate error callback args

pvaneck:
This creates a new public class called `UploadLogsError` that will be used to contain the exception encountered and the list of failed logs. This follows the same pattern as error callbacks in other language ingestion SDKs, and will allow for some flexibility, allowing us to add more attributes without breaking user callbacks.


sdk/monitor/azure-monitor-ingestion/azure/monitor/ingestion/_operations/_patch.py
```diff
@@ -31,7 +32,7 @@ def upload(  # pylint: disable=arguments-renamed, arguments-differ
         stream_name: str,
         logs: Union[List[JSON], IO],
         *,
-        on_error: Optional[Callable[[Exception, List[JSON]], None]] = None,
+        on_error: Optional[Callable[[UploadLogsError], None]] = None,
```
xiangyan99, line 35:
```suggestion
        on_error: Optional[Callable[UploadLogsError]] = None,
```
?

xiangyan99 (approved):

pvaneck:
Thanks for the review, @xiangyan99!

PR #28438, opened 2023-01-21
Fix verifytypes opt-out check

pvaneck:
Was wondering why my verifytypes runs were opting-out.


pvaneck:
@scbedd PTAL :)

scbedd (approved):
Thank you @pvaneck

PR #28439, opened 2023-01-21
[Monitor][Ingestion] Enable typing checks

pvaneck:
This enables mypy, pyright, and verifytypes checks.


tools/azure-sdk-tools/ci_tools/environment_exclusions.py
```diff
@@ -97,7 +97,6 @@
     "azure-mixedreality-authentication",
     "azure-ai-ml",
     "azure-iot-modelsrepository",
-    "azure-monitor-ingestion",
```
kristapratico, line 100: This is not blocking the PR, but I know @scbedd was working on translating these opt-out lists to pyproject.toml, but I'm not sure the status. Scott - should we continue to use this exclusion list or add pyproject.toml to our libraries?

kristapratico (approved):

PR #28462, opened 2023-01-24
[Perf] Refactor tests.yml into one file per package per language

mikeharder:
- Depends on Azure/azure-sdk-tools#5083


mccoyp (commented):
Other data-plane KV packages each have performance tests as well -- is it intentional to only have a `perf-tests.yml` file for `azure-keyvault-secrets`?

mikeharder:
> Other data-plane KV packages each have performance tests as well -- is it intentional to only have a `perf-tests.yml` file for `azure-keyvault-secrets`?

Yes, it's intentional.  We don't have perf pipelines running for KeyVault yet.  Making them work will require using test-proxy instead of the live service, and it may be a fairly large work item to add test-proxy to our perf pipelines.

However, we did have some tests for keyvault-secrets in the large `tests.yml` that is being deleted, so I wanted to migrate them to `perf-tests.yml` to avoid losing the data.

mccoyp (approved):
Key Vault changes LGTM

pvaneck (approved):
LGTM!

sdk/eventhub/perf-tests.yml
```diff
@@ -0,0 +1,28 @@
+Service: eventhubs
+
+Project: sdk/eventhub/azure-eventhub
+
+PrimaryPackage: azure-eventhub
+
+PackageVersions:
+- azure-core: 1.26.1
+  azure-eventhub: 5.11.1
```
kashifkhan, line 9: will these values have to be updated every time we release a new version or will it continue to work how it does today by picking the latest version in pypi & main using the regex?

mikeharder: Currently, these need to be updated manually, but it's debatable whether this is a feature or a bug.  We update the baselines versions after we verify there is no regression in main.

I have an issue open to add a special version "latest-ga" which would automatically use the latest GA version from package manager, however I am hesitant to do this because it might hide regressions if the baseline is updated after a regression we missed.

https://github.com/Azure/azure-sdk-tools/issues/3553

kashifkhan (approved):
Lgtm 🙂

PR #28471, opened 2023-01-24
[Monitor][Ingestion] Regen using stable API spec

pvaneck:
The package was regenerated using the newly created [stable API spec](https://github.com/Azure/azure-rest-api-specs/pull/22287) for the ingestion API and a small update in autorest.

One minor additional change was to change UploadLogsError `failed_logs` typing to be a list of `Mapping` instead of `MutableMapping` since we don't expect changes to be made to the objects.


pvaneck:
Pending deployment on service side to accept new API version.

kristapratico (approved):
🚢

Issue #28477, opened 2023-01-25
AioHttpStreamDownloadGenerator doesn't work with deflate, gives bad error


AsafMah:
- **Package Name**: azure-core
- **Package Version**:  "1.26.1"
- **Operating System**: Windows
- **Python Version**:  3.7

**Describe the bug**
AioHttpStreamDownloadGenerator only works with gzip encoding, produces an unidicative error when given deflate.
The error I get:

`Error -3 while decompressing data: incorrect header check`

Which doesn't say much.

**To Reproduce**
1. `client = AsyncPipelineClient()`
2. `request = client.post("/some/path", headers= "Accept-Encoding": "deflate")`
3. `res = client.send_request(request, stream=True)`
4. 
```python
async for chunk in res.stream_download(client._pipeline):
		pass
```

**Expected behavior**
For it to work, or at least indicate not to use deflate.

**Screenshots**

**Additional context**
The offending code:
```python
    if enc in ("gzip", "deflate"):
        import zlib
        zlib_mode = 16 + zlib.MAX_WBITS if enc == "gzip" else zlib.MAX_WBITS
        decompressor = zlib.decompressobj(wbits=zlib_mode)
        response._content = decompressor.decompress(response._content)
        response._decompressed_content = True
        return response._content
```

PR #28483, opened 2023-01-25
fix deflate decompression for aiohttp

xiangyan99:


xiangyan99:
#28477

sdk/core/azure-core/azure/core/pipeline/transport/_aiohttp.py
```diff
@@ -332,7 +332,7 @@ async def __anext__(self):
                 if not self._decompressor:
                     import zlib
 
-                    zlib_mode = 16 + zlib.MAX_WBITS if enc == "gzip" else zlib.MAX_WBITS
+                    zlib_mode = 16 + zlib.MAX_WBITS if enc == "gzip" else -zlib.MAX_WBITS
```
iscai-msft, line 335: is the `else` 16 - zlib.MAX_WBITS` or just `-zlib.MAX_WBITS`. I think it's the latter but it's not the clearest

iscai-msft (approved):

sdk/core/azure-core/azure/core/pipeline/transport/_aiohttp.py
xiangyan99: It is just -zlib.MAX_WBITS

sdk/core/azure-core/CHANGELOG.md
```diff
@@ -1,14 +1,8 @@
...
+## 1.26.3 (2023-02-02)
 
-### Features Added
-
-### Breaking Changes
-
-### Bugs Fixed
-
-### Other Changes
+- Fixed deflate decompression for aiohttp   #28483
```
kashifkhan, line 5: nit: will the heading for this be bug fix ?

kashifkhan (approved):

sdk/core/azure-core/CHANGELOG.md
xiangyan99: Good catch. It was an accident. :)

PR #28492, opened 2023-01-26
[nspkg] Remove CI yaml config

pvaneck:
This pipeline has been failing for quite a long time, and appears to no longer provide any value especially since we have moved on from explicit nspkgs.


benbp (approved):
I can delete the pipelines too after merging if desired.

scbedd (approved):
I can't come up with any reasons that we should set them to inactive versus building them. We won't ever need to release a new version because we definitely won't be doing more than one-off py27 support. 

Thanks @pvaneck

PR #28497, opened 2023-01-26
App Configuration Provider Async Support

mrm9084:
# Description

Adds Async support to the Azure App Configuration Provider.

- [ X There are a small number of commits, each of which have an informative message. This means that previously merged commits do not appear in the history of the PR. For more information on cleaning up the commits in your PR, [see this page](https://github.com/Azure/azure-powershell/blob/master/documentation/development-docs/cleaning-up-commits.md).


sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/_azureappconfigurationkeyvaultoptions.py
```diff
@@ -3,27 +3,51 @@
 # Licensed under the MIT License. See License.txt in the project root for
```
xiangyan99, line 3: Should we rename the file?

sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/_settingselector.py
```diff
@@ -15,7 +15,6 @@ class SettingSelector:
     :type label_filter: str
```
xiangyan99, line 15: Same, should we rename the file?

sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/_settingselector.py
```diff
@@ -15,7 +15,6 @@ class SettingSelector:
...
     """
 
-    def __init__(self, key_filter, label_filter="\0"):
-        # type: (str, str) -> None
+    def __init__(self, key_filter: str, label_filter:str="\0"):
```
xiangyan99, line 18: @annatisch Is it good to list "\0" in the signature?

sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/_azureappconfigurationkeyvaultoptions.py
mrm9084: Do you have a suggestion?

sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/_settingselector.py
mrm9084: Do you have a suggestion?

sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/_azureappconfigurationkeyvaultoptions.py
xiangyan99: maybe _azure_appconfiguration_keyvault_options.py?

@annatisch how do you think?

sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/aio/_azureappconfigurationproviderasync.py
```diff
@@ -0,0 +1,284 @@
...
+            connection_string, user_agent=useragent, headers=headers
+        )
+        return provider
+    #pylint:disable=protected-access
+    provider._client = AzureAppConfigurationClient(endpoint, credential, user_agent=useragent, headers=headers)
+    return provider
+
+async def __resolve_keyvault_reference(config, key_vault_options:AzureAppConfigurationKeyVaultOptions, provider) -> str:
+    if key_vault_options is None:
+        raise AttributeError("Key Vault options must be set to resolve Key Vault references.")
```
annatisch, line 147:
```suggestion
        raise ValueError("Key Vault options must be set to resolve Key Vault references.")
```

sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/aio/_azureappconfigurationproviderasync.py
```diff
@@ -0,0 +1,284 @@
...
+
+    if config.secret_id is None:
+        raise AttributeError("Key Vault reference must have a uri value.")
```
annatisch, line 150:
```suggestion
        raise ValueError("Key Vault reference must have a uri value.")
```

sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/aio/_azureappconfigurationproviderasync.py
```diff
@@ -0,0 +1,284 @@
...
+
+    if referenced_client:
+        return (
+            await referenced_client.get_secret(key_vault_identifier.name, version=key_vault_identifier.version)
+        ).value
+
+    if key_vault_options.secret_resolver is not None:
+        return key_vault_options.secret_resolver(config.secret_id)
+
+    raise AttributeError(
```
annatisch, line 171:
```suggestion
    raise ValueError(
```

sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/aio/_azureappconfigurationproviderasync.py
```diff
@@ -0,0 +1,284 @@
...
+        "No Secret Client found for Key Vault reference %s" % (key_vault_identifier.vault_url)
+    )
+
+def __is_json_content_type(content_type: str) -> bool:
```
annatisch, line 175: Is there any reason this couldn't simply be reused from the sync implementation?

sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/aio/_azureappconfigurationproviderasync.py
```diff
@@ -0,0 +1,284 @@
...
+    if key_vault_options and (
+        key_vault_options.credential or key_vault_options.secret_clients or key_vault_options.secret_resolver
+    ):
+        correlation_context += ",UsesKeyVault"
+
+    headers["Correlation-Context"] = correlation_context
+    useragent = USER_AGENT
+
+    if connection_string and endpoint:
+        raise AttributeError("Both connection_string and endpoint are set. Only one of these should be set.")
```
annatisch, line 133:
```suggestion
        raise ValueError("Both connection_string and endpoint are set. Only one of these should be set.")
```

sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/_azureappconfigurationkeyvaultoptions.py
mrm9084: Moved to `_models.py`

sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/_settingselector.py
mrm9084: `\0` is the null option, i.e. I want all configurations with a label. Where as None would be all configuration and all labels. So if a configuration existed with multiple labels it would get a result for each.

sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/_azureappconfigurationkeyvaultoptions.py
```diff
@@ -3,27 +3,53 @@
...
-     provided.
-    :type credential: ~azure.core.credentials.TokenCredential
-    :param secret_clients: A list of SecretClient from azure-keyvault-secrets. This is optional if credential is
-     provided.
-    :type secret_clients: list[~azure.keyvault.secrets.SecretClient]
-    :param secret_resolver: A function that takes a URI and returns a value.
-    :type secret_resolver: callable
-    """
+    @overload
+    def __init__(self,  credential, *, secret_clients: Optional[List[SecretClient]] = None):
```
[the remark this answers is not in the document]
mrm9084: So, the logic is this, if a secret client is provided for a store, we check it's vault url compared to the one in app config, than the value provided is used, otherwise it fallback on the other method, either a general credential or a `secret_resolver` as they are both default options you can't use both.

sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/_azureappconfigurationkeyvaultoptions.py
```diff
@@ -3,27 +3,53 @@
...
+        :type secret_resolver: Callable[[str], str]
+        """
+        ...
 
-    def __init__(self, credential=None, secret_clients=None, secret_resolver=None):
-        # type: (TokenCredential, List[SecretClient], Callable) -> None
-        self.credential = credential
-        self.secret_clients = secret_clients
-        self.secret_resolver = secret_resolver
+    def __init__(self, **kwargs):
```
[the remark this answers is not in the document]
mrm9084: I might be doing something wrong, but args never would have anything in it.

sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/aio/_azureappconfigurationproviderasync.py
```diff
@@ -0,0 +1,284 @@
...
+from azure.keyvault.secrets.aio import SecretClient
+from azure.keyvault.secrets import KeyVaultSecretIdentifier
+from .._azureappconfigurationkeyvaultoptions import AzureAppConfigurationKeyVaultOptions
+from .._settingselector import SettingSelector
+from .._constants import FEATURE_MANAGEMENT_KEY
+
+from .._user_agent import USER_AGENT
+
+@overload
+async def load_provider(endpoint: str, credential: str, **kwargs):
```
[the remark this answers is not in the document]
mrm9084: Updated, had to include the import to as it wasn't used before.

sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/aio/_azureappconfigurationproviderasync.py
```diff
@@ -0,0 +1,284 @@
...
+    """
+    Loads configuration settings from Azure App Configuration into a Python application.
+
+    :keyword endpoint: Endpoint (one of connection_string or endpoint and credential must be set)
+    :type endpoint: str
+    :keyword credential: Credential (one of connection_string or endpoint and credential must be set)
+    :type credential: Union[AppConfigConnectionStringCredential, TokenCredential]
```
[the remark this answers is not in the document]
mrm9084: Fixed

sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/aio/_azureappconfigurationproviderasync.py
```diff
@@ -0,0 +1,284 @@
...
+    :keyword selects: List of setting selectors to filter configuration settings
+    :type selects: list[~azure.appconfigurationprovider.SettingSelector]
+    :keyword trim_prefixes: List of prefixes to trim from configuration keys
+    :type trim_prefixes: list[str]
+    :keyword key_vault_options: Options for resolving Key Vault references
+    :type key_vault_options: ~azure.appconfigurationprovider.KeyVaultOptions
+    """
+    ...
+
+async def load_provider(**kwargs):
```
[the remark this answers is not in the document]
mrm9084: I must be doing something wrong here as args is alway empty when I try this and everything just ends up in kwargs.

sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/_settingselector.py
annatisch: I think that might be clearer if we replace this value in the method signature with some kind of descriptive constant name...

sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/_azureappconfigurationkeyvaultoptions.py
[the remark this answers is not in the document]
annatisch: Whether `args` or `kwargs` are populated will depend on how the method is called.
If I call `KeyVaultOptions("foo", "bar")`, then `args=["foo", "bar"]` and `kwargs={}`.
Whereas if I call `KeyVaultOptions(a="foo", b="bar")` then `args=[]` and `kwargs={'a':'foo', 'b':'bar'}`

sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/_models.py
```diff
@@ -0,0 +1,69 @@
+# ------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+# -------------------------------------------------------------------------
+from typing import overload, List, Optional, Callable
+from azure.keyvault.secrets import SecretClient
+
+class AzureAppConfigurationKeyVaultOptions:
```
annatisch, line 9: In the `load_provider` docstring we just refer to this object as `KeyVaultOptions` (which I generally prefer).
Is there a particular reason for the more verbose name here?

mrm9084: I guess the logic of the name is; 1 this is owned by App Config, not key vault. 2 we use this name in our other providers so it is nice that they match.

annatisch: We can leave it for now and wait till the archboard review.

sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/_settingselector.py
annatisch: We can revisit this during the archboard review.

sdk/appconfiguration/azure-appconfiguration-provider/tests/recordings/test_async_provider.pyTestAppConfigurationClientAADtest_add_configuration_setting.json
```diff
@@ -0,0 +1,4 @@
+{
```
YalinLi0312, line 1: I don't find the test in your latest change, you can remove the odd recording or re-record. It should have some content with the new appconfiguration release.

sdk/appconfiguration/azure-appconfiguration-provider/tests/recordings/test_async_provider.pyTestAppConfigurationClientAADtest_provider_creation.json
```diff
@@ -0,0 +1,4 @@
+{
```
YalinLi0312, line 1: Same here

sdk/appconfiguration/azure-appconfiguration-provider/tests/recordings/test_async_provider.pyTestAppConfigurationClientAADtest_add_configuration_setting.json
mrm9084: They must have gotten in while trying to test out that async issue. I have removed them.

sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/_azureappconfigurationprovider.py
```diff
@@ -3,209 +3,334 @@
...
 
+if TYPE_CHECKING:
+    from azure.core.credentials import TokenCredential
+
 
-class AzureAppConfigurationProvider:
+@overload
+def load_provider(
+        endpoint: str,
+        credential: "TokenCredential",
```
xiangyan99, line 34: (We can defer this question until arch board review). Maybe we should make all these arguments keyword only?

sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/_azureappconfigurationprovider.py
```diff
@@ -3,209 +3,334 @@
...
-                elif provider.__is_json_content_type(config.content_type):
-                    try:
-                        j_object = json.loads(config.value)
-                        provider._dict[trimmed_key] = j_object
-                    except json.JSONDecodeError:
-                        # If the value is not a valid JSON, treat it like regular string value
-                        provider._dict[trimmed_key] = config.value
-                else:
+    for select in selects:
+        configurations = provider._client.list_configuration_settings(
```
xiangyan99, line 115: Is it by design to access private members?

sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/_azureappconfigurationprovider.py
```diff
@@ -3,209 +3,334 @@
...
+        for config in configurations:
+
+            trimmed_key = config.key
+            # Trim the key if it starts with one of the prefixes provided
+            for trim in provider._trim_prefixes:
+                if config.key.startswith(trim):
+                    trimmed_key = config.key[len(trim) :]
+                    break
+
+            if config.content_type == SecretReferenceConfigurationSetting._secret_reference_content_type:
```
xiangyan99, line 127: I believe here you can use isinstance(config, SecretReferenceConfigurationSetting) rather than accessing private string.

xiangyan99: We also exposed config.kind

sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/aio/_azureappconfigurationproviderasync.py
```diff
@@ -0,0 +1,293 @@
...
+            trimmed_key = config.key
+            # Trim the key if it starts with one of the prefixes provided
+            for trim in provider._trim_prefixes:
+                if config.key.startswith(trim):
+                    trimmed_key = config.key[len(trim) :]
+                    break
+
+            if config.content_type == SecretReferenceConfigurationSetting._secret_reference_content_type:
+                secret = await _resolve_keyvault_reference(config, key_vault_options, provider)
+                provider._dict[trimmed_key] = secret
```
xiangyan99, line 123: Isn't it provider.get(trimmed_key)? :)

sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/aio/_azureappconfigurationproviderasync.py
```diff
@@ -0,0 +1,293 @@
...
+            elif config.content_type == FeatureFlagConfigurationSetting._feature_flag_content_type:
+                feature_management = provider._dict.get(FEATURE_MANAGEMENT_KEY, {})
+                feature_management[trimmed_key] = config.value
+                if FEATURE_MANAGEMENT_KEY not in provider._dict.keys():
```
xiangyan99, line 127: And provider._dict.keys() could be provider.keys()?

sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/_azureappconfigurationprovider.py
mrm9084: Yes, it was a side affect of the move out of the provider class and into the module level method.

sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/_azureappconfigurationprovider.py
mrm9084: Content type is more important to me as the content type can change, and this is to support the specific content type.

sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/aio/_azureappconfigurationproviderasync.py
mrm9084: Can't do assignment with set?

sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/_azureappconfigurationprovider.py
xiangyan99: Could you give me an example, in which case, config.content_type == SecretReferenceConfigurationSetting._secret_reference_content_type and isinstance(config, SecretReferenceConfigurationSetting) are different?

sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/_azureappconfigurationprovider.py
```diff
@@ -3,209 +3,334 @@
...
+                secret = _resolve_keyvault_reference(config, key_vault_options, provider)
```
xiangyan99, line 128: What is the behavior if _resolve_keyvault_reference raises error? We want to continue or stop?

sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/aio/_azureappconfigurationproviderasync.py
xiangyan99: My bad. What I meants was given we want to set the value out of the AzureAppConfigurationProvider object, maybe we should also expose a set method for it instead of asking those code to directly access _dict? It seems to me this is why we want to expose get, right?

sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/_azureappconfigurationprovider.py
xiangyan99: If I understand correctly, if there is a ConfigSetting that is corrupt, user always gets failure when loading the provider.

How can they get it back to work? Or this is by design that they have to use some other ways (e.g. Explorer) to get the data fixed before they can use the provider?

mrm9084: Currently we stop processing with any error. The idea is we never want to return a partial configuration as they might be worse. Eventually we will add support for Geo-Replication, which will retry with that. Also, in theory this is application start up time.

sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/aio/_azureappconfigurationproviderasync.py
mrm9084: We don't what a public API for the customer modifying the configuration values.

sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/_azureappconfigurationprovider.py
mrm9084: Can we guarantee that the 1 to 1 relationship between user agent and class will always be true?

sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/aio/_azureappconfigurationproviderasync.py
xiangyan99: Looks to me the difference is app config sdk uses 

```py
if key_value.content_type.startswith(
                    SecretReferenceConfigurationSetting._secret_reference_content_type  # pylint:disable=protected-access
                ):
```
https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/appconfiguration/azure-appconfiguration/azure/appconfiguration/_models.py#L78
while you use 
```py
if config.content_type == SecretReferenceConfigurationSetting._secret_reference_content_type:
```

xiangyan99: Well. there are some other differences in case of error handling....

But still it seems to me you (the provider and SDK) want to do same thing.

My concern is if we parse same ConfigSetting differently in provider vs in SDK, it may cause problems.

sdk/appconfiguration/azure-appconfiguration-provider/azure/appconfiguration/provider/_azureappconfigurationprovider.py
xiangyan99: Took offline and we can discuss it in arch board review.

xiangyan99 (approved):

PR #28513, opened 2023-01-27
[Storage] Fix documentation for `timeout` in all Storage packages

### LINKED ISSUE #28506: Missing docs on timeout in azure-storage
- **Package Name**: azure-storage-blob
- **Package Version**: [12.14.1]
- **Operating System**: OSX
- **Python Version**: 3.8

**Describe the bug**
The docs about timeout parameters, for instance in

https://azuresdkdocs.blob.core.windows.net/$web/python/azure-storage-blob/12.14.1/azure.storage.blob.html#azure.storage.blob.BlobClient.upload_blob

does not describe the default value and what kind of behaviour to expect if left unset.

jalauzon-msft:
Resolves #28506.

Adjusts the documentation for the `timeout` argument to better explain that it's a server-side timeout and to link to further documentation on default values and maximum values. Applied to every operation in all packages.


vincenttran-msft (approved):
Thanks for getting to this. Much more helpful docstring 😄

Issue #26476, opened 2022-09-27
Create a workload identity credential for DAC auth flow


scottaddie:
**Intro:**
Azure AD workload identity federation is a new capability. It applies to both application identities and managed identities in Azure AD. The k8s team has built additional components around this that help customers use this capability in their k8s deployments. It's essential for the Azure Identity SDK to support this pattern so customers can easily benefit from this capability.

1. Allow customers to seamlessly move their code, which uses `DefaultAzureCredential`, to also run in a k8s cluster.
1. Do this in a way that isn't confusing to customers. They could either use application identities or managed identities. Telling a customer to pass a managed identity when they're using an application identity is going to be very confusing.
1. Provide predictability on the credential. For example, a customer may want to exclude `ManagedIdentityCredential`, so they know the request isn't going through the compute endpoint like IMDS.

**Proposal:**
1) Add a new credential type named `WorkloadIdentityCredential`. This new credential type will become part of the `DefaultAzureCredential` auth flow.

2) The `WorkloadIdentityCredential` will come into play if the necessary environment variables are populated: (`AZURE_FEDERATED_TOKEN_FILE`, `AZURE_CLIENT_ID`, `AZURE_TENANT_ID`). These are set up by the k8s platform based on the assertions on a service account.

3) Allow an optional client ID parameter to be passed into `DefaultAzureCredential`, which can be used by the customers to override the `AZURE_CLIENT_ID` in the event they have more than one identity for which they intend to use this flow.

With these changes, customers can do the following:

1. Use `DefaultAzureCredential` as their main guidance to their developers.
1. Recommend excluding the necessary credential types to get predictability in their code execution.
1. Pass explicit client IDs for which they want to get an Azure AD token.

Ideally with these changes, we'd also remove the current implementation where `ManagedIdentityCredential` is overloaded to work with these env. Doing this is ideal to avoid the ensuing confusion for customers.

xiangyan99:
Done

PR #28536, opened 2023-01-31
Added `WorkloadIdentityCredential`

xiangyan99:


sdk/identity/azure-identity/azure/identity/aio/_credentials/default.py
```diff
@@ -109,6 +113,13 @@ def __init__(self, **kwargs: Any) -> None:
         credentials = []  # type: List[AsyncTokenCredential]
         if not exclude_environment_credential:
             credentials.append(EnvironmentCredential(authority=authority, **kwargs))
+        if all(os.environ.get(var) for var in EnvironmentVariables.TOKEN_EXCHANGE_VARS):
+            client_id = managed_identity_client_id or os.environ.get(EnvironmentVariables.AZURE_CLIENT_ID)
```
chlowell, line 117: The argument for this, if any, should be `workload_identity_client_id`.

sdk/identity/azure-identity/azure/identity/_credentials/workload_identity.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def __init__(
+            self,
+            tenant_id: str,
+            client_id: str,
+            file: str,
+            **kwargs: Any
+    ) -> None:
+        kwargs.pop("token_file_path", None)
+        self._credential = TokenExchangeCredential(
```
chlowell, line 31: Would be simpler to replace `TokenExchangeCredential` instead? Keeping it seems to add an unnecessary layer to this credential.

sdk/identity/azure-identity/azure/identity/aio/_credentials/default.py
xiangyan99: managed_identity_client_id will be used in the managed identity route, right?

Shouldn't we keep the same behavior to honor it?

sdk/identity/azure-identity/azure/identity/aio/_credentials/default.py
```diff
@@ -109,6 +113,13 @@ def __init__(self, **kwargs: Any) -> None:
         credentials = []  # type: List[AsyncTokenCredential]
         if not exclude_environment_credential:
             credentials.append(EnvironmentCredential(authority=authority, **kwargs))
+        if all(os.environ.get(var) for var in EnvironmentVariables.TOKEN_EXCHANGE_VARS):
```
pvaneck, line 116: Perhaps rename TOKEN_EXCHANGE_VARS to WORKLOAD_IDENTITY_VARS?

sdk/identity/azure-identity/azure/identity/_credentials/managed_identity.py
```diff
@@ -64,17 +64,17 @@ def __init__(self, **kwargs: Any) -> None:
 
                 self._credential = CloudShellCredential(**kwargs)
         elif all(os.environ.get(var) for var in EnvironmentVariables.TOKEN_EXCHANGE_VARS):
```
pvaneck, line 66: Should we remove this check/implementation from `ManagedIdentityCredential` as outlined by @scottaddie  at the bottom of the issue described in https://github.com/Azure/azure-sdk-for-python/issues/26476?

xiangyan99: Removing it from MIC is a breaking change.

sdk/identity/azure-identity/azure/identity/aio/_credentials/default.py
chlowell: That's a good point! What do you think of adding `workload_identity_client_id` and preferring its value to `managed_identity_client_id`? It's ugly but avoids breaking applications and maintains the distinction between workload- and managed identity.

xiangyan99: Let me ask this question:

Customer has the code:

```py
cred = DAC(managed_identity_client_id="xxx")
```

after they upgrade the identity library, do we want to keep exactly the same behavior?

xiangyan99: I updated the code. How about this one? @chlowell

chlowell: That looks correct to me: users setting `managed_identity_client_id` get the same behavior from the new credential, new users will (we hope) set `workload_identity_client_id` instead, and we prefer the new option when someone sets both.

sdk/identity/azure-identity/azure/identity/_credentials/default.py
```diff
@@ -76,7 +80,7 @@ class DefaultAzureCredential(ChainedTokenCredential):
         Directory work or school accounts.
```
pvaneck, line 80: Add `:keyword  str workload_identity_client_id` somewhere in here And similarly in the aio version?

sdk/identity/azure-identity/azure/identity/_credentials/default.py
```diff
@@ -64,6 +68,8 @@ class DefaultAzureCredential(ChainedTokenCredential):
         AZURE_TENANT_ID, if any. If unspecified, users will authenticate in their home tenants.
     :keyword str managed_identity_client_id: The client ID of a user-assigned managed identity. Defaults to the value
         of the environment variable AZURE_CLIENT_ID, if any. If not specified, a system-assigned identity will be used.
+    :keyword str workload_identity_client_id: The client ID of a user-assigned managed identity. Defaults to the value
+        of the environment variable AZURE_CLIENT_ID, if any. If not specified, a system-assigned identity will be used.
```
chlowell, line 72: Workload Identity supports service principals, and doesn't support managed identities yet:
```suggestion
    :keyword str workload_identity_client_id: The client ID of an identity assigned to the pod. Defaults to the value
        of the environment variable AZURE_CLIENT_ID, if any. If not specified, the pod's default identity will be used.
```

sdk/identity/azure-identity/azure/identity/aio/_credentials/default.py
```diff
@@ -58,6 +62,8 @@ class DefaultAzureCredential(ChainedTokenCredential):
         **False**.
     :keyword str managed_identity_client_id: The client ID of a user-assigned managed identity. Defaults to the value
         of the environment variable AZURE_CLIENT_ID, if any. If not specified, a system-assigned identity will be used.
+    :keyword str workload_identity_client_id: The client ID of a user-assigned managed identity. Defaults to the value
+        of the environment variable AZURE_CLIENT_ID, if any. If not specified, a system-assigned identity will be used.
```
chlowell, line 66:
```suggestion
    :keyword str workload_identity_client_id: The client ID of an identity assigned to the pod. Defaults to the value
        of the environment variable AZURE_CLIENT_ID, if any. If not specified, the pod's default identity will be used.
```

chlowell (approved):

petergaultney:
did the SDK support Workload Identity in some other form before this PR was merged? I'm using 1.12 on an AKS cluster with Workload Identity enabled and it _seems_ to be working....  but now I see that the code for it apparently does not even exist in a released version?

xiangyan99:
@petergaultney Yes. You are absolutely right! :)

PR #28546, opened 2023-01-31
[EventHub] Use Streams API For Async Socket Connections

kashifkhan:
The library was raising the following error when asyncio debug mode was set up. 

`EventProcessor instance 'eafb1af4-b6e3-4877-8f2c-0a57f7e353fa' of eventhub 'ehtest' consumer group '$Default'. An error occurred while load-balancing and claiming ownership. The exception is EventHubError('the socket must be non-blocking\nthe socket must be non-blocking'). Retrying after 10.27072635958343 seconds`

This was happening because `_transport_async.py` was using a blocking socket and passing it into [`asyncio.get_event_loop.open_connection`](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.sock_connect) which expects a non-blocking socket. For a socket to be non-blocking the timeout has to be `0.0` and the code was using `None`. 

This socket was also then passed into the high-level streams API, which should be avoided. The recommended approach to create nonblocking sockets is to use the [high level streams API](https://docs.python.org/3/library/asyncio-stream.html)

While most of the changes were 1:1 there are some deltas that need to be discussed:
* `_connect` was using a happy eyeballs algo to determine which endpoint to connect to (IPv4 or IPv6). Replaced it by using the `happy_eyeballs_delay` with a value of .25 as specified in [`loop.create_connection`](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.create_connection)
  * Update  -- had to remove this as the param is only for py3.8+ 
* Passing in a family param of `socket.AF_UNSPEC` to handle both IPv4 and IPv6.
* Some of the other settings as keep_alive, tcp options etc. were removed since there is no access to the socket object created. I could go back to creating a socket object manually, keeping it non blocking , set all those options and then pass it in to `asyncio.open_connection`. The socket would have to be connected however.

### DIFF
--- a/sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_transport_async.py
+++ b/sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_transport_async.py
@@ -271,7 +271,6 @@ class AsyncTransport(
                 family=socket.AF_UNSPEC,
                 proto=SOL_TCP,
                 server_hostname=self.host if self.sslopts else None,
-                happy_eyeballs_delay=0.25,
             )
             # we've sent the banner; signal connect
             # EINTR, EAGAIN, EWOULDBLOCK would signal that the banner
@@ -361,7 +360,6 @@ class AsyncTransport(
                 # Sometimes SSL raises APPLICATION_DATA_AFTER_CLOSE_NOTIFY here on close.
                 _LOGGER.debug("Error shutting down socket: %r", e, extra=self.network_trace_params)
             self.writer, self.reader = None, None
-        self.sock = None
         self.connected = False
 
     async def negotiate(self):
--- a/sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_transport_async.py
+++ b/sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_transport_async.py
@@ -53,10 +53,8 @@ from .._transport import (
     AMQP_FRAME,
     get_errno,
     to_host_port,
-    DEFAULT_SOCKET_SETTINGS,
     SIGNED_INT_MAX,
     _UNAVAIL,
-    set_cloexec,
     AMQP_PORT,
     TIMEOUT_INTERVAL,
 )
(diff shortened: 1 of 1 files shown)


kashifkhan:
[Perf results](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=2151333&view=logs&j=42ded549-05ee-5cdb-7ba7-7a948a0cc056&t=73750c49-3b9e-58bb-db7f-f21818dca1f5)

kashifkhan:
Latest perf [results](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=2151916&view=logs&j=42ded549-05ee-5cdb-7ba7-7a948a0cc056&t=73750c49-3b9e-58bb-db7f-f21818dca1f5)

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_transport_async.py
```diff
@@ -270,106 +279,12 @@ async def connect(self):
...
-                    # we may have depleted all our options
-                    if i + 1 >= entries_num and n + 1 >= addr_types_num:
-                        raise
-                else:
-                    # hurray, we established connection
-                    return
-
-    def _init_socket(self, socket_settings):
-        self.sock.settimeout(None)  # set socket back to blocking mode
-        self.sock.setsockopt(socket.SOL_SOCKET, socket.SO_KEEPALIVE, 1)
```
annatisch, line 331: Do we have some kind of parity with these options?

sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/aio/_transport_async.py
```diff
@@ -270,106 +279,12 @@ async def connect(self):
...
-                elif hasattr(socket, opt):
-                    tcp_opts[enum] = sock.getsockopt(SOL_TCP, getattr(socket, opt))
-        return tcp_opts
-
-    def _set_socket_options(self, socket_settings):
-        tcp_opts = self._get_tcp_socket_defaults(self.sock)
-        if socket_settings:
-            tcp_opts.update(socket_settings)
-        for opt, val in tcp_opts.items():
-            self.sock.setsockopt(SOL_TCP, opt, val)
```
annatisch, line 371: omg it's so nice to see all this code cleaned out...

PR #28558, opened 2023-01-31
[ServiceBus] Iterator Support

l0lawrence:
Adds : 
keep_alive_thread (for releasing messages),
iterator support,
a lock for auto-lock renewing on mgmt_request so that one mgmt_link is created and used by all the worker threads when specifying max_workers > 2

Bug to Fix:

#28999
#29029


sdk/servicebus/azure-servicebus/azure/servicebus/_common/receiver_mixins.py
```diff
@@ -129,8 +129,18 @@ def _populate_message_properties(self, message):
...
     def _enhanced_message_received(self, frame, message):
         # pylint: disable=protected-access
-        self._handler._was_message_received = True
+        # self._handler._was_message_received = True
+        print(f"In enhanced {message.data}")
         if self._receive_context.is_set():
             self._handler._received_messages.put((frame, message))
         else:
             self._handler.settle_messages(frame[1], 'released')
+
```
l0lawrence, line 138: shouldnt have async in sync

sdk/servicebus/azure-servicebus/azure/servicebus/_pyamqp/aio/_client_async.py
```diff
@@ -265,21 +268,19 @@ async def open_async(self, connection=None):
         self._shutdown = False
         # TODO: Looks like this is broken - should re-enable later and test
```
l0lawrence, line 269: remove TODO

sdk/servicebus/azure-servicebus/azure/servicebus/_pyamqp/aio/_client_async.py
```diff
@@ -360,15 +364,16 @@ async def mgmt_request_async(self, message, **kwargs):
         operation_type = kwargs.pop("operation_type", None)
         node = kwargs.pop("node", "$management")
         timeout = kwargs.pop('timeout', 0)
-        try:
-            mgmt_link = self._mgmt_links[node]
-        except KeyError:
-            mgmt_link = ManagementOperation(self._session, endpoint=node, **kwargs)
-            self._mgmt_links[node] = mgmt_link
-            await mgmt_link.open()
+        with self._lock:
```
l0lawrence, line 367: lock to prevent context switching when trying to create a mgmt_link -- uamqp uses the 1 mgmt lock for all auto-lock renewers,pyamqp was trying (and failing) to make multiple

sdk/servicebus/azure-servicebus/azure/servicebus/aio/_servicebus_receiver_async.py
```diff
@@ -212,40 +212,38 @@ def __init__(
...
-                with receive_trace_context_manager(self.receiver, links=links):
-                    return message
+                with receive_trace_context_manager(self, links=links):
+                    yield message
+                    # print("Yielded")
+            except StopAsyncIteration:
+                print("Stop Iteration")
+                break
             finally:
-                self.receiver._receive_context.clear()
```
l0lawrence, line 240: I am wondering if this is why they had the 3 while loops because other wise on the async side you hit this finally block for the previous iterator once the current has started and it messes up the receive_context being set so you end up releasing messages you want to keep

l0lawrence: acutally I think this is a test-side issue, but I am wondering if on the iterator some exceptions from the tests sending received messages are being hidden

sdk/servicebus/azure-servicebus/azure/servicebus/_pyamqp/aio/_client_async.py
```diff
@@ -829,6 +833,58 @@ async def receive_message_batch_async(self, **kwargs):
...
+                    message = self._received_messages.get()
+                    self._last_activity_stamp = time.time()
+                    self._received_messages.task_done()
+                    yield message
+                    await self._complete_message_async(message, auto_complete)
+        finally:
+            if self._shutdown:
+                await self.close_async()
+
+    async def _complete_message_async(self, message, auto):
```
l0lawrence, line 881: never used but supported in uamqp

sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_receiver.py
```diff
@@ -218,7 +218,6 @@ def __init__(
             else ServiceBusSession(cast(str, self._session_id), self)
         )
         self._receive_context = threading.Event()
-        self._handler: ReceiveClientSync
```
l0lawrence, line 221: typing? (add back?)

sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_receiver.py
```diff
@@ -232,17 +231,18 @@ def _iter_contextual_wrapper(self, max_wait_time=None):
             # This is not threadsafe, but gives us a way to handle if someone passes
             # different max_wait_times to different iterators and uses them in concert.
             if max_wait_time:
-                # _timeout to _idle_timeout
-                original_timeout = self._handler._idle_timeout
-                self._handler._idle_timeout = max_wait_time * 1000
+                original_timeout = self._handler._timeout
+                self._handler._timeout = max_wait_time * 1
             try:
+                self._receive_context.set()
```
l0lawrence, line 237: can remove these receive_context here -- have in iter_next()

sdk/servicebus/azure-servicebus/azure/servicebus/aio/_servicebus_receiver_async.py
```diff
@@ -268,19 +260,23 @@ async def __anext__(self):
...
-            self._auto_lock_renewer
-            and not self._session
-            and self._receive_mode != ServiceBusReceiveMode.RECEIVE_AND_DELETE
-        ):
-            self._auto_lock_renewer.register(self, message)
-        return message
+        try:
+            self._receive_context.set()
+            await self._open()
+            # TODO: Add in Recieve Message Iterator
```
l0lawrence, line 266: remove todo

sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_receiver.py
```diff
@@ -417,7 +418,8 @@ def _receive(self, max_message_count=None, timeout=None):
             self._open()
 
             amqp_receive_client = self._handler
-            received_messages_queue = amqp_receive_client._received_messages
+            received_messages_queue = amqp_receive_client._received_messages # type: ignore[attr-defined]
+            amqp_receive_client._running_iter = False # type: ignore[attr-defined]
```
l0lawrence, line 422: This is something I am not sure about, I use this boolean to show if we are using the iterator or regular receive. If you are receiving normally within an iterator for loop and you sleep, the iterator will time out, however you are still receiving messages so that should not happen. This boolean is a way to reset the last_activity_stamp so that your iterator doesn't time out when receiving via the normal _receive() method. 

uamqp did this a different way, but my thought was that this would be less intrusive into the current design?

sdk/servicebus/azure-servicebus/azure/servicebus/_common/receiver_mixins.py
l0lawrence: Update: will be moved by switch to transport files

sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_receiver.py
l0lawrence: wont need type disables then

sdk/servicebus/azure-servicebus/tests/test_sessions.py
```diff
@@ -47,7 +47,7 @@
 
 
 class ServiceBusSessionTests(AzureMgmtTestCase):
-    @pytest.mark.skip(reason="TODO: iterator support")
+    @pytest.mark.skip(reason="TODO: Pyamqp Message Serialization Error")
```
annatisch, line 50: What kind of error was this?

sdk/servicebus/azure-servicebus/tests/test_queues.py
```diff
@@ -2646,17 +2647,16 @@ def hack_iter_next_mock_error(self):
                 self._open()
                 # when trying to receive the second message (execution_times is 1), raising LinkDetach error to mock 10 mins idle timeout
                 if self.execution_times == 1:
-                    from uamqp.errors import LinkDetach
-                    from uamqp.constants import ErrorCodes
+                    from azure.servicebus._pyamqp.error import ErrorCondition, AMQPConnectionError
                     self.execution_times += 1
                     self.error_raised = True
-                    raise LinkDetach(ErrorCodes.LinkDetachForced)
+                    raise AMQPConnectionError(condition=ErrorCondition.LinkDetachForced)
```
annatisch, line 2653: Don't we have a LinkDetach error in pyamqp?

sdk/servicebus/azure-servicebus/tests/test_message.py
```diff
@@ -304,7 +303,7 @@ def test_message_backcompat_receive_and_delete_databody(self, servicebus_namespa
             outgoing_message.message.reject()
         with pytest.raises(TypeError):
             outgoing_message.message.modify(True, True)
-        assert outgoing_message.message.state == uamqp.constants.MessageState.SendComplete
+        # assert outgoing_message.message.state == uamqp.constants.MessageState.SendComplete
```
annatisch, line 306: We probably should no longer make assertions using the .message attribute - as that was the leaked uamqp message (although once Swathi's switch PR is merged, we could probably uncomment these as it should be more-or-less backwards compatible.
But we should probably make a note to refactor these tests (if for no other reason than the test outputs are about to be flooded with deprecation warnings...)

sdk/servicebus/azure-servicebus/tests/async_tests/test_queues_async.py
```diff
@@ -238,16 +238,16 @@ async def sub_test_non_releasing_messages():
...
                 sender = sb_client.get_queue_sender(servicebus_queue.name)
 
-                def _hack_disable_receive_context_message_received(self, message):
-                    # pylint: disable=protected-access
-                    self._handler._was_message_received = True
-                    self._handler._received_messages.put(message)
+                # def _hack_disable_receive_context_message_received(self, message):
+                #     # pylint: disable=protected-access
+                #     self._handler._was_message_received = True
+                #     self._handler._received_messages.put(message)
```
annatisch, line 244: Not sure what this was originally testing and why it had to be removed?

sdk/servicebus/azure-servicebus/tests/async_tests/test_queues_async.py
```diff
@@ -1783,7 +1783,7 @@ async def test_queue_receive_keep_conn_alive_async(self, servicebus_namespace_co
                 assert len(messages) == 0  # make sure messages are removed from the queue
                 assert receiver_handler == receiver._handler  # make sure no reconnection happened
 
-    @pytest.mark.skip(reason="TODO: iterator support")
+    @pytest.mark.skip(reason="TODO: _counter doesnt exist in pyamqp")
```
annatisch, line 1786: What were we counting?

sdk/servicebus/azure-servicebus/tests/async_tests/test_queues_async.py
```diff
@@ -2277,7 +2280,7 @@ async def hack_iter_next_mock_error(self):
                 assert receiver.error_raised
                 assert receiver.execution_times >= 4  # at least 1 failure and 3 successful receiving iterator
 
-    @pytest.mark.skip(reason="TODO: iterator support")
+    @pytest.mark.skip(reason="TODO: Pyamqp Message Header Time to Live")
```
annatisch, line 2283: Is this just because the attribute name changed?

sdk/servicebus/azure-servicebus/azure/servicebus/_pyamqp/client.py
```diff
@@ -208,6 +210,11 @@ def __init__(self, hostname, **kwargs):
         self._custom_endpoint_address = kwargs.get("custom_endpoint_address")
         self._connection_verify = kwargs.get("connection_verify")
 
+        # Keep Alive (release messages)
+        self._keep_alive_interval = int(kwargs.get("keep_alive_interval",0))
+        self._keep_alive_thread = None
```
annatisch, line 215: This is already on line 168-169

sdk/servicebus/azure-servicebus/azure/servicebus/_pyamqp/client.py
```diff
@@ -339,6 +364,12 @@ def close(self):
         if not self._external_connection:
             self._connection.close()
             self._connection = None
+        if self._keep_alive_thread:
+            try:
+                self._keep_alive_thread.join()
+            except RuntimeError:  # Probably thread failed to start in .open()
+                logging.info("Keep alive thread failed to join.", exc_info=True)
```
annatisch, line 371: This should be debug logging

sdk/servicebus/azure-servicebus/azure/servicebus/_pyamqp/client.py
```diff
@@ -409,16 +440,16 @@ def mgmt_request(self, message, **kwargs):
...
-            mgmt_link = self._mgmt_links[node]
-        except KeyError:
-            mgmt_link = ManagementOperation(self._session, endpoint=node, **kwargs)
-            self._mgmt_links[node] = mgmt_link
-            mgmt_link.open()
-
-            while not mgmt_link.ready():
-                self._connection.listen(wait=False)
-
+        with self._lock:
```
annatisch, line 443: I know you've discussed it in our sync-ups - but can you expand on what you were observing that made this necessary?
I'm trying to figure out where the race condition was that we need to actually lock - is it `self._mgmt_links`? Or something on the connection?

sdk/servicebus/azure-servicebus/azure/servicebus/_pyamqp/client.py
```diff
@@ -762,6 +794,11 @@ def __init__(self, hostname, source, **kwargs):
         self._max_message_size = kwargs.pop("max_message_size", MAX_FRAME_SIZE_BYTES)
         self._link_properties = kwargs.pop("link_properties", None)
         self._link_credit = kwargs.pop("link_credit", 300)
+        self._timeout = kwargs.pop("timeout", 0)
+        self._timeout_reached = False
+        self._last_activity_stamp = time.time()
+        self._running_iter = False
+        self.auto_complete = kwargs.pop("auto_complete", True)
```
annatisch, line 801: We shouldn't need to add auto-complete do we?
I think that's a bit of uamqp hangover....

sdk/servicebus/azure-servicebus/azure/servicebus/_pyamqp/client.py
```diff
@@ -895,6 +934,58 @@ def receive_message_batch(self, **kwargs):
...
+                    yield message
+                    self._complete_message(message, auto_complete)
+
+        finally:
+            if self._shutdown:
+                self.close()
+
+    def _complete_message(self, message, auto): # pylint:disable=no-self-use
+        if not message or not auto:
+            return
```
annatisch, line 985: We don't need this do we?
I don't think ServiceBus ever needs us to complete a message on behalf of users, so I would just remove it entirely.

sdk/servicebus/azure-servicebus/azure/servicebus/_pyamqp/aio/_client_async.py
```diff
@@ -263,10 +267,9 @@ async def open_async(self, connection=None):
         self._network_trace_params["amqpConnection"] = self._connection._container_id
         self._network_trace_params["amqpSession"] = self._session.name
         self._shutdown = False
-        # TODO: Looks like this is broken - should re-enable later and test
-        # correct empty frame behaviour
-        # if self._keep_alive_interval:
-        #    self._keep_alive_thread = asyncio.ensure_future(self._keep_alive_async())
+
+        if self._keep_alive_interval:
+            self._keep_alive_thread = asyncio.ensure_future(self._keep_alive_async())
```
annatisch, line 272: Regarding the empty frame behaviour - have you actually seen empty frames in the amqp trace? I don't know if it's ever been tested....

sdk/servicebus/azure-servicebus/azure/servicebus/_pyamqp/aio/_client_async.py
annatisch: Yay let's get rid of it

sdk/servicebus/azure-servicebus/tests/test_message.py
swathipil: Since these are testing the LegacyMessage, I can refactor these tests in the switch PR so that we only call `.message` once.

sdk/servicebus/azure-servicebus/tests/async_tests/test_queues_async.py
swathipil: (I believe the handler._counter.get_current_ms() was uamqp's equivalent to time.time() in pyamqp.) I'll unskip/refactor this in the switch pr.

sdk/servicebus/azure-servicebus/azure/servicebus/_pyamqp/aio/_transport_async.py
```diff
@@ -326,7 +329,7 @@ async def _connect(self, host, port, timeout):
                     # hurray, we established connection
                     return
 
-    def _init_socket(self, socket_settings):
+    def _init_socket(self, socket_settings, read_timeout=1):
         self.sock.settimeout(None)  # set socket back to blocking mode
         self.sock.setsockopt(socket.SOL_SOCKET, socket.SO_KEEPALIVE, 1)
         self._set_socket_options(socket_settings)
```
kashifkhan, line 335: this entire section will be simplified if my PR to use high level streams API is what we end up going with. You don't get access to the underlying socket object, so setting time out explicitly wont be possible. Though Im unsure if it was ever being respected in the first place when the socket is passed in to asyncio.open_connection

sdk/servicebus/azure-servicebus/azure/servicebus/_pyamqp/aio/_client_async.py
```diff
@@ -226,6 +227,9 @@ async def open_async(self, connection=None):
         :type connection: ~pyamqp.aio.Connection
         """
         # pylint: disable=protected-access
+
+        self._lock_async = asyncio.Lock() # pylint:disable=attribute-defined-outside-init
```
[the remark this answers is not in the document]
l0lawrence: the async client doesn't have its own init and I didn't want to put asyncio in the sync client. Could make an async init?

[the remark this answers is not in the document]
kashifkhan: yes, will be good to move in to its own async init since the sync classes have their own lock and its defined in the base which async also uses

sdk/servicebus/azure-servicebus/azure/servicebus/_pyamqp/client.py
l0lawrence: So the test where I saw this threading context switching was in (sync) test_queue_by_queue_client_conn_str_receive_handler_with_autolockrenew where we set AutoLockRenewer max_workers to be 8. 

The flow of the autolockrenewing (at least what I have seen so far):

In AutoLockRenewer:   

The Main Thread initializes a dispatch worker in _init_workers(). In dispatch_worker() since we have more than one worker we use the ThreadPoolExecutor to create tasks for _auto_lock_renew_task(). If the message needs to be renewed we call renew_message_lock() with one of the threads created by the ThreadPoolExecutor. In this specific example 2 threads (call them 1 and 2) enter into this function one after another. 

In the Receiver:

Entering into the SB reciever, renew_message_lock() calls self._renew_locks(). These 2 threads, 1 and 2, enter this function one after another as well. They then both enter into self._mgmt_request_response_with_retry() which calls client.py.

In pyamqp:

In client.py mgmt_request, these two threads both enter one after another (print statements back to back so context switching). Since thread 1 and thread 2 enter back to back, Thread 1 hits a KeyError and tries to create a new mgmt_link. Switching to Thread 2, it also hits the KeyError as Thread1 has not had time to open() the link yet. For both of these we wait for the client.ready() on the mgmt_link and sleep for .5 seconds if it isn't ready.


(In print Statments) 

More threads go through this process all now using existing mgmt_link (but there are 2). Then you hit this error:

`Unexpected error occurred (ValueError('Connection not open.')). Handler shutting down.`
`AttributeError: 'ServiceBusReceivedMessage' object has no attribute 'renew_lock'`
`Connection closed with error: [b'amqp:not-found', b"The session channel '1' cannot be found in connection 'Connection'.", None]
Unexpected error occurred (ValueError('Management Operation error occurred.')). Handler shutting down.`

and also in the stacktrace: 

` while not self._handler.client_ready():
AttributeError: 'NoneType' object has no attribute 'client_ready'`

So then I went and compared this against uamqp. While pyamqp initially creates 2 mgmt_links because the one thread doesn't recognize the other mgmt_link being made in time, uamqp uses the same mgmt link for all renew lock operations. By using the lock in pyamqp, we ended up with one exisitng mgmt_link.

sdk/servicebus/azure-servicebus/tests/async_tests/test_queues_async.py
l0lawrence: This was originally testing overloading the receive_mixins() enhanced_receive_message() function that releases functions if the receive_context is not set. This function just no longer released messages so you were testing that trying to complete expired messages would error, and then trying to re-receive them after with an increased delivery count . 

I commented this out because it did not seem like lines 249-250 were actually resetting the value of on_message_received and so messages were still being released.

sdk/servicebus/azure-servicebus/tests/async_tests/test_queues_async.py
l0lawrence: Yes let me fix and uncomment

sdk/servicebus/azure-servicebus/tests/test_sessions.py
l0lawrence: `Failed: [undefined]azure.servicebus.exceptions.MessageSizeExceededError: Serialization operation failed due to unsupported type System.Byte[]. `

l0lawrence: I was thinking this was related to trying to send a message that we received/peeked and the formatting being off slightly

sdk/servicebus/azure-servicebus/azure/servicebus/_pyamqp/aio/_client_async.py
l0lawrence: When running a keep receiver alive check, I am not seeing any empty frames coming in across the wire or going out

sdk/servicebus/azure-servicebus/tests/test_queues.py
l0lawrence: (Sorry missed this one) I have only seen an error condition for linkDetached

sdk/servicebus/azure-servicebus/tests/test_sessions.py
annatisch: That looks like a bug - we shouldn't skip this.
If we don't want to deal with it in this PR, please open an issue for it and put the issue number in the TODO comment.

sdk/servicebus/azure-servicebus/tests/test_queues.py
annatisch: We're both half right - there's no LinkDetactError, but we do have a link error ;)
https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/eventhub/azure-eventhub/azure/eventhub/_pyamqp/error.py#L257

sdk/servicebus/azure-servicebus/azure/servicebus/_pyamqp/aio/_client_async.py
```diff
@@ -360,15 +364,16 @@ async def mgmt_request_async(self, message, **kwargs):
...
+                mgmt_link = self._mgmt_links[node]
+            except KeyError:
+                mgmt_link = ManagementOperation(self._session, endpoint=node, **kwargs)
+                self._mgmt_links[node] = mgmt_link
+                await mgmt_link.open()
 
-            while not await mgmt_link.ready():
-                await self._connection.listen(wait=False)
+                while not await mgmt_link.ready():
+                    await self._connection.listen(wait=False)
```
annatisch, line 376: I don't think these two lines need to be inside the lock - we should only need to lock access to `self._mgmt_links`

sdk/servicebus/azure-servicebus/azure/servicebus/_pyamqp/aio/_client_async.py
```diff
@@ -726,6 +731,7 @@ async def _client_run_async(self, **kwargs):
         try:
             await self._link.flow()
             await self._connection.listen(wait=self._socket_timeout, **kwargs)
+            self._timeout_reached = False
```
annatisch, line 734: Where is this used?

sdk/servicebus/azure-servicebus/azure/servicebus/_pyamqp/aio/_connection_async.py
```diff
@@ -236,7 +236,7 @@ async def _read_frame(self, wait: Union[bool, int, float] = True, **kwargs) -> b
         """
         timeout: Optional[Union[int, float]] = None
         if wait is False:
-            timeout = 1  # TODO: What should this default be?
+            timeout = .2  # TODO: What should this default be?, are the other if statements are used
```
annatisch, line 239: What prompted this value?

sdk/servicebus/azure-servicebus/azure/servicebus/_pyamqp/aio/_transport_async.py
```diff
@@ -246,6 +248,7 @@ def __init__(
         self.host, self.port = to_host_port(host, port)
 
         self.connect_timeout = connect_timeout
+        self.read_timeout = read_timeout or READ_TIMEOUT_INTERVAL #TODO: read timeout like sync transport
```
annatisch, line 251: We don't use read timeouts in async the same as sync - instead we wrap the read call in an asyncio.wait with a timeout and handle the failure there. So we shouldn't need to configure this on the socket.

sdk/servicebus/azure-servicebus/azure/servicebus/_pyamqp/client.py
```diff
@@ -164,8 +165,9 @@ def __init__(self, hostname, **kwargs):
         self._auth_timeout = kwargs.pop("auth_timeout", DEFAULT_AUTH_TIMEOUT)
         self._mgmt_links = {}
         self._retry_policy = kwargs.pop("retry_policy", RetryPolicy())
-        self._keep_alive_interval = int(kwargs.get("keep_alive_interval") or 0)
+        self._keep_alive_interval = int(kwargs.get("keep_alive_interval", 0))
         self._keep_alive_thread = None
+        self._lock = threading.Lock()
```
annatisch, line 170: Could we maybe name this something more descriptive?
Like `_mgmt_link_lock`, and move it up next to `self._mgmt_links={}` just for readability purposes :)

sdk/servicebus/azure-servicebus/azure/servicebus/_pyamqp/client.py
annatisch: This was very helpful - thank you!
While it's possible we could do the locking somewhere else - I think this is a good first step towards making the clients more threadsafe.

sdk/servicebus/azure-servicebus/azure/servicebus/_pyamqp/client.py
```diff
@@ -409,16 +436,16 @@ def mgmt_request(self, message, **kwargs):
...
+        with self._lock:
+            try:
+                mgmt_link = self._mgmt_links[node]
+            except KeyError:
+                mgmt_link = ManagementOperation(self._session, endpoint=node, **kwargs)
+                self._mgmt_links[node] = mgmt_link
+                mgmt_link.open()
+
+                while not mgmt_link.ready():
+                    self._connection.listen(wait=False)
```
annatisch, line 448: As mentioned for sync - I don't think these two lines need to be locked.

sdk/servicebus/azure-servicebus/azure/servicebus/_pyamqp/client.py
```diff
@@ -762,6 +790,10 @@ def __init__(self, hostname, source, **kwargs):
         self._max_message_size = kwargs.pop("max_message_size", MAX_FRAME_SIZE_BYTES)
         self._link_properties = kwargs.pop("link_properties", None)
         self._link_credit = kwargs.pop("link_credit", 300)
+        self._timeout = kwargs.pop("timeout", 0)
+        self._timeout_reached = False
+        self._last_activity_stamp = time.time()
+        self._running_iter = False
```
annatisch, line 796: Are all these new properties _just_ used for the iterator?
Or do the other receive methods need to respect them as well?
If it's the former - then I think we should find a way to isolate these a bit...

sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_receiver.py
```diff
@@ -232,9 +232,8 @@ def _iter_contextual_wrapper(self, max_wait_time=None):
             # This is not threadsafe, but gives us a way to handle if someone passes
             # different max_wait_times to different iterators and uses them in concert.
             if max_wait_time:
-                # _timeout to _idle_timeout
-                original_timeout = self._handler._idle_timeout
-                self._handler._idle_timeout = max_wait_time * 1000
+                original_timeout = self._handler._timeout
+                self._handler._timeout = max_wait_time * 1
```
annatisch, line 236: This looks like it was a uamqp hack - is there a way we can handle this more naturally?

sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_receiver.py
annatisch: In pyamqp, we can prevent different receive methods being used together without needing boolean state flags (e.g. locks etc). Would that be useful?
If would make the pyamqp code safer/more resilient - however I'm not sure if it would break back compat with uamqp SB - I can't recall if that supported simultaneously receiving different ways

sdk/servicebus/azure-servicebus/azure/servicebus/_common/receiver_mixins.py
```diff
@@ -129,8 +129,16 @@ def _populate_message_properties(self, message):
 
     def _enhanced_message_received(self, frame, message):
         # pylint: disable=protected-access
-        self._handler._was_message_received = True
+        # self._handler._was_message_received = True
```
annatisch, line 132: If this is no longer needed in the pyamqp client (it used to be used for batch optimization I think, do we do that differently now?), you can remove it. Looks like it snuck into the async method below as well.

sdk/servicebus/azure-servicebus/tests/async_tests/test_queues_async.py
annatisch: hmm so the test was always broken?

sdk/servicebus/azure-servicebus/tests/async_tests/test_queues_async.py
annatisch: We should be able to update the test in this PR - it's just swapping the method call for time.time()

sdk/servicebus/azure-servicebus/azure/servicebus/_pyamqp/aio/_client_async.py
kashifkhan: I have seen empty frames come across in EH when the service wants to let the client know there aren't any messages for it to send. I would think the keep alive thread would ignore it completely and let the thread receiving it deal with it.

sdk/servicebus/azure-servicebus/azure/servicebus/_pyamqp/aio/_connection_async.py
l0lawrence: This was a change made to align the read timeout on the async side to the read timeout on the sync side that had previously been made before Event Hubs went GA. I was seeing receive statements come in at 1 second increments on the async side (significantly slower than uamqp) and after debugging with Swathi and Kashif we realized that sync and async didn't align.

sdk/servicebus/azure-servicebus/azure/servicebus/_pyamqp/aio/_client_async.py
```diff
@@ -741,6 +749,7 @@ async def _message_received_async(self, frame, message):
         :param message: Received message.
         :type message: ~pyamqp.message.Message
         """
+        self._last_activity_timestamp = time.time()
```
l0lawrence, line 752: on_receive on a message means there was activity -- this is used to measure timeout

sdk/servicebus/azure-servicebus/azure/servicebus/_common/receiver_mixins.py
```diff
@@ -129,8 +129,16 @@ def _populate_message_properties(self, message):
 
     def _enhanced_message_received(self, frame, message):
         # pylint: disable=protected-access
-        self._handler._was_message_received = True
+        self._handler._last_activity_timestamp = time.time()
```
l0lawrence, line 132: on_message receive callback for link if prefetch is 1

sdk/servicebus/azure-servicebus/azure/servicebus/_pyamqp/aio/_connection_async.py
```diff
@@ -236,7 +236,7 @@ async def _read_frame(self, wait: Union[bool, int, float] = True, **kwargs) -> b
         """
         timeout: Optional[Union[int, float]] = None
         if wait is False:
-            timeout = 1  # TODO: What should this default be?
+            timeout = .2
```
annatisch, line 239: Weren't we going to move this to a constant?

sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_receiver.py
```diff
@@ -226,35 +226,21 @@ def __iter__(self):
...
-
-    def _inner_next(self):
+
+    def _inner_next(self, wait_time=None):
         # We do this weird wrapping such that an imperitive next() call, and a generator-based iter both trace sanely.
         self._check_live()
         while True:
             try:
-                return self._do_retryable_operation(self._iter_next)
+                return self._do_retryable_operation(self._iter_next, wait_time=wait_time)
```
annatisch, line 243: Is this going to wait the "max time" for every retry attempt?
The wait time a customer passes in needs to be the cumulative total, i.e. _actually_ the longest they can expect to wait.

annatisch: And there should probably be a test for the timing if there's not one already.

sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_receiver.py
```diff
@@ -272,15 +258,16 @@ def __next__(self):
...
             self._open()
-            # TODO: Add in Recieve Message Iterator
-            # if not self._message_iter:
-            #     self._message_iter = self._handler.receive_messages_iter()
-            uamqp_message = next(self._message_iter)
-            message = self._build_message(uamqp_message)
+            if not self._message_iter:
+                self._message_iter = self._handler.receive_messages_iter(timeout=wait_time)
+            elif wait_time:
+                self._message_iter = self._handler.receive_messages_iter(timeout=wait_time)
```
annatisch, line 268: These seem to be doing the same thing - couldn't they be merged into the same if statement?

sdk/servicebus/azure-servicebus/tests/async_tests/test_queues_async.py
```diff
@@ -49,15 +49,15 @@
...
+from tests.servicebus_preparer import CachedServiceBusNamespacePreparer, CachedServiceBusQueuePreparer, ServiceBusQueuePreparer
+from tests.utilities import get_logger, print_message, sleep_until_expired
 from mocks_async import MockReceivedMessage, MockReceiver
 
 _logger = get_logger(logging.DEBUG)
 
 
 class ServiceBusQueueAsyncTests(AzureMgmtTestCase):
-    @pytest.mark.skip(reason="TODO: iterator support")
+    #@pytest.mark.skip(reason="TODO: iterator support")
```
annatisch, line 60: We can delete these now right? :)

sdk/servicebus/azure-servicebus/tests/test_queues.py
```diff
@@ -1940,14 +1941,14 @@ def message_content():
...
-                        message_id='mid',
-                        to='to',
-                        reply_to='reply_to',
-                        time_to_live=timedelta(seconds=60)
+                        # content_type='application/text',
+                        # correlation_id='cid',
+                        # message_id='mid',
+                        # to='to',
+                        # reply_to='reply_to',
+                        # time_to_live=timedelta(seconds=60)
```
annatisch, line 1951: Do we have some way to track these commented tests after the pyamqp/uamqp switch work?

annatisch (approved):

sdk/servicebus/azure-servicebus/azure/servicebus/_pyamqp/aio/_connection_async.py
l0lawrence: yes!

sdk/servicebus/azure-servicebus/azure/servicebus/_servicebus_receiver.py
l0lawrence: This is just passing the max_wait_time for the user can set for the iterator through to the _iter_next function, it should not impact the timeout property that the user sets for retries via kwargs

l0lawrence: We also have this test here: test_queue_receiver_respects_max_wait_time_overrides to test max_wait_time on the iterator!

sdk/servicebus/azure-servicebus/tests/test_queues.py
l0lawrence: I can make an issue for them

PR #28561, opened 2023-02-01
[Storage] Fix Datalake rename operations when new name contains `?`

### LINKED ISSUE #28482: DataLakeFileClient rename_file with special characters does not create or error
- **Package Name**: azure-storage-file-datalake
- **Package Version**: 12.4.0
- **Operating System**: MacOS 11.7.2
- **Python Version**: 3.7.9

**Describe the bug**
When using the `DataLakeFileClient.rename_file()` with special characters in it, it successfully returns a DataLakeFileClient that is the encompassing directory, and no file was actually renamed.

**To Reproduce**
1. create a `DataLakeFileClient`
2. rename_file with special characters and display the returned `DataLakeFileClient.get_file_properties`
```python
file_client = file_system_client.get_file_client('this-is-a-directory/this-is-a-file')
new_file_client = file_client.rename_file('filesystem/this-is-a-directory/??!!~~~~***   (┛ಠ_ಠ)┛彡┻━┻     ***~~~~!!??')
print(new_file_client.get_file_properties())
```

**Expected behavior**
I would expect an error about an InvalidURI or similar thrown for the provided new_name path since the file wasn't actually successfully renamed.

**Additional context**
Example file properties returned which was actually the directory the file was originally in:
```
{
  'name': 'this-is-a-directory',
  'etag': '"0x8DAFF0A9B5D5B96"',
  'deleted': False, 
  'metadata': {'hdi_isfolder': 'true'},
  ...
}
```

jalauzon-msft:
Resolves #28482.

This change resolves an issue where `rename_file` and `rename_directory` in the Datalake SDK would not rename to the correct name if the new path contained a `?`. This was due to how we parsed the new path to account for the presence of a SAS token. The solution was to check for the presence of `sv=XXXX-XX-XX` and `sig=` in the path segment which could contain a SAS before removing it from the new path and using it as a SAS token. This is not a perfect solution but should realistically prevent any issues with renaming in the future without needing to change the public interface for how we accept a SAS token.

Also took this opportunity to move path parsing logic to common base class function.

### DIFF
--- a/.vscode/cspell.json
+++ b/.vscode/cspell.json
@@ -736,7 +736,8 @@
         "cpksubdirectory",
         "thead",
         "tbody",
-        "racwdlmeop"
+        "racwdlmeop",
+        "newfs"
       ]
     },
     {
--- a/sdk/storage/azure-storage-file-datalake/CHANGELOG.md
+++ b/sdk/storage/azure-storage-file-datalake/CHANGELOG.md
@@ -7,6 +7,10 @@
 - Added ability to perform leasing actions on file append and flush. See new keyword `lease_action` for details.
 - Added support for `AsyncIterable` as data type for async file upload.
 
+### Bugs Fixed
+- Fixed an issue where `rename_file` and `rename_directory` would not work correctly if the new file/directory name
+contained a `?` character.
+
 ### Other Changes
 - Removed `msrest` dependency.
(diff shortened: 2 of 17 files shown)


vincenttran-msft (approved):
Thanks for getting to this! Glad we were able to find the root cause (and come up with a decently robust solution) 😄

